PR #1940, opened 2017-06-20
HHH-11773 - Add support for JDBC 4.2 type names

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-11773

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -196,7 +196,9 @@ public abstract class Dialect implements ConversionContext {
 
 		registerColumnType( Types.DATE, "date" );
 		registerColumnType( Types.TIME, "time" );
+		registerColumnType( Types.TIME_WITH_TIMEZONE, getTimeWithTimeZone() );
 		registerColumnType( Types.TIMESTAMP, "timestamp" );
+		registerColumnType( Types.TIMESTAMP_WITH_TIMEZONE, getTimestampWithTimeZone() );
 
 		registerColumnType( Types.VARBINARY, "bit varying($l)" );
 		registerColumnType( Types.LONGVARBINARY, "bit varying($l)" );
@@ -2728,6 +2730,40 @@ public abstract class Dialect implements ConversionContext {
 	 */
 	public String getNotExpression(String expression) {
 		return "not " + expression;
+	}
+	
+	/**
+	 * Gets the name of the {@code TIME WITH TIME ZONE} type for this
+	 * dialect.
+	 * <p>
+	 * If the dialect supports {@code TIME WITH TIME ZONE} this method
+	 * should return the name of that type. If the dialect does not
+	 * support {@code TIME WITH TIME ZONE} it should return the name of
+	 * the {@code TIME} type.
+	 * <p>
+	 * Defaults to {@code "time"}.
+	 * 
+	 * @return the type name
+	 */
+	protected String getTimeWithTimeZone() {
+		return "time";
+	}
+
+	/**
+	 * Gets the name of the {@code TIMESTAMP WITH TIME ZONE} type for this
+	 * dialect.
+	 * <p>
+	 * If the dialect supports {@code TIMESTAMP WITH TIME ZONE} this method
+	 * should return the name of that type. If the dialect does not
+	 * support {@code TIMESTAMP WITH TIME ZONE} it should return the name of
+	 * the {@code TIMESTAMP} type.
+	 * <p>
+	 * Defaults to {@code "timestamp"}.
+	 * 
+	 * @return the type name
+	 */
+	protected String getTimestampWithTimeZone() {
+		return "timestamp";
 	}
 
 	/**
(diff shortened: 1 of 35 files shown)


marschall:
Regarding database support in general you need up to date database drivers. To my knowledge the current situation is like this:
- HSQLDB 2.4.0 supports it but has issues with literal parsing, this is fixed in 2.4.1 (not yet released)
- Oracle needs ojdbc8-12.2.0.1
- SQL Server as of 6.1.0.jre8 only supports `#setObject(int, Object)` but not `#getObject(int, Class)`, however you can use [proprietary APIs](https://static.javadoc.io/com.microsoft.sqlserver/mssql-jdbc/6.1.7.jre8-preview/com/microsoft/sqlserver/jdbc/ISQLServerResultSet.html#getDateTimeOffset-int-)
- H2 needs 1.4.194 (1.4.193 works in theory as well)
- Postgres needs pgjdbc 9.4.1208 or later

This all comes from [this bug](https://github.com/marschall/threeten-jpa/issues/41) in [this project](https://github.com/marschall/threeten-jpa/tree/master/threeten-jpa-jdbc42-hibernate).

To be fair I only use `#setObject(int, Object)` and not `#setObject(int, Object, int)` there, I believe that's where at least the H2 failures stem from.

marschall:
I created https://github.com/h2database/h2database/pull/536 to fix the issue in H2.

marschall:
pgjdbc should definitively support it https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/jdbc/PgPreparedStatement.java#L637

vladmihalcea:
We use the following drivers:

- PostgreSQL 42.1.1 > 9.4.1208
- H2 1.4.196 > 1.4.194
- SQL Server 6.1.0.jre8, I added the datetime specific logic using Reflection
- Oracle ojdbc7 12.1.0.2, I should upgrade to ojdbc8-12.2.0.1

I'll also try with `#setObject(int, Object)` as you suggested.

vladmihalcea:
I tried with `#setObject(int, Object)`, but it only works better for PostgreSQL, and just for handling `Timestamp`, not `Time`.

I'll leave this PR open for a while hoping that JDBC drivers will better handle the new JDBC 4.2 timestamp and time with timezone types.

marschall:
I understand this decision. There exist work arounds for people who want to experiment.

Probably in time for the JPA 2.1 service release that's part of Java EE 8 this will have to be revisited, let's hope the situation improves until then. Speaking of which right now Hibernate implements support for `LocalDateTime` and `LocalTime` by converting through `java.sql` types. This is not optimal for two reasons
1. `LocalTime` has nanosecond resolution `java.sql.Time` has millisecond resolution, converting results in silent data truncation.
1. `java.sql.Timestamp` due to implementation details is affected by daylight savings time even though it shouldn't be. If you run your JVM in a time zone that has daylight savings time the following value can not be stored or loaded correctly when a conversion through `java.sql.Timestamp` happens. It can only be stored and loaded correctly when driver support for JSR-310 is used.
```java

  private static LocalDateTime getUnstorableValue() {
    ZoneId systemTimezone = ZoneId.systemDefault();
    Instant now = Instant.now();

    ZoneRules rules = systemTimezone.getRules();
    ZoneOffsetTransition transition = rules.nextTransition(now);
    assertNotNull(transition);
    if (!transition.getDateTimeBefore().isBefore(transition.getDateTimeAfter())) {
      transition = rules.nextTransition(transition.getInstant().plusSeconds(1L));
      assertNotNull(transition);
    }

    Duration gap = Duration.between(transition.getDateTimeBefore(), transition.getDateTimeAfter());
    return transition.getDateTimeBefore().plus(gap.dividedBy(2L));
  }
```



Ultimately somebody has to make the first move, if drivers are waiting for frameworks to move first and frameworks are waiting for drivers to move first we end up in a chicken and egg situation see https://github.com/pgjdbc/pgjdbc/pull/695

Yes `OffsetTime` support has not yet been implemented in pgjdbc.

vladmihalcea:
@dreab8 Andrea, check out the comment above about `LocalTime` and `Timestamp`. I guess we should open a Jira issue about it, right?

dreab8:
Hi @vladmihalcea,
I have not a great experience on this topic, better to ask @sebersole for an opinion.

vladmihalcea:
@dreab8 @sebersole I created [this Jira issue for those findings](https://hibernate.atlassian.net/browse/HHH-11835).

abcfy2:
Any update for this pull request ? I think support `Types.TIMESTAMP_WITH_TIMEZONE` is very useful when I use `OffsetDateTime` and `ZonedDateTime`, but I find this pull request is still not merge.

Is hibernate still plan to support JDBC 4.2 type names ?

marschall:
@abcfy2 until fully supported you may find the threeten-jpa-jdbc42-hibernate subproject of https://github.com/marschall/threeten-jpa to be helpful. Pleas note that:

- this requires driver support
- this does not add support for DDL
- `ZonedDateTime` only really works with Oracle

marschall:
Was fixed in https://github.com/hibernate/hibernate-orm/commit/2c80250b0e0c13fea874458bd0e59f14d1493e25 in 6.0.0. Can be closed.

PR #1941, opened 2017-06-20
HHH-11830 - test case for TransactionObserver leak

ShawnClowater:
just a simple test showing that even after session is closed it is still retained in the transaction observer list.


vladmihalcea:
Check the [comments on the Jira issue](https://hibernate.atlassian.net/browse/HHH-11830?focusedCommentId=96401). I don't understand how this can be a problem since there are many Objects confined to a `Session` but they should all be GC-ed together.

vladmihalcea:
I created a [new PR](https://github.com/hibernate/hibernate-orm/pull/1947) which includes your test case and a fix as well.

PR #1943, opened 2017-06-22
[Task #55588] Schema update error message on startup for sequences in Derby

nikolovmiroslav:
DerbyDialect does not implement the method getQuerySequencesString(). 
This produces errors on startup, because Hibernate could not find the sequences and try to create them, but they already exist.

INFO  [DatabaseMetadata] table not found: XXXX
ERROR [SchemaUpdate] Unsuccessful: create sequence XXXX start with 1 increment by 1
ERROR [SchemaUpdate] Sequence 'XXXX' already exists.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -111,7 +111,7 @@ subprojects { subProject ->
     defaultTasks 'build'
 
     group = 'com.seeburger.fork'
-    version = '4.0.1.SEE5'
+    version = '4.0.1.SEE6'
 
     // minimize changes, at least for now (gradle uses 'build' by default)..
     buildDir = "target"
--- a/build.gradle
+++ b/build.gradle
@@ -110,8 +110,8 @@ subprojects { subProject ->
 
     defaultTasks 'build'
 
-    group = 'org.hibernate'
-    version = '4.0.1.Final'
+    group = 'com.seeburger.fork'
+    version = '4.0.1.SEE1'
 
     // minimize changes, at least for now (gradle uses 'build' by default)..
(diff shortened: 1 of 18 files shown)


build.gradle
```diff
@@ -106,12 +106,12 @@ libraries = [
...
 subprojects { subProject ->
     apply plugin: 'idea'
-	apply plugin: 'eclipse'
+    apply plugin: 'eclipse'
 
     defaultTasks 'build'
 
-    group = 'org.hibernate'
-    version = '4.0.1.Final'
+    group = 'com.seeburger.fork'
```
vladmihalcea: This file does not need to be included for thiss commit.

hibernate-core/src/main/java/org/hibernate/dialect/Oracle10gDialect.java
```diff
@@ -22,26 +22,50 @@
...
+ * of the <tt>oracle.jdbc.driver</tt> package in favor of
  * <tt>oracle.jdbc</tt>.
  *
  * @author Steve Ebersole
  */
 public class Oracle10gDialect extends Oracle9iDialect {
 
 	public Oracle10gDialect() {
-		super();
+	    super();
```
vladmihalcea: Why did you change the Oracle Dialect when this issue description is about Derby?

vladmihalcea:
This issue has a list of various commits that don't follow the original comment abaout Derby.

Please create a Jira issue, proviide a replicating test case, and a fix that should address the issue alone.

PR #1944, opened 2017-06-22
HHH-11808 - Update migration guide and documentation

dreab8:
https://hibernate.atlassian.net/browse/HHH-11808

this PR targets master branch.


vladmihalcea:
Applied upstream, thanks

PR #1945, opened 2017-06-25
HHH-11838 - Id retrieving from proxy with FK leads to query execution

igor-dmitriev:
## **Description of changes**
Fix - https://hibernate.atlassian.net/browse/HHH-11838
This issue is reproducible for field-based mapping, for property-based everything works fine.
https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/proxy/pojo/BasicLazyInitializer.java#L65

for ``property`` ``getIdentifierMethod`` is not null
for ``field`` ``getIdentifierMethod`` is null

## **Value being added**
getId on proxy will not load proxy (won't trigger sql query to fetch entity from db)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/property/access/internal/PropertyAccessFieldImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/property/access/internal/PropertyAccessFieldImpl.java
@@ -7,6 +7,7 @@
 package org.hibernate.property.access.internal;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 
 import org.hibernate.internal.util.ReflectHelper;
 import org.hibernate.property.access.spi.Getter;
@@ -32,8 +33,10 @@ public class PropertyAccessFieldImpl implements PropertyAccess {
 		this.strategy = strategy;
 
 		final Field field = ReflectHelper.findField( containerJavaType, propertyName );
-		this.getter = new GetterFieldImpl( containerJavaType, propertyName, field );
-		this.setter = new SetterFieldImpl( containerJavaType, propertyName, field );
+		final Method getterMethod = ReflectHelper.getterMethodOrNull(containerJavaType, propertyName);
+		final Method setterMethod = getterMethod != null ? ReflectHelper.setterMethodOrNull(containerJavaType, propertyName, getterMethod.getReturnType()) : null;
+		this.getter = new GetterFieldImpl( containerJavaType, propertyName, field, getterMethod );
+		this.setter = new SetterFieldImpl( containerJavaType, propertyName, field, setterMethod );
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/property/access/spi/GetterFieldImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/property/access/spi/GetterFieldImpl.java
@@ -26,11 +26,13 @@ public class GetterFieldImpl implements Getter {
 	private final Class containerClass;
 	private final String propertyName;
 	private final Field field;
+	private final Method getterMethod;
 
-	public GetterFieldImpl(Class containerClass, String propertyName, Field field) {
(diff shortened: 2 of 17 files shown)


hibernate-core/src/main/java/org/hibernate/property/access/spi/GetterFieldImpl.java
```diff
@@ -98,30 +100,30 @@ public Member getMember() {
...
 	public String getMethodName() {
-		return null;
+		return getterMethod != null ? getterMethod.getName() : null;
 	}
 
 	@Override
 	public Method getMethod() {
-		return null;
+		return getterMethod;
 	}
```
igor-dmitriev: I've checked the usage:
1.  getMethod is used just for ``Modifier.isFinal( method.getModifiers() )`` 
2. getMethodName -  ``getterNames[0] = getter.getMethodName();``

hibernate-core/src/main/java/org/hibernate/property/access/internal/PropertyAccessFieldImpl.java
```diff
@@ -32,8 +33,10 @@ public PropertyAccessFieldImpl(
 		this.strategy = strategy;
 
 		final Field field = ReflectHelper.findField( containerJavaType, propertyName );
-		this.getter = new GetterFieldImpl( containerJavaType, propertyName, field );
-		this.setter = new SetterFieldImpl( containerJavaType, propertyName, field );
+		final Method getterMethod = ReflectHelper.getterMethodOrNull(containerJavaType, propertyName);
+		final Method setterMethod = getterMethod != null ? ReflectHelper.setterMethodOrNull(containerJavaType, propertyName, getterMethod.getReturnType()) : null;
+		this.getter = new GetterFieldImpl( containerJavaType, propertyName, field, getterMethod );
+		this.setter = new SetterFieldImpl( containerJavaType, propertyName, field, setterMethod );
 	}
```
igor-dmitriev: get method could be ``null`` if getter for field is not provided

hibernate-core/src/test/java/org/hibernate/test/lazyload/LazyLoadingTest.java
```diff
@@ -100,4 +105,58 @@ public void testLazyCollectionLoadingAfterEndTransaction() {
...
+			s.save(address);
+			Client client = new Client(address);
+			return s.save(client);
+		});
+
+		Serializable addressId = doInHibernate(this::sessionFactory, s -> {
+			Client client = s.get(Client.class, clientId);
+			Address address = client.getAddress();
+			address.getId();
+			assertThat(Hibernate.isInitialized(address), is(true));
```
igor-dmitriev: true cause ``id_employee`` is on Address side

beikov:
Just a few notes.

1. Make sure the license is at the beginning of every file.
2. Make sure you use the Hibernate code style. There are formatters for IntelliJ and Eclipse.
3. Make sure you have signed the CLA(contributers license agreement)

I haven't dug deeper, but one of the things you'd have to check is the bytecode of the getter method. It is important to know which fields are accessed in a getter to know whether it is legal to allow non-initialized access. Otherwise you might see strange NPE where you wouldn't expect them.

igor-dmitriev:
@beikov thank you, will do

igor-dmitriev:
hm. I've found a related issue - https://hibernate.atlassian.net/browse/HHH-3718.

igor-dmitriev:
@here is there any CI ?? Now I just can run tests locally

beikov:
There is a CI(http://ci.hibernate.org), but testing PRs is not yet automated. Please also squash the commits together to a single commit or split into a test and fix commit at the end and include the JIRA issue key in every commit message so that possible backporting gets easier.

vladmihalcea:
I reworked it a little bit and will supply a new PR which simplifies the logic a little bit while retaining the getter method resolving required to preserve the Proxy.

PR #1946, opened 2017-06-27
HHH-11838 - Id retrieving from proxy with FK leads to query execution

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-11838

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/util/ReflectHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/util/ReflectHelper.java
@@ -542,6 +542,24 @@ public final class ReflectHelper {
 		}
 	}
 
+	public static Method getterMethodOrNull(Class containerJavaType, String propertyName) {
+		try {
+			return findGetterMethod( containerJavaType, propertyName );
+		}
+		catch (PropertyNotFoundException e) {
+			return null;
+		}
+	}
+
+	public static Method setterMethodOrNull(Class containerJavaType, String propertyName, Class propertyJavaType) {
+		try {
+			return findSetterMethod( containerJavaType, propertyName, propertyJavaType );
+		}
+		catch (PropertyNotFoundException e) {
+			return null;
+		}
+	}
+
 	public static Method findSetterMethod(Class containerClass, String propertyName, Class propertyType) {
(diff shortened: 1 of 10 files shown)


vladmihalcea:
@igor-dmitriev Would you like to take a look on the issue?

igor-dmitriev:
@vladmihalcea yeah, sure

igor-dmitriev:
reviewed, lgtm, but I am wondering about that - https://github.com/hibernate/hibernate-orm/pull/1945#issuecomment-311001339

vladmihalcea:
@igor-dmitriev It's probably a duplicate of this issue.

beikov:
A test case that would probably fail is when the `getId()` method accessed other fields too. For the sake of testing, just invoke `streetName.length()` or so to verify, that such a getter should require proxy initialization i.e. it shouldn't throw a NPE, but simply load the entity when invoking `getId()`.
To determine whether the id getter or actually any method only accesses the id field, you'd have to analyze the bytecode of the method body. Determining this is probably not that easy and will only work for only the simplest methods, but that should be ok. It would be nice if the `equals` and `hashCode` methods could also avoid proxy initialization if only the id field is used.

vladmihalcea:
@beikov That's true. The only way to fix it would be if we introduce an annotation like @SideEffects and check if the getId method is annotated with it. Usually, the getId method should be straightforward so I'm inclined to think we don't really need it.

kullfar:
Hi, this change is breaking such case
```
@Entity
class A {
    @Column
    private String prop;

    public String getProp() {
        return prop;
    }

    public boolean isProp() {
        return prop = "some_value";
    }
}
```
Note, here is property-mapping. But two different getters. Starting from Hibernate 5.2.12 with this Pull Request applied we have now such exception at boostrapping:
Caused by: org.hibernate.MappingException: In trying to locate getter for property [prop], Class [A] defined both a `get` [public java.lang.String A.getProp()] and `is` [public boolean A.isProp()] variant

kullfar:
I have added comment to jira
https://hibernate.atlassian.net/browse/HHH-11838

mkurz:
You saved me some time @kullfar :smile:

PR #1948, opened 2017-07-03
HHH-11726 - Test case, passing when should fail

jonathanihm:
Could someone please take a look and see why this test case is passing even though it should fail according to the JIRA issue?  Thanks!


hibernate-core/src/test/java/org/hibernate/test/distinct/SelectDistinctHqlTest.java
```diff
@@ -108,6 +108,23 @@ public void test() {
...
+		doInHibernate( this::sessionFactory, session -> {
+			sqlStatementInterceptor.getSqlQueries().clear();
+			List<Person> persons = session.createQuery(
+					"select distinct p from Person p left join fetch p.phones ")
+					.setHint(QueryHints.HINT_PASS_DISTINCT_THROUGH, true)
+					.setMaxResults(5)
+					.getResultList();
+			assertEquals(1, persons.size());
+			String sqlQuery = sqlStatementInterceptor.getSqlQueries().getLast();
+			assertTrue(sqlQuery.contains(" distinct "));
```
vladmihalcea: When you set it to true, distinct will be passed through and it will be found din the SQL query.  You eed to set it to false if you only want to deduplicate Object references.

jonathanihm: Thanks for the input, Vlad!  I will retest and update accordingly.

jonathanihm: I updated the setHint to false and the assertTrue to assertFalse, and I am still ending up with a query with no distinct in it.  I will update the test method, but I am wondering if there's something I am missing.

vladmihalcea: if (you omit or set HINT_PASS_DISTINCT_THROUGH == true) then you'll see DISTINCT in the SQL query.

if (HINT_PASS_DISTINCT_THROUGH == false) then you will not see DISTINCT in the SQL query.

jonathanihm: Ah, then what should I be testing for this issue? It seems to be working correctly with whatever combination of HINT_PASS_DISTINCT_THROUGH and maxResults, although the patch that was submitted also looks correct.

PR #1950, opened 2017-07-11
FIX/HHH-11726 - Pass Distinct Test With False Pass Through

jonathanihm:
Still can't get this to pass


vladmihalcea:
Picked test case and fixed it as indicated. Thanks.

PR #1952, opened 2017-07-15
HHH-11871 - during metamodel generation, verify if methods respect JavaBeans conventions

BogdanStirbat:
When the JPAMetaModelEntityProcessor is executed, a new AnnotationMetaEntity is created for each Entity.
When the AnnotationMetaEntity is created, for a specific entity (Java class), each member of the class was registered as a persistent member. In this pull request, only members of class that respects the JavaBeans convention are added as persistent members.

I was thinking about this fix, by reading the [Java API specification](http://docs.oracle.com/javaee/7/tutorial/persistence-intro001.htm#BNBQA) : "If the entity uses persistent properties, the entity must follow the method conventions of JavaBeans components.". 
The same specification describes the [metamodel generation](https://docs.oracle.com/javaee/7/tutorial/persistence-criteria002.htm#GJIUP) : "For each entity class in a particular package, a metamodel class is created with a trailing underscore and with attributes that correspond to the persistent fields or properties of the entity class."

### DIFF
--- a/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/annotation/AnnotationMetaEntity.java
+++ b/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/annotation/AnnotationMetaEntity.java
@@ -15,6 +15,8 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.PackageElement;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.ExecutableType;
+import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.ElementFilter;
 import javax.tools.Diagnostic;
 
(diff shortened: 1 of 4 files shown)


ChristianCiach:
Thank you! But I think you try to fix this at the wrong place.

Your method "methodRespectsJavaBeansConvention()" is redundant, because this is already checked at `/util/StringUtil.isProperty(String methodName, String returnTypeAsString)`. This method is called by `/annotation/MetaAttributeGenerationVisitor.visitExecutable(ExecutableType t, Element p)`, which is (indirectly) called by `/annotation/AnnotationMetaEntity.addPersistentMembers(List<? extends Element> membersOfClass, AccessType membersKind)`.

I think the additional check for the method parameters (one for setters, none for getters) should be done in `MetaAttributeGenerationVisitor.visitExecutable(...)` or in `AnnotationMetaEntity.addPersistentMembers(...)`, whichever is more convinient.

BogdanStirbat:
Hi!

Than you for the observations. I will, first, explain why I added the check in that place. Then, some additional thoughts I had based on your suggestion.

I have added the check in that place, because the method call `addPersistentMembers( methodsOfClass, AccessType.PROPERTY );` adds the methods in the `members` map (please see `members.put( result.getPropertyName(), result );` ). The example `Homework` class (from tests) has, among others, in this order, 2 methods: `public List<String> getPaths()`, `public Set<String> getPaths(String startPath)`. If both methods are added to the map, the second method overwrites the first one. This is why, issue can be reproduced when the 2 methods are listed in this order: `public List<String> getPaths()`, `public Set<String> getPaths(String startPath)`; by reversing the order of declaration, issue is no longer reproducible, because the good method overwrites the bad one in the map. 

Based on your suggestion, I tried to replace the newly added method `methodRespectsJavaBeansConvention()` with `/util/StringUtil.isProperty(String methodName, String returnTypeAsString)`, thus eliminating duplicate code. But the method `/util/StringUtil.isProperty` does not check if method has parameter, thus both `List<String> getPaths()` and `Set<String> getPaths(String startPath)` would pass the check. 

However, your suggestion made me think, maybe Boolean properties can start both with 'is' and 'has'. 
According to [JPA specification](http://docs.oracle.com/javaee/7/tutorial/persistence-intro001.htm#BNBQA) and [this document](https://docstore.mik.ua/orelly/java-ent/jnut/ch06_02.htm), boolean getter may start with 'is'. Further suggestions are welcome.

BogdanStirbat:
This PR targets 5.1 branch.
Closing this PR, to add a new one targeting the master branch.

PR #1953, opened 2017-07-17
HHH-11864 - AutoCommit mode not reset after use by SchemaValidator

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-11864

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/tool/schema/internal/DdlTransactionIsolatorProvidedConnectionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/tool/schema/internal/DdlTransactionIsolatorProvidedConnectionImpl.java
@@ -9,6 +9,9 @@ package org.hibernate.tool.schema.internal;
 import java.sql.Connection;
 import java.sql.SQLException;
 
+import org.hibernate.engine.jdbc.connections.spi.JdbcConnectionAccess;
+import org.hibernate.internal.CoreLogging;
+import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.resource.transaction.spi.DdlTransactionIsolator;
 import org.hibernate.tool.schema.internal.exec.JdbcConnectionAccessProvidedConnectionImpl;
 import org.hibernate.tool.schema.internal.exec.JdbcContext;
@@ -20,6 +23,9 @@ import org.hibernate.tool.schema.spi.SchemaManagementException;
  * @author Steve Ebersole
  */
 class DdlTransactionIsolatorProvidedConnectionImpl implements DdlTransactionIsolator {
+
+	private static final CoreMessageLogger LOG = CoreLogging.messageLogger( DdlTransactionIsolatorProvidedConnectionImpl.class );
+
 	private final JdbcContext jdbcContext;
 
 	public DdlTransactionIsolatorProvidedConnectionImpl(JdbcContext jdbcContext) {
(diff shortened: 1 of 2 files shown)


vladmihalcea:
Thanks for the PR. It's superseded by [this PR](https://github.com/hibernate/hibernate-orm/pull/1953).

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/DdlTransactionIsolatorProvidedConnectionImpl.java
```diff
@@ -49,5 +55,17 @@ public void prepare() {
 
 	@Override
 	public void release() {
+		JdbcConnectionAccess connectionAccess = jdbcContext.getJdbcConnectionAccess();
```
gbadner: I see that `connectionAccessProvidedConnection` isn't used. Is the check on `connectionAccess` and the cast there just to make sure that the connection was provided?

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/DdlTransactionIsolatorProvidedConnectionImpl.java
```diff
@@ -49,5 +55,17 @@ public void prepare() {
...
+		if( !( connectionAccess instanceof JdbcConnectionAccessProvidedConnectionImpl ) ) {
+			throw new IllegalStateException( "DdlTransactionIsolatorProvidedConnectionImpl should always use a JdbcConnectionAccessProvidedConnectionImpl" );
+		}
+		JdbcConnectionAccessProvidedConnectionImpl connectionAccessProvidedConnection =
+			(JdbcConnectionAccessProvidedConnectionImpl) connectionAccess;
+		try {
```
gbadner: There should be a comment explaining why passing 'null' to `JdbcConnectionAccess#releaseConnection`is OK. 

Is it valid to pass `null` (rather than the `Connection`) because we know that that connection was provided?

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/DdlTransactionIsolatorProvidedConnectionImpl.java
vladmihalcea: The cast is not needed and should be removed. However, we need to check to make sure the connection access was provided (e.g. `JdbcConnectionAccessProvidedConnectionImpl`) and only then release the connection.

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/DdlTransactionIsolatorProvidedConnectionImpl.java
vladmihalcea: While passing the connection to the `releaseConnection` method might be suitable for the `JdbcConnectionAccessConnectionProviderImpl`, `JdbcEnvironmentInitiator` or `NonContextualJdbcConnectionAccess`, it has no meaning for `JdbcConnectionAccessProvidedConnectionImpl` because, in this case, the connection is wrapped and we don't have access to it upon releasing via the `DdlTransactionIsolatorProvidedConnectionImpl`. 

That's why I added the check above so to make sure we don't pass `null` to a connection access that's not `JdbcConnectionAccessProvidedConnectionImpl`.

vladmihalcea:
@gbadner I pushed a new update to this Pull Request which addresses your comments. Let me know if you agree with it. Thanks.

gbadner:
Looks good. Merged.

PR #1954, opened 2017-07-17
HHH-11176: Add support for Tuple results for native queries

galovics:
Basic implementation and tests added.

Now the EntityManager#createNativeQuery(String, Class) and EntityManager#createNamedQuery(String, Class) are supporting Tuple results.

One thing I was unsure about is what contract do we have for the Tuple#get(String) method regarding case-sensitivity? Currently, the implementation is case-sensitive but I feel like for native queries, this might be confusing.


vladmihalcea:
@galovics I reviewed it and made some small changes and submitted a new [Pull Request](https://github.com/hibernate/hibernate-orm/pull/1955) which I'd like you to review.

I agree that we should use lowercase aliases since, depending on the underlying DB, we might get lowercase or uppercase aliases and the users will be confused about what they'll get out of the tuple. For this reason, let's just store the lower case values and compare the key against the lowercase version of whatever the user sends us. This way it will just work no matter what the DB will return.

Thanks and looking forward to your comments.

PR #1790, opened 2017-02-09
HHH-11477 - HQL query against field marked with @Lob fails on PostgreSQL

dreab8:
https://hibernate.atlassian.net/browse/HHH-11477

PR #1957, opened 2017-07-20
HHH-11614 - Wrong result for @Lob column with postgresql DB since 5.2.9

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-11614


dreab8:
@vladmihalcea I talked also with @gbadner about this solution and we agree the best solution is just to revert the changes introduced by https://github.com/hibernate/hibernate-orm/pull/1790 and to add, in the migration guide and the release announcement, an explanation of the necessity to update PostgreSQL db records corresponding to `@Lob String` fields before upgrading from version >= 5.2.9 to 5.2.11

gbadner:
Closing because the fix for HHH-11477 will be reverted instead.

gbadner:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2001

PR #1959, opened 2017-07-20
HHH-11871 - during metamodel generation, verify if methods respect JavaBeans conventions

BogdanStirbat:
In method `AnnotationMetaEntity.init()`, function call `ElementFilter.methodsIn( element.getEnclosedElements() );` returns all methods the Entity has. One method might not respect JavaBeans convention; but, if it's listed after an other one, with the same name, following the JavaBeans convention, will overwrite the 'good' function in the `members` map (method call `addPersistentMembers`)

I fixed the issue, by checking if a method respects the convention prior to it's addition to the map. I followed the JavaBeans conventions explained in [this document](http://docs.oracle.com/javaee/7/tutorial/persistence-intro001.htm#BNBQA) .

### DIFF
--- a/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/annotation/AnnotationMetaEntity.java
+++ b/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/annotation/AnnotationMetaEntity.java
@@ -15,6 +15,8 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.PackageElement;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.ExecutableType;
+import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.ElementFilter;
 import javax.tools.Diagnostic;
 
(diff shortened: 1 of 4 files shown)


ChristianCiach:
Thanks again. Let be copy my comment from the closed pull request:

Your method "methodRespectsJavaBeansConvention()" is redundant, because this is already checked at `/util/StringUtil.isProperty(String methodName, String returnTypeAsString)`. This method is called by `/annotation/MetaAttributeGenerationVisitor.visitExecutable(ExecutableType t, Element p)`, which is (indirectly) called by `/annotation/AnnotationMetaEntity.addPersistentMembers(List<? extends Element> membersOfClass, AccessType membersKind)`.

I think the additional check for the method parameters (one for setters, none for getters) should be done in `MetaAttributeGenerationVisitor.visitExecutable(...)` or in `AnnotationMetaEntity.addPersistentMembers(...)`, whichever is more convinient.

BogdanStirbat:
Hi,


Thanks for the observation. 

In class `AnnotationMetaEntity`, method `addPersistentMembers` adds methods into a map. This map, `members`, has method name as key. Thus, if 2 methods have the same name, one is overwritten. For example, `public List<String> getPaths()` is overwritten by `public Set<String> getPaths(String startPath)`.

The proposed check can be added anywhere else. But, if a method is already overwritten by then, it's useless.

vladmihalcea:
Applied upstream, thanks

PR #1961, opened 2017-07-21
HHH-11826: ImplicitNamingStrategyComponentPathImpl generates invalid SQL for Entity with Embedded ElementCollection

simbacai:
The root cause is that by using ImplicitNamingStrategyComponentPathImpl strategy,  the internal place holder "collection&&element" which is part of the attribute path will be also taken as the part of the join column name.

The fix is just like what Ejb3Column did, in Ejb3JoinColumn:buildDefaultColumnName(), remove the place holder string "collection&&element" if it exist, and of course, only for the owner side.

Also add UT cases for this scenario.


vladmihalcea:
Applied upstream, thanks

PR #1962, opened 2017-07-23
Improved approach to providing database-specific setting overrides for testing

sebersole:
I'd actually suggest taking "the next step" here also and centralizing the main settings by either:

1. moving `hibernate.properties` into `hibernate-testing` as well, or
2. defining a "default" `h2.properties` in `hibernate-testing`

the idea of this "next step" being to remove the repetitiveness of having these settings in each module.


sebersole:
So we are all on board with this?  What about the "next step"?

dreab8:
I'm for moving ```hibernate.properties``` into ```hibernate-testing```

PR #1964, opened 2017-07-29
HHH-11897: Fix support for Tuple results for native queries

galovics:
Added couple more tests to verify the stream() behavior as well as the NPE bug.

I'm a bit unsure if the change in the Loader class (removing a method parameter and introducing the protected method) is appropriate or not, I'm open for discussion on that area.


galovics:
Are there any news here? ^^

vladmihalcea:
@galovics I reviewed your PR and think that we can avoid the inheritance which is only needed for `CustomLoader` and, instead, use a `Supplier` to provide the transformer alias array.

Check out my [Pull Request](https://github.com/hibernate/hibernate-orm/pull/1979) and let me know what you think.

PR #1968, opened 2017-07-31
HHH-11640 - NamedQuery doesn't log comment when UPDATE/DELETE

hradecek:
Prepending comment to native update, similarly as [AbstractLoadPlanBasedLoade.executeQueryStatement()#L183](https://github.com/hibernate/hibernate-orm/blob/3a813dcbb4b5cf5b13571c63ff0c66b94a83b53c/hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/AbstractLoadPlanBasedLoader.java#L183) or [Loader.executeQueryStatement()#L1909](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/loader/Loader.java#L1909)


sfragata:
Thanks @hradecek, this will definitely fix the issue

vladmihalcea:
Thanks for the PR. I continued it, add a test case and tried to unify the comment handling at the Dialect level. Therefore, this PR is replaced by [this one](https://github.com/hibernate/hibernate-orm/pull/1970).

PR #1970, opened 2017-08-01
HHH-11640 - NamedQuery doesn't log comment when UPDATE/DELETE

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-11640


dreab8:
Hi @vladmihalcea ,

may be I'm wrong but it seems with this PR the comments are always added not only when hibernate.use_sql_comments is true.

vladmihalcea:
@dreab8 That's a fine observation. I thought that only auto-generated Hibernate comments should be filtered by that property, not the explicit comments set by the users. In this case, we should probably block both implicit and explicit comments.

vladmihalcea:
@dreab8 I made the corrections. Let me know what you think.

PR #1971, opened 2017-08-02
HHH-11904  EnumExplicitTypeTest test assert fails on mariadb clusters

ChesterSVK:
https://hibernate.atlassian.net/browse/HHH-11904


vladmihalcea:
Applied upstream, thanks. 

For the 5.1 branch, @gbadner might want to integrate it.

PR #1972, opened 2017-08-02
HHH-11905 AbstractLobTest Fails on sybase

ChesterSVK:
Test will assert returned values on every DB except Sybase


vladmihalcea:
Applied upstream to master branch, thanks. For backporting, @gbadner should decide if this one should go into 5.1.

PR #1973, opened 2017-08-03
HHH-11907 - Add the getResultStream() default method in org.hibernate.query.Query

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-11907

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/Query.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/Query.java
@@ -1109,4 +1109,15 @@ public interface Query<R> extends TypedQuery<R>, org.hibernate.Query<R>, CommonQ
 		return this;
 	}
 
+	/**
+	 * JPA 2.2 defines the {@code getResultStream} method so to get a {@link Stream} from the JDBC {@link java.sql.ResultSet}.
+	 *
+	 * Hibernate 5.2 already defines the {@link Query#stream()} method, so {@code getResultStream} can delegate to it.
+	 *
+	 * @return The results Stream
+	 * @since 5.2.11
+	 */
+	default Stream<R> getResultStream() {
+		return stream();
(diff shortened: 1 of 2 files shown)


Sanne:
LGTM ! Thanks

sebersole:
As I mentioned on the mailing list, I have asked for clarification from the E.G. regarding the name of this method.  As I remember from E.G. discussions we had all agreed to name this `#stream` rather than `#getResultStream`.  `#getResultStream` was the original proposal, so I think that "changes doc" you saw may have just not been updated.

Let's wait to hear back about that.  If my memory is correct, we'd really have nothing to do here as we already define `#stream`

vladmihalcea:
@sebersole I saw that the [JPA spec page](https://www.jcp.org/en/jsr/detail?id=338) does not yet offer a final spec document. I'll check the web site from time to time and come back to this issue once we know for sure what's been decided.

sebersole:
I asked specifically.  So I'll let you know as soon as I know :)

ggam:
@vladmihalcea  the documenta are available on the JPA spec repository https://github.com/javaee/jpa-spec/tree/master/jsr338-MR

@sebersole did you post the message directly to the JPA mailing list?. I haven't received the email although I'm subscribed. Moreover, I sent a message myself this week but seems like it didn't work either.

sebersole:
I asked the lead directly as well as sending an email to the E.G. via reply to the original discussion thread.  I have not heard back from the spec lead and have not checked the email I use for the E.G. (I will tomorrow)

vladmihalcea:
@sebersole As you mentioned in the [Hibernate mailing list](http://lists.jboss.org/pipermail/hibernate-dev/2017-August/016406.html), it looks like `getResultStream` is what was chosen in the end, so we can integrate this issue.

PR #1974, opened 2017-08-08
HHH-11914 : SchemaUpdate.setHaltOnError(true) does nothing

gbadner:
I renamed the existing `SchemaUpdateHaltOnErrorTest` to `SchemaMigratorHaltOnErrorTest`, since it uses `SchemaMigrator`, not `SchemaUpdate`.

I added a new test named `SchemaUpdateHaltOnErrorTest`  that tests `SchemaUpdate.setHaltOnError(true) `.

I noticed that `hibernate.hbm2ddl.halt_on_error` applies to `SchemaMigrator`, but not `SchemaUpdate`.

Is `SchemaUpdate` considered deprecated? If not should  `hibernate.hbm2ddl.halt_on_error` apply to `SchemaUpdate` as well?


dreab8:
Hi @gbadner 
```SchemaUpdate``` delegates to ```SchemaMigrator```

https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/tool/hbm2ddl/SchemaUpdate.java#L87

dreab8:
@gbadner ,
sorry I see the problem, looking at the PR :)

vladmihalcea:
Applied upstream, thanks

PR #1975, opened 2017-08-08
HHH-11915 : DatabaseMetaData#getIndexInfo can return column names enc…

gbadner:
…losed in quotes on PostgresPlus

Should column names in the `ResultSet` returned by  `DatabaseMetaData#getIndexInfo` be stripped of quotes, even if the actual column in the index has quotes? It appears that is the case for all dialects except PostgresPlus.

Is this is bug in PostgresPlus?

Would this PR be a reasonable bugfix or workaround?

The bug is reproduced by `SchemaUpdateTest` on PostgresPlus.


gbadner:
I pushed Andrea's fix instead.

PR #1976, opened 2017-08-08
HHH-11916 - Unnecessary initialization of lazy collection on PERSIST cascade

bastii32:


vladmihalcea:
Applied upstream, thanks

PR #1977, opened 2017-08-13
Allow to use @OnDelete annotation on unidirectional @OneToMany associations

alek-sys:
Motivation for this change is that since JPA 2.x it is a valid configuration to have unidirectional one-to-many associations i.e. do not create many-to-one 'reverse' link back to parent.

However in order to allow DDL schema export to confiure foreign  keys properly and include ON DELETE CASCADE (for the supported dialects only) having many-to-one link is still required.

This change allow to add `@OnDelete` for `@OneToMany` association like this:

```java
@OneToMany(cascade = CascadeType.ALL)
@JoinColumn(name = "parent_id")
@OnDelete(action = OnDeleteAction.CASCADE)
List<Child> children;
```

This will automatically generate `ON CASCADE DELETE` on table for the entity `Child`.

Backward compatibility: as this change enable new configuration which wasn't allowed before it is backward compatible.

The only affected area is automatic DDL schema generation.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/mapping/Collection.java
+++ b/hibernate-core/src/main/java/org/hibernate/mapping/Collection.java
@@ -293,12 +293,6 @@ public abstract class Collection implements Fetchable, Value, Filterable {
 		assert getKey() != null : "Collection key not bound : " + getRole();
 		assert getElement() != null : "Collection element not bound : " + getRole();
 
-		if ( getKey().isCascadeDeleteEnabled() && ( !isInverse() || !isOneToMany() ) ) {
-			throw new MappingException(
-					"only inverse one-to-many associations may use on-delete=\"cascade\": "
-							+ getRole()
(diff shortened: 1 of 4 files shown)


vladmihalcea:
First, we need a Jira issue for this new feature.

Second, this feature breaks if you remove the `@JoinColumn` since you end up with an intermediary table and the cascade delete will not work anymore. You need to make sure this behavior is applied only when you have the `@JoinColumn` present.

alek-sys:
Superseded by https://github.com/hibernate/hibernate-orm/pull/1982

alek-sys:
Hi @vladmihalcea 
thanks for comments. I've made few changes and also found an old JIRA issue for this - https://hibernate.atlassian.net/browse/HHH-6382. 

Please take a look at the new pull request, https://github.com/hibernate/hibernate-orm/pull/1982.

PR #1980, opened 2017-08-16
HHH-11929 Improve Performance of SQLServer2012LimitHandler.hasOrderBy()

sisudev:
https://hibernate.atlassian.net/browse/HHH-11929

HHH-11929 removed excessive toLowerCase() and substring() from SQLServer2012LimitHandler.hasOrderBy(). Also reversed the direction for scanning the sql string for "order by " to start at the end of the string


Naros:
Applied upstream.

PR #1981, opened 2017-08-16
HHH-11815 - @org.hibernate.annotations.Index and NullPointerException in IndexOrUniqueKeySecondPass

lalmeras:
cf https://hibernate.atlassian.net/browse/HHH-11815

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/IndexOrUniqueKeySecondPass.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/IndexOrUniqueKeySecondPass.java
@@ -72,7 +72,7 @@ public class IndexOrUniqueKeySecondPass implements SecondPass {
 		if ( column != null ) {
 			this.table = column.getTable();
 
-			PersistentClass persistentClass = (PersistentClass) persistentClasses.get( column.getPropertyHolder().getEntityName() );
+			PersistentClass persistentClass = (PersistentClass) persistentClasses.get( column.getPropertyHolder().getEntityOwnerClassName() );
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/IndexOrUniqueKeySecondPass.java
```diff
@@ -72,7 +72,7 @@ public void doSecondPass(Map persistentClasses) throws MappingException {
 		if ( column != null ) {
 			this.table = column.getTable();
 
-			PersistentClass persistentClass = (PersistentClass) persistentClasses.get( column.getPropertyHolder().getEntityName() );
+			PersistentClass persistentClass = (PersistentClass) persistentClasses.get( column.getPropertyHolder().getEntityOwnerClassName() );
 			Property property = persistentClass.getProperty( column.getPropertyName() );
```
dreab8: not sure but probably getting the class name instead of the entity name may cause problems. 
what about https://github.com/hibernate/hibernate-orm/pull/1983/files?

lalmeras: Effectively, my solution seems buggy. It is surprising that it does not trigger any test failure.

Maybe adding an indexed field on your User entity (with its custom entity name) would be a good move to prevent regression on this use-case.

vladmihalcea:
I see that a [new PR](https://github.com/hibernate/hibernate-orm/pull/1983) was open by @dreab8. I'll close this one as it might be obsolete. Feel free to open it if this was not the case.

PR #1982, opened 2017-08-16
HHH-6382 - Allow to use OnDelete annotation on unidirectional OneToMany associations

alek-sys:
This is a reworked https://github.com/hibernate/hibernate-orm/pull/1977.

Changes made: reword commit message to include JIRA reference, add validation for `@JoinColumn` annotation when `@OnDelete` applied.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
@@ -46,6 +46,7 @@ import org.hibernate.annotations.LazyCollectionOption;
 import org.hibernate.annotations.LazyGroup;
 import org.hibernate.annotations.Loader;
 import org.hibernate.annotations.ManyToAny;
+import org.hibernate.annotations.OnDelete;
 import org.hibernate.annotations.OptimisticLock;
 import org.hibernate.annotations.OrderBy;
 import org.hibernate.annotations.Parameter;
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
```diff
@@ -480,6 +481,15 @@ public void bind() {
 			throw new AnnotationException( message );
 		}
 
+		if (!isMappedBy
```
vladmihalcea: What about XML mappings? You are searching for annotations only. For orm.xml or HBM mappings, this will not work, right?

alek-sys: Thanks for the feedback. 

I did a quick research and it seems that both HBM and ORM mappings are fine. Quick recap - validation checks that when unidirectional one-to-many association has `@OnDelete` annotation, it doesn't use *join table*, i.e. uses `@JoinColumn` annotation. 

Short summary:
1. HBM mapping doesn't need any extra validation. An unidirectional one-to-many link is defined in `hbm.xml` by having `<key column="entity_id" not-null="true" on-delete="cascade"  />`. By default it doesn't generate join table. So when `on-delete` defined `JoinColumn` is implicit and configuration is valid.
2. JPA mapping (orm.xml) doesn't need any validation at all as it doesn't allow to define `ON DELETE CASCADE` for a FK, according to the JPA spec. 

So overall conclusion is - no extra validation is needed, HBM works as intended, pure JPA `orm.xml` doesn't allow to define `ON CASCADE DELETE` for a FK at all. My suggestion is to accept the fix as is. 

A bit more details below. Happy to discuss it further if you'd like.

# HBM

HBM analogue for 

```java
@OneToMany(cascade = CascadeType.ALL)
@JoinColumn(name = "entity_id")
@OnDelete(action = OnDeleteAction.CASCADE)
...
```
is 

```xml
<list name="childEntityList" cascade="all">
    <key column="entity_id" not-null="true" on-delete="cascade"  />
    <list-index column="orderNumber" />
    <one-to-many class="ChildEntity"/>
</list>
```

Because `key` element defines both join column (via `column` attribute) **and** FK constraint (via `on-delete` attribute) there is no need in extra validation here. `on-delete` simply cannot be defined without implicit `JoinColumn` defined by `key`.

# JPA mapping

JPA Specification provides **no way** to define `ON DELETE CASCADE` for a foreign key. The only option is to use `@ForeignKey` annotation (or XML element) with `foreignKeyDefinition` property set to a SQL statement. 

Consider the following example:
```xml
<one-to-many name="childEntityList" orphan-removal="true">
    <join-column name="entity_id" nullable="false" />
    <cascade>
        <cascade-all/>
    </cascade>
</one-to-many>
```

`join-column` is defined here, but there is no way to define `on-delete` in `ORM.xml` so there is nothing to validate.

vladmihalcea:
Applied upstream, thanks

PR #1983, opened 2017-08-17
HHH-11815 - @org.hibernate.annotations.Index and NullPointerException in IndexOrUniqueKeySecondPass

dreab8:
https://hibernate.atlassian.net/browse/HHH-11815


gsmet:
@dreab8 @vladmihalcea it would be nice to drive this puppy home for 5.2.11.

@dreab8 Laurent posted a comment on the other PR:
> Maybe adding an indexed field on your User entity (with its custom entity name) would be a good move to prevent regression on this use-case.

It might be worth updating the test along these lines.

vladmihalcea:
Applied upstream, thanks

PR #1984, opened 2017-08-17
HHH-11934 Add a protected getter for the delegate in SessionFactoryDelegatingImpl

gsmet:
* https://hibernate.atlassian.net/browse/HHH-11934


gsmet:
@sebersole ah ah, I asked which style you preferred in the channel and Andrea said it was this one, I'll go change that!

dreab8:
yes my fault, I said i would have changed the field instead of introducing a get method

vladmihalcea:
Applied upstream, thanks

PR #1990, opened 2017-08-24
HHH-11938 - for not registered user defined database function calls, set type AnyType

BogdanStirbat:
If a user calls a non-registered, non-JPA standard, database function, it's type is not known.
Thus, in SelectClause, an exception is thrown: "No data type for node: MethodNode".

An example is added in QueryBuilderTest.testSinhFunction. The 'sinh' function is supported by H2 database, as described [here](http://www.h2database.com/html/functions.html#sinh). In H2Dialect class, the sinh function is not registered. Thus, test QueryBuilderTest.testSinhFunction fails, without this fix.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/MethodNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/MethodNode.java
@@ -18,6 +18,7 @@ import org.hibernate.internal.CoreLogging;
 import org.hibernate.persister.collection.CollectionPropertyMapping;
 import org.hibernate.persister.collection.CollectionPropertyNames;
 import org.hibernate.persister.collection.QueryableCollection;
+import org.hibernate.type.AnyType;
 import org.hibernate.type.Type;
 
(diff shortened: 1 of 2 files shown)


vladmihalcea:
That cannot work. Just change your test method to:

```
doInJPA( this::entityManagerFactory, entityManager -> {
	Customer customer = new Customer();
	customer.setId( "ABC123");
	customer.setAge( 35 );
	entityManager.persist( customer );

	CriteriaBuilderImpl cb = (CriteriaBuilderImpl) entityManager.getCriteriaBuilder();
	CriteriaQuery<Long> criteria = cb.createQuery( Long.class );
	criteria.from(Customer.class);
	criteria.select(cb.function("SINH", Long.class, cb.literal(2)));
	List<Long> results = entityManager.createQuery(criteria).getResultList();
	results.size();
} );
```
and you'll get a `NullPointerException` because `AnyType` was meant to select entities, not scalar results, so an entity identifier is expected.

The problem is that the result type is not passed from Criteria API to HQL. However, the HQL parser is going to change significantly in 6.0, so it's better to address it then.

PR #1991, opened 2017-08-25
HHH-11945 Make ExceptionConverterImpl use SharedSessionContractImplem…

gsmet:
…entor instead of AbstractSharedSessionContract

* https://hibernate.atlassian.net/browse/HHH-11945


vladmihalcea:
Applied upstream, thanks

PR #1993, opened 2017-08-25
HHH-11898 : test case

gbadner:
This test case illustrates why Hibernate should not support enabling empty composites for a composite with primitive values.

https://hibernate.atlassian.net/browse/HHH-11898

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/component/empty/EmptyCompositeIdAttributeTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/component/empty/EmptyCompositeIdAttributeTest.java
@@ -1,0 +1,186 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.test.component.empty;
+
+import java.io.Serializable;
+import javax.persistence.AttributeOverride;
+import javax.persistence.AttributeOverrides;
+import javax.persistence.Column;
+import javax.persistence.Embeddable;
+import javax.persistence.EmbeddedId;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.cfg.Configuration;
+import org.hibernate.cfg.Environment;
+
+import org.hibernate.testing.TestForIssue;
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.junit.Test;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInHibernate;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/test/java/org/hibernate/test/component/empty/EmptyCompositeIdAttributeTest.java
```diff
@@ -0,0 +1,186 @@
...
+				}
+		);
+
+		doInHibernate(
+				this::sessionFactory,
+				session -> {
+					final EntityA entityA = session.get( EntityA.class, entityAId );
+					if ( isEmptyEnabled ) {
+						// entityA.selectedEntityBPK should be an empty composite
+						assertNotNull( entityA.selectedEntityBPK );
```
gbadner: Currently, there is no restriction on composites that are "allowed" to be empty. With "empty composites" enabled, Hibernate will instantiate entityA.selectedEntityBPK with primitive values set to their default. As a consequence, the following statement will fail, because the empty composite has default values for the primitive values that are real values that is an ID for an existing EntityB.

hibernate-core/src/test/java/org/hibernate/test/component/empty/EmptyCompositeIdAttributeTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+		doInHibernate(
+				this::sessionFactory,
+				session -> {
+					final EntityA entityA = session.get( EntityA.class, entityAId );
+					entityA.selectedEntityBPK = new BPK();
+					// explicitly set the default int values to make clear
+					// that these are "real" values
+					entityA.selectedEntityBPK.intVal1 = 0;
+					entityA.selectedEntityBPK.intVal2 = 0;
```
gbadner: Suppose HHH-11898 is fixed so that an empty composite with primitive values are equivalent to `null` if primitive attributes are set to the primitive default and non-primitive attributes are null. 

When the session is flushed, `entityA` will *not* be updated because `entityA.selectedEntityBPK` will be equivalent to null. 

This will happen even with `hibernate.create_empty_composites.enabled=false`!

hibernate-core/src/test/java/org/hibernate/test/component/empty/EmptyCompositeIdAttributeTest.java
```diff
@@ -0,0 +1,186 @@
...
+				}
+		);
+
+		doInHibernate(
+				this::sessionFactory,
+				session -> {
+					final EntityA entityA = session.get( EntityA.class, entityAId );
```
gbadner: Suppose HHH-11898 is fixed so that an empty composite with primitive values are equivalent to null if primitive attributes are set to the primitive default and non-primitive attributes are null.

If empty composites are enabled, the following will pass because Hibernate will read null values `entityA.selectedEntityBPK` (because `entityA` was not updated in the previous transaction), then Hibernate will inject an empty composite with primitive attributes set to the default which are *real* values.

If empty composites are disabled, the following will *fail* because `entityA.selectedEntityBPK` will be null (because `entityA` was not updated in the previous transaction).

PR #1994, opened 2017-08-27
HHH-11950 Target WildFly 11 for produced hibernate-orm-modules

Sanne:
https://hibernate.atlassian.net/browse/HHH-11950


vladmihalcea:
Applied upstream, thanks

PR #1995, opened 2017-08-27
HHH-8955 - added HSQL dialect support for database trunc function

BogdanStirbat:
Hi!

In HSQLDialect, I added support for trunc function.
For HSQL database, trunc function is documented [here](http://hsqldb.org/doc/2.0/guide/builtinfunctions-chapt.html) .


vladmihalcea:
Applied upstream, thanks

PR #1997, opened 2017-08-30
HHH-11944 HHH-11956 Changes related to the OGM upgrade

gsmet:
This should be reviewed commit per commit.

The first set of commits (https://hibernate.atlassian.net/browse/HHH-11944) fix the delegating classes. I think it should be considered bugfixes as there were several hierarchy/generics issue in them. I think I got everything consistent and tested now (through concrete classes implementing the delegating classes in test/).

The last commit (https://hibernate.atlassian.net/browse/HHH-11956) is probably more controversial for a micro. I think I reduced the risk of incompatibility to the minimum. And it really helps with reducing the amount of code we have to copy to OGM.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/query/spi/NativeQueryInterpreter.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/query/spi/NativeQueryInterpreter.java
@@ -8,6 +8,8 @@ package org.hibernate.engine.query.spi;
 
 import org.hibernate.engine.query.spi.sql.NativeSQLQuerySpecification;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.loader.custom.CustomLoader;
+import org.hibernate.loader.custom.CustomQuery;
 import org.hibernate.query.internal.ParameterMetadataImpl;
 import org.hibernate.service.Service;
 
@@ -16,6 +18,7 @@ import org.hibernate.service.Service;
  *
  * @author Steve Ebersole
  * @author Gunnar Morling
+ * @author Guillaume Smet
  */
 public interface NativeQueryInterpreter extends Service {
 	/**
@@ -38,4 +41,17 @@ public interface NativeQueryInterpreter extends Service {
 	 * @return A query plan for the specified native query.
 	 */
 	NativeSQLQueryPlan createQueryPlan(NativeSQLQuerySpecification specification, SessionFactoryImplementor sessionFactory);
+
+	/**
+	 * Creates a {@link CustomLoader} for the given {@link CustomQuery}.
+	 *
+	 * @param customQuery The CustomQuery to create a loader for
+	 * @param sessionFactory The current session factory
+	 *
+	 * @deprecated This method will be removed in 6.
+	 */
+	@Deprecated
+	default CustomLoader createCustomLoader(CustomQuery customQuery, SessionFactoryImplementor sessionFactory) {
+		return new CustomLoader( customQuery, sessionFactory );
+	}
 }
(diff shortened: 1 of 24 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/SessionFactoryDelegatingImpl.java
```diff
@@ -73,11 +75,16 @@ public SessionFactoryDelegatingImpl(SessionFactoryImplementor delegate) {
 		this.delegate = delegate;
 	}
 
-	protected SessionFactoryImplementor getDelegate() {
+	protected SessionFactoryImplementor delegate() {
```
gsmet: This is not a compatibility break as this method was introduced during the 5.2.11 release cycle. I renamed it for consistency after a discussion with Steve.

sebersole:
Is "the last commit" https://github.com/hibernate/hibernate-orm/pull/1997/commits/9ccd73c7990155c2b51e42eecc191c347004f70e?  GitHub shows them in an order where that is not the last

hibernate-core/src/main/java/org/hibernate/boot/SessionFactoryBuilder.java
```diff
@@ -705,6 +706,17 @@ default SessionFactoryBuilder applyConnectionProviderDisablesAutoCommit(boolean
 	SessionFactoryBuilder enableReleaseResourcesOnCloseEnabled(boolean enable);
 
 	/**
+	 * Specify the CustomLoaderFactory to use.
+	 *
+	 * @param customLoaderFactory The CustomLoaderFactory to use.
+	 *
+	 * @return {@code this}, for method chaining
+	 */
+	default SessionFactoryBuilder applyCustomLoaderFactory(CustomLoaderFactory customLoaderFactory) {
```
sebersole: I don't think I am a fan of these kind of method defaults.  Here the default really does not do what it is documented to do - meaning that an implementation inheriting this method (thats what method defaults are for after all) will not work correctly.

sebersole: Additionally, considering this stuff all goes away in 6 I am not really a fan of making this "user facing".  *If* we do end up exposing it to users, these methods need to be deprecated.

gsmet:
@sebersole I force pushed an update with something in line with what we discussed. It's a far more limited change now thanks to your suggestions.

hibernate-core/src/main/java/org/hibernate/engine/query/internal/NativeQueryInterpreterStandardImpl.java
```diff
@@ -73,4 +75,9 @@ public NativeSQLQueryPlan createQueryPlan(
 
 		return new NativeSQLQueryPlan( specification.getQueryString(), customQuery );
 	}
+
+	@Override
```
sebersole: Can we remove this impl?  It does exactly the same as the one it overrides

hibernate-core/src/main/java/org/hibernate/engine/query/spi/NativeQueryInterpreter.java
```diff
@@ -38,4 +41,17 @@
 	 * @return A query plan for the specified native query.
 	 */
 	NativeSQLQueryPlan createQueryPlan(NativeSQLQuerySpecification specification, SessionFactoryImplementor sessionFactory);
+
```
sebersole: +1

gsmet:
Merged!

PR #1999, opened 2017-09-01
HHH-11942 - ANTLR parser should fail when providing an extra parenthesis

dreab8:
https://hibernate.atlassian.net/browse/HHH-11942


dreab8:
applied upstream

PR #2000, opened 2017-09-04
HHH-3870 - Hibernate proxies Groovy's getMetaClass method breaking proxies when used with Groovy

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-3870


vladmihalcea:
@raphw Can you take a look on this PR and let me know if the condition I added using the ByteBuddy API translates to the one I added for [`JavassistProxyFactory`](https://github.com/hibernate/hibernate-orm/pull/2000/files#diff-6009579af64063d971d4a58468cce007R36). Thanks.

@emmanuelbernard Is this what you had in mind?

raphw:
@vladmihalcea It is but I would recommend you to fully ignore the method by:

```java
new ByteBuddy()
  .ignore(isSynthetic().or(<groovy-method>))
```

vladmihalcea:
@raphw I applied the change you requested and pushed force it. Let me know if that's what you had in mind. Thanks.

PR #2001, opened 2017-09-08
HHH-11614

gbadner:
https://hibernate.atlassian.net/browse/HHH-11614

Reverts HHH-11477 and updates migration guide.


vladmihalcea:
In [this Pull Request](https://github.com/hibernate/hibernate-orm/pull/1957/files#diff-55dc5879da1d97052105082c26075701R101), I added an example of how to migrate from the Strong-based CLOB handling to the reverted one. I'm not sure if it applies to the current PR as well, but we can still try it and maybe tell our users how to migrate their Clob data for a smooth Hibernate upgrade.

dreab8:
Hi @gbadner,
the PR looks good to me.
@vladmihalcea your test is still good, what can we do it is to take the code you used in the test to update the db data and add this to the migration guide.
An example can be:
If using PostgreSql  database with @Lob annotation and example of the data migration sql that should be run before upgrading from hibernate 5.2.9 or 5.2.10 to 5.2.11 is, given the following mapping:

```
@Entity(name = "TestEntity")
@Table(name = "TEST_ENTITY")
public static class TestEntity {
	@Id
	@GeneratedValue
	private long id;

	@Lob
	@Column
	String firstLobField;

	@Lob
	@Column
	String secondLobField;

	@Lob
	@Column
	Clob clobField;

	...
}
```
the following sql should be executed:

`update test_entity set clobfield = lo_from_bytea(0, cast(clobfield as bytea)), firstlobfield = lo_from_bytea(0, cast(firstlobfield as bytea)),secondlobfield = lo_from_bytea(0, cast(secondlobfield as bytea))`

gbadner:
@vladmihalcea, @dreab8, I added the tests back in. I think they are great for giving examples and to make sure they continue to work. Thanks for mentioning them.

I've also updated the migration guide with the example that Andrea mentioned.

Thanks for your help on this!

gbadner:
Merged.

PR #2002, opened 2017-09-08
HHH-11756 - Typo in public API method name: requiresPostCommitHanding…

dreab8:
… on PostInsertEventListener
https://hibernate.atlassian.net/browse/HHH-11756

I had to introduce the `PostActionEventListener` interface in order to avoid compilation errors for classes implementing all the 3 interfaces affected by the typo due to the 3 default methods with the same name (
aN example is  `org.hibernate.cache.internal.CollectionCacheInvalidator`).


gbadner:
@dreab8, this looks good to me. Should I merge?

dreab8:
Thanks @gbadner,
I have changed the visibility of PostActionEventListener to package and rebased the PR.
If for you t is fine you can merge it.

Thanks

gbadner:
I agree package is better. Merged.

PR #1907, opened 2017-05-11
HHH-11732 - HHH000352: in StatelessSession on rollback with JDBC batch

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-11732

PR #2004, opened 2017-09-12
HHH-11732 : HHH000352: in StatelessSession on rollback with JDBC batch

gbadner:
https://hibernate.atlassian.net/browse/HHH-11732

This backs out of the fix in https://github.com/hibernate/hibernate-orm/pull/1907 and fixes in a different way.

I'm not sure which is better.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/LogicalConnectionProvidedImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/LogicalConnectionProvidedImpl.java
@@ -28,6 +28,13 @@ public class LogicalConnectionProvidedImpl extends AbstractLogicalConnectionImpl
 	private final boolean initiallyAutoCommit;
 	private boolean closed;
 
+	/**
+	 * This constructor should only be used when making a shareable connection or for testing.
+	 *
+	 * @param providedConnection - the provided connection.
+	 *
+	 * @see #makeShareableCopy()
+	 */
 	public LogicalConnectionProvidedImpl(Connection providedConnection) {
 		this( providedConnection, new ResourceRegistryStandardImpl() );
 	}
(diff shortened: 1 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/LogicalConnectionProvidedImpl.java
```diff
@@ -28,6 +28,13 @@
 	private final boolean initiallyAutoCommit;
 	private boolean closed;
 
+	/**
+	 * This constructor should only be used when making a shareable connection or for testing.
+	 *
+	 * @param providedConnection - the provided connection.
+	 *
+	 * @see #makeShareableCopy()
+	 */
```
gbadner: AFAICT, only tests and `LogicalConnectionProvidedImpl#makeShareableCopy` use this constructor.

`LogicalConnectionProvidedImpl#makeShareableCopy` does not appear to be used. Should it be deprecated?

Also, I'm not sure if the new `ResourceRegistryStandardImpl` should use the original `JdbcObserver`. If it should, then this class needs to be reworked.

vladmihalcea:
Looks good to me.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/LogicalConnectionProvidedImpl.java
gbadner: I heard back from @sebersole that `LogicalConnectionProvidedImpl#makeShareableCopy` can be deprecated.

gbadner: I'm going to remove this constructor, since the caller should provide the `ResourceRegistry`.

gbadner:
Squashed and merged.

PR #2006, opened 2017-09-15
Improved Hibernate support for SAP HANA

breglerj:
- Identity column support
- Spatial support
- Various minor improvements in HANA dialects


sebersole:
@maesenka - I understand you've already looked over the spatial related changes here?

maesenka:
@sebersole I have looked at the spatial side of this. 

I have just one remark on the code. The class `HANASpatialAggregate` seems to me superfluous (it overrides `StandardSQLFunction::renderType` but with code that looks to me equivalent to the overridden implementation).

I also have a couple of questions about the maintenance of this code. First, who is going to run the (spatial) matrix tests? Are they going to be integrated in the CI? If you want me to run this, I'd like some more information on how to set things up (I know virtually nothing about SAP HANA).  

@breglerj, are you going to be my contact if I have questions about HANA's spatial features? For example, when I intend to extend the Hibernate Spatial functionality.

libraries.gradle
```diff
@@ -100,6 +100,8 @@ ext {
             oracle:          'com.oracle.jdbc:ojdbc7:12.1.0.2',
             mssql:           'com.microsoft.sqlserver:mssql-jdbc:6.1.0.jre8',
             db2:             'com.ibm.db2:db2jcc:10.5',
+            //hana:            'com.sap.db.jdbc:ngdbc:1.120.20',
+            hana:            'com.sap.db.jdbc:ngdbc:2.2.1',
```
sebersole: Is this jar not publicly available?

hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/spi/HANAMultiTenantConnectionProvider.java
```diff
@@ -0,0 +1,137 @@
...
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.engine.config.spi.ConfigurationService;
+import org.hibernate.engine.jdbc.connections.internal.ConnectionProviderInitiator;
+import org.hibernate.service.spi.Configurable;
+import org.hibernate.service.spi.ServiceRegistryAwareService;
+import org.hibernate.service.spi.ServiceRegistryImplementor;
+import org.hibernate.service.spi.Stoppable;
+
+public class HANAMultiTenantConnectionProvider extends AbstractMultiTenantConnectionProvider
+		implements ServiceRegistryAwareService, Stoppable, Configurable {
```
sebersole: This class should be removed.  We do not provide specific MultiTenantConnectionProvider implementations

sebersole: Although I could see this generalized into a provider for multi-tenancy Connections based on providing the tenant name as URL parameter or property.

libraries.gradle
breglerj: You mean the version 1.120.20? I added it so people using HANA 1 know what the minimum supported version of the client is. If you prefer I can just add a comment stating this instead of the commented out line or leave it out completely.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/spi/HANAMultiTenantConnectionProvider.java
breglerj: OK, I'll remove it.

breglerj:
@maesenka: I'll remove the render() method implementation from HANASpatialAggregate. I'd like to keep the class for symmetry reasons if that's ok with you.

We're currently working on how we can integrate the HANA tests into the Hibernate CI infrastructure. I would assume that we can integrate the spatial tests as well. I'll keep you posted.

Yes, you can contact me if you have questions about the HANA spatial features.

sebersole:
Applied upstream

PR #2007, opened 2017-09-16
Correct a variable name in the documentation

Kwadz:
This renames the variable `number` to `customer`


Kwadz:
How come the tests failed while it is a documentation update?

vladmihalcea:
Thanks, but the docbook documentation is not used anymore. We are using the one in the asciidoc folder. I'm going to remove the docbook one to avoid confusing people.

PR #2008, opened 2017-09-18
HHH-11816 - JoinProcessor doesn't handle certain table names correctly

breglerj:
https://hibernate.atlassian.net/browse/HHH-11816


vladmihalcea:
I created a new [Pull Request](https://github.com/hibernate/hibernate-orm/pull/2015) which adds the following improvements:

1. We consider the Dialect-specific quoting character
2. The parsing logic is reduced to a single Regular Expression.

PR #2009, opened 2017-09-21
HHH-12001 - Allow ORM to be built with Java 9

dreab8:
https://hibernate.atlassian.net/browse/HHH-12001

### DIFF
--- a/libraries.gradle
+++ b/libraries.gradle
@@ -12,7 +12,7 @@ ext {
     junitVersion = '4.12'
 //    h2Version = '1.2.145'
     h2Version = '1.3.176'
-    bytemanVersion = '3.0.10'
+    bytemanVersion = '4.0.0-BETA5'
     infinispanVersion = '8.2.5.Final'
     jnpVersion = '5.0.6.CR1'
     elVersion = '2.2.4'
--- a/libraries.gradle
+++ b/libraries.gradle
@@ -12,7 +12,7 @@ ext {
     junitVersion = '4.12'
 //    h2Version = '1.2.145'
     h2Version = '1.3.176'
-    bytemanVersion = '3.0.6'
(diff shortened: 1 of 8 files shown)


libraries.gradle
```diff
@@ -12,7 +12,7 @@ ext {
     junitVersion = '4.12'
 //    h2Version = '1.2.145'
     h2Version = '1.3.176'
-    bytemanVersion = '3.0.6'
+    bytemanVersion = '3.0.10'
```
Sanne: I would recommend using Byteman version `4.0.0-BETA5` : only version 4 is intended to work on Java 9.
I understand Byteman `3.0.10` seems to work, but that's pure luck ..

Sanne:
The task `:hibernate-c3p0:findbugsMain` fails with several errors. Should we have the `build` task skip it automatically? Alternatively maybe there is a fixed version of the findbugs plugin already?

libraries.gradle
dreab8: I'm going to upgrade it

dreab8:
unfortunately it seems findbugs has no support for Java 9 and not sure if it will never have, based on https://github.com/gradle/gradle/issues/720. I think we will have to switch to SpotBugs in the meantime I will disable for Java 9.

dreab8:
Thanks @Sanne,  pushed the changed you suggested.

Sanne:
Thanks @dreab8 !
One more problem:

`:documentation:aggregateJavadocs` seems to fail as well with a weird error. It's late for me now so I won't check why :) But we might want to skip javadocs as well?

Or should I simply avoid to invoke the `build` target  and stick to `test` instead?

Sanne:
merged it! The javadocs problem can wait, being able to test it w/o other weird workarounds is great progress already.

PR #2010, opened 2017-09-22
fix test failures on SAP HANA

breglerj:
- skip testLoad() of org.hibernate.jpa.test.ops.GetLoadTest
- add sort to testRevisionsGtWithPropertyQuery() of org.hibernate.envers.test.integration.query.RevisionConstraintQuery

### DIFF
--- a/hibernate-envers/src/test/java/org/hibernate/envers/test/integration/query/RevisionConstraintQuery.java
+++ b/hibernate-envers/src/test/java/org/hibernate/envers/test/integration/query/RevisionConstraintQuery.java
@@ -126,10 +126,9 @@ public class RevisionConstraintQuery extends BaseEnversJPAFunctionalTestCase {
 				.addProjection( AuditEntity.revisionNumber() )
 				.add( AuditEntity.revisionNumber().gt( 1 ) )
 				.add( AuditEntity.property( "number" ).lt( 10 ) )
-				.addOrder( AuditEntity.revisionNumber().asc() )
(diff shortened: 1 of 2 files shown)


hibernate-envers/src/test/java/org/hibernate/envers/test/integration/query/RevisionConstraintQuery.java
```diff
@@ -126,6 +126,7 @@ public void testRevisionsGtWithPropertyQuery() {
 				.addProjection( AuditEntity.revisionNumber() )
 				.add( AuditEntity.revisionNumber().gt( 1 ) )
 				.add( AuditEntity.property( "number" ).lt( 10 ) )
+				.addOrder( AuditEntity.revisionNumber().asc() )
```
Naros: I am personally not a fan of adding an `#addOrder` here.  The method name defines we're testing two cases, the greater-than and property value check.  

Instead I would suggest we alter the assertion check like we do in other methods where we compare the contents of two sets as follows:
<pre>
Assert.assertEquals( TestTools.makeSet( 3, 4 ), new HashSet( result ) );
</pre>

breglerj: That's fine with me. I'll make the change and update the PR.

vladmihalcea:
Applied upstream, thanks

PR #2011, opened 2017-09-22
HHH-7842 - Hibernate Criteria does not respect fetch mode, when alias is used

ThomasReinhardt:


vladmihalcea:
Applied upstream, thanks

PR #2012, opened 2017-09-25
HHH-11980 MultiTenantConnectionProvider is required for MultiTenancyS…

msvticket:
…trategy.DISCRIMINATOR

Instead of checking that multiTenancyStrategy is NONE start using MultiTenancyStrategy.requiresMultiTenantConnectionProvider to distinguish whether MultiTenantConnectionProvider is to be used.


vladmihalcea:
Applied upstream, thanks

vladmihalcea:
Applied upstream, thanks

PR #2013, opened 2017-09-25
HHH-11838 : Added tests for getting the ID from a proxy for an entity…

gbadner:
… that does not exist

https://hibernate.atlassian.net/browse/HHH-11838

I don't think the fix is consistent with JPA.

Documentation for `EntityNotFoundException` says:

"The EntityNotFoundException is thrown by the persistence provider when an entity
reference obtained by getReference is accessed but the entity does not exist. ...
The current transaction, if one is active and the persistence context has been joined to
it, will be marked for rollback.

I've added a test that follows what is defined by JPA. It fails with the current fix. After reverting the commit, the test will pass (after resolving conflicts).

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/ops/GetLoadTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/ops/GetLoadTest.java
@@ -8,6 +8,7 @@ package org.hibernate.jpa.test.ops;
 
 import java.util.Map;
 import javax.persistence.EntityManager;
+import javax.persistence.EntityNotFoundException;
 
 import org.hibernate.Hibernate;
 import org.hibernate.Session;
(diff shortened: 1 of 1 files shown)


sebersole:
I'm trying to understand the concern here.  Is your concern, Gail, that calling the identifier getter on a proxy does not trigger that rest of the proxy state to be loaded?

sebersole:
After discussing with @dreab8 I see that that is indeed the concern.  And while I get where you get that idea from, I do not agree with the behavior you propose because it is a regression from what we had always done.  In that regard HHH-11838 is simply reporting a regression.  I am not aware of when that original behavior was changed, but it should not have.

Aside from being an obvious performance question and a regression behavior, I think it a logical assumption that calling #getId on a thing i just got by id does not need to "do anything" to be able to return the id value.

To me, this is all correct - as long as accessing other state on the proxy leads to the EntityNotFoundException.

@dreab8 says he asked @vladmihalcea to amend his commit to add a setting to control this.  I can definitely see value in that, mainly in case the TCK ever decides to try to add a test for this[1].  However that setting's default should absolutely be to not trigger the query when accessing the id.

[1] However I am pretty sure it does not and will not because other providers make this same assumption.

vladmihalcea:
@sebersole @gbadner @dreab8 After analyzing the issue, I came to the same conclusion as Steve.

I remembered that Hibernate was not supposed to trigger a Proxy initialization when calling the id getter. What I didn't remember was that this behavior has been working when using property-based access.

The HHH-11838 issue has actually done the same for field-based access, so now, they behave in the very same way.

Therefore, even if we reak the JPA spec, I think it's the spec that's broken since we've been having this behavior for a very long time now, and I don't think that any user will complain about not initializing the Proxy when calling the identifier getter.

To be JPA compliant, I'll create a new issue which will combine Gail's test and a new one that proves that we've been having this issue for property-based access even before we fixed HHH-11838.

vladmihalcea:
@gbadner @sebersole @dreab8 I created [HHH-12034](https://hibernate.atlassian.net/browse/HHH-12034) and pushed [a commit](https://github.com/hibernate/hibernate-orm/commit/a77c48dc111132eba0b5c64b339c144abebe4b5f) which demonstrates that we have this issue for both field-based and property-based access.

It's better to address both using a new issue instead of reopening HHH-11838.

PR #2014, opened 2017-09-25
HHH-11970 : @NotFound(IGNORE) and @BatchSize

gbadner:
https://hibernate.atlassian.net/browse/HHH-11970

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/batchfetch/BatchFetchNotFoundIgnoreDefaultStyleTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/batchfetch/BatchFetchNotFoundIgnoreDefaultStyleTest.java
@@ -1,0 +1,339 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.test.batchfetch;
+
+import java.util.ArrayList;
+import java.util.List;
+import javax.persistence.ConstraintMode;
+import javax.persistence.Entity;
+import javax.persistence.ForeignKey;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.OneToOne;
+
+import org.hibernate.Session;
+import org.hibernate.annotations.BatchSize;
+import org.hibernate.annotations.NotFound;
+import org.hibernate.annotations.NotFoundAction;
+import org.hibernate.cfg.Configuration;
+import org.hibernate.cfg.Environment;
+import org.hibernate.engine.spi.BatchFetchQueue;
+import org.hibernate.engine.spi.EntityKey;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.persister.entity.EntityPersister;
+import org.hibernate.resource.jdbc.spi.StatementInspector;
+
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInHibernate;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+/**
+ * @author Gail Badner
+ * @author Stephen Fikes
+ */
+public class BatchFetchNotFoundIgnoreDefaultStyleTest extends BaseCoreFunctionalTestCase {
+	private static final AStatementInspector statementInspector = new AStatementInspector();
+	private static final int NUMBER_OF_EMPLOYEES = 8;
+
+	private List<Task> tasks = new ArrayList<>();
+
+	@Override
+	protected Class<?>[] getAnnotatedClasses() {
+		return new Class[] { Employee.class, Task.class };
+	}
+
+	@Override
+	protected void configure(Configuration configuration) {
+		super.configure( configuration );
+		configuration.getProperties().put( Environment.STATEMENT_INSPECTOR, statementInspector );
+	}
+
+	@Before
+	public void createData() {
+		tasks.clear();
+		tasks = doInHibernate(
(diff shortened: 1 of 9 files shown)


hibernate-core/src/test/java/org/hibernate/test/batchfetch/BatchFetchNotFoundIgnoreDefaultStyleTest.java
```diff
@@ -0,0 +1,339 @@
...
+				}
+		);
+
+		statementInspector.clear();
+
+		final List<Employee> employees = doInHibernate(
+				this::sessionFactory, session -> {
+					List<Employee> results =
+							session.createQuery( "from Employee e order by e.id", Employee.class ).getResultList();
+					for ( int i = 0 ; i < tasks.size() ; i++ ) {
```
gbadner: Without the fix, the `EntityKey` for each entity that as not found would remain in the `BatchFetchQueue`.

The following would fail for `i == 0`.

hibernate-core/src/test/java/org/hibernate/test/batchfetch/BatchFetchNotFoundIgnoreDefaultStyleTest.java
```diff
@@ -0,0 +1,339 @@
...
+		assertEquals( 0, paramterCounts.get( 0 ).intValue() );
+
+		// query specifically for Task with ID == 0 will result in 1st batch;
+		// query should have 5 parameters for [0,1,2,3,4];
+		// Task with IDs == [0,1,2,3,4] won't be found
+		assertEquals( 5, paramterCounts.get( 1 ).intValue() );
+
+		// query specifically for Task with ID == 1 will result in 2nd batch;
+		// query should have 4 parameters [1,5,6,7];
+		// Task with IDs == [1,5,6] won't be found; Task with ID == 7 will be found.
```
gbadner: Without the fix, the following IDs would be queried in the batch: [1,5.2,3,4], causing the following to fail.

hibernate-core/src/test/java/org/hibernate/test/batchfetch/BatchFetchNotFoundIgnoreDefaultStyleTest.java
```diff
@@ -0,0 +1,339 @@
...
+		assertEquals( 4, paramterCounts.get( 2 ).intValue() );
+
+		// query specifically for Task with ID == 2 will result in just querying
+		// Task with ID == 2 (because the batch is empty).
+		// query should have 1 parameter [2];
+		// Task with ID == 2 won't be found.
```
gbadner: Without the fix, the following IDs would be queried in the batch: [2,5.6,3,4], causing the following to fail.

hibernate-core/src/test/java/org/hibernate/test/batchfetch/BatchFetchNotFoundIgnoreDefaultStyleTest.java
```diff
@@ -0,0 +1,339 @@
...
+		assertEquals( 1, paramterCounts.get( 3 ).intValue() );
+
+		// query specifically for Task with ID == 3 will result in just querying
+		// Task with ID == 3 (because the batch is empty).
+		// query should have 1 parameter [3];
+		// Task with ID == 3 won't be found.
```
gbadner: Without the fix, the following IDs would be queried in the batch: [3,5,6,7,4], causing the following to fail.

After the fix, the batch would be empty, so Hibernate would simply try to load the entity with id == 3 (since it was not found in the last batch). This would perform much better than before the fix.

hibernate-core/src/test/java/org/hibernate/test/batchfetch/BatchFetchNotFoundIgnoreDefaultStyleTest.java
```diff
@@ -0,0 +1,339 @@
...
+		assertEquals( 1, paramterCounts.get( 4 ).intValue() );
+
+		// query specifically for Task with ID == 4 will result in just querying
+		// Task with ID == 4 (because the batch is empty).
+		// query should have 1 parameter [4];
+		// Task with ID == 4 won't be found.
```
gbadner: Without the fix, the following IDs would be queried in the batch: [4,5,6,2,3], causing the following to fail.

After the fix, the batch would be empty, so Hibernate would simply try to load the entity with id == 4 (since it was not found in the last batch). This would perform much better than before the fix.

hibernate-core/src/test/java/org/hibernate/test/batchfetch/BatchFetchNotFoundIgnoreDefaultStyleTest.java
```diff
@@ -0,0 +1,339 @@
...
+		assertEquals( 1, paramterCounts.get( 5 ).intValue() );
+
+		// query specifically for Task with ID == 5 will result in just querying
+		// Task with ID == 5 (because the batch is empty).
+		// query should have 1 parameter [5];
+		// Task with ID == 5 won't be found.
```
gbadner: Without the fix, the following IDs would be queried in the batch: [5,4,6,2,3], causing the following to fail.

After the fix, the batch would be empty, so Hibernate would simply try to load the entity with id == 5 (since it was not found in the last batch). This would perform much better than before the fix.

hibernate-core/src/test/java/org/hibernate/test/batchfetch/BatchFetchNotFoundIgnoreDefaultStyleTest.java
```diff
@@ -0,0 +1,339 @@
...
+		assertEquals( 1, paramterCounts.get( 6 ).intValue() );
+
+		// query specifically for Task with ID == 6 will result in just querying
+		// Task with ID == 6 (because the batch is empty).
+		// query should have 1 parameter [6];
+		// Task with ID == 6 won't be found.
```
gbadner: Without the fix, the following IDs would be queried in the batch: [6,4,5,2,3], causing the following to fail.

After the fix, the batch would be empty, so Hibernate would simply try to load the entity with id == 6 (since it was not found in the last batch). This would perform much better than before the fix.

dreab8:
Rebased and applied upstream.
Thanks @gbadner

PR #2015, opened 2017-09-28
HHH-11816 - JoinProcessor considers table names with colons dynamic filter parameters

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-11816


vladmihalcea:
@breglerj Let me know what you think of this Pull Request.

breglerj:
@vladmihalcea Looks good to me.

One more thing I thought of is that hasDynamicFilterParam() should also check if the match ends with a single quote. Otherwise the same error occurs with a join condition like this:

... inner join User u on r.lastUpdateBy = u.username and u.username!='user:name' ...

vladmihalcea:
@breglerj You can grab this PR and add a new test that fails for this new use case. Send the PR when you have it and I'll integrate it.

breglerj:
@vladmihalcea I've created PR #2016

PR #2017, opened 2017-09-29
HHH-11965 - Fix cascade of orphan-removal on one-to-one associations with proxies.

Naros:
https://hibernate.atlassian.net/browse/HHH-11965


gbadner:
Looks good; merged. Thanks!

PR #2021, opened 2017-10-11
HHH-12026 Make sure that search icon is rendered correctly in TOC

marko-bekhta:
- https://hibernate.atlassian.net/browse/HHH-12026

It turned out that that search input in TOC wasn't using the FontAwesome font-family hence rendering the strange symbol instead of a search icon. I've also fixed a small typo in JavaDoc. 

As for the icon I'm not 100% sure that that's enough. Should the url for the font itself (https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/fonts/fontawesome-webfont.ttf) be added to `Hibernate_User_Guide-docinfo.html` or is it enough...


vladmihalcea:
Applied upstream, thanks

PR #2022, opened 2017-10-13
HHH-12033 README.md links should use Markdown notation instead of AsciiDoc

koppor:
This PR fixes a small typo in the links in README.md's markdown.

Before:

grafik

After:

grafik

Also http is replaced by https in that area.


vladmihalcea:
Applied upstream, thanks

PR #2026, opened 2017-10-14
HHH-11971 - QueryParameterBindingValidator does not handle primitive types

dreab8:
https://hibernate.atlassian.net/browse/HHH-11971

The check is needed because the primitive parameter is automatically boxed, so by example true become an instance of `Boolean` while the expected type is a `boolean` this cause the `expectedType.isInstance( value )` returning false.


dreab8:
applied upstream

PR #2027, opened 2017-10-15
HHH-11845 - Warn user when multiple persistence-units use the same name

dreab8:
https://hibernate.atlassian.net/browse/HHH-11845

The PU names clash was performed for names belonging to the same persistence.xml file but  if  the same PU name is present in two different files, like in case of testing, the clash does not produce any warning.  The a solution is to have a  class `private final Map<String,ParsedPersistenceXmlDescriptor> persistenceUnits; `field, so for every PU we can check if there is already one with the same name.


vladmihalcea:
Applied upstream, thanks

PR #2028, opened 2017-10-18
HHH-12040 Add simple validation for CriteriaQuery and groupBy/having

marko-bekhta:
- https://hibernate.atlassian.net/browse/HHH-12040

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/CriteriaQueryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/CriteriaQueryImpl.java
@@ -258,6 +258,11 @@ public class CriteriaQueryImpl<T> extends AbstractNode implements CriteriaQuery<
 		if ( getSelection() == null && !hasImplicitSelection() ) {
 			throw new IllegalStateException( "No explicit selection and an implicit one could not be determined" );
 		}
+
+		// check if there's a having clause without the group by defined.
+		if ( queryStructure.getHaving() != null && queryStructure.getGroupings().isEmpty() ) {
+			throw new IllegalStateException( "Having clause defined without the grouping" );
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/criteria/internal/CriteriaQueryImpl.java
```diff
@@ -258,6 +258,11 @@ public void validate() {
 		if ( getSelection() == null && !hasImplicitSelection() ) {
 			throw new IllegalStateException( "No explicit selection and an implicit one could not be determined" );
 		}
```
gbadner: I believe this check is too strict. SQL 92 says: 

         "A grouped table is a set of groups derived during the evaluation
         of a <group by clause> or a <having clause>. A group is a multiset
         of rows in which all values of the grouping column or columns are
         equal if a <group by clause> is specified, or the group is the
         entire table if no <group by clause> is specified. "

IIUC, HAVING does not require GROUP BY.

marko-bekhta: hmmm, yes... it seems so. In that case I think this part is not good:
```java
if ( ! getGroupings().isEmpty() ) {
	jpaqlQuery.append( " group by " );
	String sep = "";
	for ( Expression grouping : getGroupings() ) {
		jpaqlQuery.append( sep )
				.append( ( (Renderable) grouping ).render( renderingContext ) );
		sep = ", ";
	}

	if ( getHaving() != null ) {
		jpaqlQuery.append( " having " )
				.append( ( (Renderable) getHaving() ).render( renderingContext ) );
	}
}
```
It's from `QueryStructure#render` method. See how it's checking the groupings and only if they are present - adding having clause. Shouldn't that "having" if block be outside the grouping one? WDYT?

gbadner: I agree that doesn't look right.

gbadner:
@marko-bekhta, I'm closing this because your proposed fix is too strict. 

Could you try a test case that uses HAVING with no GROUP BY? If it fails, then please create a jira for that issue.

PR #1872, opened 2017-04-07
HHH-11615 - Fix Envers to run under WildFly's SecurityManager.

Naros:
https://hibernate.atlassian.net/browse/HHH-11615

PR #2029, opened 2017-10-19
HHH-11615

gbadner:
https://hibernate.atlassian.net/browse/HHH-11615

Includes the commit from https://github.com/hibernate/hibernate-orm/pull/1872/.

It adds a commit that avoids wrapping an AuditException in an AuditException.

I've run the WildFly test from https://issues.jboss.org/browse/WFLY-8377 and it ran successfully with the added commit.


hibernate-envers/src/main/java/org/hibernate/envers/internal/entities/mapper/relation/component/MiddleEmbeddableComponentMapper.java
```diff
@@ -52,6 +52,10 @@ public Object mapToObjectFromFullMap(
 			);
 			return componentInstance;
 		}
```
gbadner: An alternative is to remove the try-catch. 

I chose to keep the try-catch and just throw the ```AuditException``` because I was concerned that ```delegate.mapToEntityFromMap``` could throw an ```Exception``` that should be wrapped by the ```AuditException```.

PR #2030, opened 2017-10-19
Use correct logger.

jvdvegt:
This class was using the logger for DriverManagerConnectionProviderImpl, which causes confusion when debugging log messages.


vladmihalcea:
Actually, the problem here is that `PooledConnections` is not a nested class in `DriverManagerConnectionProviderImpl`. This class is only meant to be used by `DriverManagerConnectionProviderImpl`.

vladmihalcea:
I created [HHH-12052](https://hibernate.atlassian.net/browse/HHH-12052) for this.

PR #2031, opened 2017-10-20
HHH-12046 - Don't consider static methods for isFoo()/getFoo() conflicts

archiecobbs:


archiecobbs:
Revoking, need to add unit test.

PR #2032, opened 2017-10-20
HHH-12046 - Don't consider static methods for isFoo()/getFoo() conflicts

archiecobbs:
https://hibernate.atlassian.net/browse/HHH-12046


vladmihalcea:
The unit test runs fine even if I revert your changes to the `ReflectHelper`. Please provide an example which proves that, without your change, we have a failing use case.

archiecobbs:
Oops, sorry. That should be fixed now.

vladmihalcea:
Applied upstream, thanks

PR #2034, opened 2017-10-23
HHH-12055 Revert to using the Javassist version included in WildFly 11

Sanne:
This is a small step but it also simplifies the packaging needs for end users.

I originally started looking at this as apparently the JDK9 CI build started failing after we upgraded Javassist; it only failed on WildFly so this is kinda of a workaround: don't use our "latest" on WildFly tests yet.

Incidentally, I tried to switch the integration tests to ByteBuddy but the Javassist supporting code is still killing the test, so we'll have to work on making Javassist truely optional.


Sanne:
merged

PR #2035, opened 2017-10-25
HHH-12036 Use cache.clear() for bulk removal on non-tx cache

rvansa:
https://hibernate.atlassian.net/browse/HHH-12036
https://issues.jboss.org/browse/ISPN-8411

@gbadner Should I create a fix for 5.2 as well?


vladmihalcea:
@rvansa If this also affects 5.2, we usually apply the fix on the master and then backport it to older branches.

gbadner:
This is an improvement, so closing because only critical bugfixes are being backported to 5.1 branch.

PR #2036, opened 2017-10-25
HHH-12036 Use cache.clear() for bulk removal on non-tx cache

rvansa:
https://hibernate.atlassian.net/browse/HHH-12036
https://issues.jboss.org/browse/ISPN-8411

The same changes as #2035


vladmihalcea:
Applied upstream, thanks

PR #2037, opened 2017-10-25
HHH-12051 - Filter dirty but non-updatable atributes

barreiro:


vladmihalcea:
Applied upstream, thanks

PR #2038, opened 2017-10-27
HHH-10677 Local Timestamp cache is not updated right after cache.put

rvansa:
https://hibernate.atlassian.net/browse/HHH-10677


rvansa:
The 5.1 backport will be simple though modifications will be needed due to the flush out of tx.

vladmihalcea:
Applied upstream, thanks

PR #2039, opened 2017-10-27
HHH-12068 - Fix generated @SuppressWarnings to work for all compilers

archiecobbs:


tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/ClassWriter.java
```diff
@@ -210,7 +210,7 @@ private static String writeGeneratedAnnotation(MetaEntity entity, Context contex
 	}
 
 	private static String writeSuppressWarnings() {
-		return "@SuppressWarnings(\"all\")";
+		return "@SuppressWarnings({ \"deprecation\", \"rawtypes\", \"serial\" })";
```
vladmihalcea: Thanks for the tip. I have one question. Why did you pick these [3 our of all supported warnings](https://docs.oracle.com/javase/7/docs/technotes/tools/solaris/javac.html#xlintwarnings)?

I wonder if we should also add  `varargs` to the list.

archiecobbs: Admittedly I'm not exactly sure which ones can occur. We know we need `"deprecation"` and `"rawtypes"`, but I'm not sure if/when others may occur.

I don't think it's possible for a metamodel class to implement `Serializable` so I don't think we need `"serial"`. I'll update the pull request.

And if the only thing possible in a generated metamodel class is the `public static volatile ...` declarations (appears to be the case from looking at `ClassWriter.generateBody()`), then I don't think any other warnings could occur.

Re `"varargs"` .. how could that happen? I don't see it but could be missing something.

archiecobbs:
Hmm... somehow when I updated the `HHH-12046` branch the previous comments associated with this PR got deleted... ?

First comment was:

> 
> @vladmihalcea commented on this pull request.
> 
> In tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/ClassWriter.java:
> ```
> > @@ -210,7 +210,7 @@ private static String writeGeneratedAnnotation(MetaEntity entity, Context contex
>  	}
>  
>  	private static String writeSuppressWarnings() {
> -		return "@SuppressWarnings(\"all\")";
> +		return "@SuppressWarnings({ \"deprecation\", \"rawtypes\", \"serial\" })";
> ```
> Thanks for the tip. I have one question. Why did you pick these 3 our of all supported warnings?
> 
> I wonder if we should also add varargs to the list.

My response was basically all we should need is `"deprecation"` and "`rawtypes`" because the only content within generated classes is the `public static volatile` attribute fields.

vladmihalcea:
Applied upstream, thanks

PR #2041, opened 2017-10-28
HHH-12067: NativeQueryTupleTransformer doesn't handle null values in returned ResultSet

msymonov:


vladmihalcea:
Applied upstream, thanks

PR #2042, opened 2017-10-30
HHH-12054 - Handle UNFETCHED_PROPERTY in CollectionType.replace()

barreiro:
with this change all types support replace() where the target is UNFETCHED_PROPERTY, removing the need for special handling in TypeHelper.replace()

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/TypeHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/TypeHelper.java
@@ -156,25 +156,8 @@ public class TypeHelper {
 			final Map copyCache) {
 		Object[] copied = new Object[original.length];
 		for ( int i = 0; i < types.length; i++ ) {
-			if ( original[i] == LazyPropertyInitializer.UNFETCHED_PROPERTY
-				|| original[i] == PropertyAccessStrategyBackRefImpl.UNKNOWN ) {
+			if ( original[i] == LazyPropertyInitializer.UNFETCHED_PROPERTY || original[i] == PropertyAccessStrategyBackRefImpl.UNKNOWN ) {
 				copied[i] = target[i];
-			}
-			else if ( target[i] == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
-				// Should be no need to check for target[i] == PropertyAccessStrategyBackRefImpl.UNKNOWN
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/type/TypeHelper.java
```diff
@@ -160,22 +160,6 @@ else if ( row[i] == LazyPropertyInitializer.UNFETCHED_PROPERTY || row[i] == Prop
 				|| original[i] == PropertyAccessStrategyBackRefImpl.UNKNOWN ) {
 				copied[i] = target[i];
 			}
```
gbadner: The following change will cause ClassCastException when target is a lazy property of a mutable type that extends ```AbstractStandardBasicType```.

Both implementations of ```AbstractStandardBasicType#replace``` use ```AbstractStandardBasicType#getReplacement``` (one implementation of ```AbstractStandardBasicType#replace``` is ```final```; the other is not overridden by any subclasses):

```
       protected T getReplacement(T original, T target, SharedSessionContractImplementor session) {
		if ( !isMutable() ) {
			return original;
		}
		else if ( isEqual( original, target ) ) {
			return original;
		}
		else {
			return deepCopy( original );
		}
	}
```
```AbstractStandardBasicType#isEqual``` will cast ```LazyPropertyInitializer.UNFETCHED_PROPERTY``` to the expected type, which will throw ```ClassCastException```.

I updated your test to reproduce this [here](https://github.com/gbadner/hibernate-core/commit/860cef19039ede01b27e5da4b740075d093ba1c0).

gbadner: Can embedded values be lazy? If so, may be a problem with lazy ```ComponentType```, because `implementations of ```ComponentTuplizer``` do not handle ```LazyPropertyInitializer.UNFETCHED_PROPERTY```.

There may also be problems with ```CustomType``` and```CompositeCustomType``` properties as well, depending on how applications implement ```UserType#replace```.

barreiro:
Added some more changes to address @gbadner comments. 
I don't see `ComponentType` as being an issue. As for user type, it may be, but only when bytecode enhacement is enabled.
Added a check in `TypeTest`to make sure that basic types support unloaded properties.

vladmihalcea:
Applied upstream, thanks

PR #2043, opened 2017-10-31
HHH-12072 BasicFormatterImpl throws a NPE if native SQL begins with a parentheses

ageery:
This can happen, for example, where there are two select statements
being union'ed together and the select statements are placed in
parentheses for readability.


vladmihalcea:
For every Pull Request, we need:

1. A Jira issue so that we can track changes
2. A test case to replicate the problem the PR is fixing. Just take one test from hibernate-core and use it as a template for your test case.

ageery:
Done: https://hibernate.atlassian.net/browse/HHH-12072

vladmihalcea:
@ageery Thanks for creating the issue. Now, please provide a replicating test case.

ageery:
It's attached to the JIRA ticket.

vladmihalcea:
@ageery Thanks, I'll take it from there,

vladmihalcea:
Applied upstream, thanks

PR #1901, opened 2017-05-08
Test for https://hibernate.atlassian.net/browse/HHH-11721

graemerocher:
Example demonstrating NullPointerException described in issue https://hibernate.atlassian.net/browse/HHH-11721

PR #2044, opened 2017-11-01
HHH-11721 - PreInsertEventListener that just vetoed ends up throwing NullPointerException because the entity is still managed

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-11721


vladmihalcea:
@Naros This is a replacement for this [Pull Request](https://github.com/hibernate/hibernate-orm/pull/1901) which only provided a test case. If you have time, let me know if you agree with the proposed fix.

I checked it for PRE_UPDATE, and if a Listener vetoes, the update will be skipped while the rest of actions will proceed just fine.

vladmihalcea:
@gbadner If you have time, would you mind taking a look at this issue. The idea is that if an `EventListener` vetoes the `InsertAction` we should throw an exception.

PR #2046, opened 2017-11-03
HHH-5409: Implement PersistentBad.equals as intended in Collections

JasperHorn:
Implements an equals with bag semantics.

I have some tests in org.hibernate.jpa.test.cdi are failing, but they are also failing on master, so I think those are false positives (or would that be false negatives?).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
@@ -515,67 +515,56 @@ public class PersistentBag extends AbstractPersistentCollection implements List 
 	@Override
 	public boolean equals(Object obj) {
 		read();
-		
-		if (!obj.getClass().equals(PersistentBag.class))
-		{
-			return false;
-		}
-		
-		PersistentBag other = (PersistentBag)obj;
-		
-		if (this.bag.size() != other.bag.size())
-		{
+
+		if ( !obj.getClass().equals( PersistentBag.class ) ) {
 			return false;
 		}
 
-		Map<Object, Integer> elementCounts = new HashMap<>(this.bag.size());
+		PersistentBag other = (PersistentBag) obj;
 
-		for (Object element : this.bag)
-		{
-			if (!elementCounts.containsKey(element))
-			{
-				elementCounts.put(element, 1);
+		if ( this.bag.size() != other.bag.size() ) {
+			return false;
+		}
+
+		Map<Object, Integer> elementCounts = new HashMap<>( this.bag.size() );
+
+		for ( Object element : this.bag ) {
+			if ( !elementCounts.containsKey( element ) ) {
+				elementCounts.put( element, 1 );
 			}
-			else
-			{
-				elementCounts.put(element, elementCounts.get(element) + 1);
+			else {
+				elementCounts.put( element, elementCounts.get( element ) + 1 );
 			}
 		}
 
-		for (Object element : other.bag)
-		{
-			if (!elementCounts.containsKey(element))
-			{
+		for ( Object element : other.bag ) {
+			if ( !elementCounts.containsKey( element ) ) {
 				return false;
 			}
-			else
-			{
-				elementCounts.put(element, elementCounts.get(element) - 1);
+			else {
+				elementCounts.put( element, elementCounts.get( element ) - 1 );
 			}
 		}
 
-		for (Integer count : elementCounts.values())
-		{
-			if (count != 0)
-			{
+		for ( Integer count : elementCounts.values() ) {
+			if ( count != 0 ) {
 				return false;
 			}
 		}
-		
+
 		return true;
 	}
 
 	@Override
 	public int hashCode() {
-		
+
 		int code = 0;
-		
-		for (Object element : bag)
-		{
+
+		for ( Object element : bag ) {
 			// Any commutative operation will be order independent
 			code += element.hashCode();
 		}
-		
+
 		return code;
 	}
 
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
@@ -11,9 +11,11 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
+import java.util.Map;
 
 import org.hibernate.HibernateException;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
@@ -510,18 +512,57 @@ public class PersistentBag extends AbstractPersistentCollection implements List 
 		return bag.toString();
 	}
 
-	/**
-	 * Bag does not respect the collection API and do an
-	 * JVM instance comparison to do the equals.
-	 * The semantic is broken not to have to initialize a
-	 * collection for a simple equals() operation.
-	 * @see java.lang.Object#equals(java.lang.Object)
-	 *
-	 * {@inheritDoc}
-	 */
 	@Override
 	public boolean equals(Object obj) {
-		return super.equals( obj );
+		read();
+		
+		if (!obj.getClass().equals(PersistentBag.class))
+		{
+			return false;
+		}
+		
+		PersistentBag other = (PersistentBag)obj;
+		
+		if (this.bag.size() != other.bag.size())
+		{
+			return false;
+		}
+
+		Map<Object, Integer> elementCounts = new HashMap<>(this.bag.size());
+
+		for (Object element : this.bag)
+		{
+			if (!elementCounts.containsKey(element))
+			{
+				elementCounts.put(element, 1);
+			}
+			else
+			{
+				elementCounts.put(element, elementCounts.get(element) + 1);
+			}
+		}
+
+		for (Object element : other.bag)
+		{
+			if (!elementCounts.containsKey(element))
+			{
+				return false;
+			}
+			else
+			{
+				elementCounts.put(element, elementCounts.get(element) - 1);
+			}
+		}
+
+		for (Integer count : elementCounts.values())
+		{
+			if (count != 0)
+			{
+				return false;
+			}
+		}
+		
+		return true;
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
@@ -567,7 +567,16 @@ public class PersistentBag extends AbstractPersistentCollection implements List 
 
 	@Override
 	public int hashCode() {
-		return super.hashCode();
+		
+		int code = 0;
+		
+		for (Object element : bag)
+		{
+			// Any commutative operation will be order independent
+			code += element.hashCode();
+		}
+		
+		return code;
 	}
 
 	final class Clear implements DelayedOperation {
--- a/hibernate-core/src/test/java/org/hibernate/test/collection/bag/PersistentBagEqualsTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/collection/bag/PersistentBagEqualsTest.java
@@ -5,118 +5,114 @@ import static org.junit.Assert.assertEquals;
 import java.util.ArrayList;
 
 import org.hibernate.collection.internal.PersistentBag;
-import org.hibernate.collection.spi.PersistentCollection;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.testing.TestForIssue;
 import org.junit.Test;
 import org.mockito.Mockito;
-import org.mockito.invocation.InvocationOnMock;
-import org.mockito.stubbing.Answer;
 
 /**
  * Tests collection equals behavior given the semantic meaning of a bag
  *
  * @author Jasper Horn
  */
-@TestForIssue( jiraKey="HHH-5409" )
+@TestForIssue(jiraKey = "HHH-5409")
 public class PersistentBagEqualsTest {
-	
+
 	@Test
-	public void testTwoEmptyBags()
-	{
+	public void testTwoEmptyBags() {
 		PersistentBag bag1 = createBag();
 		PersistentBag bag2 = createBag();
-		
-		assertEquals(bag1.equals(bag2), true);
+
+		assertEquals( bag1.equals( bag2 ), true );
 	}
-	
+
 	@Test
 	public void testTwoBagsWithDifferentContents() {
 		PersistentBag bag1 = createBag();
-		bag1.add("item1");
-		
+		bag1.add( "item1" );
+
 		PersistentBag bag2 = createBag();
-		bag2.add("item2");
-		
-		System.out.println("Added items");
-		
-		assertEquals(bag1.equals(bag2), false);
+		bag2.add( "item2" );
+
+		System.out.println( "Added items" );
+
+		assertEquals( bag1.equals( bag2 ), false );
 	}
 
 	@Test
 	public void testTwoBagsWithSameContents() {
 		PersistentBag bag1 = createBag();
-		bag1.add("item");
-		
+		bag1.add( "item" );
+
 		PersistentBag bag2 = createBag();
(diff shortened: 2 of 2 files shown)


vladmihalcea:
Did you read the comments on the issue? The reason why Bag uses that particular implementation is to avoid initializing the collection for [equals](https://github.com/hibernate/hibernate-orm/commit/23326608f7929676c315e85d14ea59d2607dfc5c).

If the entities implement equals and hashCode properly, that shouldn't be an issue as [demonstrated by this test](https://github.com/hibernate/hibernate-orm/commit/23326608f7929676c315e85d14ea59d2607dfc5c).

JasperHorn:
I did read the comments. As stated in the issue:

- PersistentMap, PersistentSet and PersistentList actually implement equals this way (they have it slightly simpler, because their underlying collections have the same semantics, but they all load unloaded data from the database)
- The implementation on master breaks the contract that goes with the Collection interface.
- Because the code using Hibernate probably does not actually contain any reference to the PersistentBag class (just to an annotated List), it can be hard to debug this problem (the reason is my own interpretation, but others mentioned how much time debugging this can cost)
- Other libraries (e.g. lombok through its generated equals method) may depend on proper implementation of the equals method (alright, that's not from the issue or its comments)

To counter your "use constraints or keys instead" argument: if I want to see if something is the same entry from your database, I write an equals function that does not call the collection's equals, so the implementation of the PersistentSet does not matter for performance or correctness. If, on the other hand, I want to compare two object to see if they contain the same data (which can be either two entries from the database or an entry from the database and an object not from the database), I write an equals that does call the collection's equals, and this breaks because the the collection's equals does not function as expected.

The following is a code sample that shows how unintuitive the current master implementation is:

    class A
    {
	@Id
	@Column(name = "ID", updatable = false, nullable = false)
	public String id;

	@ManyToOne
	@JoinColumn(name = "A_ID")
        public List<B> bs;

        public A(List<B> bs)
        {
            this.bs = bs;
        }

        @Override
        public boolean equals(Object o)
        {
            // Skipping type checking and null checking for simplicity
            return texts.equals(());
        }
    }

    class B
    {
	@Id
	@Column(name = "ID", updatable = false, nullable = false)
	public String id;
    }
    
    // sample code fragment
    
    A a1 = new A(new ArrayList<String>());
    A a2 = new A(new ArrayList<String>());

    assertTrue(a1.equals(a2)); // succeeds

    a1 = crudRepository.save(a1);

    assertTrue(a1.equals(a2)); // fails!

vladmihalcea:
But you are showing me that a bad implementation of equals does not work properly. Lombok is a nice framework, but that does not mean it works for JPA entities.

So, please show me how the current implementation breaks your data access logic when you are implementing `equals` and `hashCode` the right way:

1. Either [using a natural identifier, when you have one](https://vladmihalcea.com/2013/10/23/hibernate-facts-equals-and-hashcode/)
2. Either [using a the entity identifier using a consistent `equals` and `hashCode` implementation, when you don't have a natural identifier](https://vladmihalcea.com/2016/06/06/how-to-implement-equals-and-hashcode-using-the-jpa-entity-identifier/)

JasperHorn:
No, I'm showing you that an implementation that doesn't follow *Vlad's preferred way™* breaks. I also explained that anything using *Vlad's preferred way™* isn't affected by this change at all (neither in performance nor in effect) because it never calls this function. So, the change to the equals makes the code more versatile as it no longer only works for what some people consider the only correct usage.

To give an example that's less abstract, consider a blogging system. The author of a blog post edits their post and a new version of that blog post is sent to the backend. The backend has to decide whether it will actually create a new revision or that it doesn't have to because the author didn't actually make any changes. In other words, I want to see if the newly sent post is equal to the one stored in the database.

Now, I can see how you might prefer a different implementation which doesn't use equals to do this comparison, or one that copies the data over to another class, call equals on that, then if the two are not equal, copies the data to the managed class and store that in the database. However, I don't think that Hibernate should be forcing me to do that and should support a proper implementation of equals for collections.

> The current implementation was chosen for performance reasons.

But the same choice (i.e. not loading data from the database) was not made for other collections. Can you explain to me why that is? 

(Also, note that this change does not affect performance as long as you're following Vlad's preferred way™ and the equals method doesn't give the expected result to begin with when you're not, so the performance argument doesn't really combine well with the other argument you are making.)

> That's documented in the Javadoc:

It is. That doesn't mean it's not a problem, though. 

A documented flaw is still a flaw and it's still better not to have that flaw. You can argue that this is not a flaw, but you can't say it's not a flaw because it is documented.

> Just like getting a `LazyInitializationException` is an indication that the fetching policy is wrong, bumping into this problem is usually a code smell that you're doing something wrong. in your data access layer.

Whether or not we agree on this being a code smell, I think we can agree that an equals method not giving the result you were expecting is a pretty poor way to _indicate_ anything at all.

> Maybe @sebersole can provide you more context into why this decision was taken a very long time ago.

I would love to hear the reasoning behind this.

My impression was that the choice was made (and contested) a very long time and that the lack of the same choice in other persistent collections indicated that the current situation was more about the fact that the implementation was more complicated (because a List isn't a Bag, while the other collections are backed by implementation that matches their semantics) than about the original reason for the implementation. I may very well be wrong about this.

vladmihalcea:
@sebersole Can you tell us about the reason why PersistentBag is the way it is and whether we should change the `equals` implementation?

Thanks

sebersole:
> My impression was that the choice was made (and contested) a very long time and that the lack of the same choice in other persistent collections indicated that the current situation was more about the fact that the implementation was more complicated

Simply not true -

```
@Override
public void equals(Object other) {
    if ( obj instanceof java.util.Collection ) {
        final Collection other = (Collection) obj;
        // has to be initialized for this "proper" check to work, right?
        initialize( false );
        return bag.containsAll( other );
    }
    return false;
}
```

Because that's really the semantic.  No the point is really about performance.  Where is this "contention" you speak of?  Link?  Pics or it never happened.

Personally, I don't care.

gavinking:
> Using a Collection in your equals is just wrong. Why would ever want to do that? This is not consistent with the way tuples are unique in the context of the Relational Algebra. Is the Primary Key of a table ever associated with the result of running a query or a JOIN?

That's quite correct. An `equals()` that depends on a collection is almost certainly completely broken. And imagine what happens if the collection is uninitialized when `equals()` is called.

> The implementation on master breaks the contract that goes with the `Collection` interface.

I don't think that's right, actually. Here's the contract of `Collection.equals()`:

> While the Collection interface adds no stipulations to the general contract for the Object.equals, programmers who implement the Collection interface "directly" (in other words, create a class that is a Collection but is not a Set or a List) must exercise care if they choose to override the `Object.equals`. It is not necessary to do so, and **the simplest course of action is to rely on `Object`'s implementation**, but the implementor may wish to implement a "value comparison" in place of the default "reference comparison." (The List and Set interfaces mandate such value comparisons.)

Now, it's true that you can also type `Bag`s as `List`s, but they're *not* lists, and this is actually an abuse of the `List` interface.

gavinking:
I mean, if there's anything to criticize here I think it's that we let you treat a field of type `List` as a bag. We shouldn't.

Probably what we should have done, many years ago, was introduce our own `interface org.hibernate.Bag extends Collection`. Hell, we still could do it in H6.

sebersole:
Meh, Collection itself captures the idea fine.

I do agree maybe it is time to reconsider allowing a List attribute to be a
bag.

JasperHorn:
> Because that's really the semantic.

From a quick look, I don't think that implementation is entirely right (a bag should not equal a different type of collection, you'd need to initialize the other bag as well to make sure the equals is symmetric), but I will definitely admit that looks a lot simpler than what I wrote. 

> Where is this "contention" you speak of? Link? Pics or it never happened.

https://hibernate.atlassian.net/plugins/servlet/mobile?originPath=%2Fbrowse%2FHHH-5409#issue/HHH-5409

(Also, the fact that PersistentSet does not behave the same.) 

> I don't think that's right, actually. Here's the contract of `Collection.equals()`

Good point. I'd still say it violates the expectation that an implementation of Collection brings (as well as the List contract, abuse or not), but you're right that the actual contract of the Collection interface is not broken.

> That's quite correct. An entity equals() that depends on a collection is almost certainly completely broken.

A bold statement. I clearly disagreed three years ago, but I'm not sure yet where I stand today. However, that claim was made without any arguments to back it up.

JasperHorn:
(To be clear: I still think that implementing equals in way that a user would expect is better and that the inconsistency between the different collection types is a bad idea. Also, comments like "any code that does x must be broken" (without by arguments why that would be so) and "pics or it didn't happen" defenitely drew me in emotionally. However I have little to do with the code that was affected by this behavior these days, and I'm not actually that invested in whether Hibernate implements bags in a way that I think is sensible anymore.)

gavinking:
> Also, comments like "any code that does x must be broken" (without by arguments why that would be so)

I already wrote those arguments down in a book almost 20 years ago, and I'm not going to copy and paste them here.

gavinking:
> I do agree maybe it is time to reconsider allowing a `List` attribute to be a bag.

Doesn't the JPA spec require us to support `List`s where the indices are not persistent?

gavinking:
A potential solution to this would be to have a different implementation of `equals()` when the attribute has declared type `List`. i.e. we keep the current `equal()` when its declared type is `Collection`, but specialize the implementation comply with the semantics of the `List` interface when the program explicitly declares that type.

That would give people a choice.

(But we certainly shouldn't make a change like that in H5, since it would break truly ancient behavior that *many* programs rely on.)

sebersole:
> > I do agree maybe it is time to reconsider allowing a `List` attribute to be a bag.
> 
> Doesn't the JPA spec require us to support `List`s where the indices are not persistent?

Its been a while since I looked but I believe the expectation is to treat that a

sebersole:
> > I do agree maybe it is time to reconsider allowing a `List` attribute to be a bag.
> 
> Doesn't the JPA spec require us to support `List`s where the indices are not persistent?

It does, but for JPA that is specifically what we would call a PersistentList just with an implicit ordering

gavinking:
Ah OK! For years I assumed that those would be bags!

So with annotation-based mapping when do you actually get a bag? Only when you declare the type `Collection`?

If that’s the case, then I agree, we *could* remove the ability to narrow to `List` in H6.

JasperHorn:
> I already wrote those arguments down in a book almost 20 years ago, and they're very well-known in the Hibernate community. I don't feel the need to copy and paste them here.

"I don't want to take five minutes to even briefly list my arguments, so you should just take hours to read my book (and I'm not even going to tell you what the name of my book is)" and "others know my arguments, you should too" are absolutely two of the strongest arguments I've ever heard. /s

JasperHorn:
I should probably just unsubscribe from this thread. I don't appreciate the lack of willingness to participate in constructive dialogue I'm experiencing and you guys probably don't appreciate my sarcasm either, so a world in which this conversation isn't continued is probably the better one for each of us.

gavinking:
Look, Vlad already very politely and constructively explained to you the recommended way to implement `equals()` for an entity class. He even linked to a blog post which covers the same topic. If you want a more extensive  discussion of this issue, you can find it in the book.

But this is an issue tracker. It’s not the appropriate place to rediscuss stuff that we already figured out almost 20 years ago and have no good reason to change now.

This isn’t a discussion forum. I hope you can be understanding of that.

gavinking:
Anyway, I’m confident that if you yourself spend about an hour thinking deeply and clearly about what “equality” means for two objects which both represent rows of a database table, you’ll arrive at the exact same conclusion that we all have. Hint: Lombok can’t possibly generate a correct definition.

sebersole:
> Ah OK! For years I assumed that those would be bags!
> 
> So with annotation-based mapping when do you actually get a bag? Only when you declare the type `Collection`?
> 
> If that’s the case, then I agree, we _could_ remove the ability to narrow to `List` in H6.

We interpret it as a bag unless asked to "be JPA compliant" (`org.hibernate.jpa.spi.JpaCompliance#isJpaListComplianceEnabled`)

We should change this in 6.0 such that -

* `Collection` -> BAG
* `List` -> LIST

To be honest I thought I had made that change already.  Maybe it was one of the previous 6.0 incarnations.

gavinking:
@sebersole that sounds good to me

gavinking:
Actually on second thoughts let me just document here for posterity why I think this is the best resolution:

1. First of all, I don't think there's a _whole lot_ a priori wrong with the original proposal to having `PersistentBag` implement bag semantics in it's `equals()` methods. The only thing that really _is_ wrong with it is that to initially create a bag, you have to use `new ArrayList()`, which means instantiating something which _doesn't_ implement bag semantics in its `equals()` method. And that makes it a bit risky to depend on the semantics of `bag.equals()`.
2. On the other hand, there's nothing wrong with the existing semantics either, at least not as long as the bag is typed to `Collection`. By _not_ providing by-value equality for bags, we're discouraging people from relying on it. Which seems better to me. 

If we were starting from scratch I guess either choice could be reasonable, but we're not, and changing from 2 to 1 would break existing code for no good reason.

Now, the original allegation that `equals()` is currently broken isn't correct for bags typed to `Collection`, for all the reasons pointed to above. (In particular, the `Collection` itself explicitly allows this implementation.)

On the other hand, it _is_ quite correct to say that it's broken for bags typed to `List`, simply because the `List` interface comes with a well-defined contract for by-value equality. So that looks bad.

But then one immediately has to ask the question of why the hell do we even let people use `List` to represent a bag in the first place? This was a truly ancient decision, and I don't recall clearly that why I decided on that, but I think I can reconstruct my reasoning:

It was likely because you anyway needed to use `new ArrayList()` as the most convenient way to initially construct a bag. And I guess I figured that bags aren't really *that* different to lists, apart from the expectation that lists have a well-defined order. So I would have figured that the only real difference between a `List` and a bag was the _persistency_ of the list order.

But now that we're using `PersistentList` in that same circumstance, at least when you set `isJpaListComplianceEnabled()`. And of course the only really visible difference there is that it gives us a nonbroken implementation of `List.equals()`. And that seems better to me. So I don't see a good reason to not just turn it on by default, or even better simply remove the setting so that it's "always on" in H6.

You'll still be able to get a `PersistentBag` by using the type `Collection`, so I don't think we lose anything by making that change.

sebersole:
I agree that *if we start making the distinction here between `List` and `Collection`* then we can drop this setting and simply make the interpretation we already stated (above)[https://github.com/hibernate/hibernate-orm/pull/2046#issuecomment-789659761]

gavinking:
> I agree that _if we start making the distinction here between `List` and `Collection`_ then we can drop this setting and simply make the interpretation we already stated [above](https://github.com/hibernate/hibernate-orm/pull/2046#issuecomment-789659761)

Yup.

sebersole:
BTW, we are sort of close to that already - 

```

		else if ( java.util.Collection.class.equals( clazz ) ) {
			if ( property.isAnnotationPresent( CollectionId.class ) ) {
				return new IdBagBinder();
			}
			else {
				return new BagBinder();
			}
		}
		else if ( java.util.List.class.equals( clazz ) ) {
			if ( isIndexed ) {
				if ( property.isAnnotationPresent( CollectionId.class ) ) {
					throw new AnnotationException(
							"List do not support @CollectionId and @OrderColumn (or @IndexColumn) at the same time: "
									+ StringHelper.qualify( entityName, property.getName() ) );
				}
				return new ListBinder();
			}
			else if ( property.isAnnotationPresent( CollectionId.class ) ) {
				return new IdBagBinder();
			}
			else {
				return new BagBinder();
			}
		}
```

I think the only part we'd really want to change is:

1) disallow List + CollectionId
2) the final else-block as part of List handling returning BagBinder

gavinking:
Sounds good to me.

vladmihalcea:
> It was likely because you anyway needed to use new ArrayList() as the most convenient way to initially construct a bag. And I guess I figured that bags aren't really that different to lists, apart from the expectation that lists have a well-defined order. So I would have figured that the only real difference between a List and a bag was the persistency of the list order, which is something that arguably doesn't need to be represented in the type system.

Yes, that's also how I've been thinking of Bags and Lists. The `PersistentList` allowed us to materialize the element position while the `PersistentBag` didn't.

Now, most often, a `List` collection is not mapped to persist the ordering, and that's exactly what @sebersole said with _PersistentList just with an implicit ordering_. 

I also agree that the current `equals` behavior was not adhering to the _principle of least astonishment_, so Steve's change will surely provide the expected behavior.

gavinking:
Seems to me we're all in violent agreement.

PR #2048, opened 2017-11-07
HHH-12083: Fix test failure in InsertOrderingWithBidirectionalOneToManyFlushProblem on SAP HANA

breglerj:
- "TOP" is a reserved key word on HANA, so the name of an entity named "Top" must be quoted.


vladmihalcea:
Applied upstream, thanks

PR #2050, opened 2017-11-09
HHH-9241 - Allow declaring non-java.util Collection interfaces

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-9241


vladmihalcea:
We can only integrate this one as soon as [HCANN-65](https://hibernate.atlassian.net/browse/HCANN-65) gets fixed.

PR #2051, opened 2017-11-09
Pull Request for HHH-12076

randymay:


vladmihalcea:
Test case, applied upstream, thanks. It only fails for HBM. It works just fine for annotations.

PR #2052, opened 2017-11-10
HHH-12086 Batch order_inserts: flush during transaction causes incorrect insert ordering and subsequent constraint violation

awallgren:


vladmihalcea:
Cherry-picked the test case, fixed it, and applied upstream. 

Thanks for providing the new test case.

PR #2053, opened 2017-11-11
Test case for https://hibernate.atlassian.net/browse/HHH-12088

awallgren:


vladmihalcea:
We have lots of integration tests with all sorts of complex mappings that prove that flushing works when JDBC batching is not activated. Hence, I'm going to reject this test case.

PR #2054, opened 2017-11-14
HHH-12007 : Executing JPA query results in persistence of new/unmanaged entity

gbadner:
https://hibernate.atlassian.net/browse/HHH-12007

The code in ```IncrediblySillyJpaMapsIdMappedIdentifierValueMarshaller#persistTransientEntity``` is getting executed when a transient entity with a transient or detached key-many-to-one is bound to a query.

This PR removes ```IncrediblySillyJpaMapsIdMappedIdentifierValueMarshaller#persistTransientEntity```.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/tuple/entity/AbstractEntityTuplizer.java
+++ b/hibernate-core/src/main/java/org/hibernate/tuple/entity/AbstractEntityTuplizer.java
@@ -7,7 +7,6 @@
 package org.hibernate.tuple.entity;
 
 import java.io.Serializable;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/tuple/entity/AbstractEntityTuplizer.java
```diff
@@ -404,6 +403,13 @@ public void setIdentifier(Object entity, Serializable id, EntityMode entityMode,
...
 						// otherwise look for an initialized version
 						association = persistenceContext.getEntity( entityKey );
+						if ( association == null ) {
+							// get the association out of the entity itself
+							association = sessionFactory.getMetamodel().entityPersister( entityName ).getPropertyValue(
+									entity,
+									virtualIdComponent.getPropertyNames()[i]
+							);
+						}
 					}
```
gbadner: Without the code added above, ```org.hibernate.test.annotations.derivedidentities.e1.a.DerivedIdentitySimpleParentIdClassDepTest``` fails because ```association == null```.

gbadner:
Merged.

PR #2056, opened 2017-11-15
HHH-9933 - Uninitialized PersistentSet with lazy="extra" doesn't respect added elements

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-9933


vladmihalcea:
@gbadner Would you mind review this PR and let me know if you agree with the proposed change.

PR #2057, opened 2017-11-16
Hhh 12076

randymay:


vladmihalcea:
Thanks for the new use case, but you need to create a new Jira issue for this one because it's a different problem.

beikov:
Thanks for the test. This is superseded by https://github.com/hibernate/hibernate-orm/pull/3818

PR #2059, opened 2017-11-18
HHH-12102 ReaderInputStream.read() return incorrect value when reader…

cntsw:
[https://hibernate.atlassian.net/browse/HHH-12102](https://hibernate.atlassian.net/browse/HHH-12102)


vladmihalcea:
Check out the [comment on Jira](https://hibernate.atlassian.net/browse/HHH-12102). It's not clear why you need this change.

PR #2060, opened 2017-11-20
HHH-12105 Batch order_inserts: flush during transaction causes incorrect insert ordering and subsequent constraint violation

awallgren:
Test case for https://hibernate.atlassian.net/browse/HHH-12105


vladmihalcea:
Applied upstream, thanks. It's fixed by [this commit](https://github.com/hibernate/hibernate-orm/commit/7a32ea62b610e3a0fb4580471f37730791d452c7).

PR #2061, opened 2017-11-20
HHH-12120 Update StandardDialectResolver to reflect the latest Dialects of PostgreSQL and MySQL

quaff:
Use PostgreSQL95Dialect instead of default PostgreSQL81Dialect

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/dialect/internal/StandardDialectResolver.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/dialect/internal/StandardDialectResolver.java
@@ -32,6 +32,7 @@ import org.hibernate.dialect.PostgreSQL81Dialect;
 import org.hibernate.dialect.PostgreSQL82Dialect;
 import org.hibernate.dialect.PostgreSQL92Dialect;
 import org.hibernate.dialect.PostgreSQL94Dialect;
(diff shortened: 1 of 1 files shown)


gsmet:
It makes me think that maybe we should have the default the other way around i.e. return the latest version if the version is unknown (and be careful to return the old dialect for the old versions).

This way, we would have best effort support for the new versions of a RDBMS instead of using the oldest versions of the dialect.

What do you all think?

quaff:
Totally agree!

ajrepp:
Just curious -- are there plans to add dialects for Postgres 9.6, 10.0, and 10.1? How is it decided when a new dialect is needed?

- 9.6: https://www.postgresql.org/docs/9.6/static/release-9-6.html
- 10.0: https://www.postgresql.org/docs/10/static/release-10.html
- 10.1: https://www.postgresql.org/docs/10/static/release-10-1.html

vladmihalcea:
A new Dialect is added if there is a new database feature that Hibernate could use for implementing some specific persisting mechanism.

For instance, Oracle 12c now supports SQL Standard pagination. That's a good indication that we need a new Dialect for 12c.

vladmihalcea:
@ajrepp Superseded by this [PR](https://github.com/hibernate/hibernate-orm/pull/2066).

PR #2062, opened 2017-11-20
HHH-12087: Oracle dialect fixed for paging data from complex views

gunboxer:


vladmihalcea:
Applied upstream, thanks

PR #2064, opened 2017-11-23
HHH-12121 testcases for lazy loaded entities with final property getters

ikettu:
2/3 of these tests should fail now.


ikettu:
Accidentally create PR from wrong branch. Sorry.

PR #2066, opened 2017-11-23
HHH-12120 - Update StandardDialectResolver to reflect the latest Dialects of PostgreSQL and MySQL

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12120


vladmihalcea:
@ajrepp @gsmet This PR is based on the discussions that came from [this original PR](https://github.com/hibernate/hibernate-orm/pull/2061).

Let me know what you think.

hibernate-core/src/test/java/org/hibernate/dialect/resolver/DialectFactoryTest.java
```diff
@@ -136,9 +142,20 @@ public void testPreregisteredDialects() {
...
 		testDetermination( "PostgreSQL", PostgreSQL81Dialect.class, resolver );
 		testDetermination( "PostgreSQL", 8, 2, PostgreSQL82Dialect.class, resolver );
+		testDetermination( "PostgreSQL", 8, 3, PostgreSQL82Dialect.class, resolver );
 		testDetermination( "PostgreSQL", 9, 0, PostgreSQL9Dialect.class, resolver );
+		testDetermination( "PostgreSQL", 9, 1, PostgreSQL9Dialect.class, resolver );
+		testDetermination( "PostgreSQL", 9, 2, PostgreSQL92Dialect.class, resolver );
+		testDetermination( "PostgreSQL", 9, 3, PostgreSQL92Dialect.class, resolver );
+		testDetermination( "PostgreSQL", 9, 4, PostgreSQL94Dialect.class, resolver );
+		testDetermination( "PostgreSQL", 9, 5, PostgreSQL95Dialect.class, resolver );
+		testDetermination( "PostgreSQL", 10, 0, PostgreSQL95Dialect.class, resolver );
```
ajrepp: This change seems good to me. I'm a little unclear on the goal of this test class -- is it trying to check the dialect for every major.minor version, or just a few? For example, we're not checking versions 8.1, 9.6, or 10.1 of PostgreSQL. I know that new versions are released periodically, which would mean coming back to this test to add a test case for that version, so perhaps we're just trying to check some sample?

vladmihalcea: The goal of this test case is to prove that versions are picked correctly. I haven't listed every possible version, but, if 10.0 is resolved, 10.1 should be resolved as well.

ajrepp: LGTM then!

PR #2067, opened 2017-11-27
HHH-12115 - attribute converter add mapping of Java Short type to JDB…

frankggyy:
HHH-12115 - attribute converter add mapping of Java **Short** type to JDBC **SMALLINT** type.


vladmihalcea:
Applied upstream, thanks

PR #2068, opened 2017-11-28
HHH-10418 : test case

gbadner:
Test case for https://hibernate.atlassian.net/browse/HHH-10418.

This test case reproduces a failure on startup when a collection region uses the same name as an entity region.

A similar test case works on 4.2: https://github.com/gbadner/hibernate-core/tree/HHH-10418-4.2-test


galderz:
Added test and fix to #2083

PR #2071, opened 2017-11-28
HHH-12140 Allow session scoped interceptors to be managed by DI providers (e.g Spring)

ed0906:
This allows Session Scoped Interceptors to be managed by third party IOC frameworks e.g. Spring/Guice

- Instead of providing `"hibernate.session_factory.interceptor={class name}"` and the interceptor being created by `{class name}.newInstance()` in the session builder, you can now provide a `Supplier<Interceptor>` in your hibernate properties/Spring config XML/other framework configuration.

e.g. in Spring:

```
@Autowired
private Supplier<Interceptor> customSessionScopedInterceptorSupplier;

@Bean
public LocalContainerEntityManagerFactoryBean entityManagerFactory() {
    //... 
    final Map<String, Object> properties = new HashMap<>();
    //.... Add all JPA Properties
    properties.put("hibernate.session_factory.interceptor", customSessionScopedInterceptorSupplier);
    factory.setJpaPropertyMap(properties);
    return factory;
}
```

Note: This change maintains existing functionality allowing a user to specify a `Class` name and just extends this functionality to allow an `Interceptor<Supplier>` as well.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/SessionFactoryBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/SessionFactoryBuilder.java
@@ -6,8 +6,6 @@
  */
 package org.hibernate.boot;
 
-import java.util.Map;
-
 import org.hibernate.ConnectionReleaseMode;
 import org.hibernate.CustomEntityDirtinessStrategy;
 import org.hibernate.EntityMode;
@@ -27,6 +25,9 @@ import org.hibernate.resource.jdbc.spi.PhysicalConnectionHandlingMode;
 import org.hibernate.resource.jdbc.spi.StatementInspector;
 import org.hibernate.tuple.entity.EntityTuplizer;
 import org.hibernate.tuple.entity.EntityTuplizerFactory;
+
+import java.util.Map;
+import java.util.function.Supplier;
 
 /**
  * The contract for building a {@link org.hibernate.SessionFactory} given a number of options.
@@ -128,16 +129,16 @@ public interface SessionFactoryBuilder {
 	SessionFactoryBuilder applyInterceptor(Interceptor interceptor);
 
 	/**
-	 * Names an interceptor Class to be applied to the SessionFactory, which in turn means it will be used by all
+	 * Names an interceptor Supplier to be applied to the SessionFactory, which in turn means it will be used by all
 	 * Sessions unless one is explicitly specified in {@link org.hibernate.SessionBuilder#interceptor}
 	 *
-	 * @param statelessInterceptorClass The interceptor class
+	 * @param statelessInterceptorSupplier The interceptor supplier
 	 *
 	 * @return {@code this}, for method chaining
 	 *
 	 * @see org.hibernate.cfg.AvailableSettings#SESSION_SCOPED_INTERCEPTOR
 	 */
-	SessionFactoryBuilder applyStatelessInterceptor(Class<? extends Interceptor> statelessInterceptorClass);
+	SessionFactoryBuilder applyStatelessInterceptor(Supplier<? extends Interceptor> statelessInterceptorSupplier);
 
 	/**
 	 * Names a StatementInspector to be applied to the SessionFactory, which in turn means it will be used by all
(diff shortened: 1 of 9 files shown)


hibernate-core/src/main/java/org/hibernate/boot/SessionFactoryBuilder.java
```diff
@@ -128,16 +129,16 @@
...
 	 *
-	 * @param statelessInterceptorClass The interceptor class
+	 * @param statelessInterceptorSupplier The interceptor supplier
 	 *
 	 * @return {@code this}, for method chaining
 	 *
 	 * @see org.hibernate.cfg.AvailableSettings#SESSION_SCOPED_INTERCEPTOR
 	 */
-	SessionFactoryBuilder applyStatelessInterceptor(Class<? extends Interceptor> statelessInterceptorClass);
+	SessionFactoryBuilder applyStatelessInterceptor(Supplier<? extends Interceptor> statelessInterceptorSupplier);
```
vladmihalcea: We should not modify the existing API. We could add a new method, preferably a default one.

vladmihalcea:
As mentioned in the comment I added during the review, we don't want to break backward compatibility.

So, we need to retain the previous functionality where the user provides a Class. Internally, we can make it work as a Supplier, even when the user has provided a Class.

hibernate-core/src/main/java/org/hibernate/boot/SessionFactoryBuilder.java
ed0906: Would you have a `SessionFactoryBuilder applyStatelessInterceptor(Supplier<? extends Interceptor> statelessInterceptorSupplier);` on the interface above too?

ed0906: I'm not sure how we could have the `SessionFactoryOptions` as it was and incorporate the `Supplier` as its old signature is

`Class<? extends Interceptor> getStatelessInterceptorImplementor();`

Meaning we'd have to convert from supplier to` Class<? extends Interceptor>`

vladmihalcea: For `applyStatelessInterceptor`, we can have 2 overloaded methods. The same for the get part:

- `getStatelessInterceptorImplementor`
- `getStatelessInterceptorSupplier`

So, when building an Interceptor, we check both. If the user supplies both, we should raise an Exception. Otherwise, we use whatever was provided.

This way, we extend the current functionality by adding the new feature instead of modifying the existing interface API or the existing logic that is already in use.

ed0906: I'll have to add a `getStatelessInterceptorSupplier` method to the `SessionFactoryOptionsState` and `SessionFactoryOptions` interfaces otherwise it wont work. Either way this will still break backwards compatibility as upgrading users will have to implement the new supplier methods.

vladmihalcea: Not necessarily. We could add a default method to the interfaces that simply returns a lazy Supplier that delegates the call to `getStatelessInterceptorImplementor`.
Then, in the `SessionFactoryBuilderImpl` we overwrite the default method and add your logic.

ed0906: @vladmihalcea I've pushed the changes you reccommended

vladmihalcea:
Applied upstream, thanks

sebersole:
I'd have much preferred that this been handled via our "managed bean" (CDI) support.  This is generic enough that it should have been fairly simple to implement for Spring at least (Spring at least is a CDI container anyway iiuc).

vladmihalcea:
@sebersole I think the Spring and DI in the title are not really representative for this feature. Basically, we can just pass a `Supplier<Interceptor>` via the standard bootstrapping mechanisms.

So this can be used even without Spring or Java EE, via the native bootstrapping mechanism.

sebersole:
As can our CDI support via `ManagedBeanRepository`.  See? ;)

sebersole:
Think of `ManagedBeanRepository` being an abstraction over various things we directly instantiate.  If CDI support is enabled we first will try to locate the classes as CDI beans.  If there are not, we will fallback to normal instantiation

PR #2072, opened 2017-11-28
HHH-12132: HANA boolean type mapping doesn't work for existing schema definitions

breglerj:
https://hibernate.atlassian.net/browse/HHH-12132

- introduce parameter hibernate.dialect.hana.use_legacy_boolean_type to enable
  switching between new and legacy behavior.


Naros:
Applied upstream.

PR #2073, opened 2017-11-29
HHH-12107 - ClassCastException when using L2Cache with "structured_cache"=true

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12107


vladmihalcea:
The problem could be summarized as follows:

1. First, the `disassembledStateText` does not really store a description of the `disassembledState`, but of the `assembledState`. 
2.For the structured cache, when we reconstruct the `CacheEnty` from the underlying `Map`, we get the `disassembledState` back. However, we can't recreate the same  `disassembledStateText` which was set prior to saving to the 2nd-level cache since we don't really have the `assembledState`.
3. For the structured cache, we can set the `disassembledState` to null firsts, and recreate the `disassembledStateText` only after we know the `assembledState`.

vladmihalcea:
@sebersole @dreab8 @Naros Would you mind reviewing this issue and let me know if the approach is fine with you? Thanks.

PR #2074, opened 2017-11-30

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12106

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/mapping/Table.java
+++ b/hibernate-core/src/main/java/org/hibernate/mapping/Table.java
@@ -447,8 +447,24 @@ public class Table implements RelationalModel, Serializable, Exportable {
 		
 		final JdbcEnvironment jdbcEnvironment = metadata.getDatabase().getJdbcEnvironment();
 
+		Identifier quotedCatalog = catalog != null && catalog.isQuoted() ?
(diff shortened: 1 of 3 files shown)


vladmihalcea:
@dreab8 Do you mind reviewing it? Thanks.

hibernate-core/src/main/java/org/hibernate/mapping/Table.java
```diff
@@ -447,8 +447,24 @@ public Iterator sqlAlterStrings(
...
+
 		final String tableName = jdbcEnvironment.getQualifiedObjectNameFormatter().format(
-				tableInfo.getName(),
+				new QualifiedTableName(
+					quotedCatalog,
+					quotedSchema,
+					quotedTable
+				),
 				dialect
 		);
```
dreab8: I think instead of creating a new `QualifiedTableName` from the `TableInformation` may be enough to use the `Table#getQualifiedTableName() `.
```
final String tableName = jdbcEnvironment.getQualifiedObjectNameFormatter().format(
				getQualifiedTableName(),
				dialect
		);
```

dreab8:
a PR https://github.com/hibernate/hibernate-orm/pull/2091 with the change I proposed.
@vladmihalcea what do you think?

dreab8:
applied upstream
Thanks @vladmihalcea

PR #2075, opened 2017-11-30
HHH-12138: HANA dialect doesn't support procedure parameter mode REF_CURSOR

breglerj:
Added REF_CURSOR support


vladmihalcea:
Applied upstream, thanks

PR #2076, opened 2017-12-01
HHH-12143 - Using ConcurrentHashMap to implement properties

akarsh-jain:
When running with high concurrency, Hashtable.get() blocks multiple threads for properties. This causes failure when a single thread takes too much time to execute. A simple fix is to use ConcurrentHashMap that is not synchronized for get statements. 

Also created a JIRA : https://hibernate.atlassian.net/browse/HHH-12143

Thread Dump is attached that clearly shows multiple threads being blocked at Hashtable.get(). The corresponding Test Case that generates the Thread Dump too has been attached.

[thread-dump-100_threads.txt](https://github.com/hibernate/hibernate-orm/files/1520994/thread-dump-100_threads.txt)
[HibernateTestSample.zip](https://github.com/hibernate/hibernate-orm/files/1520997/HibernateTestSample.zip)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
@@ -136,11 +136,11 @@ public final class TypeFactory implements Serializable {
 	public static void injectParameters(Object type, Properties parameters) {
 		if ( ParameterizedType.class.isInstance( type ) ) {
 			if ( parameters == null ) {
-				ConcurrentHashMapBackedProperties	concurrentParams = new ConcurrentHashMapBackedProperties(parameters);
-				( (ParameterizedType) type ).setParameterValues( concurrentParams );
+				( (ParameterizedType) type ).setParameterValues( EMPTY_PROPERTIES );
 			}
 			else {
-				( (ParameterizedType) type ).setParameterValues( parameters );
+				ConcurrentHashMapBackedProperties concurrentParams = new ConcurrentHashMapBackedProperties(parameters);
+				( (ParameterizedType) type ).setParameterValues( concurrentParams );
 			}
 		}
 		else if ( parameters != null && !parameters.isEmpty() ) {
--- a/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
@@ -131,12 +131,13 @@ public final class TypeFactory implements Serializable {
 	}
 
 	// todo : can a Properties be wrapped in unmodifiable in any way?
-	private final static Properties EMPTY_PROPERTIES = new Properties();
+	private final static ConcurrentHashMapBackedProperties EMPTY_PROPERTIES = new ConcurrentHashMapBackedProperties();
 
 	public static void injectParameters(Object type, Properties parameters) {
 		if ( ParameterizedType.class.isInstance( type ) ) {
 			if ( parameters == null ) {
-				( (ParameterizedType) type ).setParameterValues( EMPTY_PROPERTIES );
+				ConcurrentHashMapBackedProperties	concurrentParams = new ConcurrentHashMapBackedProperties(parameters);
+				( (ParameterizedType) type ).setParameterValues( concurrentParams );
 			}
 			else {
 				( (ParameterizedType) type ).setParameterValues( parameters );
--- a/hibernate-core/src/main/java/org/hibernate/type/ConcurrentHashMapBackedProperties.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/ConcurrentHashMapBackedProperties.java
@@ -26,11 +26,11 @@ public class ConcurrentHashMapBackedProperties extends Properties {
 	 */
 	private static final long serialVersionUID = 23632462472472L;
 
-	private ConcurrentHashMap<Object,Object> properties = new ConcurrentHashMap<Object,Object>();
+	private ConcurrentHashMap<String,String> properties = new ConcurrentHashMap<String,String>();
 	
 
 	public ConcurrentHashMapBackedProperties(Properties properties){
-		this.properties.putAll(properties);
+		this.properties.putAll((Map)properties);
 		
 	}
 
--- a/hibernate-core/src/main/java/org/hibernate/type/ConcurrentHashMapBackedProperties.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/ConcurrentHashMapBackedProperties.java
@@ -1,0 +1,332 @@
+package org.hibernate.type;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.io.PrintWriter;
+import java.io.Reader;
+import java.io.Writer;
+import java.util.Collection;
+import java.util.Enumeration;
+import java.util.InvalidPropertiesFormatException;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+public class ConcurrentHashMapBackedProperties extends Properties {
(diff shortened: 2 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
```diff
@@ -131,12 +131,13 @@ public Type type(Class<Type> typeClass, Properties parameters) {
 	}
 
 	// todo : can a Properties be wrapped in unmodifiable in any way?
-	private final static Properties EMPTY_PROPERTIES = new Properties();
+	private final static ConcurrentHashMapBackedProperties EMPTY_PROPERTIES = new ConcurrentHashMapBackedProperties();
```
vladmihalcea: Why not using a `ConcurrentHashMap<String, String>` instead?

akarsh-jain: @vladmihalcea Added additional commit to use ConcurrentHashMap<String,String> in place of ConcurrentHashMap<Object,Object>. Since properties are of type Hashtable and always contain String types we can simply typecast to Map.

hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
```diff
@@ -131,12 +131,13 @@ public Type type(Class<Type> typeClass, Properties parameters) {
...
 
 	public static void injectParameters(Object type, Properties parameters) {
 		if ( ParameterizedType.class.isInstance( type ) ) {
 			if ( parameters == null ) {
-				( (ParameterizedType) type ).setParameterValues( EMPTY_PROPERTIES );
+				ConcurrentHashMapBackedProperties	concurrentParams = new ConcurrentHashMapBackedProperties(parameters);
```
vladmihalcea: Why do you use this new `ConcurrentHashMapBackedProperties` when the `parameters` is null, but not when the `parameters` is not null?

vladmihalcea:
First, we need a Jira issue to describe the problem and the context that generates it.

From the thread dump, it looks like the Properties#get call comes from your custom `UserType`:

    at com.myapp.entities.SecuredStringType.nullSafeGet(SecuredStringType.java:51)

not from Hibernate code.

We need a [replicating test case](http://in.relation.to/2016/01/14/hibernate-jpa-test-case-template/) with multiple threads to prove the problem is in Hibernate code and not your custom type.

hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
```diff
@@ -136,11 +136,11 @@ public Type type(Class<Type> typeClass, Properties parameters) {
 	public static void injectParameters(Object type, Properties parameters) {
 		if ( ParameterizedType.class.isInstance( type ) ) {
 			if ( parameters == null ) {
-				ConcurrentHashMapBackedProperties	concurrentParams = new ConcurrentHashMapBackedProperties(parameters);
-				( (ParameterizedType) type ).setParameterValues( concurrentParams );
+				( (ParameterizedType) type ).setParameterValues( EMPTY_PROPERTIES );
```
vladmihalcea: That's why we need a replicating test case. Otherwise, how do you know you are actually fixing an issue? This last commit is the perfect example of why we need a replicating test case.

akarsh-jain: @vladmihalcea The JIRA link and the replicating test case is already included in the pull request description. The last commit was made to rectify a miscommit i had made in the original.

vladmihalcea: Thanks. I'll investigate it next week and let you know of my findings.

akarsh-jain:
@vladmihalcea Added a JIRA issue. The Properties object that we are using in our custom user type comes from hibernate. The properties object uses HashTable which causes the blockage. If hibernate uses a properties object which better supports concurrent calls, this blockage would not occur. Hence, we propose that we change TypeFactory in hibernate to support concurrent calls.

vladmihalcea:
@akarsh-jain I checked your test case sample, and it's not what we need. That's a web application sample, and we need an isolated JUnit test case. Use [our template as described in this article](http://in.relation.to/2016/01/14/hibernate-jpa-test-case-template/).

The reason why we need a JUnit test is that we need to add that to Hibernate ORM Core and have the test running on every build.

Thanks.

vladmihalcea:
After reviewing your code, I realize that your problem comes from the way you are implementing the `UserType`, not from the choice of using `Properties` in Hibernate.

So, instead of:

```
public Object nullSafeGet(ResultSet rs, String[] names, SharedSessionContractImplementor session, Object owner)
		throws HibernateException, SQLException {
	String encryptedEntity = rs.getString(names[0]);
	String decryptedEntity = null;
	if (encryptedEntity != null) {
		if (encryptedEntity.startsWith(parameters.getProperty("key") + APPENDER)) {
			String key = parameters.getProperty("key");
			decryptedEntity = encryptedEntity.replaceFirst("^" + key + APPENDER, "");
		}
		else
		{
			decryptedEntity = encryptedEntity;
		}
	}
	return decryptedEntity;
}
```

You could move the `parameters.getProperty("key")` in a field:

```
private String key;

@Override
public void setParameterValues(Properties parameters) {
	this.parameters = parameters;
	this.key = parameters.getProperty("key");
}
```

And, you no longer need to access the `Properties` collection every time you use the `UserType`:

```
public Object nullSafeGet(ResultSet rs, String[] names, SharedSessionContractImplementor session, Object owner)
		throws HibernateException, SQLException {
	String encryptedEntity = rs.getString(names[0]);
	String decryptedEntity = null;
	if (encryptedEntity != null) {
		if (encryptedEntity.startsWith(key + APPENDER)) {
			String key = key;
			decryptedEntity = encryptedEntity.replaceFirst("^" + key + APPENDER, "");
		}
		else
		{
			decryptedEntity = encryptedEntity;
		}
	}
	return decryptedEntity;

}
```

That's it!

akarsh-jain:
@vladmihalcea According to design perspective such provision of a properties object is not sound. We have seen in several existing projects that the properties object is being used directly, rather than being set into more concurrence friendly objects. Now we can use the existing properties in our custom user types, but that would cause everyone to make such changes every time in every custom parameterized type. The identification of this issue would now require us to make several changes across the board including other open source projects. This is why we propose that we change TypeFactory in hibernate to provide a heavy-load friendly property object by default to the parameterized types.

vladmihalcea:
> According to design perspective such provision of a properties object is not sound. 

Please elaborate this statement. Doing the initialization in the `setParameterValues` phase is the right way to achieve your goal.

> We have seen in several existing projects that the properties object is being used directly, rather than being set into more concurrence friendly objects.

Open a new Jira issue for every use case you found, and we will investigate it.

> Now we can use the existing properties in our custom user types, but that would cause everyone to make such changes every time in every custom parameterized type. The identification of this issue would now require us to make several changes across the board including other open source projects. This is why we propose that we change TypeFactory in hibernate to provide a heavy-load friendly property object by default to the parameterized types.

That's the Hibernate `Type` contract. It was intended to be used like that.

akarsh-jain:
@vladmihalcea the fix that you have suggested does make sense, but it does not solve the issue of latency in all concurrent calls. We are merely suggesting an improvement wherein hibernate also uses concurrent classes. This will only improve the current state of affairs where instead of explicitly changing the underlying data structure after it is passed from hibernate, we make it implicit.

vladmihalcea:
We need a real life use case to justify this change.

If you can provide an example where multiple threads would have to do that
and we can't fix it by accessing the properties during bootstrap, than you
should open a new Jira issue and we will investigate it.

akarsh-jain:
Of course we can fix it by using your methodology and it will work in every scenario. It is not wrong, rather it is an extra effort for every using parametrised type assuming hibernate will handle the problem of concurrency whereas it doesn't implicitly. Why cant we upgrade the Type contract itself to include Concurrent Hash Map?

sebersole:
Changing the actual Map impl used simply makes the situation somewhat
better.  But regardless, accessing a Map over and over and over for the
same key is never going to as efficient as accessing it once.  This is not
a compelling argument imo

PR #2077, opened 2017-12-04
HHH-2558 - Allow batching inserts for multi-table entities

dannnnn:
HHH-5797 - Improve batching for entity updates or deletes that use secondary tables

So, added small modifications to support Batching for Joined Entities

BatchBuilderImpl.java

IMO, No need to use NoBatchingBatch class, because
1.It duplicates the if/else condition for example from AbstractEntityPersister:3163 and
2.useBatch - variable is kind of responsible to use batch or not - keep single responsibility
jdbcBatchSizeToUse > 1 - condition can be checked in AbstractEntityPersister and adjust the useBatch variable
So: here we always return BatchingBatch object and we batch only when useBatch is TRUE:
if ( useBatch ) {
 update = session
 .getJdbcCoordinator()
 .getBatch( updateBatchKey ) --> Here we call the builder
 .getBatchStatement( sql, callable );
 }
 else {
 update = session
 .getJdbcCoordinator()
 .getStatementPreparer()
 .prepareStatement( sql, callable );
 }

 BatchingBatch.java
 configuredBatchSize is the property copied from batchSize passed in constructor.
 Having Batching per Joined tables we would have to multiply the batch size on the number of tables in inheritance.
 Since we don't have access to this info here in this class, we can multiply by number of statements in the AbstractBatchImpl.statements.
 A batch is per saved Entity which might have 2 or more levels of inheritance.
 Meaning that the AbstractBatchImpl.statements will have 1 statement per level of inheritance.

 StatementPreparerImpl.java
 There is no need to execute batch here because :
 1. It breaks of putting SQL statements condition from AbstractBatchImpl.getBatchStatement:
 PreparedStatement statement = statements.get( sql );
 if ( statement == null ) {
 statement = buildBatchStatement( sql, callable ); --> as soon as we get different SQL insert for example, it will execute the batch.
 statements.put( sql, statement );
 }
 else {
 LOG.debug( "Reusing batch statement" );
 sqlStatementLogger().logStatement( sql );
 }
 Having a save of a 2 level inheritance Entity: Employee -> Person. Batch will always have 1 statement.
 So by removing this we allow multiple PreparedStatements in AbstractBatchImpl.statements.
 So now in BatchingBatch.performExecution(), the for loop makes sense(there where always 1 interation because of 1 statement).

 AbstractCollectionPersister.java
 use batch when its property is configured > 1

 BasicCollectionPersister.java
 use batch when its property is configured > 1

 AbstractEntityPersister.java
 for insert and update use batch when its property is configured > 1

 NonBatchingBatchFailureTest.java
 BatchBuilder was the only place where we use NonBatchingBatch.class.
 Having the usage of it removed,class can be marked @depricated and imo the tests can be ignored

 InsertOrderingWithJoinedTableInheritance.java
 InsertOrderingWithJoinedTableMultiLevelInheritance.java
 Fixed the test, it does less prepared statements

 BatchingTest.java
 Ignore temporarily the test using NonBatchingBatch.java

 JoinedSubclassBatchingTest.java
 Add test Joined Entities, same as for Single table Entities

I posted in forum, hopefully we can make Hibernte better
https://forum.hibernate.org/viewtopic.php?f=1&t=1045088

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchBuilderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchBuilderImpl.java
@@ -66,9 +66,26 @@ public class BatchBuilderImpl implements BatchBuilder, Configurable, Manageable,
 		final int jdbcBatchSizeToUse = sessionJdbcBatchSize == null ?
 				this.jdbcBatchSize :
 				sessionJdbcBatchSize;
-		return jdbcBatchSizeToUse > 1
-				? new BatchingBatch( key, jdbcCoordinator, jdbcBatchSizeToUse )
-				: new NonBatchingBatch( key, jdbcCoordinator );
+/**
+IMO, No need to use NoBatchingBatch class, because
+1.It duplicates the if/else condition for example from AbstractEntityPersister:3163 and
+2.useBatch - variable is kind of responsible to use batch or not - keep single responsibility
+jdbcBatchSizeToUse > 1 - condition can be checked in AbstractEntityPersister and adjust the useBatch variable
+So: here we always return BatchingBatch object and we batch only when useBatch is TRUE:
+if ( useBatch ) {
+ update = session
+ .getJdbcCoordinator()
+ .getBatch( updateBatchKey ) --> Here we call the builder
+ .getBatchStatement( sql, callable );
+ }
+ else {
+ update = session
+ .getJdbcCoordinator()
+ .getStatementPreparer()
+ .prepareStatement( sql, callable );
+ }
+ **/
+		return new BatchingBatch( key, jdbcCoordinator, jdbcBatchSizeToUse );
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
@@ -29,6 +29,7 @@ import org.hibernate.cache.spi.entry.CacheEntryStructure;
 import org.hibernate.cache.spi.entry.StructuredCollectionCacheEntry;
 import org.hibernate.cache.spi.entry.StructuredMapCacheEntry;
 import org.hibernate.cache.spi.entry.UnstructuredCacheEntry;
+import org.hibernate.cfg.Environment;
 import org.hibernate.collection.spi.PersistentCollection;
 import org.hibernate.dialect.Dialect;
 import org.hibernate.engine.jdbc.batch.internal.BasicBatchKey;
@@ -48,6 +49,7 @@ import org.hibernate.internal.FilterAliasGenerator;
 import org.hibernate.internal.FilterHelper;
 import org.hibernate.internal.util.StringHelper;
 import org.hibernate.internal.util.collections.ArrayHelper;
+import org.hibernate.internal.util.config.ConfigurationHelper;
 import org.hibernate.jdbc.Expectation;
 import org.hibernate.jdbc.Expectations;
 import org.hibernate.loader.collection.CollectionInitializer;
@@ -183,6 +185,7 @@ public abstract class AbstractCollectionPersister
 	private final boolean isMutable;
 	private final boolean isVersioned;
 	protected final int batchSize;
+	protected final Integer configuredBatchSize;
 	private final FetchMode fetchMode;
 	private final boolean hasOrphanDelete;
 	private final boolean subselectLoadable;
@@ -290,6 +293,7 @@ public abstract class AbstractCollectionPersister
 		}
 		batchSize = batch;
 
+		this.configuredBatchSize = ConfigurationHelper.getInt(Environment.STATEMENT_BATCH_SIZE, this.factory.getProperties(),1);
 		isVersioned = collectionBinding.isOptimisticLocked();
 
 		// KEY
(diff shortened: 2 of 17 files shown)


hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchBuilderImpl.java
```diff
@@ -66,9 +66,26 @@ public Batch buildBatch(BatchKey key, JdbcCoordinator jdbcCoordinator) {
...
 				this.jdbcBatchSize :
 				sessionJdbcBatchSize;
-		return jdbcBatchSizeToUse > 1
-				? new BatchingBatch( key, jdbcCoordinator, jdbcBatchSizeToUse )
-				: new NonBatchingBatch( key, jdbcCoordinator );
+/**
+IMO, No need to use NoBatchingBatch class, because
+1.It duplicates the if/else condition for example from AbstractEntityPersister:3163 and
+2.useBatch - variable is kind of responsible to use batch or not - keep single responsibility
+jdbcBatchSizeToUse > 1 - condition can be checked in AbstractEntityPersister and adjust the useBatch variable
```
vladmihalcea: What if the dev provided a batch size of 1? I don't really see a need to change this right now.

hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
```diff
@@ -290,6 +293,7 @@ public AbstractCollectionPersister(
 		}
 		batchSize = batch;
 
+		this.configuredBatchSize = ConfigurationHelper.getInt(Environment.STATEMENT_BATCH_SIZE, this.factory.getProperties(),1);
```
vladmihalcea: Did you take into consideration that we can override that on a per-Session basis now? Check `SharedSessionContract#setJdbcBatchSize`

hibernate-core/src/test/java/org/hibernate/test/insertordering/InsertOrderingWithJoinedTableInheritance.java
```diff
@@ -97,8 +97,10 @@ public void testBatchingAmongstSubClasses() {
 			}
 			connectionProvider.clear();
 		} );
-
-		assertEquals( 26, connectionProvider.getPreparedStatements().size() );
+		/**
+		 * Having the fix, the test does less statements to the DB
+		 */
+		assertEquals( 4, connectionProvider.getPreparedStatements().size() );
```
vladmihalcea: Why are fewer statements executed for inheritance join table now?

hibernate-core/src/test/java/org/hibernate/test/joinedsubclassbatch/Person.hbm.xml
```diff
@@ -0,0 +1,82 @@
+<?xml version="1.0"?>
```
vladmihalcea: Better use annotations and embed the entity in the test class itself.

hibernate-core/src/test/java/org/hibernate/test/joinedsubclassbatch/Person.java
```diff
@@ -0,0 +1,109 @@
...
+import org.hibernate.annotations.AttributeAccessor;
+import org.hibernate.annotations.GenericGenerator;
+
+import javax.persistence.*;
+
+/**
+ * @author dcebotarenco
+ */
+public class Person {
+	@Id
```
vladmihalcea: Why did you choose to use HBM since you already use annotations?

vladmihalcea:
@dannnnn The changes are worth integrating. I'll do all the required modifications to integrate them, so you don't have to change anything. Thanks.

dannnnn:
Hi @vladmihalcea , THX for the reply and accepting the my changes, I was thinking to answer your questions tomorrow cause i have to dive back in IDEA and hibernate tests + code. So soon, you'll get the answers. Overall I think you got the idea about batching solution. Its all about keeping as a batch a list of statements and reuse them, so that the batch definition in hibernate can be redefined as:

The hibernate.jdbc.batch_size is the number of inserts, updates or deletes batched in prepared statements per an entity not per tables of that entities. Example:
Having Employee extending Person and Saving 20 Employees Hibernate will do 20 inserts in Person table and 20 inserts in Employee table, using only 2 prepared statements

hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchBuilderImpl.java
dannnnn: The idea is that i'd preffer `BatchBuilderImpl` to have the `buildBatch` method passing and the jdbcBatchSize. 
`public Batch buildBatch(BatchKey key, JdbcCoordinator jdbcCoordinator,int jdbcSize){}`
so that here we only create the batch. The check itself of creating a batch or not is moved on abstract layer in `AbstractEntityPersister`. Besides this , i noticed that the `NonBatchingBatch.class` kind of duplicates the logic from `AbstractEntityPersister.check()` method, that's it why makes no sense to use NonBatchingBatch class when we cant batch

hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
dannnnn: No i don't, I was searching for a better API from where i can get current configured jdbc batchsize, but i couldnt find it.

hibernate-core/src/test/java/org/hibernate/test/joinedsubclassbatch/Person.hbm.xml
dannnnn: It was a copy paste from an existing test related to batching where i just added my annotations

hibernate-core/src/test/java/org/hibernate/test/joinedsubclassbatch/Person.java
dannnnn: copy/paste

vladmihalcea:
Closing this Pull Request since I opened [this one](https://github.com/hibernate/hibernate-orm/pull/2082) instead.

PR #2078, opened 2017-12-05
HHH-9641 Publish JavaDoc JARs

smoothreggae:
[HHH-9641](https://hibernate.atlassian.net/browse/HHH-9641) is an old ticket and isn't assigned to anyone, but the required changes didn't seem too hard so I gave it a shot. I tested my changes with a `./gradlew clean assemble publishToMavenLocal` (which, before my changes only created the binary and source JARs). I'd be glad to do any more required testing and make any other changes necessary.

### DIFF
--- a/hibernate-entitymanager/hibernate-entitymanager.gradle
+++ b/hibernate-entitymanager/hibernate-entitymanager.gradle
@@ -30,3 +30,8 @@ sourcesJar {
 	manifest = null
 }
 
+javadocJar {
+	// The OSGi JAR manifest support does not like a non-existent classes dir,
+	// so make sure we dont use the OSGi one :)
+	manifest = null
+}
--- a/hibernate-java8/hibernate-java8.gradle
+++ b/hibernate-java8/hibernate-java8.gradle
@@ -22,3 +22,9 @@ sourcesJar {
     // so make sure we dont use the OSGi one :)
     manifest = null
 }
+
+javadocJar {
+    // The OSGi JAR manifest support does not like a non-existent classes dir,
+    // so make sure we dont use the OSGi one :)
+    manifest = null
+}
--- a/build.gradle
+++ b/build.gradle
@@ -353,6 +353,15 @@ subprojects { subProject ->
 	sourcesJar {
 		manifest = jar.manifest
 	}
+
+	task javadocJar(type: Jar) {
+		from new File(project(":documentation").buildDir, "/javadocs")
+		classifier = 'javadoc'
+		manifest = jar.manifest
+	}
+
+	javadocJar.dependsOn(":documentation:aggregateJavadocs")
+
 	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 
@@ -465,6 +474,7 @@ subprojects { subProject ->
 					artifact( sourcesJar ) {
 						classifier 'sources'
 					}
+					artifact javadocJar
 				}
 				// http://issues.gradle.org/browse/GRADLE-2966
 				// Once ^^ is resolved:


vladmihalcea:
Looks good. I wonder if there is anything we should change.

vladmihalcea:
I started a [discussion on the mailing list and will attach it here](http://lists.jboss.org/pipermail/hibernate-dev/2017-December/016838.html). Basically, this was intended and that's why we have aggregated Java Docs.

However, since this is already tested and it works, we might integrate this change as well. I'm waiting to see if there's anyone against it.

sebersole:
If you can resolve the conflicts I will integrate this next week.

hibernate-entitymanager/hibernate-entitymanager.gradle
```diff
@@ -30,3 +30,8 @@ sourcesJar {
 	manifest = null
 }
 
+javadocJar {
+	// The OSGi JAR manifest support does not like a non-existent classes dir,
+	// so make sure we dont use the OSGi one :)
+	manifest = null
+}
```
sebersole: Why publishing empty JavaDocs?

smoothreggae: I just replicated what was being done for the `sources` JAR. I could supress the generation of the JavaDoc JAR -- it's a simple change. I'll amend the PR.

hibernate-java8/hibernate-java8.gradle
```diff
@@ -21,4 +21,10 @@ sourcesJar {
     // The OSGi JAR manifest support does not like a non-existent classes dir,
     // so make sure we dont use the OSGi one :)
     manifest = null
-}
\ No newline at end of file
+}
+
+javadocJar {
```
smoothreggae: I also realised that it might be better to _suppress_ the generation of JavaDoc for the deprecated module `hibernate-java8` (it's getting the JavaDoc for `hibernate-core`). I'll amend the PR to do this.

hibernate-entitymanager/hibernate-entitymanager.gradle
sebersole: It is more a question.  No one on the team actually ever uses javadoc jars - in our opinion its silly when you can instead just use the sources jar.  So I am really asking.

sebersole:
I hope to do 5.3.0.Beta1 this week.  No rush, but if you want it in there...

smoothreggae:
I've updated the branch based on the latest in the upstream `master`.

I just realized something that should really have been obvious when I made the changes: the same all-encompassing JavaDoc JAR is being created for each of the sub-projects. Although this serves the need for a JavaDoc JAR, it seems _wrong_ to include JavaDoc for classes that are *not* part of a given artifact and it seems _wrong_ to publish such a large JAR for each sub-project.

I could do one of the following:
* decline the PR (since the team doesn't really use the JavaDoc JARs)
* see what it takes to get a more correct JavaDoc JAR for each sub-project without upsetting the existing build and then update the PR

Let me know which option you'd prefer.

sebersole:
To publish module-specific JavaDoc jars, obviously those JavaDoc should be
module-specific :)

smoothreggae:
Thanks for not choosing to _decline_ it outright :) Let me see what I can do to get some module-specific JavaDoc JARs.

sebersole:
Have you checked the Gradle documentation?  Generating javadoc jar is pretty straight-forward.

smoothreggae:
@sebersole, my changes initially reused what the build in this project already does for JavaDoc. I do plan to use the standard `javdoc` task from Gradle (I'll likely duplicate some of the configuration from `aggregateJavadocs` in `documentation/documentation.gradle`). I just haven't been able to get back to this over the last few days :)

sebersole:
There really should not be much to do for this - I'm not sure what you want to "duplicate".

Anyway, as we are also looking to move to Bintray for publishing release artifacts and syncing to Central we will need to start publishing javadocs as well since Central requires it.  We had previously gotten around that, but we will no longer be able to.  So I'll likely just do this

smoothreggae:
The duplication should only be in the configuration for the JavaDoc.

sebersole:
I implemented this myself : https://github.com/hibernate/hibernate-orm/pull/2097

smoothreggae:
Cool, @sebersole. I was just about to push a working alternative.

PR #2079, opened 2017-12-05
Test case and proposed fix for HHH-12011

reva:
Test case and proposed fix for HHH-12011 - Field annotated with target TYPE_USE break metamodel generation, see https://hibernate.atlassian.net/browse/HHH-12011

We probably shouldn't use com.sun.tools.javac.code.Type.AnnotatedType, but I found no other solution short of parsing the output of toString().


vladmihalcea:
First, using SUN classes is not the way to go and the code doe not even compile.

If we modify it to:

	if ( type.getKind().isPrimitive() ) {
		if ( type.getKind().isPrimitive() ) {
			typeName = context.getTypeUtils().getPrimitiveType(type.getKind()).toString();
		}
	}

The test still fails:

```
TypeUseAnnotationEntity_.java:3: error: <identifier> expected
import (@org.hibernate.jpamodelgen.test.annotation.TypeUseAnnotation :: java.lang.Byte);
      ^
```

Naros:
This PR is superseded by https://github.com/hibernate/hibernate-orm/pull/2270.

PR #2080, opened 2017-12-06
HHH-12149 - Allow to use predicate operators in HQL order by clauses

hangpark:
PR for [HHH-12149](https://hibernate.atlassian.net/browse/HHH-12149).


beikov:
Oracle doesn't support boolean expressions in the order by clause, so they should be emulated properly for that dialect by putting a `case when ... then 1 else 0 end` around the predicate.

PR #2085, opened 2017-12-14
Checkstyle licence header

dreab8:
I have added the shared/config/checkstyle/license.header file that contains the regex used inside shared/config/checkstyle/checkstyle.xml to match both version of the header 
/*
 * Hibernate, Relational Persistence for Idiomatic Java
 *
 * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
 * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
 */
or
/*
 * Hibernate, Relational Persistence for Idiomatic Java
 *
 * License: GNU Lesser General Public License (LGPL), version 2.1 or later
 * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
 */

The other changes add a new line at the end of some files that did't have it causing a checkstyle error.


dreab8:
The only think I do not like is that with this change the error messages contain the regexp.
By example if the header does not match  "* License: GNU Lesser General Public License (LGPL), version 2.1 or later." the error message will be
"Line does not match expected header line of '^ \* License: GNU Lesser General Public License \(LGPL\), version 2.1 or later(\.)?$'"
but not sure if this is a real problem.

sebersole:
I think the error message is fine.  It is clear that the issue is the license header, which is the important part.

sebersole:
@Naros You said you had a follow up to this?

Naros:
I have already integrated a solution on master and missed this PR.  Closing.

PR #2087, opened 2017-12-15
HHH-12175: ParameterParser doesn't handle JPA positional parameters correctly

breglerj:


sebersole:
Thanks Jonathan!

BTW, I have been re-thinking about allowing the Dialect define how parameters in native queries are recognized.  While I still do not like the idea of the Dialect just contributing a prefix to look for, I think I would be ok with allowing the Dialect to define the default NativeQueryInterpreter to use.  This is the contract that controls parameter parsing as well as other aspects of interpreting a native query.  I lost the PR # - could you give me that?

breglerj:
I haven't created a PR for the other issue (HHH-12174) yet. I can do it when I'm back in the office on Monday. Or, if you prefer, you can take the issue and I just implement the HANA-specific part.

PR #1152, opened 2015-11-16
HHH-10294 EntityGraph improvement: For each jpa attribute, generate also a String constant holding the attribute field name

reda-alaoui:
Fix for https://hibernate.atlassian.net/browse/HHH-10294

PR #2088, opened 2017-12-17
HHH-10294 EntityGraph improvement: For each jpa attribute, generate also a String constant holding the attribute field name

reda-alaoui:
Fix for https://hibernate.atlassian.net/browse/HHH-10294
Replaces https://github.com/hibernate/hibernate-orm/pull/1152.

Attributes are not interweaved anymore.
A use case has been added to `Student.java`

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/graphs/Student.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/graphs/Student.java
@@ -31,7 +31,7 @@ import org.hibernate.annotations.CacheConcurrencyStrategy;
 	@NamedEntityGraph(
 		name = "Student.Full",
 		attributeNodes = {
-			@NamedAttributeNode(value = "courses")
+			@NamedAttributeNode(value = Student_.COURSES)
 		}
 	)
 })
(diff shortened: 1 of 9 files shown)


hibernate-core/src/test/java/org/hibernate/jpa/test/graphs/Student.java
```diff
@@ -31,7 +31,7 @@
 	@NamedEntityGraph(
 		name = "Student.Full",
 		attributeNodes = {
-			@NamedAttributeNode(value = "courses")
+			@NamedAttributeNode(value = Student_.COURSES_)
```
vladmihalcea: Now, I understand why you needed a constant. I wonder if we should also append the underscore at the end.

reda-alaoui: You want me to remove the underscore?

sebersole: I personally find it odd as well.

reda-alaoui: I removed the underscore

reda-alaoui:
Thank you!
Can we land this in 5.x?

reda-alaoui:
Oh didn't see the fixed version in JIRA.
It seems ok for 5.x.

sebersole:
5.3 *is* 5.x ;)  Porting things back to 5.2 is Gail's decision.  I have
reached out to her about it

PR #2090, opened 2017-12-18
HHH-12177 - Drop hibernate-infinispan module

dreab8:
https://hibernate.atlassian.net/browse/HHH-12177

I'm not sure about the relocation GAV is correct but if so I think we need to update the documentation because `org.hibernate.cache.infinispan.InfinispanRegionFactory` is now `org.infinispan.hibernate.cache.InfinispanRegionFactory` and also add a note in the migration-guide.

@sebersole please try to publish it locally to check if what I have done is correct.

Thanks


dreab8:
Disabled the hibernate-infinispan jar generation.
Not sure about the change I made to hibernate-osgi/hibernate-osgi.gradle

sebersole:
About OSGi... I guess we should talk to the infinispan team about that.  That's likely something they should do.  For now, I'm inclined to just remove that OSGi module.  

@brmeyer what do you think?  the concern is that even if we include the new infinispan-hibernate jar, the class names are still wrong.  so its not a "drop in" replacement

sebersole:
I went ahead and committed this as is.  I am going to apply some small changes on top of that.  Thinking about it some more, the OSGi changes seem fine, unless we hear back differently from Brett.

PR #2091, opened 2017-12-19

dreab8:
https://hibernate.atlassian.net/browse/HHH-12106

an alternative solution to https://github.com/hibernate/hibernate-orm/pull/2074


sebersole:
@dreab8 I leave it up to you which to use.  IIUC however, both result in the generation of a QualifiedTableName

dreab8:
@sebersole I have double checked and my solution may have some issues related with schema and/or catalog. I'm dropping this PR.

PR #2092, opened 2017-12-20
ManagedBeanRegistry CDI lifecycle fixes

yrodiere:
@sebersole Here are the fixes I mentioned yesterday regarding the `ManagedBeanRegistry`. Mainly this is about making sure that, when `shouldRegistryManageLifecycle` is `false`, the registry really doesn't manage the CDI lifecycle at all and delegates to the CDI runtime.

All of the commits in this PR are follow-ups to already-fixed tickets:

 * https://hibernate.atlassian.net/browse/HHH-12133
 * https://hibernate.atlassian.net/browse/HHH-12171

**WARNING**: I had to use CDI 2.0 APIs, and those APIs are not available in WildFly 11. So I added a build step to patch the WildFly server.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/resource/beans/internal/JpaCdiLifecycleManagementStrategy.java
+++ b/hibernate-core/src/main/java/org/hibernate/resource/beans/internal/JpaCdiLifecycleManagementStrategy.java
@@ -1,0 +1,111 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.resource.beans.internal;
+
+import javax.enterprise.context.spi.CreationalContext;
+import javax.enterprise.inject.spi.AnnotatedType;
+import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.inject.spi.BeanManager;
+import javax.enterprise.inject.spi.InjectionTarget;
+
+import org.hibernate.resource.beans.spi.ManagedBean;
+
+class JpaCdiLifecycleManagementStrategy implements CdiLifecycleManagementStrategy {
+
+	static final JpaCdiLifecycleManagementStrategy INSTANCE = new JpaCdiLifecycleManagementStrategy();
+
+	private JpaCdiLifecycleManagementStrategy() {
+		// private constructor, do not use
+	}
+
+	@Override
+	public <T> ManagedBean<T> createBean(BeanManager beanManager, Class<T> beanClass) {
+		AnnotatedType<T> annotatedType = beanManager.createAnnotatedType( beanClass );
+		InjectionTarget<T> injectionTarget = beanManager.createInjectionTarget( annotatedType );
+		CreationalContext<T> creationalContext = beanManager.createCreationalContext( null );
+
+		T beanInstance = injectionTarget.produce( creationalContext );
+		injectionTarget.inject( beanInstance, creationalContext );
+
+		injectionTarget.postConstruct( beanInstance );
+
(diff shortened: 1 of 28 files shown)


yrodiere:
@sebersole Following @Sanne's advice, I managed to avoid the switch to WildFly 12. This PR now targets WildFly 11, with only a patch file applied to the server to upgrade Weld to a version supporting CDI 2.0.

hibernate-core/src/main/java/org/hibernate/resource/beans/internal/JpaCdiLifecycleManagementStrategy.java
```diff
@@ -0,0 +1,127 @@
...
+ * (i.e. when the bean lifecycle is to be managed by the JPA runtime, not the CDI runtime).
+ *
+ * The main characteristic of this strategy is that each requested bean is instantiated directly
+ * and guaranteed to not be shared in the CDI context.
+ *
+ * In particular, @Singleton-scoped or @ApplicationScoped beans are instantiated directly by this strategy,
+ * even if there is already an instance in the CDI context.
+ * This means singletons are not really singletons, but this seems to be the behavior required by
+ * the JPA 2.2 spec.
+ */
```
sebersole: But see, this is not true.  For a given class we only ever generate one ManagedBean instance = they are singletons (unless of course they are used again in a non-JPA context).  Notice the caching of them initially in a Map keyed by class...

yrodiere: > unless of course they are used again in a non-JPA context

My point exactly. They are singleton in the JPA context only, not in the CDI context overall. I can rephrase the comment to clarify that if you like.

yrodiere: Also, even in the JPA context, such "not-actually-singleton" beans will be instantiated once per `ServiceRegistry`, which unless I'm mistaken means once per `SessionFactory`/persistence unit. So even from the JPA point of view, they cannot always be viewed as singletons.

yrodiere:
Now working on tests as discussed; I couldn't finish today, but I'll probably be able to push something on Monday.

yrodiere:
Added non-regression tests as requested.

sebersole:
@yrodiere Just because it is easier to ask than to see it in the PR.... you are using CDI 2.0's SE bootstrap now, right?  Instead of using WildFly?

yrodiere:
@sebersole Yes. I did exactly what you did in your own tests.

sebersole:
I do still see CDI 2.0 patches against WildFly though :
https://github.com/hibernate/hibernate-orm/pull/2092/files#diff-1d01abe59e76ef186ae0415fefca546d

Is that just for completeness?  If we are not testing "CDI in WildFly" (and
we should not be), then this should not be needed.

yrodiere:
> Is that just for completeness? 

It is. If anything relies on CDI in your WildFly ITs, without this patch the IT will probably fail.

If you don't think you'll ever use CDI in your WildFly ITs, then yes you can skip this commit. I can remove it if you want.

sebersole:
We currently do not do any in-WF CDI tests (we used to, but I changed that
to use the SE bootstrapping).  IMO that's outside the scope of what we
should be testing.

I can just remove that as I integrate your PR locally.  Or you can if you
really want to :)

yrodiere:
> I can just remove that as I integrate your PR locally.  Or you can if you
really want to :)

I rebased on master and removed the commit.

sebersole:
Ok, thanks.  I am integrating this now.

sebersole:
@yrodiere I did some work on top of this as discussed and pushed upstream in prep for tomorrow's Beta1 release.  It's just a Beta, so if you find anything we need to change its not a big deal - we'll get it into the next one

PR #2093, opened 2017-12-20
HHH-12164 Upgrade Hibernate Validator used for testing to 6.0.7.Final

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12164

@sebersole here it is!


dreab8:
applied upstream.

Thanks @gsmet

PR #2097, opened 2017-12-22
HHH-12187 - Drop custom javadoc css;

sebersole:
HHH-9641 - Resume uploading Javadoc JARs to Maven


sebersole:
I'm going to just incorporate this.  There is nothing here that we need to discuss I think

PR #2099, opened 2017-12-27
HHH-12192: include the causing DDL-statement into the message of org.…

bbodnar:
…hibernate.tool.schema.spi.CommandAcceptanceException


vladmihalcea:
Applied upstream, thanks

PR #2084, opened 2017-12-14
HHH-10418 Support entity/collection sharing same region (5.1.x)

galderz:
https://hibernate.atlassian.net/browse/HHH-10418

PR #2100, opened 2018-01-05
HHH-10418 (5.1)

gbadner:
https://hibernate.atlassian.net/browse/HHH-10418

Supersedes https://github.com/hibernate/hibernate-orm/pull/2084

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
@@ -8,6 +8,9 @@ package org.hibernate.internal;
 
 import org.hibernate.HibernateException;
 import org.hibernate.boot.spi.SessionFactoryOptions;
+import org.hibernate.cache.spi.CollectionRegion;
+import org.hibernate.cache.spi.EntityRegion;
+import org.hibernate.cache.spi.NaturalIdRegion;
 import org.hibernate.cache.spi.QueryCache;
 import org.hibernate.cache.spi.Region;
 import org.hibernate.cache.spi.RegionFactory;
@@ -39,7 +42,10 @@ public class CacheImpl implements CacheImplementor {
 	private final transient RegionFactory regionFactory;
 	private final transient UpdateTimestampsCache updateTimestampsCache;
 	private final transient ConcurrentMap<String, QueryCache> queryCaches;
-	private final transient ConcurrentMap<String, Region> allCacheRegions = new ConcurrentHashMap<String, Region>();
+	private final transient ConcurrentMap<String, EntityRegion> entityRegionMap = new ConcurrentHashMap<String, EntityRegion>();
+	private final transient ConcurrentMap<String, CollectionRegion> collectionRegionMap = new ConcurrentHashMap<String, CollectionRegion>();
+	private final transient ConcurrentMap<String, NaturalIdRegion> naturalIdRegionMap = new ConcurrentHashMap<String, NaturalIdRegion>();
+	private final transient ConcurrentMap<String, Region> otherRegionMap = new ConcurrentHashMap<String, Region>();
 
 	public CacheImpl(SessionFactoryImplementor sessionFactory) {
 		this.sessionFactory = sessionFactory;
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
```diff
@@ -39,7 +42,10 @@
 	private final transient RegionFactory regionFactory;
 	private final transient UpdateTimestampsCache updateTimestampsCache;
 	private final transient ConcurrentMap<String, QueryCache> queryCaches;
-	private final transient ConcurrentMap<String, Region> allCacheRegions = new ConcurrentHashMap<String, Region>();
+	private final transient ConcurrentMap<String, EntityRegion> entityRegionMap = new ConcurrentHashMap<String, EntityRegion>();
+	private final transient ConcurrentMap<String, CollectionRegion> collectionRegionMap = new ConcurrentHashMap<String, CollectionRegion>();
```
scottmarlow: @Sanne @rvansa any memory consumption concerns about this PR adding three ConcurrentHashMaps + three HashMaps per SessionFactory?

hibernate-infinispan/src/main/java/org/hibernate/cache/infinispan/impl/BaseTransactionalDataRegion.java
```diff
@@ -148,8 +144,20 @@ protected void prepareForValidation() {
 			assert strategy == Strategy.VALIDATION;
 			return;
 		}
-		validator = new PutFromLoadValidator(cache, factory);
-		strategy = Strategy.VALIDATION;
+		// If two regions share the same name, they should use the same cache.
```
scottmarlow: @rvansa does this BaseTransactionalDataRegion change look correct for ORM 5.1?  I'm asking in case BaseTransactionalDataRegion is dependent on other code that might be different between ORM 5.1 + 5.2 (not that I know of any differences).  Thanks!  -Scott

hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
rvansa: Nope, this does not cause any allocations during runtime. And users don't have thousands of regions.

hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
```diff
@@ -306,7 +309,19 @@ public QueryCache getQueryCache(String regionName) throws HibernateException {
 
 	@Override
 	public void addCacheRegion(String name, Region region) {
-		allCacheRegions.put( name, region );
+		if ( EntityRegion.class.isInstance( region ) ) {
+			entityRegionMap.put( name, (EntityRegion) region );
+		}
+		else if ( CollectionRegion.class.isInstance( region ) ) {
```
rvansa: The instance might implement multiple region types; I'd let it be added to all collections if it matches (dropping the `else`s).

hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
```diff
@@ -323,18 +338,46 @@ public void evictQueries() throws HibernateException {
...
+		Region region = entityRegionMap.get(regionName);
+		if ( region != null ) {
+			return region;
+		}
+		region = collectionRegionMap.get( regionName );
+		if ( region != null ) {
+			return region;
+		}
+		// Check for another type of Region of name regionName
+		return getNonEntityNonCollectionRegions().get( regionName );
```
rvansa: `getNonEntityNonCollectionRegions` creates a copy of all the collections, let's lookup the other maps...

hibernate-infinispan/src/main/java/org/hibernate/cache/infinispan/impl/BaseTransactionalDataRegion.java
rvansa: Not sure why are you concerned... The concept of the fix should apply to both branches imo.

hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
gbadner: I don't think that happens in 5.1 currently, but I see that it could happen. I'll make the suggested change.

hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
gbadner: OK, I'll make the suggested change and I'll move the code from ```getNonEntityNonCollectionRegions``` into ```getAllSecondLevelCacheRegions```. 

I'll also document that to keep ```getSecondLevelCacheRegion``` and ```getAllSecondLevelCacheRegions``` consistent, the order of lookups in ```getSecondLevelCacheRegion``` needs to be the opposite of the order in which ```Map``` contents are added to ```getAllSecondLevelCacheRegions```

gbadner:
@rvansa, thanks for your review. I've made changes as you suggested. Could you please take another quick look?

gbadner:
Merged. Thanks @galderz for the fix, and @rvansa for the reviews!

PR #2103, opened 2018-01-09
Fix test failures on HANA

breglerj:
- HANA doesn't support non-numeric identity columns used in
  IdentityJoinedSubclassBatchingTest


breglerj:
Fixed by commit 829b33822f259d553e87b13bb34278e01790590c

PR #2104, opened 2018-01-09
HHH-12203: NUMERIC column type is not handled correctly on HANA

breglerj:
https://hibernate.atlassian.net/browse/HHH-12203


vladmihalcea:
Applied upstream, thanks

PR #2107, opened 2018-01-10
HHH-10840 Serializable SessionFactory

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-10840

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/CustomEntityDirtinessStrategy.java
+++ b/hibernate-core/src/main/java/org/hibernate/CustomEntityDirtinessStrategy.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate;
 
+import java.io.Serializable;
+
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.type.Type;
 
@@ -20,7 +22,7 @@ import org.hibernate.type.Type;
  *
  * @author Steve Ebersole
  */
-public interface CustomEntityDirtinessStrategy {
+public interface CustomEntityDirtinessStrategy extends Serializable {
 	/**
 	 * Is this strategy capable of telling whether the given entity is dirty?  A return of {@code true} means that
 	 * {@link #isDirty} will be called next as the definitive means to determine whether the entity is dirty.
--- a/hibernate-core/src/main/java/org/hibernate/boot/CacheRegionDefinition.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/CacheRegionDefinition.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate.boot;
 
+import java.io.Serializable;
+
 /**
  * Models the definition of caching settings for a particular region.  Generally as found in either:<ul>
  *     <li>{@code cfg.xml}</li>
@@ -16,7 +18,7 @@ package org.hibernate.boot;
  *
  * @author Steve Ebersole
  */
-public class CacheRegionDefinition {
+public class CacheRegionDefinition implements Serializable {
 	public static enum CacheRegionType {
 		ENTITY,
 		COLLECTION,
--- a/hibernate-core/src/main/java/org/hibernate/boot/MetadataBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/MetadataBuilder.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.boot;
 
+import java.io.Serializable;
 import javax.persistence.AttributeConverter;
 import javax.persistence.SharedCacheMode;
 
@@ -37,7 +38,7 @@ import org.jboss.jandex.IndexView;
  *
  * @since 5.0
  */
-public interface MetadataBuilder {
+public interface MetadataBuilder extends Serializable {
 	/**
 	 * Specify the implicit catalog name to apply to any unqualified database names.
 	 * <p/>
--- a/hibernate-core/src/main/java/org/hibernate/boot/archive/scan/spi/ScanOptions.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/archive/scan/spi/ScanOptions.java
@@ -6,12 +6,14 @@
  */
 package org.hibernate.boot.archive.scan.spi;
 
+import java.io.Serializable;
+
 /**
  * Options for performing scanning
  *
  * @author Steve Ebersole
  */
-public interface ScanOptions {
+public interface ScanOptions extends Serializable {
 	/**
 	 * Is detection of managed classes from root url allowed?  In strict JPA
 	 * sense, this would be controlled by the {@code <exclude-unlisted-classes/>}
--- a/hibernate-core/src/main/java/org/hibernate/boot/cfgxml/spi/LoadedConfig.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/cfgxml/spi/LoadedConfig.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.boot.cfgxml.spi;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -35,7 +36,7 @@ import org.jboss.logging.Logger;
  * A LoadedConfig is built via {@link #consume}.  An aggregated representation
  * can be maintained through calls to {@link #merge}
  */
-public class LoadedConfig {
+public class LoadedConfig implements Serializable {
 	private static final Logger log = Logger.getLogger( LoadedConfig.class );
 
 	private String sessionFactoryName;
--- a/hibernate-core/src/main/java/org/hibernate/boot/cfgxml/spi/MappingReference.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/cfgxml/spi/MappingReference.java
@@ -7,6 +7,7 @@
 package org.hibernate.boot.cfgxml.spi;
 
 import java.io.File;
+import java.io.Serializable;
 
 import org.hibernate.boot.MetadataSources;
 import org.hibernate.boot.jaxb.cfg.spi.JaxbCfgMappingReferenceType;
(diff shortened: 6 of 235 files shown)


vladmihalcea:
@kedzie I rebased and reworked your initial PR a little bit.

winterbe:
Thanks for implementing! Hopefully this PR gets merged soon. 

Creation of Hibernate SessionFactory is a huge performance bottleneck for us (takes up to 30s which is 40% of our server startup time).

vladmihalcea:
@winterbe I'm not sure if that will get integrated soon because it's just a proof of concept. In the end, we should identify what takes long and try to offer an improvement for specific bootstrapping parts instead of having to serialize the whole SF.

Serialization makes it harder to provide backward compatibility too.

winterbe:
@vladmihalcea Thanks for your response. I suspect in our case bad performance is due to the huge amount of entities. We currently have 393 entities and 562 tables (including join tables). We use a lot of different value classes which could be combined into a single entity by introducing a discriminator column to reduce the amount of entities. But I was in hope that we could just wait for this PR to get merged.

In my opinion such a caching feature for SessionFactory would greatly improve developer experience for many hibernate based projects. I was kinda confused that the priority of the related JIRA ticket is low.

Anyways, keep up the great work and I hope one day this feature will make its way into production.

Sanne:
@winterbe we all agree it would be very nice to have the ORM boot much faster but there are many risks in this.
It's slow because it:
- verifyies the RDBMS table structure is matching the expectations
- scans your classpath to discover extension points you might need to have loaded
- has to verify the structure of your entities

Essentially if we serialize things to disk, you would have to guarantee that you didn't make change in any part of the system: not in the database, not in the project code, not even its dependencies.
_Would it still be useful for developers?_

For production, you'd have the risk of mixing a wrongly configured file - possibly not matching the database you intended to use. Imagine the pain caused of mixing up any detail, and you don't even have a validation phase at boot.
_Would it still be useful for production?_

On the hibernate-dev mailing list I mentioned I profiled the bootstrap code and created a little benchmark for it; turns out the "slow operations" might be solvable in other ways. The slowest operation is to scan a large classpath for entities, extensions, etc...

There are about 20 full classpath scans necessary to bootstrap Hibernate ORM so the most effective solution would be to trim the classpath.

Would love to hear more about your use case. Do you run in a container? Ironically, when running things in an application server they might be abel to tell us if the deployment changed, trim the visibility on the classpath, or even have knowledge about the datasource being unchanged. I guess some might even have kept the datasource pool with connections alive.

winterbe:
@Sanne Thanks for your response. 

> Essentially if we serialize things to disk, you would have to guarantee that you didn't make change in any part of the system: not in the database, not in the project code, not even its dependencies.
Would it still be useful for developers?

It would still be useful for us, let me explain way: Developers on our team usually restart the server on their local machines very often on an average work day, but the JPA data-model changes less frequently. Very often Hibernate could just blindly reuse cached SessionFactory during startups.
I understand that cache invalidation would be hard to get right 100% but for our case it would be completely ok if cache invalidation would be handled either manually e.g. by executing a command line tool to remove serialized SessionFactory from disk or by disabling cache for next start by setting a system property. So if you make hibernate-relevant code changes you disable caching and afterwards you re-enable caching again.

> For production, you'd have the risk of mixing a wrongly configured file - possibly not matching the database you intended to use. Imagine the pain caused of mixing up any detail, and you don't even have a validation phase at boot.
Would it still be useful for production?

Caching SessionFactory on production servers is no priority for us. For our use case it would be absolutely ok to completely disable caching in production.

> Would love to hear more about your use case. Do you run in a container? Ironically, when running things in an application server they might be abel to tell us if the deployment changed, trim the visibility on the classpath, or even have knowledge about the datasource being unchanged. I guess some might even have kept the datasource pool with connections alive.

Our application is a web app running on Apache Tomcat 8. Class path scanning is indeed a problem because it has grown kinda big over the years. Spring Framework 5 recently solved this problem by scanning the class path during compile time via [an annotation processor](https://docs.spring.io/spring/docs/5.0.3.RELEASE/spring-framework-reference/core.html#beans-scanning-index).

Is there something we could do to reduce the amount of class path scans Hibernate is performing on startup?

Sanne:
@winterbe that's awesome feedback, thanks it will definitely make me think again on my objections.

Scanning classpath is the strongest contributor to boot time in my own measurements, which actually happened on a relatively small set of dependencies, and it's supposedly growing *at least* linearly with the amount of classes to scan.

So.. I wasn't aware of the Spring 5 pre-scan capabilities; we should talk with the Spring team about that - or perhaps they already inject the necessary things in ORM? We do provide the injection points as e.g. WildFly uses them, but they are intentionally designed to be platform agnostic. So maybe Spring already takes advantage of it, or maybe they don't know about it. I'll ask...

winterbe:
> that's awesome feedback, thanks it will definitely make me think again on my objections.

@Sanne Thanks. Please keep me informed on that topic if something changes. Keep up the great work! 👍

Sanne:
Thanks :)

FYI conversation is here: https://twitter.com/SanneGrinovero/status/960497968527958016

I guess I'll need to experiment a bit with all the things Oliver mentioned. @winterbe have you used those?

snicoll:
For the record, that facility in spring Framework also scan JPA entities as long as you are asking Spring Boot to create the persistence unit for you.

winterbe:
Spring Boot DevTools Sound pretty cool but we use raw Spring Framework (not Boot) in the relevant product. So we can’t benefit from this unfortunately.

snicoll:
@winterbe you should be able to replicate what spring boot does to avoid using classpath scanning for your JPA entities.

jonnermut:
I agree with @winterbe that hibernate startup time is a major pain point for teams working on a server. Typically the model changes slowly, but the dev team is starting up the server many, many times a day, and the startup time (~20 seconds on my project) really adds up. 
It also really blows out the time of spring integration tests.
I think a serialisation mechanism could be really useful. Turn it off by default, caveat the hell out of it, and let teams build their own use of it to try and increase their productivity.
I personally would take a hash of all the source files which I know make up the model, plus the version config, and name the serialisation file off that. I would be happy to use it for dev, tests and maybe production (build the file in the release build, and key it off the git commit), depending on exactly how it interacted with auto schema update.

sebersole:
The default serialization for a SessionFactory will never be anything other
than what it is.  It is critical for many Session, Type, etc serialization
that this work exactly as it does.

I'm not against a specific, dedicated API for doing this.

jonnermut:
> I'm not against a specific, dedicated API for doing this.

Which would obviously be fine for the use case for custom dev/build tools.
Looks like this PR is ageing badly though given the number of conflicts...

jonnermut:
@Sanne Thanks for the analysis and the work on this. I will definitely look into AppCDS.

I Agree that the session serialisation approach totally depends on build tools, or some runtime knowledge of the dev environment.

We start up our container using embedded Tomcat, from a main() method. Similar to Spring Boot, just a couple of years before it came out. And then we bootstrap Hibernate using the Spring Hibernate support classes plus a bunch of code to apply our environment properties.

The cold start is by far the most important to us. As you are developing features you need to restart the JVM often, so a developer on a tear will be restarting dozens if not hundreds of times a day. Similarly, we isolate most of our integration tests per test class, so each test class has to do a full spring and hibernate startup.

I can envision writing our own method in our startup code which:
1. In our gradle build, in every build (dev or release), always calculate a hash or fingerprint  of the model sources and config, based on either file content or mod times, version number, whatever, and put it in a property file. (Gradle actually has very good inbuilt support for doing this, come to think of it, and could quite easily not do the work if nothing has changed). 
2. In our startup code, read the property, and if the file named with that hash exists, read it and use it to boot the session factory with it
3. Otherwise boot the session factory and immediately serialise it to that file, so its there for the next cold start
4. And if the serialisation process proves stable and solid enough, then in our Gradle build process, for a release build, run the startup code, and include the produced file in the build as a resource in the release build.

If we can gain anything out of annotation processors or other build time work it would also be worth investigating. For the integration tests, for example, the build runs once and the runtime startup runs many times, so gains are amplified.

Sanne:
Thanks! very interesting feedback and nice to compare some ideas.

You're right Gradle has great support for incremental builds so it might be possible to take advantage of it to automatically recognize when the model has changed.

I'm surprised that you're most interested in cold JVM bootstrap times. I thought people were able to avoid a full restart of the container nowadays by working mostly with integration tests, or at most using tools such as JRebel or FakeReplace. Neither of these is (yet!) able to adjust a booted Hibernate model but AFAIK they should be able to restart Hibernate or make other changes in a running container without needing to reboot it all.

Another question for you: did you try Hibernate ORM 5.3?
Several improvements were made; for example I expect people to use `hibernate.hbm2ddl.auto=validate` (at least), which used to query the database once for each table but is now able to fetch the whole schema with a single query.

### connection pools

Also quite some time is "wasted" in starting the connection pool; each connection in the pool needs to be established which implies several roundtrips and handshakes with the database, so another thing I'd like to explore is to let the runtime re-inject the existing pool. (that's not going to help you for cold-boot but it might help integration tests?)

### entity enhancement

we also need the entities to be enhanced. This requires additional temporary classloaders and that's also contributing to bootstrap times - so it might help to use the build tools to perform entity enhancement at compile time.

parthsaarthi001:
are there any plans to merge this PR ?

jianzhichun:
are there any plans to merge this PR ?

Sanne:
> are there any plans to merge this PR ?

I won't work on this myself for the following reasons:

- we improved bootstrap _a lot_ already in recent versions.
- I believe there are better and safer wasy to achieve the "frozen state" ideas, as mentioned above.

Please checkout what I did for [Quarkus](https://quarkus.io/); some optimisations are specific to Quarkus but many others are now part of Hibernate ORM. If you need optimisations for other runtimes, we'd be happy to help crafting optimisations which are specific to other runtimes too - but the potential of such optimisations depends on how willing people are to break compatibility with previous versions.

Do people here still have problems with bootstrap times? If you could share applications / benchmarks or even just profiling data showing what is the problem, I'd be happy to look at that.

Specifically this PR has many conflicts so I'm afraid I will close it. We can still discuss improvement ideas, but I'd suggest making a proposal on the mailing list before spending too much time on a patch.
 - https://lists.jboss.org/mailman/listinfo/hibernate-dev

PR #2108, opened 2018-01-10
HHH-12190 - General tidying of Gradle scripts

sebersole:
This is a pretty significantly different way to define the Gradle build scripts.  Curious if you guys also find this easier to read?


dreab8:
for me it's a +1, a lot of duplicated code removed, I like it.
Only one question, are we removing eclipse and IntelliJ plugin?

sebersole:
Not so much removing it.  More that it seems like these are not really needed anymore.  But either way, it is simple enough to add them back in if need be, in fact they are just commented out for now.

PR #2110, opened 2018-01-12
HHH-12207 test

gbadner:
Test case for https://hibernate.atlassian.net/browse/HHH-12207


gbadner:
Created for master; was supposed to be for 5.2

PR #2114, opened 2018-01-15
HHH-12216 - Improve logging for when Hibernate throws the "illegally attempted to associate a proxy with two open Sessions" Exception

mnachmia31:
Update AbstractLazyInitializer.java to include debugging information regarding the entity being proxied when illegally attempting to associate it with two open sessions. Right now the stack trace of the exception does not give any information on the entity itself which makes debugging this issue cumbersome.


vladmihalcea:
Applied upstream, along with a test case.

PR #2117, opened 2018-01-18
HHH-12141 - SQL insert in stateless session causes javax.persistence.TransactionRequiredException

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12141


vladmihalcea:
@dreab8 @sebersole Let me know if you agree with the approach I used in this PR. Thanks.

dreab8:
with 5.2 we introduced a new setting `hibernate.allow_update_outside_transaction`, `SessionImpl#checkTransactionNeeded() `uses this value before a every flush.

I may be wrong but I probably the same check can be done for StatelessSession

vladmihalcea:
@dreab8 Thanks for the hint. I'll have to investigate. I'm changing the label.

PR #2118, opened 2018-01-18
HHH-12190 - General tidying of Gradle scripts

dreab8:
@sebersole , not sure if you want this, but the PR adds to the relocation POM of the infinispan project the info JBoss Nexus is checking for


sebersole:
I'd rather JBoss not do their over-zealous validation, but that won't change soon so yeah I think we need this

PR #2119, opened 2018-01-19
HHH-12227 {h-schema} is not replaced in @Formula

MykhayloGnylorybov:
https://hibernate.atlassian.net/browse/HHH-12227


vladmihalcea:
Applied upstream, thanks

PR #2120, opened 2018-01-19
test-case for HHH-12114: queries containing multiple .class/TYPE() clauses on union-subclass aliases

pholvs:
test-case for HHH-12114: queries containing multiple .class/TYPE() clauses on union-subclass aliases


vladmihalcea:
Applied upstream along with a fix, thanks.

PR #2121, opened 2018-01-19
add testcase for HHH-12228: duplicate fields across subclasses

pholvs:


vladmihalcea:
Applied upstream, thanks.

I closed the issue as a duplicate of HHH-11686, but the problem is still valid if we use TREAT.

PR #2122, opened 2018-01-20
HHH-12231 Use the org.wildfly.build.provision Gradle plugin to fetch …

Sanne:
…WildFly instances for integration testing

 - https://hibernate.atlassian.net/browse/HHH-12231

Thanks for @dreab8 for sketching the patch first and @sebersole for all the guidance :)

See also:
 - https://plugins.gradle.org/plugin/org.wildfly.build.provision
 - https://github.com/Sanne/wildfly-gradle-tools


Sanne:
Added a second commit, now using the "overrides" capability we discussed to already include the JPA 2.2 API - this removes the need of subsequently patching the server.

Sanne:
merged

PR #2123, opened 2018-01-21
HHH-12233: Fix NPE in CacheImpl.

langfr:
Fix NPEs in CacheImpl.java seen when using Hibernate 5.1.11 with Wildfly 11.x.
Varaibles queryCache, queryCaches and updateTimestampsCache can be null.

Simply omitted the calculation of the exact HashMap size.
Protected the use of the possibly null regions like done at other places in this source.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
@@ -406,11 +406,9 @@ public class CacheImpl implements CacheImplementor {
 		if ( settings.isQueryCacheEnabled() ) {
 			allCacheRegions.put( updateTimestampsCache.getRegion().getName(), updateTimestampsCache.getRegion() );
 			allCacheRegions.put( queryCache.getRegion().getName(), queryCache.getRegion() );
-		}
-		// keys in queryCaches may not be equal to the Region names
-		// obtained from queryCaches values; we need to be sure we are adding
-		// the actual QueryCache region name as the key in allCacheRegions.
-		if ( settings.isQueryCacheEnabled() ) {
+			// keys in queryCaches may not be equal to the Region names
+			// obtained from queryCaches values; we need to be sure we are adding
+			// the actual QueryCache region name as the key in allCacheRegions.
 			for ( QueryCache queryCacheValue : queryCaches.values() ) {
 				allCacheRegions.put( queryCacheValue.getRegion().getName(), queryCacheValue.getRegion() );
 			}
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
```diff
@@ -402,17 +402,18 @@ public Region getNaturalIdCacheRegion(String regionName) {
...
 		// keys in queryCaches may not be equal to the Region names
 		// obtained from queryCaches values; we need to be sure we are adding
 		// the actual QueryCache region name as the key in allCacheRegions.
-		for ( QueryCache queryCacheValue : queryCaches.values() ) {
-			allCacheRegions.put( queryCacheValue.getRegion().getName(), queryCacheValue.getRegion() );
+		if ( settings.isQueryCacheEnabled() ) {
+			for ( QueryCache queryCacheValue : queryCaches.values() ) {
+				allCacheRegions.put( queryCacheValue.getRegion().getName(), queryCacheValue.getRegion() );
+			}
 		}
```
gbadner: The 2 ```if ( settings.isQueryCacheEnabled() ) {``` statements can be combined.

hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
```diff
@@ -402,17 +402,18 @@ public Region getNaturalIdCacheRegion(String regionName) {
 		// which Map contents are added to getAllSecondLevelCacheRegions.
 		// In addition, if there is a CollectionRegion and an EntityRegion with the same name, then we
 		// want the EntityRegion to be in the Map that gets returned.
-		final Map<String, Region> allCacheRegions = new HashMap<String, Region>(
-				2 + queryCaches.size() + otherRegionMap.size() + naturalIdRegionMap.size()
-						+ collectionRegionMap.size() + entityRegionMap.size()
-		);
-		allCacheRegions.put( updateTimestampsCache.getRegion().getName(), updateTimestampsCache.getRegion() );
-		allCacheRegions.put( queryCache.getRegion().getName(), queryCache.getRegion() );
+		final Map<String, Region> allCacheRegions = new HashMap<String, Region>();
```
gbadner: @rvansa, is it worthwhile to instantiate the ```HashMap``` to it's ultimate size, since it can be computed beforehand?

AFAIK, Hibernate only uses ```getNaturalIdCacheRegion``` when generating statistics. An application could use this method as well.

gbadner:
@langfr, thanks much for finding the regression and creating the PR! In the future, please be sure to start commit comments with the HHH issue ID.

Don't worry about it for this PR; I'll update the comment when I squash the commits.

hibernate-core/src/main/java/org/hibernate/internal/CacheImpl.java
rvansa: @gbadner I am not a profiler :) IMO this is not a perf-critical method, I'd personally rather keep it simple.

gbadner: @rvansa, thanks.

gbadner:
Test case added, commits merged, and pushed to 5.1 branch.

@langfr, hanks for the PR!

PR #2125, opened 2018-01-22
Test for issue HHH-12225

randymay:


vladmihalcea:
Applied upstream along with the fix Thanks.

PR #2126, opened 2018-01-22
HHH-12034 - According to JPA, a Proxy should be loaded even when accessing the identifier

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12034

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
@@ -924,6 +924,8 @@ public interface AvailableSettings extends org.hibernate.jpa.AvailableSettings {
 	 * pass the BeanManager to use via {@link #CDI_BEAN_MANAGER} and
 	 * optionally specify {@link #DELAY_CDI_ACCESS}.  This setting is more meant to
 	 * integrate non-CDI bean containers such as Spring.
+	 *
+	 * @since 5.3
 	 */
 	String BEAN_CONTAINER = "hibernate.resource.beans.container";
 
@@ -1772,6 +1774,7 @@ public interface AvailableSettings extends org.hibernate.jpa.AvailableSettings {
 	 * since it extends the JPA one.
 	 *
 	 * @see JpaCompliance#isJpaTransactionComplianceEnabled()
+	 * @since 5.3
 	 */
 	String JPA_TRANSACTION_COMPLIANCE = "hibernate.jpa.compliance.transaction";
 
@@ -1785,6 +1788,7 @@ public interface AvailableSettings extends org.hibernate.jpa.AvailableSettings {
 	 * Deviations result in an exception if enabled
 	 *
 	 * @see JpaCompliance#isJpaQueryComplianceEnabled()
+	 * @since 5.3
 	 */
 	String JPA_QUERY_COMPLIANCE = "hibernate.jpa.compliance.query";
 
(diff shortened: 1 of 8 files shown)


vladmihalcea:
@sebersole @gbadner I added a fix to this issue. Just like other settings encapsulated in `JpaCompliance`, this one is also disabled by default. However, users can enable it if they want strict compliance.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -1810,21 +1815,40 @@
...
 
+	/**
+	 * JPA spec says that an {@link javax.persistence.EntityNotFoundException}
+	 * should be thrown when accessing an entity Proxy which does not have an associated
+	 * table row in the database.
+	 *
+	 * Traditionally, Hibernate does not initialize an entity Proxy when accessing its
+	 * identifier since we already know the identifier value, hence we can save a database roundtrip.
+	 *
+	 * If enabled Hibernate will initialize thr entity Proxy even when accessing its identifier.
```
ajrepp: Nit: Typo: "thr" instead of "the".

hibernate-core/src/main/java/org/hibernate/jpa/JpaCompliance.java
```diff
@@ -63,4 +63,18 @@
...
+
+	/**
+	 * JPA spec says that an {@link javax.persistence.EntityNotFoundException}
+	 * should be thrown when accessing an entity Proxy which does not have an associated
+	 * table row in the database.
+	 *
+	 * Traditionally, Hibernate does not initialize an entity Proxy when accessing its
+	 * identifier since we already know the identifier value, hence we can save a database roundtrip.
+	 *
+	 * If enabled Hibernate will initialize thr entity Proxy even when accessing its identifier.
```
ajrepp: Nit: Typo: "thr" instead of "the".

hibernate-core/src/main/java/org/hibernate/proxy/AbstractLazyInitializer.java
```diff
@@ -67,6 +67,8 @@ protected AbstractLazyInitializer(String entityName, Serializable id, SharedSess
 		}
 		else {
 			setSession( session );
+			initializeProxyWhenAccessingIdentifier = session.getFactory().getSessionFactoryOptions()
+					.getJpaCompliance().isJpaProxyComplianceEnabled();
 		}
```
gbadner: I don't think it's enough for ```initializeProxyWhenAccessingIdentifier``` to be a field.

Here is a scenario where there could be problems:
1) the object is constructed, using a ```SessionFactory``` with ```hibernate.jpa.compliance.proxy=false``` configured;
2) the object is detached and ```session`` is set to null;
3) the object is attached to a ```Session``` for a different ```SessionFactory``` with ```hibernate.jpa.compliance.proxy=true``` configured;
4) the application calls ```getIdentifier()```, which returns the ID without initializing the entity.

Since ```hibernate.jpa.compliance.proxy=true``` is configured in the new ```SessionFactory```, the entity should have been initialized before returning the ID.

I think that the ```sessionFactoryUuid``` should also be assigned, and the ```SessionFactory``` should be checked in ```getIdentifier()```, similar to what is done for ```allowLoadOutsideTransaction```.

IMO, when the object gets reattached to a ```Session```, ```sessionFactoryUuid``` should be updated to the UUID for ```Session```'s ```SessionFactory```.

gbadner:
@vladmihalcea, this should also be fixed in 5.2 branch. Do you plan to fix it the same way (by backporting ```JpaCompliance```?

vladmihalcea:
@gbadner I think of using a simple property instead which we already support in 5.3. This way, we don't need to backport `JpaCompliance`.

PR #2127, opened 2018-01-22
HHH-12238 - Avoid confusing ClassCastException

dgtombs:
Ensure an object is actually a Map before treating it like one.


vladmihalcea:
Applied upstream, thanks

PR #2128, opened 2018-01-23
HHH-9460 - Removing non-optional bidirectional @OneToOne association with cascade

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-9460


vladmihalcea:
@gbadner I stumbled on this issue while trying to replicate a `@MapsId` + `@JoinColumn` mapping [mentioned in our forum](https://discourse.hibernate.org/t/set-primary-key-and-foreign-key-on-a-single-column-in-hibernate-mapping-having-type-string/108/4).

After analyzing it a little bit, I came to the following change request and noticed we already have the HHH-9460 issue for this one.

Let me know what you think of it. Thanks.

PR #2129, opened 2018-01-23
HHH-12244 HHH-12245 HV + JDK 9 related fixes

gsmet:
This was discussed with Andrea as part of his ongoing JDK 9 work.

 * https://hibernate.atlassian.net/browse/HHH-12244
 * https://hibernate.atlassian.net/browse/HHH-12245

### DIFF
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -16,6 +16,7 @@ ext {
     jnpVersion = '5.0.6.CR1'
 
     hibernateValidatorVersion = '6.0.7.Final'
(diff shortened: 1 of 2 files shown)


sebersole:
Was these something specific you wanted us to check here?  If not, our workflow is to just push simple changes like this without other people on the team looking at it.  If we didn't trust you we wouldn't have given you write priv ;)

gsmet:
Yeah, this is mostly to have @dreab8 take a look at it before merging.

dreab8:
Thnaks @gsmet , 
applied upstream.

PR #2130, opened 2018-01-24
HHH-12248 Clustered 2nd level cache sometimes throws NullPointerExcep…

tomdearman:
This should take care of the race condition in PutFromLoadValidator.

### DIFF
--- a/hibernate-infinispan/src/main/java/org/hibernate/cache/infinispan/access/PutFromLoadValidator.java
+++ b/hibernate-infinispan/src/main/java/org/hibernate/cache/infinispan/access/PutFromLoadValidator.java
@@ -156,6 +156,10 @@ public class PutFromLoadValidator {
 		else {
 			throw log.pendingPutsMustHaveMaxIdle();
 		}
+
+		this.cache = cache;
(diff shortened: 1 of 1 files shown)


vladmihalcea:
@tomdearman 5.2 is now used only for bug fixing. All fixes must be approved by @gbadner for older branches.

The `hibernate-infinispan` module was moved to `org.infinispan:infinispan-hibernate-cache`. You should provide the fix there as well. Thanks.

tomdearman:
Ok, I had notice it was moved, but wasn’t sure what the procedure was, I will provide the fix there too.  This is a bug so hopefully @gbadner will approve it.

gbadner:
@vladmihalcea, this does not require me to make a decision on this.

vladmihalcea:
@gbadner Yes. I wrote that comment before I was told that we can backport in 5.2 until we have a 5.3 Final release.

vladmihalcea:
Applied upstream, thanks.

Don't forget to send this PR to Infinispan project, as otherwise, the issue will fail in 5.3 since the Hibernate Infinispan caching implementation is provided by Infinispan now.

tomdearman:
This has already been accepted and applied to infinispan 5.2.Final.

PR #2131, opened 2018-01-24
Add test for HHH-9296

codylerum:
https://hibernate.atlassian.net/browse/HHH-9296


vladmihalcea:
Applied upstream, thanks. The issue does not replicate on master so, you might consider upgrading to 5.2 or 5.3.

codylerum:
@vladmihalcea Thanks. Unfortuantly I'm using the hibernate bundled into Wildfly and I'd rather not start upgrading the internals of Wildfly as it makes CI testing more difficult.

Looks like WF 12 will be 5.1 still as well so will be stuck with this one for a bit I guess.

sebersole:
We offer a Wildfly "bundle" for replacing Hibernate in WildFly :
hibernate-orm-modules.  Its not that hard.  In fact we use that in ORM
itself to be able to download a version of WidFly and apply our current ORM
version to do some in-container testing.

PR #2132, opened 2018-01-25
HHH-12252 - Agroal integration

barreiro:
[HHH-12252](https://hibernate.atlassian.net/browse/HHH-12252)


vladmihalcea:
Applied upstream, thanks

PR #2135, opened 2018-01-26
HHH-8382 - Fix lob column ordering when using HBM mappings.

Naros:
https://hibernate.atlassian.net/browse/HHH-8382

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.boot.model.source.internal.hbm;
 
+import java.sql.Types;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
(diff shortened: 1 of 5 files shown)


Naros:
@h2m Fair enough, I updated the PR accordingly.

h2m:
@Naros thanks! Sorry for not having thought about this already the last time, but I fear your LobAsLastValueTest would already have passed without the fix, because for such short Strings, Oracle does not bother about the order at all. I am not sure if it is worth the extra effort, there should either be assertions on the order of the column names in the SQL statement or we should ensure we have test data that fails the test without the fix and makes it pass only after.

Naros:
@h2m All valid points.  I'm updating the PR's test case so that it fails.  I also found a small bug in my implementation for detecting this use case and the updated PR will address that too.

Naros:
@h2m I just pushed the changes for this PR; could you give this a quick test/look and see if you find anything else of concern?  I ran this against Oracle 12 and it failed prior to the fix & passes after.

h2m:
Hi @Naros ,

my setup in the reproducer was a bit different, I had only a single LOB column combined with a VARCHAR(2000) and Oracle complained only when the value for the VARCHAR column was very long.

But if your data setup reproduces the fix, I am happy :-D !

BTW, I am curious where your integration tests against different database versions run in CI.

Thanks for the fix
Marc

`
<hibernate-mapping package="org.hibernate.bugs.entity">

    <class name="MyEntity" table="my_entity">
        <id name="id">
            <generator class="sequence">
                <param name="sequence">my_entity_seq</param>
            </generator>
        </id>
        <property name="details" type="string" length="4000">
            <column name="details" sql-type="clob"/>
        </property>
        <property name="title" length="2000"/>
    </class>
</hibernate-mapping>
`

Naros:
> BTW, I am curious where your integration tests against different database versions run in CI.
What exactly do you mean?

Naros:
Applied upstream.

PR #2136, opened 2018-01-28
HHH-12264 - Updated Vibur dependency from 21.2 to 21.3

simeonmalchev:


vladmihalcea:
Applied upstream, thanks

PR #2138, opened 2018-01-30
HHH-12203 for 5.2

breglerj:
https://hibernate.atlassian.net/browse/HHH-12203

If possible I would like to have this fix backported to version 5.2.


vladmihalcea:
Applied upstream, thanks

PR #2139, opened 2018-01-30
HHH-11571 - JTA platform for WebSphere Liberty and OpenLiberty

aguibert:
This will allow Hibernate to properly interoperate with Liberty's transaction service out of the box.  

All tests for the `hibernate-core` project have been run locally (`../gradlew build`) and are passing.

Based on the precedent set by other JtaPlatforms, no documentation or unit tests will be added.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/WebSphereLibertyJtaPlatform.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/WebSphereLibertyJtaPlatform.java
@@ -1,0 +1,83 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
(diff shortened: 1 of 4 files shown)


gsmet:
@sebersole wondering if it could be a good candidate for 5.3?

hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/WebSphereLibertyJtaPlatform.java
```diff
@@ -0,0 +1,83 @@
...
+
+	public Object getTransactionIdentifier(Transaction transaction) {
+		return transaction;
+	}
+
+	public void registerSynchronization(Synchronization synchronization) {
+		try {
+			retrieveTransactionManager().getTransaction().registerSynchronization(synchronization);
+		} 
+		catch (IllegalStateException x) {
```
vladmihalcea: This catch is not needed since `IllegalStateException` extends `RuntimeException`.

vladmihalcea:
Applied upstream, thanks

PR #2140, opened 2018-01-31
HHH-12267 - Update migration guide to cover Generators name scope changes

dreab8:
https://hibernate.atlassian.net/browse/HHH-12267


vladmihalcea:
Applied upstream, thanks

PR #2141, opened 2018-01-31
HHH-12266 - The release task does not upload the documentation

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12266


vladmihalcea:
@sebersole That's simpler indeed. In my PR, I've just reused the task we've been having before on 5.2. I'll redo the PR to simplify it and push it.

PR #2142, opened 2018-02-05
HHH-12273 - Load Proxy by its identifier should consider the Session UUID

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12273


vladmihalcea:
@gbadner I created this issue and provided a fix to cover [the situation you described previously](https://github.com/hibernate/hibernate-orm/pull/2126). Let me know if you agree with the change. I think it's easier to just fetch the `initializeProxyWhenAccessingIdentifier` flag when we need it instead of caching the Session identifier as well.

PR #2137, opened 2018-01-30
HHH-12260: Detach of entity with lazy-loaded collection and orphan re…

breglerj:
…moval leads to exception during flush of session

https://hibernate.atlassian.net/browse/HHH-12260

PR #2143, opened 2018-02-06
HHH-12260: refactor org.hibernate.event.internal.EvictVisitor#evictCollection

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12260

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultEvictEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultEvictEventListener.java
@@ -73,9 +73,8 @@ public class DefaultEvictEventListener implements EvictEventListener {
 			li.unsetSession();
 		}
 		else {
-			EntityEntry e = persistenceContext.removeEntry( object );
+			EntityEntry e = persistenceContext.getEntry( object );
 			if ( e != null ) {
-				persistenceContext.removeEntity( e.getEntityKey() );
 				doEvict( object, e.getEntityKey(), e.getPersister(), source );
 			}
 			else {
@@ -119,7 +118,7 @@ public class DefaultEvictEventListener implements EvictEventListener {
 
 		// remove all collections for the entity from the session-level cache
 		if ( persister.hasCollections() ) {
-			new EvictVisitor( session ).process( object, persister );
+			new EvictVisitor( session, object ).process( object, persister );
 		}
 
 		// remove any snapshot, not really for memory management purposes, but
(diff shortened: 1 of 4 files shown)


vladmihalcea:
@breglerj I created this Pull Request which is based on [your PR](https://github.com/hibernate/hibernate-orm/pull/2137), adds more tests, and removes some changes and the tests pass just fine.

Let me know why you needed all the previous changes because they cannot be replicated by the associated unit tests.

breglerj:
@vladmihalcea I made the additional changes to be consistent with the changes required for making the test pass. My assumption was that an entity should not be removed from the persistence context before EvictVisitor#process is called. Otherwise an uninitialized collection can't be resolved. If you're sure that this can't be a problem in the places where you removed my changes (DefaultRefreshEventListener, the object to be evicted is a HibernateProxy) then that's fine with me.

vladmihalcea:
@breglerj I added a similar test for refresh. However, I'm not sure what was the specific use case so that we can emulate the problem which requires us to make those changes in `DefaultRefreshEventListener`.

If you can think about a test case, let me know, and we should add that too. Otherwise, all tests are running fine without those additional changes.

breglerj:
@vladmihalcea Unfortunately I don't have a test case for the other changes. Like I was trying to say, my reasoning for those other changes was that there is a theoretical possibility which I couldn't rule out that `CollectionType#resolve` will be called via `EvictListener#process` from those places. The call to `CollectionType#resolve` will return `null` if the owning entity has already been evicted, so I adapted these calls to `EvictListener#process` as well purely for consistency and symmetry reasons. Anyway, I'm fine with merging the PR as-is.

vladmihalcea:
@gbadner Can you take a look at it and say if we should include the changes in this PR or the ones in [the original PR](https://github.com/hibernate/hibernate-orm/pull/2137).

breglerj:
@vladmihalcea I did some more experimenting and I'm almost sure that the error can't happen in the other places. 

In `DefaultRefreshEventListener#onRefresh` the entity entry is not removed from the persistence context which means that the collection can still be resolved. Anyway, even if the entity entry is removed it doesn't seem to cause a problem. 

The first case of `DefaultEvictEventListener#onEvict` is the case that the entity is not byte code enhanced. In this case it seems that the collection type is not uninitialized, so it get evicted correctly.

Assuming what I stated is correct, the difference between my original PR and this one is really only a matter of style. But in any case I guess @gbadner should decide which one to merge.

PR #2144, opened 2018-02-07
HHH-12230 - SelectCase does not work when simultaneously exists in select and group by sections

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12230


vladmihalcea:
@sebersole let me know what you think of this PR. Thanks.

PR #2146, opened 2018-02-08
Updated Vibur dependency from 21.3 to 22.0

simeonmalchev:


vladmihalcea:
Applied upstream, thanks

PR #2147, opened 2018-02-09
HHH-12280 - Resolve {alias} in @Formula

quaff:


Naros:
Could you also include a test case that shows that "{alias}" did not resolve prior to your code change so that we can avoid future regressions with this fix?

quaff:
TestCase added.

vladmihalcea:
The test case doe snot work. This is the logged SQL query:

`select formulawit0_.id as id1_0_, formulawit0_.balance as balance2_0_, formulawit0_.vip as vip3_0_, formulawit0_.balance/(select sum(c.balance) from Customer c where c.vip=.vip) as formula0_ from Customer formulawit0_`

Notice that the `{alias}` was replaced by a whitespace.

vladmihalcea:
Applied upstream, thanks

PR #2148, opened 2018-02-13
HHH-12292: nulls are valid values for Objects inside array

coladict:


vladmihalcea:
Added test case and applied to master and 5.2.

PR #2149, opened 2018-02-13
HHH-12306 - Fix org.hibernate.envers.test.integration.manytomany.MappedByEmbeddableAttributeTest on HANA

breglerj:
- The order of the elements in Container#bList is undefined on the database
  side. This fix adds a hasItem matcher instead of explicitly checking each item
  of the list which depends on the order of the elements.


vladmihalcea:
Applied upstream, thanks

PR #2150, opened 2018-02-14
HHH-12285 - DB connection exception on rollback causes connection leak

dreab8:
https://hibernate.atlassian.net/browse/HHH-12285

The db connection is not released because when the `EntityManager#close() `method is called the transaction status results active.
`SessionImpl#close() `method executes the following code:
```
if ( discardOnClose || !isTransactionInProgress( false ) ) {
  super.close();
}
else {
 //Otherwise, session auto-close will be enabled by shouldAutoCloseSession().
 waitingForAutoClose = true;
 closed = true;
}
```
and  `isTransactionInProgress( false )` is true then the super.close() method is not called .

The patch introduces the `TransactionStatus.FAILED_ROLLBACK` value that is set when a rollback fails.


vladmihalcea:
Applied upstream, thanks

PR #2151, opened 2018-02-14
HHH-12295 Restore compatibility with Java 6 on branch 5.1

Sanne:
https://hibernate.atlassian.net/browse/HHH-12295

N.B. this PR is for branch 5.1 only.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
@@ -17,7 +17,6 @@ import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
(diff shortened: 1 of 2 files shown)


Sanne:
@vladmihalcea , could you have a look please? I removed the Pattern.UNICODE_CHARACTER_CLASS flag in your regex but I'm not sure I understand the implications.

`Pattern.UNICODE_CHARACTER_CLASS` is not available in Java 6, while ORM 5.1 should still be compatible with Java 6, so this flag has to go. Maybe we need a different replacement?

vladmihalcea:
@Sanne I remember that I added the `UNICODE_CHARACTER_CLASS` after someone complained that his Java Constants written with Chinese Glyphs were not caught by that Regex. 

If all tests pass just fine `ReflectHelperTest#test_getConstantValue_constant_digits` passes just fine, then no worries.

Sanne:
thanks Vlad. I'll merge this then.

Sanne:
ah no.. forgot this is 5.1.

I'll let Gail merge it so she can check it all.

Sanne:
@gbadner ? I thought this was urgent, and should be trivial to check. Just looking for your ACK.

gbadner:
@Sanne , ```ReflectHelperTest#test_getConstantValue_constant_digits``` is failing.

@vladmihalcea , any suggestions?

rstancel:
@gbadner the issue here is that the pattern is requiring class with 2 uppercase letters and the test class contains only one. So the solution can be to update test data. Something like that: https://github.com/rstancel/hibernate-orm/commit/f0806b67515d0ab1a1c461ed5ebd25f99aabd51f

gbadner:
@rstancel, IIUC, you've just change the test code. The bug remains.

rstancel:
@gbadber I don't understand. Sanne has fixed the issue itself but his fix was failing one test. So I took his fix and added also a fix for this test. Is there something else which should be fixed?

PR #2152, opened 2018-02-15
HHH-12278

benoitbesson:
This pull request includes : 
- the full change for classes AnnotationBinder and IdGeneratorInterpreterImpl
- additional classes to delegate on in package org.hibernate.cfg.schema
- test classes in org.hibernate.cfg.annotations.schema illustrating usage

I did not find how to check test results and automate assertions. When running tests with Gradle, I see a lot of error traces not related to this test. Maybe I need a little help... 

I hope this code could be enough to discuss on the proposal.

see https://hibernate.atlassian.net/browse/HHH-12278

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/schema/SchemaNamingProviderLocator.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/schema/SchemaNamingProviderLocator.java
@@ -1,0 +1,58 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.cfg.schema;
+
+import org.hibernate.annotations.common.reflection.XClass;
+
+/**
+ * SchemaNamingProvider singleton locator.<br>
+ *
+ * TODO : integrate into class AnnotationSettings as "official" configuration. <br>
+ * TODO (later) : migrate into a non static approach... (complex)
+ *
+ * @author Benoit Besson
+ */
+public class SchemaNamingProviderLocator {
+
+	/** static instance to delegate on. */
+	private static SchemaNamingProvider instance;
+
+	public static void setInstance(SchemaNamingProvider instance) {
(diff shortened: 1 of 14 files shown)


vladmihalcea:
The `SchemaNamingStrategyTest` fails with a NPE.

hibernate-core/src/main/java/org/hibernate/cfg/schema/SchemaNamingProviderLocator.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * TODO : integrate into class AnnotationSettings as "official" configuration. <br>
+ * TODO (later) : migrate into a non static approach... (complex)
+ *
+ * @author Benoit Besson
+ */
+public class SchemaNamingProviderLocator {
+
+	/** static instance to delegate on. */
+	private static SchemaNamingProvider instance;
```
vladmihalcea: A static variable without any thread safety blocks is risky. Why not use a local variable and wrap it inside an object that confined to the caller.

hibernate-core/src/test/java/org/hibernate/cfg/annotations/schema/SchemaNamingStrategyTest.java
```diff
@@ -0,0 +1,115 @@
...
+			schemaExport.setDelimiter(";");
+			schemaExport.setHaltOnError(true);
+			schemaExport.setImportFiles(null);
+			File outputFile = File.createTempFile("schema-screate", "sql");
+			schemaExport.setOutputFile(outputFile.getAbsolutePath());
+
+			// write to stdout and file
+			schemaExport.execute(EnumSet.of(TargetType.SCRIPT, TargetType.STDOUT), Action.CREATE, metadata);
+
+			// TODO : add assertion for expected file content
```
vladmihalcea: The  asserts are a must have.

benoitbesson: I agree.

I made this first proposal to discuss on functional usage without making the code too complex to understand. After discussing on how this change could be proposed, we could discuss on how making the good asserts.

Several approches may be done:  testing the generated schema file content, and/or testing metadata or other asserts. Can you help me deciding the good assert(s) to be added ?

sebersole:
We decided to not do this - the issue was rejected, I just forgot to close the PR

PR #2153, opened 2018-02-16
HHH-12297 - Relations are not loaded when using Fetch Profiles

ladislavovic:
* fix of the issue
* tests

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
@@ -7,6 +7,8 @@
 package org.hibernate.engine.internal;
 
 import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collection;
 
 import org.hibernate.AssertionFailure;
 import org.hibernate.CacheMode;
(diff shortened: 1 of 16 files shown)


hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
```diff
@@ -288,15 +296,78 @@ else if ( value!= PropertyAccessStrategyBackRefImpl.UNKNOWN ) {
...
+		for ( String fetchProfileName : loadQueryInfluencers.getEnabledFetchProfileNames() ) {
+			FetchProfile fp = session.getFactory().getFetchProfile( fetchProfileName );
+			result.add( fp );
+		}
+		return result;
+	}
+
+	private static boolean isEagerFetching(Fetch fetch) {
+		Fetch.Style style = fetch.getStyle();
+		return style == Fetch.Style.JOIN || style == Fetch.Style.SELECT;
```
vladmihalcea: FetchStyle.SELECT isn't associated to `FetchType.LAZY`?

ladislavovic: I am not sure I understand the question but generally this method decides if @FetchOverride section of Fetch Profile should cause eager loading or not.
Actually the method does not have to be here because org.hibernate.engine.profile.Fetch.Style has only two possible values - JOIN and SELECT, there is nothing like LAZY, so it always cause eager fetching, it is impossible now to change eager association to lazy by Fetch Profile. Maybe in the future?

vladmihalcea: EAGER can be overridden with Hibernate. Only Lazy can. For FetchProfile, SELECT is not even supported.

vladmihalcea:
@ladislavovic I created a new [PR](https://github.com/hibernate/hibernate-orm/pull/2163) based on yours after doing some changes.

PR #2154, opened 2018-02-16
HHH-12132: Schema creation uses non-unicode string types on SAP HANA

breglerj:
- add Parameter hibernate.dialect.hana.use_unicode_string_types that allows
  switching the database string types to unicode (nvarchar, nchar, nclob)


vladmihalcea:
Applied upstream, thanks

PR #2155, opened 2018-02-16
HHH-12059 HHH-11440 HHH-11286 HHH-10333 - 

dreab8:
bm2ddl.auto=update and update do not work with Oracle and SQLServer when Jdbc driver Connection implementation does not implement getSchema()

The issue is related with the `SchemaNameResolver`, if the `Connection` implementation does not implement the `getSchema()` method introduced with Java 7, but at runtime the `jdbcConnectionClass.getMethod( "getSchema" );`  may return not null ( I think depends on the JRE version used) and then when the 
`getSchemaMethod.invoke( connection )` is later executed an exception is thrown.

This is causing issues with the` InformationExtractorJdbcDatabaseMetaDataImpl#getTables()` method, due to the previous behaviour the value of the `schemaFilter` in 
```
extractionContext.getJdbcDatabaseMetaData().getTables(
					catalogFilter,
					schemaFilter,
					"%",
					tableTypes
			);
```
is an empty string causing the search to be limited to tables with no schema.

.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
@@ -744,6 +744,11 @@ public class Oracle8iDialect extends Dialect {
 	}
 
 	@Override
+	public String getCurrentSchemaCommand() {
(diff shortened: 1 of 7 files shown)


vladmihalcea:
Applied upstream, thanks

PR #2158, opened 2018-02-19
HHH-12290: convert all position parameters to named ones

bdshadow:


bdshadow:
argh, hurried up. `org.hibernate.test.hql.WithClauseTest.testWithJoinRHS` is failing. Working on it

gsmet:
@bdshadow I haven't tested your patch but keep in mind that a query mixing positional parameters and named parameters should throw an exception per the JPA spec. In the JPA 2.2 TCK, there are tests exercising that.

So you must be very careful to not replace all positional parameters by named ones too early (i.e. before the consistency check), otherwise the error won't be thrown.

I don't know how it's feasible but IMHO the core of the issue is to be able to move the consistency check before we expand the parameters as:
- we only care about throwing the consistency errors for the parameters of the original query passed to the API
- the internal query remodeled by ORM can indeed contain both positional parameters and named parameters, we don't really care about it.

BTW, +100 for your efforts on fixing this, I think it's a very important issue to fix before Final.

sebersole:
Also, we really should not be converting "all position parameters to named ones" across all queries... just cases where:

1) the query uses JPA "positional" parameters
2) the query uses multi-valued parameters

That union is the only time we should be doing this special handlng

sebersole:
Also, "the internal query remodeled by ORM can indeed contain both positional parameters and named parameters, we don't really care about it" is actually completely false.  The internals *do* care about it

gsmet:
@sebersole 

> The internals do care about it

Well, the internals did execute a mixed query before without any issue. So not sure I understand your "do care about it". AFAICS, it's just a matter of being JPA compliant so we just care about the original query which shouldn't contain mixed parameters, not the one we have transformed. At least that's my understanding.

sebersole:
@gsmet that depends.  HQL rendering did not care.  JPQL rendering does.  And you also have to keep in mind the switch in what a positional parameter is.  Again, before these changes, to Hibernate a positional parameter is, you know, a positional parameter (`?`) not a labelled (named) parameter (`?1`).  It has always been illegal in HQL as well to mix positional (real positional - `?`) and labelled. 

And again, JPA utterly disalows this as @bdshadow and I discussed on HipChat yesterday

bdshadow:
@sebersole @gsmet @vladmihalcea i think "Requires changes" tag can be removed already and pr can be merged, if you don't have any objections. I don't have access to manage labels

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/HqlSqlWalker.java
```diff
@@ -1089,9 +1089,9 @@ protected void beforeSelectClause() throws SemanticException {
 	@Override
 	protected AST generatePositionalParameter(AST delimiterNode, AST numberNode) throws SemanticException {
 		// todo : we check this multiple times
-		if ( namedParameters != null ) {
+		if (  getSessionFactoryHelper().isStrictJPAQLComplianceEnabled() && namedParameters != null ) {
 			throw new SemanticException(
-					"Cannot define positional and named parameterSpecs : " + queryTranslatorImpl.getQueryString()
+					"Cannot mix positional and named parameters : " + queryTranslatorImpl.getQueryString()
```
gsmet: I know it was preexisting but can you remove the space before the colon? It should be `and named parameters: " +`

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/HqlSqlWalker.java
```diff
@@ -1151,6 +1151,11 @@ else if ( existingValue instanceof Integer ) {
 
 	@Override
 	protected AST generateNamedParameter(AST delimiterNode, AST nameNode) throws SemanticException {
+		if ( getSessionFactoryHelper().isStrictJPAQLComplianceEnabled() && positionalParameters != null ) {
+			throw new SemanticException(
+					"Cannot mix positional and named parameters : " + queryTranslatorImpl.getQueryString()
```
gsmet: Same here.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
```diff
@@ -704,6 +704,138 @@ public void testJpaPositionalParameters() {
 			em.close();
 		}
 	}
+	
+	/**
+	 * Collection parameters are internally rewritten to named parameters (one named parameter by
+	 * collection item) and HQL parser rejects mixed positional and named parameters queries.
```
gsmet: The description of the test is not very accurate. Reading it, I thought it would be an error case but it is not.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
```diff
@@ -704,6 +704,138 @@ public void testJpaPositionalParameters() {
...
+	 */
+	@Test
+	@TestForIssue(jiraKey = "HHH-12290")
+	public void testParameterCollectionAndPositional() {
+		final Item item = new Item( "Mouse", "Micro$oft mouse" );
```
gsmet: Let's be neutral here and use `Microsoft`.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
```diff
@@ -704,6 +704,138 @@ public void testJpaPositionalParameters() {
...
+	}
+	
+	/**
+	 * Collection parameters are internally rewritten to named parameters (one named parameter by
+	 * collection item) and HQL parser rejects mixed positional and named parameters queries.
+	 */
+	@Test
+	@TestForIssue(jiraKey = "HHH-12290")
+	public void testParameterCollectionParenthesesAndPositional() {
+		final Item item = new Item( "Mouse", "Micro$oft mouse" );
```
gsmet: Same remark here about neutrality.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
```diff
@@ -704,6 +704,138 @@ public void testJpaPositionalParameters() {
...
+			throw e;
+		}
+		finally {
+			em.close();
+		}
+	}
+	
+	/**
+	 * Collection parameters are internally rewritten to named parameters (one named parameter by
+	 * collection item) and HQL parser rejects mixed positional and named parameters queries.
```
gsmet: Same remark here about the test description.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
```diff
@@ -704,6 +704,138 @@ public void testJpaPositionalParameters() {
...
+			em.getTransaction().commit();
+
+			em.getTransaction().begin();
+			Query q = em.createQuery( "select item from Item item where item.name in (?1) and item.descr = ?2" );
+			//test hint in value and string
+			q.setHint( "org.hibernate.fetchSize", 10 );
+			q.setHint( "org.hibernate.fetchSize", "10" );
+			List params = new ArrayList();
+			// for this case, 1-item collection is OK, but 2 or more is broken
+			// as 1-item collection are not "rewritten"
```
gsmet: This comment is not accurate anymore as it's supposed to work now.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
```diff
@@ -704,6 +704,138 @@ public void testJpaPositionalParameters() {
...
+			em.persist( item );
+			em.persist( item2 );
+			assertTrue( em.contains( item ) );
+			em.getTransaction().commit();
+
+			em.getTransaction().begin();
+			Query q = em.createQuery( "select item from Item item where item.name in (?1) and item.descr = ?2" );
+			//test hint in value and string
+			q.setHint( "org.hibernate.fetchSize", 10 );
+			q.setHint( "org.hibernate.fetchSize", "10" );
```
gsmet: Same remark about the hints.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
```diff
@@ -704,6 +704,138 @@ public void testJpaPositionalParameters() {
...
+			em.persist( item );
+			em.persist( item2 );
+			assertTrue( em.contains( item ) );
+			em.getTransaction().commit();
+
+			em.getTransaction().begin();
+			Query q = em.createQuery( "select item from Item item where item.name in ?1 and item.descr = ?2" );
+			//test hint in value and string
+			q.setHint( "org.hibernate.fetchSize", 10 );
+			q.setHint( "org.hibernate.fetchSize", "10" );
```
gsmet: I don't think these hints have any value here? I think they should be removed to simplify the test case.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
```diff
@@ -704,6 +704,138 @@ public void testJpaPositionalParameters() {
...
+	}
+	
+	/**
+	 * Collection parameters are internally rewritten to named parameters (one named parameter by
+	 * collection item) and HQL parser rejects mixed positional and named parameters queries.
+	 */
+	@Test
+	@TestForIssue(jiraKey = "HHH-12290")
+	public void testParameterCollectionSingletonParenthesesAndPositional() {
+		final Item item = new Item( "Mouse", "Micro$oft mouse" );
```
gsmet: Same remark about neutrality.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
```diff
@@ -704,6 +704,138 @@ public void testJpaPositionalParameters() {
...
+			em.persist( item );
+			em.persist( item2 );
+			assertTrue( em.contains( item ) );
+			em.getTransaction().commit();
+
+			em.getTransaction().begin();
+			Query q = em.createQuery( "select item from Item item where item.name in (?1) and item.descr = ?2" );
+			//test hint in value and string
+			q.setHint( "org.hibernate.fetchSize", 10 );
+			q.setHint( "org.hibernate.fetchSize", "10" );
```
gsmet: Same remark about the hints.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
```diff
@@ -704,6 +704,138 @@ public void testJpaPositionalParameters() {
...
+			List params = new ArrayList();
+			// for this case, 1-item collection is OK, but 2 or more is broken
+			// as 1-item collection are not "rewritten"
```
gsmet: Same remark about the outdated comment.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
bdshadow: @gsmet there are 33 occurrences of `Micro$oft` in this file. Should i change all of them?

gsmet: Argh OK, it's an existing pattern. Sigh. I really don't like this sort of things.

I would be for not adding more of them.

And maybe send another PR cleaning this up, once this one gets accepts (better not to add noise to this one).

sebersole: These have been here since we have moved from CVS which is what, 10 years now?  I think we have enough work to not fret over such silly things.  But by all means, if someone wants to take the time I won't stop them.

sebersole:
I'm really not sure what you are saying in this last comment @gsmet - the changes done by @bdshadow in fact do exactly what you state emphatically you believe should not happen.  The exception only happens when the user mixes them.  What we do on "expansion" no longer will be the cause of such exceptions.  Unless I misunderstood @bdshadow earlier, and I actually agree that should change then.  But like I said, I think his changes already address that

gsmet:
@sebersole from reading the patch, I thought what you said is only true if the JPA strict compliance mode is disabled. That's what worried me as I think this should work also if the JPA strict compliance mode is enabled.

If it works in both cases, then we're all set and I'm a very happy camper!

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
gsmet: Yeah, I just think it does not give a good image of the project. Anyway, I don't think it should pollute the discussion we have here so let's keep it aside.

sebersole: I don't disagree with you.  Just pointing out the matter of resources.

sebersole:
@gsmet my comment was simply that it is actually valid to mix either of the "labeled" param types in HQL.  JPA however *explicitly* disallows it.  That is what should be protected by compliance.

Again it has to do with Hibernate's legacy (real) positional parameters.   Historically HQL would not allow a query like `... where a.name = :someName and a.position = ?` - notice specifically the *un*-labeled `?`.  HQL has always (and as I understand it still does with this change) supported `... where a.name = :someName and a.position = ?1` - `?` versus `?1`.

Both however are illegal according to JPA

gsmet:
Yup, but in the case Laurent reported initially, the original query is `where a.name = ?1 and a.category in(?2)` thus entirely written with positional parameters.

This should work, even in the JPA compliance case, as the way the original query is rewritten is an implementation detail of ORM and shouldn't be of any concern to the user. So even if the query is rewritten with named parameters replacing ?2 and mix positional and named parameters, it should work in the strict JPA compliance mode as the original query is valid.

That's the specific case I'm worried about and AFAICS the final PR doesn't solve it as Dmitrii reverted all of his previous work in the last commit.

That's what I'm trying to explain from the beginning in my comments and, maybe I missed something obvious, as it seems clear to me from the beginning.

bdshadow:
@gsmet i updated the PR.

> This should work, even in the JPA compliance case, as the way the original query is rewritten is an implementation detail of ORM and shouldn't be of any concern to the user. So even if the query is rewritten with named parameters replacing ?2 and mix positional and named parameters, it should work in the strict JPA compliance mode as the original query is valid.
That is what i'm actually checking in `JPAQLComplianceTest.java`.

sebersole:
Its not re-written with positional parameters.. it is re-written with named
parameters - still labeled.

sebersole:
^^ unless that is something this PR changes... I did not look that deeply

gsmet:
> Its not re-written with positional parameters.. it is re-written with named parameters - still labeled.

That's what I said: `even if the query is rewritten with named parameters replacing ?2`. But you still have the `?1` parameter so the rewritten query contains both `?1` and `:x_1` (or similar).

@bdshadow can you explain to me what happens in strict JPA compliance mode with the tests you added? Do they throw an exception or do they work OK?

bdshadow:
@gsmet ok, let's take a look at [this one](https://github.com/hibernate/hibernate-orm/pull/2158/files#diff-241e47aeba826831f99638f0638adc62R114). We know that `(?1)` will be translated to a list of named parameters internally.  So after this translation all parameters are named, which seems to be ok, BUT initial query still contains both kinds of parameters so must fail. That's what we check in a `catch` block.

bdshadow:
linking @gsmet 's pr: https://github.com/hibernate/hibernate-orm/pull/2191

sebersole:
Applied #2191 upstream

PR #2161, opened 2018-02-21
HHH-12313: org.hibernate.jpa.test.transaction.TransactionCommitFailureTest fails on HANA

breglerj:


vladmihalcea:
Applied upstream, thanks

hibernate-core/src/main/java/org/hibernate/dialect/AbstractHANADialect.java
```diff
@@ -1497,8 +1501,11 @@ public void contributeTypes(TypeContributions typeContributions, ServiceRegistry
 					}
 				}
 			}
-			catch (SQLException e) {
-				// ignore
+			catch (Exception e) {
+				LOG.debug(
+						"An error occurred while trying to determing the value of the HANA parameter indexserver.ini / session / max_lob_prefetch_size. Using the default value "
```
gsmet: @vladmihalcea just noticed a typo here: s/determing/determine/

gsmet: Yeah, don't bother, I'd better push a fix myself :). Too used to the NoORM process :]

gsmet: Just pushed the fix.

breglerj: ok :-)

PR #2163, opened 2018-02-22
HHH-12297 - Relations are not loaded when using Fetch Profiles

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12297

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/fetchprofiles/EntityLoadedInTwoPhaseLoadTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/fetchprofiles/EntityLoadedInTwoPhaseLoadTest.java
@@ -1,0 +1,284 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.test.fetchprofiles;
+
+import javax.persistence.CascadeType;
+import javax.persistence.Column;
+import javax.persistence.Entity;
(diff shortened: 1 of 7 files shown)


ladislavovic:
It is OK according to me. I also used it in my project integration tests (which also test fetch profiles) and it passed.

hibernate-core/src/test/java/org/hibernate/test/fetchprofiles/EntityLoadedInTwoPhaseLoadTest.java
```diff
@@ -0,0 +1,284 @@
...
+import org.hibernate.annotations.FetchMode;
+import org.hibernate.annotations.FetchProfile;
+import org.hibernate.cfg.Configuration;
+import org.hibernate.cfg.Environment;
+
+import org.hibernate.testing.TestForIssue;
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.junit.Test;
+
+import static com.sun.tools.internal.ws.wsdl.parser.Util.fail;
```
breglerj: Is this intentional? Shouldn't this be `import static org.junit.Assert.fail`?

vladmihalcea: Thanks for noticing it. I'll fix it.

breglerj: While you’re fixing it, could you please also quote the entity name of the entity “Start”? That’ll fix the test on HANA.

vladmihalcea: Sure thing.

PR #2165, opened 2018-02-26
HHH-11867 - @UpdateTimestamp not working with @Inheritance( strategy = JOINED )

dreab8:
https://hibernate.atlassian.net/browse/HHH-11867.

The field is not updated when a field in the child entity is modified because` getTableUpdateNeeded( dirtyFields, hasDirtyCollection )` states the parent  table, where the `@UpdateTimestamp` field is declared, does not require to be updated.

Changes to `EntityMetamodel` class are deletions of private unused methods.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
@@ -95,7 +95,6 @@ import org.hibernate.loader.entity.UniqueEntityLoader;
 import org.hibernate.mapping.Column;
 import org.hibernate.mapping.Component;
 import org.hibernate.mapping.Formula;
-import org.hibernate.mapping.Join;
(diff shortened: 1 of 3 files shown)


dreab8:
Hi @vladmihalcea I have included your solution but changed a little bit the test, what do u think?

vladmihalcea:
In your latest commit there is no change to the test:

```
Showing  with 43 additions and 108 deletions.
+38 −21  hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+5 −87  hibernate-core/src/main/java/org/hibernate/tuple/entity/EntityMetamodel.java
```

dreab8:
@vladmihalcea  https://github.com/hibernate/hibernate-orm/pull/2165/commits/b5bd7ad890688a91063a0409068273353f223ed1

dreab8:
Thanks @vladmihalcea,

I'm going to integrate it.

dreab8:
applied upstream

PR #2167, opened 2018-02-28
HHH-12257 : Refreshing an entity clears the lock mode

gbadner:
https://hibernate.atlassian.net/browse/HHH-12257

This fix does not work when refreshing an entity locked with LockMode.WRITE.

Please see comments on code for details.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
@@ -173,16 +173,29 @@ public class DefaultRefreshEventListener implements RefreshEventListener {
 		String previousFetchProfile = source.getLoadQueryInfluencers().getInternalFetchProfile();
 		source.getLoadQueryInfluencers().setInternalFetchProfile( "refresh" );
 
-		// Use the entity's current LockMode if it is greater than event.getLockMode()
+		// Use the entity's current LockMode if it is greater than event.getLockMode(),
+		// unless the current LockMode is WRITE (special case described below).
 		final LockMode currentLockMode = e == null ? null : e.getLockMode();
 		LockOptions lockOptionsToUse = event.getLockOptions();
+		boolean resetLockModeToWrite = false;
 		if ( currentLockMode != null && currentLockMode.greaterThan( event.getLockMode() ) ) {
 			lockOptionsToUse = LockOptions.copy( event.getLockOptions(), new LockOptions() );
 			lockOptionsToUse.setLockMode( currentLockMode );
+			if ( currentLockMode == LockMode.WRITE ) {
+				// LockMode.WRITE is not a valid lock mode for loading, so
+				// set the LockMode to READ to load the entity. After loading,
+				// the LockMode must be set back to WRITE (below).
+				lockOptionsToUse.setLockMode( LockMode.READ );
+				resetLockModeToWrite = true;
+			}
+			else {
+				lockOptionsToUse.setLockMode( currentLockMode );
+			}
 		}
 		Object result = persister.load( id, object, lockOptionsToUse, source );
 		// Keep the same read-only/modifiable setting for the entity that it had before refreshing;
 		// If it was transient, then set it to the default for the source.
+		// Also, set the LockMode to WRITE if resetLockModeToWrite == true.
 		if ( result != null ) {
 			if ( !persister.isMutable() ) {
 				// this is probably redundant; it should already be read-only
@@ -190,6 +203,11 @@ public class DefaultRefreshEventListener implements RefreshEventListener {
 			}
 			else {
 				source.setReadOnly( result, ( e == null ? source.isDefaultReadOnly() : e.isReadOnly() ) );
+			}
+			if ( resetLockModeToWrite ) {
+				// Get the new EntityEntry and set the LockMode to WRITE.
+				final EntityEntry eLoaded = source.getPersistenceContext().getEntry( result );
+				eLoaded.setLockMode( LockMode.WRITE );
 			}
 		}
 		source.getLoadQueryInfluencers().setInternalFetchProfile( previousFetchProfile );
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
@@ -10,7 +10,11 @@ import java.io.Serializable;
 import java.util.IdentityHashMap;
 import java.util.Map;
 
+import javax.persistence.LockModeType;
+
 import org.hibernate.HibernateException;
+import org.hibernate.LockMode;
+import org.hibernate.LockOptions;
 import org.hibernate.PersistentObjectException;
 import org.hibernate.UnresolvableObjectException;
 import org.hibernate.action.spi.AfterTransactionCompletionProcess;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
```diff
@@ -173,7 +177,15 @@ public void doAfterTransactionCompletion(boolean success, SharedSessionContractI
...
-		Object result = persister.load( id, object, event.getLockOptions(), source );
+
+		// Use the entity's current LockMode if it is greater than event.getLockMode()
+		final LockMode currentLockMode = e == null ? null : e.getLockMode();
+		LockOptions lockOptionsToUse = event.getLockOptions();
+		if ( currentLockMode != null && currentLockMode.greaterThan( event.getLockMode() ) ) {
+			lockOptionsToUse = LockOptions.copy( event.getLockOptions(), new LockOptions() );
+			lockOptionsToUse.setLockMode( currentLockMode );
+		}
+		Object result = persister.load( id, object, lockOptionsToUse, source );
```
gbadner: The statement above will fail if ```lockOptionsToUse.getLockMode() == LockMode.WRITE```, because there is no ```UniqueEntityLoader``` that is appropriate for ```LockMode.WRITE```.

This is not a surprise; Javadoc for ```LockMode.WRITE``` says it is not valid for for loading. [1]

The problem is that this will happen whenever an entity that is inserted or updated gets refreshed,  because the ```LockMode``` for the inserted/updated entity is set to ```LockMode.WRITE```.

Which ```UniqueKeyLoader``` should be used in this case? Without the fix, it is the ```UniqueKeyLoader``` with key "refresh". 

After the entity is refreshed, should the entity's ```LockMode``` still be ```LockMode.WRITE```, or should it have some other value?

[1] https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/LockMode.java#L61-L67

hibernate-core/src/test/java/org/hibernate/test/locking/LockModeTest.java
```diff
@@ -153,6 +159,75 @@ public void testQueryLockModePessimisticWriteWithAlias() {
...
+	@Test
+	@TestForIssue(jiraKey = "HHH-12257")
+	public void testRefreshAfterUpdate() {
+		doInHibernate( this::sessionFactory, session -> {
+			A a = session.get( A.class, id );
+			checkLockMode( a, LockMode.READ, session );
+			a.setValue( "new value" );
+			session.flush();
+			checkLockMode( a, LockMode.WRITE, session );
+			session.refresh( a );
```
gbadner: The statement above will fail because the ```LockMode``` for the updated entity is ```LockMode.WRITE```.

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
sebersole: If the LockMode is WRITE, we already own the exclusive lock on that row, its impossible for another process to update it meaning that there ought to be no changes to "pull back".  From this perspective we could just skip the refresh.

However, the one use-case we do need to worry about (at least the only one I can think of) are db generated values where the user is not taking advantage of Hibernate's generated-value support.  The JPA-standard way to deal with that is to refresh after the persist/merge.

So I think the general case for currentLockMode == WRITE could be to fallback to READ for the load, but to reset the entry's LockMode to WRITE after.

BTW, wouldn't we have the same problem for PESSIMISTIC_WRITE (and PESSIMISTIC_READ possibly)?

sebersole:
@gbadner was there something else you were waiting on for feedback here?

sebersole:
@gbadner ?

gbadner:
@sebersole, sorry for the delay. Thanks for the review. I'm wrapping it up now.

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
gbadner: There isn't a problem with PESSIMISTIC_WRITE or PESSIMISTIC_READ because ```AbstractEntityPersister#loaders``` has entries for both of those lock modes.

gbadner:
@sebersole, I've added a commit to deal with LockMode.WRITE. Please let me know if this looks OK to you.

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
sebersole: I think this can be made much simpler.  Working on an alternate PR for just this last little piece (WRITE)

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
```diff
@@ -168,9 +172,30 @@ public void onRefresh(RefreshEvent event, Map refreshedAlready) {
 
 		String previousFetchProfile = source.getLoadQueryInfluencers().getInternalFetchProfile();
 		source.getLoadQueryInfluencers().setInternalFetchProfile( "refresh" );
-		Object result = persister.load( id, object, event.getLockOptions(), source );
+
+		// Use the entity's current LockMode if it is greater than event.getLockMode(),
+		// unless the current LockMode is WRITE (special case described below).
+		final LockMode currentLockMode = e == null ? null : e.getLockMode();
+		LockOptions lockOptionsToUse = event.getLockOptions();
+		boolean resetLockModeToWrite = false;
```
sebersole: My last review comment should have been attached here, but as I replied to our previous conversations (line comments on out-of-date lines) it is not showing as part of the review...

I think this can be made much simpler. Working on an alternate PR for just this last little piece (WRITE)

gbadner: Oops, sorry about that. Here's your comment:

"If the LockMode is WRITE, we already own the exclusive lock on that row, its impossible for another process to update it meaning that there ought to be no changes to "pull back". From this perspective we could just skip the refresh.

However, the one use-case we do need to worry about (at least the only one I can think of) are db generated values where the user is not taking advantage of Hibernate's generated-value support. The JPA-standard way to deal with that is to refresh after the persist/merge.

So I think the general case for currentLockMode == WRITE could be to fallback to READ for the load, but to reset the entry's LockMode to WRITE after.

BTW, wouldn't we have the same problem for PESSIMISTIC_WRITE (and PESSIMISTIC_READ possibly)?"

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
```diff
@@ -191,6 +204,11 @@ public void onRefresh(RefreshEvent event, Map refreshedAlready) {
 			else {
 				source.setReadOnly( result, ( e == null ? source.isDefaultReadOnly() : e.isReadOnly() ) );
 			}
+			if ( resetLockModeToWrite ) {
+				// Get the new EntityEntry and set the LockMode to WRITE.
+				final EntityEntry eLoaded = source.getPersistenceContext().getEntry( result );
```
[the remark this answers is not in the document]
gbadner: That won't work because there is a new ```EntityEntry``` for the refreshed entity (different from the ```e```).

sebersole:
Yes, which surprised me.  And I wrote a note down to come back and see
why.  Seems like it ought to be the same entry instance

gbadner:
#2256 supersedes this PR.

PR #2169, opened 2018-03-01
HHH-12221: incorrect formatting of SQL Server statements when escaping

bdshadow:


vladmihalcea:
Applied upstream, thanks

PR #2172, opened 2018-03-02
HHH-12327 - Remove the Envers dependency on Javassist.

Naros:
https://hibernate.atlassian.net/browse/HHH-12327


Naros:
Applied upstream.

PR #2173, opened 2018-03-02
HHH-7809: Use SessionFactory.name for the CacheManager... or a UUID

alexsnaps:
Closes HHH-7809 // @chrisdennis wdyt?

### DIFF
--- a/hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/ConfigSettings.java
+++ b/hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/ConfigSettings.java
@@ -12,8 +12,6 @@ import net.sf.ehcache.CacheManager;
  * @author Steve Ebersole
  */
 public interface ConfigSettings {
-	String SIMPLE_FACTORY_NAME = "jcache";
-
 	String PROP_PREFIX = "hibernate.cache.ehcache.";
 
 	/**
--- a/hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/ConfigSettings.java
+++ b/hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/ConfigSettings.java
@@ -23,4 +23,6 @@ public interface ConfigSettings {
 	 * This is the legacy property name.  No need to change it to fit under {@link #PROP_PREFIX}
 	 */
 	String EHCACHE_CONFIGURATION_RESOURCE_NAME = "net.sf.ehcache.configurationResourceName";
+
+	String EHCACHE_CONFIGURATION_CACHE_MANAGER_NAME = "net.sf.ehcache.cacheManagerName";
 }
--- a/hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/internal/HibernateEhcacheUtils.java
+++ b/hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/internal/HibernateEhcacheUtils.java
@@ -7,12 +7,17 @@
 package org.hibernate.cache.ehcache.internal;
 
 import java.net.URL;
+import java.util.Map;
+import java.util.UUID;
 
 import net.sf.ehcache.config.CacheConfiguration;
 import net.sf.ehcache.config.Configuration;
 import net.sf.ehcache.config.ConfigurationFactory;
 import net.sf.ehcache.config.NonstopConfiguration;
 import net.sf.ehcache.config.TimeoutBehaviorConfiguration.TimeoutBehaviorType;
+import org.hibernate.boot.spi.SessionFactoryOptions;
+
+import static org.hibernate.cache.ehcache.ConfigSettings.EHCACHE_CONFIGURATION_CACHE_MANAGER_NAME;
 
 
 /**
@@ -60,6 +65,28 @@ public final class HibernateEhcacheUtils {
 			}
 		}
 		return config;
+	}
+
+	static void setCacheManagerNameIfNeeded(SessionFactoryOptions settings, Configuration configuration, Map properties) {
+		overwriteCacheManagerIfConfigured( configuration, properties );
+		if (configuration.getName() == null) {
+			String sessionFactoryName = settings.getSessionFactoryName();
+			if (sessionFactoryName != null) {
+				configuration.setName( sessionFactoryName );
+			} else {
+				configuration.setName( "Hibernate " + UUID.randomUUID().toString() );
+			}
+		}
+	}
+
+	static Configuration overwriteCacheManagerIfConfigured(final Configuration configuration, final Map properties) {
+		if (properties != null) {
+			final String cacheManagerName = (String) properties.get( EHCACHE_CONFIGURATION_CACHE_MANAGER_NAME );
+			if (cacheManagerName != null) {
+				configuration.setName( cacheManagerName );
+			}
+		}
+		return configuration;
 	}
 
 	private static void setupHibernateTimeoutBehavior(NonstopConfiguration nonstopConfig) {
--- a/hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/internal/HibernateEhcacheUtils.java
+++ b/hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/internal/HibernateEhcacheUtils.java
@@ -73,7 +73,8 @@ public final class HibernateEhcacheUtils {
 			String sessionFactoryName = settings.getSessionFactoryName();
 			if (sessionFactoryName != null) {
 				configuration.setName( sessionFactoryName );
-			} else {
+			}
+			else {
 				configuration.setName( "Hibernate " + UUID.randomUUID().toString() );
 			}
 		}
--- a/hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/internal/EhcacheRegionFactory.java
+++ b/hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/internal/EhcacheRegionFactory.java
@@ -32,6 +32,7 @@ import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.jboss.logging.Logger;
 
 import static org.hibernate.cache.ehcache.ConfigSettings.EHCACHE_CONFIGURATION_RESOURCE_NAME;
+import static org.hibernate.cache.ehcache.internal.HibernateEhcacheUtils.setCacheManagerNameIfNeeded;
 
 /**
  * @author Steve Ebersole
@@ -144,11 +145,13 @@ public class EhcacheRegionFactory extends RegionFactoryTemplate {
 			}
 			if ( configurationResourceName == null || configurationResourceName.length() == 0 ) {
 				final Configuration configuration = ConfigurationFactory.parseConfiguration();
+				setCacheManagerNameIfNeeded( settings, configuration, properties );
 				return new CacheManager( configuration );
 			}
 			else {
 				final URL url = loadResource( configurationResourceName, settings );
 				final Configuration configuration = HibernateEhcacheUtils.loadAndCorrectConfiguration( url );
+				setCacheManagerNameIfNeeded( settings, configuration, properties );
 				return new CacheManager( configuration );
 			}
 		}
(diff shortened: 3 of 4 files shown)


hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/EhCacheRegionFactory.java
```diff
@@ -106,4 +109,14 @@ public void stop() {
 		}
 	}
 
+	static void setCacheManagerNameIfNeeded(SessionFactoryOptions settings, Configuration configuration) {
+		if ( configuration.getName() == null ) {
+			String sessionFactoryName = settings.getSessionFactoryName();
+			if ( sessionFactoryName != null ) {
+				configuration.setName( sessionFactoryName );
+			} else {
+				configuration.setName( "Hibernate " + UUID.randomUUID().toString() );
```
alexsnaps: Meh :/

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/EhCacheRegionFactory.java
```diff
@@ -106,4 +109,14 @@ public void stop() {
 		}
 	}
 
+	static void setCacheManagerNameIfNeeded(SessionFactoryOptions settings, Configuration configuration) {
```
sebersole: Longer term I wonder if we should instead pass in the SessionFactory reference.  SessionFactory already defines a UUID...

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/SingletonEhCacheRegionFactory.java
```diff
@@ -81,6 +83,7 @@ public void start(SessionFactoryOptions settings, Properties properties) throws
 					url = loadResource( configurationResourceName );
 				}
 				final Configuration configuration = HibernateEhcacheUtils.loadAndCorrectConfiguration( url );
+				setCacheManagerNameIfNeeded( settings, configuration );
```
chrisdennis: This seems a bit needless.  It's using the singleton cache manager anyway.

alexsnaps: Yep. I thought it'd make it clearer for users, but the singleton one could be used for … god knows what else… 🔥 this

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/EhCacheRegionFactory.java
```diff
@@ -66,11 +67,13 @@ public void start(SessionFactoryOptions settings, Properties properties) throws
 			}
 			if ( configurationResourceName == null || configurationResourceName.length() == 0 ) {
 				final Configuration configuration = ConfigurationFactory.parseConfiguration();
+				setCacheManagerNameIfNeeded( settings, configuration );
```
chrisdennis: Is this right... or should this method instead be using newInstance() and returning the same cache manager?

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/EhCacheRegionFactory.java
alexsnaps: Yeah, that'd be great… I saw the name could be `null` in the `SessionFactoryOptions`, so… free styled. But if the `SessionFactory` _always_ named, it be nicer indeed.

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/EhCacheRegionFactory.java
alexsnaps: You talk about line 71 now, yes?
```java
manager = new CacheManager( configuration );
```

My memory somewhat fails me here, but I think one `SessionFactory` instance is bound to one `CacheManager` instance. I don't know we'd want to share ever… but maybe I'm wrong. 
I _could_ change that, I don't think it'd be a big deal… previously throwing config would now reuse an existing `CacheManager` instance if one'd be available. Could make sense. I let you @chrisdennis make the call… You're the man in charge here ;)

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/EhCacheRegionFactory.java
sebersole: It would be great.  The problem though is that Hibernate itself does not always have good visibility into a good name.  A "container" (WidFly, Spring, etc) has a much better idea of that.  Wildfly, e.g., does in deed set the SF name if the user has not.

It's hard to enforce that that always happen uniquely though.

However, there is always a UUID : `org.hibernate.engine.spi.SessionFactoryImplementor#getUuid`

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/EhCacheRegionFactory.java
chrisdennis: Okay - that makes sense. The last thing I have to say is: https://jira.terracotta.org/jira/browse/EHC-885

sebersole: Some providers do allow multiple SessionFactories to "bind" to a single cache instance.  This is actually why the notion of a "region name prefix" was added.

Up to you guys if you want to support that.

chrisdennis: I really don't *want* to support anything on the 2.x line any more ;)

sebersole: `hibernate-ehcache` currently pulls in `net.sf.ehcache:ehcache:2.10.3` while `hibernate-jcache` pulls in `org.ehcache:ehcache:3.0.0` (although only as a test dep).

Did you guys want to update that to only use 3?

chrisdennis: My understanding is that the last place for all this was "lets try and make Ehcache 3 not need it's own provider".  I don't think thats changed, which leaves the hibernate-ehcache in this weird place. @ljacomet care to join this party with an opinion?

alexsnaps: I'll check wrt EHC-885 this weekend and see how we can have this work as part of this. 
Expect the thing to be final by Monday. wrt 3.x et al… I'll let the Terracotta folks call the shot :)

sebersole: FWIW `hibernate-infinispan` got "relocated" to the Infinispan project for all of the same reasons.

We could do a similar relocation for Ehcache, basically dropping support for Ehcache 2

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/ConfigSettings.java
```diff
@@ -12,8 +12,6 @@
  * @author Steve Ebersole
  */
 public interface ConfigSettings {
-	String SIMPLE_FACTORY_NAME = "jcache";
```
alexsnaps: @sebersole Nothing requires that `SIMPLE_FACTORY_NAME` to be defined in these interfaces right? No magic I could find, but…

alexsnaps:
@henri-tremblay Here's what I _think_ we want. I'll look into adding _some_ test later this afternoon

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/ConfigSettings.java
sebersole: @alexsnaps It is just the name under which we register the factory for users to reference in config.  We recommend users use "short naming" to avoid refactoring impacts.  

This one is particular though is just a bad copy/paste.  The Ehcache provider actually registers 2 factories (`ehcache` and `ehcache-singleton`) which we could add to this ConfigSettings.

The intention is for users to have a static reference, much like we do with the larger  {{AvailableSettings}}.

henri-tremblay: Don't we need an "ehcache" constant somewhere?

henri-tremblay:
Thanks @alexsnaps. Everything seems meaningful. I'll let @chrisdennis play with it since he's the one on the topic.

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/internal/HibernateEhcacheUtils.java
```diff
@@ -62,6 +67,29 @@ public static Configuration loadAndCorrectConfiguration(URL url) {
...
 
+	static void setCacheManagerNameIfNeeded(SessionFactoryOptions settings, Configuration configuration, Map properties) {
+		overwriteCacheManagerIfConfigured( configuration, properties );
+		if (configuration.getName() == null) {
+			String sessionFactoryName = settings.getSessionFactoryName();
+			if (sessionFactoryName != null) {
+				configuration.setName( sessionFactoryName );
+			}
+			else {
+				configuration.setName( "Hibernate " + UUID.randomUUID().toString() );
```
sebersole: IMO it would be better to use the SessionFactory's UUID.  However, currently this code does not have access to that because only the `SessionFactoryOptions` are passed, not the `SessionFactoryImplementor`.

@dreab8 @Naros what do you think about either:

1. Managing the SF UUID on `SessionFactoryOption` instead (pro - wider usage/availability; con -?)
2. Passing the UUID to the `RegionFactory` start up.
3. Passing the `SessionFactoryImplementor` to the `RegionFactory` start up.

(3) is my least favorite

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/ConfigSettings.java
sebersole: Define "need"

sebersole: See `org.hibernate.cache.ehcache.internal.StrategyRegistrationProviderImpl`

henri-tremblay: @sebersole Yes. You answer arrived while I was typing. So all good. We said the same thing.

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/internal/HibernateEhcacheUtils.java
sebersole: https://hibernate.atlassian.net/browse/HHH-12440

sebersole: Unless we decide to drop `hibernate-ehcache` per other discussion

sebersole: @alexsnaps @henri-tremblay @chrisdennis What do you guys think of using the SessionFactory's UUID instead?  Worth the (small) effort?

henri-tremblay: It's nice to match UUIDs. If it seems useful for other purposes, why not. Otherwise, it's probably not essential

sebersole: Right, that was my thought.  It might help in other places and we can still expose it from SessionFactory.. so why not

sebersole:
Applied upstream with UUID changes discussed here.  Thanks guys!

PR #2174, opened 2018-03-03
HHH-12346: Replace StringHelper#join by Java's String#join

kinow:
This pull request replaces a couple of `StringHelper#join` methods and its occurrences by Java 8's `String#join`.

As I mentioned in the JIRA ticket, this is a left-over after I submitted another pull request. I had noted this class methods, and noticed at least the join could be replaced.

Had to use streams in three places to convert from `Integer` to `String`.

Not a problem if it is decided not to touch this method for being in core module, or because it just works :-) Otherwise, let me know if further testing, syntax, formatting, etc, is necessary. Happy to update the ticket with new commits, squash, rebase, etc.

Cheers
Bruno

### DIFF
--- a/hibernate-c3p0/src/main/java/org/hibernate/c3p0/internal/C3P0ConnectionProvider.java
+++ b/hibernate-c3p0/src/main/java/org/hibernate/c3p0/internal/C3P0ConnectionProvider.java
@@ -21,7 +21,6 @@ import org.hibernate.boot.registry.classloading.spi.ClassLoadingException;
 import org.hibernate.cfg.Environment;
 import org.hibernate.engine.jdbc.connections.internal.ConnectionProviderInitiator;
 import org.hibernate.engine.jdbc.connections.spi.ConnectionProvider;
-import org.hibernate.internal.util.StringHelper;
 import org.hibernate.internal.util.config.ConfigurationHelper;
 import org.hibernate.service.UnknownUnwrapTypeException;
 import org.hibernate.service.spi.Configurable;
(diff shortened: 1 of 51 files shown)


kinow:
Travis CI failed in the `check` part. I did only `clean test` (more familiar with Maven, sorry). Updated code, re-run with check, and confirmed this part passed now. Let's see if Travis CI can build it fine now.

sebersole:
I'd prefer to not use Streams for the conversion.  We have run perf tests on using Streams for such operations and they perform significantly worse.  I did not see such calls in my quick look over the changes - just based in your comment.

A better option might have been to simply have overloaded methods on StringHelper to take Strings versus Objects.  Calls involving Strings could have delegated to String#join

kinow:
Good point @sebersole . I've found situations where debugging problems was trickier because of streams, so I don't see why not try a different approach.

Reviewing the changes, also noted another part that could be simplified. Now instead of streams, or another method, simply calling `#iterator()` appears to work fine. WDYT?

Updated, `./gradlew clean test check` passed after some 20 minutes locally, so hopefully Travis build will be OK as well.

sebersole:
> simply calling #iterator() appears to work fine. WDYT?

I'm not sure what this means

kinow:
Sorry for not being clear on the iterator use @sebersole 

I meant that instead of maintaining a mehod for iterable, used in only a couple of places, we can simply maintain an iterator method, and call .iterator ou iterables, keySet, etc... like in here https://github.com/kinow/hibernate-orm/blob/918a50326ff8c7b893edf6aee74638c82c10d29b/hibernate-core/src/main/java/org/hibernate/query/internal/ParameterMetadataImpl.java#L68

PR #2175, opened 2018-03-05
HHH-11924 ElementCollection ignore converter for XML m…

lrsrhn:
Fix for orm xml mapping where convert tags where ignored when child to element-collection tags.
https://hibernate.atlassian.net/browse/HHH-11924


vladmihalcea:
Applied upstream, thanks

PR #2176, opened 2018-03-05
User Guide: fix adoc typo in "Using table identifier generator"

rillig-ite:


vladmihalcea:
Applied upstream, thanks

PR #2177, opened 2018-03-05
User Guide: fix SQL for bidirectional OneToMany

rillig-ite:
The documented SQL didn't match what the test actually produces.
It made me wonder why the person was not persisted at all.


vladmihalcea:
Applied upstream, thanks

PR #2178, opened 2018-03-05
Documentation: reword many instances of "going to"

rillig-ite:
"going to" is used for plans that might happen in the future. In the
User Guide and similar documents, this is not the right wording in many
places where the reader expectes a strict guarantee instead of a
"maybe".


vladmihalcea:
Thanks, but [`going to`](https://en.wikipedia.org/wiki/Going-to_future) does not imply a possibility. It just expresses an action that will occur in future. More, I noticed that you modified some Time clauses where the future must be used instead of the present tense.

Thanks, but I think we should close this issue.

@sebersole @Naros @gbadner Since you are native English speakers, can you validate my assumptions. Thanks.

vladmihalcea:
Closing as I don't consider this an issue.

PR #2179, opened 2018-03-05
HHH-12357 Demonstrate that FK names depend on default encoding

rillig-ite:


vladmihalcea:
Applied upstream, thanks

PR #2180, opened 2018-03-05
HHH-12332 Fix for NPE in AbstractPropertyMapping.getSuperCollection

beikov:
https://hibernate.atlassian.net/browse/HHH-12332

Still running tests locally, but this should do it. Essentially, I compare the element types for "equality" to determine whether a collection was defined with a type variable on an abstract entity base type and if it is, I bail out. I was sketching a proper fix for type variable use with collections as well here: https://github.com/beikov/hibernate-orm/tree/HHH-12332-2

but it's not so easy since there is no `CollectionBinding` for type variable defined collections on abstract entity types.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractPropertyMapping.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractPropertyMapping.java
@@ -17,9 +17,7 @@ import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.internal.util.StringHelper;
 import org.hibernate.internal.util.collections.ArrayHelper;
-import org.hibernate.mapping.Collection;
-import org.hibernate.mapping.MappedSuperclass;
-import org.hibernate.mapping.PersistentClass;
+import org.hibernate.mapping.*;
 import org.hibernate.sql.Template;
 import org.hibernate.type.AnyType;
 import org.hibernate.type.AssociationType;
@@ -186,7 +184,7 @@ public abstract class AbstractPropertyMapping implements PropertyMapping {
 				Collection thisCollection = metadata.getCollectionBinding( ( (CollectionType) existingType ).getRole() );
 				Collection otherCollection = metadata.getCollectionBinding( ( (CollectionType) type ).getRole() );
 
-				if ( thisCollection == otherCollection ) {
+				if ( thisCollection.isSame( otherCollection ) ) {
 					logDuplicateRegistration(
 							path,
 							existingType,
@@ -195,14 +193,7 @@ public abstract class AbstractPropertyMapping implements PropertyMapping {
 					return;
 				}
 
-				Collection commonCollection = getSuperCollection(
-						metadata,
-						thisCollection.getOwner(),
-						otherCollection.getOwner(),
-						thisCollection.getReferencedPropertyName()
-				);
-
-				newType = commonCollection.getType();
+                throw new IllegalStateException( "Collection mapping in abstract entity type with a type variable is unsupported! Couldn't add property '" + path + "' with type: " + type );
 			}
 			else if ( type instanceof EntityType ) {
 				EntityType entityType1 = (EntityType) existingType;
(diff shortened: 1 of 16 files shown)


gsmet:
@beikov I rebased the PR and pushed your branch as the build was failing due to the Agroal snapshot dependency (which has been removed in master). Hopefully, Travis will be happy now.

gsmet:
Added a commit to fix some formatting issues.

dreab8:
Hi @beikov ,the following test
```
public class JoinedInheritanceTest extends BaseCoreFunctionalTestCase {
	@Override
	protected Class<?>[] getAnnotatedClasses() {
		return new Class[] { BaseEntity.class, EntityA.class, EntityB.class, EntityC.class, EntityD.class };
	}

	@Test
	public void testIt() {
		TransactionUtil.doInHibernate( this::sessionFactory, session -> {

		} );
	}

	@Entity(name = "BaseEntity")
	@Inheritance(strategy = InheritanceType.JOINED)
	public static class BaseEntity {
		@Id
		private long id;
	}

	@Entity(name = "EntityA")
	public static class EntityA extends BaseEntity {
		@OneToMany(fetch = FetchType.LAZY)
		private Set<EntityC> attributes;
	}

	@Entity(name = "EntityB")
	public static class EntityB extends BaseEntity {
		@OneToMany(fetch = FetchType.LAZY)
		private Set<EntityD> attributes;
	}

	@Entity(name = "EntityC")
	public static class EntityC {
		@Id
		private long id;
	}

	@Entity(name = "EntityD")
	public static class EntityD {
		@Id
		private long id;
	}
```
fails with an
`Caused by: java.lang.IllegalStateException: Collection mapping in abstract entity type with a type variable is unsupported! Couldn't add property 'attributes' with type: org.hibernate.type.SetType(org.dreab8.playground.Inheritance.JoinedInheritanceTest$EntityB.attributes)`

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractPropertyMapping.java
```diff
@@ -195,14 +193,16 @@ else if ( type instanceof CollectionType ) {
...
-
-				newType = commonCollection.getType();
+				// When we discover incompatible types, we register "null" as property type to signal that the property is not resolvable on the parent type
+				newType = null;
+				if ( LOG.isTraceEnabled() ) {
+					LOG.tracev(
+							"Collection mapping in abstract entity type with a type variable is unsupported! Couldn't add property [{0}], existing type = [{1}], incoming type = [{2}]",
+							path,
+							existingType,
+							type
```
dreab8: am I wrong or most of the time the log does not convey the right information, like in the case of the added test?

beikov: RIght, adapting the message to say something like "Skipping property..." would be more appropriate.

sebersole: I agree with @dreab8 here that the message should not be logged if the super does not define this attribute.  It would be confusing to me as a user to see that log message in the case Andrea brings up

beikov: In 5.2.13 and before a "duplicate attribute" trace log is logged. I could omit the log if you like or adapt the message.

sebersole: Well @dreab8 's use case is such that it is actually not a duplicated attribute. E.g.

````
Entity(name = "BaseEntity")
@Inheritance(strategy = InheritanceType.JOINED)
public static class BaseEntity {
	@Id
	private long id;
}

@Entity(name = "EntityA")
public static class EntityA extends BaseEntity {
	@OneToMany(fetch = FetchType.LAZY)
	private Set<EntityC> attributes;
}

@Entity(name = "EntityB")
public static class EntityB extends BaseEntity {
	@OneToMany(fetch = FetchType.LAZY)
	private Set<EntityD> attributes;
}

@Entity(name = "EntityC")
public static class EntityC {
	@Id
	private long id;
}

@Entity(name = "EntityD")
public static class EntityD {
	@Id
	private long id;
}
````

Here the 2 `#attributes` attributes are not duplicate.  And I'm not so sure that the old code logged that as you say.  I could be wrong - in which case imo even the old code would have been wrong

beikov: It's a duplicate attribute for `BaseEntity`. Note that the attributes are added to the base types as well. I agree that it's not _correct_ that it's a duplicate, but technically we are adding a duplicate. At the end of the day, this is a trace message which IMO is only for internal purposes and rarely interesting for the end user/developer.

beikov:
@dreab8 @sebersole Hope this message is ok?

dreab8:
@beikov yes the message is better :)
Thanks a lot for all the work done.

toughpheeckouse:
Hi @beikov 

I get this NPE for attached sample
[hibernate-sandbox.zip](https://github.com/hibernate/hibernate-orm/files/1789512/hibernate-sandbox.zip)
Could you test it for your PR?

Abount sample - this is Spring Boot Application with sample model that emulate real model

if you run it like this
mvn clean install
java -jar target/hibernate-sandbox-0.0.1-SNAPSHOT.jar

then you'll get NPE

But if you run it this way
mvn clean install -P hibernate-12
java -jar -Dspring.profiles.active=hibernate-12 target/hibernate-sandbox-0.0.1-SNAPSHOT.jar

then everething will ok because it will use hibernate 5.2.12.Final

beikov:
@toughpheeckouse thanks for the test case. I verified that it works after applying this fix.

toughpheeckouse:
@beikov thanks!

sebersole:
You rock @beikov <https://github.com/beikov> !  Thanks for your work on this

beikov:
Thanks :)

dreab8:
Applied upstream.
Thanks @beikov

PR #2183, opened 2018-03-06
Remove unnecessary equals/hashCode from ManyToMany test case

rillig-ite:
Since the PersonAddress is the only class that is used inside a
collection, only that class needs to implement equals and hashCode.


vladmihalcea:
That's wrong. The `Person` and `Address` equals and hashCode are needed.

    return Objects.equals( person, that.person ) && Objects.equals( address, that.address );

rillig-ite:
But if they are needed, I wonder why the test also runs without them. Especially since the test should demonstrate exactly that equals/hashCode are needed in this case.

vladmihalcea:
The test runs even if the default equals and hashCode are used. But just because it does not fail, it does not mean you should omit those. For more details, check out [this article](https://vladmihalcea.com/the-best-way-to-implement-equals-hashcode-and-tostring-with-jpa-and-hibernate/).

PR #2184, opened 2018-03-06
User Guide: fix OrderBy example

rillig-ite:
Since the relative lengths of the article's name and its content are
the same, this bug could go unnoticed.


rillig-ite:
The text surrounding this code says that the articles are sorted by the length of their content.

The code sorted them by their name, which in this example had the same effect, but only coincidentally.

vladmihalcea:
That's why you need to create a Jira issue and describe the problem. Otherwise, it's not clear what you are fixing here.

vladmihalcea:
Fixed using [this commit instead](https://github.com/hibernate/hibernate-orm/commit/069222cf58eaa20d83cc128d95d0cb7f45437358).

PR #2187, opened 2018-03-07
HHH-12369 Fix for integer overflow in limit handler when using Integer.MAX_VALUE for maxResults on DB2

beikov:
https://hibernate.atlassian.net/browse/HHH-12369


dreab8:
Applied upstream.
Thanks @beikov

PR #2190, opened 2018-03-09
HHH-12375 Fix for eager loading same named properties with conflicting types

beikov:
https://hibernate.atlassian.net/browse/HHH-12375

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractPropertyMapping.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractPropertyMapping.java
@@ -127,6 +127,18 @@ public abstract class AbstractPropertyMapping implements PropertyMapping {
 		}
 	}
 
+	private void logIncompatibleRegistration(String path, Type existingType, Type type) {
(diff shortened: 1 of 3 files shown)


cnsgithub:
I merged your fix in hibernate-core-5.2.16-snapshot and can confirm that it works correctly again. Thanks for fixing it, great work.

Will your fix be contained in 5.2.16? And if so, when will 5.2.16 be released?

Greets to Austria

beikov:
Good to hear that this works for you. I'm not sure when 5.2.16 will be released, but this will most definitely be part of that release. :)

beikov:
@dreab8 I also added a _fix_ for HHH-12383 which apparently models something similar to the following 

```
@Entity class A {}
@Entity class B extends A { B relation; }
@Entity class C extends A { Set<C> relation; }
```

The essence is, that the attributes have the same name, but different _types_ i.e. one is an `EntityType` whereas the other is a `CollectionType`. The user actually reported a problem with `AnyType` but it's essentially the same.

dreab8:
not sure if [this](https://github.com/hibernate/hibernate-orm/pull/2208) solution can work. It avoids having paths with null type.

PR #2191, opened 2018-03-09
HHH-12290 Expand ordinal parameters to ordinal parameters in collection parameter expansion

gsmet:
Finally got a chance to take a look at this one today so here is my attempt to fix https://hibernate.atlassian.net/browse/HHH-12290 .

It's very similar to the new approach taken by @bdshadow in https://github.com/hibernate/hibernate-orm/pull/2158 but I think it's a bit less invasive.

We now replace ordinal parameters with ordinal parameters instead of using named parameters. It avoids gaps and should work pretty much OK.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
@@ -704,15 +704,11 @@ public class QueryTest extends BaseEntityManagerFunctionalTestCase {
 			em.close();
 		}
 	}
-	
-	/**
-	 * Collection parameters are internally rewritten to named parameters (one named parameter by
-	 * collection item) and HQL parser rejects mixed positional and named parameters queries.
-	 */
+
 	@Test
 	@TestForIssue(jiraKey = "HHH-12290")
 	public void testParameterCollectionAndPositional() {
-		final Item item = new Item( "Mouse", "Micro$oft mouse" );
+		final Item item = new Item( "Mouse", "Microsoft mouse" );
 		final Item item2 = new Item( "Computer", "Dell computer" );
 
 		EntityManager em = getOrCreateEntityManager();
@@ -724,10 +720,7 @@ public class QueryTest extends BaseEntityManagerFunctionalTestCase {
 			em.getTransaction().commit();
 
 			em.getTransaction().begin();
-			Query q = em.createQuery( "select item from Item item where item.name in ?1 and descr = ?2" );
-			//test hint in value and string
-			q.setHint( "org.hibernate.fetchSize", 10 );
-			q.setHint( "org.hibernate.fetchSize", "10" );
+			Query q = em.createQuery( "select item from Item item where item.name in ?1 and item.descr = ?2" );
 			List params = new ArrayList();
 			params.add( item.getName() );
 			params.add( item2.getName() );
@@ -747,15 +740,11 @@ public class QueryTest extends BaseEntityManagerFunctionalTestCase {
 			em.close();
 		}
 	}
-	
-	/**
-	 * Collection parameters are internally rewritten to named parameters (one named parameter by
-	 * collection item) and HQL parser rejects mixed positional and named parameters queries.
-	 */
+
 	@Test
 	@TestForIssue(jiraKey = "HHH-12290")
 	public void testParameterCollectionParenthesesAndPositional() {
-		final Item item = new Item( "Mouse", "Micro$oft mouse" );
+		final Item item = new Item( "Mouse", "Microsoft mouse" );
 		final Item item2 = new Item( "Computer", "Dell computer" );
 
 		EntityManager em = getOrCreateEntityManager();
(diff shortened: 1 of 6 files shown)


gsmet:
Rebased and removed the stream/lambdas in favor of 2 for loops.

gsmet:
@sebersole any comment on this one? You had a pretty strong opinion about it so I prefer to check the principles with you.

hibernate-core/src/test/java/org/hibernate/jpa/test/query/QueryTest.java
```diff
@@ -705,6 +705,113 @@ public void testJpaPositionalParameters() {
 		}
 	}
 
+	@Test
+	@TestForIssue(jiraKey = "HHH-12290")
+	public void testParameterCollectionAndPositional() {
+		final Item item = new Item( "Mouse", "Microsoft mouse" );
+		final Item item2 = new Item( "Computer", "Dell computer" );
+
+		EntityManager em = getOrCreateEntityManager();
```
sebersole: You could use `org.hibernate.testing.transaction.TransactionUtil2#inTransaction(org.hibernate.engine.spi.SessionFactoryImplementor, java.util.function.Consumer<org.hibernate.engine.spi.SessionImplementor>)` for this mundane transactional handling

bdshadow:
We had a long discussion with @gsmet about `StringHelper.replace`, but i still don't think that a new method `StringHelper.indexOfIdentifierWord` is needed. `StringHelper.replace` works fine

sebersole:
Applied upstream

gsmet:
@bdshadow once again, it's not about `StringHelper.replace()` but about getting https://github.com/hibernate/hibernate-orm/pull/2191/files#diff-6844ce9401311e9e7c7cef12ad3cf721L535 right.

sebersole:
@gsmet Initially I kind of agree with @bdshadow  here.  What specific case is this (`#indexOfIdentifierWord` + its use) addressing that `StringHelper#replace` does incorrectly?

gsmet:
Well, imagine you have: a.name in (?13) ..... and a.id in (?1).

When you're replacing `?1`, https://github.com/hibernate/hibernate-orm/pull/2191/files#diff-6844ce9401311e9e7c7cef12ad3cf721L535 will get the position of the `?1` part of `?13`, which is incorrect.

In the end `StringHelper.replace` compensates for it but I think we should better get this right. That's why I introduced this method.

sebersole:
````

		final String p1Replacement = StringHelper.replace(
				"... where x.y in (?1) and x.z in (?13)",
				"?1",
				":p1, :p2, :p3",
				true,
				true
		);
````

-> `... where x.y in (:p1, :p2, :p3) and x.z in (?13)`.

So dunno

gsmet:
Yeah once again, it's not about `StringHelper.replace`, it's about this specific call I pointed out: https://github.com/hibernate/hibernate-orm/pull/2191/files#diff-6844ce9401311e9e7c7cef12ad3cf721L535 .

bdshadow:
@gsmet ok, but this new method is unnecessary at all. I just meant that everything could be done with existing utils. No need in new one. It's just like: what's the point to write more code, when the less amount of it does the same.

gsmet:
@bdshadow which method would you use to get the right position of ?1 in: `in (?13) ..... and a.id in (?1)` at this line https://github.com/hibernate/hibernate-orm/pull/2191/files#diff-6844ce9401311e9e7c7cef12ad3cf721L535? Because, no, `indexOf()` doesn't work and obviously returns the wrong answer.

You keep talking about `StringHelper.replace()` but I haven't changed this method at all. `StringHelper.replace()` is not the point of this change.

bdshadow:
@gsmet i didn't need to get the right position of ?1. So no need in new `StringHelper.indexOfIdentifierWord`

gsmet:
OK, frankly, I don't see the point of this discussion. The code is obviously incorrect using `indexOf()` to get the position of the parameter. If you can't see that, I don't know what I can do. The fact that `StringHelper.replace()` compensates for it later does not have to do with the issue at hand. Personally, I don't like having code that is based on incorrect assumptions, especially because code lives and you might miss a bug later. If you don't care about having incorrect code and just care about the result, I see how you might not care about this issue.

Anyway, discussion closed for me, I spent more time discussing this than writing the patch. Do whatever you want.

bdshadow:
Blaming me like this is not a good way of discussion. 
I spent lots of time for this pr, trying different options. So i DO care about the result and about the issue.
You created your PR as a different approach to solve the problem, saying it's less invasive (see your first comment to pr). I can says that it isn't less invasive as more unnecessary code was created (in my opinion, yes) .
Also i didn't have any feedback about my pr from @sebersole. I hardly believe that PRs were compared at all. From my point of view it looks like this PR was just merged in 3 hours after you asked to review it, when my pr was hanging for a long time already and haven't been seen at all. If they were compared indeed, i'm very sorry for these words. 
Everything i want is some comments from @sebersole why this approach is better. And i'm totally ready to accept the decision if this approach is better of equal indeed. But not only if i'm a new, not yet trusted, contributor.

sebersole:
Dmitrii, TBH I have been very "head down" working on something else the
past 2 weeks.  I saw you 2 talking about various approaches, know that you
guys know each other and just assumed you'd all worked it out and that this
PR was the consolidation of both of your efforts.  So I apologize for that.

Guillaume, the point of the discussion is simply that there are methods
that *already* do what you are trying to do.  You may enjoy re-inventing
the wheel (heck, who of us as developers doesn't), but I assure you... the
people who come after and will be maintaining this code do in fact care
that you added a new method - extra maintenance work.  Is that reallynot
important to you?  Good to know

PR #2192, opened 2018-03-11
fixed "guide" name

62mkv:
Getting Started guide seems to be consistently using the (outdated?) term "Developer Guide", which is nowhere to be found in the "current" documentation, thus I propose this change. May be add a link to this guide as well ? which proper form should such a link have ?


vladmihalcea:
Applied upstream, thanks

PR #2193, opened 2018-03-11
HHH-12378 JDK 9 support: Remove javax.annotation.Generated import

reda-alaoui:
Fix https://hibernate.atlassian.net/browse/HHH-12378


Sanne:
merged. Thanks!

PR #2196, opened 2018-03-13
HHH-12380 - Stackoverflow when order_inserts=true

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12380


vladmihalcea:
@dreab8 I created a new PR for this issue which uses only local variables to fix this issue. The problem with the field variables is that they are not related to a particular `BatchIdentifier` we are currently checking and might stay set to true even for other `BatchIdentifier` variables.

PR #2197, opened 2018-03-13
HHH-11356 - Adjust the 2nd-Cache SPIs to better reflect supported uses

sebersole:
HHH-11356 - Adjust the 2nd-Cache SPIs to better reflect supported uses
HHH-12323 - Update Statistics API and SPI based on changes to 2nd level caching changes


sebersole:
Commit https://github.com/hibernate/hibernate-orm/pull/2197/commits/93cee7ef9b7b00c496ddad8e47e497aaab8ebd56 adds `hibernate-jcache` support for the new SPI as well as some general improvements.  

Next I'll work on getting `hibernate-ehcache` set up as a relocation.  @henri-tremblay what is the artifact GAV we should use for the relocation (Ehcache3 based provider)?  `org.ehcache:ehcache:3.0.0`?  Or should we just point them to `hibernate-jcache` + `org.ehcache:ehcache:3.0.0`?

henri-tremblay:
How can you point to two different places? What do you think is the work necessary to just migrate hibernate-ehcache?

sebersole:
https://github.com/hibernate/hibernate-orm/pull/2197/commits/4ebb70348d678f48d464d0408c65f8c46fcd353e#diff-9d7c1944c0518dcf56cdb2d04f395a81

A maven relocation cannot.  We point the relocation to hibernate-jcache.  In the artifact description we note the 2.

It would probably not be too bad creating an Ehcache 2 provider in hibernate-ehcache based on the new code.  The effort should be similar to the RegionFactory in hibernate-testing or the one in hibernate-jcache

sebersole:
Applied this upstream.  We can do any changes there if needed.

henri-tremblay:
Ok. Right now our thoughts are that relocation to `hibernate-jcache` is misleading. But might indeed be the best move right now.

Will you be willing to do the `hibernate-ehcache` changes until we take over? (and add Ehcache 3 in some way)?

sebersole:
Until there is a reason to, there is really no need for a specialized Ehcache 3 based RegionFactory.  IMO hibernate-jcache + ehcache3 are enough

I can add a VERY simplistic one for you if you tell me the calls I should make to:

1. Get an Ehcache "Cache"/"CacheManager"
2. Which methods to call on "Cache" to implement {{org.hibernate.cache.spi.support.StorageAccess}} and {{org.hibernate.cache.spi.support.DomainDataStorageAccess}}

Anything beyond that (cache provisioning, custom access strategies, etc) will be on y'all to do if you wish.  It's not "necessary" but there might be some cases you'd find it useful

PR #2198, opened 2018-03-14
HHH-10667 - Fix Envers allowing @IdClass mappings with entity primary keys.

Naros:
https://hibernate.atlassian.net/browse/HHH-10667


Naros:
> My only comment relates to the 'default' interface method on IdMapper. If you change that to a normal interface method and then just implement it in each class to call the prior method, your code changes are all compatible with java versions prior to java 1.8 which means this pull request could probably be cherry picked to older hibernate branches other than master.

While that would make cherry-pick simple, it is not uncommon for us to target our java 1.8 baseline and if it is deemed to be backported to earlier non java 1.8 releases, the cherry-pick would be adapted to be compatible.  In facts its not uncommon for us because features/fixes overlap with one another and to backport a feature or fix might involve the inclusion or adaptation to exclude another.

Naros:
Applied upstream.

PR #2199, opened 2018-03-14
HHH-12390 add hibernate-jipijapa

scottmarlow:
@Sanne how can we include running the JipiJapa unit tests in this pr?  

Do we have an existing HHH jira for this pull request?

Thanks,
Scott


gradle/libraries.gradle
```diff
@@ -12,7 +12,7 @@ ext {
     junitVersion = '4.12'
     h2Version = '1.4.196'
     bytemanVersion = '4.0.1' //Compatible with JDK10
-    infinispanVersion = '8.2.5.Final'
+    infinispanVersion = '9.2.0.Final'
```
Sanne: If you need the Infinispan upgrade, could you separate that in a specific commit and JIRA ?

hibernate-jipijapa/hibernate-jipijapa.gradle
```diff
@@ -0,0 +1,22 @@
...
+apply from: rootProject.file( 'gradle/published-java-module.gradle' )
+
+description = 'Integrate with WildFly JipiJapa '
+
+dependencies {
+    compile project( ':hibernate-core' )
+    compile( "org.wildfly:jipijapa-spi:12.0.0.Final" )
+    compile( "org.jboss:jboss-vfs:3.2.11.Final" )
+    // compile( "org.infinispan:infinispan-core:9.2.0.Final" )
+    compile( "org.infinispan:infinispan-hibernate-cache-commons:9.2.0.Final" )
```
Sanne: this should probably use the `infinispanVersion` property?

hibernate-jipijapa/src/test/bundles/overridenpar/META-INF/persistence.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Hibernate, Relational Persistence for Idiomatic Java
+  ~
+  ~ License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+  ~ See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+  -->
+
+<persistence xmlns="http://java.sun.com/xml/ns/persistence"
+             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
Sanne: Do we need all these additional integrations tests? I wonder if most aren't redundant with the ORM standard tests.

In other words: are they all specific to JipiJapa?

Sanne:
> Do we have an existing HHH jira for this pull request?

I think you mentioned on HipChat that you had created one

gradle/libraries.gradle
scottmarlow: The WildFly master branch is currently on Infinispan 9.2.0.Final, from what you said before about Infinispan needing to catch up with the ORM 5.3 cache changes, the Infinispan version will likely need to be bumped again.  Should we upgrade to Infinispan 9.2.0.Final for now?

hibernate-jipijapa/hibernate-jipijapa.gradle
scottmarlow: sure, it was easier for me to just hard code the versions in my first .gradle file, I'll fix this. :)

scottmarlow:
I'll create a HHH jira.

hibernate-jipijapa/src/test/bundles/overridenpar/META-INF/persistence.xml
scottmarlow: Hi Sanne, yes, I think we need the tests to ensure that basic deployment scanning works.

gradle/libraries.gradle
scottmarlow: I didn't remove the 9.2.0.Final upgrade yet, guess we can do that next but we will be out of sync then with WildFly master.

scottmarlow:
The hibernate-jipijapa tests do not run during "gradlew clean publishToMavenLocal test", do we need more changes in the hibernate-jipijapa/hibernate-jipijapa.gradle for that to happen?

hibernate-jipijapa/src/test/bundles/overridenpar/META-INF/persistence.xml
sebersole: We need *so many* just to test scanning/discovery?

sebersole:
I'm not sure about for you @scottmarlow but for me the tests "wont run" because they fail to compile - missing shrinkwrap deps

sebersole:
This fails badly on top of the cache SPI changes because it bundles its own Infinispan-based RegionFactory.  We need to decide what to do here...  The temporary solution, as I mentioned the earlier, is to temporarily use hibernate-jcache+infinispan.  Should I just drop the custom RegionFactory impls from hibernate-jipijapa?

Sanne:
Using `hibernate-jcache+infinispan` might be ok for standalone users but isn't good enough for WildFly users as WildFly manages those caches - since JipiJapa is only used by WildFly I don't think that's an option.

Regarding the custom `RegionFactory`, we were discussing solutions about that with @pferraro on the clustering mailing list.

He just made a great proposal, but it will need coordination with the other repositories integrating with this so I'm waiting to hear what @scottmarlow thinks of it.

Regarding this PR: we either wait for that to be clarified, or we drop the cache integration temporarily to re-introduce it shortly after.

sebersole:
I have no idea how WF would (or not) affect "Infinispan" as a JCache CachingProvider, so hard to say.

But you guys seem very tied to only using an Infinispan "native" RegionFactoryImpl... so given that and the choice between waiting and just removing all of this mess that exists now... I'll chose removing it...

Sanne:
It's not about it being "native". Maybe WF should use Infinispan over JCache, but it needs to manage the underlying caches.

I don't think it matters for WF if we use them directly or "over" JCache.

sebersole:
Hence why I said "I have no idea how WF would (or not) affect "Infinispan" as a JCache CachingProvider"

I mean based on your and Scott's reactions the choice seems to be to (1) wait for Infinispan to implement the cache changes or (2) drop the custom jipijapa handling (perhaps just temporarily).

Given those, I chose (2).

Sanne:
Ok, sounds fine to me.

We'll just have to treat the deferred Infinispan support as a blocker for WildFly integration.

sebersole:
Or use hibernate-jcache + infinispan, but meh you guys know best :P

hibernate-jipijapa/src/test/bundles/overridenpar/META-INF/persistence.xml
scottmarlow: I forget exactly why we needed so many.  I think the initial idea was to go from zero tests in hibernate-jipijapa, to many of the same tests that are in Hibernate ORM.  Having a smaller number of tests sounds fine to me.

gradle/libraries.gradle
scottmarlow: @pferraro can answer better than me, I'm not really sure what JCache configuration will look like in WildFly exactly.  is wildfly-jcache configuration generic for any of the JCache implementations, like EHCache, or is it specific to Infinispan?  I assume it would be specific to Infinispan but not really sure.

sebersole:
For JCache its not just what the CacheManager looks like... generally, I
think, we'd want to use CachingProvider instead IIUC.  WildFly would be or
provide the CachingProvider, again IIUC

sebersole:
Just to follow up on that... Hibernate *can* use a specific `CacheManager`
if one is passed to it.  So that's an integration idea too.  Jipajapa would
either:

    * register itself as the JCache `CachingProvider` to the
{{hibernate-jcache}} `RegionFactory` implementation
    * resolve the CacheManager to use from Infinispan (though I'd have a
hook to allow that for other cache impls) and provide that to Hibernate

sebersole:
See #2221

gradle/libraries.gradle
pferraro: @scottmarlow For starters, WF would need to provide its own CachingProvider, whose caches are backed by configurations created by the Infinispan subsystem.  Additionally, hibernate-jcache would need to be modified to load its CachingProvider from a specific classloader (e.g. the classloader of the deployment or the classloader of the JCacheRegionFactory class) rather than the thread context classloader (the default behavior of Caching.getCachingProvider(String)).
This is certainly a worthwhile goal - I had proposed a WF JCache provider about 3 years ago, and can resurrect/update that code for use with hibernate-jcache.
However, while this RegionFactory implementation is likely fine for single node, or read-only and non-strict-read-write caches, anyone intending to use read-write caches in a WF cluster will still want to use Infinispan's RegionFactory implementation.

sebersole:
Superseded by my PR building on this one (which  has already been integrated upstream)

PR #2200, opened 2018-03-14
HHH-12389 - Remove usage of javax.script.ScriptEngine in MixedAccessTest

barreiro:


vladmihalcea:
@barreiro Hi Luis. I tried to integrate this PR but the test fails in the `After` check as the changes done during test don't propagate to the DB.

barreiro:
@vladmihalcea that's why this test is annotated with `@Ignore`. The bytecode enhancer is designed for field access, and in particular having property access prevents dirty tracking. Therefore, although changes are made, the entity is still 'clean' and changes are not persisted.

vladmihalcea:
@barreiro On Maven, a test annotated with `@Ignore` test would not un at all. But here, with Gradle, and our runners just ignore it, at least at IDE-level.

I will have to run it from the command-line too. But if the logic is intended, we can also make the test assume that nothing has changed. Currently, we check the opposite.

vladmihalcea:
Also, we have some problems with the `MixedAccessTest` generating OOM errors from time to time.

```
java.lang.OutOfMemoryError: GC overhead limit exceeded
Dumping heap to D:\Vlad\Work\GitHub\hibernate-orm\hibernate-core\target\OOM-dump.hprof ...
<============-> 94% EXECUTING [11m 22s]
> :hibernate-core:test > 2014 tests completed, 77 skipped
> :hibernate-core:test > Executing test org.hibernate.test...enhancement.access.MixedAccessTest
```

sebersole:
So what's the outcome here?

barreiro:
@vladmihalcea The main goal for this PR is precisely to address the OOM errors.
As for the test, depending on the enhancement context they either pass or are expected to fail (it's not technically possible to make it pass). It would be desired to have the same behavior, but it's just not possible to do so.
To address your concerns, the test that now runs both scenarios could be either split, or just run the passing one.

vladmihalcea:
@barreiro I created this screen capture to show you what happens when I run `gradlew clean test`. The tests have been running for 20 minutes and the suite is stuck on the `MixedAcessTest`.

Alt Text

Does it work for you when you run all tests?

vladmihalcea:
Applied upstream, thanks

PR #2202, opened 2018-03-15
HHH-12392 - Caching SchemaResolver delegate with multiple data sources

dreab8:
https://hibernate.atlassian.net/browse/HHH-12392

`determineAppropriateResolverDelegate(Connection)` will resolve the appropriate `SchemaNameResolver`  every time without checking `if ( delegate == null )` so no caching for the  `DefaultSchemaNameResolver#delegate` field.

Another change is related with the exception that may occur trying to resolve the Java 7 resolver, in such a case the `SchemaNameResolverFallbackDelegate` is set instead of leaving the `DefaultSchemaNameResolver#delegate` field null. 

```
catch (Exception ignore) {
			log.debugf(
					"Unable to use Java 1.7 Connection#getSchema : An error occurred trying to resolve the connection default schema resolver: "
							+ ignore.getMessage() );
			delegate = SchemaNameResolverFallbackDelegate.INSTANCE;
		}
```


dreab8:
applied upstream

PR #2203, opened 2018-03-15
>>> DO NOT MERGE, It's a WIP <<<   HHH-11263 - BootstapContext + HHH-11264 Event Listeners merge 

dreab8:
@sebersole the part related with the `org.hibernate.type.spi.TypeConfiguration` is not finished [Intoduce TypeConfiguration](https://github.com/hibernate/hibernate-orm/pull/2203/commits/8ff11f167ba26e9d1c007d20ec1794de1c637f86) , anyway the code compile and all the test pass.


hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataImpl.java
```diff
@@ -84,7 +85,7 @@
 	public MetadataImpl(
 			UUID uuid,
 			MetadataBuildingOptions metadataBuildingOptions,
-			TypeResolver typeResolver,
+			TypeConfiguration typeConfiguration,
```
sebersole: Is there a reason we can't just get the `TypeConfiguration` from the `BoootstrapContext` here?

hibernate-core/src/main/java/org/hibernate/boot/model/process/spi/MetadataBuildingProcess.java
```diff
@@ -311,45 +314,34 @@ public void finishUp() {
 		return metadataCollector.buildMetadataInstance( rootMetadataBuildingContext );
 	}
 
-//	private static JandexInitManager buildJandexInitializer(
```
sebersole: Can we leave this in place?  Possibly with a `todo (7.0)` comment?

sebersole:
You may already have done it (its hard to see looking on line) but we should make sure that `TypeResolver` itself and all of its uses are depprecated

hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataImpl.java
dreab8: just an oversight. Thanks

hibernate-core/src/main/java/org/hibernate/boot/model/process/spi/MetadataBuildingProcess.java
dreab8: Added back.

dreab8:
@sebersole Thanks for the review and the comments. 
About TypeResolver deprecation,  I have to go through and deprecate all of the uses. 
Should I add all the TypeResolver methods to the TypeConfiguration and then deprecate them in favour of the TypeConfiguration ones?

sebersole:
> Should I add all the TypeResolver methods to the TypeConfiguration and then deprecate them in favour of the TypeConfiguration ones?

No.  I think you did it correctly - expose `TypeConfiguration#getTypeResolver` (deprecated).  Just wanted to make sure we got TypeResolver itself deprecated and any accesses of it (like immediatelly deprecating this `TypeConfiguration#getTypeResolver` method).

hibernate-core/src/main/java/org/hibernate/boot/spi/BootstrapContext.java
```diff
@@ -0,0 +1,169 @@
...
+	/**
+	 * Releases the "bootstrap only" resources held by this BootstrapContext.
+	 * <p/>
+	 * Only one call to this method is supported, after we have completed the process of
+	 * building the (non-inflight) Metadata impl.  We may want to delay this until we
+	 * get into SF building.  Not sure yet.
+	 *
+	 * @todo verify this ^^
+	 */
+	void release();
```
sebersole: I'm not seeing where this gets called....

sebersole: In 6.0 this happens in `InFlightMetadataCollectorImpl#buildMetadataInstance`

hibernate-core/src/main/java/org/hibernate/internal/SessionFactoryImpl.java
```diff
@@ -247,8 +253,9 @@ public SessionFactoryImpl(final MetadataImplementor metadata, SessionFactoryOpti
 		for ( SessionFactoryObserver sessionFactoryObserver : options.getSessionFactoryObservers() ) {
 			this.observer.addObserver( sessionFactoryObserver );
 		}
-
-		this.typeResolver = metadata.getTypeResolver().scope( this );
+		TypeConfiguration typeConfiguration = metadata.getTypeConfiguration();
```
sebersole: IMO we should keep the TypeConfiguration either here or (more likely) on MetamodelImpl and access things from there.  E.g., here I would just drop the `typeResolver` field - we can always just get it from TypeConfiguration

dreab8: do you mean to add a `getTypeConfiguration() ` method to `MetamodelImplementor`?

sebersole:
Yes.  On 6 we add it to Metamodel, but I think MetamodelImplementor is
probably more appropriate

sebersole:
Actually... wrt MetamodelImplementor#getTypeConfiguration... I think what I'd really like to do instead is to essentially replace `org.hibernate.Metamodel` / `org.hibernate.metamodel.spi.MetamodelImplementor` with TypeConfiguration.. that is the direction I was headed in 6.0.  To that end, I would:

1. Deprecate both `org.hibernate.Metamodel` and `org.hibernate.metamodel.spi.MetamodelImplementor`  pointing to `TypeConfiguration` as the replacement
2. Implement as many of the `org.hibernate.Metamodel` and `org.hibernate.metamodel.spi.MetamodelImplementor`  methods as possible by delegating to `TypeConfiguration

dreab8:
@sebersole  [this commit](https://github.com/hibernate/hibernate-orm/pull/2203/commits/f9835865e6e94ea5f589e86ea0600b6aa611005d) applies the changes related to yours first comments
Now I'm going to address your last comment

sebersole:
Awesome @dreab8 !

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/2225

PR #2204, opened 2018-03-15
HHH-12338 - Incorrect metamodel for basic collections

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12338

### DIFF
--- a/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/annotation/MetaAttributeGenerationVisitor.java
+++ b/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/annotation/MetaAttributeGenerationVisitor.java
@@ -138,14 +138,20 @@ public class MetaAttributeGenerationVisitor extends SimpleTypeVisitor6<Annotatio
 				accessTypeInfo.setDefaultAccessType( entity.getEntityAccessTypeInfo().getAccessType() );
 			}
 		}
-		if ( collection.equals( "javax.persistence.metamodel.MapAttribute" ) ) {
-			return createAnnotationMetaAttributeForMap( declaredType, element, collection, targetEntity );
-		}
-		else {
-			return new AnnotationMetaCollection(
-					entity, element, collection, getElementType( declaredType, targetEntity )
+		if ( TypeUtils.containsAnnotation( element, Constants.CONVERT ) ||
+				TypeUtils.containsAnnotation( element, Constants.HIBERNATE_TYPE )	) {
+			return new AnnotationMetaSingleAttribute(
+					entity,
+					element,
+					TypeUtils.toTypeString( declaredType )
 			);
 		}
+		if ( collection.equals( Constants.MAP_ATTRIBUTE ) ) {
+			return createAnnotationMetaAttributeForMap( declaredType, element, collection, targetEntity );
+		}
+		return new AnnotationMetaCollection(
+				entity, element, collection, getElementType( declaredType, targetEntity )
+		);
 	}
 
 	@Override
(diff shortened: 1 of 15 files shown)


helloztt:
I just decided to handle the ``@Converter`` case at first, but it's more difficult to handle the ``@Converter(autoApply = true)`` case while there are no annotation on fields

tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/annotation/MetaAttributeGenerationVisitor.java
```diff
@@ -138,14 +138,20 @@ private AnnotationMetaAttribute createMetaCollectionAttribute(DeclaredType decla
...
 		}
-		else {
-			return new AnnotationMetaCollection(
-					entity, element, collection, getElementType( declaredType, targetEntity )
+		if ( TypeUtils.containsAnnotation( element, Constants.CONVERT ) ||
+				TypeUtils.containsAnnotation( element, Constants.HIBERNATE_TYPE )	) {
+			return new AnnotationMetaSingleAttribute(
+					entity,
+					element,
+					TypeUtils.toTypeString( declaredType )
```
Naros: The only downside to using `TypeUtils.toTypeString( declaredType )` is that the static definition in the static model class will be generated with fully qualified class names rather than using an import clause, effectively looking like this:
````
public static volatile SingularAttribute<SomeEntity, List<java.lang.String>> someVariable;
````
I think what we can do is address this by adjusting `ImportContextImpl#importType` as follows:
````
String additionalTypePart = null;
if ( fqcn.indexOf( '<' ) >= 0 ) {
	additionalTypePart = result.substring( fqcn.indexOf( '<' ) );
	// start new functionality to recursively handle importing generic types
	String genericTypes = additionalTypePart.substring( 1, additionalTypePart.length() - 1 );
	additionalTypePart = "<" + importType( genericTypes ) + ">";
	// end new functionality
	result = result.substring( 0, fqcn.indexOf( '<' ) );
	fqcn = result;
}
else if ( fqcn.indexOf( '[' ) >= 0 ) {
	additionalTypePart = result.substring( fqcn.indexOf( '[' ) );
	result = result.substring( 0, fqcn.indexOf( '[' ) );
	fqcn = result;
}
````
We'd need to check if this does not introduce any regressions, particularly around use cases where the imported types resulted in type clashes.  I believe the subsequent part of this method handle this, but would just want to verify.  At least to me, this ends up looking like cleaner class code rather than inline fully qualified class names.

File: tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/annotation/MetaAttributeGenerationVisitor.java [hunk 2]
```diff
@@ -138,14 +138,20 @@ private AnnotationMetaAttribute createMetaCollectionAttribute(DeclaredType decla
 				accessTypeInfo.setDefaultAccessType( entity.getEntityAccessTypeInfo().getAccessType() );
 			}
 		}
-		if ( collection.equals( "javax.persistence.metamodel.MapAttribute" ) ) {
+		if ( collection.equals( Constants.MAP_ATTRIBUTE ) ) {
```
Naros: In the _else_ branch, we handle the basic collection situation when an `AttributeConverter` is applied; however they can also be applied to a `Map` type collection.  A simple example might be something like this:
````
@Convert(converter = KeyValuePairAsJsonSimpleString.class)
private Map<String, Object> keyValuePairs;
````
I'd strongly suggest if we're going to address non-map collections, we might as well address it for these too just to be consistent in the generation of the static metamodel classes.

vladmihalcea: Thanks @Naros. I'll have to investigate that too.

tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/annotation/MetaAttributeGenerationVisitor.java
vladmihalcea: If we do this change, we get the following error:

> error: ';' expected
> import java.lang.String,java.lang.String;
>                        ^

It looks like this logic is further used for imports.

PR #2205, opened 2018-03-16
HHH-12391

dreab8:
@sebersole [this](https://github.com/hibernate/hibernate-orm/pull/2205/commits/98c577b8704a37d2a9848e81f9a108dac1640b40) reverts your last commit, it was causing some test failures.

[here](https://github.com/hibernate/hibernate-orm/pull/2205/commits/76a42beac569c613bf5fd0ea51e4f880039121f6) a test that reproduces the issue.

finally [here](https://github.com/hibernate/hibernate-orm/pull/2205/commits/2e09d27692e4f783faf1307d9d2c4c5ae174886b) a possible solution

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/resource/transaction/backend/jdbc/internal/JdbcResourceLocalTransactionCoordinatorImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/resource/transaction/backend/jdbc/internal/JdbcResourceLocalTransactionCoordinatorImpl.java
@@ -288,11 +288,15 @@ public class JdbcResourceLocalTransactionCoordinatorImpl implements TransactionC
 
 		@Override
 		public void rollback() {
-			TransactionStatus status = jdbcResourceTransaction.getStatus();
-			if ( ( rollbackOnly && status != TransactionStatus.NOT_ACTIVE ) || status == TransactionStatus.ACTIVE ) {
+			try {
+				TransactionStatus status = jdbcResourceTransaction.getStatus();
+				if ( ( rollbackOnly && status != TransactionStatus.NOT_ACTIVE ) || status == TransactionStatus.ACTIVE ) {
+					jdbcResourceTransaction.rollback();
+					JdbcResourceLocalTransactionCoordinatorImpl.this.afterCompletionCallback( false );
+				}
+			}
+			finally {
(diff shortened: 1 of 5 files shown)


sebersole:
I don't follow.  How was my test causing other test failures?

dreab8:
@sebersole I haven't investigated but seems the changes you made cause
[Failing tests](http://ci.hibernate.org/job/hibernate-orm-master-h2-main/1141/)

If run individually they pass but running the entire test suite they fail

sebersole:
That's so strange - I cannot fathom how that change would cause those failures.  But clearly they do (its the only change in that build) and the test is not needed anyway.

scottmarlow:
I'm not sure if it helps with this failure but should the rollback() + commit() methods be changed to always clear the rollbackOnly flag?  Currently, they look like they are not clearing the rollbackOnly flag.

dreab8:
@scottmarlow the rollback() method sets the rollbackOnly flag to false;
anyway the [commit](https://github.com/hibernate/hibernate-orm/pull/2205/commits/2e09d27692e4f783faf1307d9d2c4c5ae174886b) seems fixing the problem

sebersole:
@scottmarlow have you tested whether the changes here fix your failure?

scottmarlow:
@sebersole I haven't, I should be able to try that.

scottmarlow:
@dreab8 I meant that rollback() may not clear rollbackOnly, if for example we enter the method and don't reach the rollbackOnly=false.  Perhaps the entire method body should be in a try/finally {rollbackOnly=false;}

Commit() does not set rollbackOnly=false, that probably also needs a try {} finally {rollbackOnly=false} around the method body, to ensure that we do not leave rollbackOnly == true.  We might call rollback() or not.

dreab8:
@scottmarlow for the `commit() `method I think there is not need, when `rollbackOnly=true `the `rollback()` method is called.
For `rollback()` you may be right, probably `rollbackOnly=false` should be the first thing the `rollback()` method does.

dreab8:
@scottmarlow forget my last comment on the `rollback()` change, your idea of try/finally for setting the `rollbackOnly=false` is correct.
Running test and then I'll push the changes.

dreab8:
@sebersole thanks, I'm going to merge it.

PR #2208, opened 2018-03-16
HHH-12375

dreab8:
https://hibernate.atlassian.net/browse/HHH-12375

@beikov what do you think?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractPropertyMapping.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractPropertyMapping.java
@@ -127,6 +127,18 @@ public abstract class AbstractPropertyMapping implements PropertyMapping {
 		}
 	}
 
+	private void logIncompatibleRegistration(String path, Type existingType, Type type) {
+		if ( LOG.isTraceEnabled() ) {
+			LOG.tracev(
+					"Skipped adding same named type incompatible property to base type [{0}] for property [{1}], existing type = [{2}], incoming type = [{3}]",
+					getEntityName(),
+					path,
+					existingType,
+					type
+			);
+		}
+	}
+
 	/**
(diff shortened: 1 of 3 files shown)


beikov:
Looks good and cleaner. Removing the path will have the same effect as putting null into the map so this should be fine.

dreab8:
@sebersole what do u think, may it work or do you see problems with this approach?

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractPropertyMapping.java
```diff
@@ -127,6 +131,18 @@ private void logDuplicateRegistration(String path, Type existingType, Type type)
 		}
 	}
 
+	private void logIncompatibleRegistration(String path, Type existingType, Type type) {
+		if ( LOG.isTraceEnabled() ) {
+			LOG.tracev(
+					"Skipped adding same named type incompatible property to base type [{0}] for property [{1}], existing type = [{2}], incoming type = [{3}]",
```
sebersole: Can we say something like this, instead?

````
"Skipped adding attribute [{1}] to base-type [{0}] as more than one sub-type defined the attribute using incompatible types (strictly speaking the attributes are not inherited); existing type = [{2}], incoming type = [{3}]"
````

dreab8: Thanks @sebersole,
I'll change it.

cnsgithub:
@dreab8 I just tested your modifications to the original bugfix and can confirm that things are still working. Now it's time to release 5.2.16. :wink:

dreab8:
Thanks @cnsgithub,

I'm going to do the release tomorrow.

dreab8:
applied upstream
Thanks @sebersole , @beikov @cnsgithub

PR #2210, opened 2018-03-17
HHH-3930 Test and fix for unnecessary secondary query when fetching inverse one-to-one

beikov:
https://hibernate.atlassian.net/browse/HHH-3930

The fix introduces the property name in the ManyToOneType. If the type is a logical one-to-one, I use the property name during the resolving of an instance to add an EntityUniqueEntry to the persistence context.


vladmihalcea:
Applied upstream, thanks

PR #2213, opened 2018-03-21
HHH-12413 Don't clear parameterListBindingMap to allow subsequent callers of expandListValuedParameters to receive the expanded query string

beikov:
https://hibernate.atlassian.net/browse/HHH-12413


gsmet:
Hi @beikov ,

The commit message refers to HHH-12143 whereas it should be HHH-12413.

gsmet:
Is it possible to add a test so that we don't break this behavior later on?

beikov:
Yeah of course, I just wanted to create the PR to see if the tests still run through.

vladmihalcea:
Applied upstream, thanks

PR #2215, opened 2018-03-22
HHH-12419 - Fix incorrect batch inserts example

quaff:


vladmihalcea:
The fix only requires moving the check prior to the persist.

PR #2216, opened 2018-03-22
HHH-12387 - Immutable entities can be updated via bulk update queries

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12387

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
@@ -17,6 +17,7 @@ import java.sql.Clob;
 import java.sql.Connection;
 import java.sql.NClob;
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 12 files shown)


hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
```diff
@@ -1518,6 +1521,25 @@ public int executeUpdate(String query, QueryParameters queryParameters) throws H
...
+	private void preventImmutableEntityUpdate(HQLQueryPlan plan) {
+		if ( plan.isUpdate() ) {
+			for ( EntityPersister entityPersister : getSessionFactory().getMetamodel().entityPersisters().values() ) {
+				if ( !entityPersister.isMutable() ) {
+					List<Serializable> entityQuerySpaces = new ArrayList<>(
+							Arrays.asList( entityPersister.getQuerySpaces() )
+					);
+					entityQuerySpaces.retainAll( plan.getQuerySpaces() );
+
+					if ( !entityQuerySpaces.isEmpty() ) {
```
sebersole: See my comment on the Jira about log versus exception.  I's like to have a consensus on that.  Other than the impact on this specific line, I think this looks good

vladmihalcea:
@sebersole I updated the PR according to your suggestions. By default, we are going to issue a Warning, but the user can also customize the behavior and trigger an Exception instead.

PR #2218, opened 2018-03-23
HHH-12226 : ObjectNotFoundException thrown when @NotFound(action = NotFoundAction.IGNORE) used with enhancement

gbadner:
https://hibernate.atlassian.net/browse/HHH-12226

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/EntityType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/EntityType.java
@@ -681,7 +681,7 @@ public abstract class EntityType extends AbstractType implements AssociationType
 				getAssociatedEntityName(),
 				id,
 				eager,
-				isNullable() && !isProxyUnwrapEnabled
+				isNullable()
 		);
 
 		if ( proxyOrEntity instanceof HibernateProxy ) {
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/type/EntityType.java
```diff
@@ -681,7 +681,7 @@ protected final Object resolveIdentifier(Serializable id, SharedSessionContractI
 				getAssociatedEntityName(),
 				id,
 				eager,
-				isNullable() && !isProxyUnwrapEnabled
+				isNullable()
 		);
```
gbadner: Before the change above, ```LoadType``` would end up being ```INTERNAL_LOAD_LAZY``` when enhancement is used with ```@LazyToOne(value = LazyToOneOption.NO_PROXY)```.  See [```SessionImpl#internalLoad```](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java#L1109-L1116).

```INTERNAL_LOAD_LAZY``` does not allow ```null``. When Hibernate tries to load the entity, it is not found and ```ObjectNotFoundException``` is thrown.

After the change, ```LoadType``` will be ```INTERNAL_LOAD_NULLABLE```, which properly returns ```null```.

gbadner: @sebersole, is proxy unwrapping still supported w/o bytecode enhancement?

sebersole: Sounds reasonable

sebersole: proxy unwrapping without bytecode enhancement would just be eager loading.  so in that sense yes and no

gbadner:
Merged.

PR #2219, opened 2018-03-23
HHH-12426: SAP HANA spatial dialect should support all SAP HANA spatial functions

breglerj:
https://hibernate.atlassian.net/browse/HHH-12426


vladmihalcea:
Applied upstream, thanks

PR #2220, opened 2018-03-23
HHH-11769: mariadb 10.0 - 10.3 dialects

bdshadow:
I hope that i haven't missed any features that appeared in these version.

Also i have a question regarding json functions in 10.2: https://mariadb.com/kb/en/library/json-functions/. Lot's of them return and work with arrays. How should i define these functions correctly.  There's a nice article by @vladmihalcea https://vladmihalcea.com/the-hibernate-types-open-source-project-is-born/, which describes how user should work with it. But isn't it done in hibernate yet? Should we do it in this commit or in the other? And what's better to do with these array\json functions in 10.2?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Database.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Database.java
@@ -257,7 +257,7 @@ public enum Database {
 	MARIADB {
 		@Override
 		public Class<? extends Dialect> latestDialect() {
-			return MariaDB53Dialect.class;
+			return MariaDB103Dialect.class;
 		}
 
 		@Override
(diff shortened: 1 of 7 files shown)


sebersole:
Awesome!  Thanks for that @bdshadow

sebersole:
WRT support for JSON and other JDBC types... IIRC we decided to wait for supporting this until 6.0.  

@vladmihalcea you remember?

Other than that, +1 from me

sebersole:
And I'll let others answer wrt how to best handle JSON type.  I've never worked with it

vladmihalcea:
@bdshadow Adding ARRAY and JSON types is not very difficult as demonstrated by this project I've contributing too:

https://github.com/vladmihalcea/hibernate-types

At some point, we can integrate them directly into Hibernate Core. By then, this project will be thoroughly tested and we can make sure that we get them right from the very beginning.

bdshadow:
@vladmihalcea ok, i see. So should this PR wait till it is implemented and then rebase?
Or is it ok to merge it and then change the MariaDB array functions when json and array function are ready? Should then other functions be implemented additionally to `json_valid`?

vladmihalcea:
Applied upstream, thanks

PR #2221, opened 2018-03-23
Hhh 12390

sebersole:
https://hibernate.atlassian.net/browse/HHH-12390

Update removing Infinispan caching support from hibernate-jipijapa (for now) - builds

Also did some code clean up and checkstyle fixing


hibernate-jipijapa/src/main/java/org/jboss/as/jpa/hibernate5/HibernateExtendedBeanManager.java
```diff
@@ -11,7 +11,7 @@
 
 import javax.enterprise.inject.spi.BeanManager;
 
-import org.hibernate.jpa.event.spi.jpa.ExtendedBeanManager;
```
sebersole: Specifically wanted to ask about this one @scottmarlow

sebersole:
@scottmarlow - any thoughts on this?

scottmarlow:
Sorry, I'll finish reviewing it Monday morning.

I did try running the WF JPA unit tests and saw a few issues, not sure if any of these are caused by your changes as I haven't run these tests for a while against 5.3.  I was more curious if the CDI integration test worked (org.jboss.as.test.integration.jpa.entitylistener.EntityListenersTestCase), which it did.

1. org.jboss.as.test.integration.jpa.cfgfile.CfgFileTestCase test with a simple hibernate.cfg.xml, failed with "java.util.ServiceConfigurationError: org.hibernate.integrator.spi.Integrator: Provider org.hibernate.envers.boot.internal.EnversIntegrator not a subtype" error.  

2.  Not surprised to see serveral NoClassDefFoundError: Failed to link org/jboss/as/jpa/hibernate5/infinispan/SharedInfinispanRegionFactory, which makes sense since that is not expected to be present yet.

3.  Various Envers tests like org.jboss.as.test.integration.jpa.hibernate.envers.AuditJoinTableoverOnetoManyJoinColumnTest failed with a ClassNotFoundException: org.hibernate.envers.AuditReaderFactory.  

4.  Test org.jboss.as.test.integration.jpa.hibernate.sessionfactorytest.SessionFactoryTestCase fails while trying to persist an entity, org.hibernate.internal.SessionImpl.checkTransactionNeeded seems to be throwing a "javax.persistence.TransactionRequiredException: no transaction is in progress", not sure why.

5. Hibernate Search test org.jboss.as.test.integration.hibernate.search.HibernateSearchJPATestCase failed with a "java.util.ServiceConfigurationError: org.hibernate.integrator.spi.Integrator: Provider org.hibernate.search.hcore.impl.HibernateSearchIntegrator not a subtype" error.

sebersole:
(1) This is strange in that nothing wrt `Integrator` changed.
(2) true true
(3) Not sure - I see that that in code base.  Is this test including hibernate-envers?  @Naros any ideas?
(4) Well, is there a transaction?
(5) Again strange - no `Integrator` changes

Curious.. did these fail with your PR?

scottmarlow:
I'm running again to see.  It looks like the envers error might have something to do with evers being a separate optional module with bidirectional dependencies.  Perhaps we need to remove the "optional" flag, which might not make sense for bidirectional module dependencies.

sebersole:
We should never ever ever ever (did I say ever?) use optional dependencies.  When I see that I immediately know something is very wrong.

scottmarlow:
Hmm, I think we tried making envers optional years ago and it didn't work correctly then either, we should remove the optional flag from https://github.com/hibernate/hibernate-orm/blob/master/hibernate-orm-modules/module-templates/hibernate-core.xml#L14.

Yes, we get the same failures with my PR.  For future reference https://paste.fedoraproject.org/paste/h2yR2zR3d7-cc2XUc9Ukdw shows WF JPA failures with my PR.  https://paste.fedoraproject.org/paste/xviEmpLxpQVzwGJc40p6Yw is from running against this PR.

sebersole:
I honestly have no idea why it is marked optional as I really do not have any experience with JBoss modules.  But in a general way (from my Maven, Gradle, OSGi, etc) experience - optional dependencies are (almost) always indicative of a bad dependency design/

hibernate-jipijapa/src/main/java/org/jboss/as/jpa/hibernate5/HibernateExtendedBeanManager.java
scottmarlow: Looks good to me.  Seeing the org.jboss.as.test.integration.jpa.entitylistener.EntityListenersTestCase test pass, gives me more confidence about this change.

sebersole:
@scottmarlow what should we do, do you think, to get some of these broader tests into hibernate-jipijapa?  If this thing is in such a bad state of affairs it would be good to know that from the tests ;)

Sanne:
@scottmarlow especially regarding your point "5": there should be no Hibernate Search at all around, so that suggests JipiJapa is adding some modules on the classpath which we shouldn't have - especially if you add the `hibernate-search:main` modules from WildFly, that's the one designed to work with the old version of Hibernate ORM so we're ending up having both Hibernate ORM 5.1 and 5.3 on the classpath.

It probably means that the `org.hibernate.integrator.spi.Integrator` it was linked to is the one from 5.1; so the error message is telling you that it's not of the right type of `Integrator`.

Sanne:
I'll check it out locally and see if I can understand how to disable Search. This need being a great example of why we need such details controlled by the ORM project ;)

Sanne:
@scottmarlow yes looks like we saying the same. No I don't think it's a problem in general, we just need to make sure to be able to handle such cases. Specifically, going forward JipiJapa should not decide to automatically inject an incompatible Hibernate Search version on the classpath, as users will have no idea what's going on. We'll deal with such details soon.

sebersole:
This work in particular has been integrated upstream, so I am going to close this.  The Jira is still open as IMO we still have some work to do here.  

@scottmarlow Are you going to move these "real" tests to hibernate-jipijapa?  

@scottmarlow Any luck remembering the details you mentioned about `JtaPlatformProviderImpl` not working.  Again, based on the tests we have in hibernate-jipijapa this code is all good.

@Sanne what should we do then to avoid the Integrator problems?  Just remove search from the module?

scottmarlow:
I updated the hibernate-search tests persistence.xml to specify the Hibernate ORM 5.1 module and that resolved the Search issue (for now).  For envers, we should try removing the optional setting.  For adding more "real" tests, great idea!  I'll give that a try soon.

scottmarlow:
For the details from years back, about JtaPlatformProviderImpl, the old [Jipijapa source mentions a few details at](https://github.com/jipijapa/jipijapa/blob/master/hibernate4_3/src/main/java/org/jboss/as/jpa/hibernate4/HibernatePersistenceProviderAdaptor.java#L59) about there being a memory leak and a [description of what the leak was is here.](http://lists.jboss.org/pipermail/hibernate-dev/2013-July/010140.html).

PR #2222, opened 2018-03-26
Fix failing tests on SAP HANA

breglerj:
- Account for identifier quoting on HANA in
  org.hibernate.test.annotations.namingstrategy.charset.Iso88591CharsetNamingStrategyTest
  and org.hibernate.test.annotations.namingstrategy.charset.Utf8CharsetNamingStrategyTest


vladmihalcea:
Applied upstream, thanks

PR #2224, opened 2018-03-28
HHH-1268 - Unidirection OneToMany causes duplicate key entry violation when removing from list

fromi:
This test case exposes the constraint violation that arise when OrderColumn is combined with a UniqueConstraint on the same column in child class.
Apparently, during a child delete, the update statements are done before the delete statement.
https://hibernate.atlassian.net/browse/HHH-5694 is the closest description for the bug.
It has been marked as a duplicate of HHH-1268, itself fixed with HHH-11587, but I can reproduce the problem on 5.3.0.RC1.


vladmihalcea:
Applied upstream, thanks

drahkrub:
See HHH-1268: Was a new issue opened due to this test case?

gbadner:
@drahkrub, yes, the new issue is https://hibernate.atlassian.net/browse/HHH-13338.

C-Otto:
Reproducer for bug in 5.4.29: https://github.com/C-Otto/HHH-13338

PR #2226, opened 2018-03-29
HHH-12436 - Attempted to assign id from null one-to-one property

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12436


vladmihalcea:
@gbadner Would you mind reviewing this fix. You know this part much better than I do.

hibernate-core/src/main/java/org/hibernate/type/AbstractType.java
```diff
@@ -134,9 +135,14 @@ public Object replace(
 			ForeignKeyDirection foreignKeyDirection) 
 	throws HibernateException {
 		boolean include;
-		if ( isAssociationType() ) {
+		if ( isAssociationType()) {
 			AssociationType atype = (AssociationType) this;
-			include = atype.getForeignKeyDirection()==foreignKeyDirection;
+			if ( atype instanceof OneToOneType && atype.useLHSPrimaryKey()) {
```
gbadner: I think the source of the bug is that the foreign key direction for ```atype``` is incorrect. It should be ```FROM_PARENT```, but it is ```TO_PARENT```.

I'm seeing both sides of the bidirectional associations have foreign key direction set to ```TO_PARENT```, which is clearly incorrect.

If the foreign key direction for ```Segunda#parent``` were correct, this fix would not be necessary.

I think the proper fix would be to fix the foreign key direction instead of disregarding it.

vladmihalcea: @gbadner Thanks, Gail. I'll investigate to see why the `getForeignKeyDirection()` returns `TO_PARENT` then.

gbadner:
I'm seeing changes in the generated SQL in the following tests due to this PR:
org.hibernate.test.annotations.derivedidentities.e4.b.DerivedIdentitySimpleParentSimpleDepMapsIdTest
org.hibernate.test.annotations.derivedidentities.e5.c.ForeignGeneratorViaMapsIdTest
org.hibernate.test.annotations.notfound.NotFoundOneToOneNonInsertableNonUpdateableTest
org.hibernate.test.annotations.onetoone.OneToOneTest
org.hibernate.test.annotations.onetoone.OptionalOneToOneMappedByTest
org.hibernate.test.annotations.onetoone.OptionalOneToOnePKJCTest
org.hibernate.test.criteria.mapsid.MapsIdOneToOneSelectTest

Before the PR is applied, the entity and its one-to-one association are loaded in a single query.

After the PR is applied, they are loaded using separate queries.

I'll have to look into why this is happening, as this can impact performance.

gbadner:
I debugged ```ForeignGeneratorViaMapsIdTest``` and found out why the association is loaded by a separate query.

This test has a unidirectional one-to-one association that shares the primary key column.

@Entity
public class MedicalHistory implements Serializable {
	@Id
	Integer id;

	@MapsId
	@JoinColumn(name = "patient_id")
	@OneToOne
	Person patient;
}

Before the PR is applied, the foreign key direction for the association is ```ForeignKeyDirection.TO_PARENT ```, which is obviously incorrect.

After the PR is applied, the foreign key direction  for the association is ```ForeignKeyDirection.FROM_PARENT```, which is correct.

When the ```EntityReturn``` for ```MedicalHistory``` is built, its ```AssociationKey``` is registered by ```AbstractLoadPlanBuildingAssociationVisitationStrategy#startingEntity```, which is the entity table and primary key for ```MedicalHistory```. 

Later, the ```AssociationKey``` for ```MedicalHistory#patient``` is registered by ```EntityBasedAssociationAttribute#getAssociationKey```.

Before the PR is applied, the ```AssociationKey``` will be the entity table and column for the associated entity (```Person```) because the association has the incorrect ```ForeignKeyDirection```.  Because Hibernate does not detect a duplicated ```AssociationKey```, an ```EntityFetch``` is ultimately added to the ```EntityReturn```. As a result, the SQL statement generated will include what is needed to fetch both the ```MedicalHistory``` and its association (```#patient```).

After the PR is applied, the association's ```AssociationKey``` will be the same table and column name as was previously registered for ```MedicalHistory``` (because they share the same column). Hibernate determines that it is a "duplicate" ```AssociationKey```, so an```EntityFetch``` for the association is not generated. Instead, the SQL statement generated will only load ```MedicalHistory```. Because the association is non-lazy, it gets loaded by a second query.

Some change needs to be made so that an ```EntityFetch``` gets added when the association uses the primary key(s).

gbadner:
I have a tentative fix that appears to also fix:
https://hibernate.atlassian.net/browse/HHH-10745

It may also fix:
https://hibernate.atlassian.net/browse/HHH-10842

I need to refine the fix, then I'll create a new PR.

JWGmeligMeyling:
@gbadner What is the status of your fix? The code in this branch also resolves [HHH-12842](https://hibernate.atlassian.net/browse/HHH-12842), another regression which was caused by HHH-12842, but then related to fetching and not persisting OneToOne associations. See also my analysis I've done at https://github.com/hibernate/hibernate-orm/pull/2451#issuecomment-406882521 while I was trying to resolve this issue myself in PR #2451. If there's no change in getting the fix for HHH-12436 and HHH-12842 in time, please consider reverting 2bfe300 so that at least both regressions are resolved.

JWGmeligMeyling:
@gbadner 

What about:

```diff
diff --git a/hibernate-core/src/main/java/org/hibernate/tuple/entity/EntityBasedAssociationAttribute.java b/hibernate-core/src/main/java/org/hibernate/tuple/entity/EntityBasedAssociationAttribute.java
index 2365eaef5f..52fda3a186 100644
--- a/hibernate-core/src/main/java/org/hibernate/tuple/entity/EntityBasedAssociationAttribute.java
+++ b/hibernate-core/src/main/java/org/hibernate/tuple/entity/EntityBasedAssociationAttribute.java
@@ -30,6 +30,7 @@ import org.hibernate.tuple.BaselineAttributeInformation;
 import org.hibernate.type.AnyType;
 import org.hibernate.type.AssociationType;
 import org.hibernate.type.ForeignKeyDirection;
+import org.hibernate.type.OneToOneType;
 
 import static org.hibernate.engine.internal.JoinHelper.getLHSColumnNames;
 import static org.hibernate.engine.internal.JoinHelper.getLHSTableName;
@@ -76,7 +77,9 @@ public class EntityBasedAssociationAttribute
 
 		final Joinable joinable = type.getAssociatedJoinable( sessionFactory() );
 
-		if ( type.getForeignKeyDirection() == ForeignKeyDirection.FROM_PARENT ) {
+		boolean oneToOnePrimaryKey = type instanceof OneToOneType && ((OneToOneType) type).isReferenceToPrimaryKey();
+
+		if ( type.getForeignKeyDirection() == ForeignKeyDirection.FROM_PARENT && !oneToOnePrimaryKey ) {
 			final String lhsTableName;
 			final String[] lhsColumnNames;
 
```

Just keeps most of the logic as is, but treats the `OneToOne` mapping as a `TO_PARENT` relation if it references the primary key. I don't think we have to consider `EntityType#isReferenceToPrimaryKey`, as a `@PrimaryJoinColumn` is only allowed for `@OneToOne` relations.

What kind of fix were you working on?

JWGmeligMeyling:
Branch should be rebased, only a very minor merge conflict (visibility change of method).

PR #2227, opened 2018-03-31
HHH-12453 - Update Vibur dependency from 22.0 to 22.1

simeonmalchev:


vladmihalcea:
Applied upstream, thanks

PR #2228, opened 2018-04-03
HHH-12454 - Offer flag to consider id generator with local scope (legacy non JPA behavior)

dreab8:
https://hibernate.atlassian.net/browse/HHH-12454

@sebersole any idea for a better name than ID_GENERATOR_NAMES_GLOBAL_SCOPE = "hibernate.id.generator.global_scope_name";?

Backporting to 5.2 do we have to keep the default value to true?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
@@ -975,7 +975,7 @@ public class SessionFactoryOptionsBuilder implements SessionFactoryOptions {
 
 	@Override
 	public JpaCompliance getJpaCompliance() {
-		return jpaCompliance;
+		return jpaCompliance.immutableCopy();
 	}
 
 	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
@@ -15,7 +15,6 @@ import java.util.Map;
 import java.util.TimeZone;
 import java.util.function.Supplier;
 
-import org.hibernate.AssertionFailure;
 import org.hibernate.ConnectionAcquisitionMode;
 import org.hibernate.ConnectionReleaseMode;
 import org.hibernate.CustomEntityDirtinessStrategy;
@@ -51,8 +50,8 @@ import org.hibernate.id.UUIDGenerator;
 import org.hibernate.id.uuid.LocalObjectUuidHelper;
 import org.hibernate.internal.log.DeprecationLogger;
 import org.hibernate.internal.util.config.ConfigurationHelper;
-import org.hibernate.jpa.JpaCompliance;
-import org.hibernate.jpa.spi.JpaComplianceImpl;
+import org.hibernate.jpa.spi.JpaCompliance;
+import org.hibernate.jpa.spi.MutableJpaCompliance;
 import org.hibernate.loader.BatchFetchStyle;
 import org.hibernate.proxy.EntityNotFoundDelegate;
 import org.hibernate.query.ImmutableEntityUpdateQueryHandlingMode;
@@ -235,7 +234,7 @@ public class SessionFactoryOptionsBuilder implements SessionFactoryOptions {
 
 	private Map<String, SQLFunction> sqlFunctions;
 
-	private JpaComplianceImpl jpaCompliance;
+	private MutableJpaCompliance jpaCompliance;
 
 	private boolean failOnPaginationOverCollectionFetchEnabled;
 
@@ -465,7 +464,7 @@ public class SessionFactoryOptionsBuilder implements SessionFactoryOptions {
 		);
 
 		// added the boolean parameter in case we want to define some form of "all" as discussed
-		this.jpaCompliance = new JpaComplianceImpl( configurationSettings, false );
+		this.jpaCompliance = context.getJpaCompliance();
 
 		this.failOnPaginationOverCollectionFetchEnabled = ConfigurationHelper.getBoolean(
 				FAIL_ON_PAGINATION_OVER_COLLECTION_FETCH,
@@ -978,8 +977,6 @@ public class SessionFactoryOptionsBuilder implements SessionFactoryOptions {
 	public JpaCompliance getJpaCompliance() {
 		return jpaCompliance;
 	}
-
-
 
 	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 	// In-flight mutation access
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
@@ -9,6 +9,7 @@ package org.hibernate.cfg;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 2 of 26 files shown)


sebersole:
`hibernate.id.generator.allow_illogical_jpa_global_definitions`? ;)

vladmihalcea:
@dreab8 Maybe `hibernate.id.generator.enable_legacy_entity_scope`?

sebersole:
`hibernate.id.generator.global_scoping_enabled`

dreab8:
hibernate.id.generator.global_scoping_enabled sound good to me
Thanks @sebersole and @vladmihalcea

sebersole:
Shame.. I really liked `allow_illogical_jpa_global_definitions` :)

vladmihalcea:
@dreab8 @sebersole If we are to backport this to 5.2, maybe we should use the `false` value by default.

From 5.3, we should have it enabled by default, but users will pay more attention to migration guides when upgrading to a major/minor version, than when they migrate to a patch verson.

What do you think?

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
```diff
@@ -760,7 +762,13 @@ else if ( InheritanceType.SINGLE_TABLE.equals( inheritanceState.getType() ) ) {
 		}
 
 		// try to find class level generators
-		HashMap<String, IdentifierGeneratorDefinition> classGenerators = buildGenerators( clazzToProcess, context );
+		HashMap<String, IdentifierGeneratorDefinition> classGenerators;
+		if ( context.getBuildingOptions().isIdGeneratorNameScopeGlobal() ) {
```
vladmihalcea: Here, there is no difference between the if and else branches. Did you have something in mind and forgot to add it?

sebersole:
> If we are to backport this to 5.2, maybe we should use the false value by default.

This is always tricky.  On the one hand is JPA compatibility and the other is upgradeability (legacy expectations).  TBH I am not exactly sure of the legacy expectations here.  @emmanuelbernard  probably knows better since he wrote all of the annotation binding/processing code.

I also think it is bad form to change these default values from release to release.  Whatever default we pick for 5.2 ought to the be the default moving forward.  IMO

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
dreab8: good catch, Thanks @vladmihalcea

dreab8:
@sebersole I'm torn but probably it is better to keep true as a default value also for 5.2.
@vladmihalcea I'm going to create a PR also for 5.2

sebersole:
Actually, I change my mind on this if I still can... this is really nothing more that a jpa compliance this.  So I think this should be one of those `"hibernate.jpa.compliance.<>"` / `JpaCompliance` things.  

`"hibernate.jpa.compliance.global_id_generators"`?

dreab8:
@sebersole in order to make `JpaCompliance `available to the `AnnotationBinder`, is it fine to add it to the `BootStrapContext`?
Currently the `JpaComliance` is available only through `SessionFactoryOptions`

sebersole:
Well, first do you agree that modeling it as a JpaCompliance setting is the correct/logical/natural thing to do?  

If not, no worries.

If so, yes, we will need to make it available during the earlier bootstrap phases - which is not a bad thing.  I do worry about the `JpaCompliance` object being mutated after its value(s) have been used.  E.g. if the user is allowed to change `JpaCompliance#isGlobalGeneratorScopeEnabled` (proposed name) after we have fully built the Metadata object - that would have a surprising outcome (no effect).  We just need to make sure that we properly "lock" the `JpaCompliance` at the point of Metadata -> SessionFactoryBuilder

dreab8:
I think the property is JPA compliance related  and I agree also with not allowing the user to change this property . 
Not sure but probably a simple solution for the immutability can be to initialise the property in the JpaComplianceImpl constructor and do not provide a setter method for it, what do you think?

sebersole:
Well its the same for the others.  They are all accessed via setters that
are just not exposed from `JpaCompliance` (only `JpaComplianceImpl`).  But
they are still there - even at runtime.  If we move this thing to
`BootstrapContext` then a user's only option to mutate these is through
some sort of reference to `JpaCompliance` (or `MutableJpaCompliance`). I'd
really hate for the "API" calls to be `( (JpaComplianceImpl)
metadataBuilder.getBootstrapContext().getJpaCompliance()
).setSomething(..)` - that's fugly.  Much better as something like just `
metadataBuilder.getBootstrapContext().getJpaCompliance().setSomething()`
which means that whatever is returned from
`BootstrapContext.getJpaCompliance()`
needs to be a mutable view.

dreab8:
For letting users to mutate these properties I was thinking to leave the `SessionFactoryOptionBuilder#enableJpaQueryCompliance` and the others methods you added.

The `BootstrapContext` build and return a `JpaCompliance` that does not provide any setter. 
The `SessionFactoryOptionBuilder` gives the user a way to mutate the properties values.

`JpaCompliance` can be an spi interface, but I may be wrong.

dreab8:
@sebersole   after thinking a little more :) , I realized your idea of having a mutable and immutable interface is  good, I'll work on this idea and make another PR.

Thanks

dreab8:
@sebersole tried to follow your suggestions, let me know what do u think.
Thanks

sebersole:
@dreab8 I see 2 PRs linked to this Jira - this one, and https://github.com/hibernate/hibernate-orm/pull/2232.  I assume one is out-of-date?  Anything else you are waiting on?

dreab8:
@sebersole can you give a look at this [last commit](https://github.com/hibernate/hibernate-orm/pull/2228/commits/a5fa173491ca9b98b5b01dc678203ef7eb1f78b1), it's related with JpaCompliance immutability.
The other [PR](https://github.com/hibernate/hibernate-orm/pull/2232) is for the 5.2 branch

sebersole:
Sure.. but first, do you think I'm being overly paranoid about the mutability?

dreab8:
@sebersole no I don't think you are :)

hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
```diff
@@ -975,7 +975,7 @@ public boolean isFailOnPaginationOverCollectionFetchEnabled() {
 
 	@Override
 	public JpaCompliance getJpaCompliance() {
-		return jpaCompliance;
+		return jpaCompliance.immutableCopy();
```
sebersole: Is this only ever called once?  If not, maybe we want to change this so we do only call it once.  Doubt it has a huge perf impact, but every bit helps

sebersole:
Applied upstream

icyerasor:
Found the property name _global_id_generators_ really confusing today; as to me it implied that the generators themselves would be shared globally, not only their definitions / names.
https://forum.hibernate.org/viewtopic.php?p=2338260 and https://vladmihalcea.com/hibernate-hidden-gem-the-pooled-lo-optimizer/ finally helped me to grasp the concept I guess (hope) :D

FireRedDev:
Hey @icyerasor ! Does this issue's flag fix problems with higher Hibernate versions with specifying @SequenceGenerator for extending a base entity with an id as mentioned in the comments in the top-answer here: https://stackoverflow.com/questions/8589928/mappedsuperclass-change-sequencegenerator-in-subclass

icyerasor:
Which comment are you referring to? The one that it doesn't work anymore with hibernate 5.4 ([here](https://stackoverflow.com/questions/8589928/mappedsuperclass-change-sequencegenerator-in-subclass#comment130111215_12151985))?
It does still work for me with hibernate 5.6.10. Seems it was only broken for a few minor versions, see info ([here](https://stackoverflow.com/questions/4560813/specifying-distinct-sequence-per-table-in-hibernate-on-subclasses#comment92100438_4561233) and the links [1](https://hibernate.atlassian.net/browse/HHH-12329), [2](https://in.relation.to/2018/04/26/hibernate-orm-5217-final-release/)).

Warning is logged as expected: `HHH000069: Duplicate generator name`.
Haven't tried 6+ yet though.

PR #2230, opened 2018-04-05
Use SAP HANA JDBC driver from Maven Central

breglerj:


vladmihalcea:
Applied upstream, thanks

PR #2231, opened 2018-04-07
HHH-12464 - NPE upon insert & delete with identity generated id

dreab8:
https://hibernate.atlassian.net/browse/HHH-12464


vladmihalcea:
@dreab8 I'd change it to:

`"The Persistence Context does not contain any entry for the [%s] entity instance. This may occur if an entity is created and then deleted in the same JPA EntityManager or Hibernate Session".`

sebersole:
"Skipping action - the persistence context does not contain any entry for the entity [%s]. This may occur if an entity is created and then deleted in the same transaction/flush."

dreab8:
Thanks @sebersole , applying your suggestion.

PR #2233, opened 2018-04-09
HHH-9663 added failing test case for Embeddable orphanRemoval

jaanushansen:
Created a failing test, which demonstrates, that https://hibernate.atlassian.net/browse/HHH-9663 is not fully solved:

When `@OneToOne(orphanRemoval = true)` is used in `@Embeddable` class, then setting the property to `null` does not delete the orphan entity.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/jpa/orphan/one2one/embedded/OneToOneLazyOrphanRemovalInEmbeddedEntityTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/jpa/orphan/one2one/embedded/OneToOneLazyOrphanRemovalInEmbeddedEntityTest.java
@@ -1,0 +1,184 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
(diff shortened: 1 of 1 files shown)


Naros:
@jaanushansen Thanks for the test, I'll give a look.

Naros:
@jaanushansen, while debugging your use case, I noticed that it falls into a block of code at [org.hibernate.engine.internal.Cascade.java#L263](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/engine/internal/Cascade.java#L263) which highlights that this use case presently is unsupported.  Did this ever work in the past that you're aware?

jaanushansen:
hi!

I have no knowledge how it worked in the past. I added the case because it feels incorrect that cascading is not working in this case.

vladmihalcea:
If this is very important to your application, you should try to send a Pull Request with a fix. Looking forward to your contribution.

beikov:
Thanks for the test case. Superseded by https://github.com/hibernate/hibernate-orm/pull/3817

PR #2235, opened 2018-04-11
HHH-12474 - Make sure the JPA version is defined by a single property across build files

dreab8:
https://hibernate.atlassian.net/browse/HHH-12474

@sebersole a strange thing is the `published-java-module.gradle` where using `project.jpaVersion` cause a
"Could not find method instruction() for arguments [Hibernate-JpaVersion, 2.2] on object of type org.gradle.api.internal.plugins.osgi.DefaultOsgiManifest" exception, while using `project.jpaVersion.name` works fine, any idea of the reason?

### DIFF
--- a/gradle/base-information.gradle
+++ b/gradle/base-information.gradle
@@ -10,7 +10,7 @@ apply plugin: 'base'
 ext {
 	hibernateVersion = '5.3.0-SNAPSHOT'
 	baselineJavaVersion = '1.8'
-	jpaVersion = '2.2'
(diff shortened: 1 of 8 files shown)


Sanne:
@dreab8 I'm guessing the `instruction` method expects a String, while `class JpaVersion` isn't. In some of the other places you're using it - like string appending - it happens to implicitly call the `toString()` method.

sebersole:
Yes, Sanne is right

Sanne:
merged. Thanks all!

PR #2238, opened 2018-04-12
HHH-12476 - NativeQuery with EntityGraphs assumes a HQL query

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12476


vladmihalcea:
@sebersole I applied your suggestion and pushed it.

PR #2239, opened 2018-04-12
HHH-12390 add DefaultJtaPlatform back in for per JVM TM 'org.hibernate.engine.transaction.jta.platform.spi.JtaPlatformProvider' service

scottmarlow:


scottmarlow:
I brought DefaultJtaPlatform back, to track per JVM default JTAPlatform.  

In WildFly, if the transaction manager is stopped and started again, any deployed (JPA) applications should undeploy/redeploy with the updated transaction manager.  In theory, we could handle this differently but the current JipiJapa spi supports the current way.

scottmarlow:
Mostly, so I can pass the JBossAppServerJtaPlatform instance in, as well as update the JBossAppServerJtaPlatform instance later as well.  

Perhaps an alternative could be to change DefaultJtaPlatform to JNDI lookup/cache the TM + TSR, then we wouldn't need any of the static delegation nonsense.

sebersole:
Or write a specialized JtaPlatform that does this.  See
`org.hibernate.engine.transaction.jta.platform.internal.AbstractJtaPlatform`.
Although may be better to write a whole new JtaPlatform impl from scratch
using `javax.naming.event.NamespaceChangeListener` to control reference
caching; or add the ability to clear the cached reference in
AbstractJtaPlatform which we could control from
`javax.naming.event.NamespaceChangeListener`.

sebersole:
Unless the reference we obtain from JNDI already hides this restart from
us...

scottmarlow:
My knowledge of what happens when the WildFly transaction manager service is restarted, may be dated.  When I last tried that, any applications using container managed JPA, would be (automatically) undeployed + redeployed.  More specifically, when any WildFly services that the JPA deployment depends on (e.g. datasource, transaction manager) is stopped, the application is undeployed, when the needed WildFly service is started again, the app will deploy again.  This undeploy/redeploy action, may help us enough when the reference we get from JNDI changes to a different value in WildFly.

scottmarlow:
out of date, closing...

In WildFly [https://github.com/wildfly/wildfly/blob/master/jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/service/WildFlyCustomJtaPlatformInitiator.java#L34], we just use org.hibernate.engine.transaction.jta.platform.internal.JBossAppServerJtaPlatform.

PR #2241, opened 2018-04-13
HHH-12469 - Add support for IN clause parameter padding to better reuse cached statements

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12469

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
@@ -1904,4 +1904,22 @@ public interface AvailableSettings extends org.hibernate.jpa.AvailableSettings {
 	 * @see org.hibernate.query.ImmutableEntityUpdateQueryHandlingMode
 	 */
 	String IMMUTABLE_ENTITY_UPDATE_QUERY_HANDLING_MODE = "hibernate.query.immutable_entity_update_query_handling_mode";
+
+	/**
+	 * By default, the IN clause expands to include all bind parameter values.
+	 * </p>
+	 * However, for database systems supporting execution plan caching,
+	 * there's a better chance of hitting the cache if the number of possible IN clause parameters lowers.
+	 * </p>
+	 * For this reason, we can expand the bind parameters to power-of-two: 4, 8, 16, 32, 64.
+	 * This way, an IN clause with 5, 6, or 7 bind parameters will use the 8 IN clause,
+	 * therefore reusing its execution plan.
+	 * </p>
+	 * If you want to activate this feature, you need to set this property to {@code true}.
+	 * </p>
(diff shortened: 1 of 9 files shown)


vladmihalcea:
@sebersole @dreab8 When you have time, can you review this PR? Thanks.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -1883,4 +1883,22 @@
 	 * @see org.hibernate.query.ImmutableEntityUpdateQueryHandlingMode
 	 */
 	String IMMUTABLE_ENTITY_UPDATE_QUERY_HANDLING_MODE = "hibernate.query.immutable_entity_update_query_handling_mode";
+
+	/**
+	 * By default, the IN clause expends to include all bind parameter values.
```
gsmet: Was curious about the feature so I read the patch: s/expends/expands/

vladmihalcea: Thanks for the tip. The IN clause expends quite some resources too :D

hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingsImpl.java
```diff
@@ -528,6 +527,25 @@ public String expandListValuedParameters(String queryString, SharedSessionContra
...
 			final Collection bindValues = entry.getValue().getBindValues();
 
+			int bindValueCount = bindValues.size();
+			int bindValueMaxCount = bindValueCount;
+
+			boolean inClauseParameterPaddingEnabled =
+					session.getFactory().getSessionFactoryOptions().inClauseParameterPaddingEnabled() &&
+					bindValueCount > 2;
+
+			if ( inClauseParameterPaddingEnabled ) {
```
sebersole: Not sure which is ultimately faster, but have you tried `final int bindValueMaxCount = (int) (1L << -Long.numberOfLeadingZeros(count - 1) );`?

sebersole: BTW, I got that from Guava Math classes

sebersole: Also, no matter which way we choose, I'd much prefer this be moved into a method.  The intent is not very clear in either approach - but calling a method like `#ceilingPowerOfTwo` (Guava's name) is much more readable, imo

vladmihalcea: I'll move it to some Utils class and check which one is faster as well.

PR #2243, opened 2018-04-15
HHH-12487 - Calling getTransaction() on a closed Session

dreab8:
https://hibernate.atlassian.net/browse/HHH-12487

trying to create the `TransactionDriverControl` in the `TransactionImpl` constructor when 
the `EntityManager` is closed causes an Exception to be thrown because the `LogicalConnection` is closed.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
@@ -12,6 +12,7 @@ import org.hibernate.HibernateException;
 import org.hibernate.TransactionException;
 import org.hibernate.engine.spi.ExceptionConverter;
 import org.hibernate.engine.transaction.spi.TransactionImplementor;
+import org.hibernate.internal.AbstractSharedSessionContract;
 import org.hibernate.internal.CoreLogging;
 import org.hibernate.jpa.spi.JpaCompliance;
 import org.hibernate.resource.transaction.spi.TransactionCoordinator;
@@ -31,29 +32,38 @@ public class TransactionImpl implements TransactionImplementor {
 	private final TransactionCoordinator transactionCoordinator;
 	private final ExceptionConverter exceptionConverter;
 	private final JpaCompliance jpaCompliance;
+	private final AbstractSharedSessionContract session;
 
 	private TransactionDriver transactionDriverControl;
 
 	public TransactionImpl(
 			TransactionCoordinator transactionCoordinator,
 			ExceptionConverter exceptionConverter,
-			JpaCompliance jpaCompliance) {
+			AbstractSharedSessionContract session) {
 		this.transactionCoordinator = transactionCoordinator;
 		this.exceptionConverter = exceptionConverter;
-		this.jpaCompliance = jpaCompliance;
+		this.jpaCompliance = session.getFactory().getSessionFactoryOptions().getJpaCompliance();
+		this.session = session;
 
-		this.transactionDriverControl = transactionCoordinator.getTransactionDriverControl();
+		if ( session.isOpen() && transactionCoordinator.isActive() ) {
+			this.transactionDriverControl = transactionCoordinator.getTransactionDriverControl();
+		}
+		else {
+			LOG.debug( "TransactionImpl created on closed Session/EntityManager" );
+		}
 
-		LOG.debugf(
-				"On TransactionImpl creation, JpaCompliance#isJpaTransactionComplianceEnabled == %s",
-				jpaCompliance.isJpaTransactionComplianceEnabled()
-		);
+		if ( LOG.isDebugEnabled() ) {
+			LOG.debugf(
+					"On TransactionImpl creation, JpaCompliance#isJpaTransactionComplianceEnabled == %s",
+					jpaCompliance.isJpaTransactionComplianceEnabled()
+			);
+		}
 	}
 
 	@Override
 	public void begin() {
-		if ( !transactionCoordinator.isActive() ) {
-			throw new TransactionException( "Cannot begin Transaction on closed Session/EntityManager" );
+		if ( !session.isOpen() ) {
+			throw new IllegalStateException( "Cannot begin Transaction on closed Session/EntityManager" );
 		}
 
 		if ( transactionDriverControl == null ) {
@@ -66,6 +76,7 @@ public class TransactionImpl implements TransactionImplementor {
 		}
 
 		LOG.debug( "begin" );
+
 		this.transactionDriverControl.begin();
 	}
 
@@ -82,6 +93,7 @@ public class TransactionImpl implements TransactionImplementor {
 		}
 
 		LOG.debug( "committing" );
+
 		try {
 			internalGetTransactionDriverControl().commit();
 		}
@@ -100,8 +112,9 @@ public class TransactionImpl implements TransactionImplementor {
 
 	@Override
 	public void rollback() {
-		if ( jpaCompliance.isJpaTransactionComplianceEnabled() ) {
-			if ( !isActive() ) {
+		if ( !isActive() ) {
+			if ( jpaCompliance.isJpaTransactionComplianceEnabled() ) {
+
 				throw new IllegalStateException(
 						"JPA compliance dictates throwing IllegalStateException when #rollback " +
 								"is called on non-active transaction"
@@ -121,6 +134,7 @@ public class TransactionImpl implements TransactionImplementor {
 		}
 
 		LOG.debug( "rolling back" );
+
 		if ( status != TransactionStatus.FAILED_COMMIT || allowFailedCommitToPhysicallyRollback() ) {
 			internalGetTransactionDriverControl().rollback();
 		}
@@ -136,7 +150,12 @@ public class TransactionImpl implements TransactionImplementor {
 	@Override
 	public boolean isActive(boolean isMarkedForRollbackConsideredActive) {
 		if ( transactionDriverControl == null ) {
-			transactionDriverControl = transactionCoordinator.getTransactionDriverControl();
+			if ( session.isOpen() ) {
+				transactionDriverControl = transactionCoordinator.getTransactionDriverControl();
+			}
+			else {
+				return false;
+			}
 		}
 		return transactionDriverControl.isActive( isMarkedForRollbackConsideredActive );
 	}
@@ -144,7 +163,12 @@ public class TransactionImpl implements TransactionImplementor {
 	@Override
 	public TransactionStatus getStatus() {
 		if ( transactionDriverControl == null ) {
-			transactionDriverControl = transactionCoordinator.getTransactionDriverControl();
+			if ( session.isOpen() ) {
+				transactionDriverControl = transactionCoordinator.getTransactionDriverControl();
+			}
+			else {
+				return TransactionStatus.NOT_ACTIVE;
+			}
 		}
 		return transactionDriverControl.getStatus();
 	}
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
```diff
@@ -392,13 +392,14 @@ public Transaction getTransaction() throws HibernateException {
 
 	@Override
 	public Transaction accessTransaction() {
-		if ( this.currentHibernateTransaction == null || this.currentHibernateTransaction.getStatus() != TransactionStatus.ACTIVE ) {
+		if ( this.currentHibernateTransaction == null
```
sebersole: I'm not sure this is correct from a JPA perspective.  JPA says that `EntityTransaction` is used to "serially ... begin and commit multiple transactions".  It sure seems like JPA expects that the same `EntityTransaction` be returned each time that `EntityManager#getTransaction` is called:

    EntityTransaction t = manager().getTransaction();
    assertSame( t, manager().getTransaction() );
    assertFalse( t.isActive() );
    t.begin();
    assertSame( t, manager().getTransaction() );
    assertTrue( t.isActive() );
    t.commit();
    assertSame( t, manager().getTransaction() );
    assertFalse( t.isActive() );
    manager().close();
    assertSame( t, manager().getTransaction() );
    assertFalse( t.isActive() );

As it stands we create a new `EntityTransaction` as soon the transaction is accessed after the commit.

I'm not sure I can say that the spec requires this (^^) behavior though, and the TCK surely seems to not enforce it.

hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
```diff
@@ -37,12 +37,20 @@
 	public TransactionImpl(
 			TransactionCoordinator transactionCoordinator,
 			ExceptionConverter exceptionConverter,
-			JpaCompliance jpaCompliance) {
+			JpaCompliance jpaCompliance,
+			boolean sessionIsClosed) {
```
sebersole: I wonder if I got too cute here.  I wonder if we should just pass (some form of) the Session into the Transaction here.  Then we could just ask the Session if it is open.

I did it this way for "testability", but I'm not sure that goal worth not having the Session available here.

hibernate-core/src/test/java/org/hibernate/jpa/test/transaction/CloseEntityManagerWithActiveTransactionTest.java
```diff
@@ -83,6 +86,23 @@ public void tearDown() throws Exception {
 		}
 	}
 
+	@Test(expected = IllegalStateException.class)
+	@TestForIssue( jiraKey = "HHH-12487")
+	public void getTransactionOnClosedJtaEntityManagerShouldReturnNUll() {
```
sebersole: We should change the name of this test... it should not return null.  So either `testJtaGetTransaction` or `testJtaGetTransactionThrowsError`

hibernate-core/src/test/java/org/hibernate/jpa/test/transaction/CloseEntityManagerWithActiveTransactionTest.java
```diff
@@ -83,6 +86,23 @@ public void tearDown() throws Exception {
...
+		EntityManager em = createEntityManager();
+		em.close();
+		em.getTransaction();
+		fail( "Calling getTransaction on a JTA entity manager should throw an IllegalStateException" );
+	}
+
+	@Test(expected = IllegalStateException.class)
+	@TestForIssue( jiraKey = "HHH-12487")
+	public void getTransactionOnAJtaEntityManagerShouldReturnNUll() {
```
sebersole: Same regarding name - this should not return null, it should throw an exception.  The assertions are correct, the method name should change.

hibernate-core/src/test/java/org/hibernate/jpa/test/transaction/CloseEntityManagerWithActiveTransactionTest.java
sebersole: Also, neither this method nor `#getTransactionOnAJtaEntityManagerShouldReturnNUll` actually deal with an active transaction - so not sure they belong in a test named `CloseEntityManagerWithActiveTransactionTest`

hibernate-core/src/test/java/org/hibernate/jpa/test/transaction/ClosedEntityManagerWithJpaTransactionComplianceTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+	@Test
+	public void getTransactionOnClosedEntityManagerShouldReturnNotNUll() {
+		EntityManager em = createEntityManager();
+		em.close();
+		assertThat( em.getTransaction(), CoreMatchers.notNullValue() );
+	}
+
+	@Test
+	public void getTransactionOnClosedEntityManagerCallIsActive() {
```
sebersole: Method naming again.. we are expecting *not* null.

sebersole: Also, in the interest of shortening names... not sure the `getTransactionOnClosedEntityManager...` "prefix" is needed for the test names when they are part of a class named `ClosedEntityManagerWithJpaTransactionComplianceTest`.

Same for the methods on `ClosedEntityManagerWithoutJpaTransactionComplianceTest` and `ClosedEntityManagerWithoutJpaTransactionComplianceTest`

dreab8:
@sebersole I agree with all your comments, so I changed my PR to meet all of them.
I hope the changes are fine.

Thanks for the review.

hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
```diff
@@ -31,29 +32,36 @@
...
 	private final JpaCompliance jpaCompliance;
+	private final AbstractSharedSessionContract session;
 
 	private TransactionDriver transactionDriverControl;
 
 	public TransactionImpl(
 			TransactionCoordinator transactionCoordinator,
 			ExceptionConverter exceptionConverter,
-			JpaCompliance jpaCompliance) {
+			JpaCompliance jpaCompliance,
```
sebersole: If we are passing Session, is there a reason to continue to pass along `JpaCompliance`?  We can just get that from `session.getFactory().getSessionFactoryOptions().getJpaCompliance()`

hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
```diff
@@ -31,29 +32,36 @@
...
+			AbstractSharedSessionContract session) {
 		this.transactionCoordinator = transactionCoordinator;
 		this.exceptionConverter = exceptionConverter;
 		this.jpaCompliance = jpaCompliance;
+		this.session = session;
 
-		this.transactionDriverControl = transactionCoordinator.getTransactionDriverControl();
+		if ( session.isOpen() && transactionCoordinator.isActive() ) {
+			this.transactionDriverControl = transactionCoordinator.getTransactionDriverControl();
+		}
```
sebersole: else {
        LOG.debugf( "Transaction initiated on closed Session/SessionFactory" );
    }

?

sebersole: This is a question :)

hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
```diff
@@ -64,8 +72,9 @@ public void begin() {
 		if ( isActive() ) {
 			throw new IllegalStateException( "Transaction already active" );
 		}
-
-		LOG.debug( "begin" );
+		if ( LOG.isDebugEnabled() ) {
```
sebersole: Why are we changing these?  Generally this is done to avoid expensive resolution of the message to log.  Here, that is not the case.  In fact, doesn't this perform worse?  IIUC this simply forces 2 "is debug enabled" checks

hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
```diff
@@ -80,8 +89,9 @@ public void commit() {
 			// has not been marked for rollback only
 			throw new IllegalStateException( "Transaction not successfully started" );
 		}
-
-		LOG.debug( "committing" );
+		if ( LOG.isDebugEnabled() ) {
```
sebersole: Same as above wrt "logging protection"

hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
```diff
@@ -112,15 +123,19 @@ public void rollback() {
 		TransactionStatus status = getStatus();
 		if ( status == TransactionStatus.ROLLED_BACK || status == TransactionStatus.NOT_ACTIVE ) {
 			// Allow rollback() calls on completed transactions, just no-op.
-			LOG.debug( "rollback() called on an inactive transaction" );
+			if ( LOG.isDebugEnabled() ) {
```
sebersole: Same as above wrt "logging protection"

hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
```diff
@@ -112,15 +123,19 @@ public void rollback() {
...
+			}
 			return;
 		}
 
 		if ( !status.canRollback() ) {
 			throw new TransactionException( "Cannot rollback transaction in current status [" + status.name() + "]" );
 		}
 
-		LOG.debug( "rolling back" );
+		if ( LOG.isDebugEnabled() ) {
```
sebersole: Same as above wrt "logging protection"

hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
```diff
@@ -175,25 +200,29 @@ public void markRollbackOnly() {
...
+			// Since this is the JPA-defined one, we make sure the txn is active first
+			// so long as compliance (JpaCompliance) has not been defined to disable
+			// that check - making this active more like Hibernate's #markRollbackOnly
+			if ( jpaCompliance.isJpaTransactionComplianceEnabled() ) {
 				throw new IllegalStateException(
 						"JPA compliance dictates throwing IllegalStateException when #setRollbackOnly " +
 								"is called on non-active transaction"
 				);
 			}
+			else if ( LOG.isDebugEnabled() ) {
```
sebersole: Same as above wrt "logging protection"

hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
dreab8: no reason to pass `JpaCompliance`

hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
dreab8: here I'm not sure about the message, the transaction is not really initiated .

dreab8: what about 
`LOG.debug("Transaction acquired on closed Session/EntityManager")`
?

sebersole: `LOG.debug( "TransactionImpl created on closed  Session/EntityManager")`?  That would match the phrasing of the other part of this method ("On TransactionImpl creation, JpaCompliance#isJpaTransactionComplianceEnabled == %s").

dreab8: +1

dreab8:
@sebersole added the log message, I'm going to push it upstream.

Thanks for the review

sebersole:
Sweet.  Thanks :)

PR #2244, opened 2018-04-16
HHH-12365 - User Guide: call_key should change to call_timestamp_epoch

Turini:
fix SQL statement using an invalid key name


vladmihalcea:
Applied upstream, thanks

PR #2245, opened 2018-04-16
HHH-12451 - Fixing JDBC and Java Type order for CurrencyType

Turini:


vladmihalcea:
Applied upstream, thanks

PR #2246, opened 2018-04-16
HHH-12484: improved error logs

nikowitt:
Hi there,

my pull request for
https://hibernate.atlassian.net/browse/HHH-12484

Regards,
Niko

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/proxy/AbstractLazyInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/proxy/AbstractLazyInitializer.java
@@ -115,7 +115,7 @@ public abstract class AbstractLazyInitializer implements LazyInitializer {
 					entityName,
 					id
 				);
-				throw new HibernateException( "illegally attempted to associate a proxy with two open Sessions" );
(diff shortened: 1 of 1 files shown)


vladmihalcea:
Applied upstream, thanks

hibernate-core/src/main/java/org/hibernate/proxy/AbstractLazyInitializer.java
```diff
@@ -115,7 +115,7 @@ else if ( isConnectedToSession() ) {
 					entityName,
 					id
 				);
-				throw new HibernateException( "illegally attempted to associate a proxy with two open Sessions" );
+				throw new HibernateException( "illegally attempted to associate proxy (" + entityName + "/" + id + ") with two open Sessions" );
```
sebersole: The standard way we refer to entity "instances" in log messages is `<entity-name>#<id-value>`.  We should be consistent

nikowitt: Changed output: https://github.com/hibernate/hibernate-orm/pull/2248

PR #2248, opened 2018-04-17
Hhh 12484 improved error logs

nikowitt:
Adapted identifier output accordingly.


vladmihalcea:
This is already integrated as suggested in Steve's comment.

PR #2250, opened 2018-04-18
HHH-12523 Invoke CacheTransactionSynchronization.transactionCompleting

rvansa:


vladmihalcea:
Applied upstream, thanks

PR #2251, opened 2018-04-19
HHH-12505 - Add possibility to disable metadata scanning

petteripitkanen:
https://hibernate.atlassian.net/browse/HHH-12505


Naros:
@petteripitkanen Could you also include a test case with this PR that enables this scanner and tests if the behavior is expected?

petteripitkanen:
I extended ScanningCoordinatorTest with two tests:

- testManagedResourcesAfterCoordinateScanWithDisabledScanner() runs coordinateScan() with DisabledScanner and asserts that ManagedResources is empty.
- testManagedResourcesAfterCoordinateScanWithCustomEnabledScanner() instantiates custom scanner that returns non-empty result, runs coordinateScan() with that, and asserts that ManagedResources is not empty.

Naros:
Applied upstream

PR #2255, opened 2018-04-25
HHH-12508 HHH-12520 (5.1)

gbadner:
This PR applies to 5.1 only.

https://hibernate.atlassian.net/browse/HHH-12508
https://hibernate.atlassian.net/browse/HHH-12520


gbadner:
Merged.

PR #2257, opened 2018-04-27
Minor fixes for SAP HANA

breglerj:
- Use "SYS.DUMMY" instead of "DUMMY" in queries to avoid errors due to a table
  named dummy in the current schema
- Remove override for Dialect#supportsCircularCascadeDeleteConstraints from
  AbstractHANADialect


Naros:
@breglerj If there is not a jira for this, could you add and reference it in your commit?

breglerj:
Superseded by #2263 and #2264

PR #2258, opened 2018-04-27
Fix Typo in PersistEvent

Klauswk:
Just fixing a typo i found while using Hibernate, feel free to accept or just fix in master.

Cheers.


Naros:
Applied upstream to master and 5.2.

PR #2262, opened 2018-04-30
Updated ParameterBindImpl to fix null error

matsebas:
Fixes the error that occurs when the value of the stored procedure parameter is set to null. 
Now when the value is null it checks if procedure's parameter allows null.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/procedure/internal/ParameterBindImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/procedure/internal/ParameterBindImpl.java
@@ -78,10 +78,12 @@ public class ParameterBindImpl<T> implements ParameterBind<T> {
 			throw new IllegalStateException( "Can only bind values for IN/INOUT parameters : " + procedureParameter );
 		}
 
-		if ( procedureParameter.getParameterType() != null ) {
-			if ( !procedureParameter.getParameterType().isInstance( value ) ) {
-				throw new IllegalArgumentException( "Bind value [" + value + "] was not of specified type [" + procedureParameter.getParameterType() );
-			}
+		if ( value == null && !procedureParameter.isPassNullsEnabled() ) {
+			throw new IllegalArgumentException( "Parameter '" + procedureParameter.getName() +  "' doesn't allow NULL value" );
+		}
+		
+		if ( value != null && !procedureParameter.getParameterType().isInstance( value ) ) {
+			throw new IllegalArgumentException( "Bind value [" + value + "] was not of specified type [" + procedureParameter.getParameterType() );
 		}
 
 		this.value = value;
--- a/_config.yml
+++ b/_config.yml
@@ -1,0 +1,1 @@
+theme: jekyll-theme-hacker


hibernate-core/src/main/java/org/hibernate/procedure/internal/ParameterBindImpl.java
```diff
@@ -78,10 +78,12 @@ private void internalSetValue(T value) {
...
 
-		if ( procedureParameter.getParameterType() != null ) {
-			if ( !procedureParameter.getParameterType().isInstance( value ) ) {
-				throw new IllegalArgumentException( "Bind value [" + value + "] was not of specified type [" + procedureParameter.getParameterType() );
-			}
+		if ( value == null && !procedureParameter.isPassNullsEnabled() ) {
+			throw new IllegalArgumentException( "Parameter '" + procedureParameter.getName() +  "' doesn't allow NULL value" );
+		}
+		
+		if ( value != null && !procedureParameter.getParameterType().isInstance( value ) ) {
```
Naros: Is there any particular reason why your change dropped the original outer if-check here?

matsebas: Yes, because It is unnecessary due to that the parameter type at this code point never be null.
If you register the parameter using annotation `@StoredProcedureParameter` it is mandatory, and if you try to create it programmatically avoiding to set the type, the app will throw a `java.lang.NullPointerException` because of code `type.getName()` at line 322 of class `org.hibernate.procedure.internal.ProcedureCallImpl`

Naros: Thanks, could you please attach a test case that reproduces the NPE with your fix and open a JIRA if you have not already done so for this change?

matsebas: Ok, I will have a delay to attach the test because I need to recreate the same problem in a in-memory db, building a stored procedure with nullable parameter, and calling the sp using a null value in the parameter.

Naros: @matsebas Any update?

matsebas:
Sorry Chris I’ve been working full-time, I will try to do unit tests next
week. Regards.

hibernate-core/src/main/java/org/hibernate/procedure/internal/ParameterBindImpl.java
matsebas: Sorry Chris I’ve been working full-time, I will try to do unit tests next week.

gsmet: @matsebas still interested in a test case. It looks like a good fix but we really need a test case to include it.

Thanks!

matsebas:
Hi, I'm getting some problems to compile the whole hibernate project, and it is delaying me to finish tests. As soon as I resolve compilation problems I will commit fix tests.

hibernate-core/src/main/java/org/hibernate/procedure/internal/ParameterBindImpl.java
matsebas: Hi Guillaume (@gsmet), I return to code time, and I look that this PR should be closed because it was resolved in commit https://github.com/hibernate/hibernate-orm/commit/cc1ffbaaa066df990d34aea6cb4ae59f3bbfbaf4

gsmet: @matsebas OK, perfect, thanks for the follow-up!

PR #2263, opened 2018-04-30
HHH-12535: SAP HANA dialect doesn't support circular cascade delete constraints

breglerj:
Fix for [HHH-12535](https://hibernate.atlassian.net/browse/HHH-12535)


vladmihalcea:
Applied upstream, thanks

PR #2264, opened 2018-04-30
HHH-12534: SAP HANA dialects use unqualified dummy table in queries

breglerj:
Fix for [HHH-12534](https://hibernate.atlassian.net/browse/HHH-12534)


vladmihalcea:
Applied upstream, thanks

PR #2265, opened 2018-05-02
HHH-12362: fix test case

breglerj:
- org.hibernate.test.queryhint.QueryHintTest uses
  assertNotNull(String#contains(String)) which always evaluates to true


vladmihalcea:
Applied upstream, thanks

PR #2266, opened 2018-05-02
HHH-12537: Query hint test for SAP HANA

breglerj:
- add org.hibernate.test.queryhint.QueryHintHANATest


vladmihalcea:
Applied upstream, thanks

PR #2267, opened 2018-05-03
HHH-12541: Test for SAP HANA calculation views

breglerj:


vladmihalcea:
Applied upstream, thanks

PR #2270, opened 2018-05-04
HHH-12011 - Fix static metamodel generation with TYPE_USE targetted annotations.

Naros:
https://hibernate.atlassian.net/browse/HHH-12011

### DIFF
--- a/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/util/TypeUtils.java
+++ b/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/util/TypeUtils.java
@@ -17,6 +17,7 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.ArrayType;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.ExecutableType;
 import javax.lang.model.type.TypeKind;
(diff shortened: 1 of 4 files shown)


Naros:
@vladmihalcea 

You'll notice a comment in the code in the PR which basically explains the following scenario where a primitive array is encountered it generates a static metamodel with the type as-is.  e.g.

```
@Entity
public class TheEntity {
  private byte[] byteArrayNonAnnotated;
}
```
generates
```
public static volatile SingularAttribute<TheEntity, byte[]> byteArrayNonAnnotated;
```

My concern here is that if that had not been an array but instead just a primitive type, the static metamodel gets generated using the object wrappers instead.  e.g.
```
@Entity
public class TheEntity {
  private byte byteField;
}
```
generates
```
public static volatile SingularAttribute<TheEntity, Byte> byteField;
```

Personally I think it should work the same in both instances, either always generating the metamodel with object wrappers or not converting them to the object-equivalents if its a primitive.  I think this is something we likely should discuss as a group and see what the spec mandates here.

In order to remain backward compatible here, I simply had the logic adhere to what its always done but to handle the TYPE_USE annotated-type properly.  e.g.
```
@Entity
public class TheEntity {
  @MySize
  private byte[] byteArray;
}
```
generates
```
public static volatile SingularAttribute<TheEntity, byte[]> byteArray;
```
Let me know what you think.

tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/util/TypeUtils.java
```diff
@@ -60,6 +61,40 @@ private TypeUtils() {
...
+			// that primitive non-array types.  For example, primitive non-arrays are returned as their
+			// autoboxed object types where primitive arrays are returned as-is.
+			//
+			// In short this means that a field of type "byte" is generated as "Byte" in the metamodel
+			// generated class but a "byte[]" is generated as "byte[]" rather than "Byte[]".
+			//
+			final TypeMirror typeMirror = ( (ArrayType) type ).getComponentType();
+			final List<? extends AnnotationMirror> annotationMirrors = typeMirror.getAnnotationMirrors();
+			if ( !annotationMirrors.isEmpty() ) {
+				String typeName = typeMirror.toString();
```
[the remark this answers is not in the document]
Naros: I'll give a look and see.

tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/util/TypeUtils.java
```diff
@@ -61,46 +61,29 @@ private TypeUtils() {
...
+						@Override
+						protected TypeMirror defaultAction(TypeMirror e, Void aVoid) {
+							return e;
+						}
+					},
+					null
+			);
+			return extractClosestRealTypeAsString( component, context ) + "[]";
+		}
+		return toTypeString( type );
```
gsmet: So in this case, you're not returning an array type?

Naros: It does that inherently with how the `toString()` implementation works for `ArrayType`.

In other words, `ArrayType` for a `byte[]` will return `byte[]` when calling `toString()`.  When the `ArrayType` is not annotated with a TYPE_USE annotation, we fallback to this default behavior.  The implementation I added specifically catches the TYPE_USE case and uses the visitor to address the shortcomings of the `toString()` implementation.

I am still looking at ways to best eliminate the `toString()` usage all together here if we can.

PR #2271, opened 2018-05-05
HHH-12090 improving ReflectHelper support for Java 8 default methods

Turini:


vladmihalcea:
Applied upstream, thanks

PR #2272, opened 2018-05-08
HHH-10435 Create class loaders in a privileged block

dmlloyd:


Sanne:
adapted to code style and merged. Thanks @dmlloyd !

I'll fix JIRA once I've merged it in the other branches.

PR #2274, opened 2018-05-08
HHH-10435 Create class loaders in a privileged block (5.1)

dmlloyd:


gbadner:
The test mentioned in HHH-10435 does not fail in EAP 7.1, so there is no test to reproduce this issue.

This fix is a first step to fixing HHH-12542, so I'm going to put off merging this PR until HHH-12542/WFLY-9064 is fixed.

dmlloyd:
https://ci.wildfly.org/viewLog.html?buildId=96925&tab=buildResultsDiv&buildTypeId=WF_MasterSecurityManager#testNameId219335434146728075 does fail; I'm not certain that EAP is doing full test runs with the security manager enabled though.  So, I would say there is a test that reproduces the issue.

dmlloyd:
In any event the problem is statically verifiable (there is an execution path that goes from deployment to class loader creation with no intervening doPrivileged block), so I'd say a reproducing test isn't necessary to show that there is a problem.

gbadner:
Closing because 5.1 is no longer maintained. HHH-10435 has been fixed in newer versons.

PR #2275, opened 2018-05-08
HHH-12553 - Prevent ConcurrentModificationException in OsgiClassLoader

MrEasy:
Synchronizing the Sets to retain insertion-order.
Making caches thread-safe.
See issue for details.


hibernate-osgi/src/main/java/org/hibernate/osgi/OsgiClassLoader.java
```diff
@@ -51,32 +60,37 @@ public OsgiClassLoader() {
...
-				final Class clazz = classLoader.loadClass( name );
-				if ( clazz != null ) {
-					classCache.put( name, clazz );
-					return clazz;
-				}
-			}
-			catch ( Exception ignore ) {
-			}
+
+	    synchronized (bundles) {
```
vladmihalcea: If we use `synchronized` every time we access `bundles`, then `bundles` can be a `HashSet` instead of a `Collections.synchronized` wrapper

hibernate-osgi/src/main/java/org/hibernate/osgi/OsgiClassLoader.java
```diff
@@ -29,12 +32,18 @@
...
+	// HHH-12553: must be thread-safe. Concurrent impl. would be best, but we have to retain insertion-order.
+	private Set<ClassLoader> classLoaders = Collections.synchronizedSet(new LinkedHashSet<ClassLoader>());
+	private Set<Bundle> bundles = Collections.synchronizedSet(new LinkedHashSet<Bundle>());
+
+	private ConcurrentMap<String, Class<?>> classCache = new ConcurrentHashMap<String, Class<?>>();
+	private ConcurrentMap<String, URL> resourceCache = new ConcurrentHashMap<String, URL>();
+
+	static
+	{
+		ClassLoader.registerAsParallelCapable();
```
vladmihalcea: Why is this change needed since the issue is about preventing the `ConcurrentModificationException`?

hibernate-osgi/src/main/java/org/hibernate/osgi/OsgiClassLoader.java
MrEasy: @vladmihalcea Hi - I left it intentionally a LinkedHashSet, since this was implemented before with HHH-12553 to retain the insertion-order. Using a HashSet would break this (and not be synchronized per-se as well, maybe you meant Hashmap?)

hibernate-osgi/src/main/java/org/hibernate/osgi/OsgiClassLoader.java
MrEasy: It is not needed to fix the CMFEx. But since the class became thread-safe with the change, I think it makes sense to mark is as such.
See HHH-12553 for some more comments.

hibernate-osgi/src/main/java/org/hibernate/osgi/OsgiClassLoader.java
brmeyer: Hey @vladmihalcea, @MrEasy is correct.  Super important that we maintain insertion order for these maps.

MrEasy: @brmeyer thanks for reviewing

hibernate-osgi/src/main/java/org/hibernate/osgi/OsgiClassLoader.java
```diff
@@ -8,33 +8,41 @@
...
-	private Set<Bundle> bundles = new LinkedHashSet<Bundle>();
+	// HHH-12553: must be thread-safe. Concurrent impl. would be best, but we have to retain insertion-order.
+	private Set<ClassLoader> classLoaders = Collections.synchronizedSet(new LinkedHashSet<ClassLoader>());
+	private Set<Bundle> bundles = Collections.synchronizedSet(new LinkedHashSet<Bundle>());
+
+	private ConcurrentMap<String, Class<?>> classCache = new ConcurrentHashMap<String, Class<?>>();
+	private ConcurrentMap<String, URL> resourceCache = new ConcurrentHashMap<String, URL>();
+
+	static {
+		ClassLoader.registerAsParallelCapable();
```
brmeyer: We might consider leaving this out, for now.  Honestly not sure if there are implications for `registerAsParallelCapable` in OSGi environments, nor am I sure if most OSGi containers even implement this.  @timothyjward, happen to know off the top of your head?

MrEasy: Depends on the ClassLoaders in use - JBOsgi's CL e.g. does implement it (KAraf as well afaik).
If in other containers the CL for the OSGi-bundles does not implement it, it becomes ineffective anyway.
Also fine with leaving it out.

timothyjward: Apache Felix certainly does - see https://github.com/apache/felix/blob/1878353ae257681e72e50c1c87f3900a13baed4e/framework/src/main/java/org/apache/felix/framework/BundleWiringImpl.java#L1982

brmeyer: @timothyjward Bonus points for the super fast answer ;)
@MrEasy We primarily target Felix and Equinox, and it looks like Equinox is supporting this as well.

brmeyer:
Nicely done @MrEasy, and honestly a little surprised this hasn't come up before now.  Thanks for the fix!

hibernate-osgi/src/main/java/org/hibernate/osgi/OsgiClassLoader.java
Sanne: Hi, not an OSGi user myself so I can only guess what `registerAsParallelCapable` means.. no objections to include it but it would be appreciate for such enhancements to have their separate JIRA and separeate PRs.

It would help with our flow and also help PRs being merged faster.

hibernate-osgi/src/main/java/org/hibernate/osgi/OsgiClassLoader.java
```diff
@@ -8,33 +8,41 @@
...
 public class OsgiClassLoader extends ClassLoader implements Stoppable {
 	// Leave these as Sets -- addClassLoader or addBundle may be called more
 	// than once if a SF or EMF is closed and re-created.
-	private Set<ClassLoader> classLoaders = new LinkedHashSet<ClassLoader>();
-	private Set<Bundle> bundles = new LinkedHashSet<Bundle>();
+	// HHH-12553: must be thread-safe. Concurrent impl. would be best, but we have to retain insertion-order.
+	private Set<ClassLoader> classLoaders = Collections.synchronizedSet(new LinkedHashSet<ClassLoader>());
+	private Set<Bundle> bundles = Collections.synchronizedSet(new LinkedHashSet<Bundle>());
+
+	private ConcurrentMap<String, Class<?>> classCache = new ConcurrentHashMap<String, Class<?>>();
```
Sanne: It looks like some caching is introduced. Was that really necessary? Can someone reassure me about leaks?

Also, it would help to have some comments about synchronization expecations. Why making these two use `ConcurrentHashMap` and then extensively `synchronize` on `bundles` and/or `classLoaders` ?

Sanne:
Is there a way to verify that the mentioned `ConcurrentModificationException` is addressed by this PR?

I might be missing some knowledge of OSGi but it looks like to address different aspects.

hibernate-osgi/src/main/java/org/hibernate/osgi/OsgiClassLoader.java
MrEasy: @Sanne Thanks for looking into it.
The registerAsParallelCapable you commented on above is not specific to OSGi, but to ClassLoaders in general.
See my comments in HHH-12553 for some more details.
The caching was not introduced by this change, but was already there and I retained it of course.
There is no synchronization on the concurrent maps, only on the not-thread-safe maps, see comment in code: "must be thread-safe. Concurrent impl. would be best, but we have to retain insertion-order."

MrEasy:
@Sanne since it is a race-condition, I have no deterministic way to prove it. What could be done is to access ClassLoaderServiceImpl to get resources from Hibernate in a concurrent manner to provoke the situation. Only a heuristic approach though.

Sanne:
@MrEasy ok, thanks for the explanations!

Sanne:
I finally had some time to fetch the patch and have a better look. This makes perfect sense, no need to add further tests as it's a straight forward fix.

Running integration tests, will merge it soon!

hibernate-osgi/src/main/java/org/hibernate/osgi/OsgiClassLoader.java
brmeyer: +1 -- I doubt it affects much in this specific case, but the separation of commits gives more traceability if side effects kick in.

hibernate-osgi/src/main/java/org/hibernate/osgi/OsgiClassLoader.java
brmeyer: @Sanne, no new caching here -- @MrEasy is purely changing the data types.  This is where we hold all of the discovered OSGi bundles/CLs for use in our aggregated CL.  These need to be discovered during bootstrapping and held permanently for the ORM bundles.

brmeyer:
@Sanne, thanks for the additional eyes on this.  @MrEasy, sincere apologies for not having gotten to it already...

PR #2277, opened 2018-05-09
HHH-12556 Share data structures between similar LoadPlan based EntityLoaders

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12556

This is part of a series of patch to improve our memory consumption when dealing with legacy batch fetch style.

With the OpenConcerto test case contributed by the OP, this change alone brings the memory used from **1002 MB** to **461 MB**.

I think this change is pretty safe API wise, that's why I isolated it.


Sanne:
Looks great, I'd like to merge it.
Minor polish opportunity: you could reuse the instances of `QueryBuildingParametersImpl` being created in org.hibernate.loader.entity.plan.EntityLoader.Builder#byUniqueKey as well?

Sanne:
nevermind, ignore my comment about QueryBuildingParametersImpl. it's including the batch size so no point in reusing that :)

gsmet:
@Sanne btw, I'm wondering if we should consider backporting it to 5.2.

gsmet:
@Sanne @sebersole merged and currently backporting to 5.2.

@gbadner once this one gets a bit of field testing, it might be a good idea to backport it to 5.1.

Sanne:
awesome! thanks @gsmet

PR #2278, opened 2018-05-09
[Discussion] Batching entity loader memory issue

gsmet:
So apart from https://github.com/hibernate/hibernate-orm/pull/2277, which seems to be pretty safe, here are another batch of changes on top of it.

1. Vanilla ORM: **1002 MB**
2. With HHH-12556: **461 MB**
3. With more template initialization (https://github.com/hibernate/hibernate-orm/pull/2278/commits/832fca76ac5d4cccd1565def44f19ed6923f9270): **420 MB**
-> these are without any lazy loading so this is the maximal memory that will be used when all the loaders are initialized
-> **UPDATE: 3/ is unsafe if `AbstractEntityPersister` is subclassed and the loader creation methods are overridden, which is the case for OGM for instance.**  Therefore, changes withdrawn.
4. With lazy loading (https://github.com/hibernate/hibernate-orm/pull/2278/commits/20ceebfc9abb8da31e84fdb88b94e40f0d18f1e0): **264MB**
-> the lazy loading is only about the loaders loaded per lock mode, we still load all the LoadPlan based entity loaders for all batch sizes. I still load eagerly the loaders for `LockMode.NONE` and `LockMode.READ`.

The issue I have with 3. is that a user might not have the same behavior if he has subclassed AbstractEntityPersister and overridden `createEntityLoader(LockMode lockMode)` as I use another method injecting a template.

The issue with 4. is:
- we lazy load so the user might get surprised he doesn't have enough memory at some point during the application lifecycle. That being said, we really have **a lot** of ``LockMode``s so for many (most?) application usage, it will be a gain as the user won't use all the ``LockMode``s.
- for now I use a `Map` as the `Map` is exposed to subclasses so I have the same doubts as for 3. as we change slightly the behavior of methods that could be overridden or used in user subclasses.

I have added GitHub comments to underline what might be an issue.

What you have in the OP's case is:
```
 \_ [LockMode 1] batching entity loader 1
     \_ [batch size 1] LoadPlan entity loader 1.1
     \_ ...
     \_ [batch size n] LoadPlan entity loader 1.n
 \_ ...
 \_ [LockMode n] batching entity loader n
     \_ [batch size 1] LoadPlan entity loader n.1
     \_ ...
     \_ [batch size n] LoadPlan entity loader n.n
```
 * Point 2. shares common data structures between the various LoadPlan entity loaders for a given lock mode
 * Point 3. reuses the common data structure of the first lock mode for the other lock modes: to be more precise, it reinjects the common data structure of LoadPlan entity loader 1.1 inside the LoadPlan entity loaders of batching entity loader n
 * Point 4. lazy loads the batching entity loader per lock mode (the LoadPlan entity loader are still eagerly loaded)

I haven't created JIRA issues for 3. and 4. as I think the approach is open to discussion.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/entity/BatchingEntityLoaderBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/entity/BatchingEntityLoaderBuilder.java
@@ -54,11 +54,21 @@ public abstract class BatchingEntityLoaderBuilder {
 			LockMode lockMode,
 			SessionFactoryImplementor factory,
 			LoadQueryInfluencers influencers) {
+		return buildLoader( persister, null, batchSize, lockMode, factory, influencers );
+	}
+
+	public UniqueEntityLoader buildLoader(
+			OuterJoinLoadable persister,
+			UniqueEntityLoader entityLoaderTemplate,
+			int batchSize,
+			LockMode lockMode,
+			SessionFactoryImplementor factory,
+			LoadQueryInfluencers influencers) {
 		if ( batchSize <= 1 ) {
 			// no batching
 			return buildNonBatchingLoader( persister, lockMode, factory, influencers );
 		}
-		return buildBatchingLoader( persister, batchSize, lockMode, factory, influencers );
+		return buildBatchingLoader( persister, entityLoaderTemplate, batchSize, lockMode, factory, influencers );
 	}
 
 	protected UniqueEntityLoader buildNonBatchingLoader(
@@ -75,6 +85,16 @@ public abstract class BatchingEntityLoaderBuilder {
 			LockMode lockMode,
 			SessionFactoryImplementor factory,
 			LoadQueryInfluencers influencers);
+
+	protected UniqueEntityLoader buildBatchingLoader(
+			OuterJoinLoadable persister,
+			UniqueEntityLoader entityLoaderTemplate,
+			int batchSize,
+			LockMode lockMode,
+			SessionFactoryImplementor factory,
+			LoadQueryInfluencers influencers) {
+		return buildBatchingLoader( persister, batchSize, lockMode, factory, influencers );
+	}
 
 	/**
 	 * Builds a batch-fetch capable loader based on the given persister, lock-options, etc.
@@ -93,11 +113,21 @@ public abstract class BatchingEntityLoaderBuilder {
 			LockOptions lockOptions,
 			SessionFactoryImplementor factory,
 			LoadQueryInfluencers influencers) {
+		return buildLoader( persister, null, batchSize, lockOptions, factory, influencers );
+	}
+
+	public UniqueEntityLoader buildLoader(
+			OuterJoinLoadable persister,
+			UniqueEntityLoader entityLoaderTemplate,
+			int batchSize,
+			LockOptions lockOptions,
+			SessionFactoryImplementor factory,
+			LoadQueryInfluencers influencers) {
 		if ( batchSize <= 1 ) {
 			// no batching
 			return buildNonBatchingLoader( persister, lockOptions, factory, influencers );
 		}
-		return buildBatchingLoader( persister, batchSize, lockOptions, factory, influencers );
+		return buildBatchingLoader( persister, entityLoaderTemplate, batchSize, lockOptions, factory, influencers );
 	}
 
 	protected UniqueEntityLoader buildNonBatchingLoader(
@@ -114,4 +144,14 @@ public abstract class BatchingEntityLoaderBuilder {
 			LockOptions lockOptions,
 			SessionFactoryImplementor factory,
 			LoadQueryInfluencers influencers);
+
+	protected UniqueEntityLoader buildBatchingLoader(
+			OuterJoinLoadable persister,
+			UniqueEntityLoader initialEntityLoader,
+			int batchSize,
+			LockOptions lockOptions,
+			SessionFactoryImplementor factory,
+			LoadQueryInfluencers influencers) {
+		return buildBatchingLoader( persister, batchSize, lockOptions, factory, influencers );
+	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/loader/entity/plan/LegacyBatchingEntityLoaderBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/entity/plan/LegacyBatchingEntityLoaderBuilder.java
@@ -47,6 +47,40 @@ public class LegacyBatchingEntityLoaderBuilder extends AbstractBatchingEntityLoa
 		return new LegacyBatchingEntityLoader( persister, batchSize, lockOptions, factory, influencers );
 	}
 
+	@Override
+	protected UniqueEntityLoader buildBatchingLoader(
+			OuterJoinLoadable persister,
+			UniqueEntityLoader entityLoaderTemplate,
+			int batchSize,
+			LockMode lockMode,
+			SessionFactoryImplementor factory,
+			LoadQueryInfluencers influencers) {
+		if (entityLoaderTemplate instanceof LegacyBatchingEntityLoader) {
+			return new LegacyBatchingEntityLoader( persister, (LegacyBatchingEntityLoader) entityLoaderTemplate, batchSize, lockMode, null, factory,
+					influencers );
+		}
+		else {
+			return new LegacyBatchingEntityLoader( persister, batchSize, lockMode, factory, influencers );
+		}
+	}
+
+	@Override
+	protected UniqueEntityLoader buildBatchingLoader(
+			OuterJoinLoadable persister,
+			UniqueEntityLoader entityLoaderTemplate,
+			int batchSize,
+			LockOptions lockOptions,
+			SessionFactoryImplementor factory,
+			LoadQueryInfluencers influencers) {
+		if (entityLoaderTemplate instanceof LegacyBatchingEntityLoader) {
+			return new LegacyBatchingEntityLoader( persister, (LegacyBatchingEntityLoader) entityLoaderTemplate, batchSize, null, lockOptions, factory,
+					influencers );
+		}
+		else {
+			return new LegacyBatchingEntityLoader( persister, batchSize, lockOptions, factory, influencers );
+		}
+	}
+
 	public static class LegacyBatchingEntityLoader extends BatchingEntityLoader  {
 		private final int[] batchSizes;
 		private final EntityLoader[] loaders;
@@ -57,7 +91,7 @@ public class LegacyBatchingEntityLoaderBuilder extends AbstractBatchingEntityLoa
 				LockMode lockMode,
 				SessionFactoryImplementor factory,
 				LoadQueryInfluencers loadQueryInfluencers) {
-			this( persister, maxBatchSize, lockMode, null, factory, loadQueryInfluencers );
+			this( persister, null, maxBatchSize, lockMode, null, factory, loadQueryInfluencers );
 		}
 
 		public LegacyBatchingEntityLoader(
@@ -66,11 +100,12 @@ public class LegacyBatchingEntityLoaderBuilder extends AbstractBatchingEntityLoa
 				LockOptions lockOptions,
 				SessionFactoryImplementor factory,
 				LoadQueryInfluencers loadQueryInfluencers) {
-			this( persister, maxBatchSize, null, lockOptions, factory, loadQueryInfluencers );
+			this( persister, null, maxBatchSize, null, lockOptions, factory, loadQueryInfluencers );
 		}
 
-		protected LegacyBatchingEntityLoader(
+		private LegacyBatchingEntityLoader(
 				OuterJoinLoadable persister,
+				LegacyBatchingEntityLoader batchingEntityLoaderTemplate,
 				int maxBatchSize,
 				LockMode lockMode,
 				LockOptions lockOptions,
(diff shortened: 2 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -236,7 +237,9 @@
 
 	private final Map uniqueKeyLoaders = new HashMap();
 	private final Map lockers = new HashMap();
-	private final Map loaders = new HashMap();
+	private UniqueEntityLoader noneLockLoader;
+	private UniqueEntityLoader readLockLoader;
+	private final Map<Object, UniqueEntityLoader> loaders = new ConcurrentHashMap<>();
```
gsmet: So the `Object` here is weird because apparently in the `loaders`, we put  loaders with a `LockMode` key but also with a `String` key: at least we try to get some using a `String` key. Not sure if it is a bug, decided to be on the safe side for this patch.

See https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java#L4260 .

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -4136,60 +4139,13 @@ protected Map getLoaders() {
...
 		final Map loaders = getLoaders();
 
-		UniqueEntityLoader readLoader = createEntityLoader( LockMode.READ );
-		loaders.put( LockMode.READ, readLoader );
+		noneLockLoader = createEntityLoader( LockMode.NONE );
+		loaders.put( LockMode.NONE, noneLockLoader );
 
-		loaders.put( LockMode.NONE, createEntityLoader( readLoader, LockMode.NONE ) );
+		readLockLoader = createEntityLoader( noneLockLoader, LockMode.READ );
+		loaders.put( LockMode.READ, readLockLoader );
```
gsmet: It's not necessary to add these 2 loaders to the map but I thought it might be safer considering the map is exposed to the subclasses.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -4201,6 +4157,42 @@ protected void createLoaders() {
 		);
 	}
```
gsmet: What worries me a bit here is this type of pattern:
```
public class MyEntityPersister extends AbstractEntityPersisters {
    @Override
    protected void createLoaders() {
        super.createLoaders();

        Map loaders = getLoaders();
        // do something with the loaders
    }
}
```
With the new code, the user would only have the 2 eagerly loaded loaders and not the others which might change slightly the behavior.

I don't know if this use case si supported by ORM.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
gsmet: It's not necessary to add these 2 loaders to the map but I thought it might be safer considering the map is exposed to the subclasses.

hibernate-core/src/main/java/org/hibernate/loader/entity/BatchingEntityLoaderBuilder.java
```diff
@@ -114,4 +144,14 @@ protected abstract UniqueEntityLoader buildBatchingLoader(
 			LockOptions lockOptions,
 			SessionFactoryImplementor factory,
 			LoadQueryInfluencers influencers);
```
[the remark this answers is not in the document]
sebersole: We currently do not provide a way for a user to supply a custom `BatchingEntityLoaderBuilder`.  All Hibernate's access to a `BatchingEntityLoaderBuilder` is done via `BatchingEntityLoaderBuilder#getBuilder` which simply picks a builder based on the configured `BatchFetchStyle`.  So in that sense we do not the user providing a custom `BatchingEntityLoaderBuilder`. 

Now users could *call* these methods...

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
sebersole: This is correct

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
sebersole: It is unfortunately just a line comment, not a javadoc comment, but the intention for `#getLoaders` is clearly just for access during `#createLoaders` - the expectation being that it is a reference usable to add loaders created.  And in fact, that is consistent with all of ORM's use of that method.

In retrospect we maybe should have instead defined `#createLoaders` as passing in the Map and not exposed `#getLoaders`.  But I think its not worth the effort to do that now considering the massive changes here in 6 anyway.

Really, the only supported access to these loaders from a consumer (of the persister) perspective is via `#getAppropriateLoader`.  Long story short, I think we can say that the usage of `#getLoaders` outlined here (really any usage besides putting a created Loader) is not supported.  The method is protected though and the intention comment is not part of the javadoc, so it is conceivable that user code does follow that expectation.  

Personally I say we just disallow that usage ppattern

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -4113,60 +4138,14 @@ protected Map getLoaders() {
 	//Relational based Persisters should be content with this implementation
 	protected void createLoaders() {
 		final Map loaders = getLoaders();
-		loaders.put( LockMode.NONE, createEntityLoader( LockMode.NONE ) );
 
-		UniqueEntityLoader readLoader = createEntityLoader( LockMode.READ );
-		loaders.put( LockMode.READ, readLoader );
+		noneLockLoader = createEntityLoader( LockMode.NONE );
+		loaders.put( LockMode.NONE, noneLockLoader );
```
sebersole: Why do both (field + Map entry) for NONE and READ?  Not understanding that part.  Is it so that callers can continue to get access to those Loaders via `#getLoaders`?  Just like discussed above, `#getLoaders` is intended to have a very specific use - a reference by which to add a new Loader.  If that's true there, its true here too

Just drop those 2 from

gsmet: I thought I added a comment about that on GitHub: I wasn't really sure about how safe we should be with the `loaders` map.

But sure, if we don't expect the user to use this map, we can drop those 2 from the map.

hibernate-core/src/main/java/org/hibernate/loader/entity/BatchingEntityLoaderBuilder.java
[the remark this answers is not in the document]
gsmet: > Now users could call these methods...

Yes, that's why I kept the old methods. And they are working as they used to.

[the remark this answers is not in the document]
sebersole: Just answering your implicit question - the user decidedly cannot "override this method".

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -2469,6 +2472,24 @@ protected UniqueEntityLoader createEntityLoader(
 				.buildLoader( this, batchSize, lockOptions, getFactory(), loadQueryInfluencers );
```
gsmet: @sebersole ah, I misplaced the comments about the overriding method.

The issue is here: if the user has its own `AbstractEntityPersister` overriding the two `createEntityLoaders()` methods above, with this patch, they won't be called for most lock modes as it's the 2 methods below with the template that will be called.

That's what was bothering me. I don't know if it's a use case we support: I saw a documentation example with a custom class extending `SingleTableEntityPersister` somewhere so I suppose it can happen.

And this part would definitely breaks it.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
gsmet: OK, so given that, the patch with the lazy loading would work. Considering how the code is organized, I think we are stuck with using a `Map` and thus a CHM.

@Sanne any insights to share about the concurrency?

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
gsmet: Indeed. It even overrides `createLoaders()` due to a TODO we should probably fix. I should have thought about OGM!

hibernate-core/src/main/java/org/hibernate/loader/entity/plan/LegacyBatchingEntityLoaderBuilder.java
```diff
@@ -47,6 +47,40 @@ protected UniqueEntityLoader buildBatchingLoader(
...
 
+	@Override
+	protected UniqueEntityLoader buildBatchingLoader(
+			OuterJoinLoadable persister,
+			UniqueEntityLoader entityLoaderTemplate,
+			int batchSize,
+			LockMode lockMode,
+			SessionFactoryImplementor factory,
+			LoadQueryInfluencers influencers) {
+		if (entityLoaderTemplate instanceof LegacyBatchingEntityLoader) {
```
golovnin: As far as I know this violates the stupid code style used by Hibernate. You must add spaces after opening brace and before closing brace.

hibernate-core/src/main/java/org/hibernate/loader/entity/plan/LegacyBatchingEntityLoaderBuilder.java
```diff
@@ -47,6 +47,40 @@ protected UniqueEntityLoader buildBatchingLoader(
...
+
+	@Override
+	protected UniqueEntityLoader buildBatchingLoader(
+			OuterJoinLoadable persister,
+			UniqueEntityLoader entityLoaderTemplate,
+			int batchSize,
+			LockOptions lockOptions,
+			SessionFactoryImplementor factory,
+			LoadQueryInfluencers influencers) {
+		if (entityLoaderTemplate instanceof LegacyBatchingEntityLoader) {
```
golovnin: Code style again.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -4178,6 +4157,42 @@ protected void createLoaders() {
 		);
 	}
 
+	private UniqueEntityLoader getLoaderByLockMode(LockMode lockMode) {
+		if ( LockMode.NONE.equals( lockMode ) ) {
```
golovnin: LockMode is an enum. You can use == to compare values. Same applies to the line 4164.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -4178,6 +4157,42 @@ protected void createLoaders() {
...
+			return noneLockLoader;
+		}
+		else if ( LockMode.READ.equals( lockMode ) ) {
+			return readLockLoader;
+		}
+
+		return loaders.computeIfAbsent( lockMode, k -> createLazyLoadedEntityLoader( lockMode ) );
```
golovnin: Please use the method reference "this:: createLazyLoadedEntityLoader".

gsmet: I can't do that because, unfortunately, the key of the `loaders` map is an `Object` not a `LockMode`. This is a bad idea to begin with but we can't really fix it without breaking some expectations about `AbstractEntityPersister`, which can be subclassed by the user.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
gsmet: I prefer it like that. Old habit from my early Sonar days.

hibernate-core/src/main/java/org/hibernate/loader/entity/plan/LegacyBatchingEntityLoaderBuilder.java
gsmet: Ah yes. I like this codestyle :). Missing the strict checkstyle rules we have in Search, OGM and Validator :). Will fix.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
gsmet: Yeah sure... Not sure it's cleaner to be honest but why not.

FWIW, we got the confirmation from our JDK team that a method reference is as bad as a lambda so it won't change anything IMHO.

At some point they might improve the situation so let's get rid of the lambda.

gsmet:
I extracted the lazy loading in https://github.com/hibernate/hibernate-orm/pull/2279 . The other change is not safe due to backwards compatibility issues so it should be discarded.

Closing this PR.

PR #2279, opened 2018-05-09
HHH-12558 Lazy load entity loaders for the less common lock modes

gsmet:
@sebersole @Sanne I extracted the other "rather" safe change from https://github.com/hibernate/hibernate-orm/pull/2278 .

Two things:
* @Sanne interested in your insights about concurrency. I don't think we can do better because we have to keep the map for backwards compatibility purposes.
* I don't think we should backport this to 5.2, let's keep it for 5.3.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
@@ -21,6 +21,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.hibernate.AssertionFailure;
 import org.hibernate.EntityMode;
@@ -236,7 +237,9 @@ public abstract class AbstractEntityPersister
 
 	private final Map uniqueKeyLoaders = new HashMap();
 	private final Map lockers = new HashMap();
-	private final Map loaders = new HashMap();
+	private UniqueEntityLoader noneLockLoader;
+	private UniqueEntityLoader readLockLoader;
+	private final Map<Object, UniqueEntityLoader> loaders = new ConcurrentHashMap<>();
 
 	// SQL strings
 	private String sqlVersionSelectString;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -4115,58 +4118,10 @@ protected void createLoaders() {
 		final Map loaders = getLoaders();
 		loaders.put( LockMode.NONE, createEntityLoader( LockMode.NONE ) );
 
-		UniqueEntityLoader readLoader = createEntityLoader( LockMode.READ );
-		loaders.put( LockMode.READ, readLoader );
+		noneLockLoader = createEntityLoader( LockMode.NONE );
```
golovnin: This is the second instance of an entity loader for LockMode.NONE. Please move this line and maybe the next line before the line 4118 and reuse the instance.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -4178,6 +4133,46 @@ protected void createLoaders() {
 		);
 	}
 
+	private UniqueEntityLoader getLoaderByLockMode(LockMode lockMode) {
+		if ( LockMode.NONE.equals( lockMode ) ) {
```
golovnin: It is common practice in Hibernate to use '==' to compare enum values for equality. You shall follow this practice here too.

gsmet:
@Sanne following your email about CHM, I'm not sure I share your concern.

It's not a large `Map` so the memory overhead is negligible.

Once the entity loaders are initialized (and that should happen quickly in the application lifecycle), the `CHM` is readonly and from the JMH benchmarks I just did (and a few others I found on the Internet), the performances of `HashMap` and `CHM` in a readonly workload are pretty similar (CHM being just a bit slower).

My JMH benchmarks results (the operation is 3 reads of different key values consumed into a blackhole to prevent JVM optimisation):
CHM.testCHM  thrpt   20  251440.305 ± 2636.864  ops/ms
CHM.testHM   thrpt   20  254533.230 ± 1114.489  ops/ms
(this is with JDK 8)

Considering the operation itself will be negligible in the execution of a query (we will have to execute the query on the datastore at some point), I think this patch makes sense and should be seriously considered. I don't think the `CHM` will be a point of contention in this case.

sebersole:
Even though the key types are mixed, there is still a discrete set for key values - its `LockMode.* + ["merge","refresh"]`.  We could make that set into its own enumeration (e.g. `EntityLoaderType`) with a mapping from the Object form (LockMode/String) to the enum form.  EnumMap should be faster than either HM or CHM.

sebersole:
I disagree.  You only *have to* synchronize this single get-or-create access pattern to the map which can pretty easily be hidden behind a single method and protected with re-entrant read-write locking.  I think that is likely to be faster.  Now, if you don't want to do it.. ok, that's fine...

As far as this problem not existing in 6, so why worry about it in 5.3... so then why are we doing these PRs at all?

gsmet:
> I disagree. You only have to synchronize this single get-or-create access pattern to the map which can pretty easily be hidden behind a single method and protected with re-entrant read-write locking. I think that is likely to be faster. Now, if you don't want to do it.. ok, that's fine...

Well, you end up synchronizing the accesses to the map as it's the only ones we do. Frankly, I'm not convinced it will perform better. If someone wants to prove me wrong, feel free to do it.

AFAIK it has never been a hotspot with the HashMap and, as explained, I don't expect it to be more of a hotspot with a CHM.

> As far as this problem not existing in 6, so why worry about it in 5.3... so then why are we doing these PRs at all?

I'm just not comfortable changing the semantic of this map considering it has been exposed to the end user.

The point of these PRs is to improve the situation in the less intrusive way possible.

gsmet:
@Sanne about your concern about it potentially being considered as some sort of memory leak, I'm not really convinced.

An application will use a very finite set of lock modes and it's not going to use more and more of them. So it will take some more time to warm up but at some point you have initialized the entity loaders for all the lock modes you will ever use.

Note that for a given lock mode and a given entity, it's not going to need a lot of memory. The issue is the combination of all the lock modes with all the entities whereas in a typical application, you will mostly use 2-3 lock modes and maybe 2 more for special cases on particular entities.

I think the whole thing of not initializing them all eagerly makes sense.

Sanne:
My concern about CHM is not related to contention (as it will be almost read-only after initialization) nor memory size of course.

I also agree it's not really a memory leak but I am concerned that some people (many are unexperienced, and even the most expert will not read the whole handbook) will see it as such.

Or worse even, they might not _see_ the whole memory requirements manifest even with pre-production load tests as while we all know that load tests need to behave exactly like production load, this is often not  the case.

Clearly we could argue that this is not our problem, but in some way it is as people are just not used to this. So... I love the general proposal but I would recommend some caution, for example introducing at least a feature flag.

In terms of CHM performance: my concern is extremely weak and should not stop you.
My only concern is that it will introduce read barriers; they are no-ops in our main architectures so I have no doubts that you will not see any difference in most cases, especially not with a micro-benchmark run on x86_64.
Still, we've seen in the past that Hibernate ORM benefits significantly from wider JVM optimisations such as heavy inlining and such barriers will prevent some optimisations from happening.
It's absolutely premature to think this would be a problem, I just meant to point out: normally ORM has a very low ration of concurrent code / normal code; the few points in which there needs to be some barrier are well known and very limited - this has been a very nice property of the project so when you introduce a new barrier ... it might not be free in the bigger picture.

So that's not to stop introducing a CHM when there's good reason, but our performance team should verify what impact and tradeoffs this might have.

In terms of verifying the impact, it might be much easier for them to keep the changes separate: it would be hard to figure out the impact of this combined with the significant memory reduction.

Sanne:
@gsmet and myself discussed some more on Hipchat.

Our conclusion:

- let's introduce a flag to disable the lazy allocations, but have this optimisation enabled by default
- ignore my thoughts about CHM impact, proper load testing will follow and we doubt this would be a problem
- this further optimisation not to be backported for now, or if we reconsider maybe to not enable it by default on maintenance branches

sebersole:
Integrated manually to add setting and minor changes

PR #2283, opened 2018-05-11
When spring.jpa.hibernate.ddl-auto use the sequence of entity fields

free212:
modify TreeMap to LinkedHashMap


vladmihalcea:
1. You need to add a Jira issue explaining why you need this change.
2. You need to add a test case proving that your code works as expected.
3. That comment `order so that property are used in the same order when binding native query` tell you that a `TreeMap` was chosen with a goal in mind. Did you run all tests and made sure they pass successfully with this change?

PR #2284, opened 2018-05-11
HHH-12565 Allow to use TYPE function on leaf subtypes of table per class hierarchies

beikov:
https://hibernate.atlassian.net/browse/HHH-12565


sebersole:
Applied by command line to fix checkstyle.  Thanks!

PR #2288, opened 2018-05-17
HHH-12589 - Add support for registering custom SQL functions when bootstrapping via JPA

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12589


coladict:
I realize I'm a year late, but isn't this much the same as `MetadataBuilderInitializer` used here: https://github.com/hibernate/hibernate-orm/blob/5.0/hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataBuilderImpl.java#L140

The one difference I see is in how it's loaded, so instead of creating a new interface for it, you could have used the old one, and added the per-configuration options, instead of just the global loading through `META-INF/services/org.hibernate.boot.spi.MetadataBuilderInitializer`.

PR #2290, opened 2018-05-19
HHH-12604 Replace EqualsHelper.equals with Objects.equals

marschall:
https://hibernate.atlassian.net/browse/HHH-12604

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/util/compare/EqualsHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/util/compare/EqualsHelper.java
@@ -16,21 +16,6 @@ import java.util.Arrays;
  */
 public final class EqualsHelper {
 
-	@SuppressWarnings("SimplifiableIfStatement")
-	public static boolean equals(final Object x, final Object y) {
-		if ( x == y ) {
-			return true;
-		}
-
-		if ( x == null || y == null ) {
-			// One is null, but the other is not (otherwise the `x == y` check would have passed).
-			// null can never equal a non-null
-			return false;
-		}
-
-		return x.equals( y );
-	}
-
 	/**
 	 * Like the legacy {@link #equals} method, but handles array-specific checks
 	 *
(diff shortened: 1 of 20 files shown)


hibernate-core/src/main/java/org/hibernate/internal/util/compare/EqualsHelper.java
```diff
@@ -16,6 +16,16 @@
...
+	 * @param x the first object, possibly null
+	 * @param y the second object, possibly null
+	 * @return true if the two objects are equal either by identity
+	 *         or using their equals method
+	 * @deprecated use {@link java.util.Objects#equals(Object, Object)}
+	 */
+	@Deprecated
 	@SuppressWarnings("SimplifiableIfStatement")
 	public static boolean equals(final Object x, final Object y) {
 		if ( x == y ) {
```
sebersole: IMO this should be changed to call `Objects#equals`

gsmet: @sebersole considering it's internal and shouldn't be used anymore in our code, shouldn't we simply remove it?

sebersole:
One or the other.  But just deprecating it misses the point imo

hibernate-core/src/main/java/org/hibernate/internal/util/compare/EqualsHelper.java
gsmet: @marschall could you simply remove this method from the code? The idea is to not use it anymore in our code and people shouldn't use our internel helpers either.

Thanks!

marschall: Sure, no problem.

gsmet:
Rebased and merged. Thanks @marschall .

FYI, we don't usually add a link to the JIRA in the commit message (we just have the reference at the beginning of the commit message). I amended the commit to remove it.

PR #2291, opened 2018-05-19
HHH-12605 Boxed variables never null

marschall:
https://hibernate.atlassian.net/browse/HHH-12605


Sanne:
that's really weird code! Thanks for fixing it.

Sanne:
merged.

PR #2292, opened 2018-05-19
HHH-12606 Container contents are never accessed

marschall:
https://hibernate.atlassian.net/browse/HHH-12606


Sanne:
merged as well. Thanks!

PR #2293, opened 2018-05-22
HHH-12612 - Fix metamodel generation with TYPE_USE based annotations.

Naros:
https://hibernate.atlassian.net/browse/HHH-12612

While this addresses issues with collections that are annotated with TYPE_USE targeted annotations, this PR is still a WIP until we've determined all TYPE_USE scenarios are covered.

### DIFF
--- a/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/util/TypeUtils.java
+++ b/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/util/TypeUtils.java
@@ -121,7 +121,14 @@ public final class TypeUtils {
 			return extractClosestRealTypeAsString( compositeUpperBound, context );
 		}
 		else {
-			return context.getTypeUtils().erasure( type ).toString();
+			final TypeMirror erasureType = context.getTypeUtils().erasure( type );
+			if ( TypeKind.ARRAY.equals( erasureType.getKind() ) ) {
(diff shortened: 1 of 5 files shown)


Naros:
Applied upstream

tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/util/TypeUtils.java
```diff
@@ -121,7 +121,14 @@ public static String extractClosestRealTypeAsString(TypeMirror type, Context con
...
 		}
 		else {
-			return context.getTypeUtils().erasure( type ).toString();
+			final TypeMirror erasureType = context.getTypeUtils().erasure( type );
+			if ( TypeKind.ARRAY.equals( erasureType.getKind() ) ) {
+				// keep old behavior here for arrays since #asElement returns null for them.
+				return erasureType.toString();
+			}
+			else {
+				return context.getTypeUtils().asElement( erasureType ).toString();
```
gsmet: @Naros why use `toString()` instead of `getQualifiedName()`? I would have expected the latter to return what we want in a more future-proof way.

Naros: @gsmet When I push the test for embeddables for our metamodel test coverage, I'll include that too.

PR #2295, opened 2018-05-23
ByteBuddy enhancement assorted fixes/improvements

gsmet:
## HHH-12614

 * https://hibernate.atlassian.net/browse/HHH-12614

This one is rather safe: it's all about keeping the protection domain when proxying classes.

## HHH-12618

 * https://hibernate.atlassian.net/browse/HHH-12618

This one is open to discussion: the idea is to use `MethodHandle` lookups when available. We have to be extra careful because the generated class has to end up in the original class package, which was not always the case before to the second commit added in this PR.

Especially, when dealing with an interface, the subclass was `Object` and thus the classes ended up in a `net.bytebuddy.renamed` package.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BasicProxyFactoryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BasicProxyFactoryImpl.java
@@ -37,6 +37,8 @@ public class BasicProxyFactoryImpl implements BasicProxyFactory {
 			throw new AssertionFailure( "attempting to build proxy without any superclass or interfaces" );
 		}
 
+		Class<?> superClassOrMainInterface = superClass != null ? superClass : interfaces[0];
+
 		Set<Class> key = new HashSet<Class>();
 		if ( superClass != null ) {
 			key.add( superClass );
@@ -55,7 +57,7 @@ public class BasicProxyFactoryImpl implements BasicProxyFactory {
 			.implement( ProxyConfiguration.class )
 			.intercept( FieldAccessor.ofField( ProxyConfiguration.INTERCEPTOR_FIELD_NAME ).withAssigner( Assigner.DEFAULT, Assigner.Typing.DYNAMIC ) )
 			.make()
-			.load( BasicProxyFactory.class.getClassLoader(), ByteBuddyState.getLoadingStrategy() )
+			.load( superClassOrMainInterface.getClassLoader(), ByteBuddyState.resolveClassLoadingStrategy( superClassOrMainInterface ) )
 			.getLoaded();
 	}
 
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BasicProxyFactoryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BasicProxyFactoryImpl.java
@@ -26,9 +26,7 @@ import net.bytebuddy.matcher.ElementMatchers;
 public class BasicProxyFactoryImpl implements BasicProxyFactory {
 
 	private static final Class[] NO_INTERFACES = new Class[0];
-	private static final NamingStrategy PROXY_NAMING = Environment.useLegacyProxyClassnames() ?
-			new NamingStrategy.SuffixingRandom( "HibernateBasicProxy$" ) :
-			new NamingStrategy.SuffixingRandom( "HibernateBasicProxy" );
+	private static final String PROXY_NAMING_SUFFIX = Environment.useLegacyProxyClassnames() ? "HibernateBasicProxy$" : "HibernateBasicProxy";
 
 	private final Class proxyClass;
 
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BytecodeProviderImpl.java
```diff
@@ -38,8 +38,8 @@
 public class BytecodeProviderImpl implements BytecodeProvider {
 
 	private static final ByteBuddyState bytebuddy = new ByteBuddyState();
-	private static final NamingStrategy.SuffixingRandom instantiatorName = new NamingStrategy.SuffixingRandom( "HibernateInstantiator" );
-	private static final NamingStrategy.SuffixingRandom optimizerName = new NamingStrategy.SuffixingRandom( "HibernateAccessOptimizer" );
+	private static final String INSTANTIATOR_PROXY_NAMING_SUFFIX = "HibernateInstantiator";
+	private static final String OPTIMIZER_PROXY_NAMING_SUFFIX = "HibernateAccessOptimizer";
```
Sanne: Just wondering if this was necessary, we allocate millions of these NamingStrategy instances so you're undoing some previous fixes.

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BasicProxyFactoryImpl.java
```diff
@@ -48,7 +46,7 @@ public BasicProxyFactoryImpl(Class superClass, Class[] interfaces, ByteBuddyStat
 		}
 
 		this.proxyClass = bytebuddy.getCurrentyByteBuddy()
-			.with( PROXY_NAMING )
+			.with( new NamingStrategy.SuffixingRandom( PROXY_NAMING_SUFFIX, new NamingStrategy.SuffixingRandom.BaseNameResolver.ForFixedValue( superClassOrMainInterface.getName() ) ) )
```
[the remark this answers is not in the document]
gsmet: It's explained in the PR description: we need to inject the classes in the package of the current class. And at the moment, if `superClass` is null, the super class end up being `Object` and thus the class is created in the `net.bytebuddy.renamed` package.

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BytecodeProviderImpl.java
gsmet: See my answer above.

I'm a bit surprised by the numbers though? Millions, really?

At the moment, the classes were created as `ReflectionOptimizer.InstantiationOptimizer` which is a bit weird if you ask me. And with the lookup method, we need the class to end up in the original class package.

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/ByteBuddyState.java
```diff
@@ -81,8 +88,41 @@ public static ByteBuddy getStaticByteBuddyInstance() {
...
+					}
+					else {
+						throw new HibernateException( LOG.bytecodeEnhancementFailed( originalClass.getName() ), exception.getCause() );
+					}
+				}
+
+				return ClassLoadingStrategy.UsingLookup.of( privateLookup );
+			}
+			catch (Throwable e) {
+				throw new HibernateException( LOG.bytecodeEnhancementFailedUnableToGetPrivateLookupFor( originalClass.getName() ), e );
```
Sanne: this new error message gives some very specific advice to users.

Yet it's catching `Throwable` .  How confident are you that this suggestion to open up the packages won't be misleading, e.g. some other error is actually happening?

I would either rephrase the error to mention that solution as one possible cause (among many) - or ideally be very specific on what we're catching.

Sanne:
Looking good.

The elephant in the room is that neither issue has tests. Do you plan to add them later, shall we create follow-up issues, or perhaps you think it's not worth it?

It's not always worth it but in that case at least some manual testing was done?

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BytecodeProviderImpl.java
Sanne: Because of a weird interaction of testing helpers and interceptors defined on the Suite, each test will trigger an enhancement attempt of many classes on the classpath - even things which have already been enhanced in a previous test run. So you have test X classes = millions.

Luis is looking at that but it's complex. For now, I just tried to make the enhancement as cheap as possible :)

Sanne:
Found a minor issue; almost funny: it still fails with entities w/o a package :)

See e.g. https://github.com/hibernate/hibernate-search/blob/26a93c13653ea70c382ad7a9bfedaba2fec8bb11/orm/src/test/java/NoPackageTest.java

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/ByteBuddyState.java
gsmet: So as for the message rationale, I followed what Rafael did on Mockito. I agree we should probably be more cautious about the advice we give.

gsmet:
> The elephant in the room is that neither issue has tests. Do you plan to add them later, shall we create follow-up issues, or perhaps you think it's not worth it?

> It's not always worth it but in that case at least some manual testing was done?

Yeah, so the protection domain issue is not easy to test as we need a signed jar. I need to do some manual testing. Hope I won't lose too much time on it...

I must admit I hoped the OP would also be able to test the fix with its code. Will try to ping him tomorrow.

For the other issue, it's tested by the whole enhancement test suite when running with JDK 9. I had some exceptions about the classes not being in the correct packages first so yeah it's working :).

> Found a minor issue; almost funny: it still fails with entities w/o a package :)

Argh. I'll check that tomorrow.

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BytecodeProviderImpl.java
gsmet: Yeah, I'm all to reduce allocations and object creations, we fought them hard in HV. Unfortunately, for now, I don't see another option.

Sanne:
Looking good. Running the build then I'll merge it.

PR #2296, opened 2018-05-23
HHH-12615 Make a few AbstractEntityPersister methods protected final

gsmet:
This is useful for OGM. I made them final so that we don't have to
account for child classes overriding them in the future.

* https://hibernate.atlassian.net/browse/HHH-12615


gsmet:
Merged, thanks for reviewing.

PR #2297, opened 2018-05-23
HHH-12620 Update JBossStandAloneJtaPlatform to use org.wildfly.transaction.client.* TM/UT

scottmarlow:
https://hibernate.atlassian.net/browse/HHH-12620
Replace Arjuna tm/ut with WildFly transaction client classes and use static getInstance() method.


scottmarlow:
I tested JBossStandAloneJtaPlatform#locateTransactionManager +  JBossStandAloneJtaPlatform#locateUserTransaction on WF13, seems to work fine now.

Naros:
It seems your change has resulted in some errors in the repository, namely:
```
org.hibernate.userguide.transactions.TransactionsTest > bmt FAILED
    org.hibernate.resource.transaction.backend.jta.internal.JtaPlatformInaccessibleException at TransactionsTest.java:171
org.hibernate.userguide.transactions.TransactionsTest > cmt FAILED
    org.hibernate.resource.transaction.backend.jta.internal.JtaPlatformInaccessibleException at TransactionsTest.java:119
```

scottmarlow:
hmm, wonder if that is any better with WildFly 13.0.0.Beta1 instead of 12.0.0.Final.

scottmarlow:
I'm not sure of what else needs to change to address the failure (failure doesn't present much more information than a JtaPlatformInaccessibleException was thrown but not why.  I tried switching ORM to use WildFly 13 but that didn't help.

scottmarlow:
I wonder if the test has access to the WildFly classes, perhaps only the Arjuna TM classes are on the test classpath.

Naros:
@scottmarlow I'm pulling down the PR and taking a peak and will let you know.

Naros:
@scottmarlow That's exactly the issue.  In `StandardJtaPlatformResolver` we check the classpath in order to resolve which JTA platform we're operating under.  It seems with this PR the classpath doesn't contain the classes you wish to use, thus it defaults to `NoJtaPlatform`.  Can you amend the PR?

Naros:
@scottmarlow Can you confirm this is the right expected library version import to use?
```
wildfly_transaction_client : 'org.wildfly.transaction:wildfly-transaction-client:1.1.0.Final
```

Naros:
So far I've only identified one issue in the ORM suite I am looking at related to this:
```
org.hibernate.jpa.test.transaction.JtaReusingEntityTransactionTest > entityTransactionShouldBeReusableTest FAILED
    org.hibernate.TransactionException at JtaReusingEntityTransactionTest.java:48
        Caused by: java.lang.UnsupportedOperationException at JtaReusingEntityTransactionTest.java:48
```
I'm still waiting on any potential issues with the other modules as they run.

scottmarlow:
@Naros @Sanne @sebersole what is the right way to allow hibernate-core to test with all of the WildFly Transaction Client dependencies covered?  Should we have a way for wildfly-core to execute a test with a launched WildFly server that has things like the new WildFly Transaction Client (transaction manager)?

Or go down the path of adding a bunch of dependencies for WildFly Transaction Client to ORM gradle/libraries.gradle?

https://github.com/wildfly/wildfly/blob/master/feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/transaction/client/main/module.xml#L34 shows the top level modules that are needed, but some of these likely bring in other modules.  I assume that we don't actually need all of them but I'm not really sure.

IMO, if we need more time to figure this out, this change could be targeted for ORM 5.3.2.  I'd like to fix this for 5.3.1, but the only users that will be impacted, are those that specify "[wildfly.jpa.jtaplatform](https://github.com/wildfly/wildfly/blob/master/jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/service/ServiceContributorImpl.java#L29
)=false" in their configuration.  I suspect this might be Spring users that set "wildfly.jpa.jtaplatform" to override the WF JPA container WildFlyCustomJtaPlatform initiator.  Or any other application that tries to use JBossStandAloneJtaPlatform.

scottmarlow:
@dreab8 please see previous comment, would like your feedback also :)

Thanks!

dreab8:
with this https://github.com/hibernate/hibernate-orm/pull/2310 all the tests pass

scottmarlow:
merged via https://github.com/hibernate/hibernate-orm/pull/2310

PR #2298, opened 2018-05-24
HHH-12622 JdbcResourceLocalTransactionCoordinatorImpl#markRollbackOnly should be ignored if there is no TX

scottmarlow:
https://hibernate.atlassian.net/browse/HHH-12622


Naros:
@scottmarlow In case you didn't see my comment on HipChat, could we get a test case for this?

scottmarlow:
Test case is now provided and for those not on hipchat, Steve pushed a change that didn't help with this issue, so my fix is still needed.

scottmarlow:
Fix is no longer needed.  Steve fixed it at a higher level.

PR #2299, opened 2018-05-24
 HHH-12579 Use the field description with type variables resolved to cast the return value of the $$_hibernate_read_entity() method

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12579

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/FieldReaderAppender.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/FieldReaderAppender.java
@@ -24,11 +24,11 @@ abstract class FieldReaderAppender implements ByteCodeAppender {
 
 	protected final TypeDescription managedCtClass;
 
-	protected final FieldDescription persistentField;
+	protected final FieldDescription persistentFieldAsDefined;
 
-	private FieldReaderAppender(TypeDescription managedCtClass, FieldDescription.InDefinedShape persistentField) {
+	private FieldReaderAppender(TypeDescription managedCtClass, FieldDescription.InDefinedShape persistentFieldAsDefined) {
 		this.managedCtClass = managedCtClass;
-		this.persistentField = persistentField;
+		this.persistentFieldAsDefined = persistentFieldAsDefined;
 	}
 
 	static ByteCodeAppender of(TypeDescription managedCtClass, FieldDescription persistentField) {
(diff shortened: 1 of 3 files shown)


gsmet:
@Sanne this one should be in 5.3.1 too.

Sanne:
I fetched it. Looks Japanese to me :) Will need to study it a bit.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/FieldReaderAppender.java
```diff
@@ -24,19 +24,19 @@
```
yrodiere: Commit message: I think you should mention the `$$_hibernate_read_<property>()` method, not the `$$_hibernate_read_entity()` method, since `entity` just happens to be the name of the property in the test case.
I might be wrong, but if so I'd like to know what calls this method exactly...

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/FieldReaderAppender.java
```diff
@@ -91,31 +93,34 @@ public Size apply(
...
 		// end if
 		methodVisitor.visitLabel( skip );
 		if ( implementationContext.getClassFileVersion().isAtLeast( ClassFileVersion.JAVA_V6 ) ) {
 			methodVisitor.visitFrame( Opcodes.F_SAME, 0, null, 0, null );
 		}
 		// return field
 		methodVisitor.visitVarInsn( Opcodes.ALOAD, 0 );
-		fieldRead( methodVisitor );
+		fieldRead( methodVisitor, persistentFieldAsDefined );
+		if ( !persistentField.getType().isPrimitive() ) {
```
yrodiere: Maybe rather `if ( persistentField.getType().asErasure().getInternalName().equals( persistentFieldAsDefined.getType().asErasure().getInternalName() ) )`?
Though I suppose it's no big deal if we introduce an unnecessary cast. Not sure if compiler optimizations are applied to this generated code.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/FieldReaderAppender.java
```diff
@@ -45,6 +45,8 @@ public Size apply(
 			MethodVisitor methodVisitor,
 			Implementation.Context implementationContext,
 			MethodDescription instrumentedMethod) {
+		FieldDescription.InDefinedShape persistentFieldAsDefined = persistentField.asDefined();
```
yrodiere: You could move this to the constructor and store this as a attribute of  `FieldReaderAppender`, alongside `persistentField`. I think you would avoid a bit of noise if you did, in particular in the `fieldRead`/`fieldWrite` methods.

gsmet:
I pushed a new version to address Yoann's comment. It should be good now.

yrodiere:
Merged, thanks!

PR #2302, opened 2018-05-24
HHH-12592 - Fix merging detached enhanced entities with orphan-removal collections

Naros:
https://hibernate.atlassian.net/browse/HHH-12592


Naros:
Applied upstream.

PR #2303, opened 2018-05-25
HHH-12592

gbadner:
Add assertions to the test to ensure that the detached and merged collection are not the same, which currently fails.

I'll try reverting the fix and seeing if I can fix it a different way.

I've added a possible fix. 

The tricky bit for this use case is that, even though the collection is lazy, the collection still gets loaded when merging the entity. This happens when the operation cascades to the collection.

The code that causes the eager load on merge is: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/loader/entity/CascadeEntityJoinWalker.java#L40-L44 .

The fix I came up with waits for the collection to be completely initialized, then set the initialized collection in the entity.

I'm not absolutely sure this is the right way to do this.

I  think it would be best if @sebersole and @barreiro both look at this.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/merge/MergeEnhancedDetachedOrphanRemovalTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/merge/MergeEnhancedDetachedOrphanRemovalTest.java
@@ -14,6 +14,7 @@ import org.hibernate.testing.bytecode.enhancement.BytecodeEnhancerRunner;
 import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
 
 import static org.hibernate.testing.transaction.TransactionUtil.doInHibernate;
+import static org.junit.Assert.assertNotSame;
 
 /**
  * @author Chris Cranford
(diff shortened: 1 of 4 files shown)


Naros:
@gbadner i had @barreiro take a look and he agreed with this in theory.  If we want to run with this I can certainly integrate it into 5.3.1 before we release.  If not, I'll open another PR to revert my original fix and we run without it.  Either way works for me at this point.

hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/merge/MergeEnhancedDetachedOrphanRemovalTest.java
```diff
@@ -37,7 +38,10 @@ public void testMergeDetachedOrphanRemoval() {
 
 		doInHibernate( this::sessionFactory, session -> {
 			entity.setName( "updated" );
-			session.merge( entity );
+			Root entityMerged = (Root) session.merge( entity );
+			assertNotSame( entity, entityMerged );
+			assertNotSame( entity, entityMerged );
```
Sanne: why would you test this twice?

gbadner: It was a cut-and-paste error. I removed it in a later commit.

Sanne:
I'll merge it. Thanks!

PR #2304, opened 2018-05-25
HHH-12584 In fact, we can provide a ReflectionOptimizer, just without the fast class instantiator

gsmet:
I was mistaken by a flawed test and the fact that the Javassist bytecode provider returns null instead of throwing an exception in my previous commit.

I noticed that while backporting the fix to the 5.2 branch.

It should be correct now.

Basically, we avoid creating a fast class instantiator in the case of an abstract class or an interface as it does not make sense. The rest of the code checks if the fast class instantiator is null and in any way shouldn't call it in the case of an abstract case (an exception is thrown in this case).

I reopened: https://hibernate.atlassian.net/browse/HHH-12584 .

I will backport this one and the first commit in 5.2.18.


Naros:
Applied upstream.

PR #2306, opened 2018-05-25
5.2

qwop:
how did it work ?

generate ddl for java object

plain java object for ddl

table name = java name, column name = java field name

The code

static class Demo {
         private String name;
         private String age;
         private String address;
     }
    public static void main(String[] args) throws Exception {

        MetadataSources metadata = new MetadataSources(
                new StandardServiceRegistryBuilder()
                        .applySetting("hibernate.dialect", org.hibernate.dialect.Oracle12cDialect.class)
                        .build());
                 metadata.addAnnotatedClass(Demo.class);
        String filename = "c:\\ddl_" + "test" + ".sql"; // dialect.getSimpleName().toLowerCase()
        MetadataImplementor metadataImplementor = (MetadataImplementor) metadata.buildMetadata();
        SchemaExport export = new SchemaExport();
        export.setHaltOnError(false)
                .setDelimiter(";")
                .setOutputFile(filename)
                .setFormat(true);

        //can change the output here`enter code here`
        EnumSet<TargetType> enumSet = EnumSet.of(TargetType.STDOUT);
//      EnumSet<TargetType> enumSet = EnumSet.of(TargetType.SCRIPT);
        export.execute(enumSet, SchemaExport.Action.CREATE, metadataImplementor);
    }


gsmet:
Doesn't look legit. If it is, please provide a proper PR based on the right branch with only your commits.

PR #2307, opened 2018-05-25
HHH-12629: PR

nikowitt:
Pull request for
https://hibernate.atlassian.net/browse/HHH-12629

Regards, Niko

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
@@ -90,7 +90,7 @@ public class DefaultLoadEventListener extends AbstractLockUpgradeEventListener i
 		doOnLoad( persister, event, loadType );
 	}
 
-	private EntityPersister getPersister( final LoadEvent event ) {
+	protected EntityPersister getPersister( final LoadEvent event ) {
 		if ( event.getInstanceToLoad() != null ) {
 			//the load() which takes an entity does not pass an entityName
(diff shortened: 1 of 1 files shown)


Sanne:
Thanks! always great to delete some dead code.

Regarding the protected methods: we'll apply this change _temporarily_ so to help you with your migration, but I can't promise they will stay protected for long (or for any other version), deal?

We'll have to figure out a better solution for your 61+ joins :)

nikowitt:
Definititely, thank you! The trouble is that we cannot change the database architecture, so I need to have a solution for this. But is there a particular reason that the methods cannot stay protected (as they used to be)? loadFromSessionCache is also still protected.

Sanne:
I'm not sure if there is a strong reason, but anything in _internal_ packages is bound to change often.

There is also some possibility that this was an important change if I look at who made it; Stale is the lead of the performance team and sometimes it implies obscure yet important optimisations - I doubt it's the case so let's apply this but I'll eventually have to confirm with him, and he might want this reverted.

Sanne:
merging

nikowitt:
Thanks for merging. Can you update me about any progress about the internal discussion?

Sanne:
FYI we don't really do internal discussions; we'll likely discuss this either on HipChat public rooms, developer mailing list, or via a new feature JIRA.

Probably best if you could start by opening a JIRA? It will need some time as half the team is away discussing ORM6 details.

nikowitt:
Sure. Created https://hibernate.atlassian.net/browse/HHH-12632.

PR #2308, opened 2018-05-25
HHH-12631 Fix invalid tracev calls in DefaultResolveNaturalIdEventListener

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12631


Sanne:
building with trace enabled... will take a while :)

PR #2309, opened 2018-05-25
HHH-12634 Make EntityPrinter more permissive regarding the parameters passed

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12634


Naros:
Applied upstream.

PR #2311, opened 2018-05-25
HHH-12633 Fix dirty tracking when previous state is unfetched

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12633 .

The issue is that `TypeHelper#findDirty()` does not take into account the fact that `previousState[i]` can also be set to `LazyPropertyInitializer.UNFETCHED_PROPERTY` while `currentState[i]` is not.

It's easily fixable but I have a question though: in this very case (`previousState[i]` not fetched and `currentState[i]` set to a value), we would always have to consider it dirty.

@barreiro interested in your feedback on this one.


barreiro:
@gsmet My understanding is the same. One would need to load the lazy attribute to check if it's dirty, but in that case it's better to just write it (even if not dirty).

Both the fix and the test LGTM

gsmet:
@barreiro thanks for the review! Merged!

PR #2312, opened 2018-05-25
HHH-12637 : Improvement to fix for HHH-12592

gbadner:
The fix for HHH-12592 will call the owning entity {{BytecodeEnhancementMetadata#isAttributeLoaded}} for a collection contained within an embeddable, using a property name that is qualified by the embeddable (e.g., grouping.things). The method simply returns {{false}} in this case, because there is no lazy property with this (qualified) name.

Depending on how HHH-10480 is fixed, Hibernate may need to be changed to call {{BytecodeEnhancementMetadata#isAttributeLoaded}} for the embeddable (not the entity) with an unqualified property name.

I made a small change to the fix for HHH-12592 so that Hibernate does not bother calling {{BytecodeEnhancementMetadata#isAttributeLoaded}}  if the collection is in an embeddable.
I will also mention this code needs to be revisited when HHH-10480 is fixed.

I also added a couple of tests involving lazy collections in embeddables that shows that things work properly currently, before HHH-10480 is fixed.


gsmet:
This one has been merged, closing the PR.

PR #2315, opened 2018-05-28
HHH-12180: Updated Javadoc

asutosh936:


Naros:
I don't agree with this change as it stands.  

While the caller can specify an entity type which will be applied to the underlying query using `#addEntity(String, String, LockMode)`; the implementation also allows the caller to specify `Tuple.class` to support tuple-based native queries.

vladmihalcea:
The `resultClass` can also be a `Tuple`, and that's why I closed the issue. Fr more details, checkout the Jira issue.

PR #2317, opened 2018-05-29
HHH-12640 Deprecate WildFlyStandAloneJtaPlatform and use JBossStandAloneJtaPlatform for loading WildFly Transaction Client or Arjuna transaction manager secondarily.

scottmarlow:


scottmarlow:
Updated description to be a bit more clear. 

Why for this pr:  WildFlyStandAloneJtaPlatform was added by me, to specifically work with WildFly deployments that bootstrap Hibernate directly.  It turned out to be the wrong fix, as JBossStandAloneJtaPlatform is already referenced by existing applications, so better to also have JBossStandAloneJtaPlatform work with the WildFly Transaction Client (or Arjuna tm for non-WildFly environments).

Since WildFlyStandAloneJtaPlatform is not really needed anymore, I deprecated it, for future removal.

Sanne:
Looking good, I'll test it later this afternoon.

Just one doubt: maybe we should keep the `WildFlyStandAloneJtaPlatform` as well and not mark it deprecated? I see no problem in allowing people to have a more precise choice.

Sanne:
proposal: replace this with pr #2370 ?

scottmarlow:
proposal looks good!

PR #2321, opened 2018-05-31
HHH-12650 - Allow tests to run on Java 9 in IntelliJ IDEA

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12650

### DIFF
--- a/gradle/java-module.gradle
+++ b/gradle/java-module.gradle
@@ -227,6 +227,13 @@ processTestResources {
 			into file( "${buildDir}/resources/test" )
 			filter( ReplaceTokens, tokens: dbBundle[db] )
 		}
+		if( new File( '${buildDir}/../out/test/' ).exists() ) {
+			copy {
+				from file( 'src/test/resources/hibernate.properties' )
+				into file( "${buildDir}/../out/test/resources" )
+				filter( ReplaceTokens, tokens: dbBundle[db] )
+			}
+		}
 	}
 }
 


vladmihalcea:
@sebersole I've uncommented my changes, then run:

    > gradlew clean tC

And when a run a `diff`, this is what I get:

```
$ diff target/resources/test/hibernate.properties out/test/resources/hibernate.properties
8,12c8,12
< hibernate.dialect org.hibernate.dialect.H2Dialect
< hibernate.connection.driver_class org.h2.Driver
< hibernate.connection.url jdbc:h2:mem:db1;DB_CLOSE_DELAY=-1;LOCK_TIMEOUT=10000
< hibernate.connection.username sa
< hibernate.connection.password
---
> hibernate.dialect @db.dialect@
> hibernate.connection.driver_class @jdbc.driver@
> hibernate.connection.url @jdbc.url@
> hibernate.connection.username @jdbc.user@
> hibernate.connection.password @jdbc.pass@
```

So, the `hibernate.properties` in the `out` folder has placeholders while the one is `target` does not.

I wonder why it does not work for me. Is there anything else I need to do?

vladmihalcea:
According to the comment for the `copyResourcesToIntelliJOutFolder` task:

>  A solution is to enable the 'After Build' Execution of the copyResourcesToIntelliJOutFolder task
>  from the 'Gradle project' IntelliJ tool window ( The task can be found under hibernate-orm > Task > other)

Does it mean we need to configure IDEA to call the `copyResourcesToIntelliJOutFolder` task?

vladmihalcea:
@dreab8 Can you confirm if we need to do some manual steps to configure IDEA in order to have this task run and copy resources to the out folder?

sebersole:
No need for @dreab8  to confirm.  That's exactly what that means.  In IntelliJ mark that task to run "After build"

sebersole:
Going to reject this.

Now if you find a way for us to tell IntelliJ to do that for us on import that would be awesome.  Something through the Gradle `idea` plugin most likely - they pick up that info and use it to customize the import.

vladmihalcea:
I'll check how we can automate it via the idea plugin then.

PR #2322, opened 2018-06-01
HHH-12396 - Fix problem with EnumType resolution on GlassFish platforms.

Naros:
https://hibernate.atlassian.net/browse/HHH-12396


Naros:
Applied upstream.

PR #2325, opened 2018-06-05
HHH-12662 - JPQL queries fail when using the Java attribute type which has an associated @Converter (only the DB column type works)

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12662

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/annotations/converter/ConverterTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/annotations/converter/ConverterTest.java
@@ -1,0 +1,182 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.test.annotations.converter;
+
+import javax.persistence.AttributeConverter;
+import javax.persistence.Convert;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+import org.hibernate.SessionFactory;
+import org.hibernate.jpa.test.BaseEntityManagerFunctionalTestCase;
+import org.hibernate.metamodel.spi.MetamodelImplementor;
+import org.hibernate.query.Query;
+import org.hibernate.type.Type;
+import org.hibernate.type.descriptor.converter.AttributeConverterTypeAdapter;
+
+import org.hibernate.testing.TestForIssue;
+import org.junit.Test;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Vlad Mihalcea
+ */
+@TestForIssue( jiraKey = "HHH-12662")
+public class ConverterTest extends BaseEntityManagerFunctionalTestCase {
+
+	@Override
+	protected Class[] getAnnotatedClasses() {
+		return new Class[] { Photo.class };
+	}
+
+	@Override
+	protected void afterEntityManagerFactoryBuilt() {
+		doInJPA( this::entityManagerFactory, entityManager -> {
+			Photo photo = new Photo();
+			photo.setId( 1 );
+			photo.setName( "Dorobanțul" );
+			photo.setCaption( new Caption( "Nicolae Grigorescu" ) );
+
+			entityManager.persist( photo );
+		} );
+	}
+
+	@Test
+	public void testJPQLUpperDbValueBindParameter() throws Exception {
+		doInJPA( this::entityManagerFactory, entityManager -> {
+			Photo photo = entityManager.createQuery(
+				"select p " +
+				"from Photo p " +
+				"where upper(caption) = upper(:caption) ", Photo.class )
+			.setParameter( "caption", "Nicolae Grigorescu" )
+			.getSingleResult();
+
+			assertEquals( "Dorobanțul", photo.getName() );
+		} );
+	}
+
+	@Test
+	public void testJPQLUpperAttributeValueBindParameterType() throws Exception {
+		doInJPA( this::entityManagerFactory, entityManager -> {
+			SessionFactory sessionFactory = entityManager.getEntityManagerFactory().unwrap( SessionFactory.class );
+			MetamodelImplementor metamodelImplementor = (MetamodelImplementor) sessionFactory.getMetamodel();
+
+			Type captionType = metamodelImplementor.entityPersister( Photo.class.getName() ).getPropertyType( "caption" );
+
+				Photo photo = (Photo) entityManager.createQuery(
+				"select p " +
+				"from Photo p " +
+				"where upper(caption) = upper(:caption) ", Photo.class )
+			.unwrap( Query.class )
+			.setParameter( "caption", new Caption("Nicolae Grigorescu"), captionType)
+			.getSingleResult();
+
+			assertEquals( "Dorobanțul", photo.getName() );
+		} );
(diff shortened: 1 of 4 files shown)


gsmet:
Hi @vladmihalcea ,

Not a proper review, sorry, but maybe the commit message should be more descriptive of what you are fixing in the commit? It's hard to say from the current one.

vladmihalcea:
The Jira issue title is not very clear indeed. I could change that first and then alter the commit message as well.

hibernate-core/src/test/java/org/hibernate/test/annotations/converter/ConverterTest.java
```diff
@@ -0,0 +1,182 @@
...
+public class ConverterTest extends BaseEntityManagerFunctionalTestCase {
+
+	@Override
+	protected Class[] getAnnotatedClasses() {
+		return new Class[] { Photo.class };
+	}
+
+	@Override
+	protected void afterEntityManagerFactoryBuilt() {
+		doInJPA( this::entityManagerFactory, entityManager -> {
```
sebersole: If you create data in a test, please make sure you delete it after

hibernate-core/src/test/java/org/hibernate/test/annotations/converter/ConverterTest.java
```diff
@@ -0,0 +1,182 @@
...
+import org.junit.Test;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Vlad Mihalcea
+ */
+@TestForIssue( jiraKey = "HHH-12662")
+public class ConverterTest extends BaseEntityManagerFunctionalTestCase {
```
sebersole: We already have a package with tons of AttributeConverter-based tests : `org.hibernate.test.converter`.  This test would be much better put there...

vladmihalcea:
@sebersole Thanks for the review. I'll investigate it further and try to fix it according to your indications.

sebersole:
@vladmihalcea Locally I added a new method to `ExpectedTypeAwareNode` to help implement this for functions...

````
	default void setExpectedType(Node otherSide) {
		if ( SqlNode.class.isInstance( otherSide ) ) {
			setExpectedType( ( (SqlNode) otherSide ).getDataType() );
		}
	}
````

That allows existing impls and hooks to work as-is, but gives `FunctionNode` the extra context it needs.  Really this should have been the signature all along for `FunctionNode`....

**Comment by sebersole** [comment]
BTW, this is incorrectly classified as a bug.  It's not.  It's an enhancement request.  From the spec:

> The persistence provider must apply any conversion methods to
instances of attribute values in path expressions used within Java Persistence query language queries or
criteria queries (such as in comparisons, bulk updates, etc.) before sending them to the database for the
query execution. When such converted attributes are used in comparison operations with literals or
parameters, the value of the literal or parameter to which they are compared must also be converted. If
the result of a Java Persistence query language query or criteria query includes one or more entity
attributes for which conversion mappings have been specified, the persistence provider must apply the
specified conversions to the corresponding values in the query result before returning them to the appli-
cation. The use of functions, including aggregates, on converted attributes is undefined.

Specifically the last sentence.  Support for this is not required for compliance.

File: hibernate-core/src/test/java/org/hibernate/test/annotations/converter/ConverterTest.java (hunk 1 above)
Comment (vladmihalcea) [reply hibernate-core/src/test/java/org/hibernate/test/annotations/converter/ConverterTest.java]: When extending the `BaseEntityManagerFunctionalTestCase`, the database schema is created before every test execution. The behavior is different than `BaseNonConfigCoreFunctionalTestCase` where we are building the `SessionFactory` in the `@BeforeClassOnce` phase.

File: hibernate-core/src/test/java/org/hibernate/test/annotations/converter/ConverterTest.java (hunk 2 above)
Comment (vladmihalcea) [reply hibernate-core/src/test/java/org/hibernate/test/annotations/converter/ConverterTest.java]: I'll move it to that package.

**Comment by vladmihalcea** [comment]
@sebersole I investigated the approach you are proposing and tried to apply it to this issue, but it will not help unless we also change the parameter binding logic.

The problem is we don't get to the query execution because the exception is thrown when we try to bind a parameter whose `Type` is not recognized:

```
org.hibernate.HibernateException: Could not determine a type for class: org.hibernate.test.converter.ConverterTest$Caption

	at org.hibernate.internal.SessionFactoryImpl.resolveParameterBindType(SessionFactoryImpl.java:1094)
	at org.hibernate.internal.SessionFactoryImpl.resolveParameterBindType(SessionFactoryImpl.java:1077)
	at org.hibernate.query.internal.QueryParameterBindingImpl.bindValue(QueryParameterBindingImpl.java:85)
	at org.hibernate.query.internal.QueryParameterBindingImpl.setBindValue(QueryParameterBindingImpl.java:57)
	at org.hibernate.query.internal.AbstractProducedQuery.setParameter(AbstractProducedQuery.java:494)
	at org.hibernate.query.internal.AbstractProducedQuery.setParameter(AbstractProducedQuery.java:107)
	at org.hibernate.test.converter.ConverterTest.lambda$testJPQLUpperAttributeValueBindParameter$3(ConverterTest.java:91)
```

Basically, upon binding the parameters we get to this:

```
TypedValue typedValue = qp.getNamedParameters().get( name );
typedValue.getType().nullSafeSet( statement, typedValue.getValue(), position, session );
```

Here, we need the `TypedValue` to handle that Java object that's handled by the `Converter`.

Therefore, I'm not sure how we could solve that in the HQL parser if the problem arises in the binding phase.

sebersole:
The parameter value/type are managed on the query via
`QueryParameterBinding` who's bind type is inferred[1] from the
`QueryParameter`.  The type of the `QueryParameter` is determined by the
query parser, which is where this ought to be fixed (if we do).

Your solution simply will not work.  It works in your "test tube".  But if
you expand you test a bit you should be able to see where it is problematic
- basically your solution assumes that the converter is applied to each and
every use of that converted class.  A test where the model uses that same
"converted class" multiple times in a mix of non-converted and multiple
converters will illustrate it.

[1] The bind type can alternatively be explicitly set when binding the
parameter - that is the Class, Type and TemporalType parts.

vladmihalcea:
I'll try it further to see why the type was not properly inferred in my implementation.

sebersole:
Simple - because "type inferring" is not visited in to function-call nodes.  The spec says it is undefined and we have never supported it.  

I think that maybe you need to take a step back here and think of this in general terms.. 

As a counter-point, when I have a query that says `... where m.email = :email` I know that if `m.email` is converted, that converter needs to be applied to the "other side" of the binary operator.  JPA requires application of the converter in this case and we have actually supported that b4 JPA.  It's actually pretty straight-forward in the binary operator case.

The problem though is how to *generically* apply type inferring across function calls?  I understand that this is a very specific case described in the Jira, but we have to define a generic algorithm that applies in all cases.  So, e.g., consider: `...where m.email = my_function(:anEmail)` how should we *generically* understand that `:anEmail` "relates" back to the converted `email` attribute?  How about `.... where m.email = my_other_function( 2, :anEmail )`, etc?

And.. if we cannot find a complete solution, I kind of think we should not implement this at all.  It is a lot easier to explain to a user how type inferring works and say simply that the inference works directly across binary operators only as opposed to trying to list out the few exceptions where inferring also happens to work such <x>, <y> and <z> functions.

vladmihalcea:
I've just debugged it through and figured out that this is not going to be easy to handle.

I'm going to close this issue and propose a new one which does not auto-register types but allows us to provide the type explicitly when setting the parameter.

sebersole:
I think this could be  cool:

````
Query q = session.createQuery( "... where upper(m.email) = upper( cast(:email as m.email) )" );
````

PR #2330, opened 2018-06-07
HHH-12348 - Hibernate ORM Document user guide, architecture of Hibernate class is out of date

aepedraza:
`documentation/testDiagram.ucls` is a temp file needed to generate the diagram. It would be deleted before approval and merging with a cleaning commit.

I don't know if there is some reason why `SVG` format is preferred, I made a `JPG` because of convenience, but it can be converted in the same cleaning commit, as you may know.

The criteria for the updated image was to state the `javax.persistence`and `org.hibernate` main interfaces with the main implementations and dependencies. Also, I wanted to show the hierarchy tree as it's the main concept in an architectural design (or at least is my point of view).

To sumarize:

- `Architecture.adoc` will only change only if the `JPG` format is accepted
- `testDiagram.ucls` will be deleted when the PR is accepted


vladmihalcea:
@aepedraza Thanks for the PR. I'm not sure what exactly is that `ucls` file, but I can easily regenerate that diagram using IntelliJ IDEA.

vladmihalcea:
Fixed by [this commit](https://github.com/hibernate/hibernate-orm/commit/0ed516e11370e07f27526000c349d06dadac2433). Thanks for spotting it.

aepedraza:
@vladmihalcea `ucls` is a file used by a Eclipse plugin to generate diagrams.

This is my first accepted PR, so let me ask: should I delete or merge `HHH-12348` from my fork and local?

vladmihalcea:
You can delete it since I regenerated the diagram and committed it.

PR #2331, opened 2018-06-07
toString - remove overly verbose output

walterjwhite:
Is toString used, if so, when do we need to see the actionQueue or persistenceContext?
TRACE?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
@@ -2316,9 +2316,13 @@ public final class SessionImpl
 		StringBuilder buf = new StringBuilder( 500 )
 				.append( "SessionImpl(" );
 		if ( !isClosed() ) {
+			if(TRACE_ENABLED) {
 			buf.append( persistenceContext )
 					.append( ";" )
 					.append( actionQueue );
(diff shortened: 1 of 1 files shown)


vladmihalcea:
For every Pull Request, we need a Jira issue. Now, I'd like to know why exactly do you need to call `toString` on the `Session`? What use case do you have that requires calling this method?

walterjwhite:
This is strictly for logging purposes, I mainly want to see that I have an instance of an entity manager and be able to differentiate it from another instance and whether the session is open or closed.  I don't want to see the action queue or context as it could be large (unless I want TRACE).

vladmihalcea:
If we apply this change, then the `toString` method will just give you the open/close status which defeats the purpose of having a meaningful `String` representation.

Now, in your case, if you are only interested in the open/close status, just log this instead:

    LOGGER.debug("Session open status: {}", session.isOpen());

Therefore, I'm going to close this issue since you can easily achieve your goal without modifying the `toString` method which might be useful for other users that really want to inspect the content of a particular `Session` from its `String` representation.

walterjwhite:
I put trace in there, to me, that is trace level information, open / close is debug level information.  Again, this isn't something I would use in production logs; however, for development, I would use it as DEBUG / TRACE.

PR #2332, opened 2018-06-07
HHH-12670 - Allows native SQL queries that take a given result Class to map the result set to the required type

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12670


vladmihalcea:
@sebersole If you have time, please take a look at this issue too. I think it's worth implementing this mechanism for native SQL queries.

gsmet:
@vladmihalcea maybe I misunderstood the change but aren't you changing how the results of a query are organized? That would break compatibility for existing applications, wouldn't it?

vladmihalcea:
@gsmet Without my change, you get an Exception. With my code, it works. It would break compatibility only if previously passing a non-entity class to a query actually worked.

gsmet:
@vladmihalcea ah ok, that makes sense :).

PR #2333, opened 2018-06-07
[HHH-12671] Fix persisting generated properties on identity insert

JWGmeligMeyling:
Relates to issue: https://hibernate.atlassian.net/browse/HHH-12671


vladmihalcea:
Applied upstream, thanks

PR #2335, opened 2018-06-08
HHH-12597 Support Hibernate Dialect for InterSystems IRIS Data Platform

InterSystemsCorporation:


vladmihalcea:
@InterSystemsCorporation Usually, we rely on our community to let us know if a given Dialects should be integrated.

Now, for our users, there is little advantage of integrating new Dialects into the hibernate-core since they already need the database client dependency which could also provide the Hibernate Dialect.

InterSystemsCorporation:
@vladmihalcea InterSystems is the same company the created the Cache71Dialect.
CACHE is our database products, and we created the existing Hibernate dialect for CACHE.
CACHE was introduced 20 years ago and there is a large community of CACHE users around the world today.
Last year we announced InterSystems IRIS, the next-generation data platform product from InterSystems.
InterSystems IRIS the the evolution of our CACHE product. It leverages much of the database capability of our CACHE product. It is targeted to both new customers and existing CACHE customers.
The InterSystems IRIS dialect for Hibernate allows existing users of the CACHE dialect to continue to use Hibernate when they migrate to InterSystems IRIS. It also provides Hibernate access to new InterSystems IRIS users.
So while InterSystems IRIS can be thought of as a "new" dialect, it can also be thought as the next-generation of the CACHE dialect.

sebersole:
@InterSystemsCorporation Yes, I have worked with Jonathan over the years on the CACHE dialect.

What Vlad is explaining is the unnecessary explosion of Dialect classes we maintain in Hibernate source code.  For really no reason.  The normal argument is that because including it somehow makes it more "supported" or more "easily accessible".  But neither of those is actually correct.  We actually do not "support" the CACHE dialect - we do a best effort support, generally delegating stuff to y'all/Jonathan.  And it is not any more easily accessible than if the Dialect were in your CACHE/IRIS diallect.

Since this sounds like essentially an update of CACHE to IRIS, we could simply drop the CACHE dialect for this new one.  Would that work?

InterSystemsCorporation:
We really would like to move this PR forward. Our users do prefer to include the dialect in the code base, otherwise the experience has proven to be suboptimal in the past. What exactly does this vote look like, how long will it run and what are the criteria to include the dialect? Thanks, SW

JWGmeligMeyling:
Hi @InterSystemsCorporation, just my two cents here, but frankly there hasn't been any user interest on either this PR or in the mailing list, so I don't think there's a point in adding this dialect regardless of whether the old one gets removed.

Imo the dialects provided by Hibernate should be supported to some extent and preferable included in the test suite, which is something thats just not sensible to do for the less frequently used DBMSes. So I think this is best shipped as a separate artefact that is maintained outside of Hibernate. I also don't really see what burden that would be to users.

sebersole:
Dealing with Jonathan for Cache dialect stuff was always super positive experience so, for me personally, I am inclined to do the swap - replace the Cache[XY]Diallect with this new one.

For sure, not both dialects.

sebersole:
@JWGmeligMeyling Generally speaking I agree with everything you said 100%.  

Intersystems is just a bit different.  They are very active and involved in testing, maintaining and improving their dialect.  And they do host a CI job running the Hibernate test suite against their database.

vladmihalcea:
+1 for replacing the old Dialect with this new one. But I guess that can be done in 6.x. We could deprecate the old one now.

quanyinrc:
Hi all,
I'm a developer at ReadyComputing, and we use Hibernate on the IRIS data platform as the persistence for one of our applications. I vote for the IRIS dialect to be included in Hibernate.
We use Maven to build our Java application, and the IRIS dialect is not in the Maven repository, so we had to attach the jar locally to our project, and reference it via <systemPath> in the pom.xml.
By the way, according to our testing, the old Cache71Dialect does not work with Hibernate 5. Having a built-in dialect support for the new IRIS data platform in Hibernate would be greatly helpful to us.
Thank you all.

rajrsingh:
@sebersole and @vladmihalcea I'm now managing this project at InterSystems, and I'd like to move forward with replacing the old Dialect (CACHE) with this new one. Since last year we of course have updates to make to this PR but before we do that, I'd just like to make sure the offer still stands. I appreciate your attention to this and we'll make it as easy as possible.

beikov:
We are discussing the way forward for new Dialects in https://github.com/hibernate/hibernate-orm/discussions/3783

beikov:
As per https://hibernate.atlassian.net/browse/HHH-14693 and https://github.com/hibernate/hibernate-orm/discussions/3783 we decided not to include any dialects in 5.5 anymore for now and focus on 6.0.

IMO it would be fine to add it to 5.5 though if the Dialect for 6 qualifies for hibernate-core according to the requirements document: https://github.com/hibernate/hibernate-orm/blob/wip/6.0/dialects.adoc

PR #2336, opened 2018-06-08
HHH-12662 - JPQL queries fail when using the Java attribute type which has an associated AttributeConverter (only the DB column type works)

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12662

### DIFF
--- a/documentation/src/main/asciidoc/userguide/chapters/domain/basic_types.adoc
+++ b/documentation/src/main/asciidoc/userguide/chapters/domain/basic_types.adoc
@@ -3,6 +3,7 @@
 :modeldir: ../../../../../main/java/org/hibernate/userguide/model
 :sourcedir: ../../../../../test/java/org/hibernate/userguide/mapping
 :resourcedir: ../../../../../test/resources/org/hibernate/userguide/
+:converter-sourcedir: ../../../../../../../hibernate-core/src/test/java/org/hibernate/test/converter
 :extrasdir: extras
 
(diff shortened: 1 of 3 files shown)


vladmihalcea:
@sebersole I changed the PR so that now we don't auto-register anything but we allow the user o supply the type. The change is minimal and the user can supply the type explicitly like this:

```
SessionFactory sessionFactory = entityManager.getEntityManagerFactory().unwrap( SessionFactory.class );
MetamodelImplementor metamodelImplementor = (MetamodelImplementor) sessionFactory.getMetamodel();
Type captionType = metamodelImplementor.entityPersister( Photo.class.getName() ).getPropertyType( "caption" );

Photo photo = (Photo) entityManager.createQuery(
	"select p " +
	"from Photo p " +
	"where upper(caption) = upper(:caption) ", Photo.class )
.unwrap( Query.class )
.setParameter( "caption", new Caption("Nicolae Grigorescu"), captionType)
.getSingleResult();
```

sebersole:
What you describe is basically exactly what 6.0 does - it defines "types"
in terms of "any part of the domain model" whether that be a class
(Caption), an attribute (Photo#caption), etc.

It's where the `cast( :param as p.caption)` comes from too

vladmihalcea:
Then it will be a good transition to 6.0. I think the current way of resolving the right converter Type is a little bit verbose and doing it with just one `cast` in 6.0 is much better.

PR #2337, opened 2018-06-10
4.3 When executing the saveOrUpdate method, it can be executed successfully but there is an error log print prompt null pointer

jidianzhixia:
Don't know what this is


vladmihalcea:
Please use the forum if you have questions like this one:

 https://discourse.hibernate.org

PR #2338, opened 2018-06-10
HHH-12692 - Modify SessionImpl.toString to be quiet by default and verbose when trace is enabled

walterjwhite:
Change default behavior to show instance id and whether it is open/closed unless TRACE_ENABLED is true.


walterjwhite:
I created jira: https://hibernate.atlassian.net/browse/HHH-12692

Sanne:
Looks good! I rebased it, fixed the style a bit and applied. Thanks!

PR #2339, opened 2018-06-10
HHH-12675 - Respect inverse property for JoinedSubclassEntityPersister

JWGmeligMeyling:
See: https://hibernate.atlassian.net/browse/HHH-12675

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
@@ -101,6 +101,8 @@ public class JoinedSubclassEntityPersister extends AbstractEntityPersister {
 
 	private final boolean[] subclassTableSequentialSelect;
 	private final boolean[] subclassTableIsLazyClosure;
+	private final boolean[] isInverseSubclassTable;
+	private final boolean[] isNullableSubclassTable;
 
 	// subclass discrimination works by assigning particular
 	// values to certain combinations of null primary key
@@ -123,6 +125,7 @@ public class JoinedSubclassEntityPersister extends AbstractEntityPersister {
 	private final int coreTableSpan;
 	// only contains values for SecondaryTables, ie. not tables part of the "coreTableSpan"
 	private final boolean[] isNullableTable;
+	private final boolean[] isInverseTable;
 
 	//INITIALIZATION:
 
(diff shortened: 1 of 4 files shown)


gsmet:
Hi @JWGmeligMeyling ,

Looks very nice from a quick look. Could you separate the new tests instead of inserting it into an existing one? It will be easier to debug and see what's going on with your test case.

Thanks!

JWGmeligMeyling:
@gsmet 

I can do that, but a complete separation of the test cases would require the introduction of new `Pool` and `SwimmingPool` test entities which would be completely duplicates of the existing ones besides the `inverse` property. This is because one superclass has to map the secondary table with `inverse=true`, and the other one with `inverse=false`. Keeping the mappings as is (both mappings on one superclass type) would not really simplify debugging, as the EntityPersister will still have to deal with both cases separately anyway. I can duplicate the test domain classes as well, but this will result in 3 mostly duplicate classes.

So do you just want to split out a separate test case operating on the same entities (which has to be an entity with two secondary tables, one of which is mapped as inverse table)? This would make the test easier to read, show where its different from the other scenario, and provide better test adequacy (as the non-inverse and inverse test pass or fail separately). But I am not sure how this would be easier to debug, as the EntityPersister still has to handle both cases. Or should I just copy and alter all related test classes instead?

JWGmeligMeyling:
I've updated the PR so that it has two separate test cases.

JWGmeligMeyling:
@gsmet  Any reason why this is removed from the milestone? Are there any blockers?

My current workaround requires no-op queries to be executed for every update, which is really quite nasty:

```java
@org.hibernate.annotations.Table(appliesTo = "joined_view", inverse = true,
    // Ugly, but required, as inverse=true is broken... See: https://hibernate.atlassian.net/browse/HHH-12675
    sqlInsert = @SQLInsert(sql = "INSERT INTO noop_table (id) SELECT ? WHERE FALSE", check = ResultCheckStyle.NONE ),
    sqlUpdate = @SQLUpdate(sql = "UPDATE noop_table SET id = ? WHERE FALSE", check = ResultCheckStyle.NONE),
    sqlDelete = @SQLDelete(sql = "DELETE FROM noop_table WHERE discount_code_id = ? AND FALSE", check = ResultCheckStyle.NONE)
)
```

gsmet:
@JWGmeligMeyling the rationale was mostly that I found it a bit too risky for a corner case which was probably an issue all along. It seems more reasonable to commit it to 5.4.

JWGmeligMeyling:
@gsmet Any plans to commit this to 5.4 now masters on 5.4?

PR #2341, opened 2018-06-10
proposed fix for HHH-9403

cristhiank:
Adds check for WildcardType when resolving a generic type attribute
Adds test case

Signed-off-by: cristhiank <calovi86@gmail.com>


vladmihalcea:
Applied upstream, thanks

PR #2342, opened 2018-06-11
 HHH-12676 - Initial draft for SecondaryTable support on UnionSubclassPersister

JWGmeligMeyling:
Initial draft for supporting `SecondaryTables` on the `UnionSubclassPersister`. I have had to change quite a lot of the persister, as the assumption of a single table throughout the class turned out to be plain wrong. I could reuse quite some parts from the `SingleTableClassPersister` and have decided to reuse most of the code (similarly to what is done in the `JoinedTableClassPersister`) and not abstract the code to another layer.

I'd like to point out that this PR is an initial draft and I mostly seek some early feedback. There are still a couple of things that probably should be addressed, such as whether the secondary tables (or any other join tables, if these exist?) should be added to the query spaces.

Jira ticket: https://hibernate.atlassian.net/browse/HHH-12676

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/UnionSubclassEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/UnionSubclassEntityPersister.java
@@ -15,9 +15,7 @@ import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
 
-import org.hibernate.AssertionFailure;
 import org.hibernate.HibernateException;
-import org.hibernate.LockMode;
 import org.hibernate.MappingException;
 import org.hibernate.boot.model.relational.Database;
 import org.hibernate.cache.spi.access.EntityDataAccess;
@@ -35,12 +33,14 @@ import org.hibernate.internal.util.collections.ArrayHelper;
 import org.hibernate.internal.util.collections.JoinedIterator;
 import org.hibernate.internal.util.collections.SingletonIterator;
 import org.hibernate.mapping.Column;
+import org.hibernate.mapping.Join;
 import org.hibernate.mapping.PersistentClass;
+import org.hibernate.mapping.Property;
+import org.hibernate.mapping.Selectable;
 import org.hibernate.mapping.Subclass;
 import org.hibernate.mapping.Table;
 import org.hibernate.persister.spi.PersisterCreationContext;
 import org.hibernate.sql.SelectFragment;
-import org.hibernate.sql.SimpleSelect;
 import org.hibernate.type.StandardBasicTypes;
 import org.hibernate.type.Type;
 
@@ -55,6 +55,12 @@ public class UnionSubclassEntityPersister extends AbstractEntityPersister {
 	// the class hierarchy structure
 	private final String subquery;
 	private final String tableName;
+	private final int tableSpan;
+	private final String[] tableNames;
+	private final boolean[] isInverseTable;
+	private final boolean[] isNullableTable;
+	private final String[][] keyColumnNames;
+	private final boolean[] cascadeDeleteEnabled;
 	//private final String rootTableName;
 	private final String[] subclassClosure;
 	private final String[] spaces;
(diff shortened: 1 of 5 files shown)


JWGmeligMeyling:
@vladmihalcea What is the clarification required here?

Prior to this change `@SecondaryTable(s)` do not work on the `UnionSubclassPersister`, afterwards they do, although I am not sure whether I've covered all edge cases (such as caching, sequential select and table spaces). The problem is that I had to change as much as 500 lines of code in parts that haven't been touched in at least 11 years 😇 

I'd love to get some early feedback as on what needs to be added or tested more thoroughly before we can merge this.

sebersole:
I'll take a look.  And yes, implementing one of the central contracts in all of Hibernate takes a lot of work.  Imagine :P

Anyway, I wanted to say before I took a peek that this really is a complicated contract to implement and because of that, kudos for trying your hand at it.

sebersole:
Well, as you said this is all mainly copied/pasted and/or reused code, so it all seems reasonable :)

It is the best it will get with how persisters are defined and used, which we have re-worked dramatically in 6

JWGmeligMeyling:
Yes, I saw on the roadmap as well as on the mailing list that entity persisters are quite heavily reworked in 6. Is Hibernate 6.0 already in development? Is there a branch or repository for 6.0?

sebersole:
It is under very active development :
https://github.com/sebersole/hibernate-core/tree/wip/6.0/

For entity "persisters" we have only worked on no/discrim inheritance thus
far

hibernate-core/src/main/java/org/hibernate/persister/entity/UnionSubclassEntityPersister.java
```diff
@@ -125,19 +169,143 @@ public UnionSubclassEntityPersister(
...
+		}
+
+		subclassTableSequentialSelect = ArrayHelper.toBooleanArray( isDeferreds );
+		subclassTableNameClosure = ArrayHelper.toStringArray( subclassTableClosure );
+		subclassTableIsLazyClosure = ArrayHelper.toBooleanArray( isLazies );
+		subclassTableKeyColumnClosure = ArrayHelper.to2DStringArray( joinKeyColumns );
+		isClassOrSuperclassTable = ArrayHelper.toBooleanArray( isConcretes );
+		isInverseSubclassTable = ArrayHelper.toBooleanArray( isInverses );
+		isNullableSubclassTable = ArrayHelper.toBooleanArray( isNullables );
+//		hasSequentialSelects = hasDeferred; TODO
```
JWGmeligMeyling: I didn't implement sequential selects here, but I'm also unsure how it should be implemented, and whether we even technically can support sequential selects for this type of persistence.

hibernate-core/src/main/java/org/hibernate/persister/entity/UnionSubclassEntityPersister.java
```diff
@@ -246,61 +404,27 @@ public String getSubclassForDiscriminatorValue(Object value) {
 		return spaces;
 	}
 
-	protected boolean isDiscriminatorFormula() {
```
JWGmeligMeyling: This method was removed because it was unused.

hibernate-core/src/main/java/org/hibernate/persister/entity/UnionSubclassEntityPersister.java
```diff
@@ -246,61 +404,27 @@ public String getSubclassForDiscriminatorValue(Object value) {
...
-		return false;
-	}
-
-	/**
-	 * Generate the SQL that selects a row by id
-	 */
-	protected String generateSelectString(LockMode lockMode) {
```
JWGmeligMeyling: This method was removed because it was unused.

JWGmeligMeyling:
I could spend some time on reapplying these changes on 6.0. I actually had a brief look at it already but I'm still figuring out what should go where in the new API. Besides that, are there any chances to include this for 5.x?

PR #2343, opened 2018-06-11
Fix test org.hibernate.test.procedure.HANAStoredProcedureTest

breglerj:
- The addition of a field to one of the function's underlying entities caused
  the test to fail because the field wasn't part of the function's return value
  and therefore the entity couldn't be fully hydrated.
- I also added the new test introduced with #2326


vladmihalcea:
Applied upstream, thanks.

PR #2345, opened 2018-06-12
HHH-12685 ClassCastException when checking if parameter isBound in criteria query

cristhiank:
https://hibernate.atlassian.net/browse/HHH-12685

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/compile/CriteriaQueryTypeQueryAdapter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/compile/CriteriaQueryTypeQueryAdapter.java
@@ -362,7 +362,7 @@ public class CriteriaQueryTypeQueryAdapter<X> implements QueryImplementor<X> {
 			return (ExplicitParameterInfo) param;
 		}
 		else if ( ParameterExpression.class.isInstance( param ) ) {
-			return explicitParameterInfoMap.get( (ParameterExpression) param );
+			return explicitParameterInfoMap.get( param );
 		}
 		else {
 			for ( ExplicitParameterInfo parameterInfo : explicitParameterInfoMap.values() ) {
@@ -501,23 +501,27 @@ public class CriteriaQueryTypeQueryAdapter<X> implements QueryImplementor<X> {
 
 	@Override
 	public QueryImplementor<X> setEntity(String name, Object val) {
+		entityManager.checkOpen( false );
 		ExplicitParameterInfo parameterInfo = locateParameterByName( name );
-		parameterInfo.validateDateBind();
+		parameterInfo.validateBindValue( val );
 		jpqlQuery.setEntity( name, val );
 		return this;
 	}
 
 	@Override
 	public QueryImplementor<X> setParameter(String name, Object val, Type type) {
+		entityManager.checkOpen( false );
 		ExplicitParameterInfo parameterInfo = locateParameterByName( name );
-		parameterInfo.validateDateBind();
+		parameterInfo.validateBindValue( val );
 		jpqlQuery.setParameter( name, val, type );
 		return this;
 	}
 
 	@Override
 	public <T> QueryImplementor<X> setParameter(QueryParameter<T> parameter, T val) {
+		entityManager.checkOpen( false );
 		final ExplicitParameterInfo parameterInfo = resolveParameterInfo( parameter );
+		parameterInfo.validateBindValue( val );
 		if ( parameterInfo.isNamed() ) {
 			jpqlQuery.setParameter( parameterInfo.getName(), val );
 		}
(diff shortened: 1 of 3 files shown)


cristhiank:
@vladmihalcea I need your advice here... 
This bug is due to the casting in [this line](https://github.com/hibernate/hibernate-orm/blob/7938918518aca96cd3844e1119fedd006397f0c2/hibernate-core/src/main/java/org/hibernate/query/internal/AbstractProducedQuery.java#L750)
The criteria query invokes this method with an instance of [ExplicitParameterInfo](https://github.com/hibernate/hibernate-orm/blob/5.3.1/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/compile/ExplicitParameterInfo.java) that doesn't inherit from [QueryParameter<T>](https://github.com/hibernate/hibernate-orm/blob/5.3.1/hibernate-core/src/main/java/org/hibernate/query/QueryParameter.java)
My fix idea is to make ExplicitParameterInfo to inherit from QueryParameter but it's not working because [CriteriaQueryTypeQueryAdapter](https://github.com/hibernate/hibernate-orm/blob/5.3.1/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/compile/CriteriaQueryTypeQueryAdapter.java) returns a parameter that is not included in its [parameterMetadata](https://github.com/hibernate/hibernate-orm/blob/7430cdafcf520d6d7d40d362fdc42c6a5d8b141e/hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingsImpl.java#L167)

cristhiank:
Maybe @sebersole can give me a tip to solve this too.. Thanks

gsmet:
@cristhiank I pushed one possible fix to your branch. I don't think it was a good idea to expose `ExplicitParameterInfo` in the first place.

@sebersole will be able to say if it is the proper fix or not.

vladmihalcea:
@cristhiank @gsmet DO you think we should close this PR, or do you want to have the test integrated just for making sure that this problem won't ever happen?

gsmet:
@vladmihalcea why would we close it? We are waiting for @sebersole to take a look at the fix I pushed as I'm not totally sure it's the right one.

vladmihalcea:
In that case, let's wait for Steve's approval.

cristhiank:
I hope this fix can be included in the 5.3.2 release.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/compile/CriteriaQueryTypeQueryAdapter.java
```diff
@@ -443,21 +443,13 @@ private ExplicitParameterInfo locateParameterByPosition(int position) {
...
 	public <T> Parameter<T> getParameter(String name, Class<T> type) {
 		entityManager.checkOpen( false );
-		Parameter parameter = locateParameterByName( name );
-		if ( type.isAssignableFrom( parameter.getParameterType() ) ) {
-			return parameter;
-		}
-		throw new IllegalArgumentException(
-				"Named parameter [" + name + "] type is not assignanle to request type ["
-						+ type.getName() + "]"
-		);
```
dreab8: I think the type assignable check is necessary, base on the javax.persistence.Query javadoc ```
/**
     * Get the parameter object corresponding to the declared
     * parameter of the given name and type.
     * This method is required to be supported for criteria queries
     * only.
     * @param name  parameter name
     * @param type  type
     * @return parameter object
     * @throws IllegalArgumentException if the parameter of the
     *         specified name does not exist or is not assignable
     *         to the type
     * @throws IllegalStateException if invoked on a native
     *         query or Java Persistence query language query when
     *         the implementation does not support this use
     * @since Java Persistence 2.0
     */
    <T> Parameter<T> getParameter(String name, Class<T> type);
```
 an IllegalArgumentException has to be thrown if the parameter of the specified name does not exist or is not assignable to the type.
The  removed `locateParameterByName( name )` was throwing the exception in case the specified name does not exists.

cristhiank: Yes, I agree with you. It is still making this check, the only difference is that, it is delegated to the underlying jpqlQuery object `return jpqlQuery.getParameter( name, type );`

gsmet:
@dreab8 as mentioned by @cristhiank I think your comment is taken care of, it's just that it's delegated to the `jqplQuery`.

I rebased and squashed @cristhiank 's commit into one.

gsmet:
@cristhiank just to let you know the progress on this issue: we ran the JPA TCK (the test compatibility kit for the JPA specification) and it had one failure. We decided to fix another way.

We'll rerun the TCK and hopefully, it will pass this time and we will be able to integrate it in 5.3.2.

gsmet:
Merged in master.

@cristhiank thanks for preparing the test case
@dreab8 thanks for your help!

dreab8:
@gsmet thanks to you and to @cristhiank :+1:

PR #2348, opened 2018-06-14
HHH-12651 : org.hibernate.Session.*Query(Ljava/lang/String) methods r…

gbadner:
…eturn different types in 5.1 and 5.3

https://hibernate.atlassian.net/browse/HHH-12651


gbadner:
Merged.

PR #2352, opened 2018-06-18
HHH-12696 - Allow map attribute key and value subgraphs to both be added to entity / sub graphs.

aleksandarsusnjar:
We hit [HHH-12696](https://hibernate.atlassian.net/browse/HHH-12696) and decided that I can contribute the fix back. This is my first contribution back to Hibernate but, I hope, I covered everything:

1. I made minimal changes and kept the old behaviour still available to the maximum extent possible, as far as I am aware.
2. Added more JavaDoc than there was before.
3. Created unit tests. Confirmed that they fail before and succeed after the change. (No preexisting tests fail before or after.)

This is in the same area as [HHH-10378](https://hibernate.atlassian.net/browse/HHH-10378), which is of interest to me too and could fix it as well.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/graph/spi/AttributeNodeImplementor.java
+++ b/hibernate-core/src/main/java/org/hibernate/graph/spi/AttributeNodeImplementor.java
@@ -9,10 +9,17 @@ package org.hibernate.graph.spi;
 import javax.persistence.AttributeNode;
 import javax.persistence.metamodel.Attribute;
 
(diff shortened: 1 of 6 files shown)


aleksandarsusnjar:
Actually I just figured out that SubgraphImpl is an internal class and I exposed it in the AttributeNodeImplementor. In 4.3 this creates an even bigger problem (cyclic dependency between -core and -entitymanager). Trying to figure out a better way.

aleksandarsusnjar:
Created a new one.

PR #2354, opened 2018-06-19
HHH-12696 (master) - Allow map attribute key and value subgraphs to both be added to entity / sub graphs.

aleksandarsusnjar:
We hit [HHH-12696](https://hibernate.atlassian.net/browse/HHH-12696) and decided that I can contribute the fix back. This is my first contribution back to Hibernate but, I hope, I covered everything:

I made minimal changes and kept the old behaviour still available to the maximum extent possible, as far as I am aware.
Added more JavaDoc than there was before.
Created unit tests. Confirmed that they fail before and succeed after the change. (No preexisting tests fail before or after.)
This is in the same area as [HHH-10378,](https://hibernate.atlassian.net/browse/HHH-10378) which is of interest to me too and could fix it as well.

I also submitted the backport to 4.3 as a [pull request 2355](https://github.com/hibernate/hibernate-orm/pull/2355).

PR #2355, opened 2018-06-19
HHH-12696 - Allow map attribute key and value subgraphs to both be added to entity / sub graphs.

aleksandarsusnjar:
Backport of:
[pull request 2354 for master](https://github.com/hibernate/hibernate-orm/pull/2354). Description from there:

We hit HHH-12696 and decided that I can contribute the fix back. This is my first contribution back to Hibernate but, I hope, I covered everything:

I made minimal changes and kept the old behaviour still available to the maximum extent possible, as far as I am aware.
Added more JavaDoc than there was before.
Created unit tests. Confirmed that they fail before and succeed after the change. (No preexisting tests fail before or after.)
This is in the same area as HHH-10378, which is of interest to me too and could fix it as well.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/graph/spi/AttributeNodeImplementor.java
+++ b/hibernate-core/src/main/java/org/hibernate/graph/spi/AttributeNodeImplementor.java
@@ -24,6 +24,7 @@
 package org.hibernate.graph.spi;
 
 import javax.persistence.AttributeNode;
+import javax.persistence.Subgraph;
 import javax.persistence.metamodel.Attribute;
 
 /**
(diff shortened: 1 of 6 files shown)


aleksandarsusnjar:
Notes on building:

1. Had to use JDK 7.
2. Had to rebuild hibernate-matrix-testing-1.0.0-SNAPSHOT.jar locally with JDK 7 as the published version uses Java 8 bytecode incompatible with JDK 7.

aleksandarsusnjar:
The build failed with similar errors I experienced locally (:hibernate-core:runAnnotationProcessors). I cannot reproduce the error any more, but I did have to do what I described above (JDK7 + hibernate-matrix-testing build). Additionally, actual error details only show up if gradle is invoked with --debug (--info is not enough).

PR #2357, opened 2018-06-20
HHH-10661 - serialize method in StatefulPersistenceContext has duplicate code.

aepedraza:
I added a new Interface with a `serialize()` method in order to be able to add a private helper method in `StatefulPersistenceContext`. This `writeKeyMapToStream()` method is doing the same duplicated routine, so no business logic was changed.
Writing this I found `private static boolean TRACE_ENABLED = LOG.isTraceEnabled()`, which is the same as `boolean tracing` used in `StatefulPersistenceContext.serialize()` so I'll be improving that also ASAP.
**Just a little warning:** `StatefulPersistenceContext` doesn't have a test class and I don't know yet how to test the whole project, so any help with this kind of checks is appreciated, please contact me.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
@@ -1436,25 +1436,24 @@ public class StatefulPersistenceContext implements PersistenceContext {
 	 * @throws IOException serialization errors.
 	 */
 	public void serialize(ObjectOutputStream oos) throws IOException {
-		final boolean tracing = LOG.isTraceEnabled();
-		if ( tracing ) {
+		if ( TRACE_ENABLED ) {
 			LOG.trace( "Serializing persistence-context" );
 		}
 
 		oos.writeBoolean( defaultReadOnly );
 		oos.writeBoolean( hasNonReadOnlyEntities );
 
-		writeKeyMapToStream( entitiesByKey, oos, tracing, "entitiesByKey" );
-		writeKeyMapToStream( entitiesByUniqueKey, oos, tracing, "entitiesByUniqueKey" );
-		writeKeyMapToStream( proxiesByKey, oos, tracing, "proxiesByKey" );
-		writeKeyMapToStream( entitySnapshotsByKey, oos, tracing, "entitySnapshotsByKey" );
+		writeKeyMapToStream( entitiesByKey, oos, "entitiesByKey" );
+		writeKeyMapToStream( entitiesByUniqueKey, oos, "entitiesByUniqueKey" );
+		writeKeyMapToStream( proxiesByKey, oos, "proxiesByKey" );
+		writeKeyMapToStream( entitySnapshotsByKey, oos, "entitySnapshotsByKey" );
 
 		entityEntryContext.serialize( oos );
 
-		writeKeyMapToStream( collectionsByKey, oos, tracing, "collectionsByKey" );
+		writeKeyMapToStream( collectionsByKey, oos, "collectionsByKey" );
 
 		oos.writeInt( collectionEntries.size() );
-		if ( tracing ) {
+		if ( TRACE_ENABLED ) {
 			LOG.trace( "Starting serialization of [" + collectionEntries.size() + "] collectionEntries entries" );
 		}
 		for ( Map.Entry<PersistentCollection,CollectionEntry> entry : collectionEntries.entrySet() ) {
@@ -1463,7 +1462,7 @@ public class StatefulPersistenceContext implements PersistenceContext {
 		}
 
 		oos.writeInt( arrayHolders.size() );
-		if ( tracing ) {
+		if ( TRACE_ENABLED ) {
 			LOG.trace( "Starting serialization of [" + arrayHolders.size() + "] arrayHolders entries" );
 		}
 		for ( Map.Entry<Object,PersistentCollection> entry : arrayHolders.entrySet() ) {
@@ -1472,7 +1471,7 @@ public class StatefulPersistenceContext implements PersistenceContext {
 		}
 
 		oos.writeInt( nullifiableEntityKeys.size() );
-		if ( tracing ) {
+		if ( TRACE_ENABLED ) {
 			LOG.trace( "Starting serialization of [" + nullifiableEntityKeys.size() + "] nullifiableEntityKey entries" );
 		}
 		for ( EntityKey entry : nullifiableEntityKeys ) {
@@ -1490,11 +1489,10 @@ public class StatefulPersistenceContext implements PersistenceContext {
 	 *
 	 * @throws IOException Thrown by Java I/O
 	 */
-	// TODO improve parameters: is tracing and keysName needed?
-	private void writeKeyMapToStream(Map<? extends SerializableKey, ? extends Object> keyMap, ObjectOutputStream oos, boolean tracing, String keysName)
+	private void writeKeyMapToStream(Map<? extends SerializableKey, ? extends Object> keyMap, ObjectOutputStream oos, String keysName)
 			throws IOException {
 		oos.writeInt( keyMap.size() );
-		if ( tracing ) {
+		if ( TRACE_ENABLED ) {
 			LOG.trace( "Starting serialization of [" + keyMap.size() + "] " + keysName + " entries" );
 		}
 		for ( Entry<? extends SerializableKey, ? extends Object> entry : keyMap.entrySet() ) {
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
@@ -603,7 +603,7 @@ public class StatefulPersistenceContext implements PersistenceContext {
 		if ( li.getSession() != this.getSession() ) {
 			final EntityPersister persister = session.getFactory().getMetamodel().entityPersister( li.getEntityName() );
 			final EntityKey key = session.generateEntityKey( li.getIdentifier(), persister );
-		  	// any earlier proxy takes precedence
+			// any earlier proxy takes precedence
 			proxiesByKey.putIfAbsent( key, proxy );
 			proxy.getHibernateLazyInitializer().setSession( session );
 		}
@@ -1137,7 +1137,7 @@ public class StatefulPersistenceContext implements PersistenceContext {
 		final EntityPersister persister = session.getFactory().getMetamodel().entityPersister( entityName );
 		final CollectionPersister collectionPersister = session.getFactory().getMetamodel().collectionPersister( collectionRole );
 
-	    // try cache lookup first
+		// try cache lookup first
 		final Object parent = parentsByChild.get( childEntity );
 		if ( parent != null ) {
 			final EntityEntry entityEntry = entityEntryContext.getEntityEntry( parent );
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
```diff
@@ -1800,13 +1799,13 @@ private void managedSharedCacheEntries(
 							naturalIdCacheKey,
 							id,
 							null
-					);
+							);
 
 					if ( put && factory.getStatistics().isStatisticsEnabled() ) {
 						factory.getStatistics().naturalIdCachePut(
```
Naros: @aepedraza I'm not sure i necessarily agree with a number of your formatting changes you've made here.  For example, if you compare the format of this method call with its older equivalent, you'll notice that the the closing `)` on the last line is left-aligned with the word `factory`.  This is generally how we generally format multiple lines with a closing brace on a separate line.  

Such cases like this have generated quite a number of diffs that I believe are unnecessary.

aepedraza: Hi @Naros. I'm not sure about those formatting also. In fact, I disagree with some of them, specially the one you're pointing about closing brace in a new line.
I accepted them because I assume they were added due to the hibernate formatter for Eclipse, but I can review (and almost sure remove) those formatting and review the formatter afore mentioned, if the Hibernate team agree with this.
I'll be awaiting your comments. Best regards

Naros: @aepedraza To be clear, I'm saying that the `)` goes on a new line, just its indention is incorrect.  

I'm not sure what you mean by _hibernate formatter for Eclipse_ .  We use check-style to verify that the format of the code is uniform to our desired layout.  

You can find the check-style rules in the `shared/config/checkstyle` directory in the root of the repository.  In master there are 2 files where one describes style rules that won't fail the build and another which will force the build to fail should any rules be broken.  I'd suggest you check both with your local changes.

aepedraza: I expressed myself wrong, I meant I disagree with the indentation for braces and parenthesis in new lines.
By *hibernate formatter* I mean [this XML](https://raw.githubusercontent.com/hibernate/hibernate-ide-codestyles/master/eclipse/hibernate-java-formatting.xml) mentioned in the post [Contributing to Hibernate using Eclipse](https://developer.jboss.org/wiki/ContributingToHibernateUsingEclipse) (which is in the contributing guidelines).
I will review the check-style rules as you say and I'll be correcting the indentation (and the formatting in general)

aepedraza:
@Naros I'm detecting a commented `private Map<Object,EntityEntry> entityEntries` variable. I think it would be good to remove it also taking advantage of my current refactoring, what do you think?

hibernate-core/src/main/java/org/hibernate/engine/spi/EntityKey.java
```diff
@@ -29,7 +30,9 @@
  * @author Gavin King
  * @author Sanne Grinovero
  */
-public final class EntityKey implements Serializable {
+public final class EntityKey implements Serializable, SerializableKey {
+
+	private static final long serialVersionUID = 776292106989119341L;
```
Sanne: please don't add any `serialVersionUID`, we intentionally decided it's best to not have them.  It's a long story ... you can probably find the discussion on the mailing list archives.

hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
```diff
@@ -1935,11 +1917,12 @@ public void removeSharedNaturalIdCrossReference(EntityPersister persister, Seria
 			final Object naturalIdCacheKey = naturalIdCacheAccessStrategy.generateCacheKey( naturalIdValues, persister, session );
 			naturalIdCacheAccessStrategy.evict( naturalIdCacheKey );
 
-//			if ( sessionCachedNaturalIdValues != null
-//					&& !Arrays.equals( sessionCachedNaturalIdValues, deletedNaturalIdValues ) ) {
-//				final NaturalIdCacheKey sessionNaturalIdCacheKey = new NaturalIdCacheKey( sessionCachedNaturalIdValues, persister, session );
-//				naturalIdCacheAccessStrategy.evict( sessionNaturalIdCacheKey );
-//			}
+			// if ( sessionCachedNaturalIdValues != null
```
Sanne: this reformatting should probably not be applied?

Please never reformat existing code, unless it's the primary and exclusive goal of the patch.

hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
```diff
@@ -1510,14 +1470,36 @@ public void serialize(ObjectOutputStream oos) throws IOException {
...
+	 * Writes a {@link Map} into an {@link ObjectOutputStream}
+	 *
+	 * @param keyMap The map to be serialized
+	 * @param oos The stream used for serialization
+	 * @param tracing Indicates whether the operation should be logged or not
+	 * @param keysName String used for logging purposes, usually the variable name of {@code keyMap}
+	 *
+	 * @throws IOException Thrown by Java I/O
+	 */
+	private void writeKeyMapToStream(Map<? extends SerializableKey, ? extends Object> keyMap, ObjectOutputStream oos, String keysName)
```
Sanne: nice idea!

hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
aepedraza: Seems I missed that, I'll undo that commit block somehow. Thanks for the advice

hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
aepedraza: Thanks!

hibernate-core/src/main/java/org/hibernate/engine/spi/EntityKey.java
Sanne: The discourse one if for users and usage questions. Developers use http://lists.jboss.org/pipermail/hibernate-dev/

See also http://hibernate.org/community/ : you're welcome to join.

hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
Sanne: The IDE formatting guidelines are meant for new code.

Unfortunately IDEA and Eclipse - and other IDEs - don't have the capabilities to set exactly identical formatting settings, they all have their own quirks.

So the understanding is people shouldn't reformat existing code - unless something is severely broken of course.

You can see that otherwise I might open this same class in NetBeans in a week and change it back again, we'd never stop reformatting ;-)

aepedraza: I see what you mean @Sanne, and agree with you.

So, to sumarize, I'll undo all the formatting hunks in my commits. And in the future I'll have an eye on automatic formatting.

Thanks for your time.

hibernate-core/src/main/java/org/hibernate/engine/spi/EntityKey.java
aepedraza: I've just joined! Can you say when this discussion took place? The archive is long.

PS: Sorry for deviating the PR discussion, but since you mentioned it...

Sanne: lol, looks like that was more than 10y ago :)
I still remembered it..

http://lists.jboss.org/pipermail/hibernate-dev/2008-May/003097.html

FYI you can search using this service, based on Elasticsearch:
 - http://search.jboss.org/#!q=serialVersionUID&project=hibernate

aepedraza: Thank you so much! I just saw a few formatting undos I missed in my last commits. I'll be tackling that review today

aepedraza:
Hi @Sanne and @Naros, I'm seeing some checks failing in CI, is there any problem with my commits? I didn't change any significant code in `225fadc`, just tabs and spaces.

I'm running `gradle check` in my local, please let me know if you need me to do something with this.

Naros:
The problem is lines 1892-1894.  Those lines have leading space characters and lines should always only ever be indented using tabs; not spaces.

aepedraza:
Oh, that was preexistent code. I'll _"un-undo"_ that. Did you add that rule recently?

Naros:
Nope, that rule has been there since 2013 :).

aepedraza:
Done! back to normal 😃

gsmet:
@aepedraza could you please squash the formatting commits into the main one(s)? That would help to have a clear history.

@Sanne as you started with the review, I'll let you finish it. As we now have the 5.4 branch, we can now commit this sort of improvements.

aepedraza:
@gsmet I squashed the commits as you asked for leaving a cleaner history.

Just a little detail: `b3b1135: HHH-10661 - Remove repeated code with new private method` has also a few spaces-to-tabs formatting improvements as I couldn't split them because I wasn't able to do `git add -i` to stage hunks while doing an interactive rebase to squash. If you know how to do it please let me know and I'll be contacting you by private message or some other way.

As I'm writing, I see some failed analysis, but seeing the logs there is an issue parsing some `hbm.xml` mapping files

    [2018-07-15 15:50:19] [autobuild] [2018-07-15 15:50:19] [ERROR] Parsing error in file /opt/src/hibernate-core/src/test/java/org/hibernate/test/mappingexception/InvalidMapping.hbm.xml

aepedraza:
@gsmet @Sanne The checks are now successful, do you mind to review this PR and merge if everything is OK?
Thanks in advance for your time

gsmet:
Hi @aepedraza ,

Sorry for the delay. I just got back to this patch.

There is one issue with it: the `SerializableKey` interface you put in `internal` leaks into our SPI as we have SPI classes implementing it. We can't do that.

One way to fix the issue would be to rely on `writeObject()` and `writeReplace()` instead of having our own custom method but this would change the serialization format and I'm far from being sure the serialization is covered in our tests :/. So if we do that, we need to add serialization tests for `StatefulPersistenceContext` which wouldn't be a bad thing. Would you be up for it?

Another option would be to use method handles (and we could probably improve the deserialization process too) but we usually refrain from doing that too much as using method handles in our code consumes some more memory.

WDYT?

aepedraza:
Hi @gsmet:

I think that the first option is better. IMHO, adding tests is always good ;)

However, I'll be in touch because I was having some issues in my local running some tests: I was having some classes not found in the IDE. What chat channel are you attending mostly during the transition to the new platform? Tell me where I can find you easily and I'll be in touch.

gsmet:
I am on HipChat. But if you're using Eclipse, it's a real mess, I can't get the tests running there.

I use the command line:
```
cd hibernate-core
../gradlew clean test --tests=your.package.YourClass
```

And with the `--debug-jvm` option to enable debug.

aepedraza:
Hi @gsmet, sorry the delay, My payed job got me busy last week. Actually, I'm using Intellij. I'll try the console and I'll be in touch in the chat channels.

sebersole:
Most likely we need to add config to the build scripts to properly
configure Eclipse.  There is some Eclipse config, but sounds like it needs
fixing or more added.

As none of the main ORM team uses Eclipse it is not regularly checked with
new Eclipse and/or Gradle versions.  Someone using Eclipse would need to be
involved in that.

sebersole:
Alvaro - Andrea, Chris and I all use IntelliJ and it works fine if you use
IntelliJ's Gradle import feature.  You will also have to use the Gradle
`compile` task to run code generation - you have to do that step manually
ATM.

beikov:
I tried to address the concerns Guillaume brought up (exposing internal API through SPI classes). I'll close this PR as the following supersedes this: https://github.com/hibernate/hibernate-orm/pull/3844

PR #2358, opened 2018-06-20
HHH-12713 (master) - Make EntityGraph creation more convenient

aleksandarsusnjar:
Please see [HHH-12713](https://hibernate.atlassian.net/browse/HHH-12713) for details. This change adds new approaches to entity graph creation, manipulation and application, making it easier to use. Ideally this would yield their wider adoption and, with them, OxM.

Code, unit tests, JavaDoc and AsciiDoc changes/additions are included.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/graph/EntityGraphs.java
+++ b/hibernate-core/src/main/java/org/hibernate/graph/EntityGraphs.java
@@ -1,0 +1,352 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.graph;
+
+import java.text.ParseException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import javax.persistence.AttributeNode;
+import javax.persistence.EntityGraph;
+import javax.persistence.EntityManager;
+import javax.persistence.Query;
+import javax.persistence.Subgraph;
+import javax.persistence.TypedQuery;
+
+/**
+ * A collection of {@link EntityGraph} utilities.
+ * These methods really belong inside other classes that we cannot modify, hence here.
+ * 
+ * @author asusnjar
+ */
+public final class EntityGraphs {
+
+	public static String HINT_FETCHGRAPH = "javax.persistence.fetchgraph";
+	public static String HINT_LOADGRAPH = "javax.persistence.loadgraph";
+
+	/**
+	 * Configures the query hints so that ONLY the attributes specified by the graph are eagerly fetched, regardless of
+	 * what the mapping defaults indicate (even if they specify more properties to be loaded eagerly).
+	 * 
+	 * @param <T>   Root entity type of the query and graph.
+	 * 
+	 * @param query Query to configure.
+	 * @param graph Graph of properties/attributes to eagerly load.
+	 * 
+	 * @return The same (input) query.
+	 * 
+	 * @see #setFetchGraph(Query, EntityManager, Class, CharSequence)
+	 * @see #setLoadGraph(Query, EntityGraph)
+	 */
+	public static <T> Query setFetchGraph(final Query query, EntityGraph<T> graph) {
+		return query.setHint( HINT_FETCHGRAPH, graph );
+	}
+
+	/**
+	 * Configures the query hints so that ONLY the attributes specified by the graph are eagerly fetched, regardless of
+	 * what the mapping defaults indicate (even if they specify more properties to be loaded eagerly).
+	 * 
+	 * @param <T>   Root entity type of the query and graph.
+	 * 
+	 * @param query Query to configure.
+	 * @param graph Graph of properties/attributes to eagerly load.
+	 * 
+	 * @return The same (input) query.
+	 * 
+	 * @see #setFetchGraph(Query, EntityGraph)
+	 * @see #setLoadGraph(Query, EntityGraph)
+	 */
+	public static <T> Query setFetchGraph(final Query query, EntityManager em, Class<T> rootType, CharSequence graph) {
+		return setFetchGraph( query, EntityGraphParser.parse( em, rootType, graph ) );
+	}
+
+	/**
+	 * Configures the query hints so that the attributes specified by the graph are eagerly fetched IN ADDITION to those
+	 * specified to be loaded eagerly in the mapping.
+	 * 
+	 * @param <T>   Root entity type of the query and graph.
+	 * 
+	 * @param query Query to configure.
+	 * @param graph Graph of properties/attributes to eagerly load.
+	 * 
+	 * @return The same (input) query.
+	 * 
+	 * @see #setFetchGraph(Query, EntityGraph)
+	 * @see #setLoadGraph(Query, EntityManager, Class, CharSequence)
+	 */
+	public static <T> Query setLoadGraph(final Query query, EntityGraph<T> graph) {
+		return query.setHint( HINT_LOADGRAPH, graph );
+	}
+
+	/**
+	 * Configures the query hints so that the attributes specified by the graph are eagerly fetched IN ADDITION to those
+	 * specified to be loaded eagerly in the mapping.
+	 * 
+	 * @param <T>   Root entity type of the query and graph.
+	 * 
+	 * @param query Query to configure.
+	 * @param graph Graph of properties/attributes to eagerly load.
+	 * 
+	 * @return The same (input) query.
+	 * 
+	 * @see #setFetchGraph(Query, EntityGraph)
+	 * @see #setLoadGraph(Query, EntityGraph)
+	 */
+	public static <T> Query setLoadGraph(final Query query, EntityManager em, Class<T> rootType, CharSequence graph) {
+		return setLoadGraph( query, EntityGraphParser.parse( em, rootType, graph ) );
+	}
+
+	/**
+	 * Creates a query just like {@link EntityManager#createQuery(String, Class)} does but allows the query string to
+	 * begin with the "fetch" and/or "load" entity graph specifications (see {@link EntityGraphParser}) that will
+	 * internally converted to correct hints (see {@link #setFetchGraph(Query, EntityManager, Class, CharSequence)} and
+	 * {@link #setLoadGraph(Query, EntityManager, Class, CharSequence)}.
+	 * 
+	 * @param <T>      Root entity type of the query and graph.
+	 * 
+	 * @param em       EntityManager to use to create the query and entity graphs (if found).
+	 * @param rootType The root type of results.
+	 * @param qlString The query language string that can begin with fetch/load specification (see
+	 *                 {@link EntityGraphParser}).
+	 *                 
+	 * @return The typed query preconfigured as per any fetch/load specifications in the {@code qlString}.
+	 * 
+	 * @throws InvalidGraphException if a specified fetch or load graph is invalid.
+	 */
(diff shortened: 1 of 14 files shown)


aleksandarsusnjar:
Submitted a 4.3 backport [pull request 2362](https://github.com/hibernate/hibernate-orm/pull/2362).

sebersole:
This is a very good idea and based on your comment sounds very well thought out and complete.  I am going on vacation next week, so will not have chance to look before then but will definitely take a peek when I get back.

aleksandarsusnjar:
@sebersole Thanks! There are more things that could be done but may require additional engine work and/or JPA enhancements and yield powerful but potentially dangerous concepts such as allowing cycles in these fetch/load graphs.

Please note that the code I submitted is able to parse the subclass subgraphs but that fails due to [HHH-10378 - EntityGraph API: subclass subgraph support missing/broken](https://hibernate.atlassian.net/browse/HHH-10378).

I'd like to fix that too (seems easy based on findings documented in HHH-10378) but there is a risk of something else depending on current behaviour, for whatever reason, so I'd like some input on that.

sebersole:
I haven't forgotten about this.  Just been a very busy summer here plus working hard on 6.0.

WRT HHH-10378, what kind of input are you looking for?

hibernate-core/src/main/java/org/hibernate/graph/EntityGraphs.java
```diff
@@ -0,0 +1,352 @@
...
+
+/**
+ * A collection of {@link EntityGraph} utilities.
+ * These methods really belong inside other classes that we cannot modify, hence here.
+ * 
+ * @author asusnjar
+ */
+public final class EntityGraphs {
+
+	public static String HINT_FETCHGRAPH = "javax.persistence.fetchgraph";
```
sebersole: These should be unified with `org.hibernate.jpa.QueryHints#HINT_FETCHGRAPH` and `#HINT_LOADGRAPH`

aleksandarsusnjar: I agree. I remember looking into this but I think only in JPA (i.e. not Hibernate specific). It would be ideal if those are defined there... but they are not.

gsmet:
@sebersole do you plan to merge this in master? Or do you take the idea to apply it to 6?

aleksandarsusnjar:
@gsmet, @sebersole I would like to see this in v5 as well. In v6 it could be, perhaps, better integrated.

sebersole:
@aleksandarsusnjar I agree that it should go in v5.  My comment was a little incomplete, hence the confusion.  What I was trying to say is that I would make some changes to your pull request to match the corresponding contracts I have in v6 so that we do not run into avoidable complications wrt merging between v5 and v6 sources.  Hopefully that is more clear

aleksandarsusnjar:
#response_container_BBPPID{font-family: initial; font-size:initial; color: initial;} That was my understanding as well. I agreed to that.                                                                                                                                                                          From: notifications@github.comSent: August 28, 2018 10:17 AMTo: hibernate-orm@noreply.github.comReply to: reply@reply.github.comCc: github@susnjar.net; mention@noreply.github.comSubject: Re: [hibernate/hibernate-orm] HHH-12713 (master) - Make EntityGraph creation more convenient (#2358)  @aleksandarsusnjar I agree that it should go in v5.  My comment was a little incomplete, hence the confusion.  What I was trying to say is that I would make some changes to your pull request to match the corresponding contracts I have in v6 so that we do not run into avoidable complications wrt merging between v5 and v6 sources.  Hopefully that is more clear

hibernate-core/src/main/java/org/hibernate/graph/EntityGraphs.java
sebersole: Not sure what you mean here

hibernate-core/src/main/java/org/hibernate/graph/FetchAndLoadEntityGraphs.java
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Internally used class to carry a pair of graphs (fetch and load), 
+ * a result of {@link EntityGraphParser#parsePreQueryGraphDescriptors(EntityManager, Class, ParseBuffer)}.
+ * 
+ * @author asusnjar
+ *
+ * @param <T> Root entity type of both graphs.
+ */
+class FetchAndLoadEntityGraphs<T> {
```
sebersole: There can only ever be either a fetch graph or a load graph, never both.  So why do you define a class that bundles both?  IMO this really ought to be something more like `org.hibernate.engine.query.spi.EntityGraphQueryHint` with an enum (designating LOAD/FETCH) and the actual graph

sebersole:
Are you anticipating users calling EntityGraphParser directly?  It seems to me that you put a few contracts in the "API" namespace which are not necessarily contracts we want support for application usage

hibernate-core/src/main/java/org/hibernate/graph/FetchAndLoadEntityGraphs.java
aleksandarsusnjar: My understanding was the same but the JP contract statement and interface is not clear about that and does not exclude that case (that I know). Some other developer (ORM or client) could take this, for example, as the result being a union of what the two would independently specify. Thus, we made sure that both can be set. If not valid, the engine underneath would fail it. Additionally, this is just a container that knows what was set and what to pass on. It isn't a public class. If the case is truly invalid, it itself could allow only one to be set.

hibernate-core/src/main/java/org/hibernate/graph/EntityGraphs.java
aleksandarsusnjar: What I mean by "I agree"? :) ... I agree with you.

Let me paraphrase:

1. When this was written the goal was to try to be as standard JPA as possible and not Hibernate specific. Unfortunately, I didn't find these constants anywhere but in org.hibernate.*. I may also have (don't remember now) hit some cyclic references in Hibernate itself - i.e. not having that constant available in my dependencies.

It would be ideal if these constants are defined in JP. Better yet, this entire class, EntityGraphs, should really have to exist. It enriches methods of other classes only, namely Query, EntityManager, EntityGraph, Subgraph and AttributeNode, all in javax.persistence.

aleksandarsusnjar: Edited some typos/similar above. The reason we went ahead with this class is because we feel that the spirit of the approach is correct and that something like this must eventually find its way into JP. As this adds nothing on top of what JP offers now (other than convenience), it should be easy/trivial to not only reimplement this class to use the then-available JP implementation but also to update any code using this class to use JP directly.

aleksandarsusnjar:
> Are you anticipating users calling EntityGraphParser directly? 

Yes (until EntityGraph has a method/constructor to do the same instead). The reason is that we need to be able to dynamically (at runtime) collect the (configured) "needs" of multiple different bits of code to receive the data, then create a union of them. Each of these bits of code will supply a graph. 

Some of these will be static parts that will conditionally be added to the union if the user enables the product feature. Others are more dynamic - for example, users may select which columns they want in the display grid.

sebersole:
Hopefully 3rd time is the charm and I don't end up throwing away another attempted response w/ CTRL+W ;)

My concern is more that EntityGraphs and EntityGraphParser seem to duplicate effort and expose the same functionality multiple ways.  And EntityGraphs actually IMO contains a number of unnecessary methods.

Also, I am personally not a fan of the feature you added to prepend one of these EntrityGraph strings to the beginning of a JPQL/HQL query.  Personally, if the intention is to expose EntityGraphParser for users then that should be the way it is done period. 

So I will work on integrating my 6.0 work with your work based on these comments and see.  I'll post a new PR when done

aleksandarsusnjar:
#response_container_BBPPID{font-family: initial; font-size:initial; color: initial;}  Re prepending these to queries - I knew that can be controversial but put it there as an illustration. I am fine with simoly removing it.                                                                                                                                                                        Aleksandar Šušnjar                                                                                                                                                                           From: notifications@github.comSent: September 28, 2018 10:38 AMTo: hibernate-orm@noreply.github.comReply to: reply@reply.github.comCc: github@susnjar.net; mention@noreply.github.comSubject: Re: [hibernate/hibernate-orm] HHH-12713 (master) - Make EntityGraph creation more convenient (#2358)  Hopefully 3rd time is the charm and I don't end up throwing away another attempted response w/ CTRL+W ;)
My concern is more that EntityGraphs and EntityGraphParser seem to duplicate effort and expose the same functionality multiple ways.  And EntityGraphs actually IMO contains a number of unnecessary methods.
Also, I am personally not a fan of the feature you added to prepend one of these EntrityGraph strings to the beginning of a JPQL/HQL query.  Personally, if the intention is to expose EntityGraphParser for users then that should be the way it is done period.
So I will work on integrating my 6.0 work with your work based on these comments and see.  I'll post a new PR when done

sebersole:
I decided to go a little further here and extract the graph-texting parsing into a proper Antlr grammar.  Makes it much easier to see the allowable syntax.

I also made a small change in the syntax, specifically how sub-graph sub-type is specified.  Initially you had, e.g., `theAttribute:TheSubType(theSubAttribute, ...)`.  I changed this to be imo a little more readable: `theAttribute(TheSubType: theSubAttribute, ...)`.  Just finishing up, but wanted to discuss those changes up front

aleksandarsusnjar:
#response_container_BBPPID{font-family: initial; font-size:initial; color: initial;}  Will this be in Hibernate 5?We went with the original because of how these things are used, wanting to split attributes of subtypes into sections rather than allowing them to be mixed together (i.e. there can be multiple subtypes involved and we may need different specs for each). But, 6 or half a dozen, it is OK.                                                                                                                                                                        Aleksandar Šušnjar                                                                                                                                                                           From: notifications@github.comSent: October 3, 2018 11:50 AMTo: hibernate-orm@noreply.github.comReply to: reply@reply.github.comCc: github@susnjar.net; mention@noreply.github.comSubject: Re: [hibernate/hibernate-orm] HHH-12713 (master) - Make EntityGraph creation more convenient (#2358)  I decided to go a little further here and extract the graph-texting parsing into a proper Antlr grammar.  Makes it much easier to see the allowable syntax.
I also made a small change in the syntax, specifically how sub-graph sub-type is specified.  Initially you had, e.g., theAttribute:TheSubType(theSubAttribute, ...).  I changed this to be imo a little more readable: theAttribute(TheSubType: theSubAttribute, ...).  Just finishing up, but wanted to discuss those changes up front

sebersole:
Yes, like I said initially, this is me basically back-porting some of the work I have done in 6.0 to 5.4 and merging it with your work. 

The changes i back-ported are mainly in regards to the definitions of JPA type contracts.  I leverage those in the graph and graph parsing code.

And then the Antlr stuff.

About the sub-graph sub-typing... you can still do the same:
`responsibleParty(taxIdNumber), responsibleParty(Corporation: ceo,cfo), responsibleParty(NonProfit: sector,chairman)`

it's no different than what you'd have to do in your original syntax:
`responsibleParty(taxIdNumber), responsibleParty:Corporation(ceo,cfo), responsibleParty:NonProfit(sector,chairman)`

No different functionally.  Just re-defining the recognition tokens to make it easier to recognize and process the sub-type as part of the sub-graph (which it is conceptually).

sebersole:
I love the idea of being able to parse an EntityGraph from a String representation.  And many of the other ideas you added - merging, etc.  I just knew that a lot of this dove-tailed with the things we are doing in 6 and decided to back port some of that to make the graph building easier.  All the work was yours, I just moved it around a bit.  Great job on this.

gsmet:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2558 .

gsmet:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2558 .

PR #2359, opened 2018-06-21
HHH-12717 order_inserts: commit with OneToOne and OneToMany causes incrorrect ordering and subsequent constraint violation

JanHinrichs2:
Testcase for HHH-12717 order_inserts: commit with OneToOne and OneToMany causes incrorrect ordering and subsequent constraint violation

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/insertordering/InsertOrderingWithOneToOneAndOneToMany.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/insertordering/InsertOrderingWithOneToOneAndOneToMany.java
@@ -1,0 +1,84 @@
+package org.hibernate.test.insertordering;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.persistence.CascadeType;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.OneToMany;
+import javax.persistence.OneToOne;
+import javax.persistence.Table;
+
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.cfg.Configuration;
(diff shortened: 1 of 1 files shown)


vladmihalcea:
The problem is not the ordering, but your mapping.

The `Parent` also has a FK to a `Child`:

```
@OneToOne(cascade = CascadeType.ALL)
public Child child;
```

which means the `Child` needs to exist prior to the `Parent`. 

So, the `Child` entity is the parent of the `Parent` entity.

And, at the same time, the `Parent` is also a parent to the `Child`:

```
@OneToMany
private List<Child> otherChilds = new ArrayList<Child>();
```

This is a "Chicken or the Egg" problem and no sorting algorithm will figure out which one should be the first.

Looking forward to your explanation about your Domain Model.

JanHinrichs2:
Our Domain Model worked without any problems since years with hibernate 3.
With hibernate 5.2.17 the problem arises only if order_inserts is true.
With hibernate 5.2.4 the same test was green, even with activated order_inserts.
In the log output of the test I see, that the OneToMany association is mapped by a join table, for which the insert statement is executed late.
If we add @JoinColumn to the OneToMany List the association is mapped by a join column in the child table. If with this changed mapping the same child object is added to the OneToOne and at the same time to the OneToMany association there in theory a "Chicken or the Egg" problem could arrise. In reality a test showed me, that hibernate in that situation executes two insert statements an an additional update statement to avoid that problem.
But in our test case there is no @JoinColumn and additionally we never add the same child object to the OneToOne and to the OneToMany association.
In our application with thousands of entities we have a lot of constraint violations if we switch on order_inserts. With hibernate 3 this problem does'nt  exist.
What ca we do?

vladmihalcea:
Closed issue because, unless the database table model is valid and parent tables don't have FK back to child tables, there's nothing we can do to fix it for every combination of unidirectional and bidirectional associations.

PR #2360, opened 2018-06-21
HHH-12712 - Implicit Join with IS NULL does not render a LEFT JOIN

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12712


vladmihalcea:
Closed as the associated Jira issue was closed as well.

PR #2361, opened 2018-06-21
HHH-12718 Entity changes in @PreUpdate callback are not persisted when lazy loading is active for more than one field

bestmeat:
Test Case for https://hibernate.atlassian.net/browse/HHH-12718

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/callbacks/PreUpdateModificationsLazyLoadingTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/callbacks/PreUpdateModificationsLazyLoadingTest.java
@@ -1,0 +1,90 @@
+package org.hibernate.jpa.test.callbacks;
+
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.jpa.test.BaseEntityManagerFunctionalTestCase;
+import org.hibernate.testing.TestForIssue;
+import org.hibernate.testing.bytecode.enhancement.BytecodeEnhancerRunner;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import javax.persistence.*;
+import java.nio.ByteBuffer;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/test/java/org/hibernate/jpa/test/callbacks/PreUpdateModificationsLazyLoadingTest.java
```diff
@@ -0,0 +1,90 @@
...
+        @ElementCollection
+        private List<String> tags;
+
+        @Lob
+        @Basic(fetch = FetchType.LAZY)
+        private ByteBuffer image;
+
+        @PrePersist
+        void beforeCreate() {
+            createdAt = Instant.now();
```
JWGmeligMeyling: Aren't you supposed to use a `setCreatedAt` method here? I think dirty tracking enhancement is not triggered with field access (and the dirty tracking optimisation would be useless if the fields where dirty tracked anyway regardless of the optimisation being enabled).

bestmeat: Hi Jan Willem, you're right. I'll update the test case to use setters.
However I think it would make a difference with regards to the underlying bug that I wanted to isolate here. The direct field assignment p.name = "Changed Name" is recognized as dirty field. The interesting behavior is that as soon as you remove one of the lazily loaded fields the test case goes through.

JWGmeligMeyling: Thanks for editing the test case! I see it still fails indeed, interesting bug.

vladmihalcea:
Thanks. I created this [PR](https://github.com/hibernate/hibernate-orm/pull/2375) with a possible fix.

PR #2363, opened 2018-06-22
Add a test-case for HHH-12480

ptahchiev:
* https://hibernate.atlassian.net/browse/HHH-12480

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/proxy/HibernateUnproxyTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/proxy/HibernateUnproxyTest.java
@@ -8,7 +8,11 @@ package org.hibernate.test.proxy;
 
 import org.hibernate.Hibernate;
 import org.hibernate.jpa.test.BaseEntityManagerFunctionalTestCase;
+import org.hibernate.proxy.HibernateProxy;
+
 import org.jboss.logging.Logger;
+
+import org.junit.Assert;
 import org.junit.Test;
 
 import javax.persistence.*;
@@ -22,8 +26,45 @@ public class HibernateUnproxyTest extends BaseEntityManagerFunctionalTestCase {
 
     @Override
     protected Class<?>[] getAnnotatedClasses() {
-        return new Class<?>[]{Parent.class, Child.class};
+        return new Class<?>[]{Parent.class, Child.class, SubChild.class};
     }
+
+    @Test
+    public void testInitializedProxySubChildCanBeUnproxied() {
+        Parent u = new Parent();
+        SubChild c = new SubChild();
+        u.setChild(c);
+        doInJPA(this::entityManagerFactory, (entityManager -> {
+            entityManager.persist(u);
+        }));
+        doInJPA(this::entityManagerFactory, (entityManager -> {
+            Parent parent = entityManager.find(Parent.class, u.getId());
+            SubChild unproxiedChild = initializeAndUnproxy(getUser(parent));
+            assertEquals(SubChild.class, unproxiedChild.getClass());
+        }));
+    }
+
+
+    public <U extends Child> U getUser(Parent p) {
+        return (U) p.getChild();
+    }
+
+    public <T> T initializeAndUnproxy(T entity) {
+        if (entity == null) {
+            throw new NullPointerException("Entity passed for initialization is null!");
+        }
+
+        if (!Hibernate.isInitialized(entity)) {
+            Hibernate.initialize(entity);
+            if (entity instanceof HibernateProxy) {
+                entity = (T) ((HibernateProxy) entity).getHibernateLazyInitializer().getImplementation();
+            } else {
+                Assert.fail();
+            }
+        }
+        return entity;
+    }
+
 
     @Test
     public void testInitializedProxyCanBeUnproxied() {
@@ -153,5 +194,10 @@ public class HibernateUnproxyTest extends BaseEntityManagerFunctionalTestCase {
         public Parent getParent() {
             return parent;
         }
+    }
+
+    @Entity(name = "SubChild")
+    public static class SubChild extends Child {
+
     }
 }


hibernate-core/src/test/java/org/hibernate/test/proxy/HibernateUnproxyTest.java
```diff
@@ -25,6 +27,41 @@
...
+    public void testInitializedProxySubChildCanBeUnproxied() {
+        Parent u = new Parent();
+        SubChild c = new SubChild();
+        u.setChild(c);
+        doInJPA(this::entityManagerFactory, (entityManager -> {
+            entityManager.persist(u);
+        }));
+        doInJPA(this::entityManagerFactory, (entityManager -> {
+            Parent parent = entityManager.find(Parent.class, u.getId());
+            SubChild unproxiedChild = initializeAndUnproxy(getUser(parent));
```
JWGmeligMeyling: This test apparently also fails for Java 8 so this is not a true reproducer for HHH-12480. The way that the generic type for `getUser` is defined, it tries to cast `p.getChild()`, the proxy, to a `SubChild` already. This can't be done as the entity is not  initialised yet. At the time this proxy is made, only the id of the child is known, as well as its super type (i.e. the target type of the child association, which is in this case supposedly `Child`). To know the exact subtype you basically have to initialise the entire sub-child, eliminating the use for lazy loading.

You should cast to the subtype only after extracting the initialised proxy implementation in initializeAndUnproxy.

ptahchiev: @JWGmeligMeyling thank you for your response. When I run `./gradlew buiid` locally this is what I get:
```
> Task :hibernate-core:test

org.hibernate.test.proxy.HibernateUnproxyTest > testInitializedProxySubChildCanBeUnproxied FAILED
    java.lang.ClassCastException at HibernateUnproxyTest.java:40

org.hibernate.test.fileimport.CommandExtractorServiceTest > testImportFile FAILED
    org.junit.ComparisonFailure

org.hibernate.test.fileimport.MultiLineImportFileTest > testImportFile FAILED
    org.junit.ComparisonFailure at MultiLineImportFileTest.java:66

6422 tests completed, 3 failed, 350 skipped

> Task :hibernate-core:test FAILED

FAILURE: Build failed with an exception.

```

And I don't understand why pretty much the same code works fine with java8 and fails with java9 when I use the hibernate test-case templates:
https://github.com/ptahchiev/hibernate-test-case-templates/tree/HHH-12480

JWGmeligMeyling: It doesn't seem to work in Java 8 either and it doesn't fail where you expect it to fail.

Try with:

```java
    public Child getUser(Parent p) {
        return p.getChild();  // Get rid of cast to SubChild, as the proxy will be of type Child
    }

    public <T> T initializeAndUnproxy(Object entity) {
        if (entity == null) {
            throw new NullPointerException("Entity passed for initialization is null!");
        }

        if (!Hibernate.isInitialized(entity)) {
            Hibernate.initialize(entity);
            if (entity instanceof HibernateProxy) {
                entity = (T) ((HibernateProxy) entity).getHibernateLazyInitializer().getImplementation();
            }
        }
        return entity;
    }
```

hibernate-core/src/test/java/org/hibernate/test/proxy/HibernateUnproxyTest.java
```diff
@@ -25,6 +27,41 @@
         return new Class<?>[]{Parent.class, Child.class};
```
JWGmeligMeyling: Add `SubChild` here as well, perhaps `SubChild` is just persisted as a `Child` currently..

ptahchiev: @JWGmeligMeyling Woooow you were right.. i've committed the SubChild.class.

gsmet:
@ptahchiev so what is the status of this? Did you fix everything and this one is not an issue for you anymore? Or did you finally get a test case working with JDK 8 and not with JDK 9?

Thanks!

ptahchiev:
- I have a test-case using the hibernate-testcase-templates which clearly shows a test that works with java8 and fails with java9: https://github.com/ptahchiev/hibernate-test-case-templates/tree/HHH-12480
 - This pull-request adds the same test-case but integrated into the hibernate build. It works on java8, but i can't run it on java9 because some other tests are failing before my test:
````
petar@petar-XPS-15-9550:~/workspace/hibernate-orm(HHH-12480)$ ./gradlew build

> Configure project :hibernate-osgi
[WARN] Skipping all tests for hibernate-osg due to Karaf issues with JDK 9

> Task :hibernate-core:compileTestJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :documentation:compileTestJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :documentation:test

org.hibernate.userguide.envers.QueryAuditTest > test FAILED
    java.lang.NoClassDefFoundError
        Caused by: java.lang.ClassNotFoundException

org.hibernate.userguide.envers.CustomRevisionEntityTest > test FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.EntityTypeChangeAuditDefaultTrackingTest > testLifecycle FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.QueryAuditAdressCountryTest > test FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.QueryAuditWithModifiedFlagTest > test FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.EntityTypeChangeAuditTrackingRevisionListenerTest > testLifecycle FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.ModifiedFlagsAuditTest > test FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.EntityTypeChangeAuditTest > test FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.ValidityStrategyAuditTest > test FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.DefaultAuditTest > test FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.mapping.converter.hbm.MoneyConverterHbmTest > testConverterMutability FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.mapping.dynamic.DynamicEntityTest > test FAILED
    java.lang.NoClassDefFoundError

455 tests completed, 12 failed, 24 skipped

> Task :documentation:test FAILED

FAILURE: Build failed with an exception.
````
I was hoping that hibernate's jenkins has a java-9 build that would show that the code works on 8 and fails on 9

**Comment by JWGmeligMeyling** [comment]
Tests now pass on Java 8, untested on Java 9. I'm not running a Java 9 locally so I can't test that, but provided a stack trace I can have a look. My gut feeling is that this is related to the cast in `getUser`, as I would expect it to throw a `CastClassException` anyway, but seems to work here nevertheless. Perhaps in older Java versions the cast is simply omitted.

File: hibernate-core/src/test/java/org/hibernate/test/proxy/HibernateUnproxyTest.java [hunk 3]
```diff
@@ -22,9 +24,44 @@
...
+        if (entity == null) {
+            throw new NullPointerException("Entity passed for initialization is null!");
+        }
+
+        if (!Hibernate.isInitialized(entity)) {
+            Hibernate.initialize(entity);
+            if (entity instanceof HibernateProxy) {
+                entity = (T) ((HibernateProxy) entity).getHibernateLazyInitializer().getImplementation();
+            }
+        }
```
JWGmeligMeyling: Perhaps we should add  `} else { Assert.fail(); }` statements here to make sure that in both scenario's we're dealing with proxies.

ptahchiev: Done

JWGmeligMeyling: I'll have a try at j9!

gsmet:
@JWGmeligMeyling the PR seems to be in good hands. Ping me when you have validated that there is an issue so that I can plan this issue for one of our next releases.

@ptahchiev FYW, you can run a specific test with `./gradlew test --tests org.hibernate.test.bytecode.enhancement.lazy.LazyWithMappedSuperclassTest`.

Thanks for mentioning the JDK 9 build is broken. It got broken late last week due to a totally unrelated dependency upgrade. I'm in the process of fixing it.

sranjha:
I did the test for @JWGmeligMeyling on JDK 9.0.4. The tests failed with:

```
org.hibernate.userguide.envers.CustomRevisionEntityTest > test FAILED
    java.lang.NoClassDefFoundError
        Caused by: java.lang.ClassNotFoundException

org.hibernate.userguide.envers.DefaultAuditTest > test FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.EntityTypeChangeAuditDefaultTrackingTest > testLifecycle FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.EntityTypeChangeAuditTest > test FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.EntityTypeChangeAuditTrackingRevisionListenerTest > testLifecycle FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.ModifiedFlagsAuditTest > test FAILED
    java.lang.NoClassDefFoundError

org.hibernate.userguide.envers.QueryAuditAdressCountryTest > test FAILED
    java.lang.NoClassDefFoundError
```

So while the JDK 9 build is broken, this reproducer test case just seem to pass fine.

gsmet:
I fixed the JDK 9 build so it should be OK now.

Where are we on this? In the end is there an issue with JDK 9 or not? Thanks!

ptahchiev:
@gsmet I can confirm the build is fixed. The test now fails with jdk9 and passes with jdk8

gsmet:
@ptahchiev OK. Could you rebase and squash your commits to have only one? Currently it's a bit of a mess. Thanks!

ptahchiev:
@gsmet i tried but it doesn't seem to work :(

gsmet:
@ptahchiev I did it and squashed your commits into one.

I can confirm that:
```
./gradlew test --tests org.hibernate.test.proxy.HibernateUnproxyTest
```
succeeds with JDK 8 but fails with JDK 9, even with Bytebuddy.

The stacktrace is:
```
java.lang.ClassCastException: org.hibernate.test.proxy.HibernateUnproxyTest$Child$HibernateProxy$A91PflG2 cannot be cast to org.hibernate.test.proxy.HibernateUnproxyTest$SubChild
	at org.hibernate.test.proxy.HibernateUnproxyTest.lambda$testInitializedProxySubChildCanBeUnproxied$1(HibernateUnproxyTest.java:42)
	at org.hibernate.testing.transaction.TransactionUtil.doInJPA(TransactionUtil.java:205)
	at org.hibernate.testing.transaction.TransactionUtil.doInJPA(TransactionUtil.java:246)
	at org.hibernate.test.proxy.HibernateUnproxyTest.testInitializedProxySubChildCanBeUnproxied(HibernateUnproxyTest.java:40)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.hibernate.testing.junit4.ExtendedFrameworkMethod.invokeExplosively(ExtendedFrameworkMethod.java:45)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:844)
```

gsmet:
@ptahchiev I took a look and I don't think it's an Hibernate issue.

The offending line is:
```java
SubChild unproxiedChild = initializeAndUnproxy(getUser(parent));
```

Your issue is not with the proxy initialization (btw, you should use `Hibernate.unproxy()` instead of your own thing), it is with this part of the code:
```
getUser(parent)
```
With your code, apparently, it's trying to cast it to `SubChild` directly, which is wrong, because before the `unproxy()` operation, the proxy class is inheriting from `Child`, not `SubChild` as it doesn't know what the element is and only has the contract coming from the `Parent` class.

If I change your line to be:
```
SubChild unproxiedChild = (SubChild) Hibernate.unproxy( parent.getChild() );
```

Everything works and the test passes.

No idea why Java 9 interprets this differently but my IDE also interprets your `getUser(parent)` in this way: `<SubChild> SubChild org.hibernate.test.proxy.HibernateUnproxyTest.getUser(Parent p)`.

I think it's not an issue with Hibernate itself but more with the compiler/JDK doing different things.

Can you please confirm that this answer is OK with you?

ptahchiev:
I agree that maybe `Hibernate.unproxy` is better than the current implementation. However I don't want it scattered everywhere in the code because one day I might decide to migrate to EclipseLink or OpenJPA and then I will need to refactor million lines of code. That's why I have it in a separate method. 
```
With your code, apparently, it's trying to cast it to SubChild directly, which is wrong, because before the unproxy() operation
```
How is that possible? This line:
```
SubChild unproxiedChild = initializeAndUnproxy(getUser(parent));
```
does not cast anything. It simply assigns the result of `initializeAndUnproxy` which should be already unproxied to `SubChild`.

I'm not convinced by your answer, but if you say it's not a HIbernate issue I will have to agree, (since you know hibernate better than me) and take this question to the JVM team.

vladmihalcea:
@ptahchiev Maybe the problem is here, not inside `initializeAndUnproxy`:

```
public <U extends Child> U getUser(Parent p) {
        return (U) p.getChild();
}
```

Is the `HibernateProxy$A91PflG2` really implementing `SubChild`? Is there any inheritance involved? It could be a Javassist issue, not a Hibernate one.

Try setting the `hibernate.bytecode.provider` configuration property to `bytebuddy` and see if it works.

> However I don't want it scattered everywhere in the code because one day I might decide to migrate to EclipseLink or OpenJPA and then I will need to refactor million lines of code.

That's highly unlikely. :smiley: What would anyone do that? They usually go the other way around. :wink:

gsmet:
@ptahchiev as explained in my message, the issue is with:
```
getUser(parent)
```
I think that with Java 9, the compiler is now smart enough to deduce that, given:
```
SubChild unproxiedChild = initializeAndUnproxy(getUser(parent));
```
The `U` in:
```
public <U extends Child> U getUser(Parent p) {
        return (U) p.getChild();
}
```
will be `SubChild`.

At least, my IDE does that assumption and when I put a breakpoint, it's throwing the exception for the embedded `getUser(parent)` not for the `initializeAndUnproxy` code.

ptahchiev:
Guys,
thank you for your feedback. 
@vladmihalcea I tried with bytebuddy but it didn't work.
@gsmet I refactored the code like this
```
public <T> T initializeAndUnproxy(Object entity) {
```
and now it works fine. It's not a great solution because sometimes I need to explicitly specify the parameter types:
```
getEntityService().<CustomerEntity>initializeAndUnproxy(userService.getCurrentUser()).getUsername()
```
but at least it works.

Once again, thank you for your feedback

PR #2364, opened 2018-06-25
HHH-12594 Using property "hibernate.default_batch_fetch_size" crashes bootstrapping

yrodiere:
https://hibernate.atlassian.net//browse/HHH-12594

Should be backported to 5.2 if it's still maintained (not sure it is).

Commits "HHH-12594 Do not share AliasResolutionContextImpl between EntityLoadQ..." and "
HHH-12594 Properly share AliasResolutionContextImpl between EntityLoa… " can safely be squashed together if we decide to keep the latter.

The first is the safest course of action, but may have a significant impact on memory consumption, while the second is more in line with work done in [HHH-12556](https://hibernate.atlassian.net/browse/HHH-12556) and should preserve all of the memory consumption improvements from that patch.

Travis build is still in progress.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/EntityLoadQueryDetails.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/EntityLoadQueryDetails.java
@@ -87,17 +87,14 @@ public class EntityLoadQueryDetails extends AbstractLoadQueryDetails {
 	protected EntityLoadQueryDetails(
 			EntityLoadQueryDetails initialEntityLoadQueryDetails,
 			QueryBuildingParameters buildingParameters) {
-		super(
+		this(
 				initialEntityLoadQueryDetails.getLoadPlan(),
-				(AliasResolutionContextImpl) initialEntityLoadQueryDetails.getAliasResolutionContext(),
-				buildingParameters,
 				initialEntityLoadQueryDetails.getKeyColumnNames(),
-				initialEntityLoadQueryDetails.getRootReturn(),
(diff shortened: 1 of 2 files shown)


yrodiere:
I just ran the memory consumption tests from https://github.com/alostale/hibernate-test-case-templates/tree/fetch-style. 

Results:

```
# 5.3.1.Final

LEGACY/fetchSize=10: 228 MB - 9662 ms
DYNAMIC/fetchSize=10: 119 MB - 5064 ms
LEGACY/fetchSize=50: 269 MB - 5743 ms
DYNAMIC/fetchSize=50: 120 MB - 4353 ms

# First patch

LEGACY/fetchSize=10: 241 MB - 10899 ms
DYNAMIC/fetchSize=10: 119 MB - 5434 ms
LEGACY/fetchSize=50: 286 MB - 6355 ms
DYNAMIC/fetchSize=50: 120 MB - 4748 ms

# Second patch

LEGACY/fetchSize=10: 228 MB - 9795 ms
DYNAMIC/fetchSize=10: 119 MB - 5107 ms
LEGACY/fetchSize=50: 269 MB - 5955 ms
DYNAMIC/fetchSize=50: 120 MB - 4500 ms
```

So the first patch does introduce a slight increase in memory consumption, but nothing as dramatic as the previous numbers. I guess the issue is mitigated by the lazy loading you introduced.

So yeah, we can keep the second patch for another time.

yrodiere:
@gsmet I removed the last commit. It's still available there: https://github.com/yrodiere/hibernate-orm/tree/shared-alias-context

gsmet:
Looks good, merged, thanks!

PR #2365, opened 2018-06-25
JDK 9 build fixes + documentation fixes

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12724
 * https://hibernate.atlassian.net/browse/HHH-12725
 * https://hibernate.atlassian.net/browse/HHH-12731
 * https://hibernate.atlassian.net/browse/HHH-12732

### DIFF
--- a/documentation/documentation.gradle
+++ b/documentation/documentation.gradle
@@ -180,6 +180,9 @@ task renderTopicalGuides(type: AsciidoctorTask, group: 'Documentation') {
 task renderGettingStartedGuides(type: AsciidoctorTask, group: 'Documentation') {
     description = 'Renders the Getting Started Guides (quick starts) in HTML format using Asciidoctor.'
     sourceDir = file( 'src/main/asciidoc/quickstart/guides' )
+    sources {
+        include 'index.adoc'
+    }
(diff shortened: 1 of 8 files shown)


documentation/documentation.gradle
```diff
@@ -180,6 +180,9 @@ task renderTopicalGuides(type: AsciidoctorTask, group: 'Documentation') {
 task renderGettingStartedGuides(type: AsciidoctorTask, group: 'Documentation') {
     description = 'Renders the Getting Started Guides (quick starts) in HTML format using Asciidoctor.'
     sourceDir = file( 'src/main/asciidoc/quickstart/guides' )
+    sources {
+        include 'index.adoc'
+    }
```
yrodiere: Not sure it's a good idea to do that in a micro, but whatever. Let's at least make sure to add redirections to the 5.3 docs, from the individual files to the main one? We may have links to those documents in the wild, now that they've been published.

gsmet: I don't think we can have redirections on docs.jboss.org.

I still think it's better to remove them, considering they were well hidden and not very readable.

gsmet:
Merged, thanks for the review!

PR #2367, opened 2018-06-26
HHH-12699-HHH-12727 Assorted performance improvements suggested by a user

gsmet:
This PR is coming from 3 issues opened by Roger Dearnaley on JIRA with very interesting observations.  
For HHH-12699, I used a totally different approach than the one suggested. For the other issue, I tuned a bit the suggestions of Roger but it's mostly his work.

As for the third issue (https://hibernate.atlassian.net/browse/HHH-12726), I have withdrawn it as I think it's too risky, considering the case insensitive method of the dialect is not used in the code currently so I'm not sure it's reliable at all. Considering it's set to `true` for MySQL without any concern about the locale, I'm more than worried it wouldn't work very well.

 * https://hibernate.atlassian.net/browse/HHH-12699
 * https://hibernate.atlassian.net/browse/HHH-12727

About HHH-12699, from my tests, my version is 7x faster than the current code. Still far from being transparent but still a lot better. I haven't used ``MethodHandle`` for now as I preferred the code to stay consistent with the rest of the code base. We could revisit it later in 6.

About HHH-12727, it doesn't seem like a bad idea to be more cautious on the map accesses given we don't have the hand on the `Statement` implementations and their `hashCode()` and `equals()` might be costly.

@yrodiere let's see if it's a consensual subject or if you think there are issues/risks but don't waste too much time on it.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/ResultSetWrapperProxy.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/ResultSetWrapperProxy.java
@@ -6,18 +6,22 @@
  */
 package org.hibernate.engine.jdbc;
 
+import static org.hibernate.internal.CoreLogging.messageLogger;
+
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.hibernate.boot.registry.classloading.spi.ClassLoaderService;
 import org.hibernate.engine.jdbc.spi.SqlExceptionHelper;
 import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.service.ServiceRegistry;
-
-import static org.hibernate.internal.CoreLogging.messageLogger;
 
 /**
  * A proxy for a ResultSet delegate, responsible for locally caching the columnName-to-columnIndex resolution that
@@ -31,8 +35,28 @@ public class ResultSetWrapperProxy implements InvocationHandler {
 
 	private static final SqlExceptionHelper SQL_EXCEPTION_HELPER = new SqlExceptionHelper( false );
 
+	private static final Map<ResultSetMethodKey, Method> NAME_TO_INDEX_METHOD_MAPPING;
+
 	private final ResultSet rs;
 	private final ColumnNameCache columnNameCache;
+
+	static {
+		Map<ResultSetMethodKey, Method> nameToIndexMethodMapping = new HashMap<>();
+		for ( Method method : ResultSet.class.getDeclaredMethods() ) {
+			if ( isFirstArgColumnLabel( method ) ) {
+				try {
+					nameToIndexMethodMapping.put(
+							new ResultSetMethodKey( method.getName(), method.getParameterTypes() ),
+							locateCorrespondingColumnIndexMethod( method )
+					);
+				}
+				catch (NoSuchMethodException e) {
+					LOG.unableToSwitchToMethodUsingColumnIndex( method );
+				}
+			}
+		}
+		NAME_TO_INDEX_METHOD_MAPPING = Collections.unmodifiableMap( nameToIndexMethodMapping );
+	}
 
 	private ResultSetWrapperProxy(ResultSet rs, ColumnNameCache columnNameCache) {
 		this.rs = rs;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/engine/jdbc/ResultSetWrapperProxy.java
```diff
@@ -96,20 +119,19 @@ private Integer findColumn(String columnName) throws SQLException {
...
 
-	private boolean isFirstArgColumnLabel(Method method, Object[] args) {
+	private static boolean isFirstArgColumnLabel(Method method) {
 		// method name should start with either get or update
 		if ( ! ( method.getName().startsWith( "get" ) || method.getName().startsWith( "update" ) ) ) {
 			return false;
 		}
 
 		// method should have arguments, and have same number as incoming arguments
-		if ( ! ( method.getParameterCount() > 0 && args.length == method.getParameterCount() ) ) {
```
yrodiere: Not sure why you removed that. If you're trying to fix a bug while you're at it, maybe a separate commit would be better. But I wouldn't dare do that unless I had a very specific reason: I'm not sure there are lots of tests for this class.

Also you should update the comment just above.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/ResultSetWrapperProxy.java
```diff
@@ -96,20 +119,19 @@ private Integer findColumn(String columnName) throws SQLException {
...
+		if ( ! ( method.getParameterCount() > 0 ) ) {
 			return false;
 		}
 
 		// The first argument should be a String (the column name)
-		//noinspection RedundantIfStatement
-		if ( ! ( String.class.isInstance( args[0] ) && method.getParameterTypes()[0].equals( String.class ) ) ) {
+		if ( ! method.getParameterTypes()[0].equals( String.class ) ) {
```
yrodiere: Same here, this seems unrelated to your ticket.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
```diff
@@ -78,18 +81,17 @@ public void register(Statement statement, boolean cancelable) {
...
-		// proxy/wrapper around the JDBC Statement, causing excessive logging here.  See HHH-8210.
-		if ( log.isDebugEnabled() && !xref.containsKey( statement ) ) {
-			log.unregisteredStatement();
-		}
-		else {
-			final Set<ResultSet> resultSets = xref.get( statement );
-			if ( resultSets != null ) {
-				closeAll( resultSets );
-			}
+		final Set<ResultSet> resultSets = xref.get( statement );
```
yrodiere: Why not just `final Set<ResultSet> resultSets = xref.remove( statement );` here, and no `remove()` call below? Then you would only need to access the map once.

That might change the behavior slightly if `closeAll` throws an exception or somehow uses xref, though. Maybe it's not such a good idea.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
```diff
@@ -111,15 +113,17 @@ public void release(ResultSet resultSet, Statement statement) {
...
-			else {
+			if ( resultSets != null ) {
 				resultSets.remove( resultSet );
 				if ( resultSets.isEmpty() ) {
 					xref.remove( statement );
 				}
 			}
+			else if ( log.isDebugEnabled() ) {
+				// Keep this at DEBUG level, rather than warn.  Numerous connection pool implementations can return a
+				// proxy/wrapper around the JDBC Statement, causing excessive logging here.  See HHH-8210.
```
yrodiere: As far as I can tell you're changing the behavior here, not just improving performance... But I guess it's for the best.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
```diff
@@ -78,18 +81,17 @@ public void register(Statement statement, boolean cancelable) {
 	public void release(Statement statement) {
 		log.tracev( "Releasing statement [{0}]", statement );
 
-		// Keep this at DEBUG level, rather than warn.  Numerous connection pool implementations can return a
-		// proxy/wrapper around the JDBC Statement, causing excessive logging here.  See HHH-8210.
-		if ( log.isDebugEnabled() && !xref.containsKey( statement ) ) {
```
yrodiere: The `containsKey` call is only executed when the `DEBUG` level is enabled for ORM loggers.

Are we optimizing the behavior of ORM when using logging at the DEBUG level now? Maybe it's a bit too much... ?

hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
```diff
@@ -201,13 +205,15 @@ public void register(ResultSet resultSet, Statement statement) {
 			}
 		}
 		if ( statement != null ) {
+			Set<ResultSet> resultSets = xref.get( statement );
+
 			// Keep this at DEBUG level, rather than warn.  Numerous connection pool implementations can return a
 			// proxy/wrapper around the JDBC Statement, causing excessive logging here.  See HHH-8210.
-			if ( log.isDebugEnabled() && !xref.containsKey( statement ) ) {
```
yrodiere: Same here, this optimization will only change anything when the `DEBUG` level is enabled.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/ResultSetWrapperProxy.java
gsmet: It's related as I don't have the `args` parameter anymore when taking this decision: it's done only based on the method.

I will update the comment.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/ResultSetWrapperProxy.java
gsmet: Same answer.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
gsmet: Yes, you're right, missed that somehow, I think I will revert this change.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
gsmet: Well, I didn't change anything, it was more to be consistent with the new code above. As I will revert it, I'm gonna revert this part too.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/ResultSetWrapperProxy.java
yrodiere: Got it, the `args` parameter is replaced by the `parameterTypes` member in 
 `ResultSetMethodKey`.

yrodiere:
Ok, it seems safer now. Considering our current... situation (trying to fix regressions), before we merge anything that's not a fix, I'd rather have an actual benchmark tell me "yes, this improves performance in real-life scenarios", i.e. it has a significant impact to the *overall* query execution time, be it only in very specific cases.

While I agree these changes should bring performance improvements, I'm not totally convinced these improvements (probably nanosecond- to microsecond-scale) will be of any significance compared to the huge (millisecond-scale) execution times of DB queries sent over the network.

However, the improvements may be significant if the code you're improving is directly in the hot path and called hundreds or thousands of times for each query. So if you think it does change something, then ok, let's merge it.

gsmet:
I think the `ResultSet` trick is worth it: having reflection in a hot path is a very bad idea.

As for the other one, the diminished patch doesn't sound risky and is just good practice so I'll merge it too.

Thanks for the review.

gsmet:
Merged!

PR #2368, opened 2018-06-26
HHH-12630 Keep the original starting exception and use it as a cause

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12630

@yrodiere it's either that or I log the exception directly with the logger. What's your take on this?


gsmet:
Yes, not a big fan either but I suppose there's a good reason for that.

gsmet:
Merged, thanks!

PR #2369, opened 2018-06-26
HHH-12649 Auto-register entity and collection caches based on the Hibernate @Cache annotation (XML mapping) settings

yrodiere:
https://hibernate.atlassian.net//browse/HHH-12649

As discussed on HipChat, this:

* Introduces configuration properties allowing users to decide how the region factory should react when a cache has to be created "on-the-fly", i.e. when the cache manager does not have a pre-configured cache right from the start. Strategies are:
  * `fail`, the current behavior, throwing an exception
  * `create`, the old behavior before [HHH-11953](https://hibernate.atlassian.net/browse/HHH-11953)
  * `create-warn`, a new, more user-friendly behavior that does not trigger an exception, but creates the cache and logs a warning.
* Changes the default behavior for the `ehcache` and `ehcache-singleton` region factories to `create-warn`, to avoid breaking backward compatibility with applications based on ORM 5.1.
* Adds previously missing tests and documentation.

This solves [HHH-12649](https://hibernate.atlassian.net//browse/HHH-12649).

[HHH-12531](https://hibernate.atlassian.net/browse/HHH-12531), while related, involves some other annoying, breaking changes, which I discussed [in a comment](https://hibernate.atlassian.net/browse/HHH-12531?focusedCommentId=102898&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-102898). Let's try to solve it in a second PR.


hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/internal/EhcacheRegionFactory.java
```diff
@@ -122,9 +136,11 @@ protected void prepareForUse(SessionFactoryOptions settings, Map configValues) {
 			if ( this.cacheManager == null ) {
 				throw new CacheException( "Could not start Ehcache CacheManager" );
 			}
+			this.missingCacheStrategy = MissingCacheStrategy.interpretSetting(
+					configValues.get( ConfigSettings.MISSING_CACHE_STRATEGY )
+			);
 		}
 	}
-
```
gsmet: This line shouldn't be removed, should it?

hibernate-jcache/src/main/java/org/hibernate/cache/jcache/internal/JCacheRegionFactory.java
```diff
@@ -83,7 +88,19 @@ protected DomainDataStorageAccess createDomainDataStorageAccess(
 
 	@SuppressWarnings("WeakerAccess")
 	protected Cache<Object, Object> createCache(String regionName) {
-		throw new CacheException( "On-the-fly creation of JCache Cache objects is not supported [" + regionName + "]" );
+		switch ( missingCacheStrategy ) {
+			case CREATE_WARN:
+				LOG.missingCacheCreated( regionName );
+				cacheManager.createCache( regionName, new MutableConfiguration<>() );
```
gsmet: I think `createCache()` returns the cache so you don't have to get it back.

hibernate-jcache/src/test/java/org/hibernate/jcache/test/MissingCacheStrategyTest.java
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	private void doTestMissingCacheStrategyFail(Consumer<StandardServiceRegistryBuilder> additionalSettings) {
+		// first, lets make sure that the region names we think are non-existent really do not exist
+		for ( String regionName : TestHelper.allRegionNames ) {
+			assertThat( TestHelper.getCache( regionName ), nullValue() );
+		}
+
+		// and now let's try to build the standard testing SessionFactory, without pre-defining caches
+		try (SessionFactoryImplementor ignored = TestHelper.buildStandardSessionFactory( additionalSettings ) ) {
```
gsmet: Missing padding after `try (`

hibernate-jcache/src/test/java/org/hibernate/jcache/test/MissingCacheStrategyTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+	@Test
+	public void testMissingCacheStrategyCreate() {
+		// first, lets make sure that the region names we think are non-existent really do not exist
+		for ( String regionName : TestHelper.allRegionNames ) {
+			assertThat( TestHelper.getCache( regionName ), nullValue() );
+		}
+
+		// and now let's try to build the standard testing SessionFactory, without pre-defining caches
+		try (SessionFactoryImplementor ignored = TestHelper.buildStandardSessionFactory(
```
gsmet: Same here about the padding.

hibernate-ehcache/src/test/java/org/hibernate/cache/ehcache/test/MissingCacheStrategyTest.java
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	private void doTestMissingCacheStrategyFail(Consumer<StandardServiceRegistryBuilder> additionalSettings) {
+		/*
+		 * The cache manager is created per session factory, and we don't use any specific ehcache configuration,
+		 * so we know the caches don't exist before we start the session factory.
+		 */
+
+		// let's try to build the standard testing SessionFactory, without pre-defining caches
+		try (SessionFactoryImplementor ignored = TestHelper.buildStandardSessionFactory( additionalSettings ) ) {
```
gsmet: Padding.

Same below.

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/MissingCacheStrategy.java
```diff
@@ -45,7 +45,8 @@ public static MissingCacheStrategy interpretSetting(Object value) {
 
 		if ( StringHelper.isEmpty( externalRepresentation ) ) {
 			// Use the default
-			return MissingCacheStrategy.FAIL;
+			// Default is CREATE_WARN for backward compatibility reasons; we should switch to FAIL at some point.
+			return MissingCacheStrategy.CREATE_WARN;
```
gsmet: OK, I think even having warnings for a perfectly valid configuration is not acceptable.

We should probably discuss that with the others.

documentation/src/main/asciidoc/userguide/chapters/caching/Caching.adoc
```diff
@@ -615,6 +645,33 @@ shared among multiple `SessionFactory` instances in the same JVM.
 http://www.ehcache.org/documentation/2.8/integrations/hibernate#optional[Ehcache documentation] recommends using multiple non-singleton `CacheManager(s)` when there are multiple Hibernate `SessionFactory` instances running in the same JVM.
 ====
 
+[[caching-provider-ehcache-missing-cache-strategy]]
+==== JCache missing cache strategy
```
gsmet: s/JCache/Ehcache/

hibernate-ehcache/src/main/java/org/hibernate/cache/ehcache/MissingCacheStrategy.java
yrodiere: It's a warning that you can disable forever with a configuration property, and we don't know that the configuration is valid, it depends on the JCache implementation.
Note that if you don't want warnings either by default, then the whole pull request and the changes we discussed become useless, since we could just have switched the behavior to "create" and not provided any configuration option.

hibernate-ehcache/src/test/java/org/hibernate/cache/ehcache/test/MissingCacheStrategyTest.java
yrodiere: Done.

yrodiere:
Applied fixes and rebased.

gsmet:
Merged, thanks!

PR #2370, opened 2018-06-26
HHH-12640 2nd ed

Sanne:
@scottmarlow re-worked your PR a bit as a follow up of our other chat. You like this approach?


Sanne:
cool, I confirmed OGM 's testsuite is also able to run w/o all the changes the previous version was requiring... Merging.

PR #2371, opened 2018-06-26
HHH-12531 JCache existing cache not detected.

yrodiere:
https://hibernate.atlassian.net//browse/HHH-12531

Based on #2369, which should be merged first.

This simply reverts the name of the caches to their pre-5.3 value.


yrodiere:
Superseded by #2388

PR #2373, opened 2018-06-27
HHH-12741 Register new reserved words added in mysql 8.0

quaff:
https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-new-8-0-A

https://hibernate.atlassian.net/browse/HHH-12741

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/MySQL8Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/MySQL8Dialect.java
@@ -13,6 +13,26 @@ import org.hibernate.LockOptions;
  */
 public class MySQL8Dialect extends MySQL57Dialect {
 
+	public MySQL8Dialect() {
+		registerKeyword("CUME_DIST");
+		registerKeyword("DENSE_RANK");
+		registerKeyword("EMPTY");
+		registerKeyword("EXCEPT");
+		registerKeyword("FIRST_VALUE");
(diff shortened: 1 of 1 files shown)


gsmet:
@quaff thanks for the PR!

Could you create a JIRA issue here https://hibernate.atlassian.net/ and prepend your commit message with the JIRA number?

Thanks!

gsmet:
@vladmihalcea I'm surprised we don't have any reserved keywords for MySQL right now as they have a lot of them. Do you happen to know why?

vladmihalcea:
@sebersole Do you happen to know the answer to @gsmet 's question above?

quaff:
It should handled by `DatabaseMetaData.getSQLKeywords()`,  but mysql jdbc driver `com.mysql.cj.jdbc.DatabaseMetaData.ProcedureType.MYSQL_KEYWORDS` doesn't add new reserved words in order to keep back compatibility

gsmet:
@quaff interesting, thanks!

Is there a reason why you haven't added all of them? I see `WINDOW (R)` for instance that is missing in your PR.

vladmihalcea:
The SQL keyword got from the DB metadata got to the `IdentifierHelper`, but we need to check if they are available to queries as well. Maybe that's a problem too.

gsmet:
@quaff I merged your PR, thanks. Will be included in the upcoming 5.3.2.

PR #2374, opened 2018-06-27
HHH-12738 - Session/EntityManager is closed in ForeignGenerator (JTA setup)

mjesseni:
I just added the test case closing the entity manager, before the commit phase is stared. Personally, I would never close the entity manager by hand. This is just to simulate what Spring is doing when commiting a transaction. First it closes the resource (the entity manager) and afterwards performing the commit.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/idgen/foreign/ForeignGeneratorResourceLocalTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/idgen/foreign/ForeignGeneratorResourceLocalTest.java
@@ -1,8 +1,6 @@
 /*
- * Hibernate, Relational Persistence for Idiomatic Java
- *
- * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
- * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ * Hibernate, Relational Persistence for Idiomatic Java License: GNU Lesser General Public License (LGPL), version 2.1
+ * or later. See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
 package org.hibernate.test.idgen.foreign;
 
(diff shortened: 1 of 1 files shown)


hibernate-core/src/test/java/org/hibernate/test/idgen/foreign/ForeignGeneratorResourceLocalTest.java
```diff
@@ -36,275 +34,366 @@
...
+                Customer.class)
+                .setParameter("customerId", customerId)
+                .getSingleResult();
+
+            CustomerContractRelation relation = new CustomerContractRelation();
+            relation.setContractId(contractId);
+            customer.addContractRelation(relation);
+
+            // simulating spring JtaTransactionManager.triggerBeforeCompletion()
+            entityManager.close();
```
vladmihalcea: That's not a valid workflow. Do you want to span a Tx over multiple EntityManagers?

mjesseni: Hi Vlad! Thx for your quick reply. I am totally aware about that and I would not close the session as well. However, this is how Spring does the integration. 

Well, within the project I am working in, we have multiple entity managers and spawning Tx over them using Bitronix as XA transaction manager. This was the setup I originaly wanted to provid by my test case including the spring and bitronix dependencies. 

As a workaround for this issue I am able to manually flush the entity manager with the business code and everything is going well. Only if a left this work to the environment (spring), I run into this problems. However, the workaround seems to me not very practicable at all.

I am not sure how to proceed. As already mentioned, spring contributors are quite sure, that closing the entity manager before starting the commit phase should be no problem at all. And of cause, it seems to be no problem except the situation in the ForeignGenerator. 

Thx for your time and patience!

vladmihalcea:
I rewrote your PR and could replicate the issue.

hibernate-core/src/test/java/org/hibernate/test/idgen/foreign/ForeignGeneratorResourceLocalTest.java
vladmihalcea: @mjesseni I managed to replicate it, so I reopened it. I'll propose a fix today.

vladmihalcea:
Replaced by [this PR](https://github.com/hibernate/hibernate-orm/pull/2377) which provides a fix as well.

hibernate-core/src/test/java/org/hibernate/test/idgen/foreign/ForeignGeneratorResourceLocalTest.java
mjesseni: @vladmihalcea thx a lot!

PR #2375, opened 2018-06-27
HHH-12718 - Entity changes in @PreUpdate callback are not persisted when lazy loading is active for more than one field

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12718

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultFlushEntityEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultFlushEntityEventListener.java
@@ -510,7 +510,11 @@ public class DefaultFlushEntityEventListener implements FlushEntityEventListener
 		final Serializable id = entry.getId();
 		final Object[] loadedState = entry.getLoadedState();
 
-		int[] dirtyProperties = session.getInterceptor().findDirty(
+		int[] dirtyProperties;
+
+		// First try: use the interceptor
+		final boolean interceptorHandledDirtyCheck;
+		dirtyProperties = session.getInterceptor().findDirty(
 				entity,
 				id,
 				values,
@@ -518,7 +522,15 @@ public class DefaultFlushEntityEventListener implements FlushEntityEventListener
 				persister.getPropertyNames(),
 				persister.getPropertyTypes()
 		);
+		if ( dirtyProperties != null ) {
+			// the Interceptor handled the dirty checking
+			interceptorHandledDirtyCheck = true;
+		}
+		else {
+			interceptorHandledDirtyCheck = false;
+		}
 
+		// If the above did not work, try custom dirtiness strategies and bytecode enhancement
 		if ( dirtyProperties == null ) {
 			if ( entity instanceof SelfDirtinessTracker ) {
 				if ( ( (SelfDirtinessTracker) entity ).$$_hibernate_hasDirtyAttributes() ) {
@@ -564,15 +576,14 @@ public class DefaultFlushEntityEventListener implements FlushEntityEventListener
 
 		event.setDatabaseSnapshot( null );
 
-		final boolean interceptorHandledDirtyCheck;
-		boolean cannotDirtyCheck;
+		// The dirty check is considered possible unless proven otherwise (see below)
+		boolean cannotDirtyCheck = false;
 
+		// If none the above worked, try to compute dirtiness based on entity or database snapshots
 		if ( dirtyProperties == null ) {
-			// Interceptor returned null, so do the dirtycheck ourself, if possible
 			try {
 				session.getEventListenerManager().dirtyCalculationStart();
 
-				interceptorHandledDirtyCheck = false;
 				// object loaded by update()
 				cannotDirtyCheck = loadedState == null;
 				if ( !cannotDirtyCheck ) {
@@ -612,11 +623,6 @@ public class DefaultFlushEntityEventListener implements FlushEntityEventListener
 			finally {
 				session.getEventListenerManager().dirtyCalculationEnd( dirtyProperties != null );
 			}
-		}
-		else {
-			// the Interceptor handled the dirty checking
-			cannotDirtyCheck = false;
-			interceptorHandledDirtyCheck = true;
 		}
 
 		logDirtyProperties( id, dirtyProperties, persister );
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultFlushEntityEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultFlushEntityEventListener.java
@@ -332,15 +332,20 @@ public class DefaultFlushEntityEventListener implements FlushEntityEventListener
 		final boolean intercepted = invokeInterceptor( session, entity, entry, values, persister );
 
 		//now we might need to recalculate the dirtyProperties array
-		if ( intercepted && event.isDirtyCheckPossible() && !event.isDirtyCheckHandledByInterceptor() ) {
-			int[] dirtyProperties;
-			if ( event.hasDatabaseSnapshot() ) {
-				dirtyProperties = persister.findModified( event.getDatabaseSnapshot(), values, entity, session );
+		if ( intercepted && event.isDirtyCheckPossible() ) {
+			if ( !event.isDirtyCheckHandledByInterceptor() ) {
+				int[] dirtyProperties;
+				if ( event.hasDatabaseSnapshot() ) {
+					dirtyProperties = persister.findModified( event.getDatabaseSnapshot(), values, entity, session );
+				}
+				else {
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/event/internal/DefaultFlushEntityEventListener.java
```diff
@@ -332,15 +332,8 @@ protected boolean handleInterception(FlushEntityEvent event) {
...
-		if ( intercepted && event.isDirtyCheckPossible() && !event.isDirtyCheckHandledByInterceptor() ) {
-			int[] dirtyProperties;
-			if ( event.hasDatabaseSnapshot() ) {
-				dirtyProperties = persister.findModified( event.getDatabaseSnapshot(), values, entity, session );
-			}
-			else {
-				dirtyProperties = persister.findDirty( values, entry.getLoadedState(), entity, session );
-			}
-			event.setDirtyProperties( dirtyProperties );
+		if ( intercepted && event.isDirtyCheckPossible() ) {
```
JWGmeligMeyling: Why not `!event.isDirtyCheckHandledByInterceptor()`? Doesn't that effectively disable the dirty check enhancement?

vladmihalcea: Why would that disable the dirty check enhancement? The logic for which dirty checking mechanism to use is inside `dirtyCheck(final FlushEntityEvent event)` method.

gsmet: I think I agree with @JWGmeligMeyling . In this case, it's the second time you are calling `dirtyCheck()` so you might as well use the flags that are in the `event` to avoid calling `dirtyCheck()` as it has already been decided that dirty checking is using an interceptor.

Or did we miss something and the test fails when you keep ` && !event.isDirtyCheckHandledByInterceptor()`?

vladmihalcea: Exactly. The test fails if we add the check. So, we can either call dirtyChecl or duplicate it's logic. It's not just bytecode enhancement to consider. We need to consider the custom mechanisms and the interceptor as well because the user might use a custom mechanism to determine that the entity was changed by the call back.

gsmet: Thanks for the explanation. Looks good to me, please merge!

I think we should backport it to 5.2.18 too.

vladmihalcea: I'm thinking of adding, in case we don't already have, a test for the custom dirty checking mechanism/Session interceptor to make sure we test that and the logic is covered.

vladmihalcea:
@JWGmeligMeyling @gsmet As promised, I added 2 more tests for Session `Interceptor` and `CustomEntityDirtinessStrategy`. It turned out the issue could be replicated for Session `Interceptor` with `DynamicUpdate` and for `CustomEntityDirtinessStrategy` since we don't recalculate the `dirtyProperties` array.

I kept the old logic when `isDirtyCheckHandledByInterceptor` is false and just called the `dirtyCheck(event)` for when dirty checking is handled by "interceptor" (which is a bad name anyway since this can happen either via bytecode enhancement or CustomEntityDirtinessStrategy).

Let me know what you think. If you have other suggestions how we can better hadle all these dirty checking strategies, just let me know.

gsmet:
@vladmihalcea I'll take a look to this one later today.

gsmet:
@vladmihalcea please do not merge this one yet. I have my doubts about whether it's the right fix or not, especially considering that `event.isDirtyCheckHandledByInterceptor()` returns an incorrect information right now e.g. it might be true even if the dirty check is not handled by an interceptor (if we are in the enhanced case). I think the enhanced code was added at a later point without considering the consequences on the `isDirtyCheckHandledByInterceptor()` flag.

I think we will need more/different changes. I asked @yrodiere to take a closer look later this week, we will get back at you.

vladmihalcea:
No problem. I'll leave it as is for the moment.

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultFlushEntityEventListener.java
```diff
@@ -332,15 +332,20 @@ protected boolean handleInterception(FlushEntityEvent event) {
...
+				else {
+					dirtyProperties = persister.findDirty( values, entry.getLoadedState(), entity, session );
+				}
+				event.setDirtyProperties( dirtyProperties );
 			}
 			else {
-				dirtyProperties = persister.findDirty( values, entry.getLoadedState(), entity, session );
+				dirtyCheck( event );
 			}
-			event.setDirtyProperties( dirtyProperties );
```
yrodiere: This set of changes seems to fix the bug by ensuring a second dirty check is performed after a `@PreUpdate`, and even fixes an additional bug: it makes sure that, after a `FlushEntityEvent` is intercepted, bytecode enhancement or the custom dirtiness strategy (if any) is used to compute dirtiness instead of `persister.findModified` or `persister.findDirty` like we used to do.
However it's partly because another bug offsets problems in this implementation, and this "additional bug" might very well resurface later... So I'd like to propose going one step further.

But before that, a little explaining is necessary:

As Guillaume pointed out to me, the flag `isDirtyCheckHandledByInterceptor` is set to true when using bytecode enhancement or a custom dirtiness strategy: see in `dirtyCheck` how the `interceptorHandledDirtyCheck` variable is set to `true` if the properties are non-null after the bytecode enhancement/custom strategy block.

As a result, the new implementation of `handleInterception` in this patch will call `dirtyCheck` not only when an interceptor handled the previous dirty check, but also when bytecode enhancement or a custom dirty strategy handled the dirty check. Which is good... but it's because of a bug!

I pushed a few commits to this PR to demonstrate, and fix, the issue.

Commit 1: I fixed the `isDirtyCheckHandledByInterceptor`. The bytecode enhancement tests still pass, but only because, after the `@PreUpdate` is applied, we default to calling `persister.findDirty`, and it happens to work just as well as bytecode enhancement. Not sure it's a good idea to do that, though...
Commit 2: I added a test checking that, when using a custom dirtiness strategy, the strategy is used to compute dirtiness both before and after the `@PreUpdate`. The test fails.
Commit 3: I changed the `handleInterception` method to call the `dirtyCheck` method directly when dirty checking is necessary, instead of trying to inline an incomplete version of `dirtyCheck` inside `handleInterception`.
Commit 4: just a small refactorying clarifying a variable name.

Commit 3 is the core of the proposal: since we want to perform dirty checking a second time, let's just call `dirtyCheck`, which after all is here just for that. I think the behavior will be more correct, and from what I can see, there is little risk of experiencing a performance hit for the two cases that were already working:
 * the load state (used in `persister.findDirty`) is available as part of the entity entry and doesn't require any database call.
 * the database state (used in `persister.findModified`) might be loaded from the database, we're talking about the second pass (after the `@PreUpdate`), so if this method is being used, then it was already used during the first pass (before the `@PreUpdate`), and `org.hibernate.engine.internal.StatefulPersistenceContext#getDatabaseSnapshot` relies on a cache. So we won't be fetching data from the database on the second pass.

@gsmet , @vladmihalcea , WDYT?

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultFlushEntityEventListener.java
gbadner: I agree this change is correct.

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultFlushEntityEventListener.java
```diff
@@ -505,15 +498,27 @@ protected void dirtyCheck(final FlushEntityEvent event) throws HibernateExceptio
...
 				persister.getPropertyNames(),
 				persister.getPropertyTypes()
 		);
+		if ( dirtyProperties != null ) {
+			// the Interceptor handled the dirty checking
+			interceptorHandledDirtyCheck = true;
+		}
+		else {
+			interceptorHandledDirtyCheck = false;
+		}
```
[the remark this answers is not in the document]
gbadner: IMO, this method should be left as is.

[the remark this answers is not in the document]
gsmet: I kept the removal of the double negation as the logic is very hard to follow with it.

I removed the commit changing the meaning of `dirtyCheckHandledByInterceptor`. Personally I think the meaning has been changed by mistake when the bytecode enhancement was added but we never know what users are doing with it. As it's not used anymore by us, let's be conservative. I added a comment to explain its true meaning.

PR #2376, opened 2018-06-27
Fix and tests for HHH-12247

Sanne:
This makes sure that critical bootstrap errors are not just logged but propagated as runtime exceptions.

 - https://hibernate.atlassian.net/browse/HHH-12247

It's trivial, yet requires @sebersole 's approval as the semantics had quite some discussion (see JIRA).


Sanne:
thanks! Rebasing and re-testing...

PR #2377, opened 2018-06-27
HHH-12738 - Session/EntityManager is closed in ForeignGenerator (JTA setup)

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12738

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
@@ -2107,7 +2107,7 @@ public final class SessionImpl
 
 	@Override
 	public boolean contains(String entityName, Object object) {
-		checkOpen();
+		checkOpenOrWaitingForAutoClose();
 		checkTransactionSynchStatus();
 
 		if ( object == null ) {
--- a/hibernate-core/src/test/java/org/hibernate/test/idgen/foreign/ForeignGeneratorJtaTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/idgen/foreign/ForeignGeneratorJtaTest.java
@@ -6,35 +6,12 @@
  */
 package org.hibernate.test.idgen.foreign;
 
-import java.io.Serializable;
-import java.util.Date;
-import java.util.HashSet;
 import java.util.Map;
-import java.util.Set;
-import javax.persistence.CascadeType;
-import javax.persistence.Column;
-import javax.persistence.Embeddable;
-import javax.persistence.EmbeddedId;
-import javax.persistence.Entity;
-import javax.persistence.FetchType;
-import javax.persistence.GeneratedValue;
-import javax.persistence.Id;
-import javax.persistence.JoinColumn;
-import javax.persistence.ManyToOne;
-import javax.persistence.MapsId;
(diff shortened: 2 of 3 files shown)


gsmet:
@vladmihalcea looks good to me. There might still be an issue in the case of a proxy but I don't think we can do any better.

Let's merge it and to 5.2 too.

Thanks!

vladmihalcea:
The Proxy use case should be the only one guarded now. We could add another test, just to be sure.

hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
```diff
@@ -2107,7 +2107,6 @@ public boolean contains(Object object) {
 
 	@Override
 	public boolean contains(String entityName, Object object) {
-		checkOpen();
```
dreab8: what about changing to checkOpenOrWaitingForAutoClose()?

vladmihalcea: @dreab8 Good point. It works with `checkOpenOrWaitingForAutoClose`.

gsmet: @dreab8 we don't need it to be open in the proxy case?

vladmihalcea: We also need to check more use cases since `checkOpenOrWaitingForAutoClose` is tied to JPA bootstrap only, which happens to be covered by the current unit tests. I'll add some tests which bootstrap Hibernate natively so we can make sure we cover that use case as well.

vladmihalcea: Well, we can't test it for native bootstrap since the Session is closed anyway and the undelrying Connetcion as well.

So, the current solution uses `checkOpenOrWaitingForAutoClose` like many other Sesison methods. If anyone agrees, we can integrate it as is since all tests pass just fine.

dreab8: In case of Session it have to fail so if this happens I think it should work.

vladmihalcea: I don't understand what exactly you ate referring to. Everything happen inside the SessionImpl class. Are you saying that it's fine to fail for Hibernate native bootstrap?

dreab8: Yes it is what I mean.

dreab8: Anyway I want to investigate a little more the issue, so please wait before merging.

vladmihalcea: Sure thing.

dreab8: @vladmihalcea It seems good to me.

vladmihalcea: @dreab8 Cool. Then I'll integrate it into master and 5.2.

PR #2378, opened 2018-06-28
HHH-12729 : Binary and behavioral incompatibilities of org.hibernate.…

gbadner:
…Query.getFirstResult(), setFirstResult(), getMaxResults(), setMaxResults()

I still need to add a commit that adds documentation to the migration guide.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/Query.java
+++ b/hibernate-core/src/main/java/org/hibernate/Query.java
@@ -92,7 +92,7 @@ public interface Query<R> extends TypedQuery<R>, CommonQueryContract {
 	 * @see #setFirstResult(int)
 	 * @see #setHibernateFirstResult(int)
 	 *
-	 * @deprecated {@link #setFirstResult(int)} should be used instead.
+	 * @deprecated {@link #getFirstResult()} should be used instead.
 	 */
 	@Deprecated
 	default Integer getHibernateFirstResult() {
@@ -132,10 +132,10 @@ public interface Query<R> extends TypedQuery<R>, CommonQueryContract {
 	 *
 	 * @return the maximum number of query results, or {@code null} if uninitialized.
 	 *
-	 * @see #setFirstResult(int)
-	 * @see #setHibernateFirstResult(int)
+	 * @see #setMaxResults(int) (int)
+	 * @see #setHibernateMaxResults(int) (int)
 	 *
-	 * @deprecated {@link #setFirstResult(int)} should be used instead.
+	 * @deprecated {@link #setMaxResults(int)} should be used instead.
 	 */
 	@Deprecated
 	default Integer getHibernateMaxResults() {
--- a/hibernate-core/src/main/java/org/hibernate/Query.java
+++ b/hibernate-core/src/main/java/org/hibernate/Query.java
@@ -22,6 +22,7 @@ import javax.persistence.Parameter;
 import javax.persistence.TemporalType;
 import javax.persistence.TypedQuery;
 
+import org.hibernate.engine.spi.RowSelection;
 import org.hibernate.query.CommonQueryContract;
 import org.hibernate.query.ParameterMetadata;
 import org.hibernate.query.QueryParameter;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/Query.java
```diff
@@ -64,6 +65,106 @@
...
+	 * "QueryOptions" is a better name, I think, than "RowSelection" -> 6.0
+	 *
+	 * @todo 6.0 rename RowSelection to QueryOptions
+	 *
+	 * @return Return the encapsulation of this query's options, which includes access to
+	 * firstRow, maxRows, timeout and fetchSize.   Important because this gives access to
+	 * those values in their Integer form rather than the primitive form (int) required by JPA.
+	 */
+	RowSelection getQueryOptions();
+
```
gbadner: I really didn't want to move ```#getQueryOptions``` from ```org.hibernate.query.Query``` to ```org.hibernate.Query```, but it was the easiest way to get this to work.

hibernate-core/src/main/java/org/hibernate/Query.java
```diff
@@ -64,6 +65,106 @@
...
+	 * <p/>
+	 * If {@link #setHibernateFirstResult(int)} was called with a negative value, then 0
+	 * is returned.
+	 *
+	 * @return the position of the first query result, or {@code null} if uninitialized.
+	 *
+	 * @see #setFirstResult(int)
+	 * @see #setHibernateFirstResult(int)
+	 *
+	 * @deprecated {@link #setFirstResult(int)} should be used instead.
```
gsmet: This comment is incorrect.

hibernate-core/src/main/java/org/hibernate/Query.java
```diff
@@ -64,6 +65,106 @@
...
+	 * If {@link #setHibernateMaxResults(int)} was called with a value less than
+	 * or equal to 0, the value is considered to be uninitialized, and {@code null}
+	 * is returned, resulting in no limit on the number of results.
+	 *
+	 * @return the maximum number of query results, or {@code null} if uninitialized.
+	 *
+	 * @see #setFirstResult(int)
+	 * @see #setHibernateFirstResult(int)
+	 *
+	 * @deprecated {@link #setFirstResult(int)} should be used instead.
```
gsmet: This comment is incorrect.

hibernate-core/src/main/java/org/hibernate/Query.java
```diff
@@ -64,6 +65,106 @@
...
+	 * {@link #setHibernateMaxResults(int)}, then {@code null} is returned
+	 * <p/>
+	 * If {@link #setHibernateMaxResults(int)} was called with a value less than
+	 * or equal to 0, the value is considered to be uninitialized, and {@code null}
+	 * is returned, resulting in no limit on the number of results.
+	 *
+	 * @return the maximum number of query results, or {@code null} if uninitialized.
+	 *
+	 * @see #setFirstResult(int)
+	 * @see #setHibernateFirstResult(int)
```
gsmet: We should talk about MaxResults here.

Sanne:
> Or do you plan to use these methods with the WildFly bytecode transformer @Sanne was talking about in the issue?

@gsmet yes that's the plan. There are matching PRs being worked on by Scott which on the WildFly side of things.
I'm not too happy that such changes are tested in WildFly only so I had started with a way to run integration tests in ORM but Steve had concerns about the gradle changes so those integration tests got postponed - for now I guess we'll have to check on Scott's tests, at least there is some progress.

I believe @scottmarlow 's work is here, not aware of a real PR yet - it's probably too soon to review.
 - https://github.com/scottmarlow/wildfly/tree/WFLY-10520_hibernatecompatabilitytransformer

gsmet:
@Sanne OK, cool, thanks for sharing. So, We just need to fix the comments and it looks ready to be merged.

hibernate-core/src/main/java/org/hibernate/Query.java
Sanne: What about people using Hibernate ORM 5.3.0, will this break their code?
Should we keep both and deprecate one?

gsmet: `org.hibernate.query.Query` inherits from `org.hibernate.Query`. So Gail just moved this method one level up.

It shouldn't be an issue (except for people overriding `org.hibernate.Query` of course but I don't think it's a common use case).

hibernate-core/src/main/java/org/hibernate/Query.java
```diff
@@ -64,6 +65,106 @@
...
+	 * @deprecated {@link #setFirstResult(int)} should be used instead.
+	 */
+	@Deprecated
+	default Integer getHibernateMaxResults() {
+		return getQueryOptions().getMaxRows();
+	}
+
+	/**
+	 * Set the maximum number of query results to be retrieved. A value less than
+	 * or equal to 0 is considered uninitialized, resulting in no limit on the number
```
Sanne: This whole thing that "0" means "all" was clearly a bug in previous versions. I think it's ok to try to maintain the semantics, but should we really commit this promise into documentation? Not sure if sets the wrong expectations.

hibernate-core/src/test/java/org/hibernate/test/hql/HibernateFirstResultMaxResultsTest.java
```diff
@@ -0,0 +1,145 @@
...
+				this::sessionFactory,
+				session -> {
+					for ( int i = 0; i < 5; i++ ) {
+						session.persist( new Employee( i ) );
+					}
+				}
+		);
+
+		final String query = "from Employee order by id";
+		checkResults( executeQuery( query, null, null ), 0, 4 );
```
Sanne: Could we add some tests about the new API rightfully interpreting the "maxresults=0" case as a need to return zero results?

gsmet: Don't we already have some somewhere else - I suppose they have been added when the new API was introduced? This test is about validating the compatibility, I don't think we should add tests about the new API here.

hibernate-core/src/main/java/org/hibernate/Query.java
Sanne: ok, thanks for checking!

hibernate-core/src/test/java/org/hibernate/test/hql/HibernateFirstResultMaxResultsTest.java
Sanne: Agreed it's orthogonal, yet this whole mess of needing to intepret zero as all is caused by that being the semantics before. Whish is clearly wrong, and I'd like us to make sure it never happens again.

(So I'm not sure if that's tested anywhere else as clearly we had it wrong)

+1 to manage it as a follow up JIRA if that's easier, didn't mean to "block" this one.

scottmarlow: In addition to https://github.com/scottmarlow/wildfly/tree/WFLY-10520_hibernatecompatabilitytransformer, in the jira description of https://issues.jboss.org/browse/WFLY-10520, we are trying to keep current, the list of transformations to make.

hibernate-core/src/main/java/org/hibernate/Query.java
gbadner: @Sanne, why do you think it was a bug?

gbadner: In any case, this method is deprecated, so we are only committing to it until it is removed.

gbadner:
I've pushed a commit that fixes the Javadoc comments @gsmet mentioned.

hibernate-core/src/test/java/org/hibernate/test/hql/HibernateFirstResultMaxResultsTest.java
gbadner: There are a couple of tests that show that ```javax.persistence.Query#setMaxResults``` works as it should. Here are tests for ```setMaxResults( 0 )```:
* using CriteriaQuery -- [LimitExpressionTest](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/limitExpression/LimitExpressionTest.java#L36-L45)
* using Query -- [LimitExpressionTest](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/test/java/org/hibernate/test/legacy/LimitExpressionTest.java#L33-L51)

hibernate-core/src/main/java/org/hibernate/Query.java
gbadner: @Sanne, IMO, for existing applications, the difference between all results and no results is too much of a behavioral difference.

An alternative that I would find acceptable (not sure about anyone else) is for Hibernate to log a warning if ```setMaxResults( 0 )``` is called on an object that does *not* extend ```java.persistence.Query```. WDYT?

gsmet:
@gbadner looks good to me! Let's merge it for 5.3.2 so that Scott can leverage it in his ongoing patch!

gbadner:
Merged.

PR #2379, opened 2018-06-28
fixed java.lang.StringIndexOutOfBoundsException

hone-liu:
fixed java.lang.StringIndexOutOfBoundsException where value is an empty string

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/CharacterTypeDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/CharacterTypeDescriptor.java
@@ -58,7 +58,7 @@ public class CharacterTypeDescriptor extends AbstractTypeDescriptor<Character> {
 		}
 		if ( String.class.isInstance( value ) ) {
 			final String str = (String) value;
-			if(str.length()==0){
+			if(str.length() == 0){
 				return null;
 			}
 			return str.charAt( 0 );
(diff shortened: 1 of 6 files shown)


gsmet:
Hi @hone-liu ,

Thanks for this PR. Do you have a reproducible test case by any chance? That would help avoiding regression in the future.

To be honest, I'm a bit surprised you can send an empty string as a valid value for a char.

JWGmeligMeyling:
Not sure if this change is the right way to achieve this. @vladmihalcea has released a custom type for this: https://vladmihalcea.com/how-to-implement-a-custom-basic-type-using-hibernate-usertype/ . Maybe he can shed some light on this.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/CharacterTypeDescriptor.java
```diff
@@ -58,6 +58,9 @@ public Character fromString(String string) {
 		}
 		if ( String.class.isInstance( value ) ) {
 			final String str = (String) value;
+			if(str.length() == 0){
+				return null;
```
vladmihalcea: Shouldn't we use `Character.MIN_VALUE` instead?

gsmet:
Yes, I must admit I'm not very convinced. That's why I would like to see a proper use case.

beikov:
Closing due to the lack of a JIRA issue and test case

PR #2380, opened 2018-06-29
Hhh 12746

gbadner:
https://hibernate.atlassian.net/browse/HHH-12746

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/internal/CacheKeyImplementation.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/internal/CacheKeyImplementation.java
@@ -55,6 +55,7 @@ final class CacheKeyImplementation implements Serializable {
 
 	private int calculateHashCode(Type type, SessionFactoryImplementor factory) {
 		int result = type.getHashCode( id, factory );
+		result = 31 * result + entityOrRoleName.hashCode();
 		result = 31 * result + (tenantId != null ? tenantId.hashCode() : 0);
 		return result;
 	}
--- a/hibernate-core/src/test/java/org/hibernate/cache/internal/CacheKeyImplementationHashCodeTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/cache/internal/CacheKeyImplementationHashCodeTest.java
@@ -1,0 +1,79 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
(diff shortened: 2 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/cache/internal/CacheKeyImplementation.java
```diff
@@ -55,6 +55,7 @@
 
 	private int calculateHashCode(Type type, SessionFactoryImplementor factory) {
 		int result = type.getHashCode( id, factory );
+		result = 31 * result + entityOrRoleName.hashCode();
```
rvansa: Hmm, I thought that entity name is already handled in `type.getHashCode`. What is the difference between `Type` and entityOrRoleName?

gbadner: No, the entity name is not handled by ```type.getHashCode```.

The ```type``` argument is the ID type. So, if the ID is an ```int```, then its ```Type`` is ```org.hibernate.type.IntegerType```. The hashCode is ultimately the hashCode of the ID, computed [here](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/AbstractTypeDescriptor.java#L76-L79).

hibernate-core/src/test/java/org/hibernate/cache/internal/CacheKeyImplementationHashCodeTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+
+		CacheKeyImplementation anEntityCacheKey = createCacheKeyImplementation(
+				1, sessionFactory.getMetamodel().entityPersister( AnEntity.class ), sessionFactory
+		);
+		CacheKeyImplementation anotherEntityCacheKey = createCacheKeyImplementation(
+				1, sessionFactory.getMetamodel().entityPersister( AnotherEntity.class ), sessionFactory
+		);
+		assertFalse( anEntityCacheKey.equals( anotherEntityCacheKey ) );
+		assertNotEquals( anEntityCacheKey.hashCode(), anotherEntityCacheKey.hashCode() );
```
Sanne: The overall test looks nice, but this specific line is making a wrong assumption. You can't expect any two objects to guarantee having a different hashcode (unless it's hardcoded or the implementation does some really weird masking to compute it).

PR #2382, opened 2018-06-29
Reduce Criteria overhead: add implementors cache in MetamodelImpl

caristu:
A proposal to fix https://hibernate.atlassian.net/browse/HHH-11495 in Hibernate 5.3.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
@@ -111,6 +111,8 @@ public class MetamodelImpl implements MetamodelImplementor, Serializable {
 
 	private final TypeConfiguration typeConfiguration;
 
+	private final transient Map<String, String[]> implementorsCache = new ConcurrentHashMap<>();
+
 	public MetamodelImpl(SessionFactoryImplementor sessionFactory, TypeConfiguration typeConfiguration) {
 		this.sessionFactory = sessionFactory;
 		this.typeConfiguration = typeConfiguration;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
```diff
@@ -623,12 +625,20 @@ else if ( result == INVALID_IMPORT ) {
 	 */
 	public String[] getImplementors(String className) throws MappingException {
 
+		if (implementorsCache.containsKey(className)) {
+			return implementorsCache.get(className);
```
gsmet: Better call `get` then test if the result is not null, that's one less map access.

hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
```diff
@@ -623,12 +625,20 @@ else if ( result == INVALID_IMPORT ) {
...
+
 		final Class clazz;
+		String[] implementors;
+
 		try {
 			clazz = getSessionFactory().getServiceRegistry().getService( ClassLoaderService.class ).classForName( className );
 		}
 		catch (ClassLoadingException e) {
-			return new String[] { className }; //for a dynamic-class
+			implementors = return new String[] { className }; //for a dynamic-class
```
gsmet: This does not compile.

hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
```diff
@@ -641,7 +651,9 @@ else if ( result == INVALID_IMPORT ) {
 			final boolean isMappedClass = className.equals( checkQueryableEntityName );
 			if ( checkQueryable.isExplicitPolymorphism() ) {
 				if ( isMappedClass ) {
-					return new String[] { className }; //NOTE EARLY EXIT
+					implementors =  new String[] {className}; //NOTE EARLY EXIT
```
gsmet: The comment should be moved to the `return`, it does not make sense here.

hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
```diff
@@ -666,7 +678,9 @@ else if ( result == INVALID_IMPORT ) {
 				}
 			}
 		}
-		return results.toArray( new String[results.size()] );
+		implementors = (String[]) results.toArray( new String[results.size()] );
```
gsmet: Is the cast necessary? It shouldn't be.

PR #2383, opened 2018-06-29
HHH-11495 Reduce Criteria overhead: add implementors cache in MetamodelImpl

caristu:
Proposal to fix https://hibernate.atlassian.net/browse/HHH-11495 in Hibernate 5.3

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
@@ -8,6 +8,7 @@ package org.hibernate.metamodel.internal;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -624,7 +625,9 @@ public class MetamodelImpl implements MetamodelImplementor, Serializable {
 	 * @throws MappingException
 	 */
 	public String[] getImplementors(String className) throws MappingException {
-		return implementorsCache.computeIfAbsent( className, this::doGetImplementors );
+		String[] implementors = implementorsCache.computeIfAbsent( className, this::doGetImplementors );
+
+		return Arrays.copyOf( implementors, implementors.length );
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
@@ -614,17 +614,6 @@ public class MetamodelImpl implements MetamodelImplementor, Serializable {
 		}
 	}
 
-	/**
-	 * Given the name of an entity class, determine all the class and interface names by which it can be
-	 * referenced in an HQL query.
-	 *
-	 * @param className The name of the entity class
-	 *
-	 * @return the names of all persistent (mapped) classes that extend or implement the
-	 *     given class or interface, accounting for implicit/explicit polymorphism settings
-	 *     and excluding mapped subclasses/joined-subclasses of other classes in the result.
-	 * @throws MappingException
-	 */
 	@Override
 	public String[] getImplementors(String className) throws MappingException {
 		// computeIfAbsent() can be a contention point and we expect all the values to be in the map at some point so
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
@@ -625,9 +625,22 @@ public class MetamodelImpl implements MetamodelImplementor, Serializable {
 	 * @throws MappingException
 	 */
 	public String[] getImplementors(String className) throws MappingException {
-		String[] implementors = implementorsCache.computeIfAbsent( className, this::doGetImplementors );
+		// computeIfAbsent() can be a contention point and we expect all the values to be in the map at some point so
+		// let's do an optimistic check first
+		String[] implementors = implementorsCache.get( className );
+		if ( implementors != null ) {
+			return Arrays.copyOf( implementors, implementors.length );
+		}
 
-		return Arrays.copyOf( implementors, implementors.length );
+		try {
+			final Class<?> clazz = getSessionFactory().getServiceRegistry().getService( ClassLoaderService.class ).classForName( className );
+			implementors = doGetImplementors( clazz );
+			implementorsCache.putIfAbsent( className, implementors );
+			return Arrays.copyOf( implementors, implementors.length );
+		}
+		catch (ClassLoadingException e) {
+			return new String[]{ className }; // we don't cache anything for dynamic classes
+		}
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


gsmet:
Note that you can push your follow-up commits to your branch and they will be included in the PR, no need to create another one :).

hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
```diff
@@ -625,7 +625,7 @@ else if ( result == INVALID_IMPORT ) {
 	 */
 	public String[] getImplementors(String className) throws MappingException {
 
-		if (implementorsCache.containsKey(className)) {
+		if (implementorsCache.get(className) != null) {
```
gsmet: Here you're still doing 2 map accesses. Sorry, I should have been more clear: you have to store the first one in a variable then reuse the variable for the test and the return.

hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
```diff
@@ -111,6 +111,8 @@
 
 	private final TypeConfiguration typeConfiguration;
 
+	private final transient Map<String, String[]> implementorsCache = new ConcurrentHashMap<>();
```
gsmet: I don't think we need to make it transient, the other similar maps such as `jpaEmbeddableTypeMap` are not. The entity graph one is a bit special.

hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
```diff
@@ -623,12 +625,20 @@ else if ( result == INVALID_IMPORT ) {
 	 */
 	public String[] getImplementors(String className) throws MappingException {
 
+		if (implementorsCache.get(className) != null) {
+			return implementorsCache.get(className);
+		}
+
 		final Class clazz;
+		String[] implementors;
```
gsmet: Minor nitpick, but considering the flow, I would only initialize the variable where needed.

gsmet:
@caristu I added a few more comments but all in all, it looks like a good idea.

Could you put together a test case allowing to reproduce the performance issue so that I can test the benefit of this approach in a benchmark?

You just need to include it in our test case template: https://github.com/hibernate/hibernate-test-case-templates/tree/master/orm/hibernate-orm-5 and push it to JIRA.

Thanks!

caristu:
I'm reusing the "implementors" variable to retrieve the cached values.

I can use another variable, if you still prefer ti declare the variable just when needed for the other cases, let me know.

gsmet:
@caristu no it's OK for the variable. Now that I think about it, maybe we should have a simple `getImplementors()` that does a `implementorsCache.computeIfAbsent(...)` and a `doGetImplementors()` that really does the work. This way, we won't compute the map several times if concurrent accesses happen when the value is not yet computed.

This is small tuning, I can also do it while squashing the patch.

Do you think you'll have the time to put together a test case?

caristu:
@gsmet ok, sounds great

yes, I'm working it, I'll try to provide the test asap

caristu:
@gsmet I've just pushed the test case

This is the repo: https://github.com/caristu/hibernate-test-case-templates/tree/criteria-overhead
And this is the test case that can be used to measure time differences with and without the fix: https://github.com/caristu/hibernate-test-case-templates/blob/criteria-overhead/orm/hibernate-orm-5/src/test/java/org/hibernate/bugs/CriteriaOverhead.java

gsmet:
@caristu I squashed your commits and used `computeIfAbsent()` instead.

It indeed makes a very big difference when the model is large. Wondering how we could have ignored it for so long.

Thanks for the patch. I'll ask @yrodiere to take another look before merging.

caristu:
@gsmet thanks for handling the PR on so short notice. Would this patch be included in next 5.3.2 version?It would be great.

Sanne:
Verry nice!

What about memory usage, do we need to cap the cache size and/or is there any opportunity to discard entries?

gsmet:
@Sanne my understanding is that it's naturally capped by the number of entities you have in your application. And even more the ones you are explicitly querying.

So I don't think we will have an issue with this map. Except if a user executes random queries with random entity names but I don't think it makes any sense :).

Sanne:
Thanks @gsmet , sounds reasonable.

_getImplementors()_ is public JPA API tough. We should make a defensive copy of that array :/

gsmet:
@Sanne is it? I don't see it exposed with an Eclipse "Open call hierarchy".

yrodiere:
@gsmet The method implements `org.hibernate.Metamodel#getImplementors`, which is public API. Not JPA, Hibernate-native API. But still public API.

I pushed a few additional commits to address that, and more.

Regarding the patch, it seems to make sense, but I'm not convinced it's really useful in practice. It could even be damageable, for all I know.

The given benchmark has so much variance that I don't see how we could trust it. Here are the results on my machine:

* without the patch:
  ```
  *** Criteria with 1 threads done in 5317 ms ***
  *** Criteria with 2 threads done in 2 ms ***
  *** Criteria with 3 threads done in 1330 ms ***
  *** Criteria with 4 threads done in 952 ms ***
  *** Criteria with 5 threads done in 764 ms ***
  *** Criteria with 6 threads done in 3 ms ***
  *** Criteria with 7 threads done in 636 ms ***
  *** Criteria with 8 threads done in 3 ms ***
  ```
* with patch:
  ```
  *** Criteria with 1 threads done in 3063 ms ***
  *** Criteria with 2 threads done in 1203 ms ***
  *** Criteria with 3 threads done in 770 ms ***
  *** Criteria with 4 threads done in 677 ms ***
  *** Criteria with 5 threads done in 485 ms ***
  *** Criteria with 6 threads done in 395 ms ***
  *** Criteria with 7 threads done in 3 ms ***
  *** Criteria with 8 threads done in 9 ms ***
  ```
* with patch plus an additional commit adding a `clone()` call in `getImplementors(String)` to protect the cache from external changes:
  ```
  *** Criteria with 1 threads done in 4482 ms ***
  *** Criteria with 2 threads done in 1743 ms ***
  *** Criteria with 3 threads done in 1051 ms ***
  *** Criteria with 4 threads done in 861 ms ***
  *** Criteria with 5 threads done in 718 ms ***
  *** Criteria with 6 threads done in 2 ms ***
  *** Criteria with 7 threads done in 441 ms ***
  *** Criteria with 8 threads done in 410 ms ***
  ```

Or is it me? Should I use some specific configuration to make the benchmark more reliable?

yrodiere:
Also, if we really go for this, I think we should only cache actual hits, at least as a first step. We don't know how people will use this API, they could potentially iterate on all the classes in the classpath, and then we'd end up with a bloated cache.

Maybe we should cache only in cases where the class passed as a parameter actually has persisted subtypes?

gsmet:
@yrodiere I force-pushed what I had in the works and applied 2 of your commits on top of that, the third one was already taken care of.

gsmet:
@Sanne I reworked the method with your advises in mind. Could you take a look? Thanks!

Sanne:
Nice safer approach to not catch on CNFE.

Why did you decide to not cache on `implementors.length == 0` ?

gsmet:
@Sanne Yoann's idea. It's to avoid caching elements for classes not resolved as entities e.g. it won't be possible to fill in the cache with random classes.

Sanne:
ah, good point. Let's merge it ?!

gsmet:
Done!

@caristu thanks for having pushed hard for this.
@yrodiere @Sanne thanks for the reviews and ideas, great team effort!

Sanne:
yea, thanks all! A pleasure to see this shaped collaboratively, and big thanks to @caristu for the core idea.

caristu:
Thank you all for the effort!

PR #2384, opened 2018-07-02
HHH-12749 - Avoid setting the isolation level to the same value in C3P0ConnectionProvider

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12749

### DIFF
--- a/hibernate-testing/src/main/java/org/hibernate/testing/util/ReflectionUtil.java
+++ b/hibernate-testing/src/main/java/org/hibernate/testing/util/ReflectionUtil.java
@@ -1,4 +1,4 @@
-package org.hibernate.test.util;
+package org.hibernate.testing.util;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
@@ -24,6 +24,10 @@ public class ReflectionUtil {
 			return field;
 		}
 		catch ( NoSuchFieldException e ) {
+			Class superClass = clazz.getSuperclass();
+			if ( !clazz.equals( superClass ) ) {
+				return getField( superClass, name );
+			}
 			throw new IllegalArgumentException( "Class " + clazz + " does not contain a " + name + " field", e);
 		}
 	}
--- a/hibernate-c3p0/hibernate-c3p0.gradle
+++ b/hibernate-c3p0/hibernate-c3p0.gradle
@@ -14,6 +14,8 @@ dependencies {
     compile( libraries.c3p0 )
 
     testCompile project( ':hibernate-testing' )
+    testCompile( libraries.mockito )
+    testCompile( libraries.mockito_inline )
 
     testCompile( libraries.validator ) {
         // for test runtime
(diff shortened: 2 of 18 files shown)


hibernate-testing/src/main/java/org/hibernate/testing/util/ReflectionUtil.java
```diff
@@ -24,6 +24,10 @@ public static Field getField(Class clazz, String name) {
 			return field;
 		}
 		catch ( NoSuchFieldException e ) {
+			Class superClass = clazz.getSuperclass();
+			if ( !clazz.equals( superClass ) ) {
+				return getField( superClass, name );
+			}
```
gsmet: This is really weird. We are using explicitly `getDeclaredField()` here so we don't want the fields of the superclasses (and why stop to the direct superclass?).

What is the rationale of this change? Can't we do it more cleanly?

Personally, I would have a separate commit with the package change. Then a separate commit changing `getField()` to `getDeclaredField()` and I would add a new `getField()` method which uses `Class#getField()` instead of `getDeclaredField()`.

That would be more explicit and cleaner.

vladmihalcea: Class#getField only works for public field, so it's not useful.

Splitting into multiple commits is questionable since we wouldn't know the use case that lead to that change. It would look like a random commit. And it complicates cherry picking too. So, I don't agree with this change.

vladmihalcea:
It might be relevant to other CPs.

hibernate-testing/src/main/java/org/hibernate/testing/util/ReflectionUtil.java
gsmet: Changing a package name is not really relevant to this commit so you pollute it with something that has nothing to do with it. Can you explain me the link between a C3P0 issue and moving a test class to another package? You're talking about the use case that leads to the change, but your use case didn't lead you to it, it's just something you wanted to do.

If you want it to be attached to this commit, you can use the JIRA key.
Having a refactoring change mixed with a behavior change is really not something that is recommended.
And sure you will have to cherry-pick 2 commits instead of one but that does not seem like an issue to me.

Missed the public aspect of `getField()`, sorry for the noise. Still think it's not a good idea to change this generic method with this very specific behavior. The behavior is really not natural so better make it something specific to your test class, instead of introducing a weird behavior in a generic util.

vladmihalcea: This is not a refactoring in the main code base. This is moving a test utility so that we can use it in a different module. So while what you say makes sense for when you are actually refactoring code in the src/main folder, for testing is just undeeded ceremony that will require more work and provide no real benefit. All those utilities were added in the Jira issues where we needed them, and no one complained about that since we could easily cherry pick them months after in 5.1 or other beaches. 

Therefore, unless there is a consensus in the team about doing that from now on, I'm not going to split it to 2 commits just for that sake of it.

gsmet: OK, I suppose I have lost enough time on it. Please do whatever you want.

PR #2385, opened 2018-07-02
HHH-12752 - @OneToMany on non primary key cause additional select for…

mikaelst:
https://hibernate.atlassian.net/browse/HHH-12752


beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/3869

PR #2386, opened 2018-07-03
HHH-12762 No longer needing to use port-offset in WildFly integration…

Sanne:
… tests

Hi @vladmihalcea , this is a trivial one. Would you mind verifying it doesn't make the Arquillian tests troublesome for you?

 - https://hibernate.atlassian.net/browse/HHH-12762


Sanne:
I'll merge it.

PR #2387, opened 2018-07-03
[HHH-12765] don't use SQL Server reserved keyword in test

simkam:
https://hibernate.atlassian.net/browse/HHH-12765


Sanne:
merged

PR #2388, opened 2018-07-03
HHH-12531 JCache existing cache not detected.

yrodiere:
https://hibernate.atlassian.net//browse/HHH-12531

Supersedes #2371

Based on #2369, which should be merged first.

This introduces new default names for query regions. Also, this adds a defaulting mechanism that will make our own built-in region factories check if explicit configuration exists for older, "legacy" default query region names when there is no explicit configuration for the most current default query region name.


gsmet:
Merged, thanks!

PR #2389, opened 2018-07-04
HHH-12695 Incompatibility in return value for org.hibernate.procedure.ParameterRegistration.getType() 5.1 vs 5.3

yrodiere:
https://hibernate.atlassian.net//browse/HHH-12695

Note: if we merge this, we'll have to update the migration guide. In particular, we should add the following section [there](https://github.com/hibernate/hibernate-orm/wiki/Migration-Guide---5.3#known-changes) (at the end of "Known changes")

```
=== Query parameters

Method `org.hibernate.type.Type getType()` in `org.hibernate.query.QueryParameter` was renamed to `getHibernateType`.

This was done in order to allow for the re-introduction of `Class<T> getType()` in a sub-interface, `org.hibernate.procedure.ParameterRegistration`. That second method had been removed in 5.2, breaking compatibility with 5.1.
```

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/query/spi/AbstractParameterDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/query/spi/AbstractParameterDescriptor.java
@@ -42,7 +42,7 @@ public abstract class AbstractParameterDescriptor implements QueryParameter {
 	}
 
 	@Override
-	public Type getType() {
+	public Type getHibernateType() {
 		return getExpectedType();
 	}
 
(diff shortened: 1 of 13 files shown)


gsmet:
@gbadner Yoann followed the instructions you put in the JIRA issue. Interesting in your review as you are more versed in this compatibility effort than us.

Thanks!

gbadner:
@yrodiere, @gsmet, this looks good!

gsmet:
Merged, thanks @yrodiere and @gbadner !

PR #2390, opened 2018-07-04
HHH-12555 Fix merging of lazy loaded blobs/clobs/nclobs

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12555

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/AbstractStandardBasicType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/AbstractStandardBasicType.java
@@ -345,17 +345,10 @@ public abstract class AbstractStandardBasicType<T>
 	}
 
 	@Override
+	@SuppressWarnings({ "unchecked" })
 	public final Object replace(Object original, Object target, SharedSessionContractImplementor session, Object owner, Map copyCache) {
 		if ( original == null && target == null ) {
 			return null;
-		}
-
-		if ( original == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
-			return target;
-		}
-
-		if ( target == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
-			return getReplacement( (T) original, null, session );
 		}
 
 		return getReplacement( (T) original, (T) target, session );
--- a/hibernate-core/src/main/java/org/hibernate/type/AbstractStandardBasicType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/AbstractStandardBasicType.java
@@ -71,11 +71,7 @@ public abstract class AbstractStandardBasicType<T>
 	}
 
 	protected T getReplacement(T original, T target, SharedSessionContractImplementor session) {
-		if ( original == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
-			return target;
-		}
-		else if ( !isMutable() ||
-					( target != LazyPropertyInitializer.UNFETCHED_PROPERTY && isEqual( original, target ) ) ) {
+		if ( !isMutable() || ( target != null && isEqual( original, target ) ) ) {
 			return original;
 		}
 		else {
@@ -349,8 +345,19 @@ public abstract class AbstractStandardBasicType<T>
 	}
 
 	@Override
-	@SuppressWarnings({ "unchecked" })
 	public final Object replace(Object original, Object target, SharedSessionContractImplementor session, Object owner, Map copyCache) {
+		if ( original == null && target == null ) {
+			return null;
+		}
+
+		if ( original == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
+			return target;
+		}
+
+		if ( target == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
+			return getReplacement( (T) original, null, session );
+		}
+
 		return getReplacement( (T) original, (T) target, session );
 	}
 
--- a/hibernate-core/src/main/java/org/hibernate/type/BlobType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/BlobType.java
@@ -7,8 +7,12 @@
 package org.hibernate.type;
 
 import java.sql.Blob;
+import java.sql.SQLException;
 
+import org.hibernate.engine.jdbc.LobCreator;
+import org.hibernate.engine.jdbc.spi.JdbcServices;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.internal.util.collections.ArrayHelper;
 import org.hibernate.type.descriptor.java.BlobTypeDescriptor;
 
 /**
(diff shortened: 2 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/type/AbstractStandardBasicType.java
```diff
@@ -349,8 +345,19 @@ public final Type getSemiResolvedType(SessionFactoryImplementor factory) {
...
+			return null;
+		}
+
+		if ( original == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
+			return target;
+		}
+
+		if ( target == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
+			return getReplacement( (T) original, null, session );
+		}
```
yrodiere: You might want to have a look at the other `replace()` method below, which also calls `getReplacement`, but without executing this code. Probably a bug?

It might be a dumb suggestion, but couldn't we just move all this code to the caller(s) of `replace()`? Then any type would benefit from this code, not just types extending `AbstractStandardBasicType`. From what I can see it's rather simple, and it would remove the need for type implementations to refer to `LazyPropertyInitializer`, which is an SPI type...

hibernate-core/src/main/java/org/hibernate/type/BlobType.java
```diff
@@ -36,7 +40,23 @@ protected boolean registerUnderJavaType() {
 
 	@Override
 	protected Blob getReplacement(Blob original, Blob target, SharedSessionContractImplementor session) {
+		if ( target == null ) {
+			return copyOriginalBlob( (Blob) original, session );
```
gbadner: This is OK for `Dialect#NEW_LOCATOR_LOB_MERGE_STRATEGY` (by default, returned by `Dialect#getLobMergeStrategy), but not for `LEGACY_LOB_MERGE_STRATEGY` or `STREAM_XFER_LOB_MERGE_STRATEGY`. It's unknown if this would work for a custom `LobMergeStrategy`.

hibernate-core/src/main/java/org/hibernate/type/ClobType.java
```diff
@@ -36,7 +39,21 @@ protected boolean registerUnderJavaType() {
 
 	@Override
 	protected Clob getReplacement(Clob original, Clob target, SharedSessionContractImplementor session) {
-		return session.getJdbcServices().getJdbcEnvironment().getDialect().getLobMergeStrategy().mergeClob( original, target, session );
+		if ( target == null ) {
+			return copyOriginalClob( (Clob) original, session );
```
gbadner: Same [comment](https://github.com/hibernate/hibernate-orm/pull/2390/files#r241243365) here.

hibernate-core/src/main/java/org/hibernate/type/NClobType.java
```diff
@@ -36,7 +39,21 @@ protected boolean registerUnderJavaType() {
 
 	@Override
 	protected NClob getReplacement(NClob original, NClob target, SharedSessionContractImplementor session) {
+		if ( target == null ) {
+			return copyOriginalNClob( original, session );
```
gbadner: Same [comment](https://github.com/hibernate/hibernate-orm/pull/2390/files#r241243365) here.

hibernate-core/src/main/java/org/hibernate/type/TypeHelper.java
```diff
@@ -193,6 +196,9 @@ else if ( row[i] == LazyPropertyInitializer.UNFETCHED_PROPERTY || row[i] == Prop
 					|| original[i] == PropertyAccessStrategyBackRefImpl.UNKNOWN ) {
 				copied[i] = target[i];
 			}
+			else if ( target[i] == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
+				copied[i] = types[i].replace( original[i], null, session, owner, copyCache, foreignKeyDirection );
```
gbadner: Same [comment](https://github.com/hibernate/hibernate-orm/pull/2390/files#r241254494) here.

hibernate-core/src/main/java/org/hibernate/type/BlobType.java
gbadner: Actually, it should be delegated to the `LobMergeStrategy`.

gbadner:
I've created a new PR using a different approach: https://github.com/hibernate/hibernate-orm/pull/2691

hibernate-core/src/main/java/org/hibernate/type/AbstractStandardBasicType.java
```diff
@@ -71,11 +71,7 @@ public T fromStringValue(String xml) throws HibernateException {
 	}
 
 	protected T getReplacement(T original, T target, SharedSessionContractImplementor session) {
-		if ( original == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
-			return target;
-		}
-		else if ( !isMutable() ||
-					( target != LazyPropertyInitializer.UNFETCHED_PROPERTY && isEqual( original, target ) ) ) {
+		if ( !isMutable() || ( target != null && isEqual( original, target ) ) ) {
```
gbadner: I know you didn't change this, but If the property is mutable, and `isEqual( original, target ) == true`, I think `target` should be returned.

hibernate-core/src/main/java/org/hibernate/type/BlobType.java
gbadner: @gsmet, was there a reason why you didn't just delegate to the `LobMergeStrategy`?

hibernate-core/src/main/java/org/hibernate/type/ClobType.java
gbadner: @gsmet, why not just delegate to the `LobMergeStrategy`?

hibernate-core/src/main/java/org/hibernate/type/NClobType.java
gbadner: @gsmet, why not just delegate to the LobMergeStrategy?

hibernate-core/src/main/java/org/hibernate/type/TypeHelper.java
```diff
@@ -159,6 +159,9 @@ else if ( row[i] == LazyPropertyInitializer.UNFETCHED_PROPERTY || row[i] == Prop
 			if ( original[i] == LazyPropertyInitializer.UNFETCHED_PROPERTY || original[i] == PropertyAccessStrategyBackRefImpl.UNKNOWN ) {
 				copied[i] = target[i];
 			}
+			else if ( target[i] == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
+				copied[i] = types[i].replace( original[i], null, session, owner, copyCache );
```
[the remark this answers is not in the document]
gbadner: It a little ugly to knowingly do the wrong thing here if `LEGACY_LOB_MERGE_STRATEGY` is used. I think it would be reasonable to throw `UnsupportedOperationException` here if `LEGACY_LOB_MERGE_STRATEGY` is used. Unfortunately, it's not possible because `Dialect#LEGACY_LOB_MERGE_STRATEGY` is protected.

hibernate-core/src/test/java/org/hibernate/test/type/LobUnfetchedPropertyTest.java
```diff
@@ -0,0 +1,165 @@
...
+		return new Class<?>[]{ FileBlob.class, FileClob.class, FileNClob.class };
+	}
+
+	@Test
+	public void testBlob() {
+		doInHibernate( this::sessionFactory, s -> {
+			FileBlob file = new FileBlob();
+			file.setBlob( s.getLobHelper().createBlob( "TEST CASE".getBytes() ) );
+			s.save( file );
+			s.clear();
```
gbadner: If you clear the session before flushing, the entity will still be transient.

I was surprised to see that this issue arises with a transient entity.

PR #2391, opened 2018-07-04
HHH-12649 Auto-register entity and collection caches based on the Hibernate @Cache annotation (XML mapping) settings

yrodiere:
https://hibernate.atlassian.net//browse/HHH-12649

Follows up on #2369. Changes the default behavior for the JCache region factory on missing caches, to be consistent with the Ehcache region factory, following our discussion on the mailing list.


gsmet:
Merged, thanks!

PR #2392, opened 2018-07-05
HHH-12200 - Docs mention outdated APIs

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12200


documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
```diff
@@ -610,12 +610,12 @@ The default value of this setting is determined by the value for `hibernate.gene
 [[configurations-cache]]
 === Cache Properties
 
-`*hibernate.cache.region.factory_class*` (e.g. `org.hibernate.cache.infinispan.InfinispanRegionFactory`)::
+`*hibernate.cache.region.factory_class*` (e.g. `org.hibernate.cache.ehcache.internal.EhcacheRegionFactory`)::
```
gsmet: Hi Vlad,

A better example would be `jcache`. The Ehcache region factory is only supporting Ehcache 2 so it's deprecated and we'd better use the short names rather than the internal class names.

documentation/src/main/asciidoc/userguide/chapters/caching/Caching.adoc
```diff
@@ -677,401 +677,5 @@ unless an appropriate `<defaultCache>` entry is added to the Ehcache configurati
...
-In this case you would still have the additional overhead of talking across to another JVM, but it wouldn't have the latency of across a network.
-+
-The benefit of doing this is that:
-+
-** Size the cache separate from the application, since the cache and the application server have very different memory profiles.
-One has lots of short lived objects, and the other could have lots of long lived objects.
-** To pin the cache and the application server onto different CPU cores (using _numactl_ ), and even pin them to different physically memory based on the NUMA nodes.
-
+Infinispan can be used as a 2nd-level cache provider as well. For more details, check out the
+http://infinispan.org/docs/stable/user_guide/user_guide.html#jpa_hibernate_2l_cache[Infinispan User Guide].
```
gsmet: Wondering if maybe we should add a couple of details about what makes Infinispan interesting as a 2nd level cache?

vladmihalcea:
@gsmet There's a commit on master saying that "master is now 5.4", but there's no such version on Jira. Should these commits be merged manually from master to 5.3?

PR #2394, opened 2018-07-06
HHH-12775 - Avoid join on natural id property access

JWGmeligMeyling:
Implements: https://hibernate.atlassian.net/browse/HHH-12775

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
@@ -6,11 +6,6 @@
  */
 package org.hibernate.hql.internal.ast.tree;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Set;
-
 import org.hibernate.QueryException;
 import org.hibernate.engine.internal.JoinSequence;
 import org.hibernate.hql.internal.CollectionProperties;
@@ -21,7 +16,6 @@ import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.internal.log.DeprecationLogger;
 import org.hibernate.internal.util.StringHelper;
-import org.hibernate.internal.util.collections.ArrayHelper;
 import org.hibernate.persister.collection.QueryableCollection;
 import org.hibernate.persister.entity.AbstractEntityPersister;
 import org.hibernate.persister.entity.EntityPersister;
@@ -400,7 +394,7 @@ public class DotNode extends FromReferenceNode implements DisplayableNode, Selec
 			// entity's PK (because 'our' table would know the FK).
 			parentAsDotNode = (DotNode) parent;
 			property = parentAsDotNode.propertyName;
-			joinIsNeeded = generateJoin && !isReferenceToPrimaryKey( parentAsDotNode.propertyName, entityType );
+			joinIsNeeded = generateJoin && !isPropertyEmbeddedInJoinProperties( parentAsDotNode.propertyName );
 		}
 		else if ( !getWalker().isSelectStatement() ) {
 			// in non-select queries, the only time we should need to join is if we are in a subquery from clause
@@ -427,7 +421,7 @@ public class DotNode extends FromReferenceNode implements DisplayableNode, Selec
 
 	}
 
-	private boolean isDotNode(AST n) {
+	private static boolean isDotNode(AST n) {
 		return n != null && n.getType() == SqlTokenTypes.DOT;
 	}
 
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
```diff
@@ -583,39 +577,44 @@ public FromElement getImpliedJoin() {
 	}
 
 	/**
-	 * Is the given property name a reference to the primary key of the associated
-	 * entity construed by the given entity type?
+	 * Is the given property name a reference to the join key of the associated
+	 * entity constructed by the given entity type?
+	 * <p/>
+	 *
+	 * This method resolves the {@code propertyName} as an property of the entity type at the
```
gsmet: s/as an property/as a property/

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
```diff
@@ -583,39 +577,44 @@ public FromElement getImpliedJoin() {
...
+	 * {@link #propertyPath} relative to the {@link #getFromElement() FromElement}.
+	 * The implementation does so by invoking {@link EntityPersister#getPropertyType(String)},
+	 * which will resolve the property path against the entitiy's {@link org.hibernate.persister.entity.PropertyMapping}.
```
gsmet: s/entitiy/entity/

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
```diff
@@ -583,39 +577,44 @@ public FromElement getImpliedJoin() {
...
-				.getEntityPersister( owningType.getAssociatedEntityName() );
-		if ( persister.getEntityMetamodel().hasNonIdentifierPropertyNamedId() ) {
-			// only the identifier property field name can be a reference to the associated entity's PK...
-			return propertyName.equals( persister.getIdentifierPropertyName() ) && owningType.isReferenceToPrimaryKey();
+	private boolean isPropertyEmbeddedInJoinProperties(String propertyName) {
+		EntityPersister entityPersister = this.getFromElement().getEntityPersister();
+		String propertyPath = String.join( ".", this.propertyPath, propertyName );
+		try {
+			Type propertyType = entityPersister.getPropertyType( propertyPath );
+			return propertyType != null;
```
gsmet: Are we really cautious enough? I might have missed something because I don't know this part of the code at all but I find it a bit weird that we only check that the entity has a property with the same name.

You could have a property with the same name that is not at all the foreign key for the other table and has a totally different value. How does the code work in this case?

JWGmeligMeyling: When we're in this method, we're wondering whether `someEntity.someAssociation.prop` can be derived from the SQL columns of `someEntity`. This method is only invoked when `someAssociation` refers to an `EntityType` (i.e. for embedded component types we already know that we don't have to join). The code related to this is in `DotNode#resolve(boolean, boolean, String, AST, AST)`.

The only "extended" paths (paths with a dot in it) that are added to the property mapping are either:
a. Embedded component types (`someEntity.someEmbeddable.someProp`). Obviously a join is not required here, but as mentioned before, this is already accounted for earlier in the process.
b. Referenced id or unique properties of associations (i.e. `someEntity.someAssociation.id` , or in this case `someEntity.someAssociation.isbn`).

Besides these, the property mapping also contains all basic properties. However, because we are in a `DotNode`, the `propertyPath` will always contain at least one dot, so we only have to consider the above two options here.

Furthermore, the exact same `PropertyMapping` is later used when actually dereferencing the attribute. (In `DotNode#getColumns()` > `FromElement#toColumns(String, String, boolean)` > `FromElementType.toColumns()`).

I'm pretty sure we're cautious enough here.

JWGmeligMeyling:
Thanks for the feedback, I'll address the spelling errors!

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
```diff
@@ -583,39 +577,44 @@ public FromElement getImpliedJoin() {
...
-	 * entity construed by the given entity type?
+	 * Is the given property name a reference to the join key of the associated
+	 * entity constructed by the given entity type?
+	 * <p/>
+	 *
+	 * This method resolves the {@code propertyName} as an property of the entity type at the
+	 * {@link #propertyPath} relative to the {@link #getFromElement() FromElement}.
+	 * The implementation does so by invoking {@link EntityPersister#getPropertyType(String)},
+	 * which will resolve the property path against the entitiy's {@link org.hibernate.persister.entity.PropertyMapping}.
+	 * On initialisation of the {@link EntityPersister}, this {@code PropertyMapping} is filled with
```
gsmet: BTW, minor nitpick and no need to change it here but as it seems you want to contribute a lot: we usually use American English so it would be `initialization` and `optimized`. Not sure everything in the code is consistent with it tbh but that's a good rule of thumb.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
```diff
@@ -583,39 +577,44 @@ public FromElement getImpliedJoin() {
...
+			Type propertyType = entityPersister.getPropertyType( propertyPath );
+			return propertyType != null;
 		}
-		// here, we have two possibilities:
-		// 1) the property-name matches the explicitly identifier property name
-		// 2) the property-name matches the implicit 'id' property name
-		// the referenced node text is the special 'id'
-		if ( EntityPersister.ENTITY_ID.equals( propertyName ) ) {
-			return owningType.isReferenceToPrimaryKey();
+		catch (QueryException e) {
```
gsmet: I supposed you haven't found a way to do this without relying on exception logic? It's something we try to avoid.

JWGmeligMeyling: This cannot be done very easily.

`AbstractEntityPersister` implements `Queryable` which extends `PropertyMapping`. The `AbstractEntityPersister` comprises a `BasicEntityPropertyMapping` in the `propertyMapping` field. Some methods from `PropertyMapping` on the `AbstractEntityPersister`delegate directly to this `BasicEntityPropertyMapping`. So by calling `EntityPersister.getPropertyType(String)`, `BasicEntityPropertyMapping.getPropertyType(String)` is invoked under the hood. This `BasicEntityPropertyMapping` looks up the property in a map, and throws a `QueryException` if the property is not found.

However, the other methods related to properties are implemented somewhat inconsistently.  For example: `getPropertyNames()` on `EntityPersister` does not delegate to the `BasicEntityPropertyMapping`, but instead returns the property names from the metamodel, which does *not* contain the extended property paths. In fact: `PropertyMapping` does not have a method that returns all possible property names.

So in order to make this more neat, I have to add a `getPropertyNames` to `AbstractPropertyMapping`, as well as add a `PropertyMapping` accessor to the `AbstractEntityPersister` or add another delegate method to the `AbstractEntityPersister`. Furthermore, I would have to cast the `EntityPersister` to an `AbstractEntityPersister`. All these changes probably require breaking changes, plus I think that this is one of the aspects of the persister overhaul in 6.0 anyway.

beikov:
Is there anything you think should be changed before merging this @gsmet ?

JWGmeligMeyling:
@beikov  I am ready to fix the spelling errors and extend the documentation with the information provided here, I was just waiting to see whether this discussion would show the need for some functional adjustments as well.

gsmet:
@beikov sorry, I'm not really in a position to have the full picture for this issue and I'm fully focused on 5.3 at the moment.

I did review the code and haven't found any obvious logic issue.

If you think it's safe and should go in, let's fix the spelling issues (and squash them in the main commit) and you can commit it to 5.4 only.

beikov:
Sure, thanks. I just wanted to make sure there are no blockers from your side 👍

JWGmeligMeyling:
Reformatted test, fixed spelling errors and simplified logic 😄

PR #2395, opened 2018-07-08
HHH-4808 - SessionImpl.initializeCollection() does not release JDBC connection (if outside of a transaction)

dreab8:
https://hibernate.atlassian.net/browse/HHH-4808

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
@@ -1112,16 +1112,23 @@ public final class SessionImpl
 		LoadEvent event = loadEvent;
 		loadEvent = null;
 		event = recycleEventInstance( event, id, entityName );
-		fireLoad( event, LoadEventListener.IMMEDIATE_LOAD );
-		Object result = event.getResult();
-		if ( loadEvent == null ) {
-			event.setEntityClassName( null );
-			event.setEntityId( null );
-			event.setInstanceToLoad( null );
-			event.setResult( null );
-			loadEvent = event;
+		boolean success = false;
+		try {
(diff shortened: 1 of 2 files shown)


gsmet:
Haven’t reviewed the PR but the JIRA number of the second commit is incorrect :)

dreab8:
fixed
Thanks @gsmet

hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
```diff
@@ -2241,11 +2248,18 @@ public SessionFactoryImplementor getSessionFactory() {
...
+		try {
+			InitializeCollectionEvent event = new InitializeCollectionEvent( collection, this );
+			for ( InitializeCollectionEventListener listener : listeners( EventType.INIT_COLLECTION ) ) {
+				listener.onInitializeCollection( event );
+			}
+			success = true;
+		}
+		finally {
+			delayedAfterCompletion();
+			afterOperation( success );
```
gsmet: When initializing an entity and its collections in one go, we will now call `afterOperation()` several times (after each collection initialization). It won't have any consequence (e.g. it won't close the connection or close the transaction and reopen a new one for the next collection)?

dreab8: I think what could happen is that after the connection is closed any other operation will cause an exception due to the closed connection. 
I'm going to create a test for the scenario you proposed.

sebersole (changes requested):
To me the proper way to do this would be in `AbstractPersistentCollection#withTemporarySessionIfNeeded`, no?

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/3868

PR #2396, opened 2018-07-08
HHH-12647 Use JDBC arrays in batch queries if the DB supports them

agherardi:
This PR is still a work-in-progress.

I know I need to add some tests.

At this point, I'm soliciting feedback to understand if I'm on the right track.


gsmet:
@agherardi sorry for not getting back to you: we are currently quite busy with 6 development and 5.3 stabilization.

Not a review but I think what you're trying to do is worth pursuing and would be a nice improvement.

agherardi:
Thank you for your feedback.

Other than writing some unit tests, is there anything else you suggest that I do at this point?

agherardi:
I put together a couple of tests, however:
- The tests need postgres. Is there a way to 
- I don't know how the test can get a hold of the generated SQL to verify it's what I expect

Any suggestions?

agherardi:
Unit tests added. Please review. Thanks

PR #2397, opened 2018-07-09
HHH-12776 NullPointerException when executing native query. 

dreab8:
https://hibernate.atlassian.net/browse/HHH-12776

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
@@ -953,18 +953,8 @@ public abstract class AbstractSharedSessionContract implements SharedSessionCont
 				}
 			} );
 		}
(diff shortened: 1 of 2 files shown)


vladmihalcea:
@dreab8 I reverted your change in the `AbstractSharedSessionContract` class and the test still passes successfully.

dreab8:
Hi @vladmihalcea 
discussing with @gsmet we discovered my test does not really reproduce the issue.
First the Entity have to be Audited, and then depends on the order 
`getFactory().getMetamodel().getEntities().stream()` returns the values.

vladmihalcea:
@dreab8 @gsmet I see there's a provided test case in the Jira issue. Should we try to use that one instead?

dreab8:
here the test https://github.com/hibernate/hibernate-orm/pull/2398
but as I said it may not fail due to the order `getFactory().getMetamodel().getEntities().stream()` returns the entities

dreab8:
rejected

PR #2398, opened 2018-07-09
HHH-12776 Add test for issue

dreab8:
https://hibernate.atlassian.net/browse/HHH-12776

### DIFF
--- a/hibernate-envers/src/test/java/org/hibernate/envers/test/integration/nativequery/EntityResultNativeQueryTest.java
+++ b/hibernate-envers/src/test/java/org/hibernate/envers/test/integration/nativequery/EntityResultNativeQueryTest.java
@@ -1,0 +1,66 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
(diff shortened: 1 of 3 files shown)


hibernate-envers/src/test/java/org/hibernate/envers/test/integration/nativequery/EntityResultNativeQueryTest.java
```diff
@@ -0,0 +1,66 @@
...
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
```
gsmet: Are you sure you need 3 licenses? :)

gsmet:
@vladmihalcea regarding this issue, we discussed with @dreab8 and we think it's better to revert https://github.com/hibernate/hibernate-orm/pull/2332 entirely for the 5.3 branch.

Currently, we have a very clear contract with the users: we accept either a tuple or an entity and it's not that bad.

We might revisit it in the future if you can come up with a better fix. Apart from the NPE issue, we don't think it's acceptable to iterate on all the entities and call `isAssignableFrom()`. When you have a complex model, it's going to be a performance disaster.

I'll go revert it.

hibernate-envers/src/test/java/org/hibernate/envers/test/integration/nativequery/EntityResultNativeQueryTest.java
dreab8: @gsmet fixed, thanks

gsmet:
Merged, thanks for the test @dreab8 .

vladmihalcea:
@gsmet We can revert the change, but the issue is still a bummer for our users. We'll provide a fix in 5.4. for it.

vladmihalcea:
I'm going to investigate the Envers test now. What's more concerning is that in spite of having over 8000 test cases, a change like this passes successfully, meaning that the current test suite is still far from covering the vast majority of use cases our users have in place.

We will investigate a different approach. We also have a way of getting the entity by its class in the `Metamodel`, so we could use that instead.

PR #2399, opened 2018-07-09
HHH-12778 Register the cause when an exception occurs creating a proxy

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12778

To be backported to the 5.3 branch.


gsmet:
Merged!

PR #2400, opened 2018-07-09
HHH-12793 Upgrade Karaf, pax-exam and reenable the OSGi tests

fax4ever:
https://hibernate.atlassian.net/browse/HHH-12793

### DIFF
--- a/hibernate-core/hibernate-core.gradle
+++ b/hibernate-core/hibernate-core.gradle
@@ -118,10 +118,6 @@ jar {
             // TODO: Shouldn't have to explicitly list this, but the plugin
             // generates it with a [1.0,2) version.
             "javax.persistence;version=\"${project.jpaVersion.osgiName}\"",
-            // Temporarily support JTA 1.1 -- Karaf and other frameworks still
-            // use it.  Without this, the plugin generates [1.2,2).
-            // build.gradle adds javax.transaction for all modules
-            'javax.transaction.xa;version="[1.1,2)"',
             // optionals
             'javax.management;resolution:=optional',
             'javax.naming.event;resolution:=optional',
--- a/hibernate-osgi/src/test/resources/org/hibernate/osgi/test/testing-bundles.xml
+++ b/hibernate-osgi/src/test/resources/org/hibernate/osgi/test/testing-bundles.xml
@@ -7,6 +7,5 @@
 <features name="hibernate-osgi-testing-repository" xmlns='http://karaf.apache.org/xmlns/features/v1.2.0'>
     <feature name='hibernate-osgi-testing' version='5.0.0-SNAPSHOT'>
         <bundle>mvn:com.h2database/h2/1.3.170</bundle>
-        <bundle>mvn:org.ops4j.pax.url/pax-url-aether/2.3.0</bundle>
     </feature>
 </features>
--- a/hibernate-osgi/hibernate-osgi.gradle
+++ b/hibernate-osgi/hibernate-osgi.gradle
@@ -10,32 +10,14 @@ apply plugin: 'com.github.lburgazzoli.karaf'
 
 description = 'Support for running Hibernate O/RM in OSGi environments'
 
-
 ext {
-	// MUST use 4.3.1!  4.3.0 was compiled with "-target jsr14".
-	// http://blog.osgi.org/2012/10/43-companion-code-for-java-7.html
-	osgiCoreVersion = '4.3.1'
-	karafVersion = '3.0.3'
-	paxExamVersion = '4.4.0'
-	paxUrlVersion = '2.4.1'
+	osgiCoreVersion = '6.0.0'
+	osgiCompediumVersion = '5.0.0'
+	karafVersion = '4.1.0'
+	paxExamVersion = '4.12.0'
 }
 
-
-// tests were failing for me, no idea why yet:
-//
-// java.lang.RuntimeException: Cannot get the remote bundle context
-//		at org.ops4j.pax.exam.rbc.client.intern.RemoteBundleContextClientImpl.getRemoteBundleContext(RemoteBundleContextClientImpl.java:255)
-//		at org.ops4j.pax.exam.rbc.client.intern.RemoteBundleContextClientImpl.waitForState(RemoteBundleContextClientImpl.java:211)
-// 		...
-// Caused by: java.rmi.ConnectException: Connection refused to host: 198.105.254.130; nested exception is:
-//		java.net.ConnectException: Connection timed out (Connection timed out)
-//		...
-// Caused by: java.net.ConnectException: Connection timed out (Connection timed out)
-//		...
-//
-//
-// todo : fix these tests and re-enable
-test.enabled = false
+test.enabled = true
 
 if ( JavaVersion.current().isJava9Compatible() ) {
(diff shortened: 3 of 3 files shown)


hibernate-osgi/src/test/java/org/hibernate/osgi/test/OsgiIntegrationTest.java
```diff
@@ -144,34 +139,6 @@ private String testingFeatureXmlUrl() {
 	@ProbeBuilder
 	public TestProbeBuilder probeConfiguration(TestProbeBuilder probe) {
 		System.out.println( "Configuring probe..." );
-
-		// Note : I found locally that this part is not needed.  But I am leaving this here as I might
-		// 		someday have a need for tweaking the probe and I want to remember how it is done...
```
gsmet: Let's not remove this. The comment clearly says that the original author wanted us to keep it.

hibernate-osgi/src/test/resources/org/hibernate/osgi/test/testing-bundles.xml
```diff
@@ -8,5 +8,13 @@
...
         <bundle>mvn:com.h2database/h2/1.3.170</bundle>
         <bundle>mvn:org.ops4j.pax.url/pax-url-aether/2.3.0</bundle>
+
+        <!-- org.jboss.spec.javax.transaction/jboss-transaction-api_1.2_spec does not include javax.transaction.xa package anymore -->
+        <!-- adding here bundles to provide the package -->
+        <bundle>mvn:javax.transaction/javax.transaction-api/1.2</bundle>
+        <bundle>mvn:javax.enterprise/cdi-api/1.2</bundle>
+        <bundle>mvn:javax.el/javax.el-api/3.0.0</bundle>
+
+        <bundle>mvn:com.sun.activation/javax.activation/1.2.0</bundle>
```
gsmet: Some questions here:
- `javax.transaction.xa` is present in Java SE now. I suppose that somehow Karaf doesn't expose it?
- wondering if we should depend on the `transaction-api` feature described here instead of declaring the dependencies manually: https://github.com/apache/karaf/blob/master/assemblies/features/enterprise/src/main/feature/feature.xml#L32 ?
- if it's not supposed to be optional, shoudn't this dependency be part of the core ORM feature instead?

Open questions, interested in your feedback :).

hibernate-core/hibernate-core.gradle
```diff
@@ -118,10 +118,6 @@ jar {
             // TODO: Shouldn't have to explicitly list this, but the plugin
             // generates it with a [1.0,2) version.
             "javax.persistence;version=\"${project.jpaVersion.osgiName}\"",
-            // Temporarily support JTA 1.1 -- Karaf and other frameworks still
-            // use it.  Without this, the plugin generates [1.2,2).
-            // build.gradle adds javax.transaction for all modules
-            'javax.transaction.xa;version="[1.1,2)"',
```
gsmet: So if we remove that, the dependency is added automatically?

fax4ever: Yes it is added with version interval: [1.2,2)

hibernate-osgi/src/test/resources/org/hibernate/osgi/test/testing-bundles.xml
fax4ever: Do we still support Java 7 or not?

gsmet: ORM 5.3 is targeting Java 8+

fax4ever: So, may I remove some extra configurations for Java 7?

fax4ever: It is a very good idea using `transaction-api` feature already provided by the Karaf container itself, but I think that we need to update the version, in fact we are using a very old Karaf version here: `3.0.3`.

gsmet: Outch, good catch. And a very old OSGi core and a very old pax-exam too.

We should certainly try to upgrade them if it's in the way of doing things properly.

fax4ever: It is also needed to run JGroups 4 bundle. See https://groups.google.com/forum/#!topic/ops4j/mUZPQ1c-XTE.

hibernate-osgi/src/test/java/org/hibernate/osgi/test/OsgiIntegrationTest.java
```diff
@@ -99,13 +100,7 @@
...
 						.karafVersion( paxExamEnvironment.getProperty( "org.ops4j.pax.exam.container.karaf.version" ) )
 						.name( "Apache Karaf" )
-						.unpackDirectory(
-								new File(
-										paxExamEnvironment.getProperty(
-												"org.ops4j.pax.exam.container.karaf.unpackDir"
-										)
-								)
-						)
+						.unpackDirectory( new File( paxExamEnvironment.getProperty( "org.ops4j.pax.exam.container.karaf.unpackDir" ) ) )
```
gsmet: I think the whole change to this file can be reverted.

fax4ever: Yes, surprisingly, we don't need anymore to touch it.

hibernate-osgi/src/test/resources/org/hibernate/osgi/test/testing-bundles.xml
```diff
@@ -7,6 +7,6 @@
 <features name="hibernate-osgi-testing-repository" xmlns='http://karaf.apache.org/xmlns/features/v1.2.0'>
     <feature name='hibernate-osgi-testing' version='5.0.0-SNAPSHOT'>
         <bundle>mvn:com.h2database/h2/1.3.170</bundle>
-        <bundle>mvn:org.ops4j.pax.url/pax-url-aether/2.3.0</bundle>
+        <bundle>mvn:com.sun.activation/javax.activation/1.2.0</bundle>
```
gsmet: 2 questions here:
- do we really need that? What's its purpose? My main concern is that if we need to add some necessary dependency for testing, maybe our features are incorrect and are missing a jar? Or is it necessary for the H2 driver?
- if we need that, wondering if `javax.activation:javax.activation-api` wouldn't be enough. Not very fond of having `com.sun` artifacts.

fax4ever: You're definitely right. I'm going to move activation bundle to the hibernate-orm feature. I tried to change the dependency from `com.sun.activation:javax.activation:1.2.0` to `javax.activation:javax.activation-api`, but it did not work.

fax4ever:
@gsmet I removed the bundle `javax.activation:javax.activation-api` from hibernate-orm feature, the bundle was not necessary anymore. Then, we don't need `com.sun.activation:javax.activation:1.2.0` too.

gsmet:
Merged!

@fax4ever thanks a lot for taking care of it, nice work!

fax4ever:
Thank you ;)

PR #2401, opened 2018-07-09
HHH-12780 - Re-allow native SQL queries that take a given resultClass to map the result set to the required type

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12780

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
@@ -10,6 +10,7 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
+import java.lang.reflect.Array;
 import java.sql.SQLException;
 import java.util.List;
 import java.util.TimeZone;
@@ -80,6 +81,7 @@ import org.hibernate.resource.jdbc.spi.StatementInspector;
 import org.hibernate.resource.transaction.backend.jta.internal.JtaTransactionCoordinatorImpl;
 import org.hibernate.resource.transaction.spi.TransactionCoordinator;
 import org.hibernate.resource.transaction.spi.TransactionCoordinatorBuilder;
+import org.hibernate.transform.BasicTransformerAdapter;
 import org.hibernate.type.Type;
 import org.hibernate.type.descriptor.sql.SqlTypeDescriptor;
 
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
```diff
@@ -932,12 +934,36 @@ public NativeQueryImplementor createNativeQuery(String sqlString, Class resultCl
...
 			query.addEntity( "alias1", resultClass.getName(), LockMode.READ );
 		}
+		else if ( !isObjectArray ) {
+			query.setResultTransformer( new BasicTransformerAdapter() {
+				@Override
+				public Object transformTuple(Object[] tuple, String[] aliases) {
+					return resultClass.cast( tuple[0] );
+				}
+			} );
+		}
```
gsmet: What in the `else` case? Is it legit to not do anything?

vladmihalcea: In that case, the result is processed as `Object[]` and we don't have to do any transformation.

gsmet:
@vladmihalcea so about this one, I also reverted the tests you added so you would have to add them back.

hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
```diff
@@ -932,12 +934,36 @@ public NativeQueryImplementor createNativeQuery(String sqlString, Class resultCl
...
+				public Object transformTuple(Object[] tuple, String[] aliases) {
+					Object[] result = (Object[]) Array.newInstance( elementClass, tuple.length );
+					for ( int i = 0; i < tuple.length; i++ ) {
+						result[i] = elementClass.cast( tuple[i] );
+					}
+					return result;
+				}
+			} );
+		}
+		else if ( this.getFactory().getMetamodel().entityPersisters().get( resultClass.getName() ) != null ) {
```
gsmet: Maybe I'm wrong but I think the key of the map returned by the `entityPersisters()` is not always the class name but can also be the entity name if defined (at least if the mapping is done via XML, it seems that there is a possibility the class name is not used).

See https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/AbstractEntitySourceImpl.java#L114

That being said, it looks like `MetamodelImpl#entityPersister(Class entityClass)` is doing the same mistake so I'm not sure about it.

gsmet:
@vladmihalcea still requires the tests to be reintroduced: https://github.com/hibernate/hibernate-orm/commit/9fac6747ef22acff9c7350efca33286452a2c967#diff-8366ba4b8b14c099a596fe79b1ab7849

vladmihalcea:
This will be implemented in 6.0.

PR #2406, opened 2018-07-10
HHH-8805 - [SchemaUpdate] javax.persistence.ForeignKey doesn't respect ConstraintMode.NO_CONSTRAINT

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-8805

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaUpdateJoinColumnNoConstraintSecondaryTableTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaUpdateJoinColumnNoConstraintSecondaryTableTest.java
@@ -1,0 +1,106 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.test.schemaupdate;
+
+import java.io.File;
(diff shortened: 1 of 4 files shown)


vladmihalcea:
@gsmet I plan on integrating this on master and think it's worth having in 5.3 as well. Let me know if you don't want it integrated there as well.

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaUpdateJoinColumnNoConstraintSecondaryTableTest.java
```diff
@@ -0,0 +1,106 @@
...
+			"    create table Child ( " +
+			"       id bigint not null, " +
+			"        some_fk bigint, " +
+			"        primary key (id) " +
+			"    ); " +
+			" " +
+			"    create table Parent ( " +
+			"       id bigint not null, " +
+			"        primary key (id) " +
+			"    ); ";
```
gsmet: Quick question as I'm not very familiar with the tests in this area: the generated scripts will be stable across the dialects? E.g. if the tests are run with MySQL/DB2/... the script will be the same?

gsmet:
> think it's worth having in 5.3 as well

Makes totally sense. I will affect 5.3.3 as the fixVersion.

I added a small question about the tests being portable. The fix looks good.

vladmihalcea:
@gsmet The tests are set to run on H2 only:

    @RequiresDialect(H2Dialect.class)

so, there's no possibility of a test failure on other DBs.

gsmet:
@vladmihalcea OK perfect, then.

PR #2412, opened 2018-07-11
HHH-12730 : User types built using 5.1 are not binary compatible with…

gbadner:
… 5.3

https://hibernate.atlassian.net/browse/HHH-12730

No need to review this if we decide to use compatibility transformer instead.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/ProcedureParameterNamedBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/ProcedureParameterNamedBinder.java
@@ -10,7 +10,6 @@ import java.sql.CallableStatement;
 import java.sql.SQLException;
 
 import org.hibernate.HibernateException;
-import org.hibernate.engine.spi.SessionImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 
 /**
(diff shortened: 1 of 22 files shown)


hibernate-core/src/main/java/org/hibernate/type/ProcedureParameterNamedBinder.java
```diff
@@ -31,7 +32,7 @@
 	 * Bind a value to the JDBC prepared statement, ignoring some columns as dictated by the 'settable' parameter.
 	 * Implementors should handle the possibility of null values.
 	 * Does not support multi-column type
-	 *
+	 *SubTransactionType
```
gsmet: Looks like a copypasto.

gbadner:
This approach won't work, so closing.

gbadner:
Re-opening since it appears to be needed after all.

hibernate-core/src/main/java/org/hibernate/type/ProcedureParameterNamedBinder.java
gsmet: Fixed.

gsmet:
@gbadner I amended your commit with a small fix and added an additional one for the other `AbstractPersistentCollection`. If the constructor is needed for `PersistentBag`, I suspected it would be needed by the others.

I will take a look at the transformer and decide if we should remove some methods later today. I suspect the ones in user types called by ORM should be removed.

gsmet:
Pushed a few other fixes and merged.

PR #2414, opened 2018-07-11
HHH-12771 - Caused by: java.lang.UnsupportedOperationException: Cache provider [org.hibernate.cache.ehcache.internal.EhcacheRegionFactory@3271ec2a] does not support `transactional` access

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12771

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/spi/SecondLevelCacheLogger.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/spi/SecondLevelCacheLogger.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.cache.spi;
 
+import org.hibernate.cache.spi.access.AccessType;
 import org.hibernate.metamodel.model.domain.NavigableRole;
 
 import org.jboss.logging.BasicLogger;
@@ -89,5 +90,13 @@ public interface SecondLevelCacheLogger extends BasicLogger {
 			id = NAMESPACE + 7
 	)
 	void usingLegacyCacheName(String currentName, String legacyName);
+
+	@LogMessage(level = WARN)
+	@Message(
+			value = "Cache [%1$s] uses the [%2$s] access type, but [%3$s] does not support it natively." +
+					" Make sure your cache implementation supports JTA transactions.",
+			id = NAMESPACE + 8
+	)
+	void nonStandardSupportForAccessType(String key, String accessType, String regionName);
 
 }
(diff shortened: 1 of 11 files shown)


gsmet:
@sebersole I would appreciate your feedback on this one. Thanks!

sebersole:
@vladmihalcea Not sure what you mean.  `hibernate-ehcache` only works with Ehcache 2

sebersole:
From the discussion on the Jira, it seems Ehcache 2 does in fact support JTA transactions - so it seems fine to add that support back in.  

However, I am not a fan of blindly adding support for this with JCache.  It's dangerous to just allow "transactional access" when the underlying cache does not support JTA transactionality.  You end up with non-ACID access.  And the user can literally drop in any JCache implementor they want, even a custom one - not all of them support JTA transactional access.  At *the very least* I think there should be a INFO (or even WARN) message stating the danger.

vladmihalcea:
@sebersole I will change the PR to drop Transactional support for JCache. We can reconsider it in future if a newer version of JCache will support it explicitly.

sebersole:
The other option like I said is to do it, but to log a warning/info
message.  Ultimately if we could discern the difference (or user
configures) we could leverage the CacheTransactionContext approach I
mentioned earlier.

vladmihalcea:
@sebersole I added the warning message when using JCache. 

Now, when running this `TRANSACTIONAL` unit test I added to the `hibernate-jcache`, we can see the following warning messages:

```
07:25:00,426  WARN cache:40 - HHH90001008: Cache [org.hibernate.test.cache.jcache.JCacheTransactionalCacheConcurrencyStrategyTest$Parent] uses the [transactional] access type, but [JCacheRegionFactory] does not support it natively. Use it at your own risk.
07:25:00,435  WARN cache:40 - HHH90001008: Cache [org.hibernate.test.cache.jcache.JCacheTransactionalCacheConcurrencyStrategyTest$Child] uses the [transactional] access type, but [JCacheRegionFactory] does not support it natively. Use it at your own risk.
07:25:00,439  WARN cache:60 - HHH90001008: Cache [org.hibernate.test.cache.jcache.JCacheTransactionalCacheConcurrencyStrategyTest$Parent.children] uses the [transactional] access type, but [JCacheRegionFactory] does not support it natively. Use it at your own risk.
```

Let me know if you want me to tweak the message. Otherwise, I'll integrate it to 5.4, and to 5.3 if @gsmet  agrees as well.

gsmet:
Not sure I understand the issue correctly but if I do, I would have the message say:
```
Cache [org.hibernate.test.cache.jcache.JCacheTransactionalCacheConcurrencyStrategyTest$Parent] uses the [transactional] access type with JCache. Make sure your cache implementation supports JTA transactions.
```

vladmihalcea:
@sebersole Do you agree with this WARN message too?

PR #2417, opened 2018-07-11
HHH-12177 Relocation of Infinispan provider for 5.2

galderz:


rvansa:
@galderz looks like LGTM CI failed, but I can't see any error in the log...

gsmet:
I might have missed some discussion but are we sure we want to do that in the maintenance branches?

galderz:
There hasn't been a discussion on this. I was just proposing that Hibernate 5.2.x users move to a newer Infinispan second level cache provider. However, thinking it further, latest Infinispan 9.3.x drops support for Hibernate 5.2.x, leaving only 5.1.x and 5.3.x in there, so for 5.2.x maybe better just leave it where it is.

Sanne:
> However, thinking it further, latest Infinispan 9.3.x drops support for Hibernate 5.2.x, leaving only 5.1.x and 5.3.x in there

That's right we only need support for ORM 5.1 and 5.3 going forward. Users on ORM 5.2 can (should) use the existing Infinispan releases.

Sanne:
To clarify my previous comment - I didn't mean that we should not try to help people on 5.2. Just that we don't *need* to.

Either way would be nice to have a relocation in place so to drop this from maintenance.

@galderz up to you? Do you need to change the version it's relocating to?

rvansa:
It seems to me that we should relocate. We're not applying bugfixes to 5.2 anymore, but some may have been applied in Infinispan 9.2 branch.

galderz:
@Sanne 

> @galderz up to you? Do you need to change the version it's relocating to?

The PR already contains the necessary changes. The latest 5.2 provider is Infinispan 9.2.5.Final.

galderz:
Btw, the CI error is a problem with the changes here. Looking...

galderz:
Gradle layout has been refactored in 5.3. I've added missing gradle files for this, but might not be the right thing to do. If it's too much of a pain to sort this out I'm happy to leave things as they are.

galderz:
Let's see how this goes before tackling 5.1

rvansa:
@galderz Still failing.

galderz:
Hmmm, closing this, too much has changed to be able to apply the right Gradle settings. I've made sure the examples use the Infinispan source tree provider.

PR #2418, opened 2018-07-11
HHH-12177 Update relocation of Infinispan provider for 5.3

galderz:


gsmet:
@rvansa FYI I also applied this commit to master.

PR #2420, opened 2018-07-11
HHH-12797 Fix cache modes relationships table layout

diegolovison:
Fix cache modes relationships table layout

https://hibernate.atlassian.net/browse/HHH-12797


gsmet:
@diegolovison good catch!

It's merged, thanks!

PR #2421, opened 2018-07-12
HHH-12787 - SessionJdbcBatchTest hangs with DB2

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12787


vladmihalcea:
Basically, this is what we need to do to intercept JDBC operations:

- If SQL statement logging is what we need, then `SQLStatementInterceptor` is the way to go since it does not require JDBC mocking or stubbing
- If we need JDBC mocking or stubbing, then `PreparedStatementSpyConnectionProvider` is unavoidable. But then, we need the test will `@RequiresDialectFeature(DialectChecks.SupportsJdbcDriverProxying.class)` so we can skip such tests for those JDBC Drivers that cannot be proxied.

gsmet:
@vladmihalcea looks good to me. Thanks for taking the time to convert all the tests that could be converted.

We will need this in 5.3 too.

PR #2422, opened 2018-07-12
HHH-12795 FlushMode Manual as annotation on a namedquery is not used

yrodiere:
https://hibernate.atlassian.net//browse/HHH-12795

I ran the tests I added against 5.1, and they passed, so this definitely restores behavioral compatibility with 5.1.
It seems `AddNamedQueryTest` (which expects a different behavior from my own tests) was added during the development of 5.2 and has wrong expectations regarding the flush mode in newly instantiated named queries.

The last commit is not absolutely necessary, but may be a nice bonus if you think it's not too dangerous.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
@@ -3776,12 +3776,7 @@ public final class SessionImpl
 		}
 
 		if ( namedQueryDefinition.getFlushMode() != null ) {
-			if ( namedQueryDefinition.getFlushMode() == FlushMode.COMMIT ) {
-				query.setFlushMode( FlushModeType.COMMIT );
-			}
-			else {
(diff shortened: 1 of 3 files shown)


gsmet:
@yrodiere looks good to me.

As for the last commit, I think it's better if we only apply it on master and not for 5.3.

gbadner:
I think the fix in https://github.com/hibernate/hibernate-orm/pull/2422/commits/defe8c6e2c467310de1816da6d200dd1c7c21b63 is fine.

I'm not sure about removing the duplicate code in https://github.com/hibernate/hibernate-orm/pull/2422/commits/d1f85bad23820b561f8010beb2db1b5946baa8d5. I don't have time to check all the possibilities. Please don't backport this to 5.3. I'm not sure what to say about master. In general, we're more aggressive about pushing things like this when we're working on a new minor release, assuming that there is enough time to fix it later if it breaks things.

gsmet:
Pushed the first 2 commits to 5.3 and all the commits to master.

PR #2424, opened 2018-07-12
HHH-12799 Enforce version alignment of Mockito and ByteBuddy dependen…

Sanne:
…cies

- https://hibernate.atlassian.net/browse/HHH-12799

### DIFF
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -59,6 +59,7 @@ ext {
 
             // Byte Buddy
             byteBuddy:      "net.bytebuddy:byte-buddy:${byteBuddyVersion}",
+            byteBuddy_agent: "net.bytebuddy:byte-buddy-agent:${byteBuddyVersion}",
 
             // javax
             jpa:            "javax.persistence:javax.persistence-api:${project.jpaVersion}",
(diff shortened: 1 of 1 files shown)


gradle/libraries.gradle
```diff
@@ -59,6 +59,7 @@ ext {
 
             // Byte Buddy
             byteBuddy:      "net.bytebuddy:byte-buddy:${byteBuddyVersion}",
+            byteBuddy_agent: "net.bytebuddy:byte-buddy-agent:${byteBuddyVersion}",
```
gsmet: AFAICS, you don't use that one.

Sanne: I don't but it's good to have it defined as a memo - it's pulled in transitively from Mockito.

gsmet: Well, my issue with it is that we will suppose it's what sets the version but it's not at all.

Feel free to push if you think I worry for nothing but I don't like having things misleading the developer. A comment saying that the version of byte-buddy-agent is enforced below would have my preference.

Again, feel free to ignore me on this one :).

Sanne: np, I'll remove it since you think that's clearer.

Sanne:
merged

PR #2425, opened 2018-07-12
HHH-12791 Cache the Component type to avoid generating a lot of Bytebuddy proxies

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12791

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/mapping/Component.java
+++ b/hibernate-core/src/main/java/org/hibernate/mapping/Component.java
@@ -52,6 +52,9 @@ public class Component extends SimpleValue implements MetaAttributable {
 
 	private java.util.Map<EntityMode,String> tuplizerImpls;
 
+	// cache the status of the type
+	private volatile Type type;
+
 	/**
 	 * @deprecated User {@link Component#Component(MetadataBuildingContext, PersistentClass)} instead.
 	 */
@@ -209,13 +212,28 @@ public class Component extends SimpleValue implements MetaAttributable {
 
 	@Override
 	public Type getType() throws MappingException {
-		// TODO : temporary initial step towards HHH-1907
-		final ComponentMetamodel metamodel = new ComponentMetamodel(
-				this,
-				getMetadata().getMetadataBuildingOptions()
-		);
-		final TypeFactory factory = getMetadata().getTypeConfiguration().getTypeResolver().getTypeFactory();
-		return isEmbedded() ? factory.embeddedComponent( metamodel ) : factory.component( metamodel );
+		// Resolve the type of the value once and for all as this operation generates a proxy class
+		// for each invocation.
+		// Unfortunately, there's no better way of doing that as none of the classes are immutable and
+		// we can't know for sure the current state of the property or the value.
+		Type localType = type;
+
+		if ( localType == null ) {
+			synchronized ( this ) {
+				if ( type == null ) {
+					// TODO : temporary initial step towards HHH-1907
+					final ComponentMetamodel metamodel = new ComponentMetamodel(
+							this,
+							getMetadata().getMetadataBuildingOptions()
+					);
+					final TypeFactory factory = getMetadata().getTypeConfiguration().getTypeResolver().getTypeFactory();
+					localType = isEmbedded() ? factory.embeddedComponent( metamodel ) : factory.component( metamodel );
+					type = localType;
+				}
+			}
+		}
+
+		return localType;
 	}
 
 	@Override
@@ -288,15 +306,15 @@ public class Component extends SimpleValue implements MetaAttributable {
 		}
 		return result;
 	}
-	
+
 	public boolean isKey() {
 		return isKey;
 	}
-	
+
 	public void setKey(boolean isKey) {
 		this.isKey = isKey;
 	}
-	
+
 	public boolean hasPojoRepresentation() {
 		return componentClassName!=null;
 	}
--- a/hibernate-core/src/test/java/org/hibernate/test/component/proxy/ComponentBasicProxyTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/component/proxy/ComponentBasicProxyTest.java
@@ -7,11 +7,20 @@
 package org.hibernate.test.component.proxy;
 
 import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
+import static org.junit.Assert.assertEquals;
 
 import java.util.List;
 
+import org.hibernate.EntityMode;
+import org.hibernate.boot.Metadata;
+import org.hibernate.boot.MetadataSources;
+import org.hibernate.boot.registry.StandardServiceRegistry;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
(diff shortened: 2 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/test/component/proxy/ComponentBasicProxyTest.java
```diff
@@ -44,4 +53,28 @@ public void testBasicProxyingWithProtectedMethodCalledInConstructor() {
...
+					.build();
+			PersistentClass persistentClass = metadata.getEntityBinding( Person.class.getName() );
+
+			ComponentType componentType1 = (ComponentType) persistentClass.getIdentifierMapper().getType();
+			Object instance1 = componentType1.instantiate( EntityMode.POJO );
+
+			ComponentType componentType2 = (ComponentType) persistentClass.getIdentifierMapper().getType();
+			Object instance2 = componentType2.instantiate( EntityMode.POJO );
+
+			assertEquals( instance1.getClass(), instance2.getClass() );
```
yrodiere: Maybe also assert both classes implement `HibernateProxy`, otherwise we're not testing what we want to test.

hibernate-core/src/main/java/org/hibernate/mapping/Component.java
```diff
@@ -209,13 +212,28 @@ public void setDynamic(boolean dynamic) {
...
+					final ComponentMetamodel metamodel = new ComponentMetamodel(
+							this,
+							getMetadata().getMetadataBuildingOptions()
+					);
+					final TypeFactory factory = getMetadata().getTypeConfiguration().getTypeResolver().getTypeFactory();
+					localType = isEmbedded() ? factory.embeddedComponent( metamodel ) : factory.component( metamodel );
+					type = localType;
+				}
+			}
+		}
```
yrodiere: This is a bit unfortunate, but I must admit I can't think of any other solution, considering:

* there's no obvious place in the bootstrap process from where we could call some `init` method on this class
* it's not obvious what data is required to initialize the type either, so we can't just replace various setters with a big `setTypeInfo(Something, Somethingelse, ...)` setter.

I guess there are ways to solve this by changing how `org.hibernate.tuple.Instantiator` is... instantiated, but that would require changing `AbstractComponentTuplizer` and all its subclasses, and maybe also `ComponentTuplizerFactory` and `ComponentMetamodel`, to rely on a global cache at some level. But that sounds a bit dangerous, and I'm not sure it's doable without changing publicly exposed methods...

So... +1.

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/PassThroughInterceptor.java
```diff
@@ -11,57 +11,58 @@
...
 		final String name = method.getName();
-		if ( "toString".equals( name ) ) {
+
+		if ( "toString".equals( name ) && arguments.length == 0 ) {
 			return proxiedClassName + "@" + System.identityHashCode( instance );
 		}
-		else if ( "equals".equals( name ) ) {
-			return proxiedObject == instance;
+
+		if ( "equals".equals( name ) && arguments.length == 1 ) {
```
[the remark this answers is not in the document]
gsmet: Yeah, I stopped there, no special reason why but I think it should be enough.

hibernate-core/src/main/java/org/hibernate/mapping/Component.java
gsmet: Yeah, my thoughts exactly.

Another line of thought is to consider this method will be called at least once during the bootstrap in a mono-threaded environment before being used in a multi-threaded one. Thus we wouldn't need the volatile/synchronized. I'm not sure I want to go that way and expose us to weird multi-threading issues if one day, this method is not called anymore during bootstrap for whatever reason.

hibernate-core/src/test/java/org/hibernate/test/component/proxy/Adult.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Entity
+public class Adult extends Person {
+
+	public Adult() {
+		someInitMethod();
+	}
+
+	@Override
+	public void someInitMethod() {
```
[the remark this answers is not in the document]
gsmet: The proxy is generated for `Person` which is abstract due to the `PersonId` component.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/EnhancerImpl.java
```diff
@@ -170,42 +170,42 @@ private TypePool buildClassPool(final ByteBuddyEnhancementContext enhancementCon
...
-							.intercept( Advice.to( CodeTemplates.AreFieldsDirtyWithoutCollections.class ).wrap( StubMethod.INSTANCE ) )
+									.intercept( Advice.to( CodeTemplates.AreFieldsDirtyWithoutCollections.class ).wrap( StubMethod.INSTANCE ) )
 							.defineMethod( EnhancerConstants.TRACKER_CLEAR_NAME, void.class, Visibility.PUBLIC )
-							.intercept( Advice.to( CodeTemplates.ClearDirtyAttributesWithoutCollections.class ).wrap( StubMethod.INSTANCE ) )
+									.intercept( Advice.to( CodeTemplates.ClearDirtyAttributesWithoutCollections.class ).wrap( StubMethod.INSTANCE ) )
 							.defineMethod( EnhancerConstants.TRACKER_SUSPEND_NAME, void.class, Visibility.PUBLIC )
-							.withParameters( boolean.class )
-							.intercept( Advice.to( CodeTemplates.SuspendDirtyTracking.class ).wrap( StubMethod.INSTANCE ) )
+									.withParameters( boolean.class )
+										.intercept( Advice.to( CodeTemplates.SuspendDirtyTracking.class ).wrap( StubMethod.INSTANCE ) )
```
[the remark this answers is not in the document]
gsmet: Fixed.

hibernate-core/src/test/java/org/hibernate/test/component/proxy/ComponentBasicProxyTest.java
gsmet: This particular type of proxy does not implement `HibernateProxy`. The ByteBuddy generated classes implement `ProxyInterceptor` but not the Javassist ones.

gsmet:
@gbadner in this commit https://github.com/hibernate/hibernate-orm/pull/2425/commits/9cb003984888c1ff6b0f1c034c03bbdf05ccfd74 , I cache the `type` of a  `Component`. As the object is totally mutable, it's hard to know if it's safe.

Apparently, Andrea tried to cache the type on other `Value` types and it led to some issues. In any case, we would need to cache the `ComponentTuplizer` to avoid creating new proxies.

gbadner:
@gsmet, I'm really not sure about this. Please check with @sebersole.

gsmet:
Merged it. I took a close look at how the `Component` are built and apparently they are initialized all in one go so it should be safe enough.

Hopefully, we will get feedback soon if there are any issues.

PR #2429, opened 2018-07-14
Hhh 12791

dreab8:
@gsmet this PR is based on top of your https://github.com/hibernate/hibernate-orm/pull/2425 and try to just address the https://hibernate.atlassian.net/browse/HHH-12791 in a slightly different way,


gsmet:
@dreab8 I don't understand the rationale of the change? Could you explain it?

My patch is less intrusive and I think it's better to cache the full type rather than generating it each time.

dreab8:
Hi @gsmet I'm just not !00% sure that caching the Type created is completely safe, and may be also caching a tuplizer is not.
I have just noticed that for org.hibernate.mapping.ManyToOne, cashing the Type causes the failure of OneToOneWithDerivedIdentityTest but may be for a Component it is not an issue

PR #2432, opened 2018-07-16
HHH-12666 Don't wrap PersistenceException inside PersistenceException

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12666

So it might break user code for 5.2 and early 5.3 but I think it's a bug and should be treated as such: we shouldn't wrap our nicely typed exceptions in a generic `PersistenceException` when our exception is already one.

If a user is using `getCause()`, it will be an issue.

I tried to use `initCause()` with the initial exception in it to try to mitigate the issue but I got a `Self-causation not permitted` exception.

We may have to warn the Spring people about this change.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/exception/ExceptionTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/exception/ExceptionTest.java
@@ -111,8 +111,7 @@ public class ExceptionTest extends BaseEntityManagerFunctionalTestCase {
 			fail();
 		}
 		catch ( PersistenceException e ) {
-			Throwable t = e.getCause();
-			assertTrue( "Should be a constraint violation", t instanceof ConstraintViolationException );
+			assertTrue( "Should be a constraint violation", e instanceof ConstraintViolationException );
 			em.getTransaction().rollback();
 		}
 		finally {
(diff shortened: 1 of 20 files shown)


hibernate-core/src/test/java/org/hibernate/jpa/test/exception/ExceptionTest.java
```diff
@@ -111,8 +111,7 @@ public void testConstraintViolationException() throws Exception {
 			fail();
 		}
 		catch ( PersistenceException e ) {
-			Throwable t = e.getCause();
-			assertTrue( "Should be a constraint violation", t instanceof ConstraintViolationException );
+			assertTrue( "Should be a constraint violation", e instanceof ConstraintViolationException );
```
dreab8: this breaks 5.1 compatibility

hibernate-core/src/test/java/org/hibernate/jpa/test/transaction/TransactionJoiningTest.java
```diff
@@ -243,7 +243,7 @@ public void run() {
 			em.persist( new Book( "The Book of Foo", 1 ) );
 		}
 		catch ( PersistenceException e ) {
-			caught = e.getCause().getClass().equals( HibernateException.class );
+			caught = e.getClass().equals( HibernateException.class );
```
dreab8: also this breaks 5.1 compatibility

hibernate-core/src/test/java/org/hibernate/jpa/test/exception/ExceptionTest.java
gsmet: Does it?

It breaks 5.2 and early 5.3 compatibility for sure but the idea is to restore the 5.1 behavior.

gsmet:
Closing as apparently, we should wrap them in some cases but not all.

PR #2433, opened 2018-07-17
HHH-3138 - make constructor of DynamicMapEntityTuplizer public

aepedraza:
As requested in the ticket, I just changed the constructor's visibility from default (package) to public.

I commented in the ticket if this change is even desirable but got no answers. Maybe this is the proper site for discussion.


gsmet:
@aepedraza sorry, totally missed your comment on this issue. I think making it protected would be better. Subclasses could access it but it wouldn't be public.

Most of the methods in it are protected so it does not seem like a bad idea to make the constructor protected.

aepedraza:
@gsmet without knowing much about the code and not being able to reach out to the ticket reporter, seems to be the most "Solomonic" solution to make the constructor protected, which allows extensibility without exposing methods unnecessarily. I'll be changing it unless you want to argue something else.
I'll be awaiting your comments.

beikov:
Won't do that as we removed `EntityTuplizer` in 6.0

PR #2435, opened 2018-07-18
HHH-12104 @Where clause is ignored when eagerly loading ManyToOne association

gbadner:
https://hibernate.atlassian.net/browse/HHH-12104

I'll need to do a lot more testing on this, but this is what I have so far.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/LoadQueryJoinAndFetchProcessor.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/LoadQueryJoinAndFetchProcessor.java
@@ -222,16 +222,43 @@ public class LoadQueryJoinAndFetchProcessor {
 				: null;
 	}
 
-	private String resolveAdditionalJoinCondition(String rhsTableAlias, String withClause, Joinable joinable, AssociationType associationType) {
+	private String resolveAdditionalJoinCondition(
+			Join join,
+			String rhsTableAlias,
+			String withClause,
+			Joinable joinable,
+			AssociationType associationType) {
 		// turns out that the call to AssociationType#getOnCondition in the initial code really just translates to
 		// calls to the Joinable.filterFragment() method where the Joinable is either the entity or
 		// collection persister
-		final String filter = associationType!=null?
-				associationType.getOnCondition( rhsTableAlias, factory, queryInfluencers.getEnabledFilters() ):
-				joinable.filterFragment(
+
+		final String filter;
+		// If the left hand side
+		if ( associationType.isEntityType() &&
+				join.getLeftHandSide().getDisposition() != QuerySpace.Disposition.COLLECTION ) {
+			// We always need to get the filter (if any) for the following cases:
+			// * composite with entity fetch
+			// * entity with entity fetch
+			// EntityType#getOnCondition doesn't always do the right thing in this case,
+			// so explicitly call Joinable#filterFragment (instead of EntityType#getOnCondition.
+			filter = joinable.filterFragment(
 					rhsTableAlias,
 					queryInfluencers.getEnabledFilters()
-		);
+			);
+		}
+		else if ( associationType != null ) {
+			// We leave it up to the assocationType to in these cases:
+			// * entity with (element, one-to-many, or many-to-many) collection fetch
+			// * collection element with entity fetch.
+			filter = associationType.getOnCondition( rhsTableAlias, factory, queryInfluencers.getEnabledFilters() );
+		}
+		else {
+			// What falls into this category?
+			filter = joinable.filterFragment(
+						rhsTableAlias,
+						queryInfluencers.getEnabledFilters()
+				);
+		}
 
 		if ( StringHelper.isEmpty( withClause ) && StringHelper.isEmpty( filter ) ) {
 			return "";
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/LoadQueryJoinAndFetchProcessor.java
```diff
@@ -222,16 +222,43 @@ private AssociationType getJoinedAssociationTypeOrNull(Join join) {
...
 		// turns out that the call to AssociationType#getOnCondition in the initial code really just translates to
 		// calls to the Joinable.filterFragment() method where the Joinable is either the entity or
 		// collection persister
-		final String filter = associationType!=null?
-				associationType.getOnCondition( rhsTableAlias, factory, queryInfluencers.getEnabledFilters() ):
-				joinable.filterFragment(
+
+		final String filter;
+		// If the left hand side
+		if ( associationType.isEntityType() &&
```
gsmet: You will have a NPE here if `associationType` is null whereas before the code would support a null `associationType`.

hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/LoadQueryJoinAndFetchProcessor.java
```diff
@@ -222,16 +222,43 @@ private AssociationType getJoinedAssociationTypeOrNull(Join join) {
...
+				join.getLeftHandSide().getDisposition() != QuerySpace.Disposition.COLLECTION ) {
+			// We always need to get the filter (if any) for the following cases:
+			// * composite with entity fetch
+			// * entity with entity fetch
+			// EntityType#getOnCondition doesn't always do the right thing in this case,
```
gsmet: So shouldn't we fix it instead? Or is it used in other situations?

hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/LoadQueryJoinAndFetchProcessor.java
gbadner: Yes, you're right. I'll fix that.

hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/LoadQueryJoinAndFetchProcessor.java
gbadner: Changing EntityType#getOnCondition to work for this use case breaks other tests. The problem is that we we multiple ways of walking code in 5.x, and they seem to use EntityType#getOnCondition for different reasons.

@dreab8, is the walking code unified or Hibernate 6?

If so, we shouldn't spend a lot of resources on fixing this in Hibernate 5.

hibernate-core/src/test/java/org/hibernate/test/where/EagerAssociationWhereTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+
+	@Override
+	protected Class[] getAnnotatedClasses() {
+		return new Class[] { Product.class, Category.class };
+	}
+
+	@Test
+	@TestForIssue( jiraKey = "HHH-12104" )
+	//@FailureExpected( jiraKey = "HHH-12104" )
```
gsmet: This can be removed.

hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/LoadQueryJoinAndFetchProcessor.java
```diff
@@ -222,16 +222,44 @@ private AssociationType getJoinedAssociationTypeOrNull(Join join) {
...
+				// * entity with entity fetch
+				// EntityType#getOnCondition doesn't always do the right thing in this case,
+				// so explicitly call Joinable#filterFragment (instead of EntityType#getOnCondition.
+				filter = joinable.filterFragment(
+						rhsTableAlias,
+						queryInfluencers.getEnabledFilters()
+				);
+			}
+			else  {
+				// We leave it up to the assocationType to in these cases:
```
gsmet: s/to in these cases/in these cases/ ?

hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/LoadQueryJoinAndFetchProcessor.java
```diff
@@ -222,16 +222,44 @@ private AssociationType getJoinedAssociationTypeOrNull(Join join) {
...
+
+		final String filter;
+		if ( associationType != null ) {
+			if ( associationType.isEntityType() &&
+					join.getLeftHandSide().getDisposition() != QuerySpace.Disposition.COLLECTION ) {
+				// We always need to get the filter (if any) for the following cases:
+				// * composite with entity fetch
+				// * entity with entity fetch
+				// EntityType#getOnCondition doesn't always do the right thing in this case,
+				// so explicitly call Joinable#filterFragment (instead of EntityType#getOnCondition.
```
gsmet: Missing ).

hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/LoadQueryJoinAndFetchProcessor.java
gsmet: OK, that is unfortunate but you're right, it's probably better to be cautious about this one.

hibernate-core/src/main/java/org/hibernate/loader/plan/exec/internal/LoadQueryJoinAndFetchProcessor.java
gbadner: correct.

gbadner:
Closing because I don't think this is a safe fix.

PR #2436, opened 2018-07-18
[HHH-12823] CompositeIdTest.testDistinctCountOfEntityWithCompositeId …

simkam:
…fails on databases that don't support tuple distinct counts because it expects wrong exception

https://hibernate.atlassian.net/browse/HHH-12823

Could you please backport this fix to 5.3?

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/cid/CompositeIdTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/cid/CompositeIdTest.java
@@ -17,20 +17,18 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
+import javax.persistence.PersistenceException;
+
 import org.hibernate.Hibernate;
 import org.hibernate.Session;
 import org.hibernate.Transaction;
 import org.hibernate.dialect.Oracle8iDialect;
-import org.hibernate.dialect.PostgreSQL9Dialect;
-import org.hibernate.dialect.PostgresPlusDialect;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/test/java/org/hibernate/test/cid/CompositeIdTest.java
```diff
@@ -108,11 +105,11 @@ public void testDistinctCountOfEntityWithCompositeId() {
 		try {
 			long count = ( Long ) s.createQuery( "select count(distinct o) FROM Order o" ).uniqueResult();
 			if ( ! getDialect().supportsTupleDistinctCounts() ) {
-				fail( "expected SQLGrammarException" );
+				fail( "expected PersistenceException" );
 			}
 			assertEquals( 2l, count );
 		}
-		catch ( SQLGrammarException e ) {
+		catch ( PersistenceException e ) {
```
gsmet: @simkam 

I think this issue is due to the new exception handling but I think we should expect a `SQLGrammarException` wrapped in a `PersistenceException`.

Could you try with this change instead:
```
catch ( PersistenceException e ) {
    if ( !(e.getCause() instanceof SQLGrammarException) || getDialect().supportsTupleDistinctCounts() ) {
        throw e;
    }
}
```

simkam: yep, it works as well, PR updated

gsmet:
Merged, thanks!

PR #2437, opened 2018-07-18
[HHH_12824] ASTParserLoadingTest.testComponentNullnessChecks fails wi…

simkam:
…th DB2 because it uses legacy-style query parameter

https://hibernate.atlassian.net/browse/HHH-12824

Could you please backport fix also to 5.3?


gsmet:
Merged, thanks Martin!

PR #2438, opened 2018-07-18
[HHH_12825] CriteriaHQLAlignmentTest.testCountReturnValues fails on d…

simkam:
…atabases that don't support tuple distinct counts because it expects wrong exception

https://hibernate.atlassian.net/browse/HHH-12825

Could you please backport fix also to 5.3 branch? 

The fix is a bit different from #2436 only because it uses the same approach from the same test (line 350)


gsmet:
Merged, thanks!

PR #2439, opened 2018-07-19
HHH-12492 Jpa DELETE Query generated has missing table alias and thus incorrect semantics 

yrodiere:
https://hibernate.atlassian.net//browse/HHH-12492

Given the very sensitive section of the code I had to change, maybe we should wait for 5.4 before merging this. That being said, my first attempt was definitely buggy and at least 3 previously existing tests failed as a result, which is reassuring.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElement.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElement.java
@@ -341,12 +341,10 @@ public class FromElement extends HqlSqlWalkerNode implements DisplayableNode, Pa
 
 		final String propertyName = getIdentifierPropertyName();
 
-		if ( getWalker().getStatementType() == HqlSqlTokenTypes.SELECT ) {
-			return getPropertyMapping( propertyName ).toColumns( table, propertyName );
-		}
-		else {
-			return getPropertyMapping( propertyName ).toColumns( propertyName );
-		}
+		return toColumns(
+				table, propertyName,
+				getWalker().getStatementType() == HqlSqlTokenTypes.SELECT
+		);
 	}
 
 	public void setCollectionJoin(boolean collectionJoin) {
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/IdentNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/IdentNode.java
@@ -197,6 +197,15 @@ public class IdentNode extends FromReferenceNode implements SelectExpression {
 
 		String[] columnExpressions = element.getIdentityColumns();
 
+		// TODO Remove this? See HHH-12492
+		// This is legacy code that shouldn't be necessary, because getIdentityColumns is supposed to add the alias
+		if ( ! isFromElementUpdateOrDeleteRoot( element ) ) {
+			if ( StringHelper.isNotEmpty( element.getTableAlias() ) ) {
+				// apparently we also need to check that they are not already qualified.  Ugh!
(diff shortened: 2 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElement.java
```diff
@@ -172,7 +172,7 @@ public String getClassAlias() {
 		//return classAlias == null ? className : classAlias;
 	}
 
-	private String getTableName() {
+	public String getTableName() {
```
gsmet: I don't think this is needed?

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/IdentNode.java
```diff
@@ -197,14 +197,6 @@ private boolean resolveAsAlias() {
 
 		String[] columnExpressions = element.getIdentityColumns();
```
gsmet: So the general idea is to fix this `getIdentityColumns()` to avoid needing the workaround below.

yrodiere: Yep. Also, `getIdentityColumns` is only used in `IdentNode#resolveAsAlias`, so changing its behavior should not affect any other node or any other resolution of an `IdentNode`.

JWGmeligMeyling: Can confirm after some scrolling through the code that the replaced logic of `getIdentityColumn` was basically a simplified version of  the now used `toColumns(...)`, with indeed a special handling for subqueries. This change should be fine!

JWGmeligMeyling: After some scrolling through the related code, and having played with `FromElement` recently, I can confirm that the replaced code in `getIdentityColumns()` was indeed a simplified version of what happened in the `toColumns` function now. The latter indeed has a special handling for subqueries. This change looks good to me!

dreab8:
The PR looks fine to me just not 100% sure the change to the ```IdentNode``` is safe so I would limit the changes to the ```FromElement``` class
```
if ( getWalker().getStatementType() == HqlSqlTokenTypes.SELECT ) {
			return getPropertyMapping( propertyName ).toColumns( table, propertyName );
		}
		else {
			return toColumns(
					table, propertyName,
					getWalker().getStatementType() == HqlSqlTokenTypes.SELECT
			);
		}
```
this seems enough to solve the issue.

yrodiere:
> @yrodiere can you think of a case where keeping the code in IdentNode fails? I suppose it won't as there is this qualifyItNot() thing. But it's still better for correctness IMHO.

I also think it's confusing and just wrong to keep this piece of code. But I doubt it will ever get executed, so I guess in this regard it's "safer" to keep it.

I pushed a fix-up commit that restores this piece of code, let's see if any test fails.

beikov:
I haven't checked the details of the PR or specific DBMS, I just wanted to mention that I think this was done this way because some DBMS do not support an explicit alias for DML tables and instead require the use of the table name.

sebersole:
This.

yrodiere:
> I haven't checked the details of the PR or specific DBMS, I just wanted to mention that I think this was done this way because some DBMS do not support an explicit alias for DML tables and instead require the use of the table name.

Yes, this was accounted for.

yrodiere:
To be clearer: `FromElementType.toColumns` will use the table name to qualify columns when the "fromelement" is the target of an update/delete query, where aliases cannot be used.

gsmet:
Merged, thanks!

PR #2443, opened 2018-07-20
HHH-12666 : Document hibernate.native_exception_handling_51_complianc…

gbadner:
This adds documentation for hibernate.native_exception_handling_51_compliance to the migration guide.

PR #2444, opened 2018-07-20
HHH-12666

gbadner:
https://hibernate.atlassian.net/browse/HHH-12666

After discussing with @dreab8 , I've decided on a simpler approach.

`hibernate.native_exception_handling_51_compliance` will be added with possible values `true` or `false` (default). This property will indicate if exception handling for a `SessionFactory` built via Hibernate's native bootstrapping should behave the same as native exception handling in Hibernate ORM 5.1,

When set to true, `HibernateException` will not be wrapped or converted according to the JPA specification.

This setting will be ignored for a SessionFactory built via JPA bootstrapping.

I've incorporated tests from @yrodiere, which were extremely helpful. Thanks @yrodiere! I move and renamed the tests so, unfortunately, there are a lot of diffs. Sorry about that.

I've created a new jira to add more tests: https://hibernate.atlassian.net/browse/HHH-12837.

PR for 5.3 migration guide is: https://github.com/hibernate/hibernate-orm/pull/2443.


yrodiere:
@gbadner Glad the tests helped! Thanks for the fix :)

gsmet:
I reorganized the commits a bit to keep the history clean, removed an outdated comment and merged.

Thanks @gbadner and @yrodiere for your work on this!

PR #2445, opened 2018-07-20
[HHH_12838] AndNationalizedTests fails with DB2

simkam:
https://hibernate.atlassian.net/browse/HHH-12838


gsmet:
Good catch, merged, thanks!

PR #2451, opened 2018-07-20
[HHH-12842] Fix lazy loading for OneToOne associations

JWGmeligMeyling:
Suggested fix for: https://hibernate.atlassian.net/browse/HHH-12842

Todo: 
* ~Still waiting on full pass test results~
* The exact same approach should probably be applied to ManyToOne relations (currently using the `@NotFound(Ignore)` setting instead). I have a feeling that the solution should be more universal (perhaps a combination of the two).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
@@ -29,10 +29,11 @@ public class OneToOneType extends EntityType {
 	private final ForeignKeyDirection foreignKeyType;
 	private final String propertyName;
 	private final String entityName;
+	private final boolean isNullable;
 
 	/**
-	 * @deprecated Use {@link #OneToOneType(TypeFactory.TypeScope, String, ForeignKeyDirection, boolean, String, boolean, boolean, String, String)}
-	 *  instead.
+	 * @deprecated Use {@link #OneToOneType(TypeFactory.TypeScope, String, ForeignKeyDirection, boolean, String, boolean, boolean, boolean, String, String)}
+	 * instead.
 	 */
 	@Deprecated
 	public OneToOneType(
@@ -47,6 +48,11 @@ public class OneToOneType extends EntityType {
 		this( scope, referencedEntityName, foreignKeyType, uniqueKeyPropertyName == null, uniqueKeyPropertyName, lazy, unwrapProxy, entityName, propertyName );
 	}
 
+	/**
+	 * @deprecated Use {@link #OneToOneType(TypeFactory.TypeScope, String, ForeignKeyDirection, boolean, String, boolean, boolean, boolean, String, String)}
+	 * instead.
+	 */
+	@Deprecated
 	public OneToOneType(
 			TypeFactory.TypeScope scope,
 			String referencedEntityName,
@@ -57,10 +63,25 @@ public class OneToOneType extends EntityType {
 			boolean unwrapProxy,
 			String entityName,
 			String propertyName) {
+		this( scope, referencedEntityName, foreignKeyType, referenceToPrimaryKey, uniqueKeyPropertyName, lazy, unwrapProxy, foreignKeyType==ForeignKeyDirection.TO_PARENT, entityName, propertyName );
+	}
+
+	public OneToOneType(
+			TypeFactory.TypeScope scope,
+			String referencedEntityName,
+			ForeignKeyDirection foreignKeyType,
+			boolean referenceToPrimaryKey,
+			String uniqueKeyPropertyName,
+			boolean lazy,
+			boolean unwrapProxy,
+			boolean nullable,
+			String entityName,
+			String propertyName) {
 		super( scope, referencedEntityName, referenceToPrimaryKey, uniqueKeyPropertyName, !lazy, unwrapProxy );
 		this.foreignKeyType = foreignKeyType;
 		this.propertyName = propertyName;
 		this.entityName = entityName;
+		this.isNullable = nullable;
 	}
 
 	public OneToOneType(OneToOneType original, String superTypeEntityName) {
@@ -68,6 +89,7 @@ public class OneToOneType extends EntityType {
 		this.foreignKeyType = original.foreignKeyType;
 		this.propertyName = original.propertyName;
 		this.entityName = original.entityName;
+		this.isNullable = original.isNullable;
 	}
 
 	@Override
@@ -156,7 +178,7 @@ public class OneToOneType extends EntityType {
 
 	@Override
 	protected boolean isNullable() {
-		return foreignKeyType==ForeignKeyDirection.TO_PARENT;
+		return isNullable;
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/type/SpecialOneToOneType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/SpecialOneToOneType.java
@@ -27,7 +27,8 @@ import org.hibernate.engine.spi.SharedSessionContractImplementor;
 public class SpecialOneToOneType extends OneToOneType {
 	
 	/**
-	 * @deprecated Use {@link #SpecialOneToOneType(org.hibernate.type.TypeFactory.TypeScope, String, ForeignKeyDirection, boolean, String, boolean, boolean, String, String)} instead.
+	 * @deprecated Use {@link SpecialOneToOneType#SpecialOneToOneType(TypeFactory.TypeScope, String, ForeignKeyDirection, boolean, String, boolean, boolean, boolean, String, String)}
+	 * instead.
 	 */
 	@Deprecated
 	public SpecialOneToOneType(
@@ -41,7 +42,12 @@ public class SpecialOneToOneType extends OneToOneType {
 			String propertyName) {
 		this( scope, referencedEntityName, foreignKeyType, uniqueKeyPropertyName == null, uniqueKeyPropertyName, lazy, unwrapProxy, entityName, propertyName );
 	}
-	
+
+	/**
+	 * @deprecated Use {@link SpecialOneToOneType#SpecialOneToOneType(TypeFactory.TypeScope, String, ForeignKeyDirection, boolean, String, boolean, boolean, boolean, String, String)}
+	 * instead.
+	 */
+	@Deprecated
 	public SpecialOneToOneType(
 			TypeFactory.TypeScope scope,
 			String referencedEntityName,
@@ -52,7 +58,7 @@ public class SpecialOneToOneType extends OneToOneType {
 			boolean unwrapProxy,
 			String entityName,
 			String propertyName) {
-		super(
+		this(
 				scope,
 				referencedEntityName, 
 				foreignKeyType,
@@ -60,9 +66,35 @@ public class SpecialOneToOneType extends OneToOneType {
 				uniqueKeyPropertyName, 
 				lazy,
 				unwrapProxy,
+				foreignKeyType==ForeignKeyDirection.TO_PARENT,
 				entityName, 
 				propertyName
 			);
+	}
+
+	public SpecialOneToOneType(
+			TypeFactory.TypeScope scope,
+			String referencedEntityName,
+			ForeignKeyDirection foreignKeyType,
+			boolean referenceToPrimaryKey,
+			String uniqueKeyPropertyName,
+			boolean lazy,
+			boolean unwrapProxy,
+			boolean nullable,
+			String entityName,
+			String propertyName) {
+		super(
+				scope,
+				referencedEntityName,
+				foreignKeyType,
+				referenceToPrimaryKey,
+				uniqueKeyPropertyName,
+				lazy,
+				unwrapProxy,
+				nullable,
+				entityName,
+				propertyName
+		);
 	}
 
 	public SpecialOneToOneType(SpecialOneToOneType original, String superTypeEntityName) {
--- a/hibernate-core/src/test/java/org/hibernate/test/onetoone/lazy/LazyToOneTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/onetoone/lazy/LazyToOneTest.java
@@ -1,0 +1,166 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.test.onetoone.lazy;
+
+import org.hibernate.testing.TestForIssue;
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.junit.Before;
+import org.junit.Test;
+
+import javax.persistence.CascadeType;
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.MapsId;
+import javax.persistence.OneToOne;
+import javax.persistence.PrimaryKeyJoinColumn;
+import javax.persistence.Table;
+import java.util.Date;
+
+import static org.hibernate.Hibernate.isInitialized;
+import static org.hibernate.testing.transaction.TransactionUtil.doInHibernate;
+import static org.junit.Assert.assertFalse;
+
+@TestForIssue(jiraKey = "HHH-12842")
+public class LazyToOneTest extends BaseCoreFunctionalTestCase {
+
+	@Override
+	protected Class<?>[] getAnnotatedClasses() {
+		return new Class<?>[] { Post.class, PostDetails.class };
+	}
+
+	@Before
+	public void setUp() throws Exception {
+		doInHibernate( this::sessionFactory, s -> {
+			Post post = new Post();
+			post.setDetails( new PostDetails() );
+			post.getDetails().setCreatedBy( "ME" );
+			post.getDetails().setCreatedOn( new Date() );
+			post.setTitle( "title" );
+			s.persist( post );
+		} );
+	}
+
+	@Test
+	public void testOneToOneLazyLoading() {
+		doInHibernate( this::sessionFactory, s -> {
+			PostDetails post = (PostDetails) s.createQuery( "select a from PostDetails a" ).getResultList().get( 0 );
+			assertFalse( isInitialized( post.post ) );
+		} );
+	}
+
+	@Entity(name = "PostDetails")
+	@Table(name = "post_details")
+	public static class PostDetails {
+
+		@Id
+		@GeneratedValue
+		private Long id;
+
+		@Column(name = "created_on")
+		private Date createdOn;
+
+		@Column(name = "created_by")
+		private String createdBy;
+
+		@MapsId
+		@OneToOne(fetch = FetchType.LAZY, mappedBy = "details", optional = false)
+		private Post post;
+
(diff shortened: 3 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
```diff
@@ -55,19 +56,22 @@ public OneToOneType(
...
 		this.entityName = entityName;
+		this.isNullable = nullable;
 	}
 
 	public OneToOneType(OneToOneType original, String superTypeEntityName) {
 		super( original, superTypeEntityName );
 		this.foreignKeyType = original.foreignKeyType;
 		this.propertyName = original.propertyName;
 		this.entityName = original.entityName;
+		this.isNullable = original.isNullable;
```
JWGmeligMeyling: Probably we should introduce yet another constructor for extensions?

hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
```diff
@@ -55,19 +56,22 @@ public OneToOneType(
 			String uniqueKeyPropertyName,
 			boolean lazy,
 			boolean unwrapProxy,
+			boolean nullable,
```
JWGmeligMeyling: https://github.com/hibernate/hibernate-orm/pull/2451

gsmet: Yes please follow the same pattern used for the deprecated constructor above.

hibernate-core/src/main/java/org/hibernate/type/SpecialOneToOneType.java
```diff
@@ -60,6 +61,7 @@ public SpecialOneToOneType(
 				uniqueKeyPropertyName, 
 				lazy,
 				unwrapProxy,
+				nullable,
```
gsmet: Same here about adding a deprecated constructor.

hibernate-core/src/test/java/org/hibernate/test/onetoone/lazy/LazyToOneTest.java
```diff
@@ -0,0 +1,169 @@
...
+    public void setUp() throws Exception {
+        doInHibernate( this::sessionFactory, s -> {
+            Post post = new Post();
+            post.setDetails(new PostDetails());
+            post.getDetails().setCreatedBy("ME");
+            post.getDetails().setCreatedOn(new Date());
+            post.setTitle("title");
+            s.persist(post);
+            s.flush();
+            s.clear();
```
gsmet: I don't think you need to flush() and clear(). Considering this init is specific to the test method, I would have moved it to the test method and would remove this setUp() altogether.

hibernate-core/src/test/java/org/hibernate/test/onetoone/lazy/LazyToOneTest.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.Date;
+
+import static org.hibernate.Hibernate.isInitialized;
+import static org.hibernate.testing.transaction.TransactionUtil.doInHibernate;
+import static org.junit.Assert.assertFalse;
+
+@TestForIssue(jiraKey = "HHH-12842")
+public class LazyToOneTest extends BaseCoreFunctionalTestCase {
+
+    private NegativeValueSequenceTest.PositiveTwoIncrementSize a;
```
gsmet: This looks like something coming from another test and not really useful here?

hibernate-core/src/test/java/org/hibernate/test/onetoone/lazy/LazyToOneTest.java
JWGmeligMeyling: Just wanted make sure that I would not get an actual object instead of the expected proxy because the entity was already in the session. But I think this can't happen anyway as both blocks use a different session.

JWGmeligMeyling:
Addressed the feedback 👍

hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
```diff
@@ -29,6 +29,7 @@
 	private final ForeignKeyDirection foreignKeyType;
 	private final String propertyName;
 	private final String entityName;
+	private final boolean isNullable;
 
 	/**
 	 * @deprecated Use {@link #OneToOneType(TypeFactory.TypeScope, String, ForeignKeyDirection, boolean, String, boolean, boolean, String, String)}
```
gsmet: This reference should be changed to to point to the new constructor. Otherwise you kinda cascade the deprecations :).

hibernate-core/src/main/java/org/hibernate/type/SpecialOneToOneType.java
```diff
@@ -41,7 +41,12 @@ public SpecialOneToOneType(
 			String propertyName) {
 		this( scope, referencedEntityName, foreignKeyType, uniqueKeyPropertyName == null, uniqueKeyPropertyName, lazy, unwrapProxy, entityName, propertyName );
 	}
-	
+
+	/**
+	 * @deprecated Use {@link SpecialOneToOneType#SpecialOneToOneType(TypeFactory.TypeScope, String, ForeignKeyDirection, boolean, String, boolean, boolean, boolean, String, String)}
```
gsmet: Please also update the pointer of the preexisting deprecated constructor.

gsmet:
@JWGmeligMeyling added 2 more minor comments. Overall, it looks good, I'll do a proper review later today or early next week.

JWGmeligMeyling:
@gsmet Just changed the last two links. 

Still wondering whether the same should be applied for the `ManyToOne` scenerio though. Had a brief chat with @sebersole regarding the current implementation that checks against `@NotFound(Ignore)` on Gitter. Still wondering if this behaviour should be extended to somehow check for the value for `optional` as well.

gsmet:
@JWGmeligMeyling let's deal with the first case first.

I checked out your PR, a few comments:
 * you should indent with tabs, not spaces and we have some special padding rules (e.g. `String result = method( parameter )`). I can fix them and amend your commits but if you plan on contributing often, https://github.com/hibernate/hibernate-ide-codestyles might help. We usually don't reformat existing code but your could reformat entirely your new test with these rules.
 * more importantly, your test doesn't fail even without the second commit. E.g. it fails with `@FailureExpected` as it succeeds. If you remove the annotation it just works.

JWGmeligMeyling:
Hmm for some reason I'm testing against ManyToOne types with isLogicalOneToOne=true... Let my try to get a mapping that properly shows the issue.

JWGmeligMeyling:
Issue seemed to be related to: https://github.com/hibernate/hibernate-orm/pull/2451/files#diff-9aca9167532201db74410e7b0094d4a1R1794 . Switching to `PrimaryKeyJoinColumn` in my test instead showed the issue.

JWGmeligMeyling:
Still debugging it a bit more as I find this behaviour rather odd

JWGmeligMeyling:
Updated the PR.

JWGmeligMeyling:
Thanks for pointing me to the style guide by the way, must have overlooked it while I was going through the CONTRIBUTING.md, and naively just focussed on the checkstyle to pass :)

JWGmeligMeyling:
Reformatted the testcase.

JWGmeligMeyling:
@gsmet I'll extend my PR so that we use the same approach for ManyToOne's too. I thought about similar logic to what you describe.

Regarding the MapsId, AFAIK either one of the entities must persist first to generate the id value used in both. I think in old Hibernate versions I've used some sort of ForeignKeyIdentifierGenerator (is that still around?), but `MapsId` seemed to do the trick as well. I am always struggling with properly mapping OneToOnes. It should work in a more conventional mapping as well.

JWGmeligMeyling:
@gsmet 

Quickly applying this patch for `ManyToOne` relations result in quite a lot of failing test cases. Now, this may be because I haven't covered all cases yet (I didn't cover the ManyToOne binding in the `ModelBinder` and `CollectionBinding` yet, as I don't have quick access to optionality), but after a  quick look through the test failures, this seems to be related to "null" values being inserted in to what are now seen as "non-nullable" columns.

I think there is a subtle difference between "nullability" and "optionality/proxyability" (and the FK direction and "ignore-ability" that were used before for that fact). I think `isNullable` is a poor naming, because the method appears to be used differently in various scenarios, and appears to have a different meaning between the `org.hibernate.type.ManyToOneType` and `org.hibernate.mapping.ManyToOne` classes.

Now I am starting to be afraid that the same problems are applicable to OneToOne's, its just that they perhaps are poorly tested...

I'll keep debugging the failing tests and keep you informed about my process, but any suggestion as to how to proceed with the "redefinition" of the `isNullable` method (and its counterparts and potential new API additions) is very appreciated.

JWGmeligMeyling:
I've done some more research and I've found some surprising results:

- In the legacy implementation no such notion as `optional` existed
- There however was `constrained` for `one-to-one` relations, which indicates that a foreign key linking to the primary key of the associated entity should exist.

From "Java Persistence with Hibernate" (C. Bauer and G. King):

> The foreign key constraint means that a user [associated entity in the example] has to exist for a particular shipping address [association owner in this example], so a proxy can be enabled without hitting the database. Without this constraint, Hibernate has to hit the database to find out if there is a  user for the address; the proxy would then be redundant. (ch. 7, p. 280)

I think in modern Hibernate, `constrained="true"` is more or less assumed for owned associations, as I have observed hbm2ddl creating foreign key constraints by default. Also I don't see the JPA specification use `@ForeignKey` annotations in any example showing an association that is lazy-loaded. Nowhere the specification mentions that a `@ForeignKey` annotation would be required for lazy loading. So why was `constrained` ever required anyway? The documentation of the `@NotFound` annotation and its values `NotFoundAction` may shine some light on this:

>  Possible actions when an associated entity is not found in the database.  Often seen with "legacy" foreign-key schemes which do not use `NULL` to indicate a missing reference, instead using a "magic value".
> Emmanuel Bernard

This also explains the comments Steve made on Gitter, explaining why the "proxy-ability" of `ManyToOne` relations was based on the `@NotFound(Ignore.class)` annotation being absent:

> `@NotFound(Ignore)` means it cannot be proxied
> `@NotFound(Ignore)` means even though Hibernate sees a "valid" (non-null) FK value that might still not equate to an associated row
> we use the null / not-null state to decide whether to build the proxy

The handling for `@NotFound` was never ported to `OneToOne` relations. It seems it never existed in legacy (I don't see a `not-found` attribute for `one-to-one` in the hbm definition), and it's certainly missing now as the logic for handling the `@NotFound` annotation in `OneToOneType` is also missing. This also makes sense, because there is no need for backwards compatibility, and it seems we're moving in the direction of presuming `constrained="true"` anyway, which eliminates the purpose of the `@NotFound` annotation. This also explains the disparity between `OneToOneType#isNullable()` and `ManyToOneType#isNullable()`.

With this in mind, lets have another look at what these methods are supposed to do.

`ManyToOneType#isNullable()` implements `EntityType#isNullable` which is solely used in `EntityType#resolveIdentifier` and passed to `SharedSessionContractImplementor.internalLoad` as indicator whether a proxy may be created, or whether a row should exists and how to handle an empty result. It indicates whether a `null` value is allowed *provided a non-null identifier*. Emphasis added here because the latter caused some confusion to me personally. For example: it is completely irrelevant to check for the optionality of the association or nullability of its join columns here. For an empty optional association the result set would simply contain a `null` value for the foreign key, and not even attempt to create a proxy. In conclusion: `ManyToOne` associations work perfectly fine, with lazy and and all. So what changed in the scenario of the `OneToOne` association here?

`OneToOneType#isNullable()` implements  `EntityType#isNullable` as well, and does so by checking whether the foreign key type equals `TO_PARENT`. I.e. this effectively translates to: "do not proxy / lazy load the one to one relation if this association has a foreign key from child to parent." This sounds a bit like a workaround to me, but I think the idea here is to only proxy the "owned" side of the `OneToOne` association. I'm still not sure as to how this is a relevant check, because as mentioned above, at this time, we already assume that we possess a non-null FK value which is constrained. More on that later.

This foreign key type is set in `OneToOneSecondPass#doSecondPass(Map)` which is the very piece of code changed as part of the fix for [HHH-9460](https://github.com/hibernate/hibernate-orm/commit/2bfe30009c15ea59c587d11714d4be78b14f9ad9#diff-6af8d09ecfe77b1855d03b0ebc37fc28R97).

```diff
diff --git a/hibernate-core/src/main/java/org/hibernate/cfg/OneToOneSecondPass.java b/hibernate-core/src/main/java/org/hibernate/cfg/OneToOneSecondPass.java
index 1859f5586d..30d8a1b17e 100644
--- a/hibernate-core/src/main/java/org/hibernate/cfg/OneToOneSecondPass.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/OneToOneSecondPass.java
@@ -8,7 +8,6 @@ package org.hibernate.cfg;

 import java.util.Iterator;
 import java.util.Map;
-
 import javax.persistence.ConstraintMode;

 import org.hibernate.AnnotationException;
@@ -92,12 +91,19 @@ public class OneToOneSecondPass implements SecondPass {
 		value.setCascadeDeleteEnabled( cascadeOnDelete );
 		//value.setLazy( fetchMode != FetchMode.JOIN );

-		if ( !optional ) value.setConstrained( true );
-		value.setForeignKeyType(
-				value.isConstrained()
-						? ForeignKeyDirection.FROM_PARENT
-						: ForeignKeyDirection.TO_PARENT
-		);
+		if ( !optional ) {
+			value.setConstrained( true );
+		}
+		if ( value.isReferenceToPrimaryKey() ) {
+			value.setForeignKeyType( ForeignKeyDirection.TO_PARENT );
+		}
+		else {
+			value.setForeignKeyType(
+					value.isConstrained()
+							? ForeignKeyDirection.FROM_PARENT
+							: ForeignKeyDirection.TO_PARENT
+			);
+		}
 		PropertyBinder binder = new PropertyBinder();
 		binder.setName( propertyName );
 		binder.setValue( value );
```

In other words, `constrained` (which currently means `non-optional`, and I think incorrectly does so) one-to-one associations that reference a primary key, used to have `ForeignKeyDirection.FROM_PARENT` but now get `ForeignKeyDirection.TO_PARENT`, which causes `isNullable` to return true and prevents lazy loading to occur for the association.

Regardless of whether it is correct to decide nullability based on the ForeignKeyDirection, let's first get some more feeling for what this value actually should be, as there seems to be some undecidability about its definition for one-to-one associations.

The foreign key direction is for example used in ordering insert and delete statements.
For example, the foreignKeyDirection for `ManyToOneType` is always `FROM_PARENT`, because it is itself the owner of the association,
and therefore must be deleted prior to the deletion of the orphaned associations, or persisted after the persisting of the associations.

With this idea in mind, it can be easily observed what the problem is with the snippet above: simply because a value is a reference to the primary key, is not sufficient enough to decide which side owns the association. In fact, both sides of the association can
be the identifier, and this does in fact causes problems as reported in [HHH-12436](https://hibernate.atlassian.net/browse/HHH-12436), which is also a regression introduced with HHH-9460.

And in fact, the code proposed in #2226 to resolve HHH-12436 changes the logic to determine the foreign key direction based on
the owning side of the association:

```diff
diff --git a/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java b/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
index b819e699dd..9db6a6b4bf 100644
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
@@ -161,6 +161,7 @@ import org.hibernate.mapping.SingleTableSubclass;
 import org.hibernate.mapping.Subclass;
 import org.hibernate.mapping.ToOne;
 import org.hibernate.mapping.UnionSubclass;
+import org.hibernate.type.ForeignKeyDirection;
 
 import static org.hibernate.internal.CoreLogging.messageLogger;
 
@@ -3135,6 +3136,10 @@ public final class AnnotationBinder {
 				}
 			}
 		}
+		ForeignKeyDirection foreignKeyDirection = !BinderHelper.isEmptyAnnotationValue( mappedBy )
+				? ForeignKeyDirection.TO_PARENT
+				: ForeignKeyDirection.FROM_PARENT;
+
 		if ( trueOneToOne || mapToPK || !BinderHelper.isEmptyAnnotationValue( mappedBy ) ) {
 			//is a true one-to-one
 			//FIXME referencedColumnName ignored => ordering may fail.
@@ -3150,7 +3155,8 @@ public final class AnnotationBinder {
 					optional,
 					cascadeStrategy,
 					joinColumns,
-					context
+					context,
+					foreignKeyDirection
 			);
 			if ( inSecondPass ) {
 				secondPass.doSecondPass( context.getMetadataCollector().getEntityBindingMap() );
```

Which seems to be a much more solid implementation. And in fact, when I cherry-pick 97dad2816cb9a637f33f78bc250a7adf24d4299d on top of my branch, the test is fixed. Unfortunately, the fix introduces some new problems, as brought up by @gbadner in #2226:

> I debugged ForeignGeneratorViaMapsIdTest and found out why the association is loaded by a separate query.
> This test has a unidirectional one-to-one association that shares the primary key column.
> 
> ```java
> @entity
> public class MedicalHistory implements Serializable {
> @id
> Integer id;
> 
> @MapsId
> @JoinColumn(name = "patient_id")
> @OneToOne
> Person patient;
> }
> ```
> 
> Before the PR is applied, the foreign key direction for the association is `ForeignKeyDirection.TO_PARENT`, which is obviously incorrect.
> After the PR is applied, the foreign key direction for the association is `ForeignKeyDirection.FROM_PARENT`, which is correct.
> 
> When the EntityReturn for MedicalHistory is built, its AssociationKey is registered by   `AbstractLoadPlanBuildingAssociationVisitationStrategy#startingEntity`, which is the entity table and primary key for `MedicalHistory`.
> 
> Later, the `AssociationKey` for `MedicalHistory#patient` is registered by `EntityBasedAssociationAttribute#getAssociationKey`.
> 
> Before the PR is applied, the AssociationKey will be the entity table and column for the associated entity (Person) because the association has the incorrect ForeignKeyDirection. Because Hibernate does not detect a duplicated `AssociationKey`, an `EntityFetch` is ultimately added to the EntityReturn. As a result, the SQL statement generated will include what is needed to fetch both the `MedicalHistory` and its association (`#patient`).
> 
> After the PR is applied, the association's AssociationKey will be the same table and column name as was previously registered for `MedicalHistory` (because they share the same column). Hibernate determines that it is a "duplicate" `AssociationKey`, so an `EntityFetch` for the association is not generated. Instead, the SQL statement generated will only load `MedicalHistory`. Because the association is non-lazy, it gets loaded by a second query.
> 
> Some change needs to be made so that an EntityFetch gets added when the association uses the primary key(s).

In conclusion, I think we can say the following:

* The code proposed in this pull request is of no use. The code is based on some wrong assumptions on what `isNullable` is supposed to do.
* I think `isNullable` should be renamed to clearly describe its intent and expected use
* I think we should either deprecate `@NotFound`, or implement it for `@OneToOne` associations as well. (If its not applicable there, I am really wondering why?)
* The fix #2226 for HHH-12436 properly implements `getForeignKeyDirection` and uses the `constrained` value for `OneToOneType#isNullable()`, which fixes the issue I tried to resolve here ((HHH-12842)[https://hibernate.atlassian.net/browse/HHH-12842]). I hope that we can very resolve the fetch issues in that branch and merge #2226.
* If there's no change in getting the fix for HHH-12436 and HHH-12842 in time, please consider reverting https://github.com/hibernate/hibernate-orm/commit/2bfe30009c15ea59c587d11714d4be78b14f9ad9 so that at least both regressions are resolved.

PR #2453, opened 2018-07-23
HHH-12763 - Log which JtaPlatform implementation is used at startup on info level 

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12763

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/JtaPlatformInitiator.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/JtaPlatformInitiator.java
@@ -49,6 +49,7 @@ public class JtaPlatformInitiator implements StandardServiceInitiator<JtaPlatfor
 			platform = getFallbackProvider( configurationValues, registry );
 		}
 
+		LOG.usingJtaPlatform( platform != null ? platform.getClass().getName() : "null" );
 		return platform;
 	}
 
--- a/hibernate-core/src/test/java/org/hibernate/test/tool/schema/JtaPlatformLoggingTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/tool/schema/JtaPlatformLoggingTest.java
@@ -1,0 +1,80 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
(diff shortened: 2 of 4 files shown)


vladmihalcea:
@aepedraza I reworked your PR and changed it to use the Core Logger messages. Also, I wanted to print the implementation class only and not force to retrieve the TM or UT since the Service has just been created, but not initialized.

If anyone agrees, I'll close the other PR and integrate this one.

hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/JtaPlatformInitiator.java
```diff
@@ -49,6 +49,7 @@ public JtaPlatform initiateService(Map configurationValues, ServiceRegistryImple
 			platform = getFallbackProvider( configurationValues, registry );
 		}
 
+		LOG.usingJtaPlatform( platform != null ? platform.getClass().getName() : "null" );
```
gsmet: IMHO let's not log anything if null.

hibernate-core/src/test/java/org/hibernate/test/tool/schema/JtaPlatformLoggingTest.java
```diff
@@ -0,0 +1,80 @@
...
+	@Override
+	protected void addSettings(Map settings) {
+		TestingJtaBootstrap.prepare( settings );
+		settings.put( AvailableSettings.TRANSACTION_COORDINATOR_STRATEGY, "jta" );
+	}
+
+	@Test
+	public void test() {
+		assertEquals(
+				"HHH000485: Illegally attempted to associate a proxy for entity [org.hibernate.session.AssociateEntityWithTwoSessionsTest$Location] with id [1] with two open sessions.",
```
gsmet: This is weird? Why is it about a totally different message?

vladmihalcea:
@gsmet I'll add it for 5.4. The assert was fixed. Now, check out my comment about the logging `null`. If we agree on that, I'll integrate it.

hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/JtaPlatformInitiator.java
gsmet: OK, makes sense. Let's wait for the Travis build to pass before merging to master.

aepedraza:
Hi @vladmihalcea. Thanks for your review. Now I know how to add and test logging. BTW: _"add it to 5.4"_ does this mean to add it to a specific branch or something else?

gsmet:
Merged, thanks to both of you!

PR #2456, opened 2018-07-24
HHH-12792 : Document binary incompatibility of persisters and tuplizers (5.2)

gbadner:
https://hibernate.atlassian.net/browse/HHH-12792

Tuplizer and persister SPIs changed going from 5.1 to 5.2.


gsmet:
Merged, thanks!

PR #2457, opened 2018-07-25
HHH-12856 Support of sequences on DB2400

pierrickrouxel:
* https://hibernate.atlassian.net/browse/HHH-12856

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/DB2400Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/DB2400Dialect.java
@@ -32,6 +32,17 @@ public class DB2400Dialect extends DB2Dialect {
 	}
 
 	@Override
+	public String getLimitString(String sql, int offset, int limit) {
+		if ( offset > 0 ) {
+			throw new UnsupportedOperationException( "query result offset is not supported" );
+		}
+		if ( limit == 0 ) {
+			return sql;
+		}
+		return sql + " fetch first " + limit + " rows only ";
+	}
+
+	@Override
 	public String getForUpdateString() {
 		return " for update with rs";
 	}
--- a/hibernate-core/src/main/java/org/hibernate/dialect/DB2400Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/DB2400Dialect.java
@@ -21,79 +21,14 @@ import org.hibernate.engine.spi.RowSelection;
  */
 public class DB2400Dialect extends DB2Dialect {
 
-	private static final AbstractLimitHandler LIMIT_HANDLER = new AbstractLimitHandler() {
-		@Override
-		public String processSql(String sql, RowSelection selection) {
-			if ( LimitHelper.hasFirstRow( selection ) ) {
-				//nest the main query in an outer select
-				return "select * from ( select inner2_.*, rownumber() over(order by order of inner2_) as rownumber_ from ( "
-						+ sql + " fetch first " + getMaxOrLimit( selection ) + " rows only ) as inner2_ ) as inner1_ where rownumber_ > "
-						+ selection.getFirstRow() + " order by rownumber_";
-			}
-			return sql + " fetch first " + getMaxOrLimit( selection ) + " rows only";
-		}
-
-		@Override
-		public boolean supportsLimit() {
-			return true;
-		}
-
-		@Override
-		public boolean useMaxForLimit() {
-			return true;
-		}
-
-		@Override
-		public boolean supportsVariableLimit() {
-			return false;
-		}
-	};
-
 	@Override
 	public boolean supportsSequences() {
 		return false;
 	}
 
 	@Override
-	public boolean supportsLimit() {
-		return true;
-	}
-
-	@Override
 	public String getQuerySequencesString() {
 		return null;
-	}
-
-	@Override
-	@SuppressWarnings("deprecation")
-	public boolean supportsLimitOffset() {
-		return false;
-	}
-
-	@Override
-	public boolean useMaxForLimit() {
-		return true;
-	}
-
-	@Override
-	public boolean supportsVariableLimit() {
-		return false;
-	}
-
-	@Override
-	public String getLimitString(String sql, int offset, int limit) {
-		if ( offset > 0 ) {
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/DB2400V7R1Dialect.java
```diff
@@ -0,0 +1,42 @@
...
+import org.hibernate.dialect.unique.DefaultUniqueDelegate;
+import org.hibernate.dialect.unique.UniqueDelegate;
+
+/**
+ * An SQL dialect for i. This class provides support for DB2 Universal Database for i V7R1 and
+ * later, also known as DB2/400.
+ *
+ * @author Pierrick Rouxel (pierrickrouxel)
+ */
+public class DB2400V7R1Dialect extends DB2Dialect {
```
gsmet: I'm surprised it does not extend `DB2400Dialect`?

pierrickrouxel: It's voluntary because the real function of DB2400Dialect is emulate the sequences. But maybe I should have extend DB2400Dialect and revert the feature for a better understanding ?

gsmet: Agreed there is the sequence support and a lot of noise about limits that should be removed considering we have a specific limit handler but there is at least these methods which are different from the `DB2Dialect`:
```
	@Override
	public String getForUpdateString() {
		return " for update with rs";
	}

	@Override
	public IdentityColumnSupport getIdentityColumnSupport() {
		return new DB2390IdentityColumnSupport();
	}
```

pierrickrouxel:
For the DB2390IdentityColumnSupport V7R1 supports VALUES clause. There is no reason to continue to use sysibm.sysdummy1. 

I copied for update with rs.

hibernate-core/src/main/java/org/hibernate/dialect/DB2400V7R1Dialect.java
pierrickrouxel: For the DB2390IdentityColumnSupport V7R1 supports VALUES clause. There is no reason to continue to use sysibm.sysdummy1.

I copied for update with rs.

pierrickrouxel: I also replaced the unique delegate because the V7R1 supports unique constraint on column which has a different behaviour of the unique index with null. It makes unique compatible with Postgres, MySQL...

gsmet:
@pierrickrouxel I thought about it a bit more and I don't think we can include the automatic upgrade in 5.3 as the sequence won't be initialized. So the user will end up with duplicated keys (or most likely errors).

We can add this dialect to 5.3 and the user would have to enable it manually.

And we can add your PR to current master and add a note for the `migration-guide.adoc` document on how to initialize the sequence.

Would you be willing to prepare 2 differents PR along these lines:
- for master: your current patch + a note in `migration-guide.adoc` at the root of the git repo explaining how to initialize the sequence
- for 5.3: the new dialect but no automatic enablement.

pierrickrouxel:
I found a problem with sequences. qsys2.syssequences documents all sequences of system. Without filter by schema_name the test doesn't work. If there is 1 sequences in 1 schema and 2 connections on 2 different schemas, the sequence existence test will return true in both connections.

I will work on this.

hibernate-core/src/main/java/org/hibernate/dialect/DB2400Dialect.java
```diff
@@ -21,81 +21,16 @@
...
-			return sql;
-		}
-		return sql + " fetch first " + limit + " rows only ";
-	}
-
-	@Override
-	public LimitHandler getLimitHandler() {
-		return LIMIT_HANDLER;
-	}
-
```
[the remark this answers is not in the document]
pierrickrouxel: I removed this lines because the implementation is the same as the superclass. However rereading my commit I discovered I removed a restriction with offset that existed in an old OS version. I reintroduce it in the next commit.

hibernate-core/src/main/java/org/hibernate/dialect/DB2400V7R3Dialect.java
```diff
@@ -0,0 +1,51 @@
...
+import org.hibernate.dialect.unique.DefaultUniqueDelegate;
+import org.hibernate.dialect.unique.UniqueDelegate;
+
+/**
+ * An SQL dialect for i. This class provides support for DB2 Universal Database for i V7R1 and
+ * later, also known as DB2/400.
+ *
+ * @author Pierrick Rouxel (pierrickrouxel)
+ */
+public class DB2400V7R3Dialect extends DB2400Dialect {
```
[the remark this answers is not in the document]
pierrickrouxel: I changed the release number because the features used to upgrade the dialect are available in V7R1, but only installing Technology Refresh 11, also available in V7R2 with TR3 but in V7R3 they are shipped natively. I think it's more reliable. For this reason I fixed the documentation and put the dialect as default for V7R3.

hibernate-core/src/main/java/org/hibernate/dialect/DB2400Dialect.java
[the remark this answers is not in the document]
dreab8: I may be wrong but I think also 
	```
public boolean supportsLimitOffset() {	
		return false;	
	}
```
should be reintroduced.

dreab8:
applied upstream after reintroducing the `supportsLimitOffset`.

Thanks a lot @pierrickrouxel and sorry for the long time it took to be merged.

pierrickrouxel:
Thank you for the merge.

hibernate-core/src/main/java/org/hibernate/dialect/DB2400Dialect.java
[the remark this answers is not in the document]
icgUser: Hi, shouldn't it be set to true?

Otherwise offset will always come to zero in the getLimitString method, so it will always execute the query as in the getLimitString method of the DB2400Dialect dialect, without offset.

PR #2458, opened 2018-07-25
HHH-12855 - Add query plan compilation statistics

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12855

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/query/spi/QueryPlanCache.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/query/spi/QueryPlanCache.java
@@ -13,6 +13,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 
 import org.hibernate.Filter;
 import org.hibernate.MappingException;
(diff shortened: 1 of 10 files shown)


vladmihalcea:
@sebersole If you have time, can you take a look on this issue? Thanks.

hibernate-core/src/main/java/org/hibernate/engine/query/spi/QueryPlanCache.java
```diff
@@ -148,13 +149,28 @@ public HQLQueryPlan getHQLQueryPlan(String queryString, boolean shallow, Map<Str
 			throws QueryException, MappingException {
 		final HQLQueryPlanKey key = new HQLQueryPlanKey( queryString, shallow, enabledFilters );
 		HQLQueryPlan value = (HQLQueryPlan) queryPlanCache.get( key );
+		boolean stats = factory.getStatistics().isStatisticsEnabled();
+
 		if ( value == null ) {
+			final long startTime = System.nanoTime();
```
gsmet: It would be nice if we could skip that when statistics are not enabled.

`System.nanoTime()` adds some overhead.

hibernate-core/src/main/java/org/hibernate/engine/query/spi/QueryPlanCache.java
```diff
@@ -148,13 +149,28 @@ public HQLQueryPlan getHQLQueryPlan(String queryString, boolean shallow, Map<Str
...
+				final long microseconds = TimeUnit.MICROSECONDS.convert( endTime - startTime, TimeUnit.NANOSECONDS );
+				factory.getStatistics().queryCompiled( queryString, microseconds );
+			}
+
 			queryPlanCache.putIfAbsent( key, value );
 		}
 		else {
 			LOG.tracev( "Located HQL query plan in cache ({0})", queryString );
+
+			if(stats) {
```
gsmet: Style.

hibernate-core/src/main/java/org/hibernate/stat/internal/QueryStatisticsImpl.java
```diff
@@ -156,6 +182,18 @@ void executed(long rows, long time) {
 		}
 	}
 
+	/**
+	 * add plan statistics report of a DB query
+	 *
+	 * @param microseconds time taken
+	 */
+	void compiled(long microseconds) {
+		log.tracef( "QueryStatistics - query compiled : %s", query );
```
gsmet: We are in the process of removing the other logger calls so let's not add more.

hibernate-core/src/main/java/org/hibernate/stat/internal/QueryStatisticsImpl.java
```diff
@@ -174,12 +212,20 @@ void incrementCachePutCount() {
 		cachePutCount.increment();
 	}
 
+	void incrementPlanCacheHitCount() {
+		log.tracef( "QueryStatistics - plan cache hit : %s", query );
```
gsmet: We are in the process of removing the other logger calls so let's not add more.

hibernate-core/src/main/java/org/hibernate/stat/internal/StatisticsImpl.java
```diff
@@ -172,6 +175,9 @@ public void clear() {
 		deprecatedNaturalIdStatsMap.clear();
 
 		startTime = System.currentTimeMillis();
+
+		queryPlanCacheHitCount.reset();
+		queryPlanCacheMissCount.reset();
```
gsmet: It would be more consistent if these calls were before the `startTime` resolution.

hibernate-core/src/main/java/org/hibernate/stat/internal/StatisticsImpl.java
```diff
@@ -807,7 +806,42 @@ public void queryCachePut(String hql, String regionName) {
...
+		queryPlanCacheMissCount.increment();
 
+		if ( hql != null ) {
+			getQueryStatistics( hql ).compiled( microseconds );
+		}
+	}
+
+	@Override
+	public void queryPlanCacheHit(String hql) {
+		LOG.tracef( "Statistics#queryPlanCachePut( `%s`, `%s` )", hql );
```
gsmet: Same here, let's not add a call to the logger. There's no real value.

gsmet:
Oh and additional comment: we need to add them in the new documentation you recently added.

+ there's a merge conflict you should fix in the logger.

PR #2459, opened 2018-07-26
HHH-12770 - Document @NotFound(action = NotFoundAction.IGNORE) and FetchType.LAZY behavior

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12770


documentation/src/main/asciidoc/userguide/chapters/domain/associations.adoc
```diff
@@ -425,6 +425,12 @@ However, you can configure this behavior so that Hibernate can ignore such an Ex
 
 To ignore non-existing parent entity references, even though not really recommended, it's possible to use the annotation `org.hibernate.annotation.NotFound` annotation with a value of `org.hibernate.annotations.NotFoundAction.IGNORE`.
 
+[NOTE]
+====
+`@ManyToOne` and `@OneToOne` associations annotated with `@NotFound(action = NotFoundAction.IGNORE)` are always fetched eagerly
```
JWGmeligMeyling: Is this true for `OneToOne`'s though? Seems to only check the FK direction (because this is not sufficient you proposed to change it to `!constrained`, which is correct). Didn't see the handling for `@NotFound(Ignore)` there. Does `@NotFound(Ignore)` set `!constrained`?

vladmihalcea: @JWGmeligMeyling There's a test for `@OneToOne` too. It's for the child-side where it replicates just like for `@ManyToOne`.

gsmet:
+1 to merge to master only.

PR #2460, opened 2018-07-26
HHH-11209 NullPointerException in EntityType.replace() with a PersistentBag

gbadner:

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/AbstractPersistentCollection.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/AbstractPersistentCollection.java
@@ -624,6 +624,9 @@ public abstract class AbstractPersistentCollection implements Serializable, Pers
 		prepareForPossibleLoadingOutsideTransaction();
 		if ( currentSession == this.session ) {
 			if ( !isTempSession ) {
+				if ( hasQueuedOperations() ) {
+					LOG.queuedOperationWhenDetachFromSession( MessageHelper.collectionInfoString( getRole(), getKey() ) );
+				}
 				this.session = null;
 			}
 			return true;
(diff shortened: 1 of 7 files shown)


gbadner:
If it looks ok, please don't merge quite yet. 

I think there's another case that needs to be considered that will have the same issue.  If so, I'll add a failing test to the PR. I'll let you know if it turns out to not be a problem

hibernate-core/src/main/java/org/hibernate/collection/internal/AbstractPersistentCollection.java
```diff
@@ -620,6 +624,9 @@ public final boolean unsetSession(SharedSessionContractImplementor currentSessio
 		prepareForPossibleLoadingOutsideTransaction();
 		if ( currentSession == this.session ) {
 			if ( !isTempSession ) {
+				if ( hasQueuedOperations() ) {
+					LOG.queuedOperationWhenDetachFromSession( MessageHelper.collectionInfoString( getRole(), getKey() ) );
+				}
```
gbadner: Logs a warning if a collection is detached with queued operations. 

This warning is known to be logged in the following situations, as illustrated by running the unit tests:
* the application explicitly detaches the collection owner entity while it has queued operations;
* Hibernate detaches the collection with queued operations after rolling back the transaction.

hibernate-core/src/main/java/org/hibernate/collection/internal/AbstractPersistentCollection.java
```diff
@@ -647,25 +654,22 @@ public final boolean setCurrentSession(SharedSessionContractImplementor session)
...
 			}
 			else {
-				this.session = session;
-				return true;
+				LOG.logUnexpectedSessionInCollectionNotConnected( msg );
 			}
 		}
+		if ( hasQueuedOperations() ) {
+			LOG.queuedOperationWhenAttachToSession( MessageHelper.collectionInfoString( getRole(), getKey() ) );
+		}
```
gbadner: Logs a warning when a collection with queued operations is attached to the session.

This is known to happen when calling `Session#save`, `#saveOrUpdate`, or `#update` on an entity with a collection with queued operations. Provided the collection is not the target of a merge operation, the queued operations will be executed and cleared when the collection is initialized or the session is flushed.

hibernate-core/src/test/java/org/hibernate/test/collection/delayedOperation/DetachedBagDelayedOperationTest.java
```diff
@@ -0,0 +1,387 @@
...
+					);
+					triggerableIgnoreQueuedOperationsOnMerge.reset();
+
+					assertFalse( Hibernate.isInitialized( p.getChildren() ) );
+					assertFalse( ( (AbstractPersistentCollection) p.getChildren() ).hasQueuedOperations() );
+
+					// When initialized, p.children will not include the new Child ("Zeke"),
+					// because that Child was flushed without a parent before being detached
+					// along with its parent.
+					Hibernate.initialize( p.getChildren() );
```
gbadner: When initialized, the collection does not contain the `Child` that was added to the collection by the queued operations before being detached. That's because the queued operations in the entity being merged (`pWithQueuedOperations`) were ignored.

hibernate-core/src/test/java/org/hibernate/test/collection/delayedOperation/DetachedBagDelayedOperationTest.java
```diff
@@ -0,0 +1,387 @@
...
+					// Make sure nothing else got triggered
+					checkTriggerablesNotTriggered();
+
+					assertFalse( Hibernate.isInitialized( pAfterDetachWithQueuedOperations.getChildren() ) );
+					assertTrue( ( (AbstractPersistentCollection) pAfterDetachWithQueuedOperations.getChildren() ).hasQueuedOperations() );
+
+					// Queued operations will be executed when the collection is initialized,
+					// After initialization, the collection will contain the Child that was added as a
+					// queued operation before being detached above.
+					Hibernate.initialize( pAfterDetachWithQueuedOperations.getChildren() );
```
gbadner: When initialized, the collection **contains** the Child that was added to the collection by the queued operations before being detached. That's because `pAfterDetachWithQueuedOperations` was attached to the session  using `Session#saveOrUpdate`.

gbadner:
I've fixed the type, rebased, squashed and pushed to master.

PR #2461, opened 2018-07-27
HHH-7318 Hibernate incorrecly registers BIGINT as corresponding to Java's BigInteger type.

yrodiere:
https://hibernate.atlassian.net//browse/HHH-7318

The fix was trivial, though testing it was not (result type auto-discovery behaves quite differently depending on the underlying database).

I'm a bit worried about the fix, to be honest:

* The backward incompatibility is obvious: where users would previously get a `BigInteger`, they will now get a `Long`. We might be ok with that (because it's a bugfix) or not (because this bug has been around for a long, long time).
* While most databases ([PostgreSQL](https://www.postgresql.org/docs/8.2/static/datatype-numeric.html), [SQL Server](https://docs.microsoft.com/en-us/sql/t-sql/data-types/int-bigint-smallint-and-tinyint-transact-sql?view=sql-server-2017), [MySQL](https://dev.mysql.com/doc/refman/5.7/en/integer-types.html)) indeed store the equivalent of `java.lang.Long` (integer ranging from `-2^63` to `2^63-1`) into the `BIGINT` SQL type, there are exceptions:
  * MySQL allows using the `BIGINT UNSIGNED` type, which allows to store  `0` to `2^64-1`. This could be a problem in legacy schemas that were not desgined to fit the schema generated by Hibernate ORM exactly. However, we already have the same problem for `TINYINT`, `SMALLINT`, `INTEGER`, so I guess it's not such a big deal. See https://dev.mysql.com/doc/refman/5.7/en/integer-types.html
  * Oracle internally only uses a `NUMBER(precision, scale)` datatype for all integer types, so I'm not sure how the JDBC driver (which is obviously closed-source) will report the type of integer columns. I spent more than an hour trying to make our build work with Oracle locally and I couldn't, so I gave up. I guess CI will tell us.

Maybe this change would be better suited for 5.4, or even 6.0?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -213,7 +213,7 @@ public abstract class Dialect implements ConversionContext {
 		registerColumnType( Types.NCLOB, "nclob" );
 
 		// register hibernate types for default use in scalar sqlquery type auto detection
-		registerHibernateType( Types.BIGINT, StandardBasicTypes.BIG_INTEGER.getName() );
+		registerHibernateType( Types.BIGINT, StandardBasicTypes.LONG.getName() );
 		registerHibernateType( Types.BINARY, StandardBasicTypes.BINARY.getName() );
 		registerHibernateType( Types.BIT, StandardBasicTypes.BOOLEAN.getName() );
 		registerHibernateType( Types.BOOLEAN, StandardBasicTypes.BOOLEAN.getName() );
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/query/NativeQueryResultTypeAutoDiscoveryTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/query/NativeQueryResultTypeAutoDiscoveryTest.java
@@ -1,0 +1,596 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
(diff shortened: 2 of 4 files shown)


sebersole:
I'll look at this from the perspective of changing it in 6.0.  I do not think this is something we want to change prior to 6.

sebersole:
Actually I saw that Gavin had already made this change in his Dialect work.  I'll apply your test however

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -213,7 +213,7 @@ protected Dialect() {
 		registerColumnType( Types.NCLOB, "nclob" );
 
 		// register hibernate types for default use in scalar sqlquery type auto detection
-		registerHibernateType( Types.BIGINT, StandardBasicTypes.BIG_INTEGER.getName() );
+		registerHibernateType( Types.BIGINT, StandardBasicTypes.LONG.getName() );
 		registerHibernateType( Types.BINARY, StandardBasicTypes.BINARY.getName() );
```
NathanQingyangXu: Just found the above change has been done in v6 by @gavinking already

sebersole:
As mentioned, the functional change here was already handled by a PR from @gavinking - I basically just took his src/main changes and applied your test addition and changes on top.

Those tests all fail for "unrelated" reasons - mainly support for NativeQuery and import.sql handling not yet being implemented. I created a new follow-up PR (#3304) that moves these tests to the new package structure which actually runs them as part of the build.  These tests fail for the mentioned reason, so for the time being that should be considered a "draft" PR

lbenedetto:
Shouldn't this change be documented in the migration guide?

PR #2462, opened 2018-07-27
[HHH-12863] SchemaUpdateTest should be skipped with Sybase

simkam:
https://hibernate.atlassian.net/browse/HHH-12863


gsmet:
Merged, thanks!

PR #2464, opened 2018-07-27
HHH-12857 Support the security manager with ByteBuddy

gsmet:
https://hibernate.atlassian.net/browse/HHH-12857

So this is my attempt to support the security manager with ByteBuddy as the bytecode provider.

Thanks a lot to @raphw for his support along the process!

@Naros could you run the tests you were running previously with this PR? Hopefully, it will work with only minor tuning.

And I think we will need a thorough review of this one.


Naros:
@gsmet Tested this against the latest WF repo and it worked just fine.

gsmet:
Merged.

PR #2466, opened 2018-07-30
HHH-12868 Fix NPE when loading entity with CacheConcurrencyStrategy.NONE

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12868


gsmet:
Merged.

PR #2467, opened 2018-07-30
HHH-12872 Reduce memory consumption of XSD schema validations

Sanne:
For master only.

 - https://hibernate.atlassian.net/browse/HHH-12872


gsmet:
Merged, thanks!

PR #2468, opened 2018-07-31
Test for Issue HHH-12870

pb00068:
PR includes 2 test-methods which worked with version 5.3.2.
Now they pass only with the modification (commented lines) in AbstractSharedSessionContract.java


sebersole:
See https://hibernate.atlassian.net/browse/HHH-18864

PR #2469, opened 2018-07-31
HHH-12867 Lazy loaded OneToMany gets initialized on refresh

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12867

This is just a test case as I couldn't come up with a fix without triggering other test failures.


alostale:
Being this an accepted misbehavior, shouldn't the other failing test cases be fixed to be adapted to the new (and correct) behavior?

gsmet:
Sure. I think there's something wrong but unfortunately I'm not fully understanding what's going on with those failing tests, that's why I haven't pushed a fix for now.

dreab8:
Thanks @gsmet for the test case, I added a `@FailureExpected` annotation and merged it into main.

PR #2470, opened 2018-07-31
HHH-12869 Do not check the cacheManager is not null when creating it

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12869


gsmet:
Merged.

PR #2472, opened 2018-07-31
HHH-12848 Restore the original Oracle LimitHandler

gsmet:
Oracle does not support using the SQL 2008 standard limit handler with
FOR UPDATE clauses. Thus we need to get back to the old limit handler.

https://hibernate.atlassian.net/browse/HHH-12848


gsmet:
Merged.

PR #2473, opened 2018-07-31
HHH-10603 Avoid doing distinct and comparisons on byte arrays

gsmet:
They are stored as blobs starting with Oracle12cDialect and distinct and
comparisons on blobs are not supported.

Some tests were adapted, some are now skipped with Oracle12cDialect.

 * https://hibernate.atlassian.net/browse/HHH-10603


gsmet:
Merged.

PR #2450, opened 2018-07-20
[HHH-12842] Test that shows OneToOne won't be lazyloaded anymore

jwgmeligmeyling:
Testcase for https://hibernate.atlassian.net/browse/HHH-12842

PR #2474, opened 2018-08-01
[HHH-12842] Test that shows OneToOne won't be lazyloaded anymore

JWGmeligMeyling:
Testcase that shows that a non-optional `OneToOne`  to an id property mapping cannot be lazy loaded anymore due to a regression since 5.2.13.

Testcase for https://hibernate.atlassian.net/browse/HHH-12842
Duplicates  #2450 (which was closed but couldn't be reopened)
The changes suggested in #2226 fix this issue.

EDIT gsmet: so fixing https://hibernate.atlassian.net/browse/HHH-12436 might fix this issue and in this case we would just need to add the test to the test suite.


gsmet:
Thanks @JWGmeligMeyling !

PR #2475, opened 2018-08-01
HHH-12849 - QuotedIdentifierTest fails with ORA-04043 on Oracle12c

dreab8:
https://hibernate.atlassian.net/browse/HHH-12849
@gsmet @simkam I have disabled the test adding a comment about the issue.
What do you think?


gsmet:
@dreab8 if we can reproduce it with pure JDBC calls, shouldn't we open an Oracle issue? I don't know how we should report it and if they even care if you're not a customer.

gsmet:
Merged it, thanks!

PR #2476, opened 2018-08-01
HHH-12847 Consider LockOptions for getting the LockMode

gsmet:
Otherwise, it leads to null pointer exceptions.

The NPE is fairly recent and is due to a recent change but we had a bug before the NPE as the `LockMode` set in the `LockOptions` was totally ignored.

So all in all, the NPE allowed to find a hidden bug.

 * https://hibernate.atlassian.net/browse/HHH-12847


gsmet:
Merged.

PR #2477, opened 2018-08-01
HHH-12196 Implement a naive limit handler for Sybase

gsmet:
It doesn't manage all the corner cases but it should be safe enough as
only triggered in the simple cases.

https://hibernate.atlassian.net/browse/HHH-12196

@simkam we have a customer asking for it in 5.3 and it's arguably a bug fix considering the current behavior is very bad (it loads all the rows even if you specify a limit). Could you test this branch with Sybase to check it doesn't introduce any regression in our compatibility matrix? That would be very helpful!


simkam:
@gsmet sure, I queued build but queue is quite long now (1500 builds). We will have results tomorrow if we are lucky (and I didn't make configuration error)

gsmet:
Thanks for the reviews, merged!

PR #2479, opened 2018-08-02
HHH-12875 Class level where="..." clause in hbm.xml mappings is not enforced on collections of that class

gbadner:
https://hibernate.atlassian.net/browse/HHH-12875

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
@@ -1527,7 +1527,6 @@ public class ModelBinder {
 			binding.getSynchronizedTables().add( name );
 		}
 
-		binding.setWhere( source.getWhere() );
 		binding.setLoaderName( source.getCustomLoaderName() );
 		if ( source.getCustomSqlInsert() != null ) {
 			binding.setCustomSQLInsert(
@@ -3448,9 +3447,15 @@ public class ModelBinder {
 
 				final PersistentClass referencedEntityBinding = mappingDocument.getMetadataCollector()
 						.getEntityBinding( elementSource.getReferencedEntityName() );
+				collectionBinding.setWhere(
+					StringHelper.getNonEmptyOrConjunctionIfBothNonEmpty(
+							referencedEntityBinding.getWhere(),
+							getPluralAttributeSource().getWhere()
+					)
+				);
+
 				elementBinding.setReferencedEntityName( referencedEntityBinding.getEntityName() );
 				elementBinding.setAssociatedClass( referencedEntityBinding );
-
 				elementBinding.setIgnoreNotFound( elementSource.isIgnoreNotFound() );
 			}
 			else if ( getPluralAttributeSource().getElementSource() instanceof PluralAttributeElementSourceManyToMany ) {
@@ -3469,12 +3474,17 @@ public class ModelBinder {
 						new RelationalObjectBinder.ColumnNamingDelegate() {
 							@Override
 							public Identifier determineImplicitName(final LocalMetadataBuildingContext context) {
-								return context.getMetadataCollector().getDatabase().toIdentifier( Collection.DEFAULT_ELEMENT_COLUMN_NAME );
+								return context.getMetadataCollector()
+										.getDatabase()
+										.toIdentifier( Collection.DEFAULT_ELEMENT_COLUMN_NAME );
 							}
 						}
 				);
 
-				elementBinding.setLazy( elementSource.getFetchCharacteristics().getFetchTiming() != FetchTiming.IMMEDIATE );
+				elementBinding.setLazy(
+						elementSource.getFetchCharacteristics()
+								.getFetchTiming() != FetchTiming.IMMEDIATE
+				);
 				elementBinding.setFetchMode(
 						elementSource.getFetchCharacteristics().getFetchStyle() == FetchStyle.SELECT
 								? FetchMode.SELECT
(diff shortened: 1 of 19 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
```diff
@@ -3494,7 +3494,20 @@ public Identifier determineImplicitName(final LocalMetadataBuildingContext conte
...
+						.getEntityBinding( elementSource.getReferencedEntityName() );
+				if ( StringHelper.isNotEmpty( referencedEntityBinding.getWhere() ) ) {
+					whereBuffer.append( referencedEntityBinding.getWhere() );
+				}
+				if ( StringHelper.isNotEmpty( elementSource.getWhere() ) ) {
+					if ( whereBuffer.length() > 0 ) {
+						whereBuffer.append( " and " );
+					}
+					whereBuffer.append( elementSource.getWhere() );
+				}
```
gsmet: I think you need to add extra parentheses around the conditions. Otherwise you could easily be fooled with `OR` conditions.

AFAICS, the test does not exercise the case when we have 2 conditions. It would be nice to have a test for that.

gbadner: @gsmet, I've come up with a test that reproduces incorrect results due to missing parentheses. This problem affects mapping with annotations as well as hbm.xml.

Since HHH-12875 specifically affects hbm.xml, I'll creat a new jira issue to deal with the parentheses for both annotations and hbm.xml

gbadner: I've created https://hibernate.atlassian.net/browse/HHH-12882 to add the parentheses.

gbadner:
@gsmet, I've added tests, and fixed the parentheses issue (now HHH-12882).

It causes some MultiTableTest failures because of an unrelated issue having to do with a where clause declared on the root of a joined-subclass hierarchy. The clause is getting added for the wrong table. I'll create a jira for that one tomorrow.

hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
```diff
@@ -3494,7 +3504,16 @@ public Identifier determineImplicitName(final LocalMetadataBuildingContext conte
 
 				getCollectionBinding().setElement( elementBinding );
 
-				getCollectionBinding().setManyToManyWhere( elementSource.getWhere() );
+				final PersistentClass referencedEntityBinding = mappingDocument.getMetadataCollector().getEntityBinding(
+						elementSource.getReferencedEntityName()
+				);
+				getCollectionBinding().setWhere( getPluralAttributeSource().getWhere() );
```
gsmet: When is this one used and why is it different?

hibernate-core/src/main/java/org/hibernate/internal/util/StringHelper.java
```diff
@@ -852,4 +852,38 @@ public static String nullIfEmpty(String value) {
...
+					.append( " ) and ( ")
+					.append( secondExpression )
+					.append( " )" );
+			return buffer.toString();
+		}
+		else if ( isFirstExpressionNonEmpty ) {
+			return firstExpression;
+		}
+		else if ( isSecondExpressionNonEmpty ) {
+			return secondExpression;
```
gsmet: I think I would add parentheses around the clause anyway, just to be safe. Or we are sure the query can't contain any other condition?

gsmet: I thought a bit more about it and I think it would definitely be safer to add parentheses in all cases. At some point we might add features that contribute elements to the where clause and it needs to be properly isolated.

gsmet:
@gbadner I added a commit to handle the parentheses a bit differently.

gbadner:
@gsmet, the value returned by Collection#getWhere is already enclosed by parentheses here:
https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java#L280

The value returned by Collection#getManyToManyWhere is already enclosed by parentheses here:
https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java#L573-L575

In hibernate-orm, it is generally the consumer that adds the outer-most parentheses. That's why I purposely left out the outer-most parentheses in `StringHelper#getNonEmptyOrConjunctionIfBothNonEmpty`.

Also, I seem to remember that having multiple sets of outer parentheses can cause a problem on some DB, although I could be mis-remembering.

hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
gbadner: I've added comments to clarify.

gsmet:
@gbadner removed my commit and force pushed.

gsmet:
Hi @gbadner ,

I confirm the `MultiTableTest` failures are due to https://hibernate.atlassian.net/browse/HHH-12016 .

Removing `where="id1_ is not null"` from `src/test/java/org/hibernate/test/legacy/Multi.hbm.xml` fixes the test failures and the build in general.

I took a look at HHH-12016 and I don't think it's possible to fix it without major changes. I would vote for talking to Steve about this issue to see if it will be addressed in 6.

My recommendation would be to remove the `where="id1_ is not null"` element from the mapping file for now and move on to work on the 5.3 regressions.

gbadner:
@gsmet, thanks for checking the test failures. I agree we can wait for 6 to fix HHH-12016. 

HHH-12016 was already an issue for annotations before this fix. After this fix, behavior using hbm.xml mappings will be consistent with annotations, which causes the MultiTableTest failures. I'll add a migration note.

I'm surprised all of the test passes when `where="id1_ is not null"` is removed. I'll take a look at that before removing it.

gbadner:
@gsmet, I think it would be best to add an option to allow disabling inclusion of entity class-level where="..." restrictions when the entities are included in a collection, since this was a new feature introduced in 5.2.

If we don't do that, there could be behavior changes when 5.1 applications are migrated to 5.3.

gbadner:
Also, MultiTableTest could have hibernate.use_entity_where_clause_for_collections=false, so it would no longer fail.

gsmet:
@gbadner 

As for the test, I would be for removing `where="id1_ is not null"` entirely if it's not useful to these tests (and AFAICS, it's not as all the tests pass if I remove it).

Better add specific tests for this feature as you did, and more when we'll be able to fix HHH-12016.

As for the option, it's probably a good idea to be cautious considering we have HHH-12016 in the way. So +1.

gbadner:
I've removed `where="id1_ is not null"` from `Top` used by `MultiTableTest`.

Merged.

PR #2480, opened 2018-08-05
HHH-12885 - Selecting inverse OneToOne with additional selects in query wrongly produces null

beikov:
Test and fix for https://hibernate.atlassian.net/browse/HHH-12885

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/hql/QueryLoader.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/hql/QueryLoader.java
@@ -32,7 +32,6 @@ import org.hibernate.hql.internal.ast.tree.AggregatedSelectExpression;
 import org.hibernate.hql.internal.ast.tree.FromElement;
 import org.hibernate.hql.internal.ast.tree.QueryNode;
 import org.hibernate.hql.internal.ast.tree.SelectClause;
-import org.hibernate.hql.spi.NamedParameterInformation;
 import org.hibernate.hql.spi.ParameterInformation;
 import org.hibernate.internal.IteratorImpl;
 import org.hibernate.internal.util.collections.ArrayHelper;
@@ -182,7 +181,7 @@ public class QueryLoader extends BasicLoader {
 			}
 		}
 
-		//NONE, because its the requested lock mode, not the actual! 
+		//NONE, because its the requested lock mode, not the actual!
 		defaultLockModes = ArrayHelper.fillArray( LockMode.NONE, size );
 	}
 
@@ -450,9 +449,24 @@ public class QueryLoader extends BasicLoader {
 		if ( hasScalars ) {
 			String[][] scalarColumns = scalarColumnNames;
 			int queryCols = queryReturnTypes.length;
+			int managedTypeIndex = 0;
+			SessionFactoryImplementor factory = session.getFactory();
 			resultRow = new Object[queryCols];
 			for ( int i = 0; i < queryCols; i++ ) {
-				resultRow[i] = queryReturnTypes[i].nullSafeGet( rs, scalarColumns[i], session, null );
+				// Types for inverse relationships have a zero column span in which case we have to get the
+				// result from the row array.
+				Type queryReturnType = queryReturnTypes[i];
+				if ( queryReturnType.getColumnSpan( factory ) == 0 ) {
+					resultRow[i] = row[managedTypeIndex];
+				}
+				else {
+					resultRow[i] = queryReturnType.nullSafeGet( rs, scalarColumns[i], session, null );
+				}
+				// The row array contains entity objects, we have to count up the index to access inverse relationship
+				// entities from the row array
+				if ( queryReturnType.isEntityType() ) {
+					managedTypeIndex++;
+				}
 			}
 		}
 		else {
--- a/hibernate-core/src/test/java/org/hibernate/test/onetoone/bidirectional/BiDirectionalOneToOneFetchTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/onetoone/bidirectional/BiDirectionalOneToOneFetchTest.java
@@ -30,6 +30,7 @@ import org.junit.After;
 import org.junit.Test;
 
 import javax.persistence.*;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.junit.Assert.assertEquals;
@@ -62,14 +63,14 @@ public class BiDirectionalOneToOneFetchTest extends BaseCoreFunctionalTestCase {
 	public void testEagerFetchBidirectionalOneToOneWithDirectFetching() {
 		inTransaction( session -> {
 			EntityA a = new EntityA( 1L, new EntityB( 2L ) );
-			
+
 			session.persist( a );
 			session.flush();
 			session.clear();
 
 			// Use atomic integer because we need something mutable
 			final AtomicInteger queryExecutionCount = new AtomicInteger();
-			
+
 			session.getEventListenerManager().addListener( new StatisticalLoggingSessionEventListener() {
 				@Override
 				public void jdbcExecuteStatementStart() {
@@ -77,9 +78,9 @@ public class BiDirectionalOneToOneFetchTest extends BaseCoreFunctionalTestCase {
 					queryExecutionCount.getAndIncrement();
 				}
 			} );
-			
+
 			session.find( EntityA.class, 1L );
-			
+
 			assertEquals(
 					"Join fetching inverse one-to-one didn't use the object already present in the result set!",
 					1,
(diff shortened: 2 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/loader/hql/QueryLoader.java
```diff
@@ -450,9 +450,16 @@ protected Object getResultColumnOrRow(
 		if ( hasScalars ) {
 			String[][] scalarColumns = scalarColumnNames;
 			int queryCols = queryReturnTypes.length;
+			int managedTypeIndex = 0;
 			resultRow = new Object[queryCols];
 			for ( int i = 0; i < queryCols; i++ ) {
-				resultRow[i] = queryReturnTypes[i].nullSafeGet( rs, scalarColumns[i], session, null );
+				// The scalar column names can only be empty if this is a managed type which was already materialized by it's persister
```
gsmet: Minor nitpicking: `s/it's/its/`.

Maybe I would have gone for `When the column is a managed type, the scalar column names array is empty and we have to get the result from the rows array.`.

beikov:
Yeah and IMO this is critical enough to be backported. I guess not everyone uses inverse one to ones in projections 😆 
In fact, I just used it because I was doing a synthetic test and it just so happened, that this kind of query was produced result in a strange test failure.

I amended the commit to use your wording. You can go ahead and merge it when you are ok with it.

gsmet:
@beikov you're insisting about it concerning inverse one to ones but don't you have the issue in any case if you mix scalar and entities for any type of relations?

Considering the fix, I would expect the issue to exist with any type of associations. And even with no associations (e.g. wouldn't something like `select sum(a.myColumn), a from EntityA a`).

beikov:
AFAICT this only concerns inverse one-to-ones as these are special. Object of that type are constructed by their owners id rather than a dedicated key contained in the result set.

gsmet:
@beikov I rebased and force pushed.

I told you I found it a bit too dark magic for my taste and I broke your fix with a subsequent commit.

The issue is that you cannot consider that `row` will only contain values for your very specific case. It contains values for all the elements in `SelectClause.fromElementsForLoad`.

I think one idea would be to track more info in the `SelectClause` as we need to know which elements will be part of `fromElementsForLoad` (note that it includes the elements from the select clause but also potentially the ones of the fetch clauses). Either using a specific array or wrap the query return type in an object indicating if the element is a scalar or will be loaded.

beikov:
Hmm, I'll have a look in the next days, but it looks like the `includeInSelect` array might have that information.

gsmet:
What I'm wondering is if we are not doing the work twice in the case of the first item of my updated test: it is present in `row` and we end up extracting it again from the result set.

I think that, if possible, we should exploit the elements coming from `row`.

beikov:
@gsmet I think the fix is doing the right thing now.

hibernate-core/src/main/java/org/hibernate/loader/hql/QueryLoader.java
```diff
@@ -450,9 +449,24 @@ protected Object getResultColumnOrRow(
...
 			int queryCols = queryReturnTypes.length;
+			int managedTypeIndex = 0;
+			SessionFactoryImplementor factory = session.getFactory();
 			resultRow = new Object[queryCols];
 			for ( int i = 0; i < queryCols; i++ ) {
-				resultRow[i] = queryReturnTypes[i].nullSafeGet( rs, scalarColumns[i], session, null );
+				// Types for inverse relationships have a zero column span in which case we have to get the
+				// result from the row array.
+				Type queryReturnType = queryReturnTypes[i];
+				if ( queryReturnType.getColumnSpan( factory ) == 0 ) {
```
JWGmeligMeyling: Sure that you don't have to account for `includeInSelect` here? Perhaps because this is a scalar query? What happens when you ie. add a `FETCH JOIN` to the query?

hibernate-core/src/main/java/org/hibernate/loader/hql/QueryLoader.java
```diff
@@ -450,9 +449,24 @@ protected Object getResultColumnOrRow(
...
+					resultRow[i] = row[managedTypeIndex];
+				}
+				else {
+					resultRow[i] = queryReturnType.nullSafeGet( rs, scalarColumns[i], session, null );
+				}
+				// The row array contains entity objects, we have to count up the index to access inverse relationship
+				// entities from the row array
+				if ( queryReturnType.isEntityType() ) {
```
JWGmeligMeyling: What about other managed types? I.e. selecting an embeddable in a  scalar query? Is that even possible?

hibernate-core/src/test/java/org/hibernate/test/onetoone/bidirectional/BiDirectionalOneToOneFetchTest.java
```diff
@@ -153,12 +154,40 @@ public void jdbcExecuteStatementStart() {
...
+			session.flush();
+			session.clear();
+
+			List<EntityA> list = session.createQuery(
+					"select a from EntityB b left join b.a a",
+					EntityA.class
+			).list();
+
+			List<Object[]> tupleList = session.createQuery(
+					"select b, b.id, a from EntityB b left join b.a a",
```
JWGmeligMeyling: Perhaps we should add 1 or 2 more cases with reordering the scalars a bit and gather some more confidence regarding the assumptions on the ordering of the arrays. What about `select b, b.id, a, a.id from EntityB b left join b.a a`?

JWGmeligMeyling:
Looks good, but initially I had the same worries as Guillaume. Left a couple of comments that might need some explanation.

gsmet:
About HHH-12842, I think you said Gail had a fix for it as part of other fixes. The patch will eventually get into 5.3. It's on the pending list for inclusion once Gail can finish working on the patch she initiated. Unfortunately, I doubt it will make it for WF 14, considering the feature freeze in on Wednesday.

> Also, I'd like to hear @beikov 's opinion regarding your findings as well. He seemed pretty adamant that his fix was sufficient.

For now, I'm -1 on including the patch as is. It's too fragile as relying on an unrelated source of information to deal with the elements in `row`.

gbadner:
@beikov, I just created a PR for HHH-11903: https://github.com/hibernate/hibernate-orm/pull/2600.

I believe HHH-11903 and HHH-12885 may be related. HHH-11903 only deals with a bidirectional association, where the non-mappedBy side is a derived ID. It may be possible to create a more general solution that fixes both HHH-11903 and HHH-12885.

JWGmeligMeyling:
@gbadner the issue seems unrelated to me. The problem is that this side of the one to one association is mapped by a `OneToOneType`, which has a fixed column span of [`0`](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java#L115) and [resolves the entity based on the id of the owner](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java#L199). However, in `QueryLoader`, the owner is unknown and passed as `null`. We have to work around the [`nullSafeGet`](https://github.com/hibernate/hibernate-orm/pull/2480/files#diff-187a9df4ac00fdb3f0499a25295f286aR463) invocation on the `OneToOneType`.

After looking at the code in `SelectClause`, I see where you are trying to go @gsmet . I am however a bit cautious about the suggested change. The approach doesn't sound very intuitive to me and seems to suffer from the same issue: a poor communication contract between `SelectClause` and `QueryLoader` for scalar queries with entity projections. Getting this right may require more work than is appropriate under the scope of this issue.

Regarding your two concerns about the core assumptions from @beikov : 

1. The only entity selections that are omitted from the `fromElementsForLoad` seem to be fetch joins. I already raised a [question](https://github.com/hibernate/hibernate-orm/pull/2480#discussion_r208711895) about how this behaves with questions with fetch joins. Based on that question, @beikov added an [additional test]() with a `FETCH JOIN` in https://github.com/hibernate/hibernate-orm/pull/2480/commits/17e587a3916b5ed352aa12d9da5fc3d7978b0f2d#diff-069946aa0f0181fcdefc3613a3a830a3R226 . Not sure if this covers all the cases yet. 
2. It seems `SelectClause#isReturnableEntity()` returns `true` if the `FromElement` has an `EntityPersister`, so I think the `isEntityType()` should be safe. 

The approach that @beikov suggests seem to work on the supplied test cases, and the implementation seems to be valid based on the *current* implementation of `SelectClause`. His test only affects `Types` with a column span of `0`, which as far as I am aware of can only be inverse one to one relations. He's test leaves the normal execution path unchanged. The fix only affects the case that would be broken without the change (invalidly returning `null`). Therefore I think it should be safe to integrate this as is.

gsmet:
@JWGmeligMeyling your two points exactly summarizes my issue with this patch as it is (and tbh this remark is not limited to this work but concerns a lot of the code base):

1. `seem to be` `Not sure if this covers all the cases yet.`
2. `It seems` `I think` `should be`

I don't say Christian's reasoning is wrong or the patch in itself is wrong but it relies on assumptions, assumptions that could change with the life of the code.

That's why I wanted us to tie closely the creation of the information to the fact that we get it from there. I'm not sure I'll have the time to write some code before 5.4.0.CR1 though so if someone wants to try my suggestion and see how it goes, I don't think it's that much work.

JWGmeligMeyling:
Hi @gsmet , I understand your concern, and opprove your reasoning. Obviously I prefer a clean, proper solution too. My point was that even if the assumptions change, it would only break the case that is currently already broken (that is, as long as there are no other types besides inverse true one to ones that ever return a column span of 0). 

I had started to implement your suggested approach prior to my previous commit, but I didnt get it to finish it yet.

gsmet:
@JWGmeligMeyling is it a large change? I didn't think it would be to be honest so I'm wondering if we're on the same page or if I missed something?

Maybe you could share your progress?

Thanks!

beikov:
Is there anything I can do to make progress on this PR @gsmet ?

EugenMayer:
We just ran into this very same bug and thanks to @beikov we were lucky to identify the cause and work arround using Blaze Persistence EntityViews. Still i would love to see this one getting fixed - offering testing capacities :)

PR #2481, opened 2018-08-08
HHH-10782 HHH-12896 Query plan cache minor improvements

gsmet:
* https://hibernate.atlassian.net/browse/HHH-10782
 * https://hibernate.atlassian.net/browse/HHH-12896

@Sanne it contains the comment about how the query plan cache is cleaned up.


Sanne:
Looks great! Thanks.

I'm not merging it right now as I am not sure how to deal with the JIRAs ?

I still suspect that we eventually will need to reject HHH-10782. In that case we shouldn't have commits relating to that issue being merged.

Could we merge both of these as a different jira?

Sanne:
@gsmet I decided to not wait for HHH-12896 . That's merged.

Up to you how you want to deal with the other one. +1 to commit.

+1 to apply HHH-12896 on 5.3x as well if you agree.. seems very sensible to me.

gsmet:
Merged.

PR #2482, opened 2018-08-08
HHH-12895 - Test case that shows *ToOne join tables are always joined

beikov:
https://hibernate.atlassian.net/browse/HHH-12895


dreab8:
merged upstream

dreab8:
Thanks @beikov for the test

PR #2484, opened 2018-08-10
HHH-12904 Removing some dead code in InFlightMetadataCollectorImpl an…

Sanne:
…d MetadataImpl

Should be trivial, but sending as PR to see if I'm not missing something.

 - https://hibernate.atlassian.net/browse/HHH-12904


gsmet:
@Sanne I can see how the elements are not used. Can't really be sure though is the right fix is to remove all this dead code :).

Sanne:
ok, in case of need code can always be restored. merged!

PR #2491, opened 2018-08-14
Fix Javadocs comments for Query.getHibernateMaxResults() method

sgilda:
The generated Javadocs for Query.getHibernateMaxResults [1] suggests a setter method to replace a getter method. 

[1] http://docs.jboss.org/hibernate/orm/5.3/javadocs/org/hibernate/Query.html#getHibernateMaxResults--


gsmet:
Good catch, thanks!

Merged to master and 5.3.

PR #2497, opened 2018-08-22
HHH-12921 HHH-12922 Hibernate Connection Pool Thread Improvements

caristu:
Improvements for Hibernate Connection Pool Validation Thread:

- Set the thread as a daemon thread
- Give it a name

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
@@ -455,7 +455,7 @@ public class DriverManagerConnectionProviderImpl
 		}
 	}
 
-	private class ValidationThreadFactory implements ThreadFactory {
+	private static class ValidationThreadFactory implements ThreadFactory {
 		public Thread newThread(Runnable r) {
 			Thread thread = new Thread(r);
 			thread.setDaemon(true);
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
@@ -14,6 +14,7 @@ import java.util.Properties;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
```diff
@@ -453,4 +454,13 @@ public void closeConnection(Connection conn) throws SQLException {
 		}
 	}
 
+	private class ValidationThreadFactory implements ThreadFactory {
```
gsmet: I think you are missing some imports (at least for `ThreadFactory`) here.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
```diff
@@ -384,7 +384,8 @@ private void startIfNeeded() {
 				if ( active ) {
 					return;
 				}
-				executorService = Executors.newSingleThreadScheduledExecutor();
+				ValidationThreadFactory vtf = new ValidationThreadFactory();
+				executorService = Executors.newSingleThreadScheduledExecutor(vtf);
```
gsmet: A reminder for me: we need to reformat the new code to comply with the Hibernate formatting rules.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
vladmihalcea: @gsmet I've added those when reviewing it and want to add a test for it as well. I'll push a PR to replicate this one when I'm done.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
gsmet: @vladmihalcea can you also reformat the added code properly too if you're working on this PR? Thanks!

hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
gsmet: FWIW, I'm not sure we'll be able to implement a test for that. If we don't, it's not a biggie IMHO.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
vladmihalcea: @gsmet natürlich.

gsmet:
@caristu good catch! Nobody should use this in production but I agree it's better to get it right.

caristu:
Hello @gsmet, @vladmihalcea  
Thanks for taking a look in so short time...I've just pushed two changesets I missed when preparing the PR...sorry for that

vladmihalcea:
@caristu @gsmet I created [this PR](https://github.com/hibernate/hibernate-orm/pull/2498) after reviewing this one. Let me know what you think.

gsmet:
Closed in favor of https://github.com/hibernate/hibernate-orm/pull/2498 .

PR #2498, opened 2018-08-22
HHH-12921 - Hibernate Connection Pool Validation Thread should be defined as a daemon Thread

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12921
https://hibernate.atlassian.net/browse/HHH-12922

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
@@ -14,6 +14,7 @@ import java.util.Properties;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReadWriteLock;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
```diff
@@ -453,4 +449,17 @@ public void closeConnection(Connection conn) throws SQLException {
 		}
 	}
 
+	private static class ValidationThreadFactory implements ThreadFactory {
+
+		public static final ValidationThreadFactory INSTANCE = new ValidationThreadFactory();
```
gsmet: Is it really a good idea? I mean, most of the time, we will use it only once and we'd better get it garbage collected rather than kept as a static thing forever?

caristu: If you finally decide to keep the INSTANCE variable, taking into account the scope of the class it can be enough to define it as private, at least for the moment

vladmihalcea: It's better to remove it. The only time the ConnectionProvider will be stopped is when the SF is closed, so this should be a one-time instance creation.

caristu:
Thanks for the quick fix!

PR #2500, opened 2018-08-22
HHH-12920 Fix a debug message causing an exception at debug level

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12920

@gbadner even, if it could be considered minor, I think we should backport that one to 5.3 as it could cause exceptions when someone is trying to debug an issue with the cache. WDYT?


gsmet:
Merged.

PR #2501, opened 2018-08-23
Fix for HHH-12925 and HHH-12926

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12925
https://hibernate.atlassian.net/browse/HHH-12926

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java
@@ -41,6 +41,7 @@ import org.hibernate.hql.spi.id.local.LocalTemporaryTableBulkIdStrategy;
 import org.hibernate.internal.util.JdbcExceptionHelper;
 import org.hibernate.procedure.internal.PostgresCallableStatementSupport;
 import org.hibernate.procedure.spi.CallableStatementSupport;
+import org.hibernate.type.PostgresUUIDType;
 import org.hibernate.type.StandardBasicTypes;
 import org.hibernate.type.descriptor.sql.BlobTypeDescriptor;
 import org.hibernate.type.descriptor.sql.ClobTypeDescriptor;
@@ -171,6 +172,8 @@ public class PostgreSQL81Dialect extends Dialect {
 		registerFunction( "locate", new PositionSubstringFunction() );
 
 		registerFunction( "str", new SQLFunctionTemplate(StandardBasicTypes.STRING, "cast(?1 as varchar)") );
+
+		registerHibernateType( Types.OTHER, PostgresUUIDType.INSTANCE.getName() );
 
 		getDefaultProperties().setProperty( Environment.STATEMENT_BATCH_SIZE, DEFAULT_BATCH_SIZE );
 		getDefaultProperties().setProperty( Environment.NON_CONTEXTUAL_LOB_CREATION, "true" );
--- a/hibernate-core/src/main/java/org/hibernate/type/spi/TypeConfiguration.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/spi/TypeConfiguration.java
@@ -8,7 +8,10 @@ package org.hibernate.type.spi;
 
 import java.io.InvalidObjectException;
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -17,6 +20,7 @@ import org.hibernate.Incubating;
 import org.hibernate.SessionFactory;
 import org.hibernate.SessionFactoryObserver;
 import org.hibernate.boot.cfgxml.spi.CfgXmlAccessService;
+import org.hibernate.boot.spi.BasicTypeRegistration;
 import org.hibernate.boot.spi.MetadataBuildingContext;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.id.uuid.LocalObjectUuidHelper;
@@ -25,6 +29,7 @@ import org.hibernate.internal.SessionFactoryRegistry;
 import org.hibernate.metamodel.internal.MetamodelImpl;
 import org.hibernate.metamodel.spi.MetamodelImplementor;
 import org.hibernate.service.ServiceRegistry;
+import org.hibernate.type.BasicType;
 import org.hibernate.type.BasicTypeRegistry;
 import org.hibernate.type.Type;
 import org.hibernate.type.TypeFactory;
@@ -69,6 +74,7 @@ public class TypeConfiguration implements SessionFactoryObserver, Serializable {
 
 	private final transient Map<String,String> importMap = new ConcurrentHashMap<>();
 
+	private final transient Map<Integer, String> jdbcToHibernateTypeContributionMap = new HashMap<>();
 
 	// temporarily needed to support deprecations
 	private final transient TypeResolver typeResolver;
@@ -118,6 +124,10 @@ public class TypeConfiguration implements SessionFactoryObserver, Serializable {
 
 	public Map<String, String> getImportMap() {
 		return Collections.unmodifiableMap( importMap );
+	}
+
+	public Map<Integer, String> getJdbcToHibernateTypeContributionMap() {
+		return jdbcToHibernateTypeContributionMap;
 	}
 
 	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -206,6 +216,29 @@ public class TypeConfiguration implements SessionFactoryObserver, Serializable {
 		// todo (6.0) : finish this
 		//		release Database, descriptor Maps, etc... things that are only
 		// 		valid while the TypeConfiguration is scoped to SessionFactory
+	}
+
+	public void addBasicTypeRegistrationContributions(List<BasicTypeRegistration> contributions) {
+		for ( BasicTypeRegistration basicTypeRegistration : contributions ) {
+			BasicType basicType = basicTypeRegistration.getBasicType();
+
+			basicTypeRegistry.register(
+					basicType,
+					basicTypeRegistration.getRegistrationKeys()
+			);
+
+			try {
+				int[] jdbcTypes = basicType.sqlTypes( null );
+
+				if ( jdbcTypes.length == 1 ) {
+					jdbcToHibernateTypeContributionMap.put( jdbcTypes[0], basicType.getName() );
+				}
+			}
+			catch (Exception e) {
+				log.errorf( e, "Cannot register [%s] Hibernate Type contribution", basicType.getName() );
+			}
+
+		}
 	}
 
 	/**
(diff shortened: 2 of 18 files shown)


vladmihalcea:
@dreab8 DO you have time to review this PR? Thanks

hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java
```diff
@@ -171,6 +172,8 @@ public PostgreSQL81Dialect() {
 
 		registerFunction( "str", new SQLFunctionTemplate(StandardBasicTypes.STRING, "cast(?1 as varchar)") );
 
+		registerHibernateType( Types.OTHER, PostgresUUIDType.INSTANCE.getName() );
```
sebersole: Historically we have stayed away from this because lots of db datatypes get mapped to OTHER.  Is there a specific reason you changed this?

hibernate-core/src/main/java/org/hibernate/type/spi/TypeConfiguration.java
```diff
@@ -209,6 +215,27 @@ public void sessionFactoryClosed(SessionFactory factory) {
 		// 		valid while the TypeConfiguration is scoped to SessionFactory
 	}
 
+	public String matchTypeContribution(Mapping mapping, int columnType) {
```
sebersole: Anytime you are calling this, the TypeConfiguration has access to the `Mapping` it needs - passing it is redundant and adds a requirement on a class we know it going away (`Mapping`).

hibernate-core/src/main/java/org/hibernate/type/spi/TypeConfiguration.java
```diff
@@ -209,6 +215,27 @@ public void sessionFactoryClosed(SessionFactory factory) {
...
+		return basicTypeRegistrationContributions
+				.stream()
+				.filter( type -> type.getBasicType().getColumnSpan( mapping ) == 1 && type.getBasicType().sqlTypes( mapping ).length == 1 &&
+						type.getBasicType().sqlTypes( mapping )[0] == columnType )
+				.map( type -> type.getBasicType().getName() )
+				.findAny()
```
sebersole: Why are we caching BasicTypeRegistration when we only ever use its BasicType?  Also, we have access to all of this information already via the TypeConfiguration's BasicTypeRegistry.

hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java
vladmihalcea: I did it because we already register the column in the PostgreSQL Dialect:

    registerColumnType( Types.OTHER, "uuid" );

So, for consistency sake, I thought we might want to do the same for the Hibernate Type.

vladmihalcea:
@sebersole I applied your suggestions, so I removed the `matchTypeContribution` and, besides registering the `BasicType` as we used to do before, we now also push the associated Hibernate Type to the `Dialect`. This is only done for types that are contributed explicitly so the previous default behavior should not be affected at all.

Let me know what you think.

hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java
JWGmeligMeyling: This will result in [`Dialect#getTypeName(int code)`](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java#L326-L332) always returning "uuid" for all (custom) types mapped to `Types.OTHER` too right? I remember seeing some issues with this in a Union Subclass hierarchy with custom types, where the `null` values from missing properties declared in other subtypes are casted to their respective type.

vladmihalcea: The UUID column type was mapped to Types.OTHER in Postgresql Dialect. This PR allows you to override that. Check out the associated unit tests.

gsmet:
@vladmihalcea I'm not sure of the status of this one?

vladmihalcea:
For more context about the state of this issue, I'll add this HipChat conversation with @sebersole:

```
[3:38 PM] Steve Ebersole: @VladMihalcea I am here if you wanted to discuss that PR real time
[3:41 PM] Vlad Mihalcea: @SteveEbersole Hi. Sure thing
[3:41 PM] Steve Ebersole: a threw a lot of info at you :)
[3:42 PM] Vlad Mihalcea: That's good
[3:42 PM] Steve Ebersole: i threw
[3:42 PM] Vlad Mihalcea: I hope I got it right
[3:42 PM] Steve Ebersole: i just added some comments
[3:42 PM] Vlad Mihalcea: 1s to check it out
[3:42 PM] Steve Ebersole: the OTHER stuff is just wrong imo
[3:42 PM] Steve Ebersole: yep
[3:44 PM] Steve Ebersole: more-or-less, I'd go back to mostly what you had initially - with just a few changes.  but we can discuss after you take a look
[3:44 PM] Vlad Mihalcea: I read it
[3:45 PM] Vlad Mihalcea: I didn't know about Dialect being used by multiple SF instances
[3:46 PM] Vlad Mihalcea: I could go back to the previous solution
[3:46 PM] Vlad Mihalcea: I thought that this one is simpler
[3:52 PM] Vlad Mihalcea: Actually, I could go back to the previous version while also incorporating a Map<Integer, BasicType> so that we, at least, cache the contributed BasicTypes
[3:52 PM] Vlad Mihalcea: this will improve the BasciType resolving
[3:57 PM] Steve Ebersole: Well again, the overall problem with that is its really <Integer,Set<BasicTyppe>>
[3:57 PM] Steve Ebersole: think OTHER, ARRAY, etc
[3:57 PM] Steve Ebersole: ARRAY is not just a single type on most dbs
[3:58 PM] Steve Ebersole: you have specific handling for String[], int[], etc
[3:58 PM] Steve Ebersole: yes,  creating them is ok because we ask the driver to do it based on the element type.  but think about a cast e.g.
[3:59 PM] Steve Ebersole: also, one other thing I noticed...
[4:00 PM] Steve Ebersole: your solution in your StringArrayType is ultimately to express the "element SQL type" as the database type name
[4:00 PM] Steve Ebersole: that's not ideal
[4:00 PM] Steve Ebersole: really it ought to be the "element JDBC type code"
[4:01 PM] Steve Ebersole: what if they have expressed that they want all strings to be handled as NVARCHAR instead of VARCHAR e.g.?
[4:02 PM] Steve Ebersole: thats easier to handle via the type code and some form of re-mapping instead of trying to handle that based on (potentially proprietary) database type names
[4:02 PM] Vlad Mihalcea: The way I understood it works
[4:02 PM] Vlad Mihalcea: we have all these parameter names: columnType, sqlType, jdbcType
[4:03 PM] Vlad Mihalcea: isn't it that all of them refer to the JDBC Types Integer?
[4:03 PM] Vlad Mihalcea: like Types. OTHER
[4:03 PM] Vlad Mihalcea: The mapping was done for this Integer to a Hibernate Type, which indeed is not ideal
[4:04 PM] Vlad Mihalcea: we need more info, but I don't know if we have it for native SQL queries at this point
[4:05 PM] Vlad Mihalcea: Related to your question regarding ("what if they have expressed that they want all strings to be handled as NVARCHAR instead of VARCHAR")
[4:05 PM] Vlad Mihalcea: The default behavior is the same as it was before
[4:06 PM] Vlad Mihalcea: Now, when a user provides a Type contribution, it's their responsibility to make sure that the new Hibernate Type is able to handle the underlying JDBC Type
[4:06 PM] Vlad Mihalcea: Anyway I implement it at this stage, it will be a half-solution
[4:06 PM] Vlad Mihalcea: it will patch some current limitations, but there's still room left for improvement
[4:07 PM] Vlad Mihalcea: Maybe we can better handle it in 6.0 if we could pass the expected Java Type somehow so that we can better do that matching between JDBC Type and the Hibernate Type
[4:10 PM] GitHub: vladmihalcea synchronize pull request #2501 on hibernate/hibernate-orm : Fix for HHH-12925 and HHH-12926
https://hibernate.atlassian.net/browse/HHH-12925 https://hibernate.atlassian.net/browse/HHH-12926...
[4:10 PM] Vlad Mihalcea: @SteveEbersole I reverted to the non-Dialect solution used previously
[4:11 PM] Vlad Mihalcea: If you think it's risky, we can leave the PR as-is and wait for 6.0 to see if we can integrate it properly there
[4:13 PM] Steve Ebersole: "we have all these parameter names: columnType, sqlType, jdbcType
isn't it that all of them refer to the JDBC Types Integer?"
[4:13 PM] Steve Ebersole: yep
[4:13 PM] Steve Ebersole: if you mean the integer type code as opposed to a name
[4:13 PM] Steve Ebersole: "we need more info, but I don't know if we have it for native SQL queries at this point"
[4:13 PM] Steve Ebersole: for the auto-discovery stuff we do not for these
[4:14 PM] Steve Ebersole: the metadata will tell us OTHER (or ARRAY...) and we have no idea really what to do with that
[4:15 PM] Steve Ebersole: think of the scenario I brought up - an app that uses PGSQL and UUID and one-or-more pgqsl enums
[4:15 PM] Steve Ebersole: those all are "OTHER"
[4:15 PM] Steve Ebersole: when doing auto-discovery we have no idea which to use
[4:16 PM] Steve Ebersole: we'd know that is the specific situation though with <Integer,Set<BasicTyppe>> though
[4:16 PM] Steve Ebersole: if the value (Set<Type>) has multiple we could tell them a different error (auto-discovery not allowed...) instead of the "no type mapping"
[4:17 PM] Steve Ebersole: its really a question of where we keep that info and how we build and then access it
[4:17 PM] Steve Ebersole: "Now, when a user provides a Type contribution, it's their responsibility to make sure that the new Hibernate Type is able to handle the underlying JDBC Type"
[4:18 PM] Steve Ebersole: ok, but that does not play nicely with other Hibernate features
[4:18 PM] Steve Ebersole: yes, i understand it does not now today anyway.  but this is something you want to add support for
[4:18 PM] Steve Ebersole: so it should then work properly
[4:18 PM] Steve Ebersole: that (^^) was not clear
[4:20 PM] Steve Ebersole: what I mean is that today we do not support array types at all.  I know this is all test code and what-not, but you are changing Hibernate itself also to allow support for them.  So its not really relevant that this works as before.  it is something you are trying to add support for - so now it is a concern
[4:21 PM] Steve Ebersole: does that make sense?
[4:22 PM] Steve Ebersole: if you want, I can take your original code and show you what I mean
[4:24 PM] Steve Ebersole: the initial thought for 6 was to leverage custom sql type codes
[4:25 PM] Steve Ebersole: so rather than using 2003 for any and all arrays, you'd specify unique codes for a String[], Integer[], etc
[4:25 PM] Steve Ebersole: same for OTHER
[4:26 PM] Steve Ebersole: you'd use a specific type code for UUID, another for a pgsql GENDER enum UDT, another for a STATUS enum UDT, ...
[4:26 PM] Steve Ebersole: I say UDT but I really just mean database type -- not JDBC's UDT notion
[4:27 PM] Steve Ebersole: but JDBC UDT is another possibility to help here.  The problem there is inconsistent support for UDT
[4:29 PM] Steve Ebersole: bbiab
[5:15 PM] Steve Ebersole: @Barreiro @JohnOHara you guys around?
[5:23 PM] Vlad Mihalcea: @SteveEbersole  makes sense
[5:23 PM] Vlad Mihalcea: we should wait for 6.0 then to see how we can better address it
[5:24 PM] Steve Ebersole: we may be able to handle it better in 6, but I think we can handle the majority of scenarios in 5 too
[5:24 PM] Steve Ebersole: up to you
```

vladmihalcea:
Superseded by [this PR](https://github.com/hibernate/hibernate-orm/pull/2639).

PR #2503, opened 2018-08-24
HHH-12927 Test case for false error when update subselects immutable entities

barnabycourt:
This does not include a fix for the problem but adds a test case that demonstrates the failure for HHH-12927.


vladmihalcea:
Thanks. I created [this PR](https://github.com/hibernate/hibernate-orm/pull/2505) which includes your test and a fix as well.

PR #2504, opened 2018-08-27
Fixing security manager issues (5.3)

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12931
 * https://hibernate.atlassian.net/browse/HHH-12932


gbadner:
Depending on requirements for supporting a security manager, we may not want to revert the whole commit. 

`PrivilegedAction` blocks in `ReflectHelper` should definitely be reverted. Maybe also `ClassLoaderServiceImpl.java` methods that use `PrivilegedAction` blocks (e.g., `classForName`, `loadResource`, `workWithClassLoader`). Also, `ConfigHelper#getResourceAsStream`.

The second commit for ByteBuddy looks good to me.

PR #2506, opened 2018-08-28
HHH-12933 - Generate_statistics grows QueryStatistics ConcurrentHashMap indefinitely

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12933

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
@@ -1951,4 +1951,14 @@ public interface AvailableSettings extends org.hibernate.jpa.AvailableSettings {
 	 */
 	String IN_CLAUSE_PARAMETER_PADDING = "hibernate.query.in_clause_parameter_padding";
 
+	/**
+	 * This setting controls the number of {@link org.hibernate.stat.QueryStatistics} entries
+	 * that will be stored by the Hibernate {@link org.hibernate.stat.Statistics} object.
+	 * </p>
+	 * The default value is given by the {@link org.hibernate.stat.Statistics#DEFAULT_QUERY_STATISTICS_MAX_SIZE} constant value.
+	 *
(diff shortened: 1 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -1951,4 +1951,14 @@
 	 */
 	String IN_CLAUSE_PARAMETER_PADDING = "hibernate.query.in_clause_parameter_padding";
 
+	/**
+	 * This setting controls the number of {@link org.hibernate.stat.QueryStatistics} entries
+	 * that will be stored by the Hibernate {@link org.hibernate.stat.Statistics} object.
+	 * </p>
+	 * The default value is given by the {@link #QUERY_PLAN_CACHE_MAX_SIZE}.
```
gsmet: This is incorrect.

hibernate-core/src/main/java/org/hibernate/stat/Statistics.java
```diff
@@ -15,6 +15,8 @@
  */
 public interface Statistics {
 
+	int DEFAULT_QUERY_STATISTICS_MAX_SIZE = 5000;
+
```
gsmet: I think I would have put the default in `AvailableSettings` with the others.

vladmihalcea: The `AvailableSettings` is used only for the global configuration property names. I chose the `Statistics` interface since this default value is only relevant for statistics.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
vladmihalcea: Good catch.

vladmihalcea:
@gsmet Fixed the comment issue and I think we could integrate it.

PR #2508, opened 2018-08-29
HHH-12934 - Exception handling documentation does not apply only to "Session-per-application anti-pattern"

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12934

### DIFF
--- a/documentation/src/main/asciidoc/userguide/chapters/pc/PersistenceContext.adoc
+++ b/documentation/src/main/asciidoc/userguide/chapters/pc/PersistenceContext.adoc
@@ -1,5 +1,5 @@
 [[pc]]
-== Persistence Contexts
+== Persistence Context
 :sourcedir: ../../../../../test/java/org/hibernate/userguide/pc
 :sourcedir-caching: ../../../../../test/java/org/hibernate/userguide/caching
 :extrasdir: extras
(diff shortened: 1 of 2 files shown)


documentation/src/main/asciidoc/userguide/chapters/pc/PersistenceContext.adoc
```diff
@@ -857,4 +857,54 @@ include::{sourcedir}/CascadeOnDeleteTest.java[tags=pc-cascade-on-delete-example]
...
\ No newline at end of file
+====
+
+[[pc-exception-handling]]
+=== Exception handling
+
+If the JPA `EntityManager` or the Hibernate-specific `Session` throws an exception, including any JDBC https://docs.oracle.com/javase/8/docs/api/java/sql/SQLException.html[`SQLException`], you have to immediately rollback the database transaction and close the current `EntityManager` or `Session`.
+
+Certain methods of the JPA `EntityManager` or the Hibernate `Session` will not leave the Persistence Context in a consistent state. As a rule of thumb, no exception thrown by Hibernate can be treated as recoverable. Ensure that the Session will be closed by calling the `close()` method in a finally block.
+
```
gbadner: I think the following text from the 4.2 documentation would be useful to add:

"Rolling back the database transaction does not put your business objects back into the state they were at the start of the transaction. This means that the database state and the business objects will be out of sync. Usually, this is not a problem because exceptions are not recoverable and you will have to start over after rollback anyway."

gbadner: I see that the above text is in "Session-per-application anti-pattern" section. Doesn't this apply in general?

documentation/src/main/asciidoc/userguide/chapters/pc/PersistenceContext.adoc
```diff
@@ -857,4 +857,54 @@ include::{sourcedir}/CascadeOnDeleteTest.java[tags=pc-cascade-on-delete-example]
...
+Hibernate wraps the JDBC `SQLException`, thrown while interacting with the database, in a
+https://docs.jboss.org/hibernate/orm/{majorMinorVersion}/javadocs/org/hibernate/JDBCException.html[`JDBCException`].
+In fact, Hibernate will attempt to convert the exception into a more meaningful subclass of `JDBCException`. The underlying `SQLException` is always available via https://docs.jboss.org/hibernate/orm/{majorMinorVersion}/javadocs/org/hibernate/JDBCException.html#getSQLException--[`JDBCException.getSQLException()`]. Hibernate converts the `SQLException` into an appropriate JDBCException subclass using the
+https://docs.jboss.org/hibernate/orm/{majorMinorVersion}/javadocs/org/hibernate/exception/spi/SQLExceptionConverter.html[`SQLExceptionConverter`]
+attached to the current `SessionFactory`.
+
```
gbadner: `JDBCException` may be wrapped by `PersistenceException` 

Should some of the information from HHH-12666 be added here?

Also, I think it would be good to mention setting `hibernate.native_exception_handling_51_compliance=true` to disable wrapping `HibernateException` with `PersistenceException`. I'm not sure which property should be discussed first though.

gbadner:
I'll also ask Stephen Fikes to review this, since he originally requested this change.

documentation/src/main/asciidoc/userguide/chapters/pc/PersistenceContext.adoc
vladmihalcea: I added that 4.x paragraph as well. The above text is not in the `Session-per-application anti-pattern`, but in the `Exception handling` section.

vladmihalcea:
@gbadner I added your suggestions to the PR.

documentation/src/main/asciidoc/userguide/chapters/pc/PersistenceContext.adoc
```diff
@@ -857,4 +857,63 @@ include::{sourcedir}/CascadeOnDeleteTest.java[tags=pc-cascade-on-delete-example]
...
+https://docs.jboss.org/hibernate/orm/{majorMinorVersion}/javadocs/org/hibernate/HibernateException.html[`HibernateException`] wraps most of the errors that can occur in a Hibernate persistence layer.
+
+Both the `PersistenceException` and the `HibernateException` are runtime exceptions because, in our opinion, we should not force the application developer to catch an unrecoverable exception at a low layer. In most systems, unchecked and fatal exceptions are handled in one of the first frames of the method call stack (i.e., in higher layers) and either an error message is presented to the application user or some other appropriate action is taken. Note that Hibernate might also throw other unchecked exceptions that are not a `HibernateException`. These are not recoverable either, and appropriate action should be taken.
+
+[NOTE]
+====
+Starting with Hibernate 5.1, the Hibernate `Session` extends the JPA `EntityManager`, and for this reason, the Hibernate-specific `HibernateException` is further wrapped in a JPA https://docs.oracle.com/javaee/7/api/javax/persistence/PersistenceException.html[`PersistenceException`].
```
gbadner: The change to have Hibernate `Session` extend the JPA `EntityManager` happened in 5.2 (not 5.1).

gbadner: Also, by default, if `SessionFactory` is built via Hibernate’s native bootstrapping and `HibernateException` was thrown executing a Hibernate-specific method that does not override an `EntityManager` method (e.g., `Session.saveOrUpdate`), the `HibernateException` is not wrapped by `PersistenceException`. 

`hibernate.native_exception_handling_51_compliance=true` only affects methods that override `EntityManager` methods (e.g., `EntityManager#merge`) when `SessionFactory` is built via Hibernate’s native bootstrapping.

I know this is kind of confusing. I'm not sure of the best way to present this information.

documentation/src/main/asciidoc/userguide/chapters/pc/PersistenceContext.adoc
gbadner: Hibernate will wrap `JDBCException` with `PersistenceException` unless `SessionFactory` is built via Hibernate’s native bootstrapping and:
1) HibernateException was thrown executing a Hibernate-specific method that does not override an EntityManager method (e.g., Session.saveOrUpdate),
OR 2) `hibernate.native_exception_handling_51_compliance=true`

vladmihalcea:
@gbadner I addressed your comments. Let me know if it's fine for you now.

PR #2510, opened 2018-08-30
HHH-12937 : Where clause for collections of basic, embeddable and "any" elements is ignored when mapped using hbm.xml

gbadner:
https://hibernate.atlassian.net/browse/HHH-12937

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
@@ -3386,6 +3386,11 @@ public class ModelBinder {
 		}
 
 		protected void bindCollectionElement() {
+			log.debugf(
+					"Binding [%s] element type for a [%s]",
+					getPluralAttributeSource().getElementSource().getNature(),
+					getPluralAttributeSource().getNature()
+			);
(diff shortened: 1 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
```diff
@@ -3603,6 +3616,11 @@ else if ( getPluralAttributeSource().getElementSource() instanceof PluralAttribu
 
 				);
 				getCollectionBinding().setElement( elementBinding );
+				// Collection#setWhere is used to set the "where" clause that applies to the collection table
+				// (which is the join table for a many-to-any association).
+				// This "where" clause comes from the collection mapping; e.g., <set name="..." ... where="..." .../>
+				getCollectionBinding().setWhere( getPluralAttributeSource().getWhere() );
+				getCollectionBinding().setWhere( getPluralAttributeSource().getWhere() );
```
gsmet: This line is duplicated.

gbadner: Ah yes, must have done that when I copied the comment at the last minute. I've removed it.

gbadner:
Merged.

PR #2511, opened 2018-08-30

pbruski:


gsmet:
@pbruski thanks for this PR.

We need a new JIRA issue as we can't reopen one that has already been shipped. Could you create one and reference it?

We also need a test. Updating SqlServerQuoteSchemaTest would be enough IMHO. Not sure how it's possible to check that the catalog is quoted (not sure how to enforce it) but we can already check it's not there.

pbruski:
@gsmet done: https://hibernate.atlassian.net/browse/HHH-12939 https://github.com/hibernate/hibernate-orm/pull/2512

The original test case (SqlServerQuoteSchemaTest) was red before my changes - the test case already covers the fix.

PR #2512, opened 2018-08-31
HHH-12939 - Database name not quoted at schema update

pbruski:


gsmet:
@pbruski @vladmihalcea I suggest a slightly different approach: https://github.com/hibernate/hibernate-orm/pull/2512/commits/3816c26a3a6f52ae5fd93939ae2897985b2d44d3 .

Why use the `TableInformation` elements and quote them if the elements of `Table` are quoted instead of simply using the ones from `Table`?

I think the elements of `Table` are what we want.

@vladmihalcea could you take a look and check that my assumption is right?

If so, I'll extract my commit from this PR and apply it as a fix.

vladmihalcea:
@gsmet The `QualifiedTableName` just qualifies the path (e.g. catalog.schema.table), it does not quote the path automatically. If we apply the same logic you did in [this commit](https://github.com/hibernate/hibernate-orm/commit/3816c26a3a6f52ae5fd93939ae2897985b2d44d3), how would we escape a table that contains an illegal character like `-`, as described in [HHH-12939](https://hibernate.atlassian.net/browse/HHH-12939)?

gsmet:
@vladmihalcea you are using `schema.isQuoted()` to decide if you should escape the `TableInformation` schema with quotes. So I suppose the `schema` `Identifier` is already properly quoted if it needs to be.

Otherwise the existing code wouldn't work.

vladmihalcea:
@gsmet Let's add another test case to validate your findings then. That's the best way to make sure we handle it properly.

pbruski:
@gsmet I thought that the intention was to always include the catalog in the alter table. For that to happen, the data has to come from TableInformation (Table doesn't have to have it) and has to be quoted before use.

gsmet:
I don't think that's the original intention. In any case, it's not very consistent with the rest as the other commands are not qualified with the catalog. My version improves consistency and is far more readable.

I'll see if I can add a test case.

gsmet:
@vladmihalcea done here: https://github.com/hibernate/hibernate-orm/pull/2526 .

gsmet:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2528 .

PR #2514, opened 2018-09-03
HHH-12323 - fix access to Statistics over JMX

icidel:
Access to CollectionStatistics, EntityStatistics and NaturalIdStatistics through JMX is broken because their new parent interface CacheableDataStatistics is not serializable.

### Test case
- import the sample Spring Boot application from https://github.com/icidel/spring-boot-jmx-export-of-hibernate-statistics as an Eclipse project (or your favorite)
- execute the map class "Application"
- execute jconsole from you JDK bin folder and connect to the running application
- open the MBean tab
- access the Hibernate/Statistics MBean "operations" panel
- run the getEntityStatistics opération with "com.example.hibernatestatsdemo.Customer" as entityName argument

This test case is **OK** if the pom.xml file references Hibernate 5.2.x and earlier: entity statistics are returned successfully.

This test case is **KO** if the pom.xml file references Hibernate 5.3.0 to 5.3.5: an unmarshalling exception is raised because the EntityStatistics interface is not fully serializable.

This simple patch correct this bug.


icidel:
I removed the logger because it adds an external dependency to a Serializable class.

As a consequence, accessing this class remotely implies adding to the JMX client's classpath (any) hibernate-core-5.3.x.jar (which is ok) and also org.jboss.logging.jar and org.jboss.commons.jar... but then the Logger instance initialization fails because the logger will also be initialized on the RMI client side.

Of course we can as an alternative make the Logger "Log" instance transient in order to keep Logging on the RMI server side, which will make this instance null on the client side. But the "org.jboss.logger" dependency will have to remain on the client's classpath.

icidel:
I also set some magical "serialVerisonUID" number in order to allow RMI access even for Hibernate versions prior to my patch: I took the number generated by the remote invocation classVersionError exception displayed by JConsole to do this trick.

This means that any "hibernate-core_<version>.jar" built with those fixes allow JConsole to connect to any JMX server backed by any version of Hibernate since 5.3.0.CR2 !

gsmet:
I'm a bit torn on this one:
- the `Serializable` addition should obviously go in;
- I wouldn't add the `serialVersionUID` as we decided we didn't want them. My approach would be: we had a bug, statistics couldn't be read with older versions, please upgrade;
- about the logger thing, I can see how it's appealing to remove additional dependencies to access simple beans. I'm just not sure it's an approach that could stick in time and we will **never** need a logger here. Users might use them from time to time too. Question is is this class the best place to have such a logger, I'm not sure. @Sanne any additional thoughts?

Sanne:
yeah I wouldn't include the `serialVersionUID`. I understand the good intention, but since we can't promise we will not break this "funcionality" in future versions, I'd rather make it clear that this is not intended to be stable over time.

Regarding the logger, I'm torn. Apparently there's people like @icidel who like to connect over RMI, but there's also people who just want to log it. @icidel if we keep the logger, you'd only need one extra dependency on classpath, and people would still have the option to use logging if they prefer. I don't think that's too much to ask from you and anyone having a similar need?

Frankly I would just expect that people who want to do such a thing, should be prepared to have:

1. an exact match of Hibernate ORM's jar (same exact version)
2. any dependencies

That's not too hard to generate from your project? Any build tool has some way to dump an output string in a traditional classpath format.

+1 to restore `Serializable` of course.

icidel:
The Log instance of AbstractCacheableDataStatistics is not used, even by subclasses => we can safely remove it.

The Log instance of QueryStatisticsImpl is used but:
- only when the TRACE level is enabled
- log calls have a non-zero overhead, even when the trace level is disabled : if we want to keep those trace log calls, at least we should protect them with a call to Log.istraceEnabled() to reduce this overhead.
- those log calls were added for 5.3.0.CR2 in a commit labelled "HHH-11356 - Adjust the 2nd-Cache SPIs to better reflect supported uses", but the logging addition has not been discussed in the Jira ticket HHH-11356 => it was just added for development testing purposes, were not expected, so shouldn't have been put into ready to production code.
  (direct link to QueryStatisticsImpl diff. : https://github.com/hibernate/hibernate-orm/commit/f432ecea687ce606144c8fca1f4fc7f1937e1c2a#diff-dd11f7b6650d69b1f18abddb2fc7e0da )

That's why I think the logger should be removed from AbstractCacheableDataStatistics and QueryStatisticsImpl.

icidel:
About the serialVersionUID, the java.io.Serializable documentation says:

"However, it is **_strongly recommended_** that all serializable classes explicitly declare serialVersionUID values, since the default serialVersionUID computation is highly sensitive to class details that may vary depending on compiler implementations, and can thus result in unexpected InvalidClassExceptions during deserialization."

"Therefore, to **guarantee** a consistent serialVersionUID value across different java compiler implementations, a serializable class **must declare an explicit serialVersionUID value**"

" It is also **strongly advised** that explicit serialVersionUID declarations use the **private** modifier where possible, since such declarations apply only to the immediately declaring class--serialVersionUID fields are not useful as inherited members."

So setting a **private** static final serialVersionUID is a really good thing after all :-)

sebersole:
@icidel The point made in the javadocs wrt serialVersionUID is specifically in regards to maintaining serial compatibility (can I serialize one of these things using one version of Hibernate and then deserialize it with another version).  We do not support that and so the serialVersionUID is unnecessary. 
 And in fact, as Sanne points out, defining the serialVersionUID actually gives the wrong impression.

gsmet:
@icidel 

Thought about it a bit more:
* let's add the `Serializable` (obviously)
* let's remove the loggers for now... but, if at some point, we need some real loggers here for an error or something, be prepared to add the jboss-logging dependency to your projects
* let's not add the `serialVersionUID`: we have decided a long time ago to not have them and to not maintain them.

If you can remove the `serialVersionUID`, I'll merge it.

Thanks!

hibernate-core/src/main/java/org/hibernate/stat/internal/QueryStatisticsImpl.java
```diff
@@ -24,8 +22,9 @@
  * @author Alex Snaps
  */
 public class QueryStatisticsImpl implements QueryStatistics {
-	private static final Logger log = Logger.getLogger( QueryStatisticsImpl.class );
-
+	// This magic number allow pre-5.3.7 releases RMI compatibility (which includes a JBoss Logger instance)	
+	private static final long serialVersionUID = 8569182022705792772L;
+
```
icidel: To remove : "tricky tradeoff, we generally prefer to not define one as it's almost impossible to promise that it will be back/forward compatible in all JVMs."

hibernate-core/src/main/java/org/hibernate/stat/internal/AbstractCacheableDataStatistics.java
```diff
@@ -12,14 +12,13 @@
...
-import org.jboss.logging.Logger;
-
 /**
  * @author Steve Ebersole
  */
 public abstract class AbstractCacheableDataStatistics implements CacheableDataStatistics {
-	private static final Logger log = Logger.getLogger( AbstractCacheableDataStatistics.class );
-
+	// This magic number allow pre-5.3.7 releases RMI compatibility (which includes a JBoss Logger instance)	
+	private static final long serialVersionUID = 1417304055641145372L;
```
icidel: To remove : "tricky tradeoff, we generally prefer to not define one as it's almost impossible to promise that it will be back/forward compatible in all JVMs."

icidel:
I rebased the last two commits to alter author and commiter names.

All is now OK for me :-)

Sanne:
Thanks @icidel ! And thanks @gsmet for taking care of it while I was away :+1:

PR #2515, opened 2018-09-03
HHH-12944 Add test for multiload & second level cache

barnabycourt:


vladmihalcea:
Thanks for the test case. I incorporated it in [this PR](https://github.com/hibernate/hibernate-orm/pull/2607) with a fix

PR #2516, opened 2018-09-04
Update H2Dialect.java

dbcxy:
H2Dialect，h2 database column key qualifier openQuote and closeQuote " -> · .

I think he is a problem that should be repaired.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
@@ -447,4 +447,14 @@ public class H2Dialect extends Dialect {
 	public String getQueryHintString(String query, String hints) {
 		return IndexQueryHintHandler.INSTANCE.addQueryHints( query, hints );
 	}
+	
+	@Override
+	public char openQuote() {
+		return '`';
+	}
+	
+	@Override
+	public char closeQuote() {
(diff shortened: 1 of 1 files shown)


vladmihalcea:
Please create a Jira issue for this. Thanks.

gsmet:
What is the issue you are trying to solve?

Because AFAICS H2 supports both `"` and backticks and they don't have the same meaning (from what I read names in backticks are not case sensitive). I'm not convinced switching to backticks would be an improvement.

gsmet:
Closing. Feel free to reopen with more explanation later.

dbcxy:
But still felt that he should be treated.
I still think it's a BUG.
1 ------------------------------------------------------------
DROP TABLE IF EXISTS test
create table test
(
  id         BIGINT GENERATED BY DEFAULT AS IDENTITY,
  `select`  varchar(255)  null,
  PRIMARY KEY (id)
);

select `select` from test

result: cloumn name == SELECT
---------------
SELECT



2 ------------------------------------------------------------
DROP TABLE IF EXISTS test
create table test
(
  id         BIGINT GENERATED BY DEFAULT AS IDENTITY,
  "select"  varchar(255)  null,
  PRIMARY KEY (id)
);

select "select" from test

result: cloumn name == "select"
---------------
"select"

PR #2517, opened 2018-09-04
HHH-10201 - ElementCollection not audited when revision_on_collection_change=false.

Naros:
https://hibernate.atlassian.net/browse/HHH-10201


Naros:
Appled upstream.

PR #2518, opened 2018-09-05
HHH-12956 - Add workaround for Native Query parameter type guessing

coladict:
So far this is a proof-of-concept that requires some review, and the names used may not be the final ones.


gsmet:
As mentioned on HipChat, I'm wondering if a better fix wouldn't be to implement a `Query<R> setParameter(String name, LocalDate value, TemporalType temporalType);` in our Hibernate extension `org.hibernate.query.Query`.

Then you would call:
```
((org.hibernate.query.Query<?>) q).setParameter("ed", (LocalDate) null, TemporalType.DATE);
```

Not sure if the logic is correctly implemented after that and if it would solve your issue but I think it would be a better bet.

coladict:
No, because LocalDate is just one easy example of when determining type doesn't work. There are other types as well, for which the problem remains. I didn't think of it earlier, but it also happens with `bigint` (Java Long equivalent).
If you pass a null to a bigint parameter and when the query doesn't know the type you get the same database error.

gavinking:
This work was never finished, and the original issue was closed.

PR #2521, opened 2018-09-06
Fix for criteria query/HQL ast + EntityGraph bug

zoller27osu:
Fixes a bug where adding an EntityGraph (containing an attribute annotated with @OrderBy) hint to a Criteria Query that defined a groupBy but no orderBy caused the ORDER clause to be inserted before the GROUP clause in the tree, therefore causing the SqlGenerator to ignore the GROUP clause when converting the tree to SQL.


gsmet:
@zoller27osu thanks for your contribution.

A few things:
* could you squash all your changes to the tests so that the history is more readable. You can have the fix and the test in separate commits but having 21 commits for the test seems a bit too much;
* we need a JIRA issue explaining the problem you have. You need to create it there: https://hibernate.atlassian.net/secure/Dashboard.jspa and then use the number as a prefix in your commit messages;
* I think you're missing an import in your test as it doesn't compile;
* is there a reason why you only test it with MySQL. Testing it with H2 would be nice as it's our default testing database. I took a look and I think it supports `HAVING`.

Thanks.

zoller27osu:
@gsmet done.

The last I saw, the test compiled fine, but failed due to an error on what is now Line 65. I added the MySQL dialect in an effort to fix it - I've removed that now.

I'm unable to debug the error right now as I'm having difficulty getting my clone to compile.

gsmet:
@zoller27osu I pushed a commit with some fixes to your test (formatting and renaming `Order` as it's a reserved keyword for H2).

That being said your test fails with a now unrelated error. I don't know exactly what you wanted to show and it would be nice if you could take it from there.

It can be hard to get ORM to compile with Eclipse for instance so what I usually do to run a test case is the following:
```
cd hibernate-core
../gradlew test --tests=org.hibernate.test.ast.tree.QueryNodeTest
```

You can use the --debug-jvm option to attach a debugger if you need one.

gsmet:
Closing for now. Please ask for reopening it if you end up getting a valid test case.

Thanks!

PR #2523, opened 2018-09-07
HHH-12961: Update HANA dialects Javadoc

breglerj:
- Fix the links
- add some more information
- fix some compiler warnings

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/AbstractHANADialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/AbstractHANADialect.java
@@ -102,11 +102,18 @@ import org.hibernate.type.descriptor.sql.SqlTypeDescriptor;
 import org.hibernate.type.descriptor.sql.VarcharTypeDescriptor;
 
 /**
- * An abstract base class for HANA dialects. <br/>
- * <a href="http://help.sap.com/hana/html/sqlmain.html">SAP HANA Reference</a> <br/>
- * NOTE: This dialect is currently configured to create foreign keys with <code>on update cascade</code>.
+ * An abstract base class for SAP HANA dialects.
+ * <p>
+ * For more information on interacting with the SAP HANA database, refer to the
+ * <a href="https://help.sap.com/viewer/4fe29514fd584807ac9f2a04f6754767/">SAP HANA SQL and System Views Reference</a>
+ * and the <a href=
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/AbstractHANADialect.java
```diff
@@ -1533,10 +1540,10 @@ public Integer convert(Object value) {
 			this.blobTypeDescriptor = new HANABlobTypeDescriptor( maxLobPrefetchSize );
 		}
 
-		boolean useUnicodeStringTypes = configurationService.getSetting( USE_UNICODE_STRING_TYPES_PARAMETER_NAME, StandardConverters.BOOLEAN,
+		this.useUnicodeStringTypes = configurationService.getSetting( USE_UNICODE_STRING_TYPES_PARAMETER_NAME, StandardConverters.BOOLEAN,
```
gsmet: I think it would be better to define all `useUnicodeStringTypes`, `useLegacyBooleanType` and `maxLobPrefetchSize` in the constructor once and for all.

At the moment, for a short amount of time, the values are incorrect as they are defined to the default values.

The fields should be final and defined once and for all in the constructor. That would be cleaner and less error prone.

breglerj: I'll remove the cleanup from this PR and open a separate one.

Regarding the configuration parameters, I'm not sure how to read the configuration in the constructor. I guess I could use `Environment.getProperties()`. Do you have a better idea? Also, I need a JDBC connection to set `maxLobPrefetchSize`. Do you know how I can obtain one in the constructor?

gsmet: Hum. Yes, I think I'm too accustomed to our other projects having immutable objects everywhere.

Apparently, the other dialects use `Environment.getProperties()` but I can see how it's not very appealing.

As for `maxLobPrefetchSize`, I would have resolved it with the others in the constructor and have a private field for it.

I really don't like the fact that we have incorrect values for a certain amount of time between the construction and the call to this method.

Maybe let's keep it that way for now and we might revisit it in 6. I remember the others talking about significant improvements about how to configure dialects.

Still interested in having the doc changes and the code changes in different commits though :).

Thanks.

breglerj: Sounds good. If there is a good way to get access to the configuration in Hibernate 6 I'd be happy to change it.

I've pushed a second commit that reverts the cleanup. The PR should now effectively only fix the Javadoc.

PR #2524, opened 2018-09-07
HHH-12963: HANA dialect doesn't set internal configuration for use_unicode_string_types correctly

breglerj:
- use member variable instead of local variable
- fix an "unboxing" compiler warning


vladmihalcea:
Applied upstream, thanks

PR #2525, opened 2018-09-07
HHH-12964 Upgrade to dom4j 2.1.1

SirWayne:
Upgrade dom4j to version 2.1.1.

Unfortunately there is no upgrade bundle for osgi.

HHH-12964


vladmihalcea:
Applied upstream, thanks

PR #2526, opened 2018-09-10
HHH-12939 Properly quote identifiers in alter table statements

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12939

This is a follow-up of https://github.com/hibernate/hibernate-orm/pull/2512, which is IMHO the proper fix for the original https://hibernate.atlassian.net/browse/HHH-12106 issue.

Added 2 test cases run with H2.

@vladmihalcea apart from a review, if you have SQL Server handy, I would appreciate if you could check that the modified SQL Server test works as expected.

@pbruski feedback welcome.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/schemaupdate/AlterTableQuoteDefaultSchemaTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/schemaupdate/AlterTableQuoteDefaultSchemaTest.java
@@ -6,79 +6,68 @@
  */
 package org.hibernate.test.schemaupdate;
 
+import static org.hamcrest.core.Is.is;
+import static org.hibernate.testing.transaction.TransactionUtil.doInHibernate;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.util.EnumSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
 import javax.persistence.Entity;
 import javax.persistence.Id;
 import javax.persistence.Table;
 
+import org.hibernate.boot.MetadataBuilder;
 import org.hibernate.boot.MetadataSources;
 import org.hibernate.boot.registry.StandardServiceRegistry;
 import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
 import org.hibernate.boot.spi.MetadataImplementor;
 import org.hibernate.cfg.AvailableSettings;
-import org.hibernate.dialect.SQLServerDialect;
-import org.hibernate.tool.hbm2ddl.SchemaUpdate;
-import org.hibernate.tool.schema.TargetType;
-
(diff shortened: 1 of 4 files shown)


hibernate-core/src/test/java/org/hibernate/test/schemaupdate/AlterTableQuoteDefaultSchemaTest.java
```diff
@@ -0,0 +1,188 @@
...
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.hibernate.tool.hbm2ddl.SchemaUpdate;
+import org.hibernate.tool.schema.TargetType;
+import org.junit.Test;
+
+/**
+ * @author Guillaume Smet
+ */
+@TestForIssue(jiraKey = "HHH-12939")
+@RequiresDialect(H2Dialect.class)
```
vladmihalcea: Better use:

    @RequiresDialectFeature(DialectChecks.SupportSchemaCreation.class)

So we can test it on more DBs, like SQL Server.

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/AlterTableQuoteSpecifiedSchemaTest.java
```diff
@@ -0,0 +1,170 @@
...
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.hibernate.tool.hbm2ddl.SchemaUpdate;
+import org.hibernate.tool.schema.TargetType;
+import org.junit.Test;
+
+/**
+ * @author Guillaume Smet
+ */
+@TestForIssue(jiraKey = "HHH-12939")
+@RequiresDialect(H2Dialect.class)
```
vladmihalcea: Better use:

    @RequiresDialectFeature(DialectChecks.SupportSchemaCreation.class)

So we can test it on more DBs, like SQL Server.

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/AlterTableQuoteDefaultSchemaTest.java
gsmet: That's exactly what I don't want to do: quoting rules are different for each dialect so I just want to validate the behavior works for one dialect supporting quoting.

gsmet:
The failure on SQL Server is expected as it's not quoting strings the same way as H2.

I wasn't clear but the test to validate with SQL Server is this one: `SqlServerQuoteSchemaTest`. I adapted it so I expect it to work but better be sure.

vladmihalcea:
I can send a PR based on yours with the test adapted for all major DBs supporting schemas. I need to test it on PostgreSQL too.

gsmet:
If you want but I don't think it has much value as we just want to validate the general mechanism.

But feel free to do it.

gsmet:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2528.

PR #2520, opened 2018-09-05
Add test for cascade entities removal in context

scadgek:
A test case for issue report at https://discourse.hibernate.org/t/cascade-not-cleaning-persistent-context/1347/2

PR #2527, opened 2018-09-10
Add test for cascade entities removal in context

scadgek:
This is a follow-up to https://github.com/hibernate/hibernate-orm/pull/2520. The patch from @vladmihalcea helped with that case, but then I found that the test data was being created not as expected. Fixed it with commit https://github.com/hibernate/hibernate-orm/commit/c441fa24627732b6a7cc5fe180e3ac3f4a5dd6d4. 

Also, had to uncomment the `@OnDelete` annotations, because otherwise the database constraint (which by default is `ON DELETE RESTRICT`) prevented removing the table by throwing an SQL exception about foreign key constraint failure.

P.S. I didn't really want to create a new MR for the same problem, but the original one did not pull the new commits. Please guide me if I'm missing how to merely update the previous MR.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/cascade/CascadeContextCleanupTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/cascade/CascadeContextCleanupTest.java
@@ -35,23 +35,16 @@ public class CascadeContextCleanupTest extends BaseEntityManagerFunctionalTestCa
   @Test
   public void testCascadeContextCleanup() {
 
+    // create test data: Table1 -> Column -> Relation -> Table2
+    createTablesData();
+
     final EntityManager entityManager = getOrCreateEntityManager();
 
-    // create test data: Table1 -> Column -> Relation -> Table2
-    createTablesData( entityManager );
-
-    // remove table 1
-    // expected to remove Table, Column and Relation
-    // and it does remove them from DB, but they stay in Persistence Context
     entityManager.getTransaction().begin();
     final TableEntity table = entityManager.find( TableEntity.class, "1" );
     assertEquals(1, table.getColumns().size());
     entityManager.remove( table );
-    entityManager.getTransaction().commit();
 
-    // remove table 2
-    // throws org.hibernate.TransientPropertyValueException
-    entityManager.getTransaction().begin();
     final TableEntity table2 = entityManager.find( TableEntity.class, "2" );
     entityManager.remove( table2 );
     entityManager.getTransaction().commit();
@@ -59,8 +52,9 @@ public class CascadeContextCleanupTest extends BaseEntityManagerFunctionalTestCa
     entityManager.close();
   }
 
-  private void createTablesData( EntityManager entityManager ) {
+  private void createTablesData() {
 
+    final EntityManager entityManager = getOrCreateEntityManager();
     entityManager.getTransaction().begin();
 
     final TableEntity parentTable = new TableEntity();
@@ -85,6 +79,7 @@ public class CascadeContextCleanupTest extends BaseEntityManagerFunctionalTestCa
     entityManager.persist( relation );
 
     entityManager.getTransaction().commit();
+    entityManager.close();
   }
 
   @Entity
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/cascade/CascadeContextCleanupTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/cascade/CascadeContextCleanupTest.java
@@ -1,0 +1,211 @@
+package org.hibernate.jpa.test.cascade;
+
+import org.hibernate.annotations.OnDelete;
+import org.hibernate.annotations.OnDeleteAction;
+import org.hibernate.jpa.test.BaseEntityManagerFunctionalTestCase;
+import org.junit.Test;
+
+import javax.persistence.CascadeType;
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.EntityManager;
+import javax.persistence.FetchType;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.MapsId;
+import javax.persistence.OneToMany;
+import javax.persistence.OneToOne;
+import java.util.ArrayList;
+import java.util.Collection;
+
+public class CascadeContextCleanupTest extends BaseEntityManagerFunctionalTestCase {
+
+  @Override
+  protected Class<?>[] getAnnotatedClasses() {
+    return new Class<?>[] {
+            TableEntity.class,
+            ColumnEntity.class,
+            RelationEntity.class
+    };
+  }
+
+  @Test
+  public void testCascadeContextCleanup() {
+
+    final EntityManager entityManager = getOrCreateEntityManager();
+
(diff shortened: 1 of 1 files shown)


scadgek:
I've done some deeper debugging and here's what I found.  

After creating the test data, there are 4 entities in the Persistence Context: Table1, Column, Relation and Table2, each with state `MANAGED`. This is correct by far.  

Then, a new transaction starts to remove Table1. After calling `entityManager.remove(table1)`, the cascade properly goes through the Column and Relation and changes their state to `DELETED`. This is also correct by far.  

But then on the next line, when `.commit()` is executed, a beforeCommit operation tries to flush the existing context before proceeding, and it flushes the Table2 object, since it's state is `MANAGED`. And since Table2 refers to Relation and Relation refers to Column (both relations with cascade=ALL), the Relation and Column entries in the Persistence Context change their state to `MANAGED` again. This prevents them further from being deleted along with Table1.  

And so I'm wondering, is it a correct behavior? I guess the flush-persist operation should not cascade to the entries, which are already marked to be `DELETED`, but I'm not sure about this.

vladmihalcea:
In order to have a bug, we need a valid use case or a use case that's meaningful for a real-life situation. This test case has 2 main issues:

1. The very same `EntityManager` is used for both creating and removing the entities across two transactions that actually commit the changes. When will we ever bump into a situation like this one? If this were a workflow, only the last transaction should be a RW one and do the commit.
2. The presence of both CASCADE and `@OnDelete` indicates a serious mapping issue. You should either use one or the other, but not both because. 

I personally don't think it's worth investigating it further, so in my opinion, we should just close this issue. However, if anyone else considers this might be a valid issue, feel free to create a Jira issue for it.

vladmihalcea:
> In our case, we have something like "Open Session In View" currently, and since the opening and closing of the EntityManager is done in another level of abstraction, unfortunately it cannot be changed in one run, though we do plan to change that iteratively.

OSIV is a really [bad choice](https://vladmihalcea.com/the-open-session-in-view-anti-pattern/). Not only that it might cause this issue, but it can also affect performance and stability of your application.

> What is not clear to me, is does this annotation notify the EntityManager's persistence context that the related entity is removed automatically, so that it clears it from the context, too?

No, it does not. The `ON DELETE <ACTION>` happens in the DB engine and Hibernate cannot be notified about that.

> But I'm not sure it would then work properly in the context of Hibernate. Can you please elaborate on this?

If you want to delete all that work to the DB, then you don't need Hibernate for that. Just issue the bulk `DELETE` on the entities you want to delete and rely on FK deleting then. However, you won't benefit from optimistic locking.

If you want to investigate this Hibernate behavior further and want to provide a Pull Request, I can take a look at it. Otherwise, you are better off dropping OSIV and use one `EntityManager` per service layer method.

scadgek:
> No, it does not. The ON DELETE <ACTION> happens in the DB engine and Hibernate cannot be notified about that.

Then what's the purpose of the `@OnDelete` annotation besides automatic schema generation?

> If you want to delete all that work to the DB, then you don't need Hibernate for that. Just issue the bulk DELETE on the entities you want to delete and rely on FK deleting then. However, you won't benefit from optimistic locking.

So, in this case your advice would be to drop the foreign key constraints and rely on Hibernate's cascade solely? That would be a bit fragile, because Hibernate definitely may be not the only one working with the DB at some moment.

vladmihalcea:
> Then what's the purpose of the @OnDelete annotation besides automatic schema generation?

It's only purpose is schema generation, meaning that you rarely need it. What you need is a tool like Flyway anyway.

> So, in this case your advice would be to drop the foreign key constraints and rely on Hibernate's cascade solely? That would be a bit fragile, because Hibernate definitely may be not the only one working with the DB at some moment.

That's where optimistic locking can help you. Nevertheless, if another DB adds a child record while you try to delete the parent row, the delete transaction will fail anyway. Therefore, the DB will never be left in an inconsistent state no matter what approach you are taking. But in the end, it's up to you to decide which one is best for your use case.

scadgek:
> It's only purpose is schema generation, meaning that you rarely need it. 

Thanks, I get it now. I was falsely hoping that it goes further than this =/

> That's where optimistic locking can help you. Nevertheless, if another DB adds a child record while you try to delete the parent row, the delete transaction will fail anyway. Therefore, the DB will never be left in an inconsistent state no matter what approach you are taking. But in the end, it's up to you to decide which one is best for your use case.

Yes, I understand that Hibernate is smart enough to ensure there's no inconsistency when *it* works with the database. I rather meant that some other program may contain bugs which leave DB in an inconsistent state, or some person might delete a record without cleaning up its relation, and database could make sure these things are done correct by properly executing the foreign key actions. It's just frustrating that Hibernate encourages the non-integral design of the database, though I guess it might be a fundamental problem to some extent which Hibernate is not able to solve.

Okay, so I think we'll solve our project issues by narrowing down the lifecycle of the `EntityManager`. By the way, it also fixed the problem in the current issue.

Thanks for all the assistance! If I find anything really broken or if I see something I can help with by creating a pull request, I will surely do.

vladmihalcea:
> It's just frustrating that Hibernate encourages the non-integral design of the database, though I guess it might be a fundamental problem to some extent which Hibernate is not able to solve.

No, it doesn't. Hibernate encourages an integral design of the DB. Just because you are using ON CASCADE, it does not mean you are using a proper RDBMS design. And if you are not using ON CASCADE, it does not mean your design is bad. USing ON CASCADE is also extremely dangerous. What if you accidentally delete a parent and it cascades to all children and grandchildren?

> If I find anything really broken or if I see something I can help with by creating a pull request, I will surely do.

That's great.

scadgek:
> What if you accidentally delete a parent and it cascades to all children and grandchildren?

If my constraints in the database act so, then this is exactly what I meant to do :) I see your point, but anyway you'll end up with some kind of cascades either on the database level or on the application level, and for me having them only in one place makes the probability of something being broken much less, than when you duplicate the cascading logic in every app and when working with DB manually, too.

vladmihalcea:
You ake it sound like if you don't do it on the FK-level, it could cause a situation where the DB is left in an inconsistent state. The consistency is already provided by the FK. 

The worst thing that can happen if you don't cascade at the FK level is to abort a transaction when you wanted to delete the parent, but another pending transaction has just added a new row that's not loaded in your Persistence Context. While you might think this is a bug, in reality, this is an application-level Phantom Read detection mechanism which can work even on lower isolation levels (e.g. Read Committed, Repeatable Read) which are used by default by the vast majority of DBs.

Therefore, not only that won't affect consistency, but in fact, it might even prevent some data anomalies.

scadgek:
Looks like I misunderstood you - I thought you were talking about getting rid of foreign keys completely. You're right, if I preserve the foreign keys, but with a default `ON DELETE RESTRICT` - it will ensure consistency. But at the same time I wanted to make this more convenient, so that if I delete the parent - no matter via Hibernate or using MySQL Workbench - then all the childs are cleaned up as well (of course assuming I did mean such cleanup, because the entities are logically coupled). 

Otherwise, if I don't have `ON DELETE CASCADE` but the default `ON DELETE RESTRICT`, I will just not be able to remove the parent without removing the childs first. Which I see is still realiable in the context of consistency, but less convenient.

So is that what you meant by not using foreign key cascades?

vladmihalcea:
> So is that what you meant by not using foreign key cascades?

Just read carefully my previous comments and you will surely understand what I meant.

PR #2528, opened 2018-09-10
HHH-12939 Database name not quoted at schema update

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12939


vladmihalcea:
@gsmet There's an extra commit, besides yours, which makes those two tests work on H2, PostgreSQL and SQL Server.

I chose not to activate those for every DB supporting schemas as it will fail on HANA due to lack of privileges when creating the schema.

gsmet:
Not sure I'm a big fan of the pattern. I think I would have created a `quote()` method that would have enclosed directly the element with the right quoting.

YMMV, so feel free to push it if you think your version is OK.

getamihalcea:
@gsmet Either way is fine for me.

gsmet:
@vladmihalcea I force-pushed an additional commit to your branch.

Could you check that it works OK with all the databases you tested? I also added `MySQLDialect` so we should also test the patch with MySQL or MariaDB (following this forum post https://discourse.hibernate.org/t/mysql-migration-issue/1378).

Thanks!

vladmihalcea:
I'll check it out tomorrow when I get home.

vladmihalcea:
@gsmet I tested your change but it fails on both PostgreSQL and SQL Server. For instance, on SQL Server, your RegEx pattern is `create table [my\-schema]\.[my_entity]` which does not escape the `[` and `]` RegEx reserved characters.

With your change, the tests only work on H2.

gsmet:
@vladmihalcea 

I fixed the PostgreSQL issue which was due to the fact that it includes a `if exists` in the `alter table` command.

I think I also fixed the SQL Server issue but doesn't have one to test it. Could you retry? Thanks!

vladmihalcea:
@gsmet I had to make this change to make it work on SQL Server, as otherwise the `[` and `]` quotes for the table identifier were not escaped and the RegEx was failing:

```
       protected String regexpQuote(String schema, String table) {
 -         return regexpQuote( schema ) + "\\." + quote( table );
 +         return regexpQuote( schema ) + "\\." + regexpQuote( table );
       }

```

vladmihalcea:
@gsmet If everything works fine on all major DBs, I'll squash the commits and push force it.

PR #2530, opened 2018-09-11
[HHH-12969] Identifiable.getJavaType() returns IdClass member type rather than Entity class member type

jwgmeligmeyling:
Reproduces https://hibernate.atlassian.net/browse/HHH-12969


sebersole:
For whatever reason I had conflicts trying to rebase this PR.  But I did add a test inspired by yours.  It is working as expected on main.

PR #2532, opened 2018-09-13
Fix Constructor Not Found with Lombok's @Value + JPA Projections

apapenko:
**Fix Constructor Not Found with Lombok's @Value when used with projections and Spring Data JPA**
Couldn't find bug-reporting, so suggesting this fix for the problem. 
Right now created all args constructor manually to avoid this issue.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/util/ReflectHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/util/ReflectHelper.java
@@ -315,7 +315,7 @@ public final class ReflectHelper {
 	 * @throws PropertyNotFoundException Indicates we could not locate an appropriate constructor (todo : again with PropertyNotFoundException???)
 	 */
 	public static Constructor getConstructor(Class clazz, Type[] types) throws PropertyNotFoundException {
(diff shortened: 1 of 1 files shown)


vladmihalcea:
We need 2 things for this PR:

- a Jira issue describing the problem
- a test case that replicates what Lombok does behind the scenes. I suppose it generates a private constructor, right? So please add an entity which emulates what Lombok does so we can validate your change.

apapenko:
@vladmihalcea hi! I can provide a test case, but I don't think I have access to Jira

vladmihalcea:
Anyone can create a Jira issue. Just create an account.

Vlad

gsmet:
I concur: we certainly need far more information before doing this change.

sebersole:
WRT private constructors... Hibernate supports accessing private constructors - normally.  It is conceivable that this could be some new problem stemming from Java's module visibility (assuming the JVM is 10+).

gsmet:
Closing this PR as we don't have enough information to decide if the fix is right or not.

Please provide further information if we want us to fix this at some point.

Thanks.

PR #2535, opened 2018-09-14
HHH-12977 - Update latest dialect for MySQL

quaff:
Currently MySQL8Dialect should be the latest.


vladmihalcea:
Applied upstream, thanks

PR #2539, opened 2018-09-19
HHH-12982 - Generify Hibernate#unproxy

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12982

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/Hibernate.java
+++ b/hibernate-core/src/main/java/org/hibernate/Hibernate.java
@@ -212,4 +212,16 @@ public final class Hibernate {
 			return proxy;
 		}
 	}
+
+	/**
+	 * Unproxies a {@link HibernateProxy}. If the proxy is uninitialized, it automatically triggers an initialization.
+	 * In case the supplied object is null or not a proxy, the object will be returned as-is.
+	 *
+	 * @param proxy the {@link HibernateProxy} to be unproxied
+	 * @param entityClass the entity type
+	 * @return the proxy's underlying implementation object, or the supplied object otherwise
+	 */
+	public static <T> T unproxy(T proxy, Class<T> entityClass) {
+		return entityClass.cast( unproxy( proxy ) );
+	}
 }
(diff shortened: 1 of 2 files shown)


gsmet:
@gbadner could you take a look at this one?

I think it's fine but wondering if there could be any weird corner case where people would pass a class that the implementation wouldn't implement.

I don't see any use case where it would make sense to pass the initial object as the proxy class though.

vladmihalcea:
@gbadner @gsmet I also thought of any possible backward compatibility we might break, but I think we should be fine since we already did that before with `Session#get` which we changed from `Object` to `T` as well.

gbadner:
@vladmihalcea, I believe that ```Session#get``` was changed in 5.0. We were not concerned about maintaining binary compatibility for APIs moving from 4.2 (used in EAP 6) to 5.0 (used in EAP 7). In 5.x, we need to keep APIs binary compatible (and even 6.0, IIRC), since they will likely continue to be consumed by EAP 7.x.

Another difference is that is that the application provides the `Class` to `Session#get` that it expects to be returned. That's not the case with `Hibernate#unproxy`.

When debugging the test case from ```HibernateUnproxyTest#testInitializedProxyCanBeUnproxied```, I can see that ```p.child``` is of type ```org.hibernate.test.proxy.HibernateUnproxyTest$Child$HibernateProxy$NIR55I68```. which extends ```Child``` and implements ```HibernateProxy```. It seems a bit strange to cast the return value to this subclass. It can also cause problems like the following:

An application could have: 
HibernateProxy child = (HibernateProxy) parent.getChild();

// The following would cause a compiler error
Child unproxiedChildProxy = Hibernate.unproxy( child );

// The following doesn't cause a compiler error; instead ClassCastException is thrown at runtime.
HibernateProxy unproxiedChildProxy = Hibernate.unproxy( childProxy );

My preference would be add a new static method with a parameter for the expected class to be returned: 

`public static <T> T unproxy(Class<T> clazz, T proxy)`

vladmihalcea:
I'll add the new method instead as suggested by @gbadner.

vladmihalcea:
@gbadner @gsmet I pushed the version as suggested by Gail. There's a new method now. I think it's safe to integrate it.

gbadner:
@vladmihalcea, this looks good to me now.

PR #2540, opened 2018-09-19
HHH-12978 - Enum value binding is not logged by BasicBinder

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12978

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/model/convert/internal/NamedEnumValueConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/model/convert/internal/NamedEnumValueConverter.java
@@ -13,10 +13,13 @@ import java.sql.SQLException;
 import java.sql.Types;
 import java.util.Locale;
 
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.metamodel.model.convert.spi.EnumValueConverter;
+import org.hibernate.type.descriptor.ValueBinder;
+import org.hibernate.type.descriptor.ValueExtractor;
 import org.hibernate.type.descriptor.java.EnumJavaTypeDescriptor;
-
-import org.jboss.logging.Logger;
+import org.hibernate.type.descriptor.java.StringTypeDescriptor;
+import org.hibernate.type.descriptor.sql.VarcharTypeDescriptor;
 
 /**
  * BasicValueConverter handling the conversion of an enum based on
@@ -25,12 +28,17 @@ import org.jboss.logging.Logger;
  * @author Steve Ebersole
  */
 public class NamedEnumValueConverter<E extends Enum> implements EnumValueConverter<E,String>, Serializable {
-	private static final Logger log = Logger.getLogger( NamedEnumValueConverter.class );
 
 	private final EnumJavaTypeDescriptor<E> enumJavaDescriptor;
 
+	private final transient ValueExtractor<E> valueExtractor;
+
+	private final transient ValueBinder<String> valueBinder;
+
 	public NamedEnumValueConverter(EnumJavaTypeDescriptor<E> enumJavaDescriptor) {
 		this.enumJavaDescriptor = enumJavaDescriptor;
+		this.valueExtractor = VarcharTypeDescriptor.INSTANCE.getExtractor( enumJavaDescriptor );
+		this.valueBinder = VarcharTypeDescriptor.INSTANCE.getBinder( StringTypeDescriptor.INSTANCE );
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/model/convert/internal/NamedEnumValueConverter.java
```diff
@@ -56,42 +55,19 @@ public int getJdbcTypeCode() {
...
-			return null;
-		}
-
-		final E enumValue = toDomainValue( value );
-		if ( traceEnabled ) {
-			log.trace( String.format( "Returning [%s] as column [%s]", enumValue, name ) );
-		}
-
-		return enumValue;
+		return VarcharTypeDescriptor.INSTANCE.getExtractor( enumJavaDescriptor ).extract( resultSet, name, null );
```
gsmet: Shouldn't we initialize `VarcharTypeDescriptor.INSTANCE.getExtractor( enumJavaDescriptor )` once and for all in the constructor?

Same question for the binder (and obviously the Integer version).

AFAICS, the objects are stateless.

vladmihalcea: Good point. We could definitely do that.

vladmihalcea:
@gsmet I pushed some changes according to your suggestions. Let me know what you think.

hibernate-core/src/main/java/org/hibernate/metamodel/model/convert/internal/NamedEnumValueConverter.java
```diff
@@ -25,12 +27,17 @@
  * @author Steve Ebersole
  */
 public class NamedEnumValueConverter<E extends Enum> implements EnumValueConverter<E,String>, Serializable {
-	private static final Logger log = Logger.getLogger( NamedEnumValueConverter.class );
 
 	private final EnumJavaTypeDescriptor<E> enumJavaDescriptor;
 
+	private final transient ValueExtractor<E> valueExtractor;
+
+	private final transient ValueBinder<String> valueBinder;
```
gsmet: I see you marked them as transient. Do we expect the converter to be serialized (I see it's `Serializable` so this part is true) and then reused (wondering about this specific part)?

If so, we would need to implement a `readObject()` method calling the default `readObject()` and then reinitializing the two fields. Obviously, you would have to remove the `final` qualifier in this case.

sebersole: In 5.x the converter might be serialized, so this part is fine.

Seeing the null getting passed along, I wonder if we want to change the signature of `EnumValueConverter#readValue` and `EnumValueConverter#writeValue` to pass along the Session

gsmet: @sebersole 

About `In 5.x the converter might be serialized, so this part is fine.`, you're saying that adding a `readObject()` method is unnecessary?

I don't know if we expect the converter to be reused after deserialization or if we just want it to be serializable?

About adding the session as a parameter, it seems like a good idea. `EnumValueConverter` is an SPI but I don't expect it to be implemented elsewhere.

vladmihalcea: I don't understand how changing the signature from `WrapperOptions` to `Session` would solve the null passing problem. Either way, we won't have a reference to the `Session` in the `readValue` method:

```
@Override
public E readValue(ResultSet resultSet, String name) throws SQLException {
	return valueExtractor.extract( resultSet, name, null );
}
```

sebersole: Which is why I said to pass it in to the converter.  Which is in fact exactly what we have done in 6 for just this reason.  So:

````
    @Override
    E readValue(
            ResultSet resultSet, 
            String name,  
            SharedSessionContractImplementor session) throws SQLException {
	return valueExtractor.extract( resultSet, name, session );
}
````

sebersole: @gsmet 

> About `In 5.x the converter might be serialized, so this part is fine.`, you're saying that adding a `readObject()` method is unnecessary?

I guess it depends on where the converter is serialized "from".  

If it is serialized as part of the SessionFactory (which I think is true) then the readObject is unnecessary because we never *really* serialize the SessionFactory - we simply resolve it readResolve (though in retrospect, I think Externalizable would have been better).

It is a different story if it ever gets serialized as part of a Session (outside of the Session's reference to SF).  There it would need the readObject/


> About adding the session as a parameter, it seems like a good idea. `EnumValueConverter` is an SPI but I don't expect it to be implemented elsewhere.

Right.  The "API" would be JPA's AttributeConverter.

PR #2541, opened 2018-09-23
HHH-12968 - Child tables in InheritanceType.SINGLE_TABLE not save properly

j-alimohammadi:
https://hibernate.atlassian.net/browse/HHH-12968


vladmihalcea:
Applied upstream after adding one more test and simplified the condition, thanks

PR #2542, opened 2018-09-24
HHH-12968 - Flush is not flushing inserts for inherited tables before a select within a transaction

vladmihalcea:
Extract IdentityGenerator batch support validation logic

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/id/IdentifierGenerator.java
+++ b/hibernate-core/src/main/java/org/hibernate/id/IdentifierGenerator.java
@@ -7,7 +7,6 @@
 package org.hibernate.id;
 
 import java.io.Serializable;
-
 import javax.persistence.GeneratedValue;
 
 import org.hibernate.HibernateException;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/id/IdentifierGenerator.java
```diff
@@ -60,4 +59,13 @@
...
 	Serializable generate(SharedSessionContractImplementor session, Object object) throws HibernateException;
+
+	/**
+	 * Check if JDBC batch inserts are supported.
+	 *
+	 * @return JDBC batch inserts are supported.
+	 */
+	default boolean supportsJdbcBatchInserts() {
+		return true;
+	}
```
gsmet: @sebersole could you check this out? I think it's cleaner to have a proper contract rather than relying on an `instanceof IdentityGenerator` but I would like to be sure it's not an issue to add this method here.

sebersole: With the default, I think its fine to add

sebersole: I always prefer a "proper contract"

vladmihalcea: To make it clear, was it ok to merge it or you wanted to have it implemented differently?

sebersole: The implementation is great.  I was just agreeing with @gsmet

vladmihalcea: I agree. It's much cleaner now.

PR #2543, opened 2018-09-26
HHH-7913 Placeholders substitution in subselect

OleksiiChumak:


OleksiiChumak:
https://hibernate.atlassian.net/browse/HHH-7913

gsmet:
@gbadner I think it makes a lot of sense but could you take a quick look?

gbadner:
I just pushed a commit with a failing test that maps the entities to a non-default schema. It fails because the placeholder is substituted with the default schema instead of the mapped schema.

After pushing that commit, I realized that the user guide says that "{h-schema} resolves the current hibernate.default_schema configuration property value" [1], so maybe a different placeholder should be  used if we want to allow a mapped, non-default schema.

The `Table#schema` for the subselect has the correct schema name, so the information is there. The solution should take into account what is mapped, rather than just assuming the default should be used.

[1] http://docs.jboss.org/hibernate/orm/5.3/userguide/html_single/Hibernate_User_Guide.html#sql-global-catalog-schema

OleksiiChumak:
Pushed a change. As was proposed If schema is configured on the entity level, the implementation chooses it instead of the default.

gbadner:
In [17.11. Resolving global catalog and schema in native SQL queries, the documentation explicitly](http://docs.jboss.org/hibernate/orm/5.3/userguide/html_single/Hibernate_User_Guide.html#sql-global-catalog-schema) says:

"{h-schema} 
resolves the current hibernate.default_schema configuration property value."

Documentation for [2.5.8. Mapping the entity to a SQL query](http://docs.jboss.org/hibernate/orm/5.3/userguide/html_single/Hibernate_User_Guide.html#entity-sql-query-mapping) should be updated in a similar way, but mentioning that if a schema is mapped for the entity, then it will be used instead.

gbadner:
If adding support for substituting {h-schema} is added, support for {h-catalog} and {h-domain} should probably be added as well.

OleksiiChumak:
Actually, the method that resolves {h-schema} also resolves {h-catalog} and {h-domain}. So they are supported. 
To be consistent with the previous commit I can implement the same behavior as for {h-schema} (first look at Table, if nothing is specified there than resolve placeholders against global settings)
But does it make sense to use `@Subselect` with `@Table` ?

OleksiiChumak:
Added same support for {h-schema} and {h-domain} as for {h-schema}. Changed documentation

sebersole:
See https://github.com/hibernate/hibernate-orm/pull/9293

PR #2545, opened 2018-09-27
HHH-12993 - Omit joining of superclass table when querying subclass only

ladislavovic:
* it go through AST and find out which tables are referenced. If super
  class table in "joined" inheritance strategy is not referenced, is is
  not joined
* when any filter is active then all super class table are joined, because
  filter can reference them and I can not find it out from query AST analyze.
  It could be optimized probably somehow.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
@@ -98,6 +98,7 @@ import static org.hibernate.cfg.AvailableSettings.LOG_SESSION_METRICS;
 import static org.hibernate.cfg.AvailableSettings.MAX_FETCH_DEPTH;
 import static org.hibernate.cfg.AvailableSettings.MULTI_TENANT_IDENTIFIER_RESOLVER;
 import static org.hibernate.cfg.AvailableSettings.NATIVE_EXCEPTION_HANDLING_51_COMPLIANCE;
+import static org.hibernate.cfg.AvailableSettings.OMIT_JOIN_OF_SUPERCLASS_TABLES;
 import static org.hibernate.cfg.AvailableSettings.ORDER_INSERTS;
 import static org.hibernate.cfg.AvailableSettings.ORDER_UPDATES;
 import static org.hibernate.cfg.AvailableSettings.PREFER_USER_TRANSACTION;
@@ -206,6 +207,7 @@ public class SessionFactoryOptionsBuilder implements SessionFactoryOptions {
 	private final boolean procedureParameterNullPassingEnabled;
 	private final boolean collectionJoinSubqueryRewriteEnabled;
 	private boolean jdbcStyleParamsZeroBased;
+	private final boolean omitJoinOfSuperclassTablesEnabled;
 
 	// Caching
 	private boolean secondLevelCacheEnabled;
@@ -350,6 +352,7 @@ public class SessionFactoryOptionsBuilder implements SessionFactoryOptions {
 				CONVENTIONAL_JAVA_CONSTANTS, BOOLEAN, true );
 		this.procedureParameterNullPassingEnabled = cfgService.getSetting( PROCEDURE_NULL_PARAM_PASSING, BOOLEAN, false );
 		this.collectionJoinSubqueryRewriteEnabled = cfgService.getSetting( COLLECTION_JOIN_SUBQUERY, BOOLEAN, true );
+		this.omitJoinOfSuperclassTablesEnabled = cfgService.getSetting( OMIT_JOIN_OF_SUPERCLASS_TABLES, BOOLEAN, true );
 
 		final RegionFactory regionFactory = serviceRegistry.getService( RegionFactory.class );
 		if ( !NoCachingRegionFactory.class.isInstance( regionFactory ) ) {
@@ -1042,6 +1045,11 @@ public class SessionFactoryOptionsBuilder implements SessionFactoryOptions {
 	@Override
 	public int getQueryStatisticsMaxSize() {
 		return queryStatisticsMaxSize;
+	}
+
+	@Override
+	public boolean isOmitJoinOfSuperclassTablesEnabled() {
+		return omitJoinOfSuperclassTablesEnabled;
 	}
 
 	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
@@ -1992,4 +1992,19 @@ public interface AvailableSettings extends org.hibernate.jpa.AvailableSettings {
 	 * @since 5.4
 	 */
 	String SEQUENCE_INCREMENT_SIZE_MISMATCH_STRATEGY = "hibernate.id.sequence.increment_size_mismatch_strategy";
+
+	/**
+	 * <p>
+	 * When you use {@link javax.persistence.InheritanceType#JOINED} strategy for inheritance mapping and query
+	 * a value from an entity, all superclass tables are joined in the query regardless you need them. With
+	 * this setting set to true only superclass tables which are really needed are joined.
+	 * </p>
+	 * <p>
+	 * The default value is true.
+	 * </p>
+	 *
+	 * @since 5.4
+	 */
+	String OMIT_JOIN_OF_SUPERCLASS_TABLES = "hibernate.query.omit_join_of_superclass_tables";
+
 }
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
@@ -74,10 +74,18 @@ public class DotNode extends FromReferenceNode implements DisplayableNode, Selec
 	 * The identifier that is the name of the property.
 	 */
 	private String propertyName;
+
+	/**
+	 * The identifier that is the name of the property. In comparison with {@link #propertyName}
+	 * it is always identical with identifier in the query, it is not changed during processing.
+	 */
+	private String originalPropertyName;
+
 	/**
 	 * The full path, to the root alias of this dot node.
 	 */
 	private String path;
+
 	/**
 	 * The unresolved property path relative to this dot node.
 	 */
@@ -158,6 +166,7 @@ public class DotNode extends FromReferenceNode implements DisplayableNode, Selec
 		// Set the attributes of the property reference expression.
 		String propName = property.getText();
 		propertyName = propName;
+		originalPropertyName = propName;
 		// If the uresolved property path isn't set yet, just use the property name.
 		if ( propertyPath == null ) {
 			propertyPath = propName;
@@ -690,12 +699,39 @@ public class DotNode extends FromReferenceNode implements DisplayableNode, Selec
 		return super.getDataType();
 	}
 
+	@Override
+	public String[] getReferencedTables() {
+		String[] referencedTables = null;
+		AST firstChild = getFirstChild();
+		if ( firstChild != null ) {
+			if ( firstChild instanceof FromReferenceNode ) {
+				FromReferenceNode fromReferenceNode = (FromReferenceNode) firstChild;
+				FromElement fromElement = fromReferenceNode.getFromElement();
+				if ( fromElement != null ) {
+					String table = fromElement.getPropertyTableName( getOriginalPropertyName() );
+					if ( table != null ) {
+						referencedTables = new String[] { table };
+					}
+				}
+			}
+		}
+		return referencedTables;
+	}
+
 	public void setPropertyPath(String propertyPath) {
 		this.propertyPath = propertyPath;
 	}
 
 	public String getPropertyPath() {
 		return propertyPath;
+	}
+
+	public String getPropertyName() {
+		return propertyName;
+	}
+
+	public String getOriginalPropertyName() {
+		return originalPropertyName;
 	}
 
 	public FromReferenceNode getLhs() {
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
@@ -14,21 +14,18 @@ import java.util.Set;
 import org.hibernate.MappingException;
 import org.hibernate.QueryException;
 import org.hibernate.engine.internal.JoinSequence;
-import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.hql.internal.CollectionProperties;
 import org.hibernate.hql.internal.CollectionSubqueryFactory;
 import org.hibernate.hql.internal.NameGenerator;
 import org.hibernate.hql.internal.antlr.HqlSqlTokenTypes;
-import org.hibernate.hql.internal.ast.HqlSqlWalker;
-import org.hibernate.hql.internal.ast.util.SessionFactoryHelper;
 import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
-import org.hibernate.internal.log.DeprecationLogger;
 import org.hibernate.internal.util.collections.ArrayHelper;
 import org.hibernate.param.ParameterSpecification;
 import org.hibernate.persister.collection.CollectionPropertyMapping;
 import org.hibernate.persister.collection.CollectionPropertyNames;
 import org.hibernate.persister.collection.QueryableCollection;
+import org.hibernate.persister.entity.AbstractEntityPersister;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.persister.entity.Joinable;
 import org.hibernate.persister.entity.PropertyMapping;
@@ -366,6 +363,15 @@ class FromElementType {
 
 	public QueryableCollection getQueryableCollection() {
 		return queryableCollection;
+	}
+
+	public String getPropertyTableName(String propertyName) {
+		checkInitialized();
+		if ( this.persister != null ) {
+			AbstractEntityPersister aep = (AbstractEntityPersister) this.persister;
+			return aep.getPropertyTableName( propertyName );
+		}
+		return null;
 	}
 
 	/**
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromReferenceNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromReferenceNode.java
@@ -137,4 +137,14 @@ public abstract class FromReferenceNode extends AbstractSelectExpression
 				|| getWalker().getStatementType() == HqlSqlTokenTypes.UPDATE;
 	}
 
+	/**
+	 * Returns table names which are referenced by this node. If the tables
+	 * can not be determined it returns null.
+	 *
+	 * @return table names or null.
+	 */
+	public String[] getReferencedTables() {
+		return null;
+	}
+
 }
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/ASTPrinter.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/ASTPrinter.java
@@ -10,6 +10,8 @@ import java.io.ByteArrayOutputStream;
 import java.io.PrintStream;
 import java.io.PrintWriter;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.hibernate.hql.internal.ast.tree.DisplayableNode;
@@ -25,7 +27,7 @@ import antlr.collections.AST;
  * @author Joshua Davis
  * @author Steve Ebersole
  */
-public final class ASTPrinter {
+public class ASTPrinter {
 
 	// This is a map: array index is the ANTLR Token ID, array value is the name of that token.
 	// There might be gaps in the array (null values) but it's generally quite compact.
@@ -103,15 +105,7 @@ public final class ASTPrinter {
 			return;
 		}
 
-		for ( AST parent : parents ) {
-			if ( parent.getNextSibling() == null ) {
-
-				pw.print( "   " );
-			}
-			else {
-				pw.print( " | " );
-			}
-		}
+		indentLine( parents, pw );
 
 		if ( ast.getNextSibling() == null ) {
 			pw.print( " \\-" );
@@ -121,6 +115,7 @@ public final class ASTPrinter {
 		}
 
 		showNode( pw, ast );
+		showNodeProperties( parents, pw, ast );
 
 		ArrayList<AST> newParents = new ArrayList<AST>( parents );
 		newParents.add( ast );
@@ -128,6 +123,17 @@ public final class ASTPrinter {
 			showAst( newParents, pw, child );
 		}
 		newParents.clear();
+	}
+
+	private void indentLine(List<AST> parents, PrintWriter pw) {
+		for ( AST parent : parents ) {
+			if ( parent.getNextSibling() == null ) {
+				pw.print( "   " );
+			}
+			else {
+				pw.print( " | " );
+			}
+		}
 	}
 
 	private void showNode(PrintWriter pw, AST ast) {
@@ -156,6 +162,24 @@ public final class ASTPrinter {
 			buf.append( " " ).append( displayableNode.getDisplayText() );
 		}
 		return buf.toString();
+	}
+
+	private void showNodeProperties(ArrayList<AST> parents, PrintWriter pw, AST ast) {
+		Map<String, Object> nodeProperties = createNodeProperties( ast );
+		ArrayList<AST> parentsAndNode = new ArrayList<>( parents );
+		parentsAndNode.add( ast );
+		for ( String propertyName : nodeProperties.keySet() ) {
+			indentLine( parentsAndNode, pw );
+			pw.println( propertyToString( propertyName, nodeProperties.get( propertyName ), ast ) );
+		}
+	}
+
+	public LinkedHashMap<String, Object> createNodeProperties(AST ast) {
+		return new LinkedHashMap<>();
+	}
+
+	public String propertyToString(String label, Object value, AST ast) {
+		return String.format( "%s: %s", label, value );
 	}
 
 	public static void appendEscapedMultibyteChars(String text, StringBuilder buf) {
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
@@ -8,14 +8,17 @@ package org.hibernate.hql.internal.ast.util;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
+import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.hibernate.AssertionFailure;
+import org.hibernate.cfg.AvailableSettings;
 import org.hibernate.dialect.Dialect;
 import org.hibernate.engine.internal.JoinSequence;
 import org.hibernate.engine.spi.LoadQueryInfluencers;
@@ -24,6 +27,7 @@ import org.hibernate.hql.internal.ast.HqlSqlWalker;
 import org.hibernate.hql.internal.ast.tree.DotNode;
 import org.hibernate.hql.internal.ast.tree.FromClause;
 import org.hibernate.hql.internal.ast.tree.FromElement;
+import org.hibernate.hql.internal.ast.tree.FromReferenceNode;
 import org.hibernate.hql.internal.ast.tree.ImpliedFromElement;
 import org.hibernate.hql.internal.ast.tree.ParameterContainer;
 import org.hibernate.hql.internal.ast.tree.QueryNode;
@@ -33,10 +37,15 @@ import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.internal.FilterImpl;
 import org.hibernate.internal.util.StringHelper;
 import org.hibernate.internal.util.collections.ArrayHelper;
+import org.hibernate.internal.util.collections.CollectionHelper;
 import org.hibernate.param.DynamicFilterParameterSpecification;
+import org.hibernate.persister.entity.AbstractEntityPersister;
+import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.sql.JoinFragment;
 import org.hibernate.sql.JoinType;
 import org.hibernate.type.Type;
+
+import antlr.collections.AST;
 
 /**
  * Performs the post-processing of the join information gathered during semantic analysis.
@@ -94,8 +103,85 @@ public class JoinProcessor implements SqlTokenTypes {
 		}
 	}
 
+	private <T extends AST> List<T> findAllNodes(AST node, Class<T> clazz) {
+		ArrayList<T> found = new ArrayList<>();
+		doFindAllNodes( node, clazz, found );
+		return found;
+	}
+
+	private <T extends AST> void doFindAllNodes(AST node, Class<T> clazz, List<T> found) {
+		if ( clazz.isAssignableFrom( node.getClass() ) ) {
+			found.add( (T) node );
+		}
+		if ( node.getFirstChild() != null ) {
+			doFindAllNodes( node.getFirstChild(), clazz, found );
+		}
+		if ( node.getNextSibling() != null ) {
+			doFindAllNodes( node.getNextSibling(), clazz, found );
+		}
+	}
+
+	private Set<String> findQueryReferencedTables(QueryNode query) {
+		if ( !walker.getSessionFactoryHelper()
+				.getFactory()
+				.getSessionFactoryOptions()
+				.isOmitJoinOfSuperclassTablesEnabled() ) {
+			if ( LOG.isDebugEnabled() ) {
+				LOG.debug( String.format(
+						"Finding of query referenced tables is skipped because the feature is disabled. See %s",
+						AvailableSettings.OMIT_JOIN_OF_SUPERCLASS_TABLES
+				) );
+			}
+			return null;
+		}
+
+		if ( CollectionHelper.isNotEmpty( walker.getEnabledFilters() ) ) {
+			LOG.debug( "Finding of query referenced tables is skipped because filters are enabled." );
+			return null;
+		}
+
+		if ( LOG.isDebugEnabled() ) {
+			LOG.debug( TokenPrinters.REFERENCED_TABLES_PRINTER.showAsString(
+					query,
+					"Tables referenced from query nodes:"
+			) );
+		}
+
+		Set<String> result = new HashSet<>();
+
+		// Find tables referenced by FromReferenceNodes
+		List<FromReferenceNode> fromReferenceNodes = findAllNodes( query, FromReferenceNode.class );
+		for ( FromReferenceNode node : fromReferenceNodes ) {
+			String[] tables = node.getReferencedTables();
+			if ( tables != null ) {
+				for ( String table : tables ) {
+					result.add( table );
+				}
+			}
+		}
+
+		// Find tables referenced by fromElementsForLoad
+		if ( query.getSelectClause() != null ) {
+			for ( Object element : query.getSelectClause().getFromElementsForLoad() ) {
+				FromElement fromElement = (FromElement) element;
+				EntityPersister entityPersister = fromElement.getEntityPersister();
+				if ( entityPersister != null && entityPersister instanceof AbstractEntityPersister ) {
+					AbstractEntityPersister aep = (AbstractEntityPersister) entityPersister;
+					String[] tables = aep.getTableNames();
+					for ( String table : tables ) {
+						result.add( table );
+					}
+				}
+			}
+		}
+
+		return result;
+	}
+
 	public void processJoins(QueryNode query) {
 		final FromClause fromClause = query.getFromClause();
+
+		Set<String> queryReferencedTables = findQueryReferencedTables( query );
 
 		final List fromElements;
 		if ( DotNode.useThetaStyleImplicitJoins ) {
@@ -136,6 +222,7 @@ public class JoinProcessor implements SqlTokenTypes {
 		while ( iter.hasNext() ) {
 			final FromElement fromElement = (FromElement) iter.next();
 			JoinSequence join = fromElement.getJoinSequence();
+			join.setQueryReferencedTables( queryReferencedTables );
 			join.setSelector(
 					new JoinSequence.Selector() {
 						public boolean includeSubclasses(String alias) {
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
@@ -521,6 +521,14 @@ public abstract class AbstractEntityPersister
 		return propertySelectable;
 	}
 
+	public String[] getTableNames() {
+		String[] tableNames = new String[getTableSpan()];
+		for ( int i = 0; i < tableNames.length; i++ ) {
+			tableNames[i] = getTableName( i );
+		}
+		return tableNames;
+	}
+
 	@SuppressWarnings("UnnecessaryBoxing")
 	public AbstractEntityPersister(
 			final PersistentClass persistentClass,
@@ -3850,7 +3858,8 @@ public abstract class AbstractEntityPersister
 				alias,
 				innerJoin,
 				includeSubclasses,
-				Collections.emptySet()
+				Collections.emptySet(),
+				null
 		).toFromFragmentString();
 	}
 
@@ -3863,7 +3872,19 @@ public abstract class AbstractEntityPersister
 		// NOTE : Not calling createJoin here is just a performance optimization
 		return getSubclassTableSpan() == 1
 				? ""
-				: createJoin( alias, innerJoin, includeSubclasses, treatAsDeclarations ).toFromFragmentString();
+				: createJoin( alias, innerJoin, includeSubclasses, treatAsDeclarations, null ).toFromFragmentString();
+	}
+
+	@Override
+	public String fromJoinFragment(
+			String alias,
+			boolean innerJoin,
+			boolean includeSubclasses,
+			Set<String> treatAsDeclarations,
+			Set<String> referencedTables) {
+		return getSubclassTableSpan() == 1
+				? ""
+				: createJoin( alias, innerJoin, includeSubclasses, treatAsDeclarations, referencedTables ).toFromFragmentString();
 	}
 
 	@Override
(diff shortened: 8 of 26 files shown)


ladislavovic:
Damn, some tests failed. It depends on environment. Here 5 tests failed, on my Ubuntu 18_04 with openjdk-8 3 failed and on my dev machine with win10 and oracle-javva-8 are all tests OK. So I am solving that.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -521,6 +521,8 @@ public boolean hasRowId() {
 		return propertySelectable;
 	}
 
+	public abstract String[] getTableNames();
```
beikov: Not sure if `AbstractEntityPersister` can be considered an API/SPI but IMO a default implementation should just return null here to be safe.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +103,73 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+			}
+		}
+
+		return result;
+	}
+
 	public void processJoins(QueryNode query) {
 		final FromClause fromClause = query.getFromClause();
 
+		Set<String> queryReferencedTables = findQueryReferencedTables( query );
```
beikov: IMO this should be deactivatable by a property so that one can deactivate this feature in case of a bug.

hibernate-core/src/test/java/org/hibernate/test/hql/HQLTest.java
```diff
@@ -241,10 +241,13 @@ public void testCollectionMemberDeclarations(){
 		// IN asks an alias, but the difference is that the error message from AST
 		// contains the error token location (by lines and columns), which is hardly 
 		// to get from Classic query translator --stliu
-		assertTranslation( "from Customer c, in(c.orders)" ); 
+		assertTranslation( "from Customer c, in(c.orders)" );
 	}
 
 	@Test
+	@FailureExpected(
```
beikov: When introducing a configuration option, you can re-enable these tests.

JWGmeligMeyling: Imho these tests should never be deactivated, I'd rather see them adjusted to the new PR.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
ladislavovic: I Added `AvailableSettings.OMIT_JOIN_OF_SUPERCLASS_TABLES` property. If it is false, the feature is deactivated. Default value is true.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
ladislavovic: Thanks for the implementation of getTableNames(), I've done it like so.

hibernate-core/src/main/java/org/hibernate/persister/entity/UnionSubclassEntityPersister.java
```diff
@@ -222,6 +222,11 @@ public String getTableName() {
 		return subquery;
 	}
 
+	@Override
+	public String[] getTableNames() {
+		return new String[] {tableName};
```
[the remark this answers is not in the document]
ladislavovic: I think you do not have to be afraid about `UnionSubclassEntityPersister`. It totally ignores `referencedTables` collection. Now, when the implementation of getTableNames() is in `AbstractEntityPersister`, `UnionSubclassEntityPersister` has no change in comparison with master branch.

Only `JoinedSubclassEntityPersister` use this information to optimize table joining.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -3930,9 +3948,15 @@ protected JoinType determineSubclassTableJoinType(
 			int subclassTableNumber,
 			boolean canInnerJoin,
 			boolean includeSubclasses,
-			Set<String> treatAsDeclarations) {
+			Set<String> treatAsDeclarations,
+			Set<String> referencedTables) {
```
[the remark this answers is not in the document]
ladislavovic: I added the methods.

hibernate-core/src/main/java/org/hibernate/persister/entity/UnionSubclassEntityPersister.java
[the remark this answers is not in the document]
ladislavovic: > I am not only worried about this issue with respect to subclass tables, but also for example in the context of secondary tables. I assume that if you only select a property from a secondary table, this will be the only table joined? That would be slightly odd behaviour. Can you for example add a test of entities A and B (not sharing the same root class) that reference a secondary table C, and then show the produced SQL and output for SELECT a.somePropOnC FROM A a? I assume this would now return somePropOnC values for B too, which is obviously wrong.

I added test for that, see `OmitAncestorJoinWithSecondaryTableTest`.

hibernate-core/src/test/java/org/hibernate/test/hql/HQLTest.java
ladislavovic: Tests are re-enabled already.

hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
```diff
@@ -486,6 +486,11 @@ public String getTableName() {
 		return qualifiedTableNames[0];
 	}
 
+	@Override
+	public String[] getTableNames() {
+		return qualifiedTableNames;
```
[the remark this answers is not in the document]
ladislavovic: I've briefly checked the issue an maybe this approach could be used to solve that. Now the collection `referencedTables` is passed to procedure which joining parent tables in `JoinedSubclassEntityPersister`. It could be probably passed also to other places to optimize joining.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -1961,4 +1961,18 @@
 	 */
 	String QUERY_STATISTICS_MAX_SIZE = "hibernate.statistics.query_max_size";
 
+	/**
+	 * <p>
+	 * When you use {@link javax.persistence.InheritanceType#JOINED} strategy for inheritance mapping and query
+	 * a value from a entity, all superclass tables are joined in the query regardless you need them. With
```
beikov: Typo `value from a entity` should be `value from an entity`

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -1961,4 +1961,18 @@
...
+	 * this settings set to true only superclass tables which are really needed are joined.
```
beikov: Typo `this settings` should be `this setting`

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -48,6 +59,9 @@
 public class JoinProcessor implements SqlTokenTypes {
 	private static final CoreMessageLogger LOG = CoreLogging.messageLogger( JoinProcessor.class );
 
+	// TODO where to place the default value?
+	public static final String DEFAULT_OMIT_JOIN_OF_SUPERCLASS_TABLES = Boolean.TRUE.toString();
```
beikov: Please add the configuration option to `org.hibernate.boot.spi.SessionFactoryOptions`. Take a look at how `isCollectionJoinSubqueryRewriteEnabled()` is implemented and initialized. There you can see how to set the default value.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +108,87 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+			findAllNodesCore( node.getFirstChild(), clazz, found );
+		}
+		if ( node.getNextSibling() != null ) {
+			findAllNodesCore( node.getNextSibling(), clazz, found );
+		}
+	}
+
+	private Set<String> findQueryReferencedTables(QueryNode query) {
+		Map<String, Object> sessionFactoryProperties = walker.getSessionFactoryHelper().getFactory().getProperties();
+		Object enabled = sessionFactoryProperties.getOrDefault(
```
beikov: Use `walker.getSessionFactoryHelper().getFactory().getSessionFactoryOptions().isOmitJoinOfSuperclassTablesEnabled()`

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +108,87 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+				AvailableSettings.OMIT_JOIN_OF_SUPERCLASS_TABLES,
+				DEFAULT_OMIT_JOIN_OF_SUPERCLASS_TABLES
+		);
+		boolean disabled = !Boolean.parseBoolean( enabled.toString() );
+		if ( disabled ) {
+			LOG.debug( "Finding of query referenced tables is skipped because the feature is disabled by " +
```
beikov: Put a `LOG.isDebugEnabled()` around this.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +108,87 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+			}
+		}
+
+		return result;
+	}
+
 	public void processJoins(QueryNode query) {
 		final FromClause fromClause = query.getFromClause();
 
+		Map<String, Object> properties = walker.getSessionFactoryHelper().getFactory().getProperties();
```
beikov: Remove this line

JWGmeligMeyling:
Thanks for addressing the feedback. I'll have a look at the produced SQL from the tests and test the feature in my own domain and get back to you.

sebersole:
Please make sure we have plenty of tests for this.  Implementing this in 6 should be much easier, but the implementation will be completely different.  We want to make sure they functionality still works

hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
```diff
@@ -206,6 +207,7 @@
 	private final boolean procedureParameterNullPassingEnabled;
 	private final boolean collectionJoinSubqueryRewriteEnabled;
 	private boolean jdbcStyleParamsZeroBased;
+	private boolean omitJoinOfSuperclassTablesEnabled;
```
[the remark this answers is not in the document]
ladislavovic: Please which commits do you want to squash? You mean all 13 commits?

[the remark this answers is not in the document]
beikov: Yeah so that there are only 2 commits in the end, one containing only the test and another one containing only the implementation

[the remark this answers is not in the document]
ladislavovic: final is added

hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
```diff
@@ -350,6 +352,7 @@ public SessionFactoryOptionsBuilder(StandardServiceRegistry serviceRegistry, Boo
 				CONVENTIONAL_JAVA_CONSTANTS, BOOLEAN, true );
 		this.procedureParameterNullPassingEnabled = cfgService.getSetting( PROCEDURE_NULL_PARAM_PASSING, BOOLEAN, false );
 		this.collectionJoinSubqueryRewriteEnabled = cfgService.getSetting( COLLECTION_JOIN_SUBQUERY, BOOLEAN, true );
+		this.omitJoinOfSuperclassTablesEnabled = cfgService.getSetting( OMIT_JOIN_OF_SUPERCLASS_TABLES, BOOLEAN, true );
```
gsmet: Am I the only one who wouldn't make it the default behavior but only a possible optimization knob for advanced users?

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -1976,4 +1976,19 @@
...
+	 * <p>
+	 * When you use {@link javax.persistence.InheritanceType#JOINED} strategy for inheritance mapping and query
+	 * a value from an entity, all superclass tables are joined in the query regardless you need them. With
+	 * this setting set to true only superclass tables which are really needed are joined.
+	 * </p>
+	 * <p>
+	 * The default value is true.
+	 * </p>
+	 *
+	 * @Since 5.4
```
gsmet: `@since` without a capital.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -1976,4 +1976,19 @@
...
+	 */
+	String OMIT_JOIN_OF_SUPERCLASS_TABLES = "hibernate.query.omit_join_of_superclass_tables";
```
gsmet: I'm not a native speaker but the option name sounds weird to my ear. Any native speaker with a better proposal?

I would have gone with something like `lenient_superclass_table_join` because you don't really always omit them, you are just considering only the useful ones.

@sebersole you're usually good with finding names. Any better proposal?

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
```diff
@@ -74,6 +74,12 @@ public QueryException buildIllegalCollectionDereferenceException(String property
 	 * The identifier that is the name of the property.
 	 */
 	private String propertyName;
+
+	/**
+	 * The identifier that is the name of the property. In comparison with {@link #propertyName}
+	 * it is always identical with identifier in the query, it is not changed during processing.
+	 */
+	private String originalPropertyName;
```
gsmet: missing new line below this line.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
```diff
@@ -690,6 +697,25 @@ public Type getDataType() {
...
+	public String[] getReferencedTables() {
+		String[] referencedTables = null;
+		AST firstChild = getFirstChild();
+		if ( firstChild != null ) {
+			if ( firstChild instanceof FromReferenceNode ) {
+				FromReferenceNode fromReferenceNode = (FromReferenceNode) firstChild;
+				FromElement fromElement = fromReferenceNode.getFromElement();
+				if ( fromElement != null ) {
+					String table = fromElement.getPropertyTableName( getOriginalPropertyName() );
+					if (table != null) {
```
gsmet: Missing padding. Looks like our Checkstyle rules are not correctly enforced.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
```diff
@@ -368,6 +365,15 @@ public QueryableCollection getQueryableCollection() {
 		return queryableCollection;
 	}
 
+	public String getPropertyTableName(String propertyName) {
+		checkInitialized();
+		if (this.persister != null) {
```
gsmet: Same here about Checkstyle. I suppose we are not strict on the padding rule.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromReferenceNode.java
```diff
@@ -137,4 +137,14 @@ protected boolean isFromElementUpdateOrDeleteRoot(FromElement element) {
 				|| getWalker().getStatementType() == HqlSqlTokenTypes.UPDATE;
 	}
 
+	/**
+	 * Returns a table names which are referenced by this node. If the tables
```
gsmet: s/Returns a table names/Returns table names/

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/ASTPrinter.java
```diff
@@ -120,6 +114,7 @@ private void showAst(ArrayList<AST> parents, PrintWriter pw, AST ast) {
 		}
 
 		showNode( pw, ast );
+		showNodeProperties(parents, pw, ast);
```
gsmet: Padding.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +105,84 @@ public static JoinType toHibernateJoinType(int astJoinType) {
 		}
 	}
 
+	private <T extends AST> List<T> findAllNodes(AST node, Class<T> clazz) {
+		ArrayList<T> result = new ArrayList<>();
+		findAllNodesCore( node, clazz, result );
+		return result;
+	}
+
+	private <T extends AST> void findAllNodesCore(AST node, Class<T> clazz, List<T> found) {
```
gsmet: Maybe `doFindAllNodes`? I don't think we use `Core` anywhere else.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +105,84 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+			findAllNodesCore( node.getNextSibling(), clazz, found );
+		}
+	}
+
+	private Set<String> findQueryReferencedTables(QueryNode query) {
+		if ( !walker.getSessionFactoryHelper()
+				.getFactory()
+				.getSessionFactoryOptions()
+				.isOmitJoinOfSuperclassTablesEnabled() ) {
+			if (LOG.isDebugEnabled()) {
```
gsmet: padding.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +105,84 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+				LOG.debug( "Finding of query referenced tables is skipped because the feature is disabled by " +
+							"configuration. See " + AvailableSettings.OMIT_JOIN_OF_SUPERCLASS_TABLES );
+			}
+			return null;
+		}
+
+		if ( walker.getEnabledFilters() != null && !walker.getEnabledFilters().isEmpty() ) {
```
gsmet: See `CollectionHelper.isEmpty()`.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +105,84 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+
+		Set<String> result = new HashSet<>();
+
+		// Find tables referenced by FromReferenceNodes
+		List<FromReferenceNode> fromReferenceNodes = findAllNodes( query, FromReferenceNode.class );
+		for ( FromReferenceNode node : fromReferenceNodes ) {
+			String[] tables = node.getReferencedTables();
+			if ( tables != null ) {
+				Arrays.stream( tables )
+						.filter( Objects::nonNull )
```
gsmet: Can't we filter the nulls before adding them?

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +105,84 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+		// Find tables referenced by fromElementsForLoad
+		if ( query.getSelectClause() != null ) {
+			for ( Object element : query.getSelectClause().getFromElementsForLoad() ) {
+				FromElement fromElement = (FromElement) element;
+				EntityPersister entityPersister = fromElement.getEntityPersister();
+				if ( entityPersister != null && entityPersister instanceof AbstractEntityPersister ) {
+					AbstractEntityPersister aep = (AbstractEntityPersister) entityPersister;
+					String[] tables = aep.getTableNames();
+					Arrays.stream( tables )
+							.filter( Objects::nonNull )
```
gsmet: Do we really expect null here?

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -521,6 +521,14 @@ public boolean hasRowId() {
 		return propertySelectable;
 	}
 
+	public String[] getTableNames() {
+		String[] tableNames = new String[ getTableSpan() ];
+		for ( int i = 0; i < tableNames.length; i++ ) {
+			tableNames[ i ] = getTableName(i);
```
gsmet: I don't think we usually pad the index but the parameter should be.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +105,84 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+			}
+		}
+
+		// Find tables referenced by fromElementsForLoad
+		if ( query.getSelectClause() != null ) {
+			for ( Object element : query.getSelectClause().getFromElementsForLoad() ) {
+				FromElement fromElement = (FromElement) element;
+				EntityPersister entityPersister = fromElement.getEntityPersister();
+				if ( entityPersister != null && entityPersister instanceof AbstractEntityPersister ) {
+					AbstractEntityPersister aep = (AbstractEntityPersister) entityPersister;
```
gsmet: Maybe it's a stupid question but what if `canOmitSuperclassTableJoin()` of this persister returns false?

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +105,84 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+					String[] tables = aep.getTableNames();
```
gsmet: You have `getTableNames()` here whereas you have `getReferencedTables()` a few lines above. Doesn't sound very consistent.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromReferenceNode.java
```diff
@@ -137,4 +137,14 @@ protected boolean isFromElementUpdateOrDeleteRoot(FromElement element) {
...
+	 * can not be determined it returns null.
```
gsmet: You say `If the tables can not be determined it returns null.` which IMHO would imply that for some reasons, we might not be able to determine the tables and in this case, we should disable the optimization. That's not what I saw in the rest of the code.

Is it just me misunderstanding or you had plans to take this into account?

If we don't have a special case for `null`, I think a collector pattern might be more efficient. It would avoid all these array initializations.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -521,6 +521,14 @@ public boolean hasRowId() {
 		return propertySelectable;
 	}
 
+	public String[] getTableNames() {
+		String[] tableNames = new String[ getTableSpan() ];
```
gsmet: I discussed the opportunity of using a collector pattern (not sure it's the best solution, just mentioning it) but it's typically to avoid this.

Moreover, you do this work everytime you will call the method, and AFAICS the method will be called several times (e.g. for each query referencing this persister). That should be avoided if it can be.

hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
beikov: How would we get broad testing if we wouldn't enable the optimization by default and allow it to be deactivated for further baking if necessary? I think 5.4 is a good candidate for enabling this by default.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
ladislavovic: fixed

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
ladislavovic: Fixed. I must set automatic formatting in IntelliJIdea properly...

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
ladislavovic: fixed

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromReferenceNode.java
ladislavovic: fixed

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/ASTPrinter.java
ladislavovic: fixed

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
ladislavovic: fixed

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
ladislavovic: fixed

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
ladislavovic: fixed

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
ladislavovic: Yes, `getReferencedTables()` should not returns any null elements in the array, null check is useless. Removed.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
ladislavovic: again, null check is useless, removed.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
ladislavovic: fixed

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
ladislavovic: It does not matter. This method looks for all tables which are referenced by the query. And persister here just returns all tables which are used for entity mapping because all these tables are referenced.

`canOmitSuperclassTableJoin()` is used later on when a persister creates SQL from part of the query. Then a persister, if the `canOmitSuperclassTableJoin()` is `true`, does not have to use all tables bud can omit some of them.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
ladislavovic: `getReferencedTables()` is called on nodes and it returns tables (often it is a one table) referenced by the node. It does not returns all tables of the entity. If we rename the method to `getTableNames()`, it will not be so expressive.
`getTableNames()` is called on persiter and it just returns tables the entity is mapped to. The name is consistent with other persister methods.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromReferenceNode.java
ladislavovic: It does not mean we are not able to deremine tables. Some nodes just does not refer any table. Tthen it returns null, It could also returns an empty array, there is no semantic difference between them.

Long answer:  for examle expression `tableAlias.prop` is translated to following AST:
```
DotNode
  +- IdentNode (tableAlias)
  \- IdentNode (prop)
```
The second `IdentNode `has no reference to persister and also no reference to fromElement (both are null). It is impossible to get a reference to table from that. But it does not mean I miss a reference I just get it from another node. For example here I get it from DotNode. So to sum it up I get information from a subset of nodes, not from all.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
ladislavovic: I do not know collector pattern, please suggest another implementation of the method.

ladislavovic:
I have squashed that to two commits again. I think we should finish this issue till we all have it in recent memory.

ladislavovic:
@Sanne  I have inherited from `ASTPrinter `in this issue. But you made `ASTPrinter ``final `a few days ago. I have removed `final `to make it compillable again. Must it be final? If so how can I customize AST tree printing? Thank you!

Sanne:
@ladislavovic sorry, feel free to undo my change. I was optimising a different thing, and changed it to be final just because there was no reason not to.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
JWGmeligMeyling: Usually it is preferred in the persisters to compute these values once, and then store them in a field. This because they are accessed many times.

For example: look at the method `doLateInit()` in `AbstractEntityPersister` which is invoked by `postInstantiate()`. I think this could be spot to instantiate the array as well.

hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
JWGmeligMeyling: I'm not sure whether it can be allowed. The produced query results may be different for unconstrained inheritance hierarchies (a row that exists in a subclass table but not in a superclass table, which was previously omitted by the join but is now present in result sets). No blocker for me personally, and I can hardly believe that this will be a problem for anyone, but we seem to consider this very edge case at other locations so we may have to keep it in mind.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
JWGmeligMeyling: I don't really like the property name either. Your suggestion looks fine.

JWGmeligMeyling:
Can you add a test with classes C extends B extends A and then try to project a tuple with a property of C together with a property of A? Want to make sure the right joins are produced if B is omitted.

Or is there already a test for this?

ladislavovic:
> Can you add a test with classes C extends B extends A and then try to project a tuple with a property of C together with a property of A? Want to make sure the right joins are produced if B is omitted.

I added the test for that, see changes in `OmitAncestorJoinTest`.

> Same domain, but select a property from a secondarytable from B with the join to B itself being omitted.

The join to B itself will not be omitted. This issue optimize joining of superclasses tables only, it does not optimize secondary table joining. I added test for that, see `OmitAncestorJoinWhenSecondaryTablePresentTest`.

beikov:
Since this optimization is deactivatable in case we missed something, I think this should be safe to merge. Any objections to merge this for 5.4?

beikov:
@ladislavovic If you have the time, https://github.com/hibernate/hibernate-orm/pull/2482 provides a test case that uses secondary table joins that could/should be omitted with this optimization enabled. Do you want to take over that issue after we merge this? You seem to have an idea already how this could be done and it would be a really nice improvement!

gsmet:
Frankly, I'm not very excited about merging this to 5.4.

I really want 5.4 to be as risk free as possible so that people migrate to it as quickly as possible as it's the version that will have all the community fixes.

Note that this is particular because of the early product-freeze of 5.3.

beikov:
I understand that, but this optimization is deactivatable, so isn't this _risk free_ enough? When people try out 5.4 and they see something fails because of the optimization, they can easily disable it and we can get into touch with these people to fix the problems we would most probably not find anyway because we are not able to cover all possible models that are out there.

I might be wrong, but even if we communicate this optimization through the release announcements, I don't think many people will enable that optimization. On the other hand, if 5.4 doesn't work for them they get in touch and we immediatelly can help them resolve the issue, yet still get feedback about the model that we missed to cover.
I just don't see how else we could do it, even if we defer such optimizations to later releases, if the solution fails to cover cases, this might cause problems at some point anyway, so why defer it at all?
For which release would you target this PR then? Are you planning to do a 5.5 release? Considering how much the code base changed in 6.0, this PR will not work for 6.0, so this would have been wasted effort if we didn't integrate it in 5.x.

JWGmeligMeyling:
The problem is l that I am not even sure whether this change is spec compliant. Query outcomes will be different in some cases. A lot of other cases that are not unlikely broken by this feature remain untested. Some of my questions and issues raised still remained unanswered. This change needs to be extensively assessed imo, or disabled by default.

Rest assure, I use joined inheritance extensively and Id really like this feature to make it into Hibernate soon. And im certainly willing to contribute tests etc. But Im not looking forward to a release  require a couple hotfixes or even reverts that go unnoticed until it made it into the product.

I personally want to go to 5.4 as soon as possible as I have been running local snapshots since 5.2.12 and we need a stable release without any issues with the slightest doubt of stability.

ladislavovic:
>What it doesn't test is the following:
>- A secondary table associated with subclass B, with subclass C and superclass A.
>- This secondary table will be joined against the primary key of the subclass table for B.
>- This primary key will not be available, because you've just omitted the join to B, because only values from the secondary table and A are projected.

Please write the test yourself. I do not understand what exactly should the test do. Feel free to commit that into the branch, I think I let it open for project members.

JWGmeligMeyling:
I'll add the tests for the cases that I find important to test!

ladislavovic:
>I also like to see the above described test case, plus some tests using @Where and @Filter on various levels in the class hierarchy

When a filter is enabled the feature is switched off and all tables are joined. See `JoinProcessor.findQueryReferencedTables`

`@Where` could cause problems, but I try that and got interesting results. It always joined all tables in inheritance hierarchy. So probably the program goes by different way.

Moreover @Where cause error when you use it with inheritance strategy and join a subclass. And it is not related to tables omitting, it is reproducible also on master branch. The following test ends with error because generated SQL is incorrect. Generated SQL:
```
SELECT subalist1_.valsuba AS col_0_0_ 
FROM   b_table joinedinhe0_ 
       INNER JOIN suba_table subalist1_ 
               ON joinedinhe0_.id = subalist1_.b_id 
                  AND ( subalist1_.vala = 'foo' ) 
       INNER JOIN a_table subalist1_1_ 
               ON subalist1_.id = subalist1_1_.id 
```
The test:
```
package org.hibernate.test.joinwithoutancestor;

import java.util.List;
import javax.persistence.Entity;
import javax.persistence.GeneratedValue;
import javax.persistence.Id;
import javax.persistence.Inheritance;
import javax.persistence.InheritanceType;
import javax.persistence.ManyToOne;
import javax.persistence.OneToMany;
import javax.persistence.Table;

import org.hibernate.annotations.Where;

import org.hibernate.testing.TestForIssue;
import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
import org.hibernate.testing.transaction.TransactionUtil;
import org.junit.Test;

@TestForIssue(jiraKey = "HHH-12993")
public class JoinedInheritanceStrategyAndWhereAnnotation extends BaseCoreFunctionalTestCase {

	@Override
	protected Class[] getAnnotatedClasses() {
		return new Class[] { A.class, SubA.class, B.class };
	}

	@Test
	public void test() {
		TransactionUtil.doInHibernate( this::sessionFactory, session -> {
			String q = "select suba.valSubA from B b join b.subAList suba";
			session.createQuery( q ).getResultList();
		} );
	}

	@Entity(name = "A")
	@Table(name = A.TABLE)
	@Inheritance(strategy = InheritanceType.JOINED)
	static class A {

		public static final String TABLE = "A_Table";

		@Id
		@GeneratedValue
		private Long id;

		private String valA;

		public Long getId() {
			return id;
		}

		public String getValA() {
			return valA;
		}

		public void setValA(String valA) {
			this.valA = valA;
		}

	}

	@Entity(name = "SubA")
	@Table(name = SubA.TABLE)
	static class SubA extends A {

		public static final String TABLE = "SubA_Table";

		private String valSubA;

		@ManyToOne
		private B b;

		public String getValSubA() {
			return valSubA;
		}

		public void setValSubA(String valSubA) {
			this.valSubA = valSubA;
		}

		public B getB() {
			return b;
		}

		public void setB(B b) {
			this.b = b;
		}
	}

	@Entity(name = "B")
	@Table(name = B.TABLE)
	static class B {

		public static final String TABLE = "B_table";

		@Id
		@GeneratedValue
		private Long id;

		private String valB;

		@OneToMany(mappedBy = "b")
		@Where(clause = "valA = 'foo'")
		private List<SubA> subAList;

		public Long getId() {
			return id;
		}

		public String getValB() {
			return valB;
		}

		public void setValB(String valB) {
			this.valB = valB;
		}

		public List<SubA> getSubAList() {
			return subAList;
		}

		public void setSubAList(List<SubA> subAList) {
			this.subAList = subAList;
		}

	}

}
```

ladislavovic:
Guys I need to have the feature in Hibernate 5.4. It is important for my project. It can be disabled by default but I really need it there.

ladislavovic:
Guys, let's finish this issue somehow. It works, we do not have any test which fails due to that. If you think the risk is too high, make it disabled by default, mark it as experimental, whatever... But that would be a pity not to have this feature in Hibernate, it can speed up some kind of JPQL queries.

ladislavovic:
@beikov @gsmet @JWGmeligMeyling are you there?

beikov:
I'm sorry, but as far as I understood, the 5.4 branch won't receive any new features, only bug fixes. Correct me if I got that wrong @gsmet

ladislavovic:
Hi, will be the feature which is this pull reques about part of the Hibernate 6?

ladislavovic:
@sebersole @beikov Hi, will be the feature which is this pull reques about part of the Hibernate 6?

ladislavovic:
@sebersole @beikov Please can you answer my question? Thank you!

beikov:
Sorry, but I can't answer this question. I think this kind of optimization will most likely happen in Hibernate 6. If we have the commitment from the core team, I can work on this, as I would like to see this in 5.x as well, but without that, there is not much we can do with this PR unfortunately.

Sanne:
So I can't comment on specifics of this issue as I hadn't noticed this so far, but we could reasses our intention of not merging any feature improvement yet.

We were trying to avoid to create more work on the 5.x branches so that the core team could focus on ORM6, but since that was delayed a bit I guess that as long as people help by doing most of the work we could show some flexibility - that's just my suggestion though, would need @sebersole 's approval.

@beikov if you were to want to volunteer for this would it benefit ORM 6 too? It's probably not a good idea to add a feature in 5 which would then create more work to the 6 team to have feature parity when it's released. (While I guess we never committed to feature parity, it would be odd to include a new feature here and then have a new major which takes it out again)

beikov:
This is an optimization, not a new feature ;)

Sanne:
@beikov I agree :) But above you said

> I'm sorry, but as far as I understood, the 5.4 branch won't receive any new features, [...]

Whatever we call this, I just mean that I think we could be flexibile a bit more flexible.

beikov:
Alright, when steve is on board with this then, I'll work on it. I was of the impression that 5.x will only receive bug fixes from that point on and no further optimizations.

Sanne:
you're right that was the intention. But we drifted already, in fact I'm working on several other strong optimisations which are going to appear in 5.4+ versions.

sebersole:
If it applies cleanly then +1.  

As far as conflicts with 6.0, the only change that should have an impact is `Joinable` and that has to change already so it would be safe to just ignore that.  

The PR does change the formatting incorrectly in some of the HQLTest lines but that's minor.

sebersole:
Sorry for the delayed response.  I've been very "head down" on 6.0 work

sebersole:
I agree with what @Sanne mentions wrt 6.0 delay and getting useful things into 5.4 if possible (avoiding regressions ofc).  I went ahead, cleaned this up and applied it.

JWGmeligMeyling:
Unfortunately this does seem to cause a couple of regressions as pointed out in :
* https://hibernate.atlassian.net/browse/HHH-13670
* https://hibernate.atlassian.net/browse/HHH-1631?focusedCommentId=105647&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-105647

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -94,9 +103,86 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+
+	private <T extends AST> void doFindAllNodes(AST node, Class<T> clazz, List<T> found) {
+		if ( clazz.isAssignableFrom( node.getClass() ) ) {
+			found.add( (T) node );
+		}
+		if ( node.getFirstChild() != null ) {
+			doFindAllNodes( node.getFirstChild(), clazz, found );
+		}
+		if ( node.getNextSibling() != null ) {
+			doFindAllNodes( node.getNextSibling(), clazz, found );
```
jwgmeligmeyling: @beikov This is the code that caused the `StackOverflowError` for @heruan . The code seems to be fine under the assumption of a tree structure. I'm not sure in what scenario the AST could possibly lose this tree property... Does this ring a bell for you? Perhaps the query was just fairly complex and the stack relatively limited for a recursive implementation to work on the number of nodes in the tree. Perhaps then we should rewrite the implementation to do a queue based BFS instead.


```
java.lang.StackOverflowError: null
    at org.hibernate.hql.internal.ast.util.JoinProcessor.doFindAllNodes(JoinProcessor.java:113) ~[hibernate-core-5.4.6.Final.jar:5.4.6.Final]
    at org.hibernate.hql.internal.ast.util.JoinProcessor.doFindAllNodes(JoinProcessor.java:120) ~[hibernate-core-5.4.6.Final.jar:5.4.6.Final]
    at org.hibernate.hql.internal.ast.util.JoinProcessor.doFindAllNodes(JoinProcessor.java:120) ~[hibernate-core-5.4.6.Final.jar:5.4.6.Final]
    at org.hibernate.hql.internal.ast.util.JoinProcessor.doFindAllNodes(JoinProcessor.java:120) ~[hibernate-core-5.4.6.Final.jar:5.4.6.Final]
    at org.hibernate.hql.internal.ast.util.JoinProcessor.doFindAllNodes(JoinProcessor.java:120) ~[hibernate-core-5.4.6.Final.jar:5.4.6.Final]
```

PR #2547, opened 2018-09-28
HHH-12997 - @OneToMany does not work with @JoinFormula

dominikbielinski:
A test that shows that combining @JoinFormula with @OneToMany fails during startup.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/annotations/onetomany/OneToManyJoinFormulaTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/annotations/onetomany/OneToManyJoinFormulaTest.java
@@ -1,0 +1,103 @@
+package org.hibernate.test.annotations.onetomany;
+
+import org.hibernate.annotations.JoinFormula;
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.jpa.test.BaseEntityManagerFunctionalTestCase;
+import org.hibernate.testing.TestForIssue;
+import org.junit.Before;
+import org.junit.Test;
+
+import javax.persistence.*;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
+import static org.junit.Assert.assertEquals;
+
+@TestForIssue( jiraKey = "HHH-12997" )
+public class OneToManyJoinFormulaTest
+		extends BaseEntityManagerFunctionalTestCase {
+
+	@Override
+	protected Class<?>[] getAnnotatedClasses() {
+		return new Class<?>[] {
+				Invoice.class
+		};
+	}
+
+	@Override
+	protected void addConfigOptions(Map options) {
+		options.put( AvailableSettings.GLOBALLY_QUOTED_IDENTIFIERS, Boolean.TRUE );
+	}
+
+	@Before
+	public void setUp() {
+		doInJPA( this::entityManagerFactory, entityManager -> {
+			Invoice rootInvoice = new Invoice();
+			rootInvoice.setId(1L);
+			rootInvoice.setHierarchyPath("1/");
+			entityManager.persist(rootInvoice);
+
+			Invoice childInvoice1 = new Invoice();
+			childInvoice1.setId(2L);
+			childInvoice1.setHierarchyPath("1/2/");
+			entityManager.persist(childInvoice1);
+
+			Invoice childInvoice2 = new Invoice();
+			childInvoice2.setId(3L);
+			childInvoice2.setHierarchyPath("1/3/");
+			entityManager.persist(childInvoice2);
+		});
+	}
+
+	@Test
+	public void testBulkUpdate() {
+		doInJPA( this::entityManagerFactory, entityManager -> {
+			Invoice rootInvoice = entityManager.createQuery(
+					"select inv from Invoice inv " +
+					"left join fetch inv.childInvoices " +
+					"where inv.id = 1", Invoice.class).getSingleResult();
+
+			assertEquals(2, rootInvoice.getChildInvoices().size());
+		});
+	}
+
+	@Entity(name = "Invoice")
+	public static class Invoice {
+
+		@Id
+		private Long id;
+
+		@Column(name = "hierarchy_path")
+		private String hierarchyPath;
+
+		@OneToMany(fetch = FetchType.LAZY)
+		@JoinFormula(value = "hierarchy_path like id || '/_%'", referencedColumnName = "hierarchy_path")
+		private Set<Invoice> childInvoices = new HashSet<>();
+
+		public String getHierarchyPath() {
+			return hierarchyPath;
+		}
+
+		public void setHierarchyPath(String hierarchyPath) {
+			this.hierarchyPath = hierarchyPath;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/test/java/org/hibernate/test/annotations/onetomany/OneToManyJoinFormulaTest.java
```diff
@@ -0,0 +1,103 @@
...
+	public static class Invoice {
+
+		@Id
+		private Long id;
+
+		@Column(name = "hierarchy_path")
+		private String hierarchyPath;
+
+		@OneToMany(fetch = FetchType.LAZY)
+		@JoinFormula(value = "hierarchy_path like id || '/_%'", referencedColumnName = "hierarchy_path")
```
JWGmeligMeyling: Shouldn't you annotate the inverse relation? I.e. `private Invoice parentInvoice`? Not sure whether even `JoinColumn` is allowed here.

dominikbielinski: I tried that as well:
```
@JoinFormula(value = "SUBSTR(hierarchy_path, 0, INSTR(hierarchy_path, '/') - 1)", referencedColumnName = "id")
private Invoice rootInvoice;

@OneToMany(fetch = FetchType.LAZY, mappedBy = "rootInvoice")
private Set<Invoice> childInvoices;
```
but there is same exception as described here https://discourse.hibernate.org/t/how-to-use-the-hibernate-joinformula-annotation-with-a-onetomany-jpa-association/1461 (i.e. ManyToOne works with JoinFormula, but OneToMany doesn't - even when "mappedBy" is added and JoinFormula is on ManyToOne side). 
I guess the question is just whether JoinFormula works with OneToMany - but if it doesn't, how is anyone supposed to know that?

JWGmeligMeyling: There is an existing test with a `@JoinFormula @ManyToOne` relation here: https://github.com/hibernate/hibernate-orm/blob/e5dc635a52362f69b69acb8d5b166b69b165dbbd/hibernate-core/src/test/java/org/hibernate/test/annotations/manytoonewithformula/ManyToOneWithFormulaTest.java . So what you're saying is that when this example is extended with an inverse mapping (a `@OneToMany(mappedBy = "")`, Hibernate fails to boot?

dominikbielinski: Yes, that's exactly what I was trying to say. For instance, in the test that you mentioned, if you add following to Language entity: 
```
private List<Message> messages;

@OneToMany(fetch = FetchType.LAZY, mappedBy = "language")
public List<Message> getMessages() {
    return messages;
}

public void setMessages(List<Message> messages) {
    this.messages = messages;
}
```
the test fails with exception
`Caused by: java.lang.ClassCastException: org.hibernate.mapping.Formula cannot be cast to org.hibernate.mapping.Column`

gsmet:
@vladmihalcea I would be interested in having your feedback on that.

While I'm not sure this particular test case is correct, @JWGmeligMeyling has a point here: https://github.com/hibernate/hibernate-orm/pull/2547#discussion_r221451368 e.g. that a test does not work when we just add an inverse relationship.

I'm wondering if it's even possible to support an inverse relationship with `@JoinFormula`.

WDYT?

vladmihalcea:
@gsmet @JWGmeligMeyling I think it's a valid use case and it would be convenient to add support for it.

gavinking:
Here's an updated test which fails on H7.

```java
package org.hibernate.userguide.associations;

import org.hibernate.testing.orm.junit.JiraKey;
import org.junit.Before;
import org.junit.Test;

import jakarta.persistence.CascadeType;
import jakarta.persistence.Entity;
import jakarta.persistence.FetchType;
import jakarta.persistence.GeneratedValue;
import jakarta.persistence.GenerationType;
import jakarta.persistence.Id;
import jakarta.persistence.JoinColumn;
import jakarta.persistence.ManyToOne;
import jakarta.persistence.OneToMany;
import java.util.ArrayList;
import java.util.List;

import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;

/**
 * @author Lisandro Fernandez (kelechul at gmail dot com)
 */
@JiraKey("HHH-13815")
public class BidirectionalOneToManyMergeTest extends org.hibernate.orm.test.jpa.BaseEntityManagerFunctionalTestCase {

	@Override
	protected Class<?>[] getAnnotatedClasses() {
		return new Class<?>[]{
				Post.class,
				PostComment.class,
		};
	}

	@Before
	public void setUp() {
		doInJPA(this::entityManagerFactory, entityManager -> {
			entityManager.persist(
					new Post("High-Performance Java Persistence").setId(1L)
			);
		});
	}

	@Test
	public void testFailMerge() {
		doInJPA(this::entityManagerFactory, entityManager -> {
			Post post = entityManager.find(Post.class, 1L);
			post.addComment(new PostComment("This post rocks!", post));
			post.getComments().isEmpty(); // this fails
			entityManager.merge(post);
		});
	}

	@Entity
	public static class Post {

		@Id
		private Long id;

		private String title;

		@OneToMany(mappedBy = "post", cascade = CascadeType.ALL, orphanRemoval = true)
		private List<PostComment> comments = new ArrayList<>();

		public Post() {
		}

		public Post(String title) {
			this.title = title;
		}

		public Long getId() {
			return id;
		}

		public Post setId(Long id) {
			this.id = id;
			return this;
		}

		public String getTitle() {
			return title;
		}

		public Post setTitle(String title) {
			this.title = title;
			return this;
		}

		public List<PostComment> getComments() {
			return comments;
		}

		private Post setComments(List<PostComment> comments) {
			this.comments = comments;
			return this;
		}

		public Post addComment(PostComment comment) {
			comments.add(comment);
			comment.setPost(this);

			return this;
		}

		public Post removeComment(PostComment comment) {
			comments.remove(comment);
			comment.setPost(null);

			return this;
		}
	}

	@Entity
	public static class PostComment {

		@Id
		@GeneratedValue(strategy = GenerationType.IDENTITY)
		private Long id;

		private String review;

		@JoinColumn(name = "post_id")
		private Post post;

		public PostComment() {
		}

		public PostComment(String review, Post post) {
			this.review = review;
			this.post = post;
		}

		public Long getId() {
			return id;
		}

		public PostComment setId(Long id) {
			this.id = id;
			return this;
		}

		public String getReview() {
			return review;
		}

		public PostComment setReview(String review) {
			this.review = review;
			return this;
		}

		public Post getPost() {
			return post;
		}

		public PostComment setPost(Post post) {
			this.post = post;
			return this;
		}

		@Override
		public boolean equals(Object o) {
			if (this == o) return true;
			if (!(o instanceof PostComment)) return false;
			return id != null && id.equals(((PostComment) o).getId());
		}

		@Override
		public int hashCode() {
			return 31;
		}
	}
}
```

We should fix this; it looks fairly straightforward. (Just disable the exception when processing orphan delete.)

gavinking:
I'm not sure we will ever implement this feature, but we should have a meaningful error, hence #9320.

PR #2548, opened 2018-09-28
HHH-12995: Querying DECIMAL columns via Double fields can lead to pre…

breglerj:
…cision loss on SAP HANA

- add new configuration parameter hibernate.dialect.hana.treat_double_typed_fields_as_decimal


vladmihalcea:
Applied upstream, thanks

PR #2551, opened 2018-09-30
 HHH-13000 - Fix for refreshing entities under PESSIMISTIC_WRITE lock

JWGmeligMeyling:
Testcase provided separately in #2550 

Issue: https://hibernate.atlassian.net/browse/HHH-13000

PR #2552, opened 2018-10-01
HHH-13000 - Fix for refreshing entities under PESSIMISTIC_WRITE lock

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13000

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
@@ -183,7 +183,9 @@ public class DefaultRefreshEventListener implements RefreshEventListener {
 				// the requested lock-mode is less restrictive than the current one
 				//		- pass along the current lock-mode (after accounting for WRITE)
 				lockOptionsToUse = LockOptions.copy( event.getLockOptions(), new LockOptions() );
-				if ( currentLockMode == LockMode.WRITE ) {
+				if ( currentLockMode == LockMode.WRITE ||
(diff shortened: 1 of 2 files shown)


vladmihalcea:
@JWGmeligMeyling @beikov I created this PR to replace [the original one](https://github.com/hibernate/hibernate-orm/pull/2551). The same logic should apply to `PESSIMISTIC_READ` which, on DBs that don't really support this mode, Hibernate falls back to `PESSIMISTIC_WRITE`.

If you agree with this commit, I can integrate it after your review.

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultRefreshEventListener.java
```diff
@@ -194,7 +196,7 @@ public void onRefresh(RefreshEvent event, Map refreshedAlready) {
 					lockOptionsToUse.setLockMode( LockMode.READ );
 
 					// and prepare to reset the entry lock-mode to WRITE after the refresh completes
```
gsmet: Let's update the comment. The lock mode won't always be WRITE.

gsmet:
@vladmihalcea It makes sense. Please update the comment before merging, thanks!

gsmet:
Fixed the comment and merged. Thanks all!

JWGmeligMeyling:
Thanks for merging! Is this a candidate for backport too? Or won't there be a 5.3 release anymore?

gsmet:
No plan to backport it to 5.3. It’s now in a product so the bar to get into 5.3 is very high. 

That’s why we’re working on releasing a 5.4 soon so that we can get all these fixes released.

PR #2553, opened 2018-10-02
Fix tests on SAP HANA

breglerj:
- add optional row/column modifier to the expected create schema statement
  - org.hibernate.test.schemaupdate.AlterTableQuoteDefaultSchemaTest
  - org.hibernate.test.schemaupdate.AlterTableQuoteSpecifiedSchemaTest

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/schemaupdate/AlterTableQuoteDefaultSchemaTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/schemaupdate/AlterTableQuoteDefaultSchemaTest.java
@@ -125,7 +125,7 @@ public class AlterTableQuoteDefaultSchemaTest extends AbstractAlterTableQuoteSch
 			String fileContent = new String( Files.readAllBytes( output.toPath() ) );
 
 			Pattern fileContentPattern = Pattern
-					.compile( "create table " + regexpQuote( "default-schema", "my_entity" ) );
+					.compile( "create ((column|row) )?table " + regexpQuote( "default-schema", "my_entity" ) );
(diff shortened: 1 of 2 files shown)


gsmet:
@vladmihalcea didn't we decide to only enable these tests for some named dialects? It seems that you have committed the version where it's enabled for all dialects supporting schema creation?

vladmihalcea:
@gsmet I think it's better to have the tests enabled and find out if there are problems with certain DBs because of the latest Hibernate code changes.

vladmihalcea:
@breglerj When I try to run the test on HANA, I still get this:

> 17:59:54,952 ERROR SqlExceptionHelper:142 - SAP DBTech JDBC: [386]: cannot use duplicate schema name: default-schema: line 1 col 15 (at pos 14)

How is it working for you?

breglerj:
@vladmihalcea It's most likely a permission issue. 

A schema named "default-schema" created by the Jenkins test DB user still exists, because the test cleanup doesn't work without specifying `CASCADE` in the drop statement or deleting the table first. 

When you try to execute the test with a different user, that user is not allowed to drop the existing schema and consequently the following recreation of the schema fails.

Executing the test as the Jenkins test DB user should work.

I've pushed a new commit that adds an explicit drop of the entity table. That should fix the test failure for you.

vladmihalcea:
I [restricted the tests for H2, PostgreSQL and SQL Server](https://github.com/hibernate/hibernate-orm/commit/c3febcaaedfd3aab228234307fd41d8a5a2d955a) as otherwise, the test would have become way too complicated and hinder long-term maintenance.

gsmet:
Thanks Vlad, I think that's best.

PR #2556, opened 2018-10-02
HHH-13007 No longer use net.bytebuddy.experimental=true when testing …

Sanne:
…on JDK11

https://hibernate.atlassian.net/browse/HHH-13007


gsmet:
Good catch. Merged.

PR #2558, opened 2018-10-03
Hhh 12713

sebersole:
Changes to the original PR - #2358


sebersole:
Good point @dreab8.  Initially that class really was a parameter object that then got passed elsewhere.  I obviously changed that function, but never changed the name.  Thanks for the point.

@Naros, @aleksandarsusnjar, @gsmet - any thoughts?

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
```diff
@@ -183,18 +187,56 @@ public void disableFetchProfile(String name) throws UnknownProfileException {
...
+		final EntityGraph fetchHint = (EntityGraph) properties.get( EntityGraphSemantic.FETCH.getJpaHintName() );
+		if ( fetchHint != null ) {
+			this.entityGraph = fetchHint;
+			this.entityGraphSemantic = EntityGraphSemantic.FETCH;
+		}
+		else {
+			final EntityGraph loadHint = (EntityGraph) properties.get( EntityGraphSemantic.LOAD.getJpaHintName() );
+			if ( loadHint != null ) {
+				this.entityGraph = loadHint;
+				this.entityGraphSemantic = EntityGraphSemantic.FETCH;
```
gsmet: Shouldn't it be `LOAD`?

hibernate-core/src/main/java/org/hibernate/graph/CannotBecomeEntityGraphException.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.graph;
+
+import org.hibernate.HibernateException;
+
+/**
+ * Indicates an attempt to make an {@link Graph} become
```
gsmet: s/make an/make a/

hibernate-core/src/main/java/org/hibernate/graph/EntityGraphParser.java
```diff
@@ -0,0 +1,257 @@
...
+import org.hibernate.Session;
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.graph.internal.RootGraphImpl;
+import org.hibernate.graph.internal.parse.GraphParser;
+import org.hibernate.graph.spi.GraphImplementor;
+import org.hibernate.graph.spi.RootGraphImplementor;
+
+/**
+ * todo (graph) : to me most of this seems more appropriate in documentation as opposed to here
```
gsmet: Is it still the case? In the documentation, you point to this exact class to get more information.

hibernate-core/src/main/java/org/hibernate/graph/EntityGraphs.java
```diff
@@ -0,0 +1,213 @@
...
+	 * The JPA-defined HINT key for specifying an EntityGraph to use applying
+	 * fetch semantics
+	 */
+	public static volatile String HINT_FETCHGRAPH = EntityGraphSemantic.FETCH.getJpaHintName();
+
+	/**
+	 * The JPA-defined HINT key for specifying an EntityGraph to use applying
+	 * load semantics
+	 */
+	public static volatile String HINT_LOADGRAPH = EntityGraphSemantic.FETCH.getJpaHintName();
```
gsmet: Should the two elements above be `final` instead of `volatile`?

hibernate-core/src/main/java/org/hibernate/graph/internal/Helper.java
```diff
@@ -0,0 +1,85 @@
...
+import org.hibernate.metamodel.model.domain.spi.IdentifiableTypeImplementor;
+import org.hibernate.metamodel.model.domain.spi.MapAttributeImplementor;
+import org.hibernate.metamodel.model.domain.spi.PluralAttributeImplementor;
+import org.hibernate.metamodel.model.domain.spi.SimpleTypeImplementor;
+import org.hibernate.metamodel.model.domain.spi.SingularAttributeImplementor;
+
+/**
+ * @author Steve Ebersole
+ */
+public class Helper {
```
gsmet: Do we really want a public class with such a generic name?

hibernate-core/src/main/java/org/hibernate/Incubating.java
```diff
@@ -15,6 +15,12 @@
 import static java.lang.annotation.RetentionPolicy.CLASS;
 
 /**
+ * Marks certain of Hibernate's packages (recursively), types and methods
+ * as incubating.  Incubating indicates a type or method that is still being
+ * actively developed.  Users of these types and methods should are
```
aleksandarsusnjar: "should are"

hibernate-core/src/main/antlr/graph.g
```diff
@@ -0,0 +1,136 @@
...
+			startAttribute( path );
+		}
+		else {
+			startQualifiedAttribute( path, qualifier );
+		}
+	}
+	;
+
+subGraph
+	: LPAREN (subtype:NAME COLON)? { startSubGraph( subtype ); } attributeNode (COMMA attributeNode )* RPAREN {
```
aleksandarsusnjar: Can 'subGraph' not leverage 'graph' (after 'subtype'). Ideally they should have exactly the same definition and be recursive that way.

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
```diff
@@ -183,18 +187,24 @@ public void disableFetchProfile(String name) throws UnknownProfileException {
...
 	public EntityGraph getFetchGraph() {
-		return fetchGraph;
+		return entityGraphSemantic == EntityGraphSemantic.FETCH
+				? entityGraph
+				: null;
 	}
 
 	public void setFetchGraph(final EntityGraph fetchGraph) {
-		this.fetchGraph = fetchGraph;
+		this.entityGraphSemantic = EntityGraphSemantic.FETCH;
```
aleksandarsusnjar: To an outsider setFetchGraph and setLoadGraph may sound independent but we know they aren't now. The effect is also visible via the getters. "Why did A disappear if I set B". Would it not be more "honest" and clear to just accept this way all the say and have just a "set[Entity]Graph" that takes two parameters - EntityGraph and EntityGraphSemantic and a getter for each - get[Entity]Graph and get[Entity]GraphSemantic?

Alternatively maybe differently named setters could work. Something like "set**As**FetchGraph" and "set**As**LoadGraph" but getters don't benefit from that.

aleksandarsusnjar: To an outsider setFetchGraph and setLoadGraph may sound independent but we know they aren't now. The effect is also visible via the getters. "Why did A disappear if I set B". Would it not be more "honest" and clear to just accept this way all the say and have just a "set[Entity]Graph" that takes two parameters - EntityGraph and EntityGraphSemantic and a getter for each - get[Entity]Graph and get[Entity]GraphSemantic?

Alternatively setters could be renamed to something else, such as "set**As**[Fetch|Load]Graph" but this does not help with getters.

hibernate-core/src/main/java/org/hibernate/graph/EntityGraphSemantic.java
```diff
@@ -0,0 +1,67 @@
...
+	 * subsequent select).  Attributes that are not specified are treated as
+	 * FetchType.LAZY or FetchType.EAGER depending on the attribute's definition
+	 * in metadata
+	 */
+	LOAD( "javax.persistence.loadgraph" ),
+
+	/**
+	 * No graph
+	 */
+	NONE( null );
```
aleksandarsusnjar: Not sure how this helps. Isn't the deciding factor the existence of the graph itself. If the graph is not null but this is set to NONE what would that mean?

hibernate-core/src/main/java/org/hibernate/graph/EffectiveEntityGraph.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.hibernate.graph;
+
+/**
+ * Encapsulates the entity graph and its "type" (LOAD/FETCH)
+ *
+ * @author Steve Ebersole
+ */
+public class EffectiveEntityGraph {
+	enum GraphType {
```
aleksandarsusnjar: Isn't this now duplicating EntityGraphSemantic?

hibernate-core/src/main/java/org/hibernate/graph/internal/AttributeNodeImpl.java
```diff
@@ -0,0 +1,327 @@
...
+	protected <S extends J> void internalAddKeySubGraph(Class<S> subType, SubGraph<S> subGraph) {
+		log.tracef( "Adding key sub-graph : ( (%s) %s )", subType.getName(), getAttributeName() );
+
+		if ( keySubGraphMap == null ) {
+			keySubGraphMap = new HashMap<>();
+		}
+
+		final SubGraphImplementor<? extends J> previous = keySubGraphMap.put( subType, (SubGraphImplementor) subGraph );
+		if ( previous != null ) {
+			log.debugf( "Adding key sub-graph [%s] over-wrote existing [%]", subGraph, previous );
```
aleksandarsusnjar: I don't remember what was in the code I submitted but this should really never happen. It should be handled somewhere but, when adding graph specifications they should be merged. If this is meant to be handled outside then this is an unexpected state probably warranting an exception. If it is not meant to be handled outside then it should be handled here.

hibernate-core/src/main/java/org/hibernate/graph/internal/parse/GraphParser.java
```diff
@@ -0,0 +1,167 @@
...
+		final AttributeNodeImplementor<?> attributeNode = resolveAttributeNode( attributeName );
+		attributeNodeStack.push( attributeNode );
+
+		final PathQualifierType pathQualifierType = resolvePathQualifier( qualifierName );
+
+		graphSourceStack.push( pathQualifierType.getSubGraphCreator() );
+	}
+
+	private PathQualifierType resolvePathQualifier(String qualifier) {
+		if ( "key".equalsIgnoreCase( qualifier ) ) {
```
aleksandarsusnjar: Why ignoreCase?

gsmet:
On a more general note:
 * there are a few conflicts with current master;
 * I must admit I'm a bit worried by all these new metamodel related contracts, especially about the difficulty of backporting fixes from 5.4 to 5.3. I envisioned 5.4 as a very conservative release with more ability to push fixes, compared to 5.3. But maybe it's not parts we touch often?

hibernate-core/src/main/java/org/hibernate/graph/EffectiveEntityGraph.java
sebersole: WRT `GraphType` vs `EntityGraphSemantic` yes.  However, I left `EffectiveEntityGraph` in place because I am not yet sure how to best handle this.  IMO `EffectiveEntityGraph` conveys this "pairing" much better that we do currently in `LoadQueryInfluencers`, but using `EffectiveEntityGraph` on `LoadQueryInfluencers` means needing a new instantiation of EffectiveEntityGraph to wrap the EntityGraph every time a new graph is pushed to the `LoadQueryInfluencers`.  So still thinking through this one.

hibernate-core/src/main/java/org/hibernate/graph/EntityGraphSemantic.java
sebersole: Nit-pick much lol.  How does it hurt?  And I thought it helped.  Is there a real reason to remove it?

hibernate-core/src/main/java/org/hibernate/graph/internal/Helper.java
sebersole: internal is not a "public class"

hibernate-core/src/main/java/org/hibernate/graph/internal/parse/GraphParser.java
aleksandarsusnjar: 1. Opens up freedom that would have to be maintained forever after :)
2. Java is case sensitive, not case insensitive. key, KEY, kEy, ... are all different.
3. Java calls these precisely as you spelled them out - lowercase.
4. Should not be an issue here but technically "ignoreCase" is locale-specific and can yield different results in different locales. I am not aware of characters in "key" and "value" that would suffer from this but I am aware of others. For example, `"i".equalsIgnoreCase("I")` is not always `true`.
5. Opens up variation that people will take, resulting in differing styles.
6. Although extremely unlikely, at some point in time in the future there may pop up a need for "KEY" or "Key" or similar with a different meaning but it would have been taken.
7. Completely unimportant but for the sake of foolish completeness, it is a bit slower and you took longer to type it :)

hibernate-core/src/main/java/org/hibernate/graph/EntityGraphs.java
sebersole: Perhaps.  Not to mention HINT_LOADGRAPH probably ought to use EntityGraphSemantic.LOAD :D.  Lot of copy/paste errors still.  This is just an initial preview ;)

hibernate-core/src/main/java/org/hibernate/graph/EntityGraphParser.java
sebersole: Well, as a user, would you rather:

1. Read the user guide and understand how to write these "entity graph language" strings?
2. Read the user guide and and be directed to instead go read some (not even linked) Javadocs to understand how to write these "entity graph language" strings?

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
sebersole: Yep, like I said earlier - still lots of copy/paste things.  In general this is ongoing work, just wanted to get feedback on the overall parts

hibernate-core/src/main/java/org/hibernate/graph/internal/Helper.java
sebersole: Note - I am not saying the class name is great, just pointing out that there is a difference between "public" as in:

1. Java keyword
2. What we support applications using

This does not fit the later - that's why it is in an internal package

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
sebersole: As I mentioned earlier on another of your comments, I agree.  Ideally this would use `EffectiveEntityGraph`.  But see my other comment wrt concerns there

hibernate-core/src/main/antlr/graph.g
sebersole: Well first it is recursive.  `subGraph` can repeats recursively.  The problem with what you suggest is that recognition of `graph` and `subGraph` have different action (this is called syntax versus semantics in parsing).

Also, I had considered extending this to possibly allow complete definition of an EntityGraph.  E.g.:  `(MyEntity: attr1, attr2, ..., attrN)` defines a root graph for `MyEntity`).  So again the syntax is more-or-less the same, but very different semantics.

Really the only "constant" part is the `attributeNode (COMMA attributeNode)*` part.  We *could* refactor that, but honestly, what do you gain?

hibernate-core/src/main/java/org/hibernate/graph/internal/Helper.java
gsmet: Yeah, I was referring as `public` in the Java sense e.g. a class that will be visible from all the other classes when I develop in ORM.

`GraphHelper` or `EntityGraphHelper` would be better IMHO.

hibernate-core/src/main/java/org/hibernate/graph/internal/parse/GraphParser.java
gsmet: If it's supposed to be some sort of keyword, I think I would make it case sensitive too.

hibernate-core/src/main/java/org/hibernate/graph/EntityGraphParser.java
gsmet: I would prefer having the information in the documentation. It's just that it was not clear the patch was a preview and you wanted to do that before the final inclusion so my remark was more: either we fix it or we remove the todo as it's a bit unintuitive to point the user there and start with a todo saying we should include it in the doc :).

hibernate-core/src/main/java/org/hibernate/graph/internal/parse/GraphParser.java
sebersole: 1. Um, ok.  Its not any different than we do for HQL or (I think) 
2. 

> 1. Opens up freedom that would have to be maintained forever after :)

Um, ok :) Its not any different than we do for HQL or (I think) other "paths" that end in key, index, indices, element, elements, values...  They all accept them case insensitively.

> 2. Java is case sensitive, not case insensitive. key, KEY, kEy, ... are all different.

This depends how you view "key".  You interpret it as an identifier.  And yes Java identifiers are case-sensitive.  However, JPA really defines this as a keyword (look at its keyword list) and as such it ought to be recognized case-insensitively.

> 3. Java calls these precisely as you spelled them out - lowercase.

Well, *Java* does not define "key" as anything.  You mean JPA?  Right, we should never support anything beyond JPA.  Even stuff like parsing "entity graph strings" ;)  Plus, see (2)


> 4. Should not be an issue here but technically "ignoreCase" is locale-specific and can yield different results in different locales. I am not aware of characters in "key" and "value" that would suffer from this but I am aware of others. For example, `"i".equalsIgnoreCase("I")` is not always `true`.

This is not true.  equalsIgnoreCase performs locale-independent checking

> 5. Opens up variation that people will take, resulting in differing styles.

So consistently just like HQL, etc?  I'll take it! :)

> 6. Although extremely unlikely, at some point in time in the future there may pop up a need for "KEY" or "Key" or similar with a different meaning but it would have been taken.

^^ reach ;)

> 7. Completely unimportant but for the sake of foolish completeness, it is a bit slower and you took longer to type it :)

Its slower because it does TheRightThing wrt case-sensitivity.

sebersole:
> * there are a few conflicts with current master;

Yep, I have not rebased yet.  Not sure why that is a requirement for getting people to look at some work early ;)

> * I must admit I'm a bit worried by all these new metamodel related contracts, especially about the difficulty of backporting fixes from 5.4 to 5.3. I envisioned 5.4 as a very conservative release with more ability to push fixes, compared to 5.3. But maybe it's not parts we touch often?

Well our *internal* JPA type system impls have always been internal/incubating.  So not sure why this is a concern.

hibernate-core/src/main/java/org/hibernate/graph/internal/Helper.java
sebersole: You mean because internal classes only ever get used from the package they are defined in? ;)

Look, I already said the name could be improved.  In early work a name (*especially* for an internal utility class) is utterly unimportant to me.  `GraphHelper` - great.

But I utterly disagree with this "it should be private / package-protected" stance.

hibernate-core/src/main/java/org/hibernate/graph/internal/parse/GraphParser.java
sebersole: > If it's supposed to be some sort of keyword, I think I would make it case sensitive too.

"select" is also a keyword in JPA.  So you disagree with JPA that "SELECT", "select", "SeLeCt", .. should all be the same?  I do not agree with that.  KEY is another JPA keyword - they should all behave consistently.

hibernate-core/src/main/java/org/hibernate/graph/EntityGraphParser.java
sebersole: Sure, but that's kind of the point of todos ;)

hibernate-core/src/main/java/org/hibernate/graph/internal/parse/GraphParser.java
sebersole: BTW - section _4.4.1 Identifiers_ is the one that details JPA's key/reserved words and its case-insensitive expectations for their use.

> The following are reserved identifiers: ..., KEY, ... Reserved identifiers are case insensitive.

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
sebersole: So to me, this is the biggest outstanding question.  How do we track the "currently effective EntityGraph" wrt the Session?  

1. Keep `LoadQueryInfluencers` as it is more-or-less?
2. Use `EffectiveEntityGraph` in `LoadQueryInfluencers`?
3. Another option?

What do y'all think about that?

aleksandarsusnjar: Just to make it clear w.r.t. language used, I do not think that load/fetch graphs should be at session granularity but only specific to the query. I also think that fetch/load graph distinction is rather unfortunate and causes less than desirable practices. The situation is what it is, though, and we have to deal with it.

A central question to this is can there be both a load and fetch graph at the same time. As discussed before, I don't favour such a case, but JP seems unclear about this so my original code allowed it, leaving it to the implementation to decide and untie the knot one way or another. However, let's consider a few possibilities:

**1. We need to support the case when both fetch and load graphs are specified**
Unlikely. But, if so, we should expose the API that allows this cleanly. If present implementation does not support this, as any good API, it should fail fast. For example, calling a setter of another after one has been already set to non-null could throw an Unsupported...Exception or IllegalStateException or something. If the implementation already supports this we need to consider whether that is desirable and then come back to this consideration.

**2. Presently we do not need both at the same time but we may have to in the future**
The API needs to be made in a way that discourages even a possibility of doing so now but also in a way that could be modified later to allow it.

**3. We don't and will never need both at the same time**
The API should be clear about this at first glance such that there can be no confusion even if someone does not read documentation. This may be subjective, but to me that either means a single setter (taking two arguments, so not a conventional setter) or method(s) that do(es) not look at all like setters, forcing the developer to read the documentation. Since "fetch" and "load" really do not at all say how they differ it would be nice to use better names too. FETCH/LOAD enums could be hidden from public completely. Intentionally verbose examples:

`overrideNeededAttributes(...)`

`addNeededAttributes(...)`

Note that "add" here is slightly different and implies a merge if called multiple times. Brainstorming: resetNeededAttributes() and resetNeededAttributes(EntityGraph additional), needDefaultAttributes(), needDefaultAttributesPlusAdditional(EntityGraph), ...

aleksandarsusnjar: Re (1) and (2) I agree about their (lack of) usefulness. JPA certainly does not (can not) prevent setting both and I haven't seen any text explicitly stating what the two of us agree about. However, let's not waste any more time on that.

With respect to standard JPA terms being used for this I can say that I do agree that this is indeed a fact, albeit an unfortunate and very poorly documented one, as noted above. Words "fetch" and "load" really mean the same thing. Some documentation can be found here:

https://docs.oracle.com/javaee/7/tutorial/persistence-entitygraphs001.htm

Quotes, differences emphasized:

- A **fetch** graph consists of **only** the fields explicitly specified in the EntityGraph instance, **and ignores the** default entity graph settings.
- A **load** graph consists of the fields explicitly specified in the EntityGraph instance **plus any fields** in the default entity graph.

In other words, "load" **adds** to default entity graph and "fetch" **overrides** or resets it. These were brainstorming terms only, for illustration. I appended "NeededAttributes" to avoid reusing the "fetch" or "load" words in there. 

What we need to get right is, most importantly, the API (both external and internal), to promote correct development. Internal implementation may be optimized later as well, if such optimizations are discovered and needed. I think that one graph + one enum could be sufficiently efficient from memory usage perspective. Due to likely memory alignment it would not be much or at all different from graph + graph (max one non-null) or graph + boolean (e.g. "isFetchGraph").

Another alternative to mention (but, I think, NOT to pursue), is to have Fetch and Load graphs be of different classes. In some cases this may be marginally better in terms of memory use but it complicates matters otherwise - these things are indeed equal internally. To be honest, an advantage would be that the intention that the graph was created with will be clear and using graphs for what they are not intended for could be checked at compile time.

sebersole:
I have incorporated most of what we have discussed here into the code.  I am just finishing up the documentation stuff.  I'll have a new push soon that should be more complete

sebersole:
Just pushed my most recent changes.  This feels much more complete.  It incorporates everything discussed here.

@vladmihalcea If you get a chance would you look over the documentation changes?  Thanks

hibernate-core/src/main/java/org/hibernate/engine/spi/EffectiveEntityGraph.java
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	/**
+	 * Apply the graph and semantic.  The semantic is required.  The graph
+	 * may be null, but that should generally be considered mis-use.
+	 *
+	 * @throws IllegalArgumentException Thrown if the semantic is null
+	 * @throws IllegalStateException If previous state is still available (hasn't been cleared).
+	 */
+	public void applyGraph(GraphSemantic semantic, RootGraphImplementor<?> graph) {
```
aleksandarsusnjar: I like this.

documentation/src/test/java/org/hibernate/userguide/fetching/GraphParsingTest.java
```diff
@@ -0,0 +1,235 @@
...
+			final String queryString = "select e from Employee e where e.id = 1";
+			
+			final EntityGraph<Employee> graph = GraphParser.parse(
+					Employee.class,
+					graphString,
+					entityManager
+			);
+			
+			TypedQuery<Employee> query1 = entityManager.createQuery( queryString, Employee.class );
+			query1.setHint(
```
aleksandarsusnjar: Did you not "migrate" the method(s) to properly set these graphs without having to know about FETCH/LOAD hints/enums? 

Ideal case, of course, would have been, of course, that what is in lines 162-166 here could be stated as simply as:

`TypedQuery<Employee> query1 = entityManager.createQuery( graph, queryString, Employee.class );`

or:

`TypedQuery<Employee> query1 = entityManager.createQuery( GraphSemantic.FETCH, graph, queryString, Employee.class );`

Submitted approach (that did not modify existing Hibernate classes) used a utility method in EntityGraphs such as:

`EntityGraphs.setFetchGraph(query1, graph);`

sebersole:
What you propose there is already supported in a JPA compliant way.

I just don't see how knowing to call a proprietary method and still have to
understand the difference between fetch and load is any better than the
compliant way...

hibernate-core/src/main/java/org/hibernate/engine/spi/EffectiveEntityGraph.java
sebersole: Thanks :)

aleksandarsusnjar:
Everything I proposed in already supported (and was done) in a JPA _compliant_ but not in a _convenient_ way. JPA way also looks very much like unfinished business. At least in this case it is lagging behind instead of leading the implementations. I certainly hope that the situation will improve.

sebersole:
I don't necessarily disagree that the JPA way is easy.  I'm just saying I *personally* don't see these methods you asked about as any better in this regard.  But for me this stems from JPA making these "hints".  IMO hints should have been reserved for vendor hints.  

I'll let others add their opinions.  But like I said, personally did not see the benefit.

Now, if you want to talk about adding extensions to the Hibernate Query or LoadAcess contracts to specify the graph, I'm all for that as that is somewhat more straight-forward to use: e.g. `query.setFetchGraph(graph)...list()`, `query.setGraph(GraphSemantic.FETCH, graph)...list()`, etc

sebersole:
@aleksandarsusnjar BTW, these comments are in response to https://github.com/hibernate/hibernate-orm/pull/2558#discussion_r223067761 in case that was not clear.  Email replying does not seem to attach to the replied-to comment...

aleksandarsusnjar:
> Now, if you want to talk about adding extensions to the Hibernate Query or LoadAcess contracts to specify the graph, I'm all for that as that is somewhat more straight-forward to use: e.g. `query.setFetchGraph(graph)...list()`, `query.setGraph(GraphSemantic.FETCH, graph)...list()`, etc

That would be good. Ideal is, of course, JPA to evolve but we'll have to wait a while for that.

sebersole:
I am about to push some additions back to `EntityGraphs` and additions to `Query` to make it easier to work with `EntityGraph`s and queries.  

I started thinking about loading and had an interesting idea to run past y'all... basically combining the idea of EntityGraph and our LoadAccess contracts.  E.g., what do y'all think of:

````
Session {
    ... 
    // we have this already
    <E> LoadAccess<E> byId(Class<E> entityClass);
}

// we have this contract, but having it  extend RootGraph is new....

public interface LoadAccess<E> extends ... RootGraph {
    ...
}

final LoadAccess<MyEntity> plan = session.byId( MyEntity.class );
plan.addAttribute( "someAttribute" )
        .addAttribute( MyEntity_.anotherAttribute )
        .addSubgraph( ... )
        ...;

final MyEntity it = plan.load();
````

**Comment by aleksandarsusnjar** [comment]
@sebersole That sounds very interesting. Not sure how it ties to the rest of the system, though. Also to   consider is the use of "with" method.

**Comment by sebersole** [comment]
True. That would keep the method explosion out of the equation.  I'll play with that

**Comment by sebersole** [comment]
> @sebersole That sounds very interesting. Not sure how it ties to the rest of the system, though. Also to consider is the use of "with" method.

@aleksandarsusnjar Actually, what do you mean by 'the use of "with" method'?

**Comment by aleksandarsusnjar** [comment]
@sebersole:
> @aleksandarsusnjar Actually, what do you mean by 'the use of "with" method'?

I think you got it the first time, re reducing the explosion of methods. It would also allow composition to be used instead of inheritance. I will let others decide whether `LoadAccess` objects should be polymorphically reusable/passable as `EntityGraphs`.

**Comment by sebersole** [comment]
Ok that's what I thought.  Just wanted to make sure.  That's what it is the last commits

**Comment by sebersole** [comment]
So is everyone happy with this?

File: documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc [hunk 15]
```diff
@@ -248,6 +251,144 @@ include::{extrasdir}/fetching-strategies-dynamic-fetching-entity-subgraph-exampl
...
+SubGraphs can also be sub-type specific.  Given an attribute whose value is an inheritance hierarchy,
+we can refer to attributes of a specific sub-type using the forms of sub-graph definition that accept
+the sub-type Class.
+
+
+[[fetching-strategies-dynamic-fetching-entity-graph-parsing]]
+==== Creating and applying JPA graphs from text representations
+
+Hibernate allows the creation of JPA fetch/load graphs by parsing a textual representation
+of the graph.  Generally speaking, the textual representation of an graph is a comma-separated
```
gsmet: s/of an graph/of a graph/

documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc
```diff
@@ -248,6 +251,144 @@ include::{extrasdir}/fetching-strategies-dynamic-fetching-entity-subgraph-exampl
...
+invoiceGraph.addSubgraph( "responsibleParty" ).addAttributeNode( "taxIdNumber" );
+invoiceGraph.addSubgraph( "responsibleParty", Corporation.class ).addAttributeNode( "ceo" );
+invoiceGraph.addSubgraph( "responsibleParty", NonProfit.class ).addAttributeNode( "sector" );
+----
+====
+
+[[fetching-strategies-dynamic-fetching-entity-graph-merging]]
+==== Combining multiple JPA entity graphs into one
+
+Multiple entity graphs can be combined into a single supergraph that acts as a union. Graph from the previous example can also be build by combining separate aspect graphs into one, such as:
```
gsmet: s/be build/be built/

documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc
```diff
@@ -248,6 +251,144 @@ include::{extrasdir}/fetching-strategies-dynamic-fetching-entity-subgraph-exampl
...
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/GraphParsingTest.java[tags=fetching-strategies-dynamic-fetching-entity-graph-merging-example]
+----
+====
+
+[[fetching-strategies-dynamic-fetching-entity-graph-applying]]
+==== Applying entity graphs to queries
+
+Convenience/utility methods were created to make application of entity graphs to queries easier. Following examples show
```
gsmet: Not sure the sentence is finished?

hibernate-core/src/main/java/org/hibernate/Incubating.java
```diff
@@ -9,15 +9,22 @@
...
+import static java.lang.annotation.ElementType.CONSTRUCTOR;
 import static java.lang.annotation.ElementType.PACKAGE;
 import static java.lang.annotation.ElementType.METHOD;
 import static java.lang.annotation.ElementType.TYPE;
 import static java.lang.annotation.RetentionPolicy.CLASS;
 
 /**
+ * Marks (recursively) certain of Hibernate's packages, types and methods
+ * as incubating.  Incubating indicates a type or method that is still being
+ * actively developed and therefor may change at a later time.  Users of these
```
gsmet: s/therefor/therefore/ ?

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
```diff
@@ -182,19 +183,7 @@ public void disableFetchProfile(String name) throws UnknownProfileException {
...
-		return fetchGraph;
-	}
-
-	public void setFetchGraph(final EntityGraph fetchGraph) {
-		this.fetchGraph = fetchGraph;
-	}
-
-	public EntityGraph getLoadGraph() {
-		return loadGraph;
-	}
```
gsmet: We don't expect removing these methods to cause compatibility issues?

hibernate-core/src/main/java/org/hibernate/graph/GraphParser.java
```diff
@@ -0,0 +1,224 @@
...
+import javax.persistence.EntityManagerFactory;
+import javax.persistence.Subgraph;
+
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.graph.spi.GraphImplementor;
+import org.hibernate.graph.spi.RootGraphImplementor;
+
+/**
+ * todo (graph) : to me most of this seems more appropriate in documentation as opposed to here
```
gsmet: Should we remove the todo now?

hibernate-core/src/main/java/org/hibernate/graph/GraphSemantic.java
```diff
@@ -0,0 +1,62 @@
...
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.graph;
+
+/**
+ * JPA defines 2 distinct semantics for applying an EntityGraph.  This
+ * enumeration captures those 2 semantics (plus a `null` value).
```
gsmet: I think you remove the `null` value?

hibernate-core/src/main/java/org/hibernate/graph/InvalidGraphException.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.graph;
+
+import org.hibernate.HibernateException;
+
+/**
+ * Throw by {@link GraphParser} to indicate textual entity graph representation parsing errors.
```
gsmet: s/Throw/Thrown/ ?

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
sebersole: You could add them back if you want.  I'm not going to.  As discussed above they are misleading

documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc
sebersole: Probably.  I'll have to look

hibernate-core/src/main/java/org/hibernate/graph/GraphSemantic.java
sebersole: Maybe.  I'll have to look

sebersole:
I'm going to make the fixes suggested by @gsmet that I +1'd and look into the ones I mentioned I would.  After that, unless I hear some negative feedback I plan on pushing this

documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc
sebersole: Actually I think this whole section (`fetching-strategies-dynamic-fetching-entity-graph-applying`) should just go away.  It was originally trying to discuss the proprietary methods like `org.hibernate.graph.EntityGraphs#executeList`.  Initially there were many many more such methods.  TBH I personally think even these executeList ones should maybe go away...

The previous sections that @vladmihalcea had already written cover how to apply graphs to query and load/find operations.

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
sebersole: Hm, I thought I had already replied to this one.  You are possibly right.  But, in my opinion (and @aleksandarsusnjar iiuc) leaving them is confusing - they all imply that you can set both a load graph and a fetch graph.  As a compromise I'll add them back as deprecated pointing to this "confusion" fact and the correct way to do it.  Fair?

gsmet: Yep, sounds fair.

sebersole: fwiw here is the text I added locally:

````
	/**
	 * @deprecated (since 5.4) {@link #getFetchGraph}, {@link #getLoadGraph}, {@link #setFetchGraph}
	 * and {@link #setLoadGraph} (as well as JPA itself honestly) all make it very unclear that
	 * there can be only one graph applied at any one time and that graph is *either* a load or 
	 * a fetch graph.  These have all been replaced with {@link #getEffectiveEntityGraph()}.
	 * 
	 * @see EffectiveEntityGraph
	 */
````

sebersole:
Had to force push to account for rebasing against master

sebersole:
Unless someone sees something else to change, I will squash and push these changes later today

PR #2559, opened 2018-10-04
HHH-12992 - Fix OrderBy support for audited collection-based associations.

Naros:
https://hibernate.atlassian.net/browse/HHH-12992


Naros:
Applied upstream.

PR #2562, opened 2018-10-06
add test case for HHH-13016 - NPE when using SelectCase with Enum result

codylerum:
This will obviously fail CI as the test will fail

https://hibernate.atlassian.net/browse/HHH-13016


codylerum:
I've fixed the commit message so that it starts with the HHH issue

codylerum:
Looks like this can be closed as a test case already got added 

https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/simplecase/BasicSimpleCaseTest.java#L177

PR #2567, opened 2018-10-11
HHH-13009 - No documentation for "hibernate.javax.cache.cache_manager"

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13009

### DIFF
--- a/documentation/src/main/asciidoc/userguide/chapters/caching/Caching.adoc
+++ b/documentation/src/main/asciidoc/userguide/chapters/caching/Caching.adoc
@@ -559,6 +559,30 @@ In order to control which provider to use and specify configuration for the `Cac
 
 Only by specifying the second property `hibernate.javax.cache.uri` will you be able to have a `CacheManager` per `SessionFactory`.
 
+===== Using a non-default JCache `CacheManager`
+
+If you don't want to use the default `CacheManager`, you need to set the `hibernate.javax.cache.cache_manager` configuration property
+to one of the following values:
+
+Object reference:: If the value is an Object instance implementing the `CacheManager` interface,
+the provided `CacheManager` instance will be used.
+`Class`:: If the value is a Java `Class` object that implements the `CacheManager` interface,
+Hibernate will create a new instance for that `Class` and use it instead of the default one.
++
+[NOTE]
+====
+The default no-arg constructor is going to be used, so make sure that the `CacheManager`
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/userguide/chapters/caching/Caching.adoc
```diff
@@ -559,6 +559,18 @@ In order to control which provider to use and specify configuration for the `Cac
...
+
+If you don't want to use the default `CacheManager`, you need to set the `hibernate.javax.cache.cache_manager` configuration property
+to one of the following values:
+
+Object reference:: If the value is an Object instance implementing the `CacheManager` interface,
+the provided `CacheManager` instance will be used.
+`Class`:: If the value is a Java `Class` object that implements the `CacheManager` interface,
+Hibernate will create a new instance for that `Class` and use it instead of the default one.
+`String`:: If the value is a Java `String`, Hibernate expects it to be the fully-qualified `Class` name
+of the `CacheManager` implementation which will be used to instantiate the non-default `CacheManager`.
```
gsmet: Looks good to me but I think for `Class` and `String`, we should be explicit about what we expect as a constructor in the referenced class.

Could you check and add that?

Thanks!

vladmihalcea: @gsmet Are you asking what constructor is used when building the `CacheManager` instance? It's the default one, of course.

vladmihalcea: I'll make it more obvious then, maybe that's going to be relevant when providing the `CacheManager` as a Spring bean, as mentioned in the Jira issue.

pkgonan:
@vladmihalcea  @gsmet 
Hi.

Is there any example of hibernate.javax.cache.cache_manager in yml ?
I want to use Hibernate Second Level Cache using jcache (hibernate-jcache), and cache implementation is Hazelcast.

I want to use spring bean cache manager that already created using annotated spring bean not xml configuration.
But, I can not found example of **hibernate.javax.cache.cache_manager** . (example : reference of bean name.)

Is there any other solution about that?

Thank you.

vladmihalcea:
Your solution is fine. You need to pass the actual `CacheManager` to this property. It won't work via YAML, as you need a Spring-aware bean.

PR #2568, opened 2018-10-11
HHH-13001 Avoid the binding of the returning literals

fax4ever:
https://hibernate.atlassian.net/browse/HHH-13001

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/expression/function/ParameterizedFunctionExpression.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/expression/function/ParameterizedFunctionExpression.java
@@ -9,6 +9,7 @@ package org.hibernate.query.criteria.internal.expression.function;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
+import java.util.function.BiFunction;
 import javax.persistence.criteria.Expression;
 
 import org.hibernate.query.criteria.internal.CriteriaBuilderImpl;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/criteria/internal/expression/function/ParameterizedFunctionExpression.java
```diff
@@ -111,7 +111,8 @@ public String render(RenderingContext renderingContext) {
 	protected void renderArguments(StringBuilder buffer, RenderingContext renderingContext) {
 		String sep = "";
 		for ( Expression argument : argumentExpressions ) {
-			buffer.append( sep ).append( ( (Renderable) argument ).render( renderingContext ) );
+			// invoking renderProjection to avoid the binding of the returning literals
+			buffer.append( sep ).append( ( (Renderable) argument ).renderProjection( renderingContext ) );
```
gsmet: What is the rationale leading you to this fix?

The `LiteralHandlingMode` documentation says `The BIND mode will use bind variables for any literal value.`. So the previous behavior of rendering them as parameters seems legit to me and the intended behavior.

IMHO, the issue here is that the type of the parameter cannot be determined as it's lost in the JPQL translation.

I'll create a thread on the list to discuss my findings.

fax4ever:
Thanks @gsmet, I change applied the simple idea you provided me to just implement the `renderProjection` method. Much better now.

fax4ever:
@gsmet it definitely makes sense for me to close definitively this issue, if we can provide a more general solution ;)

PR #2571, opened 2018-10-11
HHH-12374 HHH-12380

gbadner:
https://hibernate.atlassian.net/browse/HHH-12374
https://hibernate.atlassian.net/browse/HHH-12380


gbadner:
Merged.

PR #2572, opened 2018-10-12
HHH-12368 - java.sql.SQLFeatureNotSupportedException in LobCreatorBuilderImpl

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12368

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -11,6 +11,7 @@ import java.io.OutputStream;
 import java.sql.Blob;
 import java.sql.CallableStatement;
 import java.sql.Clob;
+import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.NClob;
 import java.sql.ResultSet;
@@ -2955,6 +2956,18 @@ public abstract class Dialect implements ConversionContext {
 	 */
 	public String inlineLiteral(String literal) {
 		return String.format( "\'%s\'", escapeLiteral( literal ) );
+	}
+
+	/**
+	 * Check whether the JDBC {@link java.sql.Connection} supports creating LOBs via {@link Connection#createBlob()},
+	 * {@link Connection#createNClob()} or {@link Connection#createClob()}.
+	 *
+	 * @param databaseMetaData JDBC {@link DatabaseMetaData} which can be used if LOB creation is supported only starting from a given Driver version
+	 *
+	 * @return {@code true} if LOBs can be created via the JDBC Connection.
+	 */
+	public boolean supportsJdbcConnectionLobCreation(DatabaseMetaData databaseMetaData) {
+		return true;
 	}
 
 	/**
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/LobCreatorBuilderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/LobCreatorBuilderImpl.java
@@ -13,6 +13,7 @@ import java.sql.SQLException;
 import java.util.Map;
 
 import org.hibernate.cfg.Environment;
+import org.hibernate.dialect.Dialect;
 import org.hibernate.engine.jdbc.ContextualLobCreator;
 import org.hibernate.engine.jdbc.LobCreationContext;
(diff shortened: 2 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -2957,6 +2958,17 @@ public String inlineLiteral(String literal) {
 		return String.format( "\'%s\'", escapeLiteral( literal ) );
 	}
 
+	/**
+	 * Check whether the JDBC {@link java.sql.Connection} supports creating LOBs via {@link Connection#createBlob()} or {@link Connection#createBlob()}.
```
gsmet: `{@link Connection#createBlob()} or {@link Connection#createBlob()}`

both are the same :). I think you mean Clob for one of them?

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -2957,6 +2958,17 @@ public String inlineLiteral(String literal) {
...
+	 *
+	 * @param databaseMetaData JDBC {@link DatabaseMetaData} which can be used in case only starting from a given Driver version LOB creation is supported.
```
gsmet: `used in case only starting from a given Driver version LOB creation is supported`

I think I would go with:

`used if LOB creation is supported only starting from a given Driver version`

hibernate-core/src/test/java/org/hibernate/engine/jdbc/env/internal/LobCreationCheckSkipTest.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * @author Vlad Mihalcea
+ */
+public class LobCreationCheckSkipTest extends BaseEntityManagerFunctionalTestCase {
+
+	@Rule
+	public LoggerInspectionRule logInspection = new LoggerInspectionRule( Logger.getMessageLogger( CoreMessageLogger.class,
+																								   LobCreatorBuilderImpl.class
+																										   .getName()
+	) );
```
gsmet: There's something weird with the indentation here. The code is very far to the right.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/LobCreatorBuilderImpl.java
```diff
@@ -98,6 +102,10 @@ private static boolean useContextualLobCreation(Map configValues, Connection jdb
 					LOG.disablingContextualLOBCreationSinceOldJdbcVersion( meta.getJDBCMajorVersion() );
 					return false;
 				}
+
+				if ( !dialect.supportsJdbcConnectionLobCreation( meta ) ) {
+					//return false;
```
gsmet: Sounds fishy.

Hope the tests will fail, otherwise we will need new tests :).

hibernate-core/src/test/java/org/hibernate/engine/jdbc/env/internal/LobCreationCheckSkipTest.java
vladmihalcea: Unfortunately, that's what Idea does when doing CTRL+ALT+DEL using the Hibernate code syntax.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
vladmihalcea: Good catch.

vladmihalcea:
@gsmet Pushed a new version which addresses your review comments.

laptevn:
Just created Gradle project with Spring Data and PostgreSQL in Spring Initializr. Spring Boot version is 2.1.5.RELEASE.
Run it and get the same exception.

Is there a regression introduced or the issue wasn't solved completely?

roxspring:
Looks like the relevant Hibernate change (https://github.com/hibernate/hibernate-orm/commit/a5138621a69febf4f9ce9bd305a2f88775e4c5ae) went into versions 5.4.0+ and the downstream Spring Boot change (https://github.com/spring-projects/spring-boot/commit/8efddf57a5394bc39aada380590edf0e6eb9c29c) to use that went into versions 2.2.0.M2+ - so Spring Boot 2.1.5.RELEASE simply wasn't new enough to have the fix present.

PR #2573, opened 2018-10-12
HHH-13027 (5.3)

gbadner:
https://hibernate.atlassian.net/browse/HHH-13027


Sanne:
Looks good!

Minor nitpicking: I'd not store a logger in a static field when it's extremely unlikely to ever be used. We can just retrieve the logger if and when there's real need.

Sanne:
merged in 5.3 - I changed the logger as suggested in the previous comment.

Testing for master branch now...

Sanne:
merged in both master and 5.3. Thanks!

PR #2574, opened 2018-10-14
HHH-13051 AuditStrategy: Table structure is not hardcoded any more

Rokko11:
Hi,
you could configure the Hibernate Envers AuditStrategy from outside using the property `org.hibernate.envers.audit_strategy`. But there was no way to change the table structure the way you need it for your custom AuditStrategy.

I refactored it so people can create their own AuditStrategy with their corresponding properties.

Regards, Rokko


hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/metadata/AuditMetadataGenerator.java
```diff
@@ -173,43 +172,12 @@ void addRevisionType(Element anyMapping, Element anyMappingEnd, boolean isKey) {
...
-						null
-				);
-			}
-		}
+        // Add further columns to the aud table
+        addAdditionalColumns( anyMappingEnd );
+	}
+
+	private void addAdditionalColumns(Element anyMapping) {
+		auditStrategy.addAdditionalColumns(anyMapping, revisionInfoRelationMapping, verEntCfg);
```
gsmet: There's no need to add a private method for this. Just call it directly.

BTW, be careful, you need to indent with tabs.

hibernate-envers/src/main/java/org/hibernate/envers/boot/internal/EnversServiceImpl.java
```diff
@@ -182,15 +182,8 @@ private static AuditStrategy initializeAuditStrategy(
...
-			// further initialization required
-			final Getter revisionTimestampGetter = ReflectionTools.getGetter(
-					revisionInfoClass,
-					revisionInfoTimestampData,
-					serviceRegistry
-			);
-			( (ValidityAuditStrategy) strategy ).setRevisionTimestampGetter( revisionTimestampGetter );
-		}
+		// Strategy-specific initialization
+		strategy.initializeAdditional(revisionInfoClass, revisionInfoTimestampData, serviceRegistry);
```
gsmet: Need some padding in the parentheses (shoud be `strategy.initializeAdditional( revisionInfoClass, revisionInfoTimestampData, serviceRegistry );`).

Not very excited about the name of the method :).

hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/metadata/AuditMetadataGenerator.java
```diff
@@ -173,43 +172,12 @@ void addRevisionType(Element anyMapping, Element anyMappingEnd, boolean isKey) {
...
-						null,
-						null,
-						null,
-						null,
-						null
-				);
-			}
-		}
+        // Add further columns to the aud table
+        addAdditionalColumns( anyMappingEnd );
```
gsmet: I don't think I would call this here. You end up adding a very generic method in a very specific one.

I think I would add another call in `generateMappingData()` and `CollectionMetadataGenerator#createMiddleEntityXml()`, now that we have a proper method.

hibernate-envers/src/main/java/org/hibernate/envers/strategy/AuditStrategy.java
```diff
@@ -199,4 +202,8 @@ void addAssociationAtRevisionRestriction(
 			String alias1,
 			boolean inclusive,
 			MiddleComponentData... componentDatas);
+
+	void addAdditionalColumns(Element anyMapping, Element endRevMapping, AuditEntitiesConfiguration verEntCfg);
+
+	void initializeAdditional(Class<?> revisionInfoClass, PropertyData revisionInfoTimestampData, ServiceRegistry serviceRegistry);
```
gsmet: I think we need to make them empty default methods for backwards compatibility.

hibernate-envers/src/main/java/org/hibernate/envers/strategy/AuditStrategy.java
```diff
@@ -199,4 +202,8 @@ void addAssociationAtRevisionRestriction(
 			String alias1,
 			boolean inclusive,
 			MiddleComponentData... componentDatas);
+
+	void addAdditionalColumns(Element anyMapping, Element endRevMapping, AuditEntitiesConfiguration verEntCfg);
```
gsmet: The parameter name `endRevMapping` seems a bit off.

gsmet:
@Rokko11 I squashed your commits into one and rebased. I'll ask @Naros to review as it seems to be a nice addition.

hibernate-envers/src/main/java/org/hibernate/envers/strategy/AuditStrategy.java
```diff
@@ -199,4 +202,32 @@ void addAssociationAtRevisionRestriction(
...
+	 * Method to add additional Columns to the aud-tables.
+	 *
+	 * @param anyMapping the class mapping
+	 * @param revMapping the mapping of the revInfo-relation
+	 * @param auditEntitiesConfiguration the configuration of the audit entities
+	 */
+	default void addAdditionalColumns(
+			Element anyMapping,
+			Element revMapping,
+			AuditEntitiesConfiguration auditEntitiesConfiguration) {
```
Naros: I question if long-term a parameter object would be better served here.  In 6.0, I am eliminating the dom4j library and it will be replaced by the bootstrap process actually creating the Jaxb class instances themselves as a starting point to remove a lot of string-manipulation overhead.  By using a parameter object here, we would avoid tainting this method signature as we shift from 5.4 to 6.0.  I am not sure what I'd call such an object, so I need to think on it.

hibernate-envers/src/main/java/org/hibernate/envers/strategy/AuditStrategy.java
```diff
@@ -199,4 +202,32 @@ void addAssociationAtRevisionRestriction(
...
+	 * Method do add additional initialization behavior
+	 *
+	 * @param revisionInfoClass the class of the revision info
+	 * @param revisionInfoTimestampData the timestamp of the revision info
+	 * @param serviceRegistry the service registry
+	 */
+	default void initialize(
+			Class<?> revisionInfoClass,
+			PropertyData revisionInfoTimestampData,
+			ServiceRegistry serviceRegistry) {
```
Naros: I think the same holds true here for what I mentioned about `addAdditionalColumns`.

Naros:
I do have a few changes I want to make to your implementation, which also includes a bit of refactoring to the `org.hibernate.envers.strategy` package.  I want to make sure that I have a few things in place for this package so that the transition to 6.0 is much easier and clearer for users.

Naros:
I made some changes, could you give a look @Rokko11 and @gsmet  and let me know what you think.  If these look good I think we can look to integrate these into 5.4.

Rokko11:
Hi @Naros: Great, looks fine! It felt a little bit wrong to create a dependency from AufitStrategy to the xml library, but using a context object sounds good!

Naros:
@Rokko11 I completely agree.  It's also a great way to bridge the upcoming change from 5.x to 6.x when we move away from using Dom4j entirely in the Envers bootstrap.

gsmet:
Rebased and merged.

Thanks to both of you for the great work!

PR #2575, opened 2018-10-15
HHH-12955 Avoid a warn message by updating SchemaAutoTooling values

gsmet:
https://hibernate.atlassian.net//browse/HHH-12955

Some valid values of hibernate.hbm2ddl.auto were not present in this
class, whereas they should have been.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/Settings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/Settings.java
@@ -254,7 +254,8 @@ public final class Settings {
 
 	public boolean isAutoCreateSchema() {
 		return sessionFactoryOptions.getSchemaAutoTooling() == SchemaAutoTooling.CREATE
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/Settings.java
```diff
@@ -254,7 +254,8 @@ public boolean isAutoEvictCollectionCache() {
 
 	public boolean isAutoCreateSchema() {
 		return sessionFactoryOptions.getSchemaAutoTooling() == SchemaAutoTooling.CREATE
```
sebersole: Why not a method like `SchemaAutoTooling#isCreationEnabled()`?  That helps alleviate problems like this in the future.

gsmet: Well, all this is deprecated and useless AFAICS so I don't think it's worth doing more work on it.

Frankly, I was tempted to just remove all this but I suppose Settings is API and we can't do that.

PR #2578, opened 2018-10-15
HHH-13030 Upgrade hibernate-orm-modules WildFly version

fax4ever:
https://hibernate.atlassian.net/browse/HHH-13030

### DIFF
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -28,7 +28,7 @@ ext {
     geolatteVersion = '1.4.0'
 
     // Wildfly version targeted by module ZIP; Arquillian/Shrinkwrap versions used for CDI testing and testing the module ZIP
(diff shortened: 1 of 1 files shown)


gsmet:
@fax4ever does it solve the issue you were talking about, about jipijapa and such?

fax4ever:
From what I've understood, with this patch we will generate WildFly modules distribution suitable just for WildFly 14, unfortunately breaking the compatibility with WildFly 13. I think that the current jipijapa module could be *potentially compatible* just with a single version of WildFly. So maybe we have to think more before to merge it. I agree with @yrodiere's idea to move the issue on 5.4.x branch, in order to keep the 5.3.x consistent with WildFly 13.

gsmet:
Merged, thanks.

fax4ever:
Thanks!

PR #2579, opened 2018-10-16
HHH-13001 - NPE rendering nested criteria expressions

sebersole:
Initial work on https://hibernate.atlassian.net/browse/HHH-13001.

All tests pass.  Waiting to hear about specific tests.

Also, had to backport again the `Stack` interface and impls from 6 because https://github.com/hibernate/hibernate-orm/pull/2558 has still not been applied


sebersole:
@gsmet This is the general approach I mentioned.

gsmet:
@sebersole I rebased and added the test from Fabio. The test passes and the PR looks good so I think we're good to go.

Anything else on your side?

sebersole:
Nope.  Pushed as per my comments on the issue

PR #2582, opened 2018-10-16
HHH-13011

gbadner:
https://hibernate.atlassian.net/browse/HHH-13011

The first commit has tests for HHH-12104. I did this just to show that HHH-13011 has no affect on many-to-one assocations.


gsmet:
Merged and backported to 5.3, thanks.

PR #2584, opened 2018-10-16
HHH-12946 Include JAXB as a dependency as it's not provided by JDK 11

yrodiere:
https://hibernate.atlassian.net//browse/HHH-12946

The build now passes in both JDK8 and JDK11, with no difference when it comes to JAXB dependencies.

I'd say this addresses the ticket, as long as you're ok with the following limitations:

* Just as before the patch, running ORM in JDK11 + OSGi doesn't work at all. JDK8 + OSGi works fine.
* We can't upgrade to JAXB 2.3.1 yet.

I'm going to send another (preview) PR to explain these limitations more precisely and provide some (limited) solutions.

Note there is still a small difference in dependencies when building with JDK11: the JSR-250 dependency. But that should be addressed in another ticket, [HHH-12990](https://hibernate.atlassian.net//browse/HHH-12990).

### DIFF
--- a/hibernate-core/hibernate-core.gradle
+++ b/hibernate-core/hibernate-core.gradle
@@ -21,21 +21,6 @@ configurations {
 
 dependencies {
 
-    // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    // Java 9 ftw!
-    if ( JavaVersion.current().isJava9Compatible() ) {
-        compile( libraries.jaxb_api )
-        compile( libraries.jaxb_runtime )
-
-        xjc( libraries.jaxb_runtime )
-        xjc( libraries.jaxb_xjc )
-        xjc( libraries.jaxb2_basics )
-        xjc( libraries.jaxb2_basics_ant )
-        xjc( libraries.activation )
-        xjc( libraries.jsr250_api )
-    }
-    // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
     compile( libraries.jpa )
     // Javassist is no longer the default enhancer but still required for other purposes, e.g. Scanning
     compile( libraries.javassist )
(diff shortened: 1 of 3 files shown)


gradle/java-module.gradle
```diff
@@ -113,19 +113,22 @@ dependencies {
...
+		testCompile( libraries.jaxb_api )
+		testCompile( libraries.jaxb_runtime )
 		testCompile( libraries.jaxb_xjc )
 		testCompile( libraries.jaxb2_basics )
 		testCompile( libraries.jaxb2_basics_ant )
 		testCompile( libraries.jsr250_api )
 
-		testRuntime( libraries.jaxb )
+		testRuntime( libraries.jaxb_api )
+		testRuntime( libraries.jaxb_runtime )
```
gsmet: Doesn't jaxb_runtime depend on jaxb_api transitively?

gradle/java-module.gradle
```diff
@@ -107,35 +107,19 @@ dependencies {
...
 	// Java 9 ftw!
 	if ( JavaVersion.current().isJava9Compatible() ) {
-		// The JDK used to run Gradle is Java 9+, and we assume that that is the same
-		//		JDK for executing tasks
-		compile( libraries.jaxb_api )
-		compile( libraries.jaxb_runtime )
-		compile( libraries.jaxb_xjc )
-		compile( libraries.jaxb2_basics )
-		compile( libraries.jaxb2_basics_ant )
+		// In order for generated by jpamodelgen to be compiled successfully,
```
gsmet: I think a word is missing between  `for` and `generated`? Maybe `code`?

gradle/java-module.gradle
```diff
@@ -107,35 +107,19 @@ dependencies {
...
+		// we need the @Generated annotation to be in the classpath,
+		// and it's no longer in the JDK starting with JDK9.
```
gsmet: It's not exactly true. It has been moved.

gradle/java-module.gradle
yrodiere: It does, but our code actually depends on the API, not the runtime. Ideally we would not add a dependency to the runtime and we would let the user pick one, but as we know that's not a realistic expectation.

Also, now that I think of it, some applications don't use XML configuration at all, and for these applications it would be possible to exclude the dependency to the runtime while keeping the dependency to the APIs, and everything would work fine. If we only add a dependency to the runtime, they will have to exclude the dependency to the runtime, and add their own dependency to the APIs, which seems less... correct.

gradle/java-module.gradle
yrodiere: Fixed

gradle/java-module.gradle
yrodiere: Fixed

PR #2585, opened 2018-10-16
[PREVIEW, DO NOT MERGE] HHH-13022 Make OSGi integration work on JDK11

yrodiere:
https://hibernate.atlassian.net//browse/HHH-13022

This is based on #2584, which should be merged first.

**PLEASE DO NOT MERGE**, I'm just opening this so we can discuss the current state of the code, and the remaining problems.

This is far from ideal, but it's the best I could do after several days spent investigating.

On the bright side, OSGi tests are now enabled regardless of the JDK version

On the dark side, making JAXB work in OSGi is really hell. It only works at the price of an ugly hack, and only when sticking to JAXB 2.2.11.

## JAXB and OSGi

The first issue is that, for some reason, when we execute Hibernate ORM in Karaf, the JAXB APIs are loaded from some `lib/` folder provided by Karaf. I don't know why exactly, but I suspect our `OsgiClassLoader` has something to do with this: as it is currently implemented, it looks up classes in the client's bundle context first, and only if not found there we look them up in the Hibernate ORM bundle context.

This is bad enough, but with the version currently included in Karaf, we can sort of make it work.

The second issue is that JAXB JARs do not work "out of the box". Just adding the runtime JAR as a bundle to our feature is not enough, the API classes will not detect it. Apparently it's because the JAXB JARs do not declare service implementations as OSGi requires (using blueprints, for example). Here are the solutions I could think of:

* Use `com.sun.xml.bind:jaxb-osgi`, which is an OSGi bundle by the JAXB maintainers. Looks ideal? No it's not:
  * This bundle includes jaxb-xfc, which requires a lot of dependencies, among which package 'com.sun.source.tree' which is part of the Java compiler APIs and is not available by default in Karaf (I couldn't find the appropriate OSGi bundle).
  * I'm not even sure this OSGi bundle will work correctly if we solve the problem above, because the source code for this Maven artifact doesn't look like it does anything special to declare service implementations (no blueprints in particular). Maybe it will, but given the strange stuff we do in our classloader, I wouldn't bet on it.
* Wrap the JAXB JARs ourselves, adding whatever OSGi configuration that is necessary.
  * Maybe we could [create a clean bundle with all the correct configuration](http://karaf.apache.org/manual/latest/#_statically_bundling_jars), but that would essentially mean re-building the `jaxb-osgi` artifact ourselves.
  * I was able to come up with a very simple wrapping. It's an ugly and probably fragile hack, but it works in Karaf with both JRE 8 and JRE 11. The hack consists in wrapping the JAXB runtime bundle, adding in particular an "Export-Package: META-INF.services" instruction to the manifest. Due to how resource loading works in Karaf, this ends up exposing the relevant `META-INF/services/XXX` file from the runtime bundle to the API classes, and the API classes end up instantiating the runtime as we would expect.

So once again, this is bad, but I was able to make it work, at least.

The third issue is that upgrading to JAXB 2.3 will probably be quite complex. I tried it, and faced the following challenges:

* JAXB 2.3.1 requires `javax.activation` 1.2.0, which declares an OSGi dependency to package `com.sun.activation.registries`, which is not available by default in Karaf (I couldn't find the appropriate OSGi bundle).
* JAXB 2.3.1 requires `stax-ex` 1.8, which declares an OSGi dependency to Java 9+, even though it should work just fine on Java 7 (it's built using JDK7)
* ... and that's only where I stopped, there may be more.

### DIFF
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -36,6 +36,10 @@ ext {
 
     jodaTimeVersion = '2.3'
 
+    jaxbVersion = '2.2.11'
+    // Strangely, jaxb-runtime 2.2.11 depends on jaxb-api 2.2.12-b140109.1041
+    jaxbApiVersion = '2.2.12-b140109.1041'
+
     libraries = [
             // Ant
             ant:            'org.apache.ant:ant:1.8.2',
@@ -75,10 +79,9 @@ ext {
             logging_processor:  'org.jboss.logging:jboss-logging-processor:2.1.0.Final',
 
             // jaxb task
-            // Strangely, jaxb-runtime 2.2.11 depends on jaxb-api 2.2.12-b140109.1041
-            jaxb_api: 'javax.xml.bind:jaxb-api:2.2.12-b140109.1041',
-            jaxb_runtime: 'org.glassfish.jaxb:jaxb-runtime:2.2.11',
-            jaxb_xjc: 'org.glassfish.jaxb:jaxb-xjc:2.2.11',
+            jaxb_api: "javax.xml.bind:jaxb-api:${jaxbApiVersion}",
+            jaxb_runtime: "org.glassfish.jaxb:jaxb-runtime:${jaxbVersion}",
+            jaxb_xjc: "org.glassfish.jaxb:jaxb-xjc:${jaxbVersion}",
             // Note that jaxb2_basics is a set of tools on *top* of JAXB.
             // See https://github.com/highsource/jaxb2-basics
             jaxb2_basics: 'org.jvnet.jaxb2_commons:jaxb2-basics:0.11.0',
--- a/hibernate-osgi/hibernate-osgi.gradle
+++ b/hibernate-osgi/hibernate-osgi.gradle
@@ -19,13 +19,6 @@ ext {
 
 test.enabled = true
 
-if ( JavaVersion.current().isJava9Compatible() ) {
-	logger.warn( '[WARN] Skipping all tests for hibernate-osgi due to Karaf issues with JDK 9' )
-	// Hikari CP relies on Javassist which we know has issues with Java 9
-	test.enabled = false
-}
-
-
 sourceSets {
 	test {
 		// send javac output and resource copying to the same output directory for test
--- a/hibernate-osgi/hibernate-osgi.gradle
+++ b/hibernate-osgi/hibernate-osgi.gradle
@@ -41,10 +41,6 @@ configurations {
 		description = 'Feature for easily adding Infinispan-based caching support to hibernate-orm'
 		transitive = false
 	}
-	jaxb {
-		description = 'Jaxb dependencies'
-		transitive = true
-	}
 }
 
 dependencies {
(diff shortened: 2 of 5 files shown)


gsmet:
OK, thanks for the summary.

I think it would be worth opening a Karaf issue to ask for advice on how to solve this issue. We probably won't be the only one requiring a JAXB dependency. They might be more legitimate than us to ask for manifest fixes and so on.

yrodiere:
A quick update. While working on reproducers to submit to the Karaf team for advice, I noticed that two weird pieces of code in particular was necessary to reproduce our problem. In both those places, we explicitly set the Thread Context Class Loader (TCCL) to a classloader of our own:

* `org.hibernate.osgi.OsgiPersistenceProviderService#getService`:
  ```
	@Override
	public Object getService(Bundle requestingBundle, ServiceRegistration registration) {
		final OsgiClassLoader osgiClassLoader = new OsgiClassLoader();

		// First, add the client bundle that's requesting the OSGi services.
		osgiClassLoader.addBundle( requestingBundle );

		// Then, automatically add hibernate-core and hibernate-entitymanager.  These are needed to load resources
		// contained in those jars, such as em's persistence.xml schemas.
		osgiClassLoader.addBundle( FrameworkUtil.getBundle( SessionFactory.class ) );
		osgiClassLoader.addBundle( FrameworkUtil.getBundle( HibernateEntityManagerFactory.class ) );

		// Some "boot time" code does still rely on TCCL.  "run time" code should all be using
		// ClassLoaderService now.

		final ClassLoader originalTccl = Thread.currentThread().getContextClassLoader();
		Thread.currentThread().setContextClassLoader( osgiClassLoader );
		try {
			return new OsgiPersistenceProvider( osgiClassLoader, osgiJtaPlatform, osgiServiceUtil, requestingBundle );
		}
		finally {
			Thread.currentThread().setContextClassLoader( originalTccl );
		}
	}
  ```
* `org.hibernate.osgi.OsgiSessionFactoryService#getService`:
  ```
	@Override
	public Object getService(Bundle requestingBundle, ServiceRegistration registration) {
		final OsgiClassLoader osgiClassLoader = new OsgiClassLoader();

		// First, add the client bundle that's requesting the OSGi services.
		osgiClassLoader.addBundle( requestingBundle );

		// Then, automatically add hibernate-core.  These are needed to load resources
		// contained in the core jar.
		osgiClassLoader.addBundle( FrameworkUtil.getBundle(SessionFactory.class) );

		// Some "boot time" code does still rely on TCCL.  "run time" code should all be using
		// ClassLoaderService now.

		final ClassLoader originalTccl = Thread.currentThread().getContextClassLoader();
		Thread.currentThread().setContextClassLoader( osgiClassLoader );
		try {
			return buildSessionFactory( requestingBundle, osgiClassLoader );
		}
		finally {
			Thread.currentThread().setContextClassLoader( originalTccl );
		}
	}
  ```

Interestingly enough, if I remove these pieces of code, the OSGi tests now pass with JDK8 and JDK11, even if I do *not* add JAXB to our Karaf feature (it uses the Karaf-provided JAXB runtime). And no other tests fail. I added two commits to demonstrate this.

Also of interest: setting the TCCL in `OsgiSessionFactoryService` will have an effect during the whole duration of the bootstrapping process, but for `OsgiPersistenceProviderService`, it will only be effective in the `OsgiPersistenceProvider` constructor, which is dead simple and unlikely to trigger much class loading. `OsgiPersistenceProvider` itself, when it creates the `EntityManagerFactory`, does not set the TCCL. So there's something fishy here, at the very least an inconsistency between how we start in JPA and native mode.

I wonder if setting the TCCL is actually useful? I know it creates problems, but maybe it solves others? If it does, they are untested problems, though... I'll investigate a bit more, maybe try to ping Steve.

yrodiere:
Superseded by #2589

PR #2586, opened 2018-10-16
HHH-13040 - MetadataBuilderContributor should be applied before MetadataBuildingProcess.prepare

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13040


vladmihalcea:
@gsmet I think this one should also be backported to 5.3 as well. It's a small change, without major risks since it's quite a new functionality, but the benefits for the user are nice, hence we might want to backport it.

gsmet:
Thanks, merged. Marked it as a potential candidate for backporting but I wouldn't hold my breath if I were you.

Hopefully, we will get 5.4 out of the door soon.

vladmihalcea:
@gsmet Most users don't know about this new property, but it could be that they discover it later from the docs, and if we don't backport it, they won't be able to register the `AttributeConverters` dynamically. Anyway, thanks for integrating it.

PR #2587, opened 2018-10-17
HHH-13041 Mention that @Any is a unidirectional mapping

johnlinp:


documentation/src/main/asciidoc/userguide/chapters/domain/basic_types.adoc
```diff
@@ -2104,6 +2104,12 @@ To link the value of the metadata information and an actual entity type, the `@A
 The `metaType` attribute allows the application to specify a custom type that maps database column values to persistent classes that have identifier properties of the type specified by `idType`.
 You must specify the mapping from values of the `metaType` to class names.
 
+[NOTE]
+====
+`@Any` mapping can only be unidirectional.
```
vladmihalcea: Why would you assume this is bidirectional? The only bidirectional associations are the ones exposed via JPA.

documentation/src/main/asciidoc/userguide/chapters/domain/basic_types.adoc
```diff
@@ -2104,6 +2104,12 @@ To link the value of the metadata information and an actual entity type, the `@A
...
+That is, the management can only be done on the entity that holds the relational data.
```
vladmihalcea: What do you mean by `entity that holds the relational data`? Do you want to say the entity that contains the `@JoinColumn`?

vladmihalcea:
Superceded by this [Pull Request](https://github.com/hibernate/hibernate-orm/pull/2602).

documentation/src/main/asciidoc/userguide/chapters/domain/basic_types.adoc
johnlinp: I am new to Hibernate or JPA, and I am trying to map `@Any` bidirectionally. However, after some research, I found the following statements in [this article](https://javabeat.net/polymorphic-association-mapping-relationship-hibernate/):

> Any relationship cannot be implemented as bidirectional association. It will be unidirectional relationship managed by entity holding the relationship data.

So I was thinking that it's good to save others' time searching for bidirectional `@Any` by putting it on the official documentation.

vladmihalcea: [This commit](https://github.com/hibernate/hibernate-orm/commit/fecb12cff70b3bdcb749582106731d7bdf3b2f8d) adds the following clarifications:

> The `@Any` mapping is useful to emulate a unidirectional `@ManyToOne` association when there can be multiple target entities.

and

> While the `@Any` mapping is useful to emulate a `@ManyToOne` association when there can be multiple target entities, to emulate a `@OneToMany` association, the `@ManyToAny` annotation must be used.

johnlinp: Great! Thanks for your explanations.

PR #2589, opened 2018-10-17
HHH-13022 Make OSGi integration work on JDK11

yrodiere:
https://hibernate.atlassian.net//browse/HHH-13022

For the record, the initial investigation was recorded there: #2585


gsmet:
Merged, thanks!

PR #2590, opened 2018-10-17
HHH-12945 Properly support entry() criteria expression

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12945

@sebersole interested in your feedback on this, especially on the addition of the `getSelectionSpan()` method to `SelectionImplementor`.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/SelectionImplementor.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/SelectionImplementor.java
@@ -15,4 +15,8 @@ import javax.persistence.criteria.Selection;
  */
 public interface SelectionImplementor<X> extends TupleElementImplementor<X>, Selection<X>  {
 	public List<ValueHandlerFactory.ValueHandler> getValueHandlers();
+
+	default int getSelectionSpan() {
+		return 1;
+	}
 }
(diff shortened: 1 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/query/criteria/internal/SelectionImplementor.java
```diff
@@ -15,4 +15,8 @@
  */
 public interface SelectionImplementor<X> extends TupleElementImplementor<X>, Selection<X>  {
 	public List<ValueHandlerFactory.ValueHandler> getValueHandlers();
+
+	default int getSelectionSpan() {
```
yrodiere: I understand you wanted to be able to define a selection with a span of more than 1, but whose `isCompound` method would return `false`.

I wonder why exactly you don't want `MapEntryExpression` to be considered as compound? Does it break things, or do you fear that it will? After all, an entry is a key and a value, so it *is* compound...

gsmet: IIRC, that's the first thing I tried and it did break things.

gsmet:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2629 .

PR #2591, opened 2018-10-18
HHH-12979 - Setting hibernate.javax.cache.uri property value as relative path causes an error

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12979


gsmet:
@vladmihalcea I noticed that `ConfigHelper` is deprecated and it is mentioned we should use the `ClassLoaderService`. It's indeed what is done in the legacy Ehcache region factory (together with other manipulations which are IMHO useless) so I added a commit with this approach.

Note that it doesn't support the `classpath://` URI scheme as it's not something we support elsewhere. It will be supported if it is supported by the framework (Spring registers a `classpath://` handler for instance).

But it checks the classpath by default, which is what we used to do for Ehcache.

Thoughts?

sebersole:
We do support `classpath://` urls elsewhere as I listed somewhere else we
discussed this.

FWIW I did look into writing and registering a custom handler.  The problem
that I ran into there is that you can register only a
single URLStreamHandlerFactory in a given VM - as in the method to register
a URLStreamHandlerFactory can literally only be called once in a given VM
*ever*.  I suspect that works for Spring because it controls VM startup (as
a container or boot).  IMHO that is a dead-end approach for us to look at.


[1] See the javadocs -
https://docs.oracle.com/javase/8/docs/api/java/net/URL.html#setURLStreamHandlerFactory-java.net.URLStreamHandlerFactory-

vladmihalcea:
@gsmet @sebersole I thought that I don't have access to `ClassLoaderService`, but realized that I could load it via the `SessionFactoryOptions.getServiceRegistry`.

I'll try to use that instead and see how it works.

sebersole:
That's exactly how you get the ClassLoaderService, so +1 :D

vladmihalcea:
@gsmet I noticed your commit and saw you deleted the "classpath://" test. Shouldn't we add support for that in `ClassLoaderService` as well?

Due to Spring, some devs might write the URI with the "classpath" scheme, although that's not natively supported by Java.

sebersole:
> Due to Spring, some devs might write the URI with the "classpath" scheme, although that's not natively supported by Java.

I'm confused by this.  The whole premise here is that, yes, we want to add support for this *where we load the resource* (important distinction).  It sounds like you are asking if we want to do that (resounding yes).  Am I missing something?

gsmet:
As I understand it and as it is done with my commit, if Spring is used, it will register a handler for `classpath://` and it will work out of the box (the first `new URI( name )` will succeed).

If not, it won't.

But I think it's fair and we shouldn't try to play with the `classpath://` handler considering Spring might have registered one with a different classloader than ours.

Maybe we could have an additional thing in `ClassLoaderServiceImpl` that removes the `classpath://` prefix if there is one, once the first call has failed. This way, in the Spring case, it will work and use the Spring class loader and if Spring is not there, it will still work and use our class loader.

Would that work for everyone?

vladmihalcea:
@gsmet sounds good to me.

vladmihalcea:
@gsmet I think we should rebase and integrate this one. What do you think?

gsmet:
@vladmihalcea yes, I added a commit in the Monday morning GitHub confusion and was unable to add a comment to warn you. If the last commit looks good to you, it's ready to be merged.

I just rebased the PR.

PR #2593, opened 2018-10-18
HHH-13045 - Test and fix for accessing assocation id on alias of element collection in query

beikov:
https://hibernate.atlassian.net/browse/HHH-13045

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
@@ -683,7 +683,7 @@ public class DotNode extends FromReferenceNode implements DisplayableNode, Selec
 				return null;
 			}
 			// If the lhs is a collection, use CollectionPropertyMapping
-			Type propertyType = fromElement.getPropertyType( propertyName, propertyPath );
+			Type propertyType = fromElement.getPropertyType( propertyPath, propertyPath );
 			LOG.debugf( "getDataType() : %s -> %s", propertyPath, propertyType );
 			super.setDataType( propertyType );
 		}
--- a/hibernate-core/src/test/java/org/hibernate/test/annotations/collectionelement/EmbeddableCollectionElementWithLazyManyToOneTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/annotations/collectionelement/EmbeddableCollectionElementWithLazyManyToOneTest.java
@@ -128,6 +128,31 @@ public class EmbeddableCollectionElementWithLazyManyToOneTest extends BaseCoreFu
 		);
 	}
 
+	@Test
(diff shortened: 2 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/test/annotations/collectionelement/EmbeddableCollectionElementWithLazyManyToOneTest.java
```diff
@@ -128,6 +128,31 @@ public void testLazyBoth() {
 		);
 	}
 
+	@Test
+	@TestForIssue( jiraKey = "???")
```
JWGmeligMeyling: HHH-13045

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
```diff
@@ -683,7 +683,7 @@ public Type getDataType() {
 				return null;
 			}
 			// If the lhs is a collection, use CollectionPropertyMapping
-			Type propertyType = fromElement.getPropertyType( propertyName, propertyPath );
+			Type propertyType = fromElement.getPropertyType( propertyPath, propertyPath );
```
gsmet: The method is the following:
```
	public Type getPropertyType(String propertyName, String propertyPath) {
		return elementType.getPropertyType( propertyName, propertyPath );
	}
```
so it makes a clear difference between the `propertyName` and the `propertyPath`. I wonder if it's more about the underlying code using the right parameter (e.g. it's using `propertyName` whereas it should be using `propertyPath` instead).

From what I can see, the confusion about this is growing in the code.

WDYT?

JWGmeligMeyling: Throughout the persisters its not uncommon for the definition of propertyName to be stretched to allow for nested property names. Other code in `DotNode` and `BasicEntityPropertyMapping` / `PropertyMapping` is already dependent on this. I currently interpret propertyName as *a path expression, for which its value can be determined by the property mapping of the root*. I think that best describes its use throughout `DotNode` and `PropertyMapping`. The name is slightly confusing though, and its expectation could be better document. But I can't think of a better name. Just using `propertyPath` here probably just leads to new sorts of confusion as well.

beikov: The `PropertyMapping` contains the full property paths, so it should never be a problem to do it this way. The code is very old, maybe the different parameters were used differently earlier. I tried looking through the history but couldn't find anything of value. All I can say is that using the propertyPath is IMO the correct way and that the propertyName could be dropped. Also note that there are a few callers of that method that do pass the same value twice as well.

gsmet:
Merged, thanks!

PR #2597, opened 2018-10-22
HHH-13057 Prevent Byte Buddy's Advice helper to reload to many resour…

Sanne:
…ces from the ClassLoader

Best understood after reading the JIRA:
 - https://hibernate.atlassian.net/browse/HHH-13057

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/EnhancerImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/EnhancerImpl.java
@@ -72,6 +72,27 @@ public class EnhancerImpl implements Enhancer {
 	private final TypePool typePool;
 
 	/**
+	 * Extract the following constants so that enhancement on large projects
+	 * can be done efficiently: otherwise each instance use will trigger a
+	 * resource load on the ClassLoader tree, triggering allocation of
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/EnhancerImpl.java
```diff
@@ -167,18 +187,18 @@ private TypePool buildTypePool(final ByteBuddyEnhancementContext enhancementCont
 									.annotateField( AnnotationDescription.Builder.ofType( Transient.class ).build() )
 							.defineMethod( EnhancerConstants.TRACKER_CHANGER_NAME, void.class, Visibility.PUBLIC )
 									.withParameters( String.class )
-									.intercept( Advice.to( CodeTemplates.TrackChange.class ).wrap( StubMethod.INSTANCE ) )
+									.intercept( adviceTrackChange.wrap( StubMethod.INSTANCE ) )
```
gsmet: Wondering if it would make sense to also cache the `wrap()` call. It would avoid creating a new object each time.

Sanne: I was afraid you'd say that :)

On it..

gsmet:
Nice! Merged, thanks!

PR #2600, opened 2018-10-24
HHH-11903 : @OneToOne Derived ID is null when returned by query when bidirectional

gbadner:
https://hibernate.atlassian.net/browse/HHH-11903


gbadner:
@dreab8, thanks for reviewing! I resolved the conflict and pushed to master.

PR #2601, opened 2018-10-24
HHH-13060

StanislavGubanov:
Test case for issue HHH-13060


gsmet:
@StanislavGubanov it looks like there are other commits in the PR? Could you rebase on master and have only the commits with the test case?

Thanks!

PR #2603, opened 2018-10-24
HHH-13061 Introduce a 'none' option for BytecodeProvider implementati…

Sanne:
…on choice

As proposed on the mailing list.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/internal/none/BytecodeProviderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/internal/none/BytecodeProviderImpl.java
@@ -1,0 +1,47 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.bytecode.internal.none;
+
+import org.hibernate.HibernateException;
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/bytecode/internal/none/BytecodeProviderImpl.java
```diff
@@ -0,0 +1,47 @@
...
+			Class clazz,
+			String[] getterNames,
+			String[] setterNames,
+			Class[] types) {
+		throw new HibernateException( "Using the ReflectionOptimizer is not possible when the configured BytecodeProvider is 'none'. Disable " + AvailableSettings.USE_REFLECTION_OPTIMIZER + " or use a different BytecodeProvider");
+	}
+
+	@Override
+	public Enhancer getEnhancer(EnhancementContext enhancementContext) {
+		return null;
```
gsmet: Just a minor nitpick but shouldn't we throw an exception here too?

gsmet:
So, from what I understand, if you want to use it in a normal setup, you would have to use a real bytecode provider at compile time and this one at runtime?

Sanne:
> So, from what I understand, if you want to use it in a normal setup, you would have to use a real bytecode provider at compile time and this one at runtime?

Exactly.

hibernate-core/src/main/java/org/hibernate/bytecode/internal/none/BytecodeProviderImpl.java
Sanne: No the null value is actually what the API clients are expecting.

gsmet:
Merged. I don't think we should advertise it for now so let's not document it.

PR #2604, opened 2018-10-25
HHH-12713 - Make EntityGraph creation more convenient

sebersole:
- unify naming between 5.x and 6.0 wrt EntityGraph support added in 5.4

Here is the work I mentioned on HipChat this morning.  It aligns naming in the metamodel and graph contracts between 5.4 and 6.0.  It's matching 6.0 alignment is encompassed in the PR https://github.com/sebersole/hibernate-core/pull/152


gsmet:
@sebersole I don't have a strong opinion on the naming and I am not really following the 6 development so I'm neutral on this one.

PR #2608, opened 2018-10-26
HHH-12713 EntityGraph changes - Restore (most) SPI backward compatibility

yrodiere:
This PR is based on #2604, which should be merged first.

This should fix most of the backward-incompatible changes introduced in #2558 and #2604 ([HHH-12713](https://hibernate.atlassian.net/browse/HHH-12713)).

The only backward-incompatible changes I did not try to address are those in Query and Session; if we currently override the changed methods in OGM, for instance, we will have to adapt the OGM code, because some parameter types and return types changed.

For the record, here are my notes (don't bother to read it, it's just for reference):

```
First PR: git diff --patch-with-stat --cherry-pick --right-only c49ebe888a7a485b0ec04aa25aee680ee7810151~3..c49ebe888a7a485b0ec04aa25aee680ee7810151 -- ':*.java' ':!*/internal/*' ':!*/src/test/*'

## org.hibernate.Session

More constrained return type for some methods -- fine as long as we don't override them in Search/OGM.
Ex. createEntityGraph(Class<T> rootType) returns RootGraph, previously EntityGraph

## org.hibernate.graph.spi.AttributeNodeImplementor

Added methods -- fine if only ORM itself is supposed to implement this
Ex.: makeSubGraph(Class), makeKeySubGraph(Class) but also (through inheritance) makeCopy(boolean), ...


## org.hibernate.graph.spi.EntityGraphImplementor

Deleted, probably renamed to org.hibernate.graph.spi.RootGraphImplementor


## org.hibernate.graph.spi.GraphNodeImplementor

Added methods -- fine if only ORM itself is supposed to implement this
Ex.: makeCopy(boolean), isMutable()

Removed methods
Ex.: attributeImplementorNodes(), attributeNodes(), containsAttribute(String)


## org.hibernate.metamodel.spi.MetamodelImplementor

Renamed to org.hibernate.metamodel.model.domain.spi.MetamodelImplementor

More constrained parameter types
Ex.: addNamedEntityGraph(String graphName, EntityGraph<T> entityGraph) => addNamedEntityGraph(String graphName, RootGraphImplementor<T> entityGraph)


## org.hibernate.query.Query

Added methods -- fine because implementations added to AbstractProducedQuery, but might require adjustments to throw exceptions when unsupported in Search/OGM.
Ex.: applyGraph(RootGraph graph, GraphSemantic semantic)


Second PR: git diff --patch-with-stat --cherry-pick --right-only 760965b5b69d3a5436e07f3452410c0bf2ea7329~..760965b5b69d3a5436e07f3452410c0bf2ea7329 -- ':*.java' ':!*/internal/*' ':!*/src/test/*'

Seems fine -- all affected types were created in the first PR.
```


hibernate-core/src/main/java/org/hibernate/engine/query/spi/EntityGraphQueryHint.java
```diff
@@ -37,30 +39,32 @@
  *
  * @author Brett Meyer
  */
-public class EntityGraphQueryHint {
-	private final RootGraph<?> graph;
+public class EntityGraphQueryHint implements AppliedGraph {
```
sebersole: I'm confused why you say that the other HHH-12713 PR needs to be applied first.  Unless I miss something, this PR already includes those changes.  Here is a good example.  This exact change exists in that other PR - https://github.com/hibernate/hibernate-orm/pull/2604/files#diff-42aef08c22c86f0e459a475e822240a4R42

yrodiere:
> Overall it is hard to see what this PR actually does because it is mixed together with the changes from #2604

Yes, I know. It is based on #2604, and GitHub doesn't allow to submit a PR "to another PR". So we'll just have to wait until #2604 is merged, or review commit by commit.

> I really wish this is stuff you would discuss with me beforehand too.

I did send a message to hibernate-dev a few days ago, got no answer, so I figured I would propose something that we could then discuss. But yeah, discussing it with you beforehand was the idea.

>  Why the changes here at all?

Explained in my message to hibernate-dev: basically the renaming of `MetamodelImplementor` broke the Hibernate Search build, and there was some expectations that 5.4 would only bring "compatible" changes.

hibernate-core/src/main/java/org/hibernate/engine/query/spi/EntityGraphQueryHint.java
yrodiere: Yes. You can't send a PR "B" based on another PR "A" without including the commits from PR "B".

I said the other PR needs to be merged first, because we need to review that code first. If we were to merge my PR now, we would effectively merge yours too.

sebersole:
> I did send a message to hibernate-dev a few days ago, got no answer, so I figured I would propose something that we could then discuss. But yeah, discussing it with you beforehand was the idea.

Interesting.  I did not get that message at all.  More trouble with the dev ml maybe?


> > Why the changes here at all?
> 
> Explained in my message to hibernate-dev: basically the renaming of `MetamodelImplementor` broke the Hibernate Search build, and there was some expectations that 5.4 would only bring "compatible" changes.

Ok, if the only effective change is restoring `MetamodelImplementor` from `org.hibernate.metamodel.model.domain.spi` to `org.hibernate.metamodel.model.spi`, then +1 - but I would actually just drop `org.hibernate.metamodel.model.domain.spi.MetamodelImplementor`.  Again, all of this stuff has been consider incubating so we have some leeway in terms of what we want to do.  But keeping `MetamodelImplementor` in the old package is fine.  We should be sure to do the same in 6 as well - technically because this package is incubating we do not have to, but its nice if we can.

However, is that really the only effective change here?  3 commits seems like an awful lot just to handle `MetamodelImplementor`.

yrodiere:
> Interesting. I did not get that message at all. More trouble with the dev ml maybe?

Strange, because I sent it to your `@hibernate.org` address personally as well as to the mailing list. But we did experience some trouble with gmail the past few days, who knows.

> but I would actually just drop `org.hibernate.metamodel.model.domain.spi.MetamodelImplementor`

Sure, that would be even better. I can do that.

> However, is that really the only effective change here? 3 commits seems like an awful lot just to handle MetamodelImplementor.

The first commit handles `MetamodelImplementor`.

The next two handle other backward-incompatible changes that I noticed, but that do not affect Search. @gsmet wanted me to check everything, so I did, but if y'all agree that we don't care about backward compatibility on ` GraphNodeImplementor` and `EntityGraphImplementor` after all, I'm fine with that.

gsmet:
I'm fine with not considering compatibility for classes specific to entity graphs.

So let's focus on `MetamodelImplementor` then.

yrodiere:
I updated the PR, now it only restores backward compatibility for `MetamodelImplementor`.

sebersole:
+1.  I'll do the same in 6 as well to match

sebersole:
Actually that's exactly where it already is in 6 :)

PR #2544, opened 2018-09-26
 HHH-12842 - Fix lazy loading of ID OneToOne relations that are non-optional / constrained

JWGmeligMeyling:
Fix lazy loading of inverse OneToOne relations that are non-optional / constrained, by passing the OneToOne constrained value to determine nullability of type as it was prior to Hibernate 5.2.13.

Fixes [HHH-12842](https://hibernate.atlassian.net/browse/HHH-12842) and the test case that I had already provided in PR #2474 .

Partially applies PR #2226 that @vladmihalcea provided for [HHH-12436](https://hibernate.atlassian.net/browse/HHH-12436), but leaves out the changes related to the determination of the ForeignKeyDirection of the OneToOne association (the fix for HHH-12436), which caused some issues that @gbadner is still addressing.

PR #2609, opened 2018-10-27
HHH-12436 : Attempted to assign id from null one-to-one property	

gbadner:
https://hibernate.atlassian.net/browse/HHH-12436

Supersedes: https://github.com/hibernate/hibernate-orm/pull/2226

Another jira will be created to deal with associations that change from FetchMode.JOIN to FetchMode.SELECT.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
@@ -1779,7 +1779,17 @@ public final class AnnotationBinder {
 						joinColumn.setExplicitTableName( join.getTable().getName() );
 					}
 				}
-				final boolean mandatory = !ann.optional() || forcePersist;
+				// MapsId means the columns belong to the pk;
+				// A @MapsId association (obviously) must be non-null when the entity is first persisted.
+				// If a @MapsId association is not mapped with @NotFound(IGNORE), then the association
+				// is mandatory (even if the association has optional=true).
+				// If a @MapsId association has optional=true and is mapped with @NotFound(IGNORE) then
+				// the association is optional.
+				final boolean mandatory =
+						!ann.optional() ||
+								property.isAnnotationPresent( Id.class ) ||
+								( property.isAnnotationPresent( MapsId.class ) && !ignoreNotFound );
+
 				bindManyToOne(
 						getCascadeStrategy( ann.cascade(), hibernateCascade, false, forcePersist ),
 						joinColumns,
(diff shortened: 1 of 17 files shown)


hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
```diff
@@ -156,7 +195,7 @@ public Object hydrate(
 
 	@Override
 	public boolean isNullable() {
-		return foreignKeyType==ForeignKeyDirection.TO_PARENT;
```
JWGmeligMeyling: So my understanding of the `isNullable` method and how it is used is: can we safely assume, that a non-null identifier or unique key property for this association, will yield a non-null value and can it therefore be proxied?

For ManyToOnes this is rather trivial, as they are never mapped by a `PrimaryKeyJoinColumn`, but rather a `JoinColumn`, which is assumed to be `constrained` by default, which means that values must be present in the join table. The only exception being legacy schema's, with different magic values representing missing values. In this case you specify the `NotFound` action with regard to what should happen when an entity is not found under its  identifier or unique key value.

For OneToOne's this is somewhat more complicated. First, apparently, it seems in the end the `NotFound` annotation *was* ignored for `OneToOne` associations (something I had already indicated [here](https://github.com/hibernate/hibernate-orm/pull/2459#discussion_r205465299)). But another case comes with the `PrimaryKeyJoinColumn`. Even though there will always be a value for the primary key, we're not sure whether the row in our join table exists. This is indicated by the *constrainedness* or *optionality* of the `OneToOne` association. It is in my opinion irrelevant to check this based on the foreign key direction. In fact: we have seen examples with bidirectional non-optionality. It can't be persisted with this cyclic dependency , but its certainly valid for some domains to assume that a value will always be there after it is persistent and sensible to support this in lazy loading.

Therefore, I am adamant that the nullability should be determined based on the constrained value and **not** the foreign key direction. This change must be applied before my test for issue [HHH-12842](https://hibernate.atlassian.net/browse/HHH-12842) passes.

I've rebased my changes from #2544 on your work:  https://github.com/JWGmeligMeyling/hibernate-orm/pull/new/HHH-12842-fix-4 , so that we can combine our two PRs.

JWGmeligMeyling: So my understanding of the `isNullable` method and how it is used is: can we safely assume, that a non-null identifier or unique key property for this association, will yield a non-null value and can it therefore be proxied?

For ManyToOnes this is rather trivial, as they are never mapped by a `PrimaryKeyJoinColumn`, but rather a `JoinColumn`, which is assumed to be `constrained` by default, which means that values must be present in the referenced table. The only exception being legacy schema's, with different magic values representing missing values. In this case you specify the `NotFound` action with regard to what should happen when an entity is not found under its  identifier or unique key value.

For OneToOne's this is somewhat more complicated. First, apparently, it seems in the end the `NotFound` annotation *was* ignored for `OneToOne` associations (something I had already indicated [here](https://github.com/hibernate/hibernate-orm/pull/2459#discussion_r205465299)). But another case comes with the `PrimaryKeyJoinColumn`. Even though there will always be a value for the primary key, we're not sure whether the row in our referenced table exists. This is indicated by the *constrainedness* or *optionality* of the `OneToOne` association. It is in my opinion irrelevant to check this based on the foreign key direction. In fact: we have seen examples with bidirectional non-optionality. It can't be persisted with this cyclic dependency , but its certainly valid for some domains to assume that a value will always be there after it is persistent and sensible to support this in lazy loading.

Therefore, I am adamant that the nullability should be determined based on the constrained value and **not** the foreign key direction. This change must be applied before my test for issue [HHH-12842](https://hibernate.atlassian.net/browse/HHH-12842) passes.

I've rebased my changes from #2544 on your work:  https://github.com/JWGmeligMeyling/hibernate-orm/pull/new/HHH-12842-fix-4 , so that we can combine our two PRs.

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
```diff
@@ -1809,12 +1819,30 @@ else if ( property.isAnnotationPresent( OneToOne.class ) ) {
...
+				// the association is optional.
+				// @OneToOne(optional = true) with @PKJC makes the association optional.
+				final boolean mandatory =
+						!ann.optional() ||
+								property.isAnnotationPresent( Id.class ) ||
+								( property.isAnnotationPresent( MapsId.class ) && !ignoreNotFound );
+				// An optional @OneToOne with @PKJC or @PKJCs has an implicit @NotFound(IGNORE)
+				// (see HHH-4982).
+				if ( hasPkjc && !mandatory ) {
+					ignoreNotFound = true;
```
JWGmeligMeyling: Perhaps this workaround is not needed after the alternative fix for the `isNullable()` method?

hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
gbadner: @JWGmeligMeyling, thanks very much for your feedback. I'll checkout your PR, then address the statements you made above.

gbadner: @JWGmeligMeyling, also, please rest assured that I fully intend to resolve your issue for 5.4.0.GA. I'm sorry I didn't have a chance to run your test before pushing the PR on Friday, and I'm sorry to hear that it didn't work in your case. 

Also, please know that I had a long, hard look at a lot of the related code and there are very well-thought out reasons for why I addressed this the way I did. That said, I certainly could have missed something in my reasoning.

JWGmeligMeyling: Hi @gbadner  , I know that if this issue was easy to solve, the two of us would have figured it out much sooner already. I see the hard work you've done on this (also apparent by the extensive number of test cases added in this PR), and I really appreciate your efforts! I don't want to exclude the possibility of my implementation having some new flaws, but I certainly gave it a lot of thought and research too.

Looking forward to your findings with my changes applied to your PR 😄

gbadner:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2612

PR #2610, opened 2018-10-27
 HHH-12842 rebased on  HHH-12436

JWGmeligMeyling:


JWGmeligMeyling:
Will probably be super-seeded by #2612. I'll close it after my review comments on that PR are addressed.

PR #2611, opened 2018-10-30
Fixes for DelayedPostInsertIdentifier 

Naros:
https://hibernate.atlassian.net/browse/HHH-13042
https://hibernate.atlassian.net/browse/HHH-13044
https://hibernate.atlassian.net/browse/HHH-13053

Once we review this, we can squash these commits and merge.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/action/internal/EntityAction.java
+++ b/hibernate-core/src/main/java/org/hibernate/action/internal/EntityAction.java
@@ -12,6 +12,7 @@ import org.hibernate.AssertionFailure;
 import org.hibernate.action.spi.AfterTransactionCompletionProcess;
 import org.hibernate.action.spi.BeforeTransactionCompletionProcess;
 import org.hibernate.action.spi.Executable;
+import org.hibernate.engine.spi.EntityEntry;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.event.service.spi.EventListenerGroup;
 import org.hibernate.event.service.spi.EventListenerRegistry;
@@ -100,7 +101,8 @@ public abstract class EntityAction
 	 */
 	public final Serializable getId() {
 		if ( id instanceof DelayedPostInsertIdentifier ) {
-			final Serializable eeId = session.getPersistenceContext().getEntry( instance ).getId();
+			final EntityEntry entry = session.getPersistenceContext().getEntry( instance );
+			final Serializable eeId = entry == null ? null : entry.getId();
 			return eeId instanceof DelayedPostInsertIdentifier ? null : eeId;
 		}
 		return id;
--- a/hibernate-core/src/main/java/org/hibernate/action/internal/EntityAction.java
+++ b/hibernate-core/src/main/java/org/hibernate/action/internal/EntityAction.java
@@ -12,7 +12,6 @@ import org.hibernate.AssertionFailure;
 import org.hibernate.action.spi.AfterTransactionCompletionProcess;
 import org.hibernate.action.spi.BeforeTransactionCompletionProcess;
 import org.hibernate.action.spi.Executable;
-import org.hibernate.engine.spi.EntityEntry;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.event.service.spi.EventListenerGroup;
 import org.hibernate.event.service.spi.EventListenerRegistry;
@@ -101,19 +100,7 @@ public abstract class EntityAction
 	 */
 	public final Serializable getId() {
 		if ( id instanceof DelayedPostInsertIdentifier ) {
-			final EntityEntry entry = session.getPersistenceContext().getEntry( instance );
-			if ( entry == null ) {
-				if ( LOG.isDebugEnabled() ) {
-					LOG.debugf(
-							"Skipping action - the persistence context does not contain any entry for the entity [%s]. This may occur if an entity is created and then deleted in the same transaction/flush.",
-							instance
-					);
-				}
-				// If an Entity is created and then deleted in the same Transaction, when Action#postDelete() calls this method the persistence context
-				// does not contain anymore an entry.
-				return null;
-			}
-			final Serializable eeId = entry.getId();
+			final Serializable eeId = session.getPersistenceContext().getEntry( instance ).getId();
 			return eeId instanceof DelayedPostInsertIdentifier ? null : eeId;
 		}
 		return id;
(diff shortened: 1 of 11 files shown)


hibernate-core/src/main/java/org/hibernate/action/internal/EntityAction.java
```diff
@@ -101,19 +100,7 @@ public String getEntityName() {
...
-							instance
-					);
-				}
-				// If an Entity is created and then deleted in the same Transaction, when Action#postDelete() calls this method the persistence context
-				// does not contain anymore an entry.
-				return null;
-			}
-			final Serializable eeId = entry.getId();
+			final Serializable eeId = session.getPersistenceContext().getEntry( instance ).getId();
 			return eeId instanceof DelayedPostInsertIdentifier ? null : eeId;
```
gbadner: Is it possible for `session.getPersistenceContext().getEntry( instance )` to be `null`? If so, NPE will be thrown.

hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
```diff
@@ -1044,6 +1048,11 @@ public int getQueryStatisticsMaxSize() {
 		return queryStatisticsMaxSize;
 	}
 
+	@Override
+	public boolean isDelayedIdentityInsertsEnabled() {
+		return delayedIdentityInsertEnabled;
+	}
```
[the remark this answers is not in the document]
gbadner: Will this default work consistently with connection release mode `ConnectionReleaseMode.AFTER_STATEMENT`? See HHH-13050.  HHH-13050 could probably be resolved by setting `hibernate.id.disable_delayed_identity_inserts=true`. If it is desirable to delay the inserts, a solution for  `ConnectionReleaseMode.AFTER_STATEMENT` may be to skip aggressive release if a batch still has statements; for example, via `ResourceRegistryStandardImpl#hasRegisteredResources`.

hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
```diff
@@ -187,7 +188,9 @@ protected Serializable performSave(
 					throw new NonUniqueObjectException( id, persister.getEntityName() );
 				}
 			}
-			persister.setIdentifier( entity, id, source );
+			if ( ! DelayedPostInsertIdentifier.class.isInstance( id ) ) {
+				persister.setIdentifier( entity, id, source );
+			}
```
gbadner: The above is in an `if ( !useIdentityColumn ) ` block. Is it possible for `useIdentityColumn` to be `false` but `DelayedPostInsertIdentifier.class.isInstance( id )` is `true`? If so, it sounds like that would be a bug.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -4148,13 +4158,53 @@ private String substituteBrackets(String sql) {
...
+			// if the persister's identifier is assigned by insert, we need to see if we must force non-delay mode.
+			for ( NonIdentifierAttribute attribute : getEntityMetamodel().getProperties() ) {
+				final Type attributeType = attribute.getType();
+				if ( attributeType.isAssociationType() ) {
+					if ( attributeType.isEntityType() ) {
+						// In this use case, the association check is to see if we self-reference.
+						Class<?> entityClass = attributeType.getReturnedClass();
+						if ( getMappedClass().equals( entityClass ) ) {
+							canIdentityInsertBeDelayed = false;
+						}
```
gbadner: Is this a problem only when the attribute is actually a reference to the entity being persisted? 

What about an association within an embeddable that refers to the entity being persisted? It doesn't seem like the check would work for that use case.

What if the association is a reference to a different entity of the same class? Could/should it be possible to delay the insert in this case? I suppose that would require the persister to track which attributes (or sub-attributes) are of the same class, then they would have to be checked to see if any refer to the same entity being persisted. Maybe this could be an improvement fixed later?

hibernate-core/src/main/java/org/hibernate/action/internal/EntityAction.java
Naros: Since it's technically possible, I'll revert back to the null check for safety.

hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
Naros: You're right, this was unnecessary.  I will remove the if-check.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
Naros: > Is this a problem only when the attribute is actually a reference to the entity being persisted?

Yes

> What about an association within an embeddable that refers to the entity being persisted? It doesn't seem like the check would work for that use case.

Good catch and this does not work; I'll fix that use case too.

> Maybe this could be an improvement fixed later?

I think so, that's the main reason why we decided to introduce the configuration parameter for situations where we may have missed a use case that users could report but have a way to work around the problem in the short-term until we could revise the criteria this method uses.

hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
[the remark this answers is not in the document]
Naros: The configuration option has no impact on the outcome of HHH-13050.  From my investigation, it doesn't seem to be related to what this PR is trying to solve but instead a change we introduced between 5.1 and 5.3.  

I'll work on HHH-13050 separately from this PR.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -1992,4 +1992,19 @@
...
+	 * an identity-based column can be delayed until flush/commit or if the statement must be
+	 * executed early because access to the identity column is needed.
+	 * </p>
+	 * In the event that those defined rules are insufficient for a given mapping scenario, this
+	 * setting can be set to {@code true} to permanently disable delayed identity inserts for all
+	 * transactions as a short-term workaround.
+	 * </p>
+	 * The default value is {@code false}.
+	 */
+	String DISABLE_DELAYED_IDENTITY_INSERTS = "hibernate.id.disable_delayed_identity_inserts";
```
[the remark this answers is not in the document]
sebersole: Yes, it is any generator that generates identifiers as part of the insert statement sent to the database.  

However, I think the alternative names would probably be more confusing.  `DISABLE_DELAYED_POST_INSERT_IDENTIFIER_INSERTS` - meh.

This name is fine for me, unless someone has a better name suggestion.

gsmet:
@Naros a test is failing with your latest changes:
```
org.hibernate.test.cut.CompositeUserTypeTest > classMethod FAILED
    org.hibernate.testing.junit4.CallbackException
        Caused by: java.lang.ClassCastException
```

gsmet:
Rebased and squashed the last commit into the preceding one as it was a clear fix.

hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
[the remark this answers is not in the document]
gbadner: @Naros, ok, thanks for checking.

gbadner:
Merged. Thanks @Naros!

PR #2613, opened 2018-10-31
Hhh 10891

skeshavan:
We were using a very old version of hibernate & moving to the latest version broke the code as described in the bug.

The (if) code block that handles Any type in the iterator's next method has no data dependency on the joinable variable. By removing this unnecessary dependency, we stopped getting the error "any types do not have a unique referenced persister error" during hibernate load.

The simple code refactoring fix made all of our existing code to work as was working earlier (and we use Any extensively). Please note that we use polymorphic attributes only for persistence & not for retrieval. We use JDBC for retrieval (because of some application specific customization).

Therefore, I am unsure of the completeness of the fix (the retrieval of polymorphic attributes in components annotated with Embeddable is not tested).


gsmet:
Added a test case and merged. Thanks!

PR #2615, opened 2018-10-31
HHH-11096 - @CreationTimestamp doesn't works with @Column(nullable=false)

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-11096


hibernate-core/src/main/java/org/hibernate/engine/internal/Nullability.java
```diff
@@ -95,10 +94,16 @@ public void checkNullability(
...
+			final InMemoryValueGenerationStrategy[] inMemoryValueGenerationStrategies =
+					persister.getEntityMetamodel().getInMemoryValueGenerationStrategies();
+
+			boolean validValueGenerationStrategyCount = inMemoryValueGenerationStrategies.length == values.length;
 
 			for ( int i = 0; i < values.length; i++ ) {
 
-				if ( checkability[i] && values[i]!= LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
+				if ( checkability[i] && values[i] != LazyPropertyInitializer.UNFETCHED_PROPERTY &&
+						( validValueGenerationStrategyCount && GenerationTiming.NEVER == inMemoryValueGenerationStrategies[i]
```
JWGmeligMeyling: Are you sure you want this expression to return `false` if `! validValueGenerationStrategyCount `?

Otherwise:
```suggestion
						( !validValueGenerationStrategyCount || GenerationTiming.NEVER == inMemoryValueGenerationStrategies[i]
						
```

What kind of an odd situation by the way, to have `inMemoryValueGenerationStrategies.length == values.length`...

hibernate-core/src/main/java/org/hibernate/engine/internal/Nullability.java
```diff
@@ -95,10 +94,16 @@ public void checkNullability(
 					? persister.getPropertyInsertability()
 					: persister.getPropertyUpdateability();
 			final Type[] propertyTypes = persister.getPropertyTypes();
+			final InMemoryValueGenerationStrategy[] inMemoryValueGenerationStrategies =
```
JWGmeligMeyling: Isn't there the technical possibility to have non-nullable fields with a `InDatabaseValueGenerationStrategy`?

hibernate-core/src/main/java/org/hibernate/engine/internal/Nullability.java
vladmihalcea: In fact, we don't need the `validValueGenerationStrategyCount` at all. The problem was the `CustomPersister` using the wrong `Custom.hbm.xml``mapping.

JWGmeligMeyling: Cool! Thanks for verifying this!

hibernate-core/src/main/java/org/hibernate/engine/internal/Nullability.java
JWGmeligMeyling: Youre right! Didnt think of that :)

PR #2616, opened 2018-11-01
HHH-13070 perform injections on listener when using REPLACE_ORIGINAL strategy (5.3)

dohertyfjatl:
- Previously, the CallbackRegistryConsumer/JpaBootstrapSensitive injections were not performed on new listeners that were appended using duplication strategy with REPLACE_ORIGINAL, which could lead to NullPointerException when the listener was invoked (for example for a listener which is a CallbackRegistryConsumer, as the callbackRegistry field was not set)
- Added calls to checkAgainstBaseInterface and performInjections on the listener, in the case of appending a listener which replaces an existing one
- Added test cases to verify that the callbackRegistry is being injected


gsmet:
Nice contribution, thanks!

I just merged your work.

PR #2617, opened 2018-11-01
HHH-13071: use SourceSetOutput.getClassesDirs to prevent deprecation warnings

xehonk:
use SourceSetOutput.getClassesDirs to prevent deprecation warnings during build when using the gradle plugin


gsmet:
Merged, thanks!

PR #2618, opened 2018-11-01
added st_makeenvelope

berndhopp:
st_makeenvelope is a commonly used function in postgis, so I think it should be supported. st_makeenvelope is necessary for example to make queries that only include items that are inside of the bounding box of a map that the users sees in his browser. 

https://postgis.net/docs/ST_MakeEnvelope.html

### DIFF
--- a/hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/postgis/PostgisFunctions.java
+++ b/hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/postgis/PostgisFunctions.java
@@ -46,6 +46,11 @@ class PostgisFunctions extends SpatialFunctionsRegistry {
 				)
 		);
 		put(
+				"makeenvelope", new StandardSQLFunction(
+						"st_makeenvelope"
+				)
+		);
+		put(
 				"astext", new StandardSQLFunction(
 						"st_astext",
 						StandardBasicTypes.STRING


gsmet:
Hi @maesenka ,

Could you take a look at this one? It looks like a valuable addition but I'm wondering if it should be declared in `SpatialFunction` instead?

FYI, we're not far from releasing 5.4.0.CR1 and having this one in would be nice.

It at least requires a proper JIRA issue :).

berndhopp:
@gsmet Hi, do you want me to open a jira-issue for this one?

JWGmeligMeyling:
@berndhopp Yes, all PRs need a Jira issue.

berndhopp:
[jira-issue](https://hibernate.atlassian.net/browse/HHH-13083)

gsmet:
@JWGmeligMeyling not saying it's not at the right place in the PostGIS file but you'll notice that `envelope` is also defined in `SpatialFunction`.

I can't find it in the OpenGIS documentation though, even if I think it should be a standard one as it has the `ST_` prefix.

We would also need a test case for it, probably in `TestSpatialFunctions`.

@berndhopp I rebased, amended your commit message and force pushed to your branch.

I'll ping Karel so he can take a look at it, I don't have the bandwidth to drive this puppy home myself.

maesenka:
I'm on holiday right now. I'll have a look at this when I'm back in the
last week of November.

maesenka:
I merged this after rebase, cleanup and fixing of integration tests in d1d1f9f312ef5e3b74dfa935ff8ce83bd835852f

PR #2620, opened 2018-11-02
Choose the right dialect for Sybase SQL Anywhere

bobman38:
On SQL Anywhere 17.0.0.1062 the product name is "SQL Anywhere" and not "Adaptive Server Anywhere". I suggest to include this string in the test to select the right dialect.


bobman38:
Sorry I have miss the Jira case creation. Here is the case : https://hibernate.atlassian.net/browse/HHH-13078

THANKS !

gsmet:
No worries, I had planned to do the paperwork today but even better that you created the JIRA.

Amended the commit message and merged, thanks! It will be part of 5.4.0, we plan to release soon.

PR #2621, opened 2018-11-06
HHH-13084 : Querying entity with non-ID property named 'id' fails if entity has an IdClass composite key

gbadner:

This fix specifically does not fix the case where an entity has a non-ID property named "id" and a composite ID that does not have a primary key class. The user doc says that composite keys without a primary key class is deprecated, so I assume it's OK to not cover that case.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
@@ -25,6 +25,7 @@ import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.internal.log.DeprecationLogger;
 import org.hibernate.internal.util.collections.ArrayHelper;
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
```diff
@@ -685,7 +687,27 @@ public Type toType(String propertyName) throws QueryException {
...
+							String.format(
+									"[%s] has a composite key and a non-ID property named 'id'; " +
+											"Hibernate only supports this use case when " +
+											"the composite ID is mapped using a primary key class.",
+									getEntityPersister().getEntityName()
+							)
+
+					);
+				}
+			}
```
gsmet: I'm surprised this logic is only valid for this use case? All the other `EntityPersister#getIdentifierPropertyName()` calls in the code base are safe?

gbadner:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2655.

PR #2622, opened 2018-11-07
HHH-11209 : Change log message to DEBUG when due to rollback

gbadner:
Logging a warning when a collection with queued operations is detached due to rollback is too noisy. On rollback, the application should throw away data from the session because it is not consistent with persisted data, so there really isn't a need for this warning. Instead, it will be logged as a DEBUG message.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/AbstractPersistentCollection.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/AbstractPersistentCollection.java
@@ -38,6 +38,7 @@ import org.hibernate.internal.util.collections.IdentitySet;
 import org.hibernate.persister.collection.CollectionPersister;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.pretty.MessageHelper;
+import org.hibernate.resource.transaction.spi.TransactionStatus;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/collection/internal/AbstractPersistentCollection.java
```diff
@@ -625,7 +626,20 @@ public final boolean unsetSession(SharedSessionContractImplementor currentSessio
...
+					final String collectionInfoString = MessageHelper.collectionInfoString( getRole(), getKey() );
+					final TransactionStatus transactionStatus =
+							session.getTransactionCoordinator().getTransactionDriverControl().getStatus();
+					if ( transactionStatus.isOneOf(
+							TransactionStatus.ROLLED_BACK,
+							TransactionStatus.MARKED_ROLLBACK,
+							TransactionStatus.FAILED_COMMIT,
+							TransactionStatus.FAILED_ROLLBACK,
+							TransactionStatus.ROLLING_BACK
+					) )
```
gsmet: You need `{}` for the `if` here, otherwise Checkstyle fails.

gbadner:
Fixed the checkstyle failure and commit comment typo, and merged.

PR #2623, opened 2018-11-07
HHH-13085 : Use FetchMode.JOIN to load OneToOne associations that hav…

gbadner:
…e the same AssociationKey as the entity owner

https://hibernate.atlassian.net/browse/HHH-13085

I still need to look at this more deeply and create a test case.


gbadner:
Closing this PR because https://github.com/hibernate/hibernate-orm/pull/3488 is a better solution.

PR #2625, opened 2018-11-07
HHH-12850 Sybase Anywhere - Explicitly map Java boolean to SQL bits

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12850


gsmet:
Merged.

PR #2626, opened 2018-11-08
HHH-13090 Depend on Ehcache interface instead of depending on one of the implem…

mykolap:
Depend on Ehcache interface instead of depending on one of the implementation (for example BlockingCache is not working in current version)
Current ehcache2 integration uses one of the implementation of Ehcache named net.sf.ehcache.Cache

But I am using BlockingCache from net.sf.ehcache.constructs.blocking package, 
that extends EhcacheDecoratorAdapter and is an implementation of Ehcache, 
but doen't extend net.sf.ehcache.Cache 

I think it's better appoach depending on interfaces, not implementations


gsmet:
Indeed, nice contribution.

I created a JIRA issue here: https://hibernate.atlassian.net/browse/HHH-13090 and fixed a couple of formatting issues.

Will run the tests and merge.

gsmet:
Merged, thanks. Will be included in 5.4.0.Final released next week.

PR #2627, opened 2018-11-08
HHH-13092 Upgrade Hibernate Commons Annotations to 5.1.0.Final

gsmet:
* https://hibernate.atlassian.net/browse/HHH-13092

CI will fail as the artifact is not on Central yet but this way I won't forget about it.


gsmet:
Merged.

PR #2628, opened 2018-11-08
HHH-13050 - On release of batch it still contained JDBC statements logged

Naros:
https://hibernate.atlassian.net/browse/HHH-13050

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/AbstractBatchImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/AbstractBatchImpl.java
@@ -187,12 +187,17 @@ public abstract class AbstractBatchImpl implements Batch {
 		}
 	}
 
+	protected int getBatchPosition(){
+		return 0;
+	}
+
+
 	@Override
 	public void release() {
-		if ( getStatements() != null && !getStatements().isEmpty() ) {
+		if ( getStatements() != null && !getStatements().isEmpty() && getBatchPosition() != 0 ) {
 			LOG.batchContainedStatementsOnRelease();
+			releaseStatements();
 		}
-		releaseStatements();
 		observers.clear();
 	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchingBatch.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchingBatch.java
@@ -148,4 +148,9 @@ public class BatchingBatch extends AbstractBatchImpl {
 			getKey().getExpectation().verifyOutcome( rowCounts[i], ps, i );
 		}
 	}
+
+	@Override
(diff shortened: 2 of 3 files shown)


Naros:
@gbadner I integrated @dreab8 orm-proper test here that replicated the issue and his fix he proposed.  Feel free to let us know what you think.

gbadner:
I was mainly concerned about what would happen in [`AbstractBatchImpl#releaseStatements`](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/AbstractBatchImpl.java#L154-L161) if there were more than 1 statement in the batch. It looked to me that, in the first time through the loop, the connection would be released, and the other statements might not be processed properly.

I see a [comment](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchingBatch.java#L32) in BatchingBatch that says:

`"// IMPL NOTE : Until HHH-5797 is fixed, there will only be 1 statement in a batch"`

I see that HHH-5797 has been fixed, but after spending (way too much time), I see that batches still can only have one statement.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/AbstractBatchImpl.java
```diff
@@ -187,12 +187,17 @@ protected final void notifyObserversImplicitExecution() {
...
+
 	@Override
 	public void release() {
-		if ( getStatements() != null && !getStatements().isEmpty() ) {
+		if ( getStatements() != null && !getStatements().isEmpty() && getBatchPosition() != 0 ) {
 			LOG.batchContainedStatementsOnRelease();
+			releaseStatements();
 		}
-		releaseStatements();
 		observers.clear();
```
gbadner: I've pushed a [branch](https://github.com/gbadner/hibernate-core/tree/HHH-13050_naros_2) that includes a test that @dreab8 created to show that there is no statement leak.

I've added a [commit](https://github.com/gbadner/hibernate-core/commit/8c29a448d04ee37af7647ac31c139f9abb5d10c7) that adds a test with a custom `Batch` implementation similar to what @dreab8 did in his test.  In this new test, the custom batch throws an exception when `#addToBatch` is called. `#releaseStatements` does not get called because the batch gets aborted with `batchPosition` == 0. As a result the batch will still contain a `PreparedStatement` after `#release` is called. The test confirms that it is closed.

dreab8:
@gbadner I think in the new test you added in your branch, the exception thrown by the `TestBatch#addToBatch()` method prevents the creation of any `PreparedStatement` so there is no need to call the `releaseStatements()` method in such a case.

gbadner:
@Naros, @dreab8, yet another approach: https://github.com/hibernate/hibernate-orm/pull/2641

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/2643

PR #2629, opened 2018-11-09
HHH-12945

gbadner:
Follows up on https://github.com/hibernate/hibernate-orm/pull/2590.

Removes SelectionImplementor, changes MapEntryExpression to implement CompoundSelection, and restores the original validation code in CriteriaQueryImpl.

FYI, I'm not an expert on this area of the code. Maybe @sebersole or @dreab8 could take a look?


hibernate-core/src/main/java/org/hibernate/query/criteria/internal/expression/MapEntryExpression.java
```diff
@@ -25,41 +27,41 @@
...
-				+ '.'
-				+ ( (Renderable) getAttribute() ).render( renderingContext );
+	@Override
+	public boolean isCompoundSelection() {
+		return true;
 	}
+
+	@Override
+	public List<Selection<?>> getCompoundSelectionItems() {
+		return Arrays.asList( original.key(), original.value() );
```
gsmet: Force pushed to add an `@Override` here.

I remember now why I didn't implement it as a `CompoundSelection`, I didn't know how to implement this method.

It certainly looks better this way.

dreab8:
@gbadner I'm not a great expert of Criteria code but it looks good to me.

gbadner:
@gsmet, is it ok to merge this?

PR #2630, opened 2018-11-09
HHH-13076 - Hibernate “Transaction already active” behaviour with custom transaction manager

dreab8:
https://hibernate.atlassian.net/browse/HHH-13076

The fix reintroduces the old behaviour for non JPA transaction compliance settings and JTA Transactions.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/transaction/internal/TransactionImpl.java
@@ -72,7 +72,13 @@ public class TransactionImpl implements TransactionImplementor {
 
 		// per-JPA
 		if ( isActive() ) {
-			throw new IllegalStateException( "Transaction already active" );
+			if ( jpaCompliance.isJpaTransactionComplianceEnabled()
+					|| !transactionCoordinator.getTransactionCoordinatorBuilder().isJta() ) {
+				throw new IllegalStateException( "Transaction already active" );
+			}
+			else {
+				return;
+			}
 		}
 
 		LOG.debug( "begin" );
--- a/hibernate-core/src/test/java/org/hibernate/test/resource/transaction/jdbc/AlreadyStartedTransactionTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/resource/transaction/jdbc/AlreadyStartedTransactionTest.java
@@ -1,0 +1,37 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.test.resource.transaction.jdbc;
+
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+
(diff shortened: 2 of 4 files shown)


vladmihalcea:
@dreab8 @sebersole The commit looks good, especially from a JPA-compliance perspective.

However, for the non-compliance mode, I think we should not allow using the `EntityTranscation when running in JTA mode.

The JPA spec says that:

> The EntityTransaction interface is used to control resource transactions on resource-local entity
> managers. The EntityManager.getTransaction() method returns an instance of the
> EntityTransaction interface.
> When a resource-local entity manager is used, and the persistence provider runtime throws an exception
> defined to cause transaction rollback, the persistence provider must mark the transaction for rollback.
> If the EntityTransaction.commit operation fails, the persistence provider must roll back the
> transaction

So, the `EntityTransaction` is only meant for resource-local Txs. For JTA, the commit should be done via the `UserTranscation` JTA interface, which happens either manually or automatically by the EJB or Spring `TranscationInterceptor`.

That being said, I wonder whether we should disallow using the `EntityTransaction` in JTA mode.

dreab8:
@vladmihalcea I think the point is if we want to maintain the 5.1 behaviour or not.

vladmihalcea:
I understand that. My question was whether we should push the JPA compatibility even further for transaction management. We could create a new Jira issue for that.

sebersole:
We understand what the specs says. This is simply behavior Hibernate has had forever, and honestly it's more sane imo from the perspective of the user just having to use one contract regardless of the environment.  And the fact that we still protect this via JpaCompliance allows the behavior you are looking for (obviously, we need that for the TCK).

I'm just not seeing a reason to remove this behavior. Are you saying this just because of the spec?

vladmihalcea:
I was thinking of disallowing the `EntityTransaction` interactions when running in JTA mode while also using JpaCompliance=true. But then, the TCK pases just fine even without this restriction, so I get your point.

Anyway, for most of our users, transaction management is done either by Spring or the Java EE container. Manual transaction management is usually the exception than the rule. For this reason as well as backward compatibility, it's probably better to just leave it as-is.

dreab8:
@vladmihalcea  do you mean this https://github.com/hibernate/hibernate-orm/pull/2630/files#diff-d263ca999c94a30d95f67aba73828abfR44 ?

vladmihalcea:
@dreab8 Actually, it's more than that. The `EntityTransaction` was designed to be used for RESOURCE_LOCAL, so using it when running in JTA mode would bypass the JTA transaction handling which requires that the XA transaction outcome should be made via the JTA `UserTransaction` contract. 

So, it was not about this particular PR. It was just a question about interpreting the JPA and JTA spec.

dreab8:
@vladmihalcea with JTA and JPA compliance calling` EntityManager#getTransaction()` causes and `IllegalStateException`.

vladmihalcea:
@dreab8 That's exactly what I was talking about. Thanks for the clarification.

dreab8:
@vladmihalcea you are welcome :)

dreab8:
@sebersole can I merge this into master?

vladmihalcea:
@sebersole I'm good. Andrea made all the clarifications.

dreab8:
Rebased and merged into master.
@sebersole @vladmihalcea thanks for the review

PR #2634, opened 2018-11-12
HHH-13077 - Optimize query plan call count

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13077


vladmihalcea:
@sebersole If you have time, can you take a look on this PR. This is a "second" attempt, as I reverted the initial commit from the master. For more details, you can check [the original PR](https://github.com/hibernate/hibernate-orm/pull/2624). Thanks.

sebersole:
Applied upstream on master.  Thanks!

vladmihalcea:
@sebersole Cool. Thanks for integrating it.

gavinking:
This change broke HR. We need to figure out what to do about that.

PR #2637, opened 2018-11-13
HHH-13094 - Respect @Any.fetch setting to FetchType.EAGER

johnlinp:


johnlinp:
Hi @gsmet would you give me some comments? Thank you.

hibernate-core/src/main/java/org/hibernate/TypeHelper.java
```diff
@@ -105,5 +105,5 @@
 	 *
 	 * @return The type, or null
 	 */
-	public Type any(Type metaType, Type identifierType);
+	public Type any(Type metaType, Type identifierType, boolean lazy);
```
vladmihalcea: This change brings is not backward compatible so it can not be done like this. A new default method must be used instead.

PR #2641, opened 2018-11-14
HHH-13050 : On release of batch it still contained JDBC statements logged; unable to release batch statement

gbadner:
https://hibernate.atlassian.net/browse/HHH-13050

BTW, I squashed my fix commit into the commit from @dreab8, so now the commits appear to be out of order.

IMHO, the batch code is complicated by having a `#abortBatch`, `#release`, and `#releaseStatements`. 

When `#releaseStatements` is called with `AFTER_STATEMENT` connection release mode,  `jdbcCoordinator.afterStatementExecution()` ends up calling `#release`, which calls back to `#releaseStatements`.

When `#abortBatch` gets called, `#release` gets called directly, which calls `#releaseStatements`. If AFTER_STATEMENT connection release mode used, then `jdbcCoordinator.afterStatementExecution()` ends up calling back to `#release`.

IMO, adding a flag (e.g., `releaseStatementsProcessInProgress`) adds more clutter to this overly-complicated code. I think it would be better to improve this code so it's not so complicated.

The original issue here is that with AFTER_STATEMENT connection release mode, `AbstractBatchImpl#clearBatch` gets called after after the connection is released, which ends up calling `PreparedStatement#clearBatch` on a closed statement, resulting in a `SQLException`.

My approach, for now, is to simply to skip calling `PreparedStatement#clerBatch` if the `PreparedStatement` is already closed.


hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchingBatch.java
```diff
@@ -77,6 +77,7 @@ public void addToBatch() {
 			currentStatement.addBatch();
 		}
 		catch ( SQLException e ) {
+			abortBatch();
```
gbadner: The above fixes an apparent oversight.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -3174,9 +3175,8 @@ protected void insert(
 									.executeUpdate( insert ), insert, -1
 					);
 				}
-
 			}
-			catch (SQLException e) {
+			catch (SQLException | JDBCException e) {
```
gbadner: The added `JDBCException` should only be needed if an application provides its own `BatchBuilder` that builds a `Batch` that implements `#addToBatch` that does not call `#abortBatch` when it fails.

`JtaWithFailingBatchTest` throws `JDBCException`, which is handled here.

FWIW, there are lots of other places where this could happen.

dreab8:
@gbadner It looks fine to me, not sure if it would be even better to add this check to the `releaseStatements() `method
```
protected void releaseStatements() {
		try {
			for ( PreparedStatement statement : getStatements().values() ) {
				try {
					if ( !statement.isClosed() ) {
						clearBatch( statement );
						jdbcCoordinator.getResourceRegistry().release( statement );
						jdbcCoordinator.afterStatementExecution();
					}
				}catch (SQLException sql){
					LOG.unableToReleaseBatchStatement();
				}
			}
		}
		finally {
			getStatements().clear();
		}
	}
```

gbadner:
@dreab8, I think your idea is better than what I did in `AbstractBatchImpl#clearBatch`.

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/2643

PR #2642, opened 2018-11-14
HHH-13100 Make a couple of equals()/hashCode() implementations lighter

gsmet:


hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/PersistentAttributeTransformer.java
```diff
@@ -331,16 +331,11 @@ public boolean equals(final Object o) {
 			return false;
 		}
 		final PersistentAttributeTransformer that = (PersistentAttributeTransformer) o;
-		return Objects.equals( managedCtClass, that.managedCtClass ) &&
-			Objects.equals( enhancementContext, that.enhancementContext ) &&
-			Objects.equals( classPool, that.classPool ) &&
-			Arrays.equals( enhancedFields, that.enhancedFields );
```
gsmet: Here the `enhancedFields()` comes directly from the class. See https://github.com/hibernate/hibernate-orm/pull/2642/files#diff-a15011f40d71b98c54f55e9275b81bc9R71 .

Sanne: ah, so we have only one of these for each class?

+1 to remove context & pool.

Seems risky though to remove the fields from the equals though; maybe remove it from hashcode only?

gsmet: Dunno. At the moment, the object is instantiated from a static method taking only the class and the `enhancedFields` is a computed field.

I don't see how it could be different in the future.

But I won't fight for it (too much) :)

Sanne:
merged

PR #2645, opened 2018-11-14
HHH-13104 - Oracle 12c / SAP Hana insert fails when entity contains only an identity-based column.

Naros:
https://hibernate.atlassian.net/browse/HHH-13104

I stumbled onto this problem when fixing a test for HHH-13053 where the entity contained a single column that was an identity-based column.  While Oracle 12c seems to have a solution, it seems SAP Hana does not based on my research (unless I overlooked something).  Therefore SAP Hana throws a `MappingException` if we determine a mapping falls into this scenario rather than only happening during the insert at runtime.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/AbstractHANADialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/AbstractHANADialect.java
@@ -1648,4 +1648,9 @@ public abstract class AbstractHANADialect extends Dialect {
 	public boolean supportsJdbcConnectionLobCreation(DatabaseMetaData databaseMetaData) {
 		return false;
 	}
+
+	@Override
+	public String getNoColumnsInsertString() {
+		throw new MappingException( "SAP HANA requires at least one value in insert value-list clause." );
+	}
 }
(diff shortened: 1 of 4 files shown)


Naros:
I stumbled onto this problem when fixing a test for HHH-13053 where the entity contained a single column that was an identity-based column.  While Oracle 12c seems to have a solution, it seems SAP Hana does not based on my research (unless I overlooked something).  Therefore SAP Hana throws a `MappingException` if we determine a mapping falls into this scenario rather than only happening during the insert at runtime.

hibernate-core/src/main/java/org/hibernate/dialect/AbstractHANADialect.java
```diff
@@ -1648,4 +1648,9 @@ public boolean supportsNoWait() {
 	public boolean supportsJdbcConnectionLobCreation(DatabaseMetaData databaseMetaData) {
 		return false;
 	}
+
+	@Override
+	public String getNoColumnsInsertString() {
+		throw new MappingException( "SAP HANA requires at least one value in insert value-list clause." );
```
gsmet: @Naros I'm a bit annoyed by the fact that we only trigger the exception at runtime. I don't think it's worth adding a dialect method for such a corner case so I suppose we'll have to leave with it?

gsmet:
Merged, thanks!

PR #2646, opened 2018-11-15
HHH-13105 - Add registries for user-defined Types as currently a custom Type is instantiated and configured 9 times

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13105

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
@@ -8,11 +8,16 @@ package org.hibernate.type;
 
 import java.io.Serializable;
 import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 
 import org.hibernate.MappingException;
+import org.hibernate.SessionFactory;
+import org.hibernate.SessionFactoryObserver;
 import org.hibernate.classic.Lifecycle;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
```diff
@@ -52,6 +55,12 @@
 	private final TypeConfiguration typeConfiguration;
 	private final TypeScope typeScope;
 
+	private final Map<TypeKey, Type> typeRegistry = new HashMap<>();
```
sebersole: This was my concern about boot versus run time.  These Maps live on during run time even though most of them will never be used beyond boot time.  But this class is already deprecated in anticipation of 6 so it's probably not overly relevant

vladmihalcea: That's true. In fact, the Type is associated with the property in the entity Persister, and that's how it's used after boot.

I inspired myself from the basic type registry. If you think this is fine, I'll push it in order to speed up the bootstrapping a little bit.

vladmihalcea: I could also investigate whether I could use a bootstrap finished callback to clear these registers and free up memory. Do you think it's worth researching it a little bit?

vladmihalcea:
@sebersole I pushed a new version which allows us to clear the registries via the `SessionFactoryObserver`. Let me know what you think.

PR #2649, opened 2018-11-19
HHH-13103 - Allow Hibernate Types to get access to the current configuration properties

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13103

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
@@ -7,15 +7,19 @@
 package org.hibernate.type;
 
 import java.io.Serializable;
+import java.lang.reflect.Constructor;
 import java.util.Comparator;
+import java.util.Map;
 import java.util.Properties;
 
 import org.hibernate.MappingException;
 import org.hibernate.classic.Lifecycle;
+import org.hibernate.engine.config.spi.ConfigurationService;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.internal.util.ReflectHelper;
 import org.hibernate.tuple.component.ComponentMetamodel;
+import org.hibernate.type.spi.TypeBootstrapContext;
 import org.hibernate.type.spi.TypeConfiguration;
 import org.hibernate.type.spi.TypeConfigurationAware;
 import org.hibernate.usertype.CompositeUserType;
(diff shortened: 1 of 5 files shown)


vladmihalcea:
@sebersole The reason why I think it's worth to implement this feature is that we are currently configuring the Java and SQL Type Descriptors in the constructor of the `Type` instances. 

With this feature, we could allow the user to provide a specific constructor which takes the `TypeBootstrapContext` where we can pass the current `SessionFactory` properties or some other info we might need in future:

    public ArrayType(TypeBootstrapContext typeBootstrapContext) {
        super( VarcharTypeDescriptor.INSTANCE, ArrayTypeDescriptor.INSTANCE );
        this.settings = typeBootstrapContext.getConfigurationSettings();
    }

This way, we can get access to the current properties and the user can customize the custom `Type` object creation using the Hibernate properties.

One use case where I wanted this future is to customize the Jackson `ObjectMapper` used by the JSON Types. Currently, I can only do that via loading the `hibernate.properties` file externally, but not via the SessionFactory properties that might be passed by Spring or Spring Boot. This is [the original issue](https://github.com/vladmihalcea/hibernate-types/issues/35) that made me realize we could benefit from this new feature.

Let me know what you think.

cberg-zalando:
@vladmihalcea I guess you will have to resolve the merge conflict before this moves any forward.

vladmihalcea:
@cberg-zalando I no longer work for the Hibernate project, hence, I don't know what is the current status or in which branch contributions like this one have to be applied. If this one is not integrated into 5.4, then I'll provide a new PR for the 6.x branch once it becomes the main development branch.

vladmihalcea:
@dreab8 I rebased this PR. From the upvotes this PR got, I think it's very useful and worth integrating.

Do you think it could be integrated in the next release? It's a small change and shouldn't be risky at all.

hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
```diff
@@ -88,7 +92,24 @@ public Type byClass(Class clazz, Properties parameters) {
...
+
+			Constructor<Type> bootstrapContextAwareTypeConstructor = ReflectHelper.getConstructor(
+					typeClass,
+					TypeBootstrapContext.class
+			);
+			if ( bootstrapContextAwareTypeConstructor != null ) {
+				ConfigurationService configurationService = typeConfiguration.getServiceRegistry().getService(
+						ConfigurationService.class );
+				Map<String, Object> configurationSettings = configurationService.getSettings();
+				type = bootstrapContextAwareTypeConstructor.newInstance( new TypeBootstrapContext(
```
sebersole: Not a fan of creating a new `TypeBootstrapContext` every time.  I'd pass it the `TypeConfiguration` and be done with it, or have `TypeConfiguration` implement `TypeBootstrapContext` (after making it an interface - which it ought to be anyway imo)

hibernate-core/src/test/java/org/hibernate/test/type/contributor/ArrayType.java
```diff
@@ -33,4 +43,7 @@ public String getName() {
         return "comma-separated-array";
     }
 
+    public Map<String, Object> getSettings() {
```
sebersole: Not a fan of exposing these.  The whole point of the issue/PR is to allow Types to be settings-aware, not make the consumer of a Type settings-aware.  You expose it for a test, but I can guarantee with a fairly high-degree of certainty that someone will try to use this.  And at that consumer needs to protect for null every time (here for example with the static ref or non-parameterized ctor)

sebersole: Sorry, just realized this is a ArrayType from the tests, not the "main" one.

hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
sebersole: Or better yet, have TypeFactory be the TypeBootstrapContext

sebersole: I can work on this change

vladmihalcea: Thanks.

sebersole:
Applied upstream with discussed change

PR #2651, opened 2018-11-20
HHH-13105 - Add registries for user-defined Types as currently a custom Type is instantiated and configured 9 times

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13105

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
@@ -8,11 +8,16 @@ package org.hibernate.type;
 
 import java.io.Serializable;
 import java.util.Comparator;
+import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


gsmet:
This is an improvement, let’s do that in 6 instead if it still needs to be done, as per our discussions on the list. 

/cc @sebersole

vladmihalcea:
Cannot be integrated into 6.0 yet as the Type support is not fully functional.

sebersole:
Type should be close to fully functional now, albeit in a different way.  

Could you try porting and see if it works there?

vladmihalcea:
@sebersole I will try to see how it works once I get a little bit of spare time to do it.

dreab8:
Thanks @vladmihalcea,
 
I'm going to close this PR because the issue has been fixed.

PR #2653, opened 2018-11-22
HHH-11209 HHH-9933 (5.1)

gbadner:
https://hibernate.atlassian.net/browse/HHH-9933


vasanthan-m:
Hello,

Which version of 5.1.x jar would have this fix ? I see the last released version in 5.1.17 which does not have this fix in the repor [https://repo1.maven.org/maven2/org/hibernate/hibernate-core/].
Can you help with getting the 5.1.x version with this fix ?

PR #2654, opened 2018-11-22
HHH-13114 - Query "select count(h) from Human h" fails if a subclass has a non-Id property named "id"

dreab8:


dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/2655

PR #2655, opened 2018-11-22
HHH-13084 + HHH-13114

dreab8:


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElement.java
```diff
@@ -525,7 +531,7 @@ public CollectionPropertyReference getCollectionPropertyReference(String propert
 		return elementType.getCollectionPropertyReference( propertyName );
 	}
 
-	public String getIdentifierPropertyName() {
+	public String[] getIdentifierPropertyName() {
```
JWGmeligMeyling:
```suggestion
	public String[] getIdentifierPropertyNames() {
```

dreab8:
@JWGmeligMeyling Thanks for spotting it.

JWGmeligMeyling:
We have to rename the two or three uses of the method as well. Unfortunately I still cant suggest multi line changes through github :( I can however amend the commit if you wish.

dreab8:
@JWGmeligMeyling I'm doing it, Thanks

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
```diff
@@ -682,13 +683,25 @@ public Type toType(String propertyName) throws QueryException {
...
 	}
 
-	public String getIdentifierPropertyName() {
-		if ( getEntityPersister() != null && getEntityPersister().getEntityMetamodel() != null
-				&& getEntityPersister().getEntityMetamodel().hasNonIdentifierPropertyNamedId() ) {
-			return getEntityPersister().getIdentifierPropertyName();
-		}
-		else {
-			return EntityPersister.ENTITY_ID;
+	public String[] getIdentifierPropertyName() {
```
gsmet: @dreab8 I think you forgot to update this method name: it should be `getIdentifierPropertyNames()` for the sake of consistency.

Thanks!

dreab8: @gsmet 
you are right my mistake.
I'm fixing it
Thanks!

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
```diff
@@ -130,18 +129,20 @@ public Queryable getQueryable() {
...
-				buf.append( ", " );
+		for ( int j = 0; j < idPropertyName.length; j++ ) {
+			String propertyName = idPropertyName[j];
+			String[] toColumns = getPropertyMapping( propertyName ).toColumns( getTableAlias(), propertyName );
+			for ( int h = 0; h < toColumns.length; h++ ) {
+				String column = toColumns[h];
+				if ( j + h > 0 ) {
+					buf.append( ", " );
+				}
+				buf.append( column ).append( " as " ).append( NameGenerator.scalarName( i, j + h ) );
```
gbadner: Using `j+h` above can result in duplicated column aliases. For example, if `i == 0`, `idPropertyName.length == 2` and each `toColumns.length == 2`, then the column names that will be generated with be: `col_0_0_`, `col_0_1_`, `col_0_1`, `col_0_2`.

The column names should be `col_0_0_`, `col_0_1_`, col_0_2_`, `col_0_3_`.

I've created a new PR with a failing test case that illustrates this, and added a commit for the fix: 
https://github.com/hibernate/hibernate-orm/pull/2685

gbadner:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2685.

gbadner:
Thanks @dreab8 for your help on this!!!

PR #2656, opened 2018-11-22
HHH-13118 Use the provided class loader even when using the default URI

henri-tremblay:
* https://hibernate.atlassian.net/browse/HHH-13118


gsmet:
Looks good to me. Will merge and push it for 5.4.0.CR2.

FYI, I have opened https://hibernate.atlassian.net/browse/HHH-13046 a while ago as it's not the only issue we have with classloading here (there is a commit linked in a comment). Unfortunately, we exposed `JCacheHelper`.

gsmet:
And merged.

PR #2659, opened 2018-11-23
HHH-13111 Restore support of criteria subqueries in select clauses

gsmet:
When the JPA query compliance mode is disabled (the default), it should
be possible to include subqueries in select clauses of a criteria query.

This was previously accepted due to a bug. After having fixed HHH-13001, we
started to throw an IllegalStateException.

 * https://hibernate.atlassian.net/browse/HHH-13111


gsmet:
Thanks @JWGmeligMeyling and @yrodiere for the reviews, merged!

PR #2661, opened 2018-11-26
HHH-13115 - Allow define timezone per tenant when using Multitenant Database

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13115


gsmet:
Merged, thanks.

PR #2662, opened 2018-11-26
HHH-13120: Fix BeanValidationEventListener to validate dirty attribut…

mway-dmueller:
…es only, if dirty tracking is enabled via Bytecode Enhancement.


gsmet:
Unfortunately I don't think it's something we can do in a general purpose `BeanValidationEventListener`: it would ignore the potential class level constraints.

gsmet:
Closing.

PR #2663, opened 2018-11-27
fixed entity fields order

jy05568536:
fixed entity fields order

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/PropertyContainer.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/PropertyContainer.java
@@ -9,13 +9,7 @@
 
 package org.hibernate.cfg;
 
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.TreeMap;
+import java.util.*;
 import javax.persistence.Access;
 import javax.persistence.ManyToMany;
 import javax.persistence.ManyToOne;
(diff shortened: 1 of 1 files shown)


vladmihalcea:
First, we need a Jira issue to describe why you need this change. Second, we need some test case as well. Third, this change might be risky and might not be backward compatible.

gsmet:
The thing is that there is an order as we use a `TreeMap`. It's more stable as using the order in which the properties are discovered per reflection as this order is not guaranteed (it tends to be in recent versions of Java but there's still no guarantee).

What is the rationale of this change? Did you have an issue with the current order?

Naros:
I think its important to remember that column order is going to be irrelevant if you use the schema `update` option when you bootstrap Hibernate and change entity definitions.  

What I mean is, any new columns we detect in the entity mapping are added to the end of the table, regardless of order.  So you theoretically would end up with something like:

`| ID | ADDRESS | EMAIL | USERNAME | FAX_NUMBER * | PHONE_NUMBER * |`

In the above, columns marked with `*` were added after the table was created later in the entity life-cycle, so you can see that column order is somewhat irrelevant. 

That said, if this is something that people feel strongly they want changed, my suggestion would be a strategy that uses a `TreeMap` by default and user implementations can plug-in whatever they want as a container type to fit whatever goal they want.  But again that won't change the fact you'll run into the above scenario when you change an entity mapping that already exists.

gsmet:
@Naros I think he was more about having a consistent order between the entity and the metadata.

So the story is that, when using reflection in old JVM, the order of the properties was not kept and it was pretty much random. I'm pretty sure that's the reason we ended up with a `TreeMap`.

Lately the JVM tends to return the properties in order, but last time I checked, it was not something guaranteed, more like an implementation detail.

Maybe it has changed but as long as old JDK are concerned, the order is not guaranteed.

I'm going to close this PR. It could be revisited if someday the order of the properties returned by the JVM was fully guaranteed by all the JDK version we support.

PR #2664, opened 2018-11-27
HHH-13124 - The CachingRegionFactory resolution algorithm has changed

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13124


gsmet:
Merged to master. Seems like a sensible change for the next CR.

PR #2665, opened 2018-11-27
HHH-13125 - Remove the Javadoc links in the User Guide pointing internal classes

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13125

### DIFF
--- a/documentation/src/main/asciidoc/userguide/chapters/transactions/Transactions.adoc
+++ b/documentation/src/main/asciidoc/userguide/chapters/transactions/Transactions.adoc
@@ -166,13 +166,11 @@ Out-of-the-box, Hibernate comes with three implementations of this interface:
 `org.hibernate.context.internal.JTASessionContext`::
 current sessions are tracked and scoped by a `JTA` transaction.
 The processing here is exactly the same as in the older JTA-only approach.
-See the https://docs.jboss.org/hibernate/orm/{majorMinorVersion}/javadocs/org/hibernate/context/internal/JTASessionContext.html[Javadocs] for more details.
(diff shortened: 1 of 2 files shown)


documentation/src/main/asciidoc/userguide/chapters/transactions/Transactions.adoc
```diff
@@ -166,13 +166,11 @@ Out-of-the-box, Hibernate comes with three implementations of this interface:
 `org.hibernate.context.internal.JTASessionContext`::
 current sessions are tracked and scoped by a `JTA` transaction.
 The processing here is exactly the same as in the older JTA-only approach.
-See the https://docs.jboss.org/hibernate/orm/{majorMinorVersion}/javadocs/org/hibernate/context/internal/JTASessionContext.html[Javadocs] for more details.
```
gsmet: Maybe for this one and the one below, we should still indicate the class name with something like "See the `JTASessionContext` source code and Javadocs for more details". Otherwise, we lose a bit of information.

vladmihalcea: There's no Javadoc for classes in internal packages. The Gradle task simply excludes them. We could link the class on GitHub, but that should include the current branch we are referring like:

https://github.com/hibernate/hibernate-orm/blob/5.3/path-to-file

gsmet: Yeah, I knew that but I missed the fact that we referenced the class name just above so the user can find the code easily. Will merge.

gsmet:
Merged, thanks.

PR #2667, opened 2018-11-29
HHH-13102 - Document how catalog and schema attributes need to be applied based on the underlying DB capabilities

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13102


gsmet:
Merged, thanks.

PR #2669, opened 2018-11-29
HHH-13128 Add missing jaxb-runtime dependency to metamodel generator

gsmet:
It's necessary for JDK 11 and we missed it when we added the
jaxb-runtime dependency to hibernate-core.

 * https://hibernate.atlassian.net/browse/HHH-13128


gsmet:
Merged, thanks for the review.

PR #2671, opened 2018-11-29
HHH-13127 Document that JAXB jars are needed if using JDK 11

gsmet:
* https://hibernate.atlassian.net/browse/HHH-13127


gsmet:
Ok, thanks, merged.

PR #2673, opened 2018-11-30
HHH-13110 - @PreUpdate method on a Embeddable null on the parent caused NullPointerException

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13110


gsmet:
Pushed to master for 5.4.0.Final, thanks.

PR #2674, opened 2018-11-30
HHH-13133: print message about 'successfully enhanced class' as debug

fprochazka:
https://hibernate.atlassian.net/browse/HHH-13133


gsmet:
@fprochazka could you try 5.4.0.CR2? I fixed an issue with the BB enhancer that caused every single class to be logged. Now, it should only be the enhanced class and I think it makes sense to keep it as is.

gsmet:
Closing. As mentioned, I think it is fixed.

PR #2675, opened 2018-12-04
HHH-13139 - Restore StandardDialectResolver as non-final

quaff:
Developers can customize dialect resolver base on standard implementation

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/dialect/internal/StandardDialectResolver.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/dialect/internal/StandardDialectResolver.java
@@ -16,7 +16,7 @@ import org.hibernate.engine.jdbc.dialect.spi.DialectResolver;
  *
  * @author Steve Ebersole
  */
-public final class StandardDialectResolver implements DialectResolver {
+public class StandardDialectResolver implements DialectResolver {
 
 	@Override
 	public Dialect resolveDialect(DialectResolutionInfo info) {


gsmet:
@vladmihalcea so you affected this one to you. The {{StandardDialectResolver}} is an internal class (present in an internal package) so it shouldn't be extended.

Considering how simple it is, I don't think it's worth thinking about it.

@quaff Except if you have specific issues extending `DialectResolver`?

quaff:
> How about using composition instead?
> 
> ```
> public class MyDialectResolver implements DialectResolver {
> 
>     private final DialectResolver defaultDialectResolver;
> 	
> 	public MyDialectResolver(DialectResolver defaultDialectResolver) {
> 		this.defaultDialectResolver = defaultDialectResolver;
> 	}
> 
> 	@Override
> 	public Dialect resolveDialect(DialectResolutionInfo info) {
> 
> 		DatabaseProduct database = DatabaseProduct.parse(info.getDatabaseName());
> 		int majorVersion = info.getDatabaseMajorVersion();
> 	
> 		if (database == DatabaseProduct.MYSQL) {
> 			if (majorVersion == 8)
> 				return new MySQL8Dialect();
> 		}
> 		return defaultDialectResolver.resolveDialect(info);
> 	}
> }
> ```
> Extending Hibernate classes in the `internal` packages is always risky as the implementation can change from one release to the other.
> 
> Since Hibernate 5.3, you don't even java a JavaDoc for the `internal` classes. Therefore, just like you can't extend `java.lang.String` to add your custom logic, you should not extend framework classes that are not meant to be extended.
> 
> I'm closing this issue. If anyone from the Hibernate team thinks we should make this change, feel free to re-open it.

Set hibernate.dialect_resolvers=MyDialectResolver will failed since no public default constructor.

quaff:
```
private final DialectResolver defaultDialectResolver = new StandardDialectResolver();
```
I will use composition instead of inheritance, thanks.

gsmet:
Well, you can just initialize the defaultDialectResolver at initialization time:

```java
public class MyDialectResolver implements DialectResolver {

       private final DialectResolver defaultDialectResolver = new StandardDialectResolver();

	@Override
	public Dialect resolveDialect(DialectResolutionInfo info) {

		DatabaseProduct database = DatabaseProduct.parse(info.getDatabaseName());
		int majorVersion = info.getDatabaseMajorVersion();
	
		if (database == DatabaseProduct.MYSQL) {
			if (majorVersion == 8)
				return new MySQL8Dialect();
		}
		return defaultDialectResolver.resolveDialect(info);
	}
}
```

Then you have your public constructor.

vladmihalcea:
Most Hibernate object resolver settings can take actual instances, so if you build the DialectResolver programmatically, like with Spring Java-based config, you can pass the default implementation. Otherwise, you can initialize the default instance in the constructor using the Hibernate implementation.

PR #2676, opened 2018-12-04
HHH-13140 - Criteria API multiselect aliases are not passed through to the JPQL query and they are not available in ResultTransformer

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13140

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/alias/CriteriaMultiselectAliasTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/alias/CriteriaMultiselectAliasTest.java
@@ -1,0 +1,112 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/alias/CriteriaMultiselectAliasTest.java
```diff
@@ -0,0 +1,112 @@
...
+	public void test() throws Exception {
+		doInJPA( this::entityManagerFactory, entityManager -> {
+			final CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+
+			final CriteriaQuery<Object[]> query = cb.createQuery( Object[].class );
+
+			final Root<Book> entity = query.from( Book.class );
+			query.multiselect(
+					entity.get( "id" ).alias( "id" ),
+					entity.get( "name" ).alias( "name" )
```
gsmet: Wondering if maybe you should use aliases different from the column names? That would make the test more natural.

vladmihalcea: Makes sense. I'll push a new change. Maybe it's worth using a `StringBuilder` as well for the path building logic.

gsmet:
@schauder yes, that was my first instinct too when I saw your issue.

PR #2677, opened 2018-12-05
 HHH-13141 use upper-case for sequences query in SQLServer2012Dialect

dohertyfjatl:
The upper-case query should work on both case-sensitive and case-insensitive collations.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/dialect/functional/SQLServerDialectSequenceInformationTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/dialect/functional/SQLServerDialectSequenceInformationTest.java
@@ -95,7 +95,7 @@ public class SQLServerDialectSequenceInformationTest extends BaseCoreFunctionalT
 		} );
 	}
 
-	private void doInAutoCommit(String updateQuery ) {
+	private void doInAutoCommit( String updateQuery ) {
 		try ( Connection connection = serviceRegistry().getService( JdbcServices.class )
 				.getBootstrapJdbcConnectionAccess()
 				.obtainConnection();
@@ -106,7 +106,6 @@ public class SQLServerDialectSequenceInformationTest extends BaseCoreFunctionalT
 		catch ( SQLException e ) {
 			log.debug( e.getMessage() );
 		}
-
 	}
 
 }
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/test/dialect/functional/SQLServerDialectSequenceInformationTest.java
```diff
@@ -0,0 +1,111 @@
...
+				log.error( e );
+				fail( "Sequence information was not retrieved: " + e.getMessage() );
+			}
+			finally {
+				StandardServiceRegistryBuilder.destroy( ssr );
+			}
+		} );
+	}
+
+	private void doInAutoCommit( String updateQuery ) {
```
vladmihalcea: We have a `doInAutoCommit``method in `AbstractStoredProcedureTest`. I think we should move that to `TransactionUtil`.

hibernate-core/src/test/java/org/hibernate/test/dialect/functional/SQLServerDialectSequenceInformationTest.java
```diff
@@ -0,0 +1,111 @@
...
+	@TestForIssue(jiraKey = "HHH-13141")
+	public void testExtractSequenceInformationForSqlServerWithCaseSensitiveCollation() {
+		String databaseNameToken = "databaseName=";
+		String url = (String) Environment.getProperties().get( AvailableSettings.URL );
+		String[] tokens = url.split( databaseNameToken );
+		String newUrl = tokens[0] + databaseNameToken + DATABASE_NAME;
+
+		Dialect dialect = serviceRegistry().getService( JdbcServices.class ).getDialect();
+
+		doInHibernate( this::sessionFactory, session -> {
```
vladmihalcea: The call to `doInHibernate` is not needed. You can just remove it and execute its content instead.

gsmet:
@vladmihalcea I let you take care of this issue since you started reviewing it?

My only comment is that we need to squash the whitespace fix commits.

This one is needed for 5.4.0.Final I plan to release next week.

vladmihalcea:
@gsmet Sure thing. I have some changes I think we should do to this issue like verifying we extract the sequence info for a given sequence.

gsmet:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2678 .

PR #2678, opened 2018-12-05
HHH-13141 - Could not fetch the SequenceInformation from the database when using SQL Server with a case-sensitive collation

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13141


vladmihalcea:
@dohertyfjatl @gsmet This is my attempt to rework the [original PR](https://github.com/hibernate/hibernate-orm/pull/2677).

gsmet:
@vladmihalcea looks good to me.

Any chance, you could make the addition of `doInAutoCommit()` to `TransactionUtil` + update of the existing test a separate commit (which would be the first one and then you would have the fix + new test on top of that)?

vladmihalcea:
@gsmet I'll create a new Jira issue for the `doInAutocommit` change and push that change first, then proceed with this one.

dohertyfjatl:
@vladmihalcea @gsmet Many thanks for picking this up and resolving it so quickly!

PR #2681, opened 2018-12-07
 HHH-12298 Add test for join fetch with bytecode enhancement

gsmet:
* https://hibernate.atlassian.net/browse/HHH-12298
 * Also contains https://hibernate.atlassian.net/browse/HHH-13138 to fix some class loading issue that was making the test fail

Note that the real CL issue is not fixed (https://hibernate.atlassian.net/browse/HHH-13136), I only fixed the issue in the test infrastructure.

I doubt we'll be able to fix it in 5.x so let's leave it at that.


gsmet:
Merged!

PR #2682, opened 2018-12-07
HHH-13147 - Fix DelayedPostIdentifierInsert when using second-level cache.

Naros:
https://hibernate.atlassian.net/browse/HHH-13147


gsmet:
Looks good to me. And I agree we should include it in 5.4.0. 

Please push it if you have time or I’ll push it on Monday when I’m back.

PR #2683, opened 2018-12-10
HHH-13094 - Respect @Any.fetch setting to FetchType.EAGER

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13094

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/source/spi/AnyMappingSource.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/source/spi/AnyMappingSource.java
@@ -15,4 +15,8 @@ package org.hibernate.boot.model.source.spi;
 public interface AnyMappingSource {
 	AnyDiscriminatorSource getDiscriminatorSource();
 	AnyKeySource getKeySource();
+
+	default boolean isLazy() {
+		return true;
+	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
@@ -2343,6 +2343,9 @@ public class ModelBinder {
 			Any anyBinding,
 			final AttributeRole attributeRole,
 			AttributePath attributePath) {
+
+		anyBinding.setLazy( anyMapping.isLazy() );
+
 		final TypeResolution keyTypeResolution = resolveType(
 				sourceDocument,
 				anyMapping.getKeySource().getTypeSource()
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/SingularAttributeSourceAnyImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/SingularAttributeSourceAnyImpl.java
@@ -268,4 +268,9 @@ public class SingularAttributeSourceAnyImpl
 	public String getCascadeStyleName() {
 		return jaxbAnyMapping.getCascade();
 	}
+
+	@Override
+	public boolean isLazy() {
+		return isBytecodeLazy();
+	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
@@ -3272,12 +3272,14 @@ public final class AnnotationBinder {
 							+ BinderHelper.getPath( propertyHolder, inferredData )
 			);
 		}
+		boolean lazy = ( anyAnn.fetch() == FetchType.LAZY );
 		Any value = BinderHelper.buildAnyValue(
 				anyAnn.metaDef(),
 				columns,
 				anyAnn.metaColumn(),
 				inferredData,
 				cascadeOnDelete,
+				lazy,
 				nullability,
 				propertyHolder,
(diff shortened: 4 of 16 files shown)


johnlinp:
@vladmihalcea Thank you for your review! However, I have 2 separate commits in #2637, and you seem to mix them as a single commit. Can you split it?

vladmihalcea:
We squash commits when integrating a Pull Request. That's why it's one commit instead of 2.

For XML, we could add a new test case.

johnlinp:
@vladmihalcea I see. Thanks for the explanation.

I added new unit tests in another pull request: #2684. It seems that XML doesn't get fixed. I'll work on it.

johnlinp:
It seems that we should modify `ModelBinder.java` to make XML work.

Here is my proposed code change against `ModelBinder.java`:

```diff
diff --git a/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java b/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
index 2c7ec3ee1f..65b67aa8f1 100644
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
@@ -2351,6 +2351,9 @@ public class ModelBinder {
                        anyBinding.setIdentifierType( keyTypeResolution.typeName );
                }

+               final SingularAttributeSource singularAttributeSource = (SingularAttributeSource) anyMapping;
+               anyBinding.setLazy( singularAttributeSource.isBytecodeLazy() );
+
                final TypeResolution discriminatorTypeResolution = resolveType(
                                sourceDocument,
                                anyMapping.getDiscriminatorSource().getTypeSource()
```

It might not be the best fix, but at least it passes the unit tests in #2684.

hibernate-core/src/main/java/org/hibernate/boot/model/source/spi/AnyMappingSource.java
```diff
@@ -15,4 +15,8 @@
 public interface AnyMappingSource {
 	AnyDiscriminatorSource getDiscriminatorSource();
 	AnyKeySource getKeySource();
+
+	default boolean isLazy() {
+		return true;
```
johnlinp: Do we need a default value here? Even if we do, wouldn't it be `false` since [`Any.fetch` defaults to `FetchType.EAGER`](http://docs.jboss.org/hibernate/orm/5.3/javadocs/org/hibernate/annotations/Any.html#fetch--)?

hibernate-core/src/test/java/org/hibernate/test/annotations/any/AnyEagerHbmTest.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.test.annotations.any;
```
johnlinp: Since it's a unit test for HBM, not for annotation, can you move this unit test out of the annotations package?

hibernate-core/src/test/java/org/hibernate/test/annotations/any/AnyLazyHbmTest.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.test.annotations.any;
```
johnlinp: Same here.

hibernate-core/src/main/java/org/hibernate/boot/model/source/spi/AnyMappingSource.java
vladmihalcea: Yes, we do since this was the previous behavior. If we change this, we will introduce a backward incompatible change.

hibernate-core/src/test/java/org/hibernate/test/annotations/any/AnyEagerHbmTest.java
vladmihalcea: Bu then all common classes will have to be moved as well.

hibernate-core/src/main/java/org/hibernate/boot/model/source/spi/AnyMappingSource.java
johnlinp: I see. But will you add a comment for indicating that it's for backward compatibility? Otherwise it would be confusing since it is different from the documentation.

hibernate-core/src/test/java/org/hibernate/test/annotations/any/AnyEagerHbmTest.java
johnlinp: Hmm... I guess all of the test cases are kind of messy. So be it.

vladmihalcea: I've moved those to `org.hibernate.test.any`.

PR #2684, opened 2018-12-11
HHH-13094 - Add unit tests for lazy any mapping in XML

johnlinp:
According to #2683, unit tests for lazy any mapping in XML is required.


vladmihalcea:
Superseded by [this PR](https://github.com/hibernate/hibernate-orm/pull/2683).

PR #2686, opened 2018-12-12
HHH-13156 - Enhance the @AnyMetaDef annotation section with more details about the optimal placement

johnlinp:
In my project, there is only one place to use @Any. Therefore, it
makes no sense to put its @AnyMetaDef as a package metadata for me.
Before this patch, the documentation seems to recommend to place
@AnyMetaDef as a package metadata unconditionally.


vladmihalcea:
Applied upstream, thanks.

PR #2619, opened 2018-11-01
HHH-13073 - AnyMetaDef.metaType should default to "string"

johnlinp:
https://hibernate.atlassian.net/browse/HHH-13073

PR #2687, opened 2018-12-12
HHH-13073 - AnyMetaDef.metaType should default to "string"

vladmihalcea:
@johnlinp This PR continues [the original one](https://github.com/hibernate/hibernate-orm/pull/2619) and adds a test for HBM as well.


johnlinp:
@vladmihalcea Is there any update on this one? Thanks!

vladmihalcea:
@johnlinp Since this is not a critical issue, I'll push it to 6.0 and it can be backported to 5.4 if the branch maintainers want to do that.

johnlinp:
@vladmihalcea Sure, that works for me perfectly too! Thank you.

vladmihalcea:
I tried to integrate it into 6.0, but the tests fail because the annotation parser throws an exception. I'll try it when that's fixed.

sebersole:
Not knowing the nature of the failure its hard to say whether this should work now.  But adding the `6.0?` label as it targets master which is still 5.4

gavinking:
There is no longer any `@AnyMetaDef` annotation. Closing.

PR #2689, opened 2018-12-12
Fix NullPointerException for queries without parameters

DavideD:
I've added a couple of tests with named queries and adapted one of the old existing tests.


DavideD:
Merged https://github.com/hibernate/hibernate-orm/pull/2689

Thanks

PR #2691, opened 2018-12-13
HHH-12555 Fix merging of lazy loaded blobs/clobs/nclobs #2390

gbadner:
https://hibernate.atlassian.net/browse/HHH-12555

A different approach to fixing HHH-12555 than https://github.com/hibernate/hibernate-orm/pull/2390.

Tests should be added for the non-default `LobMergeStrategy` implementations (although it appears that they are completely untested at this point).

FYI, I'm not sure this covers all cases in https://github.com/hibernate/hibernate-orm/pull/2390. I'd like to get some feedback before spending more time on this approach.


hibernate-core/src/main/java/org/hibernate/type/AbstractStandardBasicType.java
```diff
@@ -364,10 +360,20 @@ public Object replace(
 			Map copyCache,
 			ForeignKeyDirection foreignKeyDirection) {
 		return ForeignKeyDirection.FROM_PARENT == foreignKeyDirection
-				? getReplacement( (T) original, (T) target, session )
+				? doReplacement( original, target, session )
 				: target;
 	}
 
+	protected Object doReplacement(Object original, Object target, SharedSessionContractImplementor session) {
+		if ( original == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
```
yrodiere: Is it absolutely necessary to do these checks in `AbstractStandardBasicType`? I think what Guillaume was trying to do in his PR (on my suggestion) was to do these checks in callers of `replace`. The main reason was that:

* `BasicType` is API if you refer only to the package it's in, so users could theoretically implement it
* `LazyPropertyInitializer.UNFETCHED_PROPERTY` is SPI
* Why would an an API  user have to care about a constant lost in the middle of SPIs?

Even for SPI implementors, it's asking a lot to expect them to be aware of such "magic values" without documenting it explicitly on the method's javadoc.

Maybe the right thing to do would be to move `LazyPropertyInitializer.UNFETCHED_PROPERTY` to API and document on the javadoc of `replace` the possibility of being passed a "magic value".

hibernate-core/src/main/java/org/hibernate/dialect/LobMergeStrategy.java
```diff
@@ -18,14 +18,25 @@
...
+
+	/**
+	 * Does this LobMergeStrategy support the merging a Lobs?
+	 *
+	 * @return true, if merging Lobs is supported; false, otherwise.
+	 */
+	default boolean supportsMerge() {
+		return true;
+	}
+
```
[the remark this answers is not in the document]
yrodiere: To be honest I don't really understand why it's okay to consider an unfetched property as null in some cases but not in others.

From what I understand what you are trying to do is prevent the `replace` methods from returning the original value (or a deep copy of it) when the target is unfetched and the lob merge strategy is the legacy one. You want it to return `LazyPropertyInitializer.UNFETCHED_PROPERTY` instead.

My question is then: why is it okay to return the original in other cases? Does the merge happen after this call to `replace`? If so, what is the return of this method considered as: the target or the original? What is used for the other?

Another question: I suppose the behavior of calling the merge strategy on these results is built in another class. Can this class not check the fetched/unfetched state of the target and decide to call `replace` or not based on that? I really think we shouldn't be dealing with "unfetched" semantics in the type itself...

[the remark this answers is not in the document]
gbadner: @yrodiere, @gsmet:

`original` is a property in the `entity` being merged, (e.g, `session.merge( entity )`). `entity` may be detached or managed (i.e. already in the `PersistenceContext`).

`target` is property value in a managed entity, `managedEntity` (e.g., `managedEntity = session.merge( entity )`). The property is either a value loaded from the database, or, if lazy, it is `LazyPropertyInitializer.UNFETCHED_PROPERTY`.

Here are the possible combinations:

`original == LazyPropertyInitializer.UNFETCHED_PROPERTY`:

Resulting property in `managedEntity`: `target`
(Since `original` was uninitialized, we know that it has not been changed)

`original != LazyPropertyInitializer.UNFETCHED_PROPERTY`:

Resulting property in `managedEntity`: 

- If the property is a Lob and the dialect's `LobMergeStrategy` is `LEGACY_LOB_MERGE_STRATEGY`, then `target` should be returned (because that strategy does not support merge; iow, changes are supposed to be ignored). Replacing `target` with `null` will ultimately change the persisted value to `null`, which is not how `LEGACY_LOB_MERGE_STRATEGY` is supposed to work.

- Otherwise, a deep copy of `original` should be returned.

hibernate-core/src/main/java/org/hibernate/type/AbstractStandardBasicType.java
gbadner: It is possible to do the check in `TypeHelper`, but it would be a bit messy, because `target` should not be replaced by `null` for Lobs that use `LEGACY_LOB_MERGE_STRATEGY`.

@yrodiere, that's a good point about users potentially implementing `doReplacement`. I'll make a quick change so it's not possible.

yrodiere: That's a good idea, but my point was more that users could even implement the interface themselves (from the javadoc, it seems we don't particularly recommend using this abstract class as a base). If they do, they won't have a clue that this magic value may be passed to `replace`, and even if they figure that out, accessing the constant would mean relying on SPIs...
So I think at the very least the constant should be moved to API and mentioned in the javadoc of `replace`; ideally the type should not have to deal with this at all, but from what you are saying this would be complex. I suppose we'd have to also expose something like `supportsMerge` on `Type`?

gbadner: @yrodiere, ah, I see what you mean now; I didn't realize `LazyPropertyInitializer.UNFETCHED_PROPERTY` was an SPI.

gbadner: @yrodiere, I was trying to avoid exposing something like, `Type#supportsMerge`, especially since `LEGACY_LOB_MERGE_STRATEGY` will likely get deprecated. Honestly, I don't know if anyone even uses it. As mentioned above, it is not tested.

gbadner: I also see that in 4.2, references to `LazyPropertyInitializer.UNFETCHED_PROPERTY` (in a different package) are not present in any `Type` implementations. Instead, it's used in `TypeHelper`.

gsmet: Maybe let's not take design decisions based on something legacy?

I would be fine with having the code in `TypeHelper` and `LEGACY_LOB_MERGE_STRATEGY` not working better than it was.

gbadner: I'll create a new PR based on @gsmet's  https://github.com/hibernate/hibernate-orm/pull/2390 (which moves references to `LazyPropertyInitializer.UNFETCHED_PROPERTY` to `TypeHelper`), with changes to treat `LEGACY_LOB_MERGE_STRATEGY` properly.

gbadner: @gsmet, missed your message before I posted mine.

I'll check to see if it was working in 4.2.

gsmet: BTW, `UNFETCHED_PROPERTY` is already partially handled in `TypeHelper`. That's one of the reasons why I would like the code to be there: it's better to centralize this handling rather than spreading it everywhere.

gbadner: @gsmet, yes I agree now. 

BTW, this is kind of complicated, but I realized that when merging a transient entity with an initialized Lob that is mapped as lazy with a Lob cannot be persisted using `LEGACY_LOB_MERGE_STRATEGY` cannot possibly work. When `target` gets instantiated, the field with be uninitialized. Even if we somehow fixed `LEGACY_LOB_MERGE_STRATEGY` to return `target`, then I think Hibernate would try to persist an entity with an uninitialized value. I'm not sure what would happen in this case. I guess Hibernate would throw an exception or the property would be persisted as `null`.

I'll go back to your PR and add a comment about this.

PR #2692, opened 2018-12-13
Fix and enable PreInsertEventListenerVetoBidirectionalTest

DavideD:


DavideD:
Merged https://github.com/hibernate/hibernate-orm/pull/2692

Thanks @dreab8

PR #2694, opened 2018-12-13
HHH-13160 - Polymorphic query for InheritanceType.TABLE_PER_CLASS is not using UNION ALL

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13160


documentation/src/main/asciidoc/userguide/chapters/domain/inheritance.adoc
```diff
@@ -310,6 +310,21 @@ include::{extrasdir}/entity-inheritance-table-per-class-query-example.sql[]
 [IMPORTANT]
 ====
 Polymorphic queries require multiple UNION queries, so be aware of the performance implications of a large class hierarchy.
```
johnlinp: I think the `UNION ` here should be `UNION ALL` or the following description would be kind of weird.

vladmihalcea: There's a following description that explains the difference. This paragraph uses the UNION operator not the SQL keyword.

PR #2696, opened 2018-12-14
HHH-12878 - Include statement when throwing StaleStateException in Expectations.BasicExpectation

danielsmith1995:
Change to allow the statement to be included in the error message when a StaleStateException is thrown from Expectations.BasicExpectation. This would make the exception much more helpful when reading the log.

https://hibernate.atlassian.net/browse/HHH-12878


gsmet:
Adjusted the commit message and merged, thanks!

PR #2697, opened 2018-12-14
Fix usage of StatementInspector

dreab8:


dreab8:
@sebersole thanks for the review, I'm going to push it upstream

dreab8:
applied upstream

PR #2698, opened 2018-12-14
Dialect not supporting Row value constructor syntax

dreab8:


hibernate-core/src/main/java/org/hibernate/sql/ast/consume/spi/AbstractSqlAstWalker.java
```diff
@@ -537,6 +540,26 @@ public void visitCurrentTimestampFunction(CurrentTimestampFunction function) {
 		appendSql( "current_timestamp" );
 	}
 
+	private void visitTuple(SqlTuple leftHandSide, SqlTuple rightHandSide, RelationalPredicate.Operator operator) {
```
sebersole: Overall, I think this is exactly the way to handle these cases.  This only covers on case though - relational comparison predicates.  Tuples can be used in other types of predicates though e.g. `IN`

And then a nit-pick... this is not really visiting tuples... its still visiting the `RelationalPredicate`.  I think we should call this something like `#renderTupleComparison` or `#visitTupleComparison`

PR #2699, opened 2018-12-17
HHH-12555 : Fix merging of lazy loaded blobs/clobs/nclobs

gbadner:
Supersedes https://github.com/hibernate/hibernate-orm/pull/2390.

I removed code in LobMergeStrategy implementations that copied original Lob when target is null. Instead `#getReplacement` delegates to the `LobMergeStrategy`.

I also added some code to the tests to check that the Lobs were persisted properly.

I found a bug with enhancement involving a property that starts with a lowercase letter followed by an uppercase letter, like `FileNClob#nclob`. I've renamed the property to `clob`. I'll create a jira for the enhancement bug on Monday.


gsmet:
Looks good to me. Squashed my first 2 commits and merged.

PR #2700, opened 2018-12-17
HHH-12780 - Allow native SQL queries that take a given resultClass to map the result set to the required type

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-12780


vladmihalcea:
This PR requires some changes to be implemented in the 6.0 branch for native queries prior to being integrated.

PR #2701, opened 2018-12-17
HHH-13084 HHH-13114 (5.3)

gbadner:

This is for 5.3 branch.


gbadner:
Squashed and merged.

PR #2703, opened 2018-12-18
HHH-13164 - Revert all changes recently made for DelayedPostInsertIdentifier

Naros:
https://hibernate.atlassian.net/browse/HHH-13164

This PR adds a test case for the user problem reported in this JIRA and reverts all changes made that are related to the following JIRAs:

https://hibernate.atlassian.net/browse/HHH-13147
https://hibernate.atlassian.net/browse/HHH-13053
https://hibernate.atlassian.net/browse/HHH-13044
https://hibernate.atlassian.net/browse/HHH-13042
https://hibernate.atlassian.net/browse/HHH-11019

This restores the behavior for `DelayedPostInsertIdentifier` back to its state prior to HHH-11019.


gsmet:
@Naros did you keep the test cases we added for the bugs you fixed? I think we should keep them.

gsmet:
Ah, I see that you did, cool.

gsmet:
@scottmarlow this PR reverts the original patch for https://hibernate.atlassian.net/browse/HHH-11019 as it triggered too many regressions. We fixed a bunch of them but we see more of them every day so we think it's better to go back to what it was for now.

We don't have any idea of the consequences on WF so as you were involved in the original discussion, please let us know if you have any concerns about it.

scottmarlow:
@gsmet regarding HHH-11019 [https://hibernate.atlassian.net/browse/HHH-11019?focusedCommentId=100853&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-100853] there was discussion about how the JPA container should do the right thing, but what does that mean?

Also, I'm not sure why PersistenceContextType is referenced in ORM, looks like that should be removed from hibernate-core/src/main/java/org/hibernate/internal/SessionFactoryImpl.java :)

gsmet:
@scottmarlow so that means you have nothing against this PR?

As for:
> I'm not sure why PersistenceContextType is referenced in ORM, looks like that should be removed from hibernate-core/src/main/java/org/hibernate/internal/SessionFactoryImpl.java :)

could you create a JIRA? It's not used in the current code but wondering if it's supposed to be used? @sebersole ?

sebersole:
@scottmarlow this is the very discussion we had when I first did this work that has just been reverted.  You said, and I quote, that WF/Jipijapa handles `PersistenceContextType#EXTENDED` itself and that it was irrelevant what Hibernate itself did.  So yes, I agree that any/all references to `PersistenceContextType` ought to be removed from Hibernate

Naros:
@scottmarlow we're just wanting to understand if there are any potential impacts to WF if we rollback the behavior implemented for HHH-11019.  As @gsmet mentioned, we've seen what some of us believe are too many regressions and I would rather see 5.3 / 5.4 stable across to board.  I need to have time to prepare the necessary ports to this PR for both 5.3 and 6.0 so any input on whether we can proceed with this or not is much appreciated sooner rather than later.

hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
```diff
@@ -320,42 +316,6 @@ protected Serializable performSaveOrReplicate(
...
-			);
-			return false;
-		}
-		else {
-			return false;
-		}
-	}
-
 	private static boolean isPartOfTransaction(EventSource source) {
 		return source.isTransactionInProgress() && source.getTransactionCoordinator().isJoined();
```
scottmarlow: I looked in 5.1 and don't see the "isJoined" check, in the reverted commit, we are now checking if the session is joined to the active JTA transaction, which seems like a good idea to me.  I think that we are reverting to code changes that may not be as well tested on WildFly, as the current ORM 5.3.x branch (with the reported bugs) or 5.1.x.  My call to action here is that we definitely should run the WF tests and the Eclipse Jakarta EE CTS tests with the 5.3.x branch (either via a snapshot release or an actual release).

Naros: @scottmarlow Just so I am clear, are you suggesting we at a minimum merge a PR into 5.3 that reverts this logic just to test?  If so, that'll be up to @gbadner.  Just bear in mind we're trying to manage keeping 3 concurrent branches synchronized with this revert and I would prefer to keep the process as clean and clear as possible.

Naros: > My only question is whether the code that we are reverting back to, will need further changes, or if we already tested that with WildFly in the past.

@scottmarlow I think in order to get this back to be identical to what has been tested with WF in the past, there is one additional revert I overlooked that you pointed out - which was the `#isJoined()` check.  

I overlooked that `#isJoined()`  was added as a part of HHH-11019 and therefore by removing it as well with the other things this PR includes, would return ORM 5.3 and 5.4 as identical to 5.1. I am fine keeping it as-is because I don't believe it will introduce any negative impact on behavior.

I'll wait to hear Gail's thoughts and go from there.  We're trying to get 5.4.1 out probably sometime next week, so just trying to plan accordingly here.

scottmarlow: @Naros I have no idea if the #isJoined() check should be removed, as there can be an active JTA transaction but the source (Hibernate Session?) may not be joined to it.  I agree that 5.1.x doesn't seem to contain that check.  I also don't know if requiresImmediateIdAccess already reflects whether the source is joined to the active JTA transaction.  @gbadner any thoughts on this?

hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
```diff
@@ -248,7 +244,7 @@ protected Serializable performSaveOrReplicate(
 
 		Serializable id = key == null ? null : key.getIdentifier();
 
-		boolean shouldDelayIdentityInserts = shouldDelayIdentityInserts( requiresImmediateIdAccess, source, persister );
+		boolean shouldDelayIdentityInserts = !isPartOfTransaction( source ) && !requiresImmediateIdAccess;
```
gbadner: Hibernate may not do the right thing when:  `!isPartOfTransaction( source ) && requiresImmediateIdAccess`.

IIUC,  `shouldDelayIdentityInserts` will be set to `false`, and Hibernate will perform the insert (using auto-commit?), even if `hibernate.allow_update_outside_transaction=false` (the default). Shouldn't `TransactionRequiredException` be thrown?

I'll address `#isPartOfTransaction` below.

Background (as I understand it):

IIUC, Hibernate requires that the entity have an ID when `Session#save` (which returns the ID) or `Session#saveOrUpdate` is used to persist a transient entity, so `requiresImmediateIdAccess == true` in those cases.

`Session#persist` does not require that the entity have an ID until after flush so `requiresImmediateIdAccess == false` in that case.

hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
gbadner: I believe checking `source.getTransactionCoordinator().isJoined()` is redundant.

`AbstractSharedSessionContract#isTransactionInProgress` calls `transactionCoordinator.isTransactionActive()`, which only returns`true` if `isJoined() == true`.

hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
gbadner: I suppose this could be handled in a separate jira issue.

gbadner:
I've created https://hibernate.atlassian.net/browse/HHH-13205 to remove `PersistenceContextType` from `SessionFactoryImpl`.

I've created https://hibernate.atlassian.net/browse/HHH-13207 to throw TransactionRequiredException when early insert is required and hibernate.allow_update_outside_transaction=false.

gbadner:
@Naros, what is the extra revert that you mentioned that would remove the redundant {{#isJoined}} check?

I'd like to see that reverted issue before approving this.

hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
Naros: @gbadner I don't believe we have those on CI (yet) afaik.

Naros:
@gbadner The extra revert was the removable of the configuration parameter I added but forgot to remove.  The latest commit I added to this PR shows its removal.

hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
gbadner: I still see the redundant `source.getTransactionCoordinator().isJoined()` check...

hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
```diff
@@ -811,8 +811,4 @@ Object createProxy(Serializable id, SharedSessionContractImplementor session)
 	int[] resolveAttributeIndexes(String[] attributeNames);
 
 	boolean canUseReferenceCacheEntries();
-
-	default boolean canIdentityInsertBeDelayed() {
-		return false;
-	}
```
gbadner: Removing this will break applications that implement this method. You should probably add it back in and deprecate it and say it's no longer used.

gbadner:
`SessionFactoryOptions#isPostInsertIdentifierDelayableEnabled` should also be deprecated.

hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
Naros: @gbadner Fixed in the last resync of the PR.

hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
Naros: Added back as a deprecated method in the last PR resync.

Naros:
The `disable_delayed_identifier_post_inserts` option was never documented, so nothing to worry with removing from there.  I also marked the method `SessionFactoryOptions#isPostInsertIdentifierDelayableEnabled` deprecated as requested.

PR #2704, opened 2018-12-19
HHH-13167 - When omitting the OTHERWISE clause in a CASE expression built with Criteria API, Hibernate throws a NullPointerException

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13167


gsmet:
Looks good, I fixed a few formatting issues and merged.

PR #2705, opened 2018-12-19
HHH-13104 - Oracle 12c / SAP Hana insert fails when entity contains only an identity-based column.

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13104


vladmihalcea:
@Naros @breglerj Let me know what you think of this PR. Thanks.

breglerj:
@vladmihalcea Looks good from my side.

gsmet:
Ah, bummer, Chris had worked on it but I forgot to close the issue when I merged. Sorry about that.

In any case, your patch looks like an improvement. I'll let @Naros decide.

gsmet:
OK, so to be clear, this patch is all about fixing the test failures we have with HANA on CI.

Naros:
Rebased and applied to upstream master.

PR #2706, opened 2018-12-19
HHH-13172 Only throw exception if incorrect inheritance type is being…

ptahchiev:
… used.

As Vlad Mihalcea mentioned in a comment here: https://hibernate.atlassian.net/projects/HHH/issues/HHH-12609
using @Inheritance and @AttributeOverride makes sense only if TABLE_PER_CLASS inheritance strategy is being used.

Issue is now: https://hibernate.atlassian.net/browse/HHH-13172

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
@@ -565,7 +565,7 @@ public final class AnnotationBinder {
 				inheritanceState
 		);
 
-		if(superEntity != null && (
+		if(inheritanceState.getType() != InheritanceType.TABLE_PER_CLASS && superEntity != null && (
 				clazzToProcess.getAnnotation( AttributeOverride.class ) != null ||
 				clazzToProcess.getAnnotation( AttributeOverrides.class ) != null ) ) {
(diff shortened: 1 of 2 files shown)


gsmet:
@ptahchiev looks like our CI is failing (test `org.hibernate.test.annotations.override.inheritance.EntityInheritanceAttributeOverrideTest`fails). Don't know if it's a code or a test problem but could you take a look?

BTW, as mentioned by Gail in the JIRA issue, it would be nice to add a new test to check the new behavior with `TABLE_PER_CLASS`.

We also need a new JIRA for this issue as the previous one is closed and part of a release.

Thanks!

ptahchiev:
@gsmet I changed the already committed test and i added my own test - now the build is green.

gsmet:
@ptahchiev so my understanding is that using `@AttributeOverride` with any type of entity inheritance is just ignored, am I correct?

If so I would lean towards changing the exception to be a warning as 5.4 was designed to be as compatible as possible with previous versions.

I created https://github.com/hibernate/hibernate-orm/pull/2710/files . Can you comment there about if it fits your current needs.

Thanks!

gsmet:
Closing, superseded by https://github.com/hibernate/hibernate-orm/pull/2710 .

PR #2708, opened 2018-12-20
different approach to writing tests

sebersole:
I started looking at using an annotation-driven approach to define how the SessionFactory-under-test should be created.  Just wanted to see if y'all had any thoughts.  As an example, look at `org.hibernate.orm.test.query.criteria.BasicCriteriaBuildingTests` (partially copied here)

````
@TestDomain(
		standardModels = StandardDomainModel.GAMBIT
)
@ServiceRegistry(
		settings = @Setting(
				name = AvailableSettings.HBM2DDL_AUTO,
				value = "create-drop"
		)
)
@SessionFactory
@SuppressWarnings("WeakerAccess")
public class BasicCriteriaBuildingTests implements SessionFactoryScopeInjectable {
	private SessionFactoryScope scope;

	@Test
	public void testParameterCollecting() {
		final SessionFactoryImplementor sf = scope.getSessionFactory();
		...
	}

	...
}
````

There are a few reasons I started down this PoC.  Ultimately I think it allows us to better manage test resources (pools, registries, schemas, etc)

### CODE REVIEW

**Comment by vladmihalcea** [comment]
@sebersole It's an interesting approach. I suppose we only need to declare the `@ServiceRegistry` if we wont to override the default settings.

For the `@TestDomain`, we could come up with some simple models that cover multiple test cases.

Personally, I didn't mind the previos way of testing which required extending a base class either.

**Comment by sebersole** [comment]
> @sebersole It's an interesting approach. I suppose we only need to declare the `@ServiceRegistry` if we wont to override the default settings.

Right.  

 
> For the `@TestDomain`, we could come up with some simple models that cover multiple test cases.

You mean like the ones in https://github.com/hibernate/hibernate-orm/tree/a44051c2c20e739bbe5200f0a6fc7731a52139ce/hibernate-testing/src/main/java/org/hibernate/testing/orm/domain ?  ;)

 
> Personally, I didn't mind the previos way of testing which required extending a base class either.

This approach just makes it easier to do some things from these JUnit extensions rather than requiring that the test class itself manage it.  E.g., these extensions could spin up docker containers, create/manage connection pools for these standard models, etc.

**Comment by sebersole** [comment]
@vladmihalcea Relatedly... I notice you've been adding tests outside of the package structure we started on the 6 branch.  Please use that structure, thanks

**Comment by sebersole** [comment]
By the way, other cool benefits of this approach include granularity and re-usability.  

Just want to test ServicRegistry as the fixture?  Easy, just use `@ServicRegistry` + `ServicRegistryAware`..

Just want to test Metamodel as the fixture?  Well again I can reuse the same `@ServicRegistry` and it works as before.  

Same for SessionFactory...

Making this piece work in the normal extend-base-test-class approach is much more difficult and fugly.

**Comment by sebersole** [comment]
FWIW I also just defined an even better approach (imo):

````
@TestDomain(
		standardModels = StandardDomainModel.GAMBIT
)
@ServiceRegistry(
		settings = @Setting(
				name = AvailableSettings.HBM2DDL_AUTO,
				value = "create-drop"
		)
)
@SessionFactory
@SuppressWarnings("WeakerAccess")
public class BasicCriteriaBuildingTests {
	@Test
	public void testSingleInjection(SessionFactoryScope scope) {
		assertThat( scope, notNullValue() );
	}

	@Test
	public void testMultipleInjections(SessionFactoryScope scope, MetadataImplementor model) {
		assertThat( scope, CoreMatchers.notNullValue() );
		assertThat( model, CoreMatchers.notNullValue() );
	}
}
````

**Comment by sebersole** [comment]
See it in action - `org.hibernate.testing.orm.infrastructure.ParameterInjectionBasedTestingTests` in `hibernate-testing`

**Comment by dreab8** [comment]
@sebersole it looks very cool to me.

what about adding a
```
public abstract class AbstractDomainModelDescriptor implements DomainModelDescriptor {

	protected abstract Class[] getDomainClasses();

	@Override
	public void applyDomainModel(MetadataSources sources) {
		for ( Class domainClass : getDomainClasses() ) {
			sources.addAnnotatedClass( domainClass );
		}
	}
}
```
so that we can have
```
public class GambitDomainModel extends AbstractDomainModelDescriptor {
	public static final GambitDomainModel INSTANCE = new GambitDomainModel();

	protected Class[] getDomainClasses() {
		return new Class[] {
				BasicEntity.class,
				Component.class,
				EmbeddedIdEntity.class,
				EntityOfBasics.class,
				EntityOfComposites.class,
				EntityOfDynamicComponent.class,
				EntityOfLists.class,
				EntityOfMaps.class,
				EntityOfSets.class,
				EntityWithLazyManyToOneSelfReference.class,
				EntityWithLazyOneToOne.class,
				EntityWithManyToOneJoinTable.class,
				EntityWithManyToOneSelfReference.class,
				EntityWithNonIdAttributeNamedId.class,
				EntityWithOneToMany.class,
				EntityWithOneToOne.class,
				EntityWithOneToOneJoinTable.class,
				EntityWithOneToOneSharingPrimaryKey.class,
				Shirt.class,
				SimpleEntity.class
		};
	}
}
```

sebersole:
@dreab8 I think that is a good idea.  I added that, but in a slightly different form:

````

public abstract class AbstractDomainModelDescriptor implements DomainModelDescriptor {
	private final Class[] annotatedClasses;

	protected AbstractDomainModelDescriptor(Class... annotatedClasses) {
		this.annotatedClasses = annotatedClasses;
	}

	@Override
	public void applyDomainModel(MetadataSources sources) {
		for ( Class annotatedClass : annotatedClasses ) {
			sources.addAnnotatedClass( annotatedClass );			
		}
	}
}


public class GambitDomainModel extends AbstractDomainModelDescriptor {
	public static final GambitDomainModel INSTANCE = new GambitDomainModel();

	public GambitDomainModel() {
		super(
				BasicEntity.class,
				Component.class,
				...
		);
	}
}

````

dreab8:
@sebersole +1

sebersole:
Another cool one (I think) is:

````
	@ParameterizedTest
	@ValueSource(strings = {"true", "false"})
	public void testInjectionPlusParameterized(boolean bool, SessionFactoryScope scope) {
		assertThat( scope, notNullValue() );
	}
````

This test gets run twice... once with `true` and once with `false`.

Once JUnit adds "parameterized test templates" we could even leverage that from annotations as we build the SF (thinking envers testing)

PR #2709, opened 2018-12-22
HHH-13175 - Add method to return null instead of throwing exception

BOOTMGR:
IllegalArgumentException is thrown during two phase load's plan creation for member relation (i.e. ManyToMany).

While resolving relation, MetamodelGraphWalker visits each attribute of target entity type and checks whether fetchgraph or loadgraph has directed to eagerly fetch that attribute. If so, it modifies load plan.

Recent changes in 5.4.0 uses contracts specified by JPA 2.0 to walk down and manage types. getAttribute(name) contract specified by JPA's ManagedType throws IllegalArgumentException when asked attribute named as 'name' is not found.

But as per Hibernate's implementation, this contract is being used to check and modify load plan if attribute matches with specified EntityGraph.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/ManagedDomainType.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/ManagedDomainType.java
@@ -6,10 +6,18 @@
  */
 package org.hibernate.metamodel.model.domain;
 
+import javax.persistence.metamodel.Attribute;
 import javax.persistence.metamodel.ManagedType;
 
 /**
  * @author Steve Ebersole
  */
 public interface ManagedDomainType<J> extends SimpleDomainType<J>, ManagedType<J> {
+	/**
+     *  Return the attribute of the managed
+     *  type that corresponds to the specified name if exists.
+     *  @param name  the name of the represented attribute
+     *  @return attribute if exists else null
+     */
+	Attribute<? super J, ?> getAttributeIfExists(String name); 
 }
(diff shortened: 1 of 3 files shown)


gsmet:
Hi @BOOTMGR ,

Could you add a test case so that we understand exactly what's your issue and make sure we keep it working in the future?

Thanks!

BOOTMGR:
@gsmet,

Test cases are attached separately in JIRA (https://hibernate.atlassian.net/browse/HHH-13175). I will be sending separate pull request to add test case in source.

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/ManagedDomainType.java
```diff
@@ -6,10 +6,18 @@
...
  * @author Steve Ebersole
  */
 public interface ManagedDomainType<J> extends SimpleDomainType<J>, ManagedType<J> {
+	/**
+     *  Return the attribute of the managed
+     *  type that corresponds to the specified name if exists.
+     *  @param name  the name of the represented attribute
+     *  @return attribute if exists else null
+     */
+	Attribute<? super J, ?> getAttributeIfExists(String name);
```
JWGmeligMeyling: Not sure if a metamodel extension is the way to go here. Especially not adding an alternative to a method already present with just slightly different handling of absent values (not to mention that that method comes from the JPA spec and not just our metamodel).

There must be an other way to prevent absent attributes being looked up...

BOOTMGR: Agreed. Hibernate should not apply entity graph on non root entities in first place.

sebersole: The method itself is a good idea.  The name is awful ;)

Also, there are a lot of extra unnecessary changes here.  I am about to push an alternative PR

gsmet:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2713 . Closing.

PR #2710, opened 2018-12-25
HHH-13172 Log a warning instead of throwing an exception when @AttributeOverride is used in conjunction with entity inheritance

gsmet:
This is a follow-up of https://github.com/hibernate/hibernate-orm/pull/2706 .

I'm not sure exactly about what is supported or not so the condition might need some tweaking. Gail, do you know more about that?

My understanding is that it's atm only supported for `@MappedSuperclass`.

In any case, let's not throw an exception as it seems to break some applications.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
@@ -568,10 +568,7 @@ public final class AnnotationBinder {
 		if(superEntity != null && (
 				clazzToProcess.getAnnotation( AttributeOverride.class ) != null ||
 				clazzToProcess.getAnnotation( AttributeOverrides.class ) != null ) ) {
-			throw new AnnotationException(
-					"An entity annotated with @Inheritance cannot use @AttributeOverride or @AttributeOverrides: " +
-							clazzToProcess.getName()
-			);
+			LOG.unsupportedAttributeOverrideWithEntityInheritance( clazzToProcess.getName() );
 		}
 
 		PersistentClass persistentClass = makePersistentClass( inheritanceState, superEntity, context );
@@ -3043,7 +3040,7 @@ public final class AnnotationBinder {
 				column.setUpdatable( false );
 			}
 		}
-		
+
 		final JoinColumn joinColumn = property.getAnnotation( JoinColumn.class );
 		final JoinColumns joinColumns = property.getAnnotation( JoinColumns.class );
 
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
```diff
@@ -568,10 +568,7 @@ public static void bindClass(
 		if(superEntity != null && (
 				clazzToProcess.getAnnotation( AttributeOverride.class ) != null ||
 				clazzToProcess.getAnnotation( AttributeOverrides.class ) != null ) ) {
-			throw new AnnotationException(
-					"An entity annotated with @Inheritance cannot use @AttributeOverride or @AttributeOverrides: " +
-							clazzToProcess.getName()
-			);
+			LOG.unsupportedAttributeOverrideWithEntityInheritance( clazzToProcess.getName() );
```
ptahchiev: The if-above must have the extra condition `inheritanceState.getType() != InheritanceType.TABLE_PER_CLASS` (just like my pull-request).

gsmet: Hmmm, did I misunderstand? I thought you said that even for `TABLE_PER_CLASS`, ORM was ignoring the `@AttributeOverride` annotations?

ptahchiev: At the moment it throws exception no matter what type of Inheritance strategy is used. In 5.2.x it was silently ignoring the attribute overrides. As I discussed it with Vlad here https://hibernate.atlassian.net/browse/HHH-12609 he said that Inheritance and AttributeOverride makes sense only when the InheritanceStrategy is TABLE_PER_CLASS. That is why my pull-request added an extra check and if the InheritanceStrategy is TABLE_PER_CLASS not to throw the exception that is now thrown. If you change the exception to warning then i believe you must also add the check for the inheritance strategy.

gsmet: @ptahchiev well, it would make sense with `TABLE_PER_CLASS` but it does not currently work. AFAICS, `@AttributeOverride` is only supported with `MappedSuperclass`. So for now, it's not supported. Do you agree with that?

If so, I maintain the warn message should state the current situation, not the situation we would like it to be.

As for supporting `TABLE_PER_CLASS`, I consider it's a new feature so it should be done in the context of ORM 6.

ptahchiev: OK

PR #2711, opened 2018-12-26
HHH-13068: Fix for FK violation if "order_inserts = true" 

harikantverma:
Following are the changes proposed against [HHH-13068](https://hibernate.atlassian.net/browse/HHH-13068) :
1. Added set of children inside BatchIdentifier to detect cycle in the entity relationship.
2. Added methods to check if cycle exists in the generated dependency graph.
3. Since the time complexity of current sorting mechanism is n*n and is not certain about the order at the end which let the FK violation to be occurred even there are chances of cycle, hence changed the sorting mechanism to sorting the batch by **topological sort** if there is no cycle in dependency graph.
4. Test cases are attached in the ticket [HHH-13068](https://hibernate.atlassian.net/browse/HHH-13068).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
@@ -1044,7 +1044,10 @@ public class ActionQueue {
 			private Set<String> childEntityNames = new HashSet<>( );
 
 			private BatchIdentifier parent;
-
+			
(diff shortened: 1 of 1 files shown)


vladmihalcea:
@harikantverma I created [this PR](https://github.com/hibernate/hibernate-orm/pull/2719) to substitute this one.

The new PR brings the following enhancements:

- the tests for the Jira issues are included
- the code is refactored to match the HIbernate coding guidelines

I still need to thoroughly verify it in order to merge it. Thanks for your contribution.

PR #2713, opened 2018-12-31
HHH-13175 - Eager subsequent-select fails when EntityGraph is specifi…

sebersole:
Jira: https://hibernate.atlassian.net/browse/HHH-13175

Alternative to https://github.com/hibernate/hibernate-orm/pull/2709

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/graph/GraphParsingTestEntity.java
+++ b/hibernate-core/src/test/java/org/hibernate/graph/GraphParsingTestEntity.java
@@ -42,7 +42,7 @@ public class GraphParsingTestEntity {
 		this.name = name;
 	}
 
-	@ManyToOne
+	@ManyToOne( fetch = FetchType.LAZY )
 	public GraphParsingTestEntity getLinkToOne() {
 		return linkToOne;
 	}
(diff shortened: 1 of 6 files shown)


hibernate-core/src/test/java/org/hibernate/graph/GraphParsingTestEntity.java
```diff
@@ -42,7 +42,7 @@ public void setName(String name) {
 		this.name = name;
 	}
 
-	@ManyToOne
+	@ManyToOne( fetch = FetchType.LAZY )
```
gsmet: This test change and the one just below look suspicious to me. Isn't the issue triggered with eager loading?

sebersole: Nope.  In fact LAZY would work perfectly fine because loading the relationships is triggered back in the test code when accessed - aka, "outside the scope" of the `#find` where the graph is applied.

It is a valid point though in that I probably should have create a new set of entities for this test and left these as they were.  I'll work on that

sebersole: Sorry, mis-read that comment.  LAZY worked all along.  With these changes it works using either.

gsmet: Yeah, so let's not rely on LAZY for the tests :).

Thanks for taking care of that.

sebersole: Hence why I am making the change

sebersole: And to be clear for posterity.... This PR is basically 2 separate changes, either of which fixes the issue on its own.  

One change is essentially the change proposed in https://github.com/hibernate/hibernate-orm/pull/2709, though using different naming (`#findAttribute` rather than `#getAttributeIfExists`).

The other change is in `SessionImpl#internalLoad` (which is triggered for eager subsequent selects) to clear effective entity graph.  

IMO, the change to `SessionImpl#internalLoad` is the proper one for this issue.  I kept the other change because it is a logical addition

gsmet:
Squashed the 2 commits and merged, thanks!

PR #2714, opened 2019-01-02
Corrects JTS version information in Hibernate 5.4 migration guide

kedar-joshi:
Migration guide for Hibernate v5.4 incorrectly specified that depends on JTS 1.6, in fact the correct JTS version is 1.16 as included in actual maven dependency.

This PR corrects the JTS version information.


gsmet:
Good catch, thanks for the fix!

PR #2716, opened 2019-01-03
HHH-13183

ptahchiev:
A PR for https://hibernate.atlassian.net/browse/HHH-13183

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
@@ -565,7 +565,7 @@ public final class AnnotationBinder {
 				inheritanceState
 		);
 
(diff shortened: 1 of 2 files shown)


gsmet:
@ptahchiev I thought we agreed on my previous PR?

I made some tests and AFAIK we don't support `@AttributeOverride` with TABLE_PER_CLASS right now. It's something we would like to support in the future but it's not working atm.

Or do you have proof that it's indeed working? I might have done something wrong in my tests but they were pretty conclusive.

ptahchiev:
@gsmet  You said that for now the solution would be a warning and `As for supporting TABLE_PER_CLASS, I consider it's a new feature so it should be done in the context of ORM 6.` This is a PR for this new feature.

gsmet:
I see how you changed the warning but I don't see any code that would make the feature actually work? Thus my question.

PR #2718, opened 2019-01-07
HHH-13189 org.hibernate.Query#setParameter(String, Object) is slow

tlamr:
With 50000 params, the time to create the params is 30s on my instance.
After proposed change, the time goes back to ~1s as it used to in 4.3.x release.

https://hibernate.atlassian.net/browse/HHH-13189

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/internal/ParameterMetadataImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/internal/ParameterMetadataImpl.java
@@ -36,8 +36,8 @@ import org.hibernate.type.Type;
 public class ParameterMetadataImpl implements ParameterMetadata {
 	private final Map<Integer,OrdinalParameterDescriptor> ordinalDescriptorMap;
 	private final Map<String,NamedParameterDescriptor> namedDescriptorMap;
-	private final Set<OrdinalParameterDescriptor> ordinalDescriptorValueCache;
-	private final Set<NamedParameterDescriptor> namedDescriptorValueCache;
+	private final Collection<QueryParameter> ordinalDescriptorValueCache;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/query/internal/ParameterMetadataImpl.java
```diff
@@ -36,16 +36,25 @@
...
+	private final Set<OrdinalParameterDescriptor> ordinalDescriptorValueCache;
+	private final Set<NamedParameterDescriptor> namedDescriptorValueCache;
 
 	public ParameterMetadataImpl(
 			Map<Integer,OrdinalParameterDescriptor> ordinalDescriptorMap,
 			Map<String, NamedParameterDescriptor> namedDescriptorMap) {
 		this.ordinalDescriptorMap = ordinalDescriptorMap == null
 				? Collections.emptyMap()
 				: Collections.unmodifiableMap( ordinalDescriptorMap );
+		this.ordinalDescriptorValueCache = ordinalDescriptorMap == null
```
gsmet: Better use `this.ordinalDescriptorMap` and test if it is empty.

Same remark below for the second one.

tlamr: ok, changing, Thanks!

Sanne:
very nice, thanks @tlamr . Testing locally, I'll merge it soon.

Sanne:
merged

Sanne:
Apologies, I just realized my last touch actually ruined your fix.. amending. I really thought I'd still have that backed by a Set but apparently it's not :)

Sanne:
Added 2ab24150f0 , closing for real now.

PR #2719, opened 2019-01-08
HHH-13068 - "order_inserts = true" causes FK Violation when inserting…

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13068

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
@@ -11,6 +11,7 @@ import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
@@ -127,7 +128,7 @@ public class ActionQueue {
 					}
 					ExecutableList<AbstractEntityInsertAction> init(ActionQueue instance) {
 						if ( instance.isOrderInsertsEnabled() ) {
-							return instance.insertions = new ExecutableList<AbstractEntityInsertAction>(
+							return instance.insertions = new ExecutableList<>(
 									new InsertActionSorter()
 							);
 						}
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -1177,70 +1192,137 @@ public void sort(List<AbstractEntityInsertAction> insertions) {
...
-				insertions.addAll( batch );
+		}
+		
+		/**
+		 * Perform the topological sort on the dependency graph such that 
+		 * children should come after the parent
+		 */
+		private void performTopologicalSort() {
+			Set<BatchIdentifier> visitedNodes = new HashSet<>();
+			Stack<BatchIdentifier> stack = new Stack<>();
```
beikov: I'd suggest to use an `ArrayList` instead of a `Stack` which is synchronized.

vladmihalcea:
There are also tests that fail, so we need to investigate those too.

vladmihalcea:
@harikantverma I have worked on fixing the original issue by enhancing the current sorting algorithm. This [PR](https://github.com/hibernate/hibernate-orm/pull/2725) is the one I'm talking about.

Now, related to your topological sort approach, I think you could try and see if you can make it run for all use cases. Currently, some tests are failing, and you might want to check those.

What I really appreciated in your solution is the fact that you allow it to fall back to the original ordering if a cycle is detected. I took that idea and applied it to our current solution. I'll leave this PR open for a while if you want to pursue it further. Thanks.

harikantverma:
> @harikantverma I have worked on fixing the original issue by enhancing the current sorting algorithm. This [PR](https://github.com/hibernate/hibernate-orm/pull/2725) is the one I'm talking about.
> 
> Now, related to your topological sort approach, I think you could try and see if you can make it run for all use cases. Currently, some tests are failing, and you might want to check those.
> 
> What I really appreciated in your solution is the fact that you allow it to fall back to the original ordering if a cycle is detected. I took that idea and applied it to our current solution. I'll leave this PR open for a while if you want to pursue it further. Thanks.

@vladmihalcea thanks for your feedback and for opting the idea i.e. keeping the original order in case of cycle in dependency graph. Meanwhile I am trying to make this solution pass all the test cases.

vladmihalcea:
@harikantverma I integrated the other PR which fixed this issue. You can open a new Jira issue for the sort algorithm, in case you still want to optimize it. Thanks.

PR #2672, opened 2018-11-29
HHH-13129 - Cascaded merge fails for detached bytecode-enhanced entity with uninitialized ToOne

bolekz:
Add test case to reproduce the issue

 * https://hibernate.atlassian.net/browse/HHH-13129

PR #2721, opened 2019-01-09
HHH-13129 : Cascaded merge fails for detached bytecode-enhanced entity with uninitialized ToOne

gbadner:
https://hibernate.atlassian.net/browse/HHH-13129

Supersedes https://github.com/hibernate/hibernate-orm/pull/2672


hibernate-core/src/main/java/org/hibernate/engine/internal/Cascade.java
```diff
@@ -102,25 +100,46 @@ public static void cascade(
...
+							// PersistenceContext. If not present, an uninitialized
+							// PersistentCollection will be added to the PersistenceContext.
+							// The action may initialize it later, if necessary.
+							// This needs to be done even when action.performOnLazyProperty() returns false.
+							final CollectionType collectionType = (CollectionType) types[i];
+							child = collectionType.getCollection(
+									collectionType.getKeyOfOwner( parent, eventSource ),
+									eventSource,
+									parent,
+									null
```
gbadner: Calling `CollectionType#getCollection` means we can remove the call to `Type#resolve` with `LazyPropertyInitializer.UNFETCHED_PROPERTY` as an argument.

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultDeleteEventListener.java
```diff
@@ -131,6 +131,7 @@ public void onDelete(DeleteEvent event, Set transientEntities) throws HibernateE
 					persister,
 					false
 			);
+			persister.afterReassociate( entity, source );
```
gbadner: After adding a detached entity to the `PersistenceContext` (just above the lines shown above), `EntityPersister#afterReassociate` needs to be called.  If the entity is enhanced for lazy loading, then `EntityPersister#afterReassociate` sets the `Session` in an enhanced entity's `LazyAttributeLoadingInterceptor`.

If that's not done, cascade-delete won't work properly.

PR #2723, opened 2019-01-09
HHH-13059 Partially revert HHH-12594

gsmet:
* https://hibernate.atlassian.net/browse/HHH-13059


gbadner:
Fixed the Javadoc comment and squashed it into fix commit, and merged.

PR #2727, opened 2019-01-10
[5.3] HHH-13164 - Revert changes for HHH-11019 and HHH-12464.

Naros:
https://hibernate.atlassian.net/browse/HHH-13164


gbadner:
Merged.

PR #2728, opened 2019-01-11
HHH-13197 Reduce bootstrap log verbosity

gsmet:
* https://hibernate.atlassian.net/browse/HHH-13197


gsmet:
@sebersole so I checked and apparently there are a few others (well, I didn't really check, I was looking for my messages by searching for DEBUG as they were supposed to be the only ones and found a bunch of others).

I would lean towards keeping it this way if you don't mind. It will be easier to revert if people complain and it keeps the existing log ids which is not bad when people look for related issues.

I don't have a strong opinion though so I can make the change if you really think it's best.

TBH, I've already spent far much time on it that I initially thought.

sebersole:
I don't have a strong opinion either so let's just leave it this way.

gsmet:
OK, thanks, will merge.

PR #2730, opened 2019-01-12
HHH-13201 - do not set empty text on collection join parent without queryable collection

Mobe91:
fixes HHH-13201

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/SelectClause.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/SelectClause.java
@@ -193,7 +193,7 @@ public class SelectClause extends SelectExpressionList {
 				FromElement fromElement = (FromElement) iterator.next();
 
 				if ( fromElement.isFetch() ) {
-					FromElement origin;
+					final FromElement origin;
 					if ( fromElement.getRealOrigin() == null ) {
 						// work around that crazy issue where the tree contains
 						// "empty" FromElements (no text); afaict, this is caused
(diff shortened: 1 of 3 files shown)


hibernate-core/src/test/java/org/hibernate/test/hql/FetchNonRootRelativeElementCollectionAndAssociationTest.java
```diff
@@ -0,0 +1,71 @@
...
+import static javax.persistence.CascadeType.ALL;
+import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
+
+/**
+ * @author Moritz Becker (moritz.becker@ordami.com)
+ * @date 12/01/2019
+ * @company ordami GmbH
+ */
+@TestForIssue(jiraKey = "HHH-13201")
+public class FetchNonRootRelativeElementCollectionAndAssociationTest extends BaseEntityManagerFunctionalTestCase {
```
gsmet: Threre are quite a lot of formatting issues in this class.

Could you set up your IDE with our styles: https://github.com/hibernate/hibernate-ide-codestyles and reformat it properly?

(Checkstyle does not enforce styles in tests but let's try to be consistent.)

Thanks!

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/SelectClause.java
```diff
@@ -193,7 +193,7 @@ public void initializeExplicitSelectClause(FromClause fromClause) throws Semanti
 				FromElement fromElement = (FromElement) iterator.next();
 
 				if ( fromElement.isFetch() ) {
-					FromElement origin = null;
+					FromElement origin;
```
gbadner: To ensure `origin` gets initialized if/when code gets changed, the declaration should be:

{{final FromElement origin;}}

Mobe91: Accessing an uninitialized local variable will result in a compile-time error so making this final does not make any difference I think - but I'll change it.

gsmet:
Moving to the 5.4 stream as we don't have the time to review it properly before 5.4.1.

PR #2732, opened 2019-01-13
HHH-13198 Introduce a global configuration flag to disable JPA callbacks

Sanne:
- https://hibernate.atlassian.net/browse/HHH-13198

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/spi/SessionFactoryOptions.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/spi/SessionFactoryOptions.java
@@ -299,4 +299,6 @@ public interface SessionFactoryOptions {
 	default boolean isPostInsertIdentifierDelayableEnabled() {
 		return true;
 	}
+
+	boolean areJPACallbacksEnabled();
 }
--- a/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
@@ -43,12 +43,12 @@ import org.hibernate.event.service.spi.DuplicationStrategy;
 import org.hibernate.event.service.spi.EventListenerRegistrationException;
 import org.hibernate.event.service.spi.EventListenerRegistry;
 import org.hibernate.event.spi.EventType;
-import org.hibernate.jpa.event.internal.CallbackBuilderLegacyImpl;
-import org.hibernate.jpa.event.internal.CallbackRegistryImpl;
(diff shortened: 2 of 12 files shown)


hibernate-core/src/main/java/org/hibernate/boot/spi/SessionFactoryOptions.java
```diff
@@ -299,4 +299,6 @@ default int getQueryStatisticsMaxSize() {
 	default boolean isPostInsertIdentifierDelayableEnabled() {
 		return true;
 	}
+
+	boolean areJPACallbacksEnabled();
```
gsmet: Wondering if we need a default implementation here as it's SPI and it will be a micro?

The above method has one.

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
```diff
@@ -131,17 +128,15 @@ SessionFactoryImplementor getSessionFactory() {
...
 
 	@Override
 	public void prepare(MetadataImplementor metadata) {
 		if ( callbackBuilder == null ) {
 			// TODO : not needed anymore when the deprecate constructor will be removed
-			this.callbackBuilder = new CallbackBuilderLegacyImpl(
-					sessionFactory.getServiceRegistry().getService( ManagedBeanRegistry.class ),
-					metadata.getMetadataBuildingOptions().getReflectionManager()
+			this.callbackBuilder = CallbacksFactory.buildCallbackBuilder( sessionFactory, metadata.getMetadataBuildingOptions().getReflectionManager()
 			);
```
gsmet: Minor code style comment: let's not have this on a single line.

Sanne: ok

hibernate-core/src/main/java/org/hibernate/boot/spi/SessionFactoryOptions.java
Sanne: good point!  I assumed this interface was not to be implemented, but I'm indeed not sure.

Sanne:
> Did you check that you still have the gains you had with your initial implementation?

yeah looks like this works.

There are two aspects; the important point is the methods in `CallbacksFactory` to ignore the `ReflectionManager` parameter, so to not have a long lived reference to it. That's what saves most memory.

The other aspect is to not include the actual implementation bitcode of a `ReflectionManager`; that seems gone as well now, however I'm surprised - the compiler wasn't _that_ smart last time I checked.

Sanne:
@gsmet thanks, this is merged now.

PR #2733, opened 2019-01-14
HHH-13202 - Add support for PostgreSQL "GENERATED BY DEFAULT AS IDENTITY"

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13202

It also contains a test case proving that [HHH-13106](https://hibernate.atlassian.net/browse/HHH-13106) doe snot replicate.


gsmet:
I moved that one to 5.4.2 as we won't have the time to properly review it before the 5.4.1 release and it doesn't seem critical to have it.

PR #2734, opened 2019-01-15
HHH-13204 Introduce a configuration flag to skip processing of XML mapping metadata

Sanne:
- https://hibernate.atlassian.net/browse/HHH-13204

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataBuilderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataBuilderImpl.java
@@ -69,7 +69,6 @@ import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.internal.util.StringHelper;
 import org.hibernate.internal.util.collections.CollectionHelper;
 import org.hibernate.service.ServiceRegistry;
-import org.hibernate.type.AbstractStandardBasicType;
 import org.hibernate.type.BasicType;
 import org.hibernate.type.descriptor.java.JavaTypeDescriptor;
 import org.hibernate.type.descriptor.sql.SqlTypeDescriptor;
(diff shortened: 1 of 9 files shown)


hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataBuilderImpl.java
```diff
@@ -627,6 +628,12 @@ public MetadataBuildingOptionsImpl(StandardServiceRegistry serviceRegistry) {
 
 			this.multiTenancyStrategy =  MultiTenancyStrategy.determineMultiTenancyStrategy( configService.getSettings() );
 
+			this.xmlMetadataEnabled =  configService.getSetting(
+					AvailableSettings.XML_MAPPING_ENABLED,
```
gsmet: Minor nitpick: could we be consistent and either choose metadata or mapping? The attribute is `xmlMetadataEnabled` whereas the setting is `XML_MAPPING_ENABLED`.

Sanne: ok. which one do you want?

gsmet: I don't really care. Mapping makes more sense IMHO but it's the one requiring the most changes so not sure it's the right choice :).

Sanne:
> Shouldn't we add a test to at least basically test this code path?

I'll try, now that I've finished the other urgent things ;)

I don't believe it's very important though. Many existing tests are covering the most important case: of old style mapping still working. If the "disabling" doesn't work, it's just slower - I tested it empirically, adding a test would somewhat protect from regressions but won't protect it to "stay fast"

gsmet:
Yeah, it was more to test that setting it to false doesn't throw an exception somewhere in the bootstrap.

Sanne:
pushed a fresh copy

Sanne:
@gsmet ok to merge?

gsmet:
Adjusted a few minor details and merged. Thanks.

Sanne:
@gsmet your change of comments is confusing.

Read this:
 > Allows to skip processing of XML Mapping. [...] By default it's enabled.

When I originally stated "By default, the XML mapping is taken into account." it's meaning the opposite.

PR #2735, opened 2019-01-16
HHH-13209 : Allow for ServiceRegistry "suspend and restore"

Sanne:
- https://hibernate.atlassian.net/browse/HHH-13209

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataImpl.java
@@ -63,7 +63,7 @@ public class MetadataImpl implements MetadataImplementor, Serializable {
 	private final MetadataBuildingOptions metadataBuildingOptions;
 	private final BootstrapContext bootstrapContext;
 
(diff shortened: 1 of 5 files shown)


gsmet:
@Sanne there is one test failure on Travis:
```
org.hibernate.test.schemaupdate.IdentifierHelperTest > testAutoQuotingDisabled FAILED
    java.lang.UnsupportedOperationException at IdentifierHelperTest.java:39
```
can you reproduce it?

Sanne:
I've not seen that before, I'll run it again..

Sanne:
nope, can't reproduce.

Sanne:
oh actually I can. Seems unstable :(

Sanne:
So it was just a regular bug. Not sure why I couldn't reproduce it initially, probably my fault.

In short: my code was expecting to be able to "clear()" the config map. That actually works fine in practice, except this single test manages to somehow sneak an immutable map as configuration.

That's actually highlighting that I need to take into account the map could have been user-injected by some weird framework over our SPI, so adjusted my patch to deal with it - even if none of our code would hit that today.

gsmet:
Rebased and merged.

PR #1361, opened 2016-05-10
HHH-10085 HHH-10143 - Fix issues with Criteria and case expressions

dreab8:
https://hibernate.atlassian.net/browse/HHH-10143

https://hibernate.atlassian.net/browse/HHH-10085

PR #2736, opened 2019-01-17
HHH-13016 HHH-13199 Fix literal enums in the select clause

gsmet:
I agree it is not pretty but I think it's the best we can do for now to fix these 2 regressions:
 * https://hibernate.atlassian.net/browse/HHH-13199 (regression introduced in 5.4 by fixing a totally valid thing (https://hibernate.atlassian.net/browse/HHH-13001): it's just the bug was hidden before)
 * https://hibernate.atlassian.net/browse/HHH-13016 (regression introduced in 5.2 hidden by the regression introduced in 5.4)

Background: in the select clause literals are supposed to be included as is and not use binding parameters as, apparently, some drivers didn't support parameters in the select clause back in the time (no idea if it's still true, couldn't find any precise info about that).

Rendering enum literals has always been buggy and the only reason it worked is that the `renderProjection()` usage (which is used in the select case) was not propagated correctly when using `case when` or functions so enums worked in this contact by using binding parameters.

I haven't found a proper way to render an enum literal (at least not an easy one) so I decided to bring back the old behavior by forcing it. I also added a few safe guards to avoid NPE. It's definitely not pretty so if you have a better idea, it's welcome...

HHH-13016 had another issue, `CaseLiteralExpression` was removed as part of #1361 and we need it for the particular case of having `case when` expressions at the root of the select (i.e. when we cannot deduce the type automatically). I introduced it back.

@dreab8 do you confirm it was removed in #1361 because you thought it was not useful anymore?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/expression/CaseLiteralExpression.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/expression/CaseLiteralExpression.java
@@ -1,0 +1,38 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.query.criteria.internal.expression;
+
+import org.hibernate.query.criteria.internal.CriteriaBuilderImpl;
+import org.hibernate.query.criteria.internal.compile.RenderingContext;
+import org.hibernate.query.criteria.internal.expression.function.CastFunction;
+
+/**
+ * @author Andrea Boriero
+ */
+public class CaseLiteralExpression<T> extends LiteralExpression<T> {
+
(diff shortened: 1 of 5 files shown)


dreab8:
@gsmet I'm not sure why I removed `CaseLiteralExpression` but I suspect I did it because the solution of wrapping the result in a cast was causing issues in some situations but really not sure, the reason why I introduced it was to solve https://hibernate.atlassian.net/browse/HHH-9731.
Apart the doubts about the reintroduction of `CaseLiteralExpression` the PR looks good to me.

gsmet:
@dreab8 from the commit, I thought maybe it would have seemed useless now that you pushed the expected types. The test suite is green and I suppose younger you would have added test cases, no?

gsmet:
@gbadner btw, not sure about the warns I added. There will be one warning per execution, maybe we should avoid that?

dreab8:
@gsmet I played a little bit with your PR and at the moment I was not able to find another way other than the solution you proposed in order to make the 2 tests you added ( `testCaseEnumResult` and `testCaseEnumInSum`) passing.
So if it is fine for @gbadner I have nothing against merging your PR.

gsmet:
Adjusted the commits a bit and merged. Thanks for the reviews.

PR #2741, opened 2019-01-22
HHH-13219 Change "e.g." to "i.e."

johnlinp:


vladmihalcea:
So, you want to change the "for example" abbreviation to "in other words". Why make this change?

johnlinp:
@vladmihalcea Because the context of the sentence should use "in other words", not "for example". Please read the original sentence.

vladmihalcea:
Applied upstream, thanks

PR #2745, opened 2019-01-22
HHH-13169 - Use exact table name for multitable update queries instead of table alias

gbadner:
https://hibernate.atlassian.net/browse/HHH-13169


gbadner:
Merged.

PR #2746, opened 2019-01-22
HHH-13129 (5.3) Cascaded merge fails for detached bytecode-enhanced entity with uninitialized ToOne 

gbadner:
https://hibernate.atlassian.net/browse/HHH-13129


gbadner:
Merged.

PR #2749, opened 2019-01-22
HHH-10891 (5.3) Exception at bootstrap when @Any is inside an @Embeddable object

gbadner:
https://hibernate.atlassian.net/browse/HHH-10891


gbadner:
Merged.

PR #2750, opened 2019-01-22
HHH-13220 Only create the PersistentAttributeTransformer if the class is enhanced

gsmet:
* https://hibernate.atlassian.net/browse/HHH-13220


gsmet:
Merged!

PR #2751, opened 2019-01-23
HHH-13225 - fix minorVersionToMatch comparison

zeratul021:
...and generify Dialect class for type safety, updated tests.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/dialect/spi/BasicDialectResolver.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/dialect/spi/BasicDialectResolver.java
@@ -21,7 +21,7 @@ public class BasicDialectResolver implements DialectResolver {
 	private final int majorVersionToMatch;
 	private final int minorVersionToMatch;
 
-	private final Class dialectClass;
+	private final Class<? extends Dialect> dialectClass;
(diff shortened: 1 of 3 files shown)


gsmet:
Looks good to me, thanks. I'm making a few changes to the tests so that the `@TestForIssue` annotation is really on the new tests and I'll merge.

gsmet:
I thought a bit more about it and unfortunately `BasicDialectResolver` is considered SPI so something the user could consume and/or extend so we cannot break it for now.

We should do that in 6 though if the class is still there. I'll do that when I forward port this patch to 6.

gsmet:
Merged the part I could merge. Thanks!

zeratul021:
@gsmet I've checked the merged commit seems to contain everything.
Or did you mean that you couldn't merge to current 5.4.x branch (not sure what branching setup/workflow is used here)?

gsmet:
Hmmm, no, it does not contain everything but you're right one one thing: it does contain one of the change I didn't want. I will push an additional commit to remove the `<? extends Dialect>` from the third constructor.

Is it what bugged you or did I miss another thing?

zeratul021:
@gsmet I think thats what mislead me. I saw the SPI change in generic signature and assumed you merged all of it. Thanks

PR #2753, opened 2019-01-23
HHH-13221 - When lazy fetching child entities outside the session DB commit is fired in a read-only transaction.

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13221


gsmet:
This one requires a proper review by Gail, please don't merge it yet.

vladmihalcea:
It's not a critical easy. I tried various scenarios and it was always just one select being triggered, even for join formula, custom loaders or using caching.

We can post pone it as long as it's necessary.

gbadner:
I just pushed a change to the test to use ```FetchMode.SELECT``` to load an association. This forces the association to loaded in a different statement.

I don't think this is a good idea.

vladmihalcea:
@gbadner Thanks for the commit. I think we should close this PR and reject the issue.

PR #2758, opened 2019-01-28
HHH-13234 Log exception causing load to fail as @Cause

galderz:
https://hibernate.atlassian.net/browse/HHH-13234

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/CoreMessageLogger.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/CoreMessageLogger.java
@@ -1115,8 +1115,8 @@ public interface CoreMessageLogger extends BasicLogger {
 	void unableToJoinTransaction(String transactionStrategy);
 
 	@LogMessage(level = INFO)
-	@Message(value = "Error performing load command : %s", id = 327)
-	void unableToLoadCommand(HibernateException e);
+	@Message(value = "Error performing load command", id = 327)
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/internal/CoreMessageLogger.java
```diff
@@ -1115,8 +1115,8 @@ void unableToInstantiateConfiguredSchemaNameResolver(
 	void unableToJoinTransaction(String transactionStrategy);
 
 	@LogMessage(level = INFO)
-	@Message(value = "Error performing load command : %s", id = 327)
-	void unableToLoadCommand(HibernateException e);
+	@Message(value = "Error performing load command :", id = 327)
```
gsmet: Let's remove the trailing ` :` as it makes no sense to have it anymore. Otherwise, looks good and it's something we sure want.

galderz: Fixed

gsmet: Can you also fix it in the 5.3 PR? Thanks!

galderz: Done

gsmet:
Merged, thanks!

PR #2762, opened 2019-01-30
HHH-13239: fix lock timeout on HANA

breglerj:
- Convert the lock wait timeout to seconds by dividing the timeout by 1000, i.e.
  ignoring all fractions of a second

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/AbstractHANADialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/AbstractHANADialect.java
@@ -29,6 +29,7 @@ import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
 import java.sql.Statement;
 import java.sql.Types;
+import java.time.Duration;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


vladmihalcea:
@breglerj I created [this PR](https://github.com/hibernate/hibernate-orm/pull/2763) which adds this change:

```
@Override
public String getWriteLockString(int timeout) {
	long timeoutInSeconds = getLockWaitTimeoutInSeconds( timeout );
	if ( timeoutInSeconds > 0 ) {
		return getForUpdateString() + " wait " + timeoutInSeconds;
	}
	else if ( timeoutInSeconds == 0 ) {
		return getForUpdateNowaitString();
	}
	else {
		return getForUpdateString();
	}
}
```

Basically, we will convert all timeouts that are shorter than 1s to a NO_WAIT.

breglerj:
@vladmihalcea That's fine with me. `NOWAIT` and `WAIT 0` are equivalent.

vladmihalcea:
Great, I'll integrate that one.

PR #2766, opened 2019-02-01
HHH-13244 test case

stoty:
setting hibernate.jpa.compliance.proxy=true and org.hibernate debug level to DEBUG breaks hibernate


gbadner:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2770

PR #2768, opened 2019-02-03
HHH-13246 - isOrdinalParameterZeroBased is unused

marschall:


vladmihalcea:
I created [this PR instead](https://github.com/hibernate/hibernate-orm/pull/2769). I think we should retain this behavior until we completely remove the legacy 0-based ordinal parameter binding.

PR #2769, opened 2019-02-04
HHH-13246 - isOrdinalParameterZeroBased is unused

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13246


vladmihalcea:
@marschall I think it's better to keep the zero-based ordinal binding logic until we remove it for good, which will probably take place in 6.x.

vladmihalcea:
After investigating it, it looks the `isOrdinalParameterZeroBased` is used in the query plan cache key. Removing it is not a good idea since the `isOrdinalParameterZeroBased` is different if we call either `createSQLQuery` or `createNativeQuery`.

PR #2772, opened 2019-02-05
HHH-13194 : Some methods returning org.hibernate.query.Query are not defined for StatelessSession

gbadner:
https://hibernate.atlassian.net/browse/HHH-13194

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/Session.java
+++ b/hibernate-core/src/main/java/org/hibernate/Session.java
@@ -1162,9 +1162,6 @@ public interface Session extends SharedSessionContract, EntityManager, Hibernate
 	void addEventListeners(SessionEventListener... listeners);
 
 	@Override
-	org.hibernate.query.Query createQuery(String queryString);
-
-	@Override
 	<T> org.hibernate.query.Query<T> createQuery(String queryString, Class<T> resultType);
 
 	@Override
@@ -1176,8 +1173,6 @@ public interface Session extends SharedSessionContract, EntityManager, Hibernate
 	@Override
 	org.hibernate.query.Query createQuery(CriteriaDelete deleteQuery);
 
-	@Override
-	org.hibernate.query.Query getNamedQuery(String queryName);
 
 	<T> org.hibernate.query.Query<T> createNamedQuery(String name, Class<T> resultType);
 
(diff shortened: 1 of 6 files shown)


yrodiere:
So, I was wrong about (3): the overrides are actually useful to force the return type to be the one of `QueryProducer` (Hibernate Query) rather than the one of `EntityManager` (JPA Query). The thing is, removing the override doesn't change anything (no compilation error) because most of our tests use `SessionImplementor` instead of `Session`... and as a result they benefit from the overrides in `SessionImplementor`.

That's probably something we should change, as it seems rather dangerous: we could introduce regressions in the Session API without noticing.

Anyway... This also means you should probably not remove the `createQuery(String)` method from `Session`, as I suspect it would effectively change the return type from the Hibernate Query to the JPA one... At least we may want to add a test for that?

gbadner:
@yrodiere, moving the following out of `SharedSessionContract`:

	`org.hibernate.query.Query createQuery(String queryString);`
	`org.hibernate.query.Query getNamedQuery(String queryName);`

and moving the following out of `StatelessSession`:

         `NativeQuery createSQLQuery(String queryString);`

into `QueryProducer` would replace the following methods:

         `org.hibernate.Query createQuery(String queryString);`
         `org.hibernate.Query getNamedQuery(String queryName);`
         `default SQLQuery createSQLQuery(String queryString) {...}`

Making the above change would introduce binary incompatibilities, because in 5.0 and 5.1, the following are defined in `SharedSessionContract`:

         `org.hibernate.Query createQuery(String queryString);`
         `org.hibernate.Query getNamedQuery(String queryName);`
         `SQLQuery createSQLQuery(String queryString);`

After making the suggested change, I see the following in the binary compatibility report:

***! MODIFIED INTERFACE: PUBLIC ABSTRACT org.hibernate.SharedSessionContract  (default serialVersionUID changed)
	---! REMOVED METHOD: PUBLIC(-) ABSTRACT(-) org.hibernate.Query createQuery(java.lang.String)
	---! REMOVED METHOD: PUBLIC(-) ABSTRACT(-) org.hibernate.SQLQuery createSQLQuery(java.lang.String)
	---! REMOVED METHOD: PUBLIC(-) ABSTRACT(-) org.hibernate.Query getNamedQuery(java.lang.String)

I'm not following what you are suggesting for 3), but I agree it is out of scope for this issue. If you think it's a worthwhile change, then please create a new issue. A PR would help clarify what you are suggesting.

gbadner:
Merged.

PR #2773, opened 2019-02-06
HHH-13265 - Remove double semicolon

wonwoo:


vladmihalcea:
Applied upstream, thanks

PR #2779, opened 2019-02-12
HHH-13261 - @ManyToAny is to emulate @ManyToMany rather than @OneToMany

johnlinp:
https://hibernate.atlassian.net/browse/HHH-13261


vladmihalcea:
I closed the issue since the current documentation is correct in regards to `@ManyToAny` behavior.

PR #2780, opened 2019-02-12
HHH-13262 - javax.persistence.TransactionRequiredException: Executing an update/delete query

dreab8:
https://hibernate.atlassian.net/browse/HHH-13262

Any suggestion for better name for the method `checkTransactionNeededForUpdateOperation` is more than welcome.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
@@ -12,6 +12,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.UUID;
 import javax.persistence.FlushModeType;
+import javax.persistence.TransactionRequiredException;
 
 import org.hibernate.CacheMode;
 import org.hibernate.Criteria;
@@ -179,6 +180,26 @@ public interface SharedSessionContractImplementor
 	 * or is there a JTA transaction in progress?
 	 */
 	boolean isTransactionInProgress();
+
+	/**
+	 * Check if an active Transaction is necessary for the update operation to be executed.
+	 * If an active Transaction is necessary but it is not then a TransactionRequiredException is raised.
+	 *
+	 * @param exceptionMessage, the message to use for the TransactionRequiredException
+	 * @param convertException, should a ExceptionConverter be used for the TransactionRequiredException
+	 */
+	default void checkTransactionNeededForUpdateOperation(String exceptionMessage, boolean convertException) {
+		if ( !isTransactionInProgress() ) {
+			if ( convertException ) {
+				throw getExceptionConverter().convert(
+						new TransactionRequiredException(
+								exceptionMessage
+						)
+				);
+			}
+			throw new TransactionRequiredException( exceptionMessage );
+		}
+	}
 
 	/**
 	 * Provides access to the underlying transaction or creates a new transaction if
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/query/internal/AbstractProducedQuery.java
```diff
@@ -1606,13 +1606,8 @@ public R getSingleResult() {
...
 	public int executeUpdate() throws HibernateException {
-		if ( ! getProducer().isTransactionInProgress() ) {
-			throw getProducer().getExceptionConverter().convert(
-					new TransactionRequiredException(
-							"Executing an update/delete query"
-					)
-			);
-		}
+		getProducer().checkTransactionNeededForUpdateOperation( "Executing an update/delete query", true );
+
```
gbadner: This above is the only time that ```#checkTransactionNeededForUpdateOperation``` is called with flag for converting the exception set to ```true```.

I see that the original code converts the exception, but I'm not sure if it is necessary to do so, since ```TransactionRequiredException``` is a JPA exception. All that calling ```ExceptionConverter#convert(RuntimeException)``` will do is mark the transaction for rollback. Since we know there is no transaction in progress, then that is a no-op.

I'm curious why the exception needs to be converted here, but not in the other places.

There is no Javadoc for ```ExceptionConverter#convert(RuntimeException)``` to indicate when the method should or should not be called, so that doesn't give guidance.

It seems that when ```#checkTransactionNeededForUpdateOperation``` is called, it should always be called with ```true``` or always with ```false```. To simplify things, I would prefer that ```#checkTransactionNeededForUpdateOperation``` not have the ```convertException``` argument, and that the method would do the right thing consistently.

hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
```diff
@@ -180,6 +181,26 @@ default long getTimestamp() {
...
 	boolean isTransactionInProgress();
 
+	/**
+	 * Check if an active Transaction is necessary for the update operation to be executed.
+	 * If an active Transaction is necessary but it is not then a TransactionRequiredException is raised.
+	 *
+	 * @param exceptionMessage, the message to use for the TransactionRequiredException
+	 * @param convertException, should a ExceptionConverter be used for the TransactionRequiredException
+	 */
+	default void checkTransactionNeededForUpdateOperation(String exceptionMessage, boolean convertException) {
```
gbadner: I don't think it should be necessary to have the ```convertException``` argument.

Please see https://github.com/hibernate/hibernate-orm/pull/2780/files#r258717383

hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
```diff
@@ -389,6 +392,20 @@ public boolean isTransactionInProgress() {
 		return !isClosed() && transactionCoordinator.isTransactionActive();
 	}
 
+	@Override
+	public void checkTransactionNeededForUpdateOperation(String exceptionMessage, boolean convertException) {
```
gbadner: I don't think it should be necessary to have the ```convertException``` argument.

Please see https://github.com/hibernate/hibernate-orm/pull/2780/files#r258717383

gbadner:
To be more precise, the name could be changed to ```#checkTransactionNeededForUpdateOrLockOperation``. I suppose a lock operation could be considered a sort of update operation though (even though it's not the data that is being updated).

hibernate-core/src/main/java/org/hibernate/query/internal/AbstractProducedQuery.java
dreab8: I checked and if I remove the conversion `org.hibernate.test.jpa.compliance.tck2_2.QueryApiTest#testUpdateRequiresTxn` fails.

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/2791

PR #2781, opened 2019-02-14
HHH-13267 - Incredibly silly reverting removal of 7-year old deprecat…

sebersole:
…ion of JDBC-style parameters in HQL queries

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
@@ -63,8 +63,6 @@ import org.hibernate.service.spi.ServiceRegistryImplementor;
 import org.hibernate.tuple.entity.EntityTuplizer;
 import org.hibernate.tuple.entity.EntityTuplizerFactory;
 
-import org.jboss.logging.Logger;
-
 import static org.hibernate.cfg.AvailableSettings.ACQUIRE_CONNECTIONS;
 import static org.hibernate.cfg.AvailableSettings.ALLOW_JTA_TRANSACTION_ACCESS;
 import static org.hibernate.cfg.AvailableSettings.ALLOW_REFRESH_DETACHED_ENTITY;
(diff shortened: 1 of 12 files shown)


gbadner:
I tried running the test using 5.1 EntityManager, and the tests operate differently. I need to look into this more deeply so I can provide more details.

gbadner:
@sebersole, ```ViburDBCPConnectionProviderTest``` is failing. See: https://travis-ci.org/hibernate/hibernate-orm/builds/493500681?utm_source=github_status&utm_medium=notification

The problem is with this query:

`List<Actor> list = session.createQuery("from Actor where firstName = ?0")`
`                .setParameter(0, "CHRISTIAN").list();`

Apparently, that worked before your fix, and I see the same query in master for 5.4.

After your fix, it needs to be changed to one of the following to work.

`List<Actor> list = session.createQuery("from Actor where firstName = ?1")`
`                .setParameter(1, "CHRISTIAN").list();`

or

```hibernate.query.sql.jdbc_style_params_base=true``` must be set, and the query changed:
`List<Actor> list = session.createQuery("from Actor where firstName = ?")`
`                .setParameter(0, "CHRISTIAN").list();`

I see that your test, ```LegacyPositionalParameterTest#testJpaParameterUsageIncorrectBase```, expects a "?0" as a parameter to fail. Should this actually work in 5.3/5.4?

gbadner:
I changed the query for ```ViburDBCPConnectionProviderTest``` to the following, which also works in master:

`List<Actor> list = session.createQuery("from Actor where firstName = ?5")`
`                .setParameter(5, "CHRISTIAN").list();`

The test, ```LegacyPositionalParameterTest#testJpaParameterUsageIncorrectLabel``` is similar, and it expects the query to fail.

gbadner:
I believe the other differences were expected. I'd like to confirm this though.

In 5.1, HQL queries using '?' require parameter numbers to start at 0.
In 5.1, EM queries using '?' require parameter numbers to start at 1.
After fix to 5.3, ```hibernate.query.sql.jdbc_style_params_base=true``` must be set in order for '?' to start numbering at 0; otherwise parameter numbers start at 1.

sebersole:
I am not sure where the expectation that `?0` in an HQL or JPQL query would ever work.  A "labelled positional" parameter (`?` followed by an int) is an idea JPA added.  JPA says that the labels must be contiguous and must start with 1.  Attempting to use `?0` *should* fail

gbadner:
@sebersole, I created some tests to show how things work in 5.1 and master.

The tests are written so they pass.

In 5.1 (https://github.com/gbadner/hibernate-core/commit/63414349c8ae6855e8624984a91e4ee11a43d7ef):

In master (https://github.com/gbadner/hibernate-core/commit/10a3da6c99bc7b0179041c067765331fb76f2514):

Summary:

Using parameter ?0 only:
* 5.1 native HQL: throws QueryParameterException
* 5.1 EM query: query executes without any problem
* master: query executes without any problem

Using parameter ?5 only:
* 5.1 native HQL: throws QueryParameterException
* 5.1 EM query: query executes without any problem
* master: query executes without any problem

Using parameters ?0 and ?5 in same query
* 5.1 native HQL: throws QueryParameterException
* 5.1 EM query: query executes without any problem
* master: throws QueryParameterException

PR #1844, opened 2017-03-11
HHH 11544 - Test and fix

beikov:
https://hibernate.atlassian.net/browse/HHH-11544

PR #2782, opened 2019-02-15
HHH-13269 - Embeddable collection regression due to HHH-11544

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13269


vladmihalcea:
@beikov I created this fix for the [HHH-13269](https://hibernate.atlassian.net/browse/HHH-13269) issue which adds a check for the `EntityQuerySpace` when creating the `Join` in the `LoadQueryJoinAndFetchProcessor`.

Let me know what you think.

gsmet:
@vladmihalcea I think we have the exact same problem here in the original PR: https://github.com/hibernate/hibernate-orm/pull/1844/files#diff-09dac74f012e7d7ef06e77ab10a33451R509

vladmihalcea:
@gsmet @beikov Good catch. I updated the PR to include a check in `DotNode` as well. All tests pass successfully.

PR #2783, opened 2019-02-16
HHH-13226 : Corrected Typo

asutosh936:


gsmet:
Thanks! Unfortunately, it's a bit more complicated: you need to keep the old ones and deprecate them as people might already use them.

Both properties would need to be supported (with the new one taking precedence over the old one if both are set).

asutosh936:
@gsmet - Have tried to implement the review comments You provided. Not sure the changes are correct but kindly review and let me know how I can help further on this issue.

Sanne:
hi @asutosh936 , 

you're going in the right direction, but you also need to ensure that:
 - the new property is being applied when people set it (search for references to the contstants in the code)
 - the old property is also applied, yet logs a warning about using a deprecated configuration property

See for example `org.hibernate.tool.schema.internal.Helper#interpretNamespaceHandling` ; there might be more such cases.

asutosh936:
@gsmet @Sanne - Kindly review and merge if changes looks good.

gsmet:
@asutosh936 are you sure you pushed your changes? I don't see any compatibility code?

asutosh936:
@gsmet - Have pushed the changes.

Sanne:
thanks @asutosh936 , I merged your fix and added a couple additional commits to address some more backwards compatibility aspects.

PR #2787, opened 2019-02-20
HHH-13233 add missing implementation to add attribute to subgraphs

zeitiger:
Hi,

I would like to fix this problem that blocks me from upgrade hibernate in my project. I hope the problem is really that easy to fix. 

My problem is that attributes of subgraphs get not fetched after 5.4, 5.3 works fine.

Have a nice day
David


gbadner:
@zeitiger, please add a test case that reproduces your issue.

gsmet:
@zeitiger as mentioned by Gail, we need a test case to be sure we don't introduce regressions.

Thanks!

zeitiger:
I add the requested test case by reusing an existing test case that used the same API differently and extend this existing test case also to verify entityGraph structure too.

zeitiger:
@gsmet are we good now?

gsmet:
Perfect, thanks, will be part of 5.4.2.

zeitiger:
@gsmet without merge?

PR #2789, opened 2019-02-21
HHH-13280 - HINT_PASS_DISTINCT_THROUGH does not work if 'hibernate.use_sql_comments' property is enabled

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13280

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
@@ -2883,8 +2883,20 @@ public abstract class Loader {
 			String sql,
 			QueryParameters parameters) {
 		if ( !parameters.isPassDistinctThrough() ) {
-			if ( sql.startsWith( SELECT_DISTINCT ) ) {
-				return SELECT + sql.substring( SELECT_DISTINCT.length() );
+			int selectDistinctIndex = sql.indexOf( SELECT_DISTINCT );
+
+			if ( selectDistinctIndex > -1 ) {
+
+				StringBuilder builder = ( selectDistinctIndex > 0 ) ?
+						new StringBuilder(
+								sql.substring( 0, selectDistinctIndex )
+						) :
+						new StringBuilder();
+
+				builder.append( SELECT )
+						.append( sql.substring( selectDistinctIndex + SELECT_DISTINCT.length() ) );
+
+				return builder.toString();
 			}
 		}
 		return sql;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/loader/Loader.java
```diff
@@ -2883,8 +2883,8 @@ protected String processDistinctKeyword(
 			String sql,
 			QueryParameters parameters) {
 		if ( !parameters.isPassDistinctThrough() ) {
-			if ( sql.startsWith( SELECT_DISTINCT ) ) {
-				return SELECT + sql.substring( SELECT_DISTINCT.length() );
+			if ( sql.contains( SELECT_DISTINCT ) ) {
+				return sql.replaceAll( SELECT_DISTINCT, SELECT );
```
gsmet: Hmmm, this changes the semantic of what we do significantly as we now replace every occurrence in the query.

There's no risk at doing that?

vladmihalcea: This option must be explicitly activated, and it's purpose is to avoid executing the DISTINCT at the SQL level. Only if the de-duplication distinct functionality is mixed with SQL-level distinct would this be an issue. Anyway, I could rework it to remove only the top-level distinct since that's the one we are interested in.

gsmet: I think it would be safer if you can find a way to do it.

vladmihalcea: @gsmet Done and updated the PR. Now, only the first occurrence is transformed leaving the option to use ANY, ALL queries where distinct is left as-is. I also a tets for that proving that as well, so I suppose it's going to be ok to proceed with this change.

gbadner:
I've added some failing tests.

If the comment contains "distinct", then it is removed from the comment only. The default comment for a query is the HQL query itself, so this fix does not work unless an explicit comment is added that does not have "distinct" in it.

Also, if there is a subquery with "distinct" and an explicit comment that does not have "distinct", then "distinct" is removed from the subquery.

IIUC, "distinct" should only be removed from the outer-most ```select```.

IMO, "distinct" should not be removed from a comment. WDYT?

vladmihalcea:
@gbadner Thanks for the tests. Your assumptions are correct. We need to rework it to make sure the "distinct" is never removed from a comment or hint and only for the top-level select clasue.

japro:
Hi all, is this issue please planned to be fixed in the near future? Currently I cannot have comments in my hibernate logs because of this. Thanks a lot! :)

beikov:
@japro If you provide an update of this PR that fixes this in a proper way (address Guillaume's comments) rebased on top of main, I can look into this.

PR #2790, opened 2019-02-21
HHH-13283: Backport HHH-13112 to branch 5.3 - Proxies on entity types in the default package lead to MappingException with JDK9+

yrodiere:
https://hibernate.atlassian.net//browse/HHH-13283

### DIFF
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -23,7 +23,7 @@ ext {
     weldVersion = '3.0.0.Final'
 
     javassistVersion = '3.23.1-GA'
-    byteBuddyVersion = '1.9.4'
+    byteBuddyVersion = '1.9.5'
 
     geolatteVersion = '1.3.0'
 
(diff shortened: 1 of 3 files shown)


yrodiere:
@Sanne I'd like your opinion on this, since I have no idea why we upgraded to bytebuddy 1.9.4 instead of the latest.

gsmet:
@yrodiere so the reason why we didn't upgrade to latest was to have some field tested version instead of the shiny new one.

@Sanne upgraded to 1.9.10 in the 5.4 branch so I think we can also upgrade to 1.9.10 here. I will release 5.4.2 soon and we will have some feedback before the inclusion of 5.3.9 in next WF.

Sanne:
N.B. the upgrade to 1.9.10 on 5.4 branch is from just a few days back. The driver was the need to run tests on JDK13.

I see no specific problem to upgrade to 1.9.10 but in spirit of "field tested version" maybe one of the versions actually shipped by a 5.4 release would be safer?

PR #2791, opened 2019-02-21
HHH-13262 - javax.persistence.TransactionRequiredException: Executing an update/delete query

dreab8:
https://hibernate.atlassian.net/browse/HHH-13262


gbadner:
@dreab8, I don't see any failures after removing the conversion. I've pushed a commit that removed them.

Am I missing something?

dreab8:
@gbadner I run the tests and no failures too, not sure why I had the failing tests may be I did something wrong on my side.

PR #2792, opened 2019-02-21
HHH-13277 Make HibernateMethodLookupDispatcher less fragile

gsmet:
And less dependent of the JVM.

 * https://hibernate.atlassian.net/browse/HHH-13277

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/HibernateMethodLookupDispatcher.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/HibernateMethodLookupDispatcher.java
@@ -10,7 +10,6 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
@@ -23,9 +22,12 @@ public class HibernateMethodLookupDispatcher {
 	private static final SecurityActions SECURITY_ACTIONS = new SecurityActions();
 
 	private static final Function<Object, Class<?>> STACK_FRAME_GET_DECLARING_CLASS_FUNCTION;
+	@SuppressWarnings("rawtypes")
+	private static final Function<Stream, Object> STACK_FRAME_EXTRACT_FUNCTION;
 	private static Object stackWalker;
 	private static Method stackWalkerWalkMethod;
 	private static Method stackFrameGetDeclaringClass;
+	private static final PrivilegedAction<Class<?>> GET_CALLER_CLASS_ACTION;
 
 	// Currently, the bytecode provider is created statically and shared between all the session factories. Thus we
 	// can't clear this set when we close a session factory as we might remove elements coming from another one.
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/HibernateMethodLookupDispatcher.java
```diff
@@ -142,25 +141,39 @@ public Void run() {
...
+					boolean hibernateMethodLookupDispatcherDetected = false;
+					// start at the 4th frame and limit that to the 16 first frames
+					for ( int i = 3; i < Math.min( 16, stackTrace.length ); i++ ) {
+						if ( stackTrace[i].getName().equals( HibernateMethodLookupDispatcher.class.getName() ) ) {
+							hibernateMethodLookupDispatcherDetected = true;
+							continue;
+						}
+						if ( hibernateMethodLookupDispatcherDetected ) {
+							return stackTrace[i];
+						}
```
yrodiere: I would swap the two ifs, but maybe that's just me.

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/HibernateMethodLookupDispatcher.java
```diff
@@ -142,25 +141,39 @@ public Void run() {
 			@SuppressWarnings({ "unchecked", "rawtypes" })
 			public Class<?> run() {
 				try {
+					Class<?>[] stackTrace;
 					if ( stackWalker != null ) {
-						Optional<Class<?>> clazzOptional = (Optional<Class<?>>) stackWalkerWalkMethod.invoke( stackWalker, new Function<Stream, Object>() {
+						stackTrace = (Class<?>[]) stackWalkerWalkMethod.invoke( stackWalker, new Function<Stream, Object>() {
```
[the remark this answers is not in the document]
yrodiere: I think that would allow you to get rid of the `stackWalker` and  `stackWalkerWalkMethod` constants, by the way.

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/HibernateMethodLookupDispatcher.java
```diff
@@ -142,25 +141,39 @@ public Void run() {
...
+						if ( stackTrace[i].getName().equals( HibernateMethodLookupDispatcher.class.getName() ) ) {
+							hibernateMethodLookupDispatcherDetected = true;
+							continue;
+						}
+						if ( hibernateMethodLookupDispatcherDetected ) {
+							return stackTrace[i];
+						}
+					}
+
+					return null;
```
yrodiere: Wouldn't the correct behavior be to throw an exception stating you couldn't determine the caller class? It looks like it was the previous behavior. Right now you're going to throw an exception saying something like "not authorized".

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/HibernateMethodLookupDispatcher.java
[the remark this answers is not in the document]
gsmet: I improved things there.

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/HibernateMethodLookupDispatcher.java
gsmet: Fixed.

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/HibernateMethodLookupDispatcher.java
gsmet: The logic wouldn't be the same. We need to ignore several `HibernateMethodLookupDispatcher` in a row.

gbadner:
I see that there are some commits after @yrodiere approved this. Is it still approved?

gsmet:
More or less. I checked it yesterday but will make a last review on Monday before merging.

Syntobe:
I can confirm that this approach works on IBM JVM (Domino 9.0.1FP10 HF383 - sorry, I don't know the exact version of the internal IBM JVM used there)

hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/HibernateMethodLookupDispatcher.java
```diff
@@ -83,99 +85,132 @@ static void registerAuthorizedClass(String className) {
...
+				: initializeGetCallerStackAction.run();
+	}
+
+	private static Class<?> getCallerClass() {
+		Class<?>[] stackTrace = System.getSecurityManager() != null
+				? AccessController.doPrivileged( GET_CALLER_STACK_ACTION )
+				: GET_CALLER_STACK_ACTION.run();
+
+		// this shouldn't happen but let's be safe
+		if ( stackTrace.length < 4 ) {
```
yrodiere: By the way, this could be `3`. Then you should be able to put it in a constant (same for the `3` in the loop below).

PR #2793, opened 2019-02-24
HHH-13285 : ClassCastException: org.dom4j.DocumentFactory cannot be cast to org.dom4j.DocumentFactory after dom4j update (5.3)

gbadner:
https://hibernate.atlassian.net/browse/HHH-13285 (5.3)


hibernate-core/src/main/java/org/hibernate/internal/util/xml/XMLHelper.java
```diff
@@ -22,23 +23,34 @@
 public final class XMLHelper {
 	private final DocumentFactory documentFactory;
 
-	public XMLHelper(ClassLoaderService classLoaderService) {
```
scottmarlow: Do Hibernate Search or other Hibernate projects new up XMLHelper via XMLHelper(ClassLoaderService classLoaderService)?

scottmarlow: Overall, the change looks good to me.

gbadner:
Merged.

hibernate-core/src/main/java/org/hibernate/internal/util/xml/XMLHelper.java
scottmarlow: If there is concern about existing references to this internal class from other Hibernate projects, we could add XMLHelper(ClassLoaderService) ctor back in.

PR #2795, opened 2019-02-24
HHH-13281 java.lang.ClassCastException: org.hibernate.internal.SessionImpl cannot be cast to org.hibernate.ejb.HibernateEntityManager (5.3)

gbadner:
https://hibernate.atlassian.net/browse/HHH-13281 (5.3)


gbadner:
Merged.

PR #2798, opened 2019-02-26
Remove superflous logging

jonasfugedi:
This log is spamming our system in a scenario where we catch the exception as it is expected behaviour. The logging should not be performed by a mapper but by the code handling the exception. Suggest possibly to set it at debug or info level if needed for diagnostics.


Sanne:
thanks @jonasfugedi , good point! Creating an issue, then I'll rebase and merge.

Sanne:
merged as https://hibernate.atlassian.net/browse/HHH-13306

PR #2800, opened 2019-02-27
HHH-13266 LocalDateTime values are wrong around 1900 (caused by JDK-8061577)

yrodiere:
https://hibernate.atlassian.net//browse/HHH-13266

Adding Gail because I think we should backport this to 5.3.

I added a comment to the ticket: https://hibernate.atlassian.net/browse/HHH-13266?focusedCommentId=104778&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-104778

In short, there's a bug in the java.util classes that leads to "drifts" when converting between year/month/day/hours/etc. to milliseconds since the epoch, and back. This only happens before a certain point in time around 1900 (different depending on the timezone).
Generally the bug is rather innocuous because we happen to run two conversions, and the two conversions fortunately drift in a way that compensate each other.
However, near 1900, there's a chance that the first conversion make the date drift after the breaking point, meaning the second conversion will not lead to a drift, and we'll end up storing the wrong, "drifted" value (and also giving this value back to the user when reading).

Anyway... The new strategy for converting LocalDateTime that was suggested will work, but there's still a bug internally in the JDK. I *think* we're relatively safe with the new approach, and we won't suffer from this bug anymore. From what I can see, the new strategy is also backward compatible with the old one: it will read back from the database exactly like the old strategy would have, and it will write the same values to the database, except for the bug we're fixing.
However, users previously affected by this bug already have stored incorrect values in the database, and there's nothing we can do about this. If they load an entity and write it back to the database, the incorrect value will still be incorrect.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/InstantJavaDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/InstantJavaDescriptor.java
@@ -61,7 +61,13 @@ public class InstantJavaDescriptor extends AbstractTypeDescriptor<Instant> {
 		}
 
 		if ( java.sql.Timestamp.class.isAssignableFrom( type ) ) {
-			return (X) Timestamp.from( instant );
+			/*
+			 * Workaround for HHH-13266 (JDK-8061577).
+			 * Ideally we'd want to use Timestamp.from(), but this won't always work.
+			 * Timestamp.from() assumes the number of milliseconds since the epoch
+			 * means the same thing in Timestamp and Instant, but it doesn't, in particular before 1900.
+			 */
+			return (X) Timestamp.valueOf( instant.atZone( ZoneId.systemDefault() ).toLocalDateTime() );
 		}
 
 		if ( java.sql.Date.class.isAssignableFrom( type ) ) {
@@ -95,7 +101,13 @@ public class InstantJavaDescriptor extends AbstractTypeDescriptor<Instant> {
 
 		if ( Timestamp.class.isInstance( value ) ) {
 			final Timestamp ts = (Timestamp) value;
-			return ts.toInstant();
+			/*
+			 * Workaround for HHH-13266 (JDK-8061577).
+			 * Ideally we'd want to use ts.toInstant(), but this won't always work.
+			 * ts.toInstant() assumes the number of milliseconds since the epoch
+			 * means the same thing in Timestamp and Instant, but it doesn't, in particular before 1900.
+			 */
+			return ts.toLocalDateTime().atZone( ZoneId.systemDefault() ).toInstant();
 		}
 
 		if ( Long.class.isInstance( value ) ) {
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocalDateTimeJavaDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocalDateTimeJavaDescriptor.java
@@ -55,8 +55,14 @@ public class LocalDateTimeJavaDescriptor extends AbstractTypeDescriptor<LocalDat
 		}
 
 		if ( java.sql.Timestamp.class.isAssignableFrom( type ) ) {
-			Instant instant = value.atZone( ZoneId.systemDefault() ).toInstant();
-			return (X) java.sql.Timestamp.from( instant );
+			/*
+			 * Workaround for HHH-13266 (JDK-8061577).
+			 * We used to do Timestamp.from( value.atZone( ZoneId.systemDefault() ).toInstant() ),
+			 * but on top of being more complex than the line below, it won't always work.
+			 * Timestamp.from() assumes the number of milliseconds since the epoch
+			 * means the same thing in Timestamp and Instant, but it doesn't, in particular before 1900.
+			 */
+			return (X) Timestamp.valueOf( value );
 		}
 
 		if ( java.sql.Date.class.isAssignableFrom( type ) ) {
@@ -98,7 +104,14 @@ public class LocalDateTimeJavaDescriptor extends AbstractTypeDescriptor<LocalDat
 
 		if ( Timestamp.class.isInstance( value ) ) {
 			final Timestamp ts = (Timestamp) value;
-			return LocalDateTime.ofInstant( ts.toInstant(), ZoneId.systemDefault() );
+			/*
+			 * Workaround for HHH-13266 (JDK-8061577).
+			 * We used to do LocalDateTime.ofInstant( ts.toInstant(), ZoneId.systemDefault() ),
+			 * but on top of being more complex than the line below, it won't always work.
+			 * ts.toInstant() assumes the number of milliseconds since the epoch
+			 * means the same thing in Timestamp and Instant, but it doesn't, in particular before 1900.
+			 */
+			return ts.toLocalDateTime();
 		}
 
 		if ( Long.class.isInstance( value ) ) {
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ZonedDateTimeJavaDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ZonedDateTimeJavaDescriptor.java
@@ -59,7 +59,13 @@ public class ZonedDateTimeJavaDescriptor extends AbstractTypeDescriptor<ZonedDat
 		}
 
 		if ( Timestamp.class.isAssignableFrom( type ) ) {
-			return (X) Timestamp.from( zonedDateTime.toInstant() );
+			/*
+			 * Workaround for HHH-13266 (JDK-8061577).
+			 * Ideally we'd want to use Timestamp.from( zonedDateTime.toInstant() ), but this won't always work.
+			 * Timestamp.from() assumes the number of milliseconds since the epoch
+			 * means the same thing in Timestamp and Instant, but it doesn't, in particular before 1900.
+			 */
+			return (X) Timestamp.valueOf( zonedDateTime.withZoneSameInstant( ZoneId.systemDefault() ).toLocalDateTime() );
 		}
 
 		if ( java.sql.Date.class.isAssignableFrom( type ) ) {
@@ -93,7 +99,13 @@ public class ZonedDateTimeJavaDescriptor extends AbstractTypeDescriptor<ZonedDat
 
 		if ( java.sql.Timestamp.class.isInstance( value ) ) {
 			final Timestamp ts = (Timestamp) value;
-			return ZonedDateTime.ofInstant( ts.toInstant(), ZoneId.systemDefault() );
+			/*
+			 * Workaround for HHH-13266 (JDK-8061577).
+			 * Ideally we'd want to use ZonedDateTime.ofInstant( ts.toInstant(), ... ), but this won't always work.
+			 * ts.toInstant() assumes the number of milliseconds since the epoch
+			 * means the same thing in Timestamp and Instant, but it doesn't, in particular before 1900.
+			 */
+			return ts.toLocalDateTime().atZone( ZoneId.systemDefault() );
 		}
 
 		if ( java.util.Date.class.isInstance( value ) ) {
--- a/hibernate-core/src/test/java/org/hibernate/test/type/InstantTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/type/InstantTest.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.test.type;
 
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
@@ -85,12 +86,12 @@ public class InstantTest extends AbstractJavaTimeTypeTest<Instant, InstantTest.E
 	}
 
 	@Override
-	protected EntityWithInstant createEntity(int id) {
-		return new EntityWithInstant( id, getExpectedPropertyValue() );
+	protected EntityWithInstant createEntityForHibernateWrite(int id) {
+		return new EntityWithInstant( id, getExpectedPropertyValueAfterHibernateRead() );
 	}
 
 	@Override
-	protected Instant getExpectedPropertyValue() {
+	protected Instant getExpectedPropertyValueAfterHibernateRead() {
 		return OffsetDateTime.of( year, month, day, hour, minute, second, nanosecond, ZoneOffset.UTC ).toInstant();
 	}
 
@@ -100,8 +101,13 @@ public class InstantTest extends AbstractJavaTimeTypeTest<Instant, InstantTest.E
 	}
 
 	@Override
-	protected Object getExpectedJdbcValue() {
-		LocalDateTime dateTimeInDefaultTimeZone = getExpectedPropertyValue().atZone( ZoneId.systemDefault() )
+	protected void setJdbcValueForNonHibernateWrite(PreparedStatement statement, int parameterIndex) throws SQLException {
+		statement.setTimestamp( parameterIndex, getExpectedJdbcValueAfterHibernateWrite() );
+	}
+
+	@Override
+	protected Timestamp getExpectedJdbcValueAfterHibernateWrite() {
+		LocalDateTime dateTimeInDefaultTimeZone = getExpectedPropertyValueAfterHibernateRead().atZone( ZoneId.systemDefault() )
 				.toLocalDateTime();
 		return new Timestamp(
 				dateTimeInDefaultTimeZone.getYear() - 1900, dateTimeInDefaultTimeZone.getMonthValue() - 1,
--- a/hibernate-core/src/test/java/org/hibernate/test/type/InstantTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/type/InstantTest.java
@@ -6,91 +6,57 @@
  */
 package org.hibernate.test.type;
 
-import java.sql.PreparedStatement;
 import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
-import java.util.Arrays;
 import java.util.List;
-import java.util.TimeZone;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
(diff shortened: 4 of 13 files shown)


hibernate-core/src/test/java/org/hibernate/test/type/LocalDateTimeTest.java
```diff
@@ -0,0 +1,209 @@
...
+		withDefaultTimeZone( zoneId, () -> {
+			inTransaction( session -> {
+				session.persist( new EntityWithLocalDateTime( 1, localDateTime ) );
+			} );
+			inTransaction( session -> {
+				Timestamp nativeRead = (Timestamp) session.createNativeQuery(
+						"SELECT thevalue FROM theentity WHERE theid = :id"
+				)
+						.setParameter( "id", 1 )
+						.uniqueResult();
```
gbadner: I just want to mention that, in general, when Hibernate executes a native query, the return type may not be exactly what gets returned by raw JDBC. In this particular case, it is.

Hibernate uses ```ResultSetMetaData``` to determine the SQL type of the column. In this case, the column is of type ```java.sql.Types.TIMESTAMP```. By default, ```java.sql.Types.TIMESTAMP``` is mapped to ```TimestampType.INSTANCE```, which delegates to ```JdbcTimestampTypeDescriptor``` for wrapping and unwrapping. When a value of type ```java.sql.Timestamp``` is wrapped or unwrapped, the value is simply cast to ```java.sql.Timestamp```.

I needed to walk through this in the debugger to convince myself that Hibernate was not converting the value that got returned by the database in any way.

This test appears to safely do what it needs to do, but I would prefer to see the native query executed and the return value obtained using raw JDBC in an implementation of the ```org.hibernate.jdbc.Work``` API.

gbadner:
I just pushed a commit that changes native query execution to use the {{Work}} API.

yrodiere:
Thanks for the fix. I will have a look at what the other types mentioned in the ticket tomorrow.

yrodiere:
I just force-pushed an update that tests and fixes the descriptors for LocalDateTime, LocalDate, ZoneDateTime, OffsetDateTime and Instant. I think that should be it for this ticket.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/InstantJavaDescriptor.java
```diff
@@ -95,7 +101,13 @@ public Instant fromString(String string) {
...
 		if ( Timestamp.class.isInstance( value ) ) {
 			final Timestamp ts = (Timestamp) value;
-			return ts.toInstant();
+			/*
+			 * Workaround for HHH-13266 (JDK-8061577).
+			 * Ideally we'd want to use ts.toInstant(), but this won't always work.
+			 * ts.toInstant() assumes the number of milliseconds since the epoch
+			 * means the same thing in Timestamp and Instant, but it doesn't, in particular before 1900.
+			 */
+			return ts.toLocalDateTime().atZone( ZoneId.systemDefault() ).toInstant();
```
yrodiere: Note this will fail with an exception for some dates... For example february 29th, 1500 exists for `java.util` (which uses the Julian calendar at that point, so 1500 is a leap year), but it does not exist for `java.time` (which uses the proleptic Gregorian calendar at that point, so 1500 is not a leap year).

As far as I can tell, this could only happen if the user bypassed Hibernate ORM to add this date in their database. But still, it's a problem.

I suppose I could add a check to turn the 29th February to the 1st March. But this feels deeply wrong.
I could also re-implement `toLocalDateTime()` and switch to another strategy before the Julian/Gregorian calendar cut-off, but the Julian/Gregorian calendar cut-off happened at a different date at different places...

Or, we could accept that such dates are not handled. I just checked, and postgresql does not accept the 29th of February, 1500 as a valid date anyway:

```
hibernate_orm_test=# select '1500-02-28T12:00:00'::timestamp;
      timestamp      
---------------------
 1500-02-28 12:00:00
(1 row)
hibernate_orm_test=# select '1500-02-29T12:00:00'::timestamp;
ERROR:  date/time field value out of range: "1500-02-29T12:00:00"
LINE 1: select '1500-02-29T12:00:00'::timestamp
               ^
```

gsmet: I suppose we will have to live with it. Let's see if we get some reports and decide then.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ZonedDateTimeJavaDescriptor.java
```diff
@@ -59,7 +59,13 @@ public ZonedDateTime fromString(String string) {
 		}
 
 		if ( Timestamp.class.isAssignableFrom( type ) ) {
-			return (X) Timestamp.from( zonedDateTime.toInstant() );
+			/*
+			 * Workaround for HHH-13266 (JDK-8061577).
+			 * Ideally we'd want to use Timestamp.from( offsetDateTime.toInstant() ), but this won't always work.
```
gbadner: The line above has a copy-paste error. it should say ```Timestamp.from( zonedDateTime.toInstant() )```

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocalDateTimeJavaDescriptor.java
```diff
@@ -55,8 +55,7 @@ public LocalDateTime fromString(String string) {
 		}
 
 		if ( java.sql.Timestamp.class.isAssignableFrom( type ) ) {
-			Instant instant = value.atZone( ZoneId.systemDefault() ).toInstant();
-			return (X) java.sql.Timestamp.from( instant );
+			return (X) Timestamp.valueOf( value );
```
gbadner: Does the JDK bug apply here? Should there be a comment about it?

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocalDateTimeJavaDescriptor.java
```diff
@@ -98,7 +97,7 @@ public LocalDateTime fromString(String string) {
 
 		if ( Timestamp.class.isInstance( value ) ) {
 			final Timestamp ts = (Timestamp) value;
-			return LocalDateTime.ofInstant( ts.toInstant(), ZoneId.systemDefault() );
+			return ts.toLocalDateTime();
```
gbadner: Does the JDK bug apply here? Should there be a comment about it?

gbadner:
Hibernate provides a couple of ways to remap a ```SqlTypeDescriptor```:

* [Dialect#remapSqlTypeDescriptor](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java#L449-L477) 
* [Dialect#getSqlTypeDescriptorOverride](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java#L479-L499)

If ```org.hibernate.type.descriptor.sql.TimeTypeDescriptor``` gets remapped to ```org.hibernate.type.descriptor.sql.TimestampTypeDescriptor```, then a ```LocalTime``` would be written to, and read from the database as a ```Timestamp```.

 ```LocalTimeJavaDescriptor#unwrap``` would convert the ```Timestamp``` to a ```ZonedDateTime``` and convert it to an instant:

`final ZonedDateTime zonedDateTime = value.atDate( LocalDate.of( 1970, 1, 1 ) ).atZone( ZoneId.systemDefault() );`

`final Instant instant = zonedDateTime.toInstant();`

Finally, it would return a ```Timestamp``` by returning ```Timestamp.from( instant )```.

When the value gets persisted by Hibernate, the date portion will be 1/1/1970, which is not affected by the JDK bug.

When the ```Timestamp``` is read from the database, ```LocalTimeJavaDescriptor#wrap``` returns  ```LocalDateTime.ofInstant( ts.toInstant(), ZoneId.systemDefault() ).toLocalTime();```.

If the ```Timestamp``` was persisted elsewhere, and is in a period that is affected the JDK bug, then ```LocalTimeJavaDescriptor#wrap``` could return an incorrect value, couldn't it?

IIUC, I think it might be a good idea to fix the ```JavaTypeDescriptor``` subclasses that can wrap stored values as a ```Timestamp```: ``LocalTimeJavaDescriptor```, ```LocalDateJavaDescriptor```, ```OffsetTimeJavaDescriptor```.

gbadner:
Also, there is a property, ```hibernate.jdbc.time_zone``` (AvailableSettings.JDBC_TIME_ZONE), that can be used to indicate the time zone used by the database. It makes most sense to set this property to use UTC to avoid problems with daylight savings time. Maybe it's overkill, but I'm wondering if the conversions will work properly with this property set.

gbadner:
I've pushed a test that overrides ```TimeTypeDescriptor``` with ```TimestampTypeDescriptor```:
https://github.com/gbadner/hibernate-core/commit/1754bbd704b83c438c0d6db393b5e11fd80c4b9d

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ZonedDateTimeJavaDescriptor.java
yrodiere: Right, will fix.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocalDateTimeJavaDescriptor.java
yrodiere: The bug applies, but I didn't add a comment because the bug-free solution is also the most obvious solution... I'll add a comment just in case.

yrodiere:
> IIUC, I think it might be a good idea to fix the ```JavaTypeDescriptor``` subclasses that can wrap stored values as a ```Timestamp```: ```LocalTimeJavaDescriptor```, ```LocalDateJavaDescriptor```, ```OffsetTimeJavaDescriptor```.

It might be. To be honest I didn't do more than what's in this PR because:

1. there are gazillions of combinations, so it would take me another week to fix them all
2. some of the conversions are obviously broken (returning an Instant when asked for a long, for example), which tends to prove at least some of them have never been used.

But fixing the timestamp conversions only might be a good compromise.

> Also, there is a property, `hibernate.jdbc.time_zone` (AvailableSettings.JDBC_TIME_ZONE), that can be used to indicate the time zone used by the database. It makes most sense to set this property to use UTC to avoid problems with daylight savings time. Maybe it's overkill, but I'm wondering if the conversions will work properly with this property set.

I'm not sure... I will try to add some tests.

yrodiere:
I added a few tests. Results:

1. `LocalDate` was indeed affected by the bug when the SQL Date type is remapped as Timestamp. I fixed it.
2. `LocalTime` and `OffsetTime` were not affected by the bug, even when their value is written to the database through JDBC with a date affected by the bug. The explanation is simple: the bug actually occurs when writing, so if Hibernate does not write, there is no bug.
3. `hibernate.jdbc.time_zone` apparently doesn't lead to more problems.

hibernate-core/src/test/java/org/hibernate/test/type/InstantTest.java
```diff
@@ -0,0 +1,144 @@
...
+/**
+ * Tests for storage of Instant properties.
+ */
+public class InstantTest extends AbstractJavaTimeTypeTest<Instant, InstantTest.EntityWithInstant> {
+
+	private static class ParametersBuilder extends AbstractParametersBuilder<ParametersBuilder> {
+		public ParametersBuilder add(int year, int month, int day,
+				int hour, int minute, int second, int nanosecond, ZoneId defaultTimeZone) {
+			if ( isNanosecondPrecisionSupported() ) {
+				nanosecond = 0;
```
gbadner: Should {{if}} condition be: {{if ( !isNanosecondPrecisionSupported() )}} ?

hibernate-core/src/test/java/org/hibernate/test/type/LocalDateTimeTest.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Tests for storage of LocalDateTime properties.
+ */
+public class LocalDateTimeTest extends AbstractJavaTimeTypeTest<LocalDateTime, LocalDateTimeTest.EntityWithLocalDateTime> {
+
+	private static class ParametersBuilder extends AbstractParametersBuilder<ParametersBuilder> {
+		public ParametersBuilder add(int year, int month, int day,
+				int hour, int minute, int second, int nanosecond, ZoneId defaultTimeZone) {
+			if ( isNanosecondPrecisionSupported() ) {
+				nanosecond = 0;
```
gbadner: Should ```if``` condition be: ```if ( !isNanosecondPrecisionSupported() )``` ?

hibernate-core/src/test/java/org/hibernate/test/type/LocalTimeTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Tests for storage of LocalTime properties.
+ */
+public class LocalTimeTest extends AbstractJavaTimeTypeTest<LocalTime, LocalTimeTest.EntityWithLocalTime> {
+
+	private static class ParametersBuilder extends AbstractParametersBuilder<ParametersBuilder> {
+		public ParametersBuilder add(int hour, int minute, int second, int nanosecond, ZoneId defaultTimeZone) {
+			if ( isNanosecondPrecisionSupported() ) {
+				nanosecond = 0;
```
gbadner: Should ```if``` condition be: ```if ( !isNanosecondPrecisionSupported() )``` ?

hibernate-core/src/test/java/org/hibernate/test/type/LocalTimeTest.java
```diff
@@ -0,0 +1,186 @@
...
+			}
+			return add( defaultTimeZone, hour, minute, second, nanosecond, 1970, 1, 1 );
+		}
+
+		public ParametersBuilder addPersistedWithoutHibernate(int yearWhenPersistedWithoutHibernate,
+				int monthWhenPersistedWithoutHibernate, int dayWhenPersistedWithoutHibernate,
+				int hour, int minute, int second, int nanosecond, ZoneId defaultTimeZone) {
+			if ( isNanosecondPrecisionSupported() ) {
+				nanosecond = 0;
```
gbadner: Should ```if``` condition be: ```if ( !isNanosecondPrecisionSupported() )``` ?

hibernate-core/src/test/java/org/hibernate/test/type/OffsetDateTimeTest.java
```diff
@@ -6,164 +6,190 @@
...
+public class OffsetDateTimeTest extends AbstractJavaTimeTypeTest<OffsetDateTime, OffsetDateTimeTest.EntityWithOffsetDateTime> {
 
-	@Override
-	protected Class[] getAnnotatedClasses() {
-		return new Class[] {OffsetDateTimeEvent.class};
+	private static class ParametersBuilder extends AbstractParametersBuilder<ParametersBuilder> {
+		public ParametersBuilder add(int year, int month, int day,
+				int hour, int minute, int second, int nanosecond, String offset, ZoneId defaultTimeZone) {
+			if ( isNanosecondPrecisionSupported() ) {
+				nanosecond = 0;
```
gbadner: Should ```if``` condition be: ```if ( !isNanosecondPrecisionSupported() )``` ?

hibernate-core/src/test/java/org/hibernate/test/type/OffsetTimeTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Tests for storage of LocalTime properties.
+ */
+public class OffsetTimeTest extends AbstractJavaTimeTypeTest<OffsetTime, OffsetTimeTest.EntityWithOffsetTime> {
+
+	private static class ParametersBuilder extends AbstractParametersBuilder<ParametersBuilder> {
+		public ParametersBuilder add(int hour, int minute, int second, int nanosecond, String offset, ZoneId defaultTimeZone) {
+			if ( isNanosecondPrecisionSupported() ) {
+				nanosecond = 0;
```
gbadner: Should ```if``` condition be: ```if ( !isNanosecondPrecisionSupported() )``` ?

hibernate-core/src/test/java/org/hibernate/test/type/OffsetTimeTest.java
```diff
@@ -0,0 +1,203 @@
...
+			}
+			return add( defaultTimeZone, hour, minute, second, nanosecond, offset, 1970, 1, 1 );
+		}
+
+		public ParametersBuilder addPersistedWithoutHibernate(int yearWhenPersistedWithoutHibernate,
+				int monthWhenPersistedWithoutHibernate, int dayWhenPersistedWithoutHibernate,
+				int hour, int minute, int second, int nanosecond, String offset, ZoneId defaultTimeZone) {
+			if ( isNanosecondPrecisionSupported() ) {
+				nanosecond = 0;
```
gbadner: Should ```if``` condition be: ```if ( !isNanosecondPrecisionSupported() )``` ?

hibernate-core/src/test/java/org/hibernate/test/type/ZonedDateTimeTest.java
```diff
@@ -6,161 +6,205 @@
...
-	protected Class[] getAnnotatedClasses() {
-		return new Class[] {ZonedDateTimeEvent.class};
+@TestForIssue(jiraKey = "HHH-10372")
+public class ZonedDateTimeTest extends AbstractJavaTimeTypeTest<ZonedDateTime, ZonedDateTimeTest.EntityWithZonedDateTime> {
+
+	private static class ParametersBuilder extends AbstractParametersBuilder<ParametersBuilder> {
+		public ParametersBuilder add(int year, int month, int day,
+				int hour, int minute, int second, int nanosecond, String zone, ZoneId defaultTimeZone) {
+			if ( isNanosecondPrecisionSupported() ) {
+				nanosecond = 0;
```
gbadner: Should ```if``` condition be: ```if ( !isNanosecondPrecisionSupported() )``` ?

gbadner:
I tried changing the condition to ```if ( !isNanosecondPrecisionSupported() )```, and the tests pass.

Other than making that change, this looks fine to me.

yrodiere:
I tested on PostgreSQL and the tests also pass. Thanks for all the reviews; merging!

PR #2801, opened 2019-02-28
HHH-13217 - Don't throw exception if both @MappedSuperclass and @Inheritance are used

vladmihalcea:
Make sure the @Inheritance annotation is ignored when used along with @MappedSuperclass

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/CoreMessageLogger.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/CoreMessageLogger.java
@@ -1864,6 +1864,6 @@ public interface CoreMessageLogger extends BasicLogger {
 	void ignoreImmutablePropertyModification(String propertyName, String entityName);
 
 	@LogMessage(level = WARN)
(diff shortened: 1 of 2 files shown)


vladmihalcea:
@gbadner This is the test case you mentioned in the [Jira issue comment](https://hibernate.atlassian.net/browse/HHH-13217).

Now, we really need to merge HHH-12653 and HHH-13217 since it makes no sense to have them both since the original behavior of HHH-12653 was reverted.

gbadner:
@vladmihalcea, I tweaked the warning message and test. If it looks OK to you, then it's fine with me.

The commit messages for HHH-12653 and HHH-13217 already refer to their respective issues, so I don't see how we can merge the issues. 

Also, you said HHH-12653 was fixed in 5.4.0.CR1. HHH-12653 is still open. It needs to have the fix version added, and it needs to be added to changelog.txt under "Changes in 5.4.0.CR1".

vladmihalcea:
@gbadner your commit looks good to me.

PR #2802, opened 2019-03-05
HHH-13300 query.getSingleResult() throws org.hibernate.NonUniqueResultException instead of javax.persistence.NonUniqueResultException

yrodiere:
https://hibernate.atlassian.net//browse/HHH-13300


gbadner:
In addition, the following do not convert the exception if the SessionFactory was JPA bootstrapped.

- ```AbstractProducedQuery.java`#executeUpdate``` (should be changed)
- ```AbstractSharedSessionContract#close``` (not sure about this one)
- ```SessionImpl#fireRefresh``` (should be changed so it is consistent with ```SessionImpl#merge```)

I'm not sure about ```AbstractSharedSessionContract#close```, because it does not implement ```EntityManager#close```.

I'm still thinking about this...

yrodiere:
> I'm not sure about `AbstractSharedSessionContract#close`, because it does not implement `EntityManager#close`.

I would argue that `AbstractProducedQuery#list` is not related to JPA either, yet it converts exceptions to JPA depending on the bootstrap and settings.

Please let me know of your decision :)

gbadner:
I agree ```AbstractProducedQuery#list``` looks inconsistent.

gbadner:
I pushed a test and fix for ```Query#executeUpdate```. The others can be dealt with later.

@yrodiere, if my commits look good to you, then I'm fine with merging this. Thanks for finding this bug.

yrodiere:
It looks good! Merged, thanks.

PR #2803, opened 2019-03-05
HHH-13285 : ClassCastException: org.dom4j.DocumentFactory cannot be cast to org.dom4j.DocumentFactory after dom4j update (5.1)

gbadner:
https://hibernate.atlassian.net/browse/HHH-13285 (5.1)


gbadner:
5.1 is not longer maintained.

PR #2805, opened 2019-03-06
HHH-7525 For @Formula, use property name when there is no column name.

r-murphy:
Also removes @FailureExpected from HHH-2225.

There's a number of ways to solve this issue. I went with a simple approach in the `ColumnEntityAliases` method which was returning `[null]` for properties with `@Formula` instead of `@Column`. I also considered checking the result of `getPropertyAliases` where it's used in `DefaultEntityAliases` but decided the issue was specific to the use of `ColumnEntityAliases` so I applied the fix there.

Other solutions should involve `instanceof` checks and casts on the perister to get more details on that property (i.e. checking if it's a formula field) that aren't available to `Loader` interface in this method, or even potentially a new subclass of `ColumnEntityAliases` and then change the usage in `SQLQueryReturnProcessor`. But I wanted to avoid extra complexity since this was my first foray into the hibernate source.


hibernate-core/src/main/java/org/hibernate/loader/ColumnEntityAliases.java
```diff
@@ -33,8 +33,12 @@ public ColumnEntityAliases(
 	protected String getDiscriminatorAlias(Loadable persister, String suffix) {
 		return persister.getDiscriminatorColumnName();
 	}
-	
+
 	protected String[] getPropertyAliases(Loadable persister, int j) {
-		return persister.getPropertyColumnNames(j);
+		String[] propertyColumnNames = persister.getPropertyColumnNames(j);
+		if (propertyColumnNames!=null && propertyColumnNames.length==1 && propertyColumnNames[0]==null) {
```
NathanQingyangXu: It would be great to use space chars consistently with existing coding format as
`
if ( propertyColumnNames != null && propertyColumnNames.length == 1 && propertyColumnNames[0] == null )
`

NathanQingyangXu: Not sure whether using `Arrays.equals()` might improve the code readability as
`
if ( Arrays.equals( propertyColumnNames, new String[] { null } ) {
    ... ...
}
`

hibernate-core/src/test/java/org/hibernate/test/annotations/formula/FormulaNativeQueryTest.java
```diff
@@ -68,11 +70,42 @@ public void cleanup() {
...
+	@Test
+	public void testNativeQueryWithAliasSyntax() throws Exception {
+		doInHibernate(
+				this::sessionFactory, session -> {
+					NativeQuery query = session.createNativeQuery(
+							"SELECT ft.id as {ft.id}, ft.locationStart as {ft.locationStart}, ft.locationEnd as {ft.locationEnd}, abs(ft.locationEnd - locationStart) as {ft.distance} FROM foo_table ft")
+							.addEntity( "ft", Foo.class );
+					Map<String,String> props = new HashMap<>();
+					props.put( "distance", "distance" );
+					query.setProperties( props );
```
NathanQingyangXu: You might well simplify a little bit as
`query.setProperties( Collections.singletonMap( "distance", "distance" ) )`

hibernate-core/src/main/java/org/hibernate/loader/ColumnEntityAliases.java
r-murphy: Inconsistent spaces fixed and switched to Arrays.equals as suggested.

hibernate-core/src/test/java/org/hibernate/test/annotations/formula/FormulaNativeQueryTest.java
r-murphy: Updated with the recommendation to simplify.

dreab8:
Hi @r-murphy,

may be I'm missing something but the code `session.createNativeQuery( "SELECT ft.* FROM foo_table ft", Foo.class );` that was the original query executed by the test `testNativeQuery()` marked with `@FailureExpected( jiraKey = "HHH-7525" )` is still failing(I tried to reintroduce it) , the test is not failing anymore in your PR because it executes a different query.

r-murphy:
@dreab8 hmm, you are correct. I wrote the original change almost a year ago now so my memory on it is fuzzy (I squashed and force pushed so it looks new).

I guess you could say my change is only a partial solution. It allows for `@Formula` and native query to be used together, provided that the native query returns values for all properties of the class (for example the modified query I used to return a distance). This would have been an acceptable workaround for my prior use case, and I think still better than not being able to use `@Formula` and native query together at all.

I can add back the original query, keeping the `@FailureExpected`.

I'm not sure what the expected result would be for the original query `SELECT ft.* FROM foo_table ft`. One option is the property with `@Formula` would just be left un-set. The other option would be to try and add the query fragment from the `@Formula` into the native query.  But then it's not really native anymore and probably has a number of edge cases and complexities. So probably the option to just leave it un-set?

dreab8:
@r-murphy not sure about the best way to deal with that case, I'm leaning towards adding the formula when the column is not specified in the query but probably not supporting that case is an acceptable solution as well. @jwgmeligmeyling, @gbadner what is your opinion?

jwgmeligmeyling:
As far as I remember the same limitation applies for almost any feature that would require manipulation of the SQL statement, i.e. secondary tables, joined table hierarchies, associations mapped by a join table, discriminator formulas.

Native queries get executed as-is, with the exception of some alias substitution. Making all these features available in native queries will be a though job, because we would have to transform the produced SQL. There is probably not a lot of reason to do so: at that point its usually easier to make the required portion of SQL available in JPQL through a custom function template.

In this case it seems we could rewrite `ft.*` to `ft.*, abs(ft.locationEnd - ft.locationStart)` with relative ease, as we know that `ft` references `foo_table`. It would perhaps be nice to support this. However the implementation will be troubled by dialect variations and might not support all types of Formulas. For one, I'm not sure if all properties will always be on `ft`, they can for example also be database constants/functions, such as `CURRENT_TIMESTAMP`.

With that regard, I think a partial fix can be integrated although we should document that the `SELECT *` case is not supported still (and while at it just keep the original test case).

r-murphy:
I've added the original test case back with FailureExpected.
The code analysis failed, but it's failing to resolve dependencies which is not related to my change. Is there a way to retrigger analysis without a new push?

jluyten:
I seem to be running into the same issue. @Formula column resulting in a column 'null' resulting in a nullpointer. Any news on merging this in?

beikov:
I have fixed this for 6.0 on a PR: https://github.com/hibernate/hibernate-orm/pull/4129

We will get back to PRs like this after we finish the final release of 6.0, so please be patient.

hibernate-core/src/main/java/org/hibernate/loader/ColumnEntityAliases.java
```diff
@@ -35,6 +36,10 @@ protected String getDiscriminatorAlias(Loadable persister, String suffix) {
 	}
 	
 	protected String[] getPropertyAliases(Loadable persister, int j) {
-		return persister.getPropertyColumnNames(j);
+		String[] propertyColumnNames = persister.getPropertyColumnNames(j);
+		if ( Arrays.equals( propertyColumnNames, new String[] { null } ) ) {
```
beikov, line 40:
```suggestion
		if ( propertyColumnNames.length == 1 && propertyColumnNames[0] == null ) {
```

PR #2807, opened 2019-03-10
JMH test skeleton for hibernate orm

carstenartur:
Just found that this pull request is violating the "Contributing guidelines" because of a number of reasons. Are you interested in a automatic performance measurement (microbenchmarking) like this sample shows for hibernate? I think for hibernate validation there is something like this part of it but not for hibernate orm as far as I know...
At the moment just would like to know your opinion regarding this question.


sebersole:
In fact we are using JMH a little with ORM 6 development.  The "problem" is that a performance number by itself is not very interesting.  It is only interesting in combination with earlier runs or compared against older versions.

Would definitely be interested though.

carstenartur:
The comparison against earlier builds is automatically part of the diagrams. Look at http://jmh.morethan.io/ and choose compare. If you collect builds for a longer time you can compare more of them over a longer period.
HHH-13311

sebersole:
Oh... I missed the part that this is a service.  With what we are doing for 6 we just used the jmh jar and ran tests locally.

carstenartur:
Steve, is https://github.com/hibernate/hibernate-orm/pull/2808 something interesting or should I close it to get out of the way?

PR #2808, opened 2019-03-11
Hhh 13311

carstenartur:
Not sure if there is a way to change https://github.com/hibernate/hibernate-orm/pull/2807 into following the convention with the separate branch. Because of this I created another pull request. Hope that is ok for you. The other one can be closed now.
see https://hibernate.atlassian.net/browse/HHH-13311


sebersole:
@carstenartur This is definitely something I'd like to incorporate.  I think @dreab8 and @Sanne will be interested as well.

I'll have to migrate this from Maven to Gradle before I can apply it upstream.

carstenartur:
> @carstenartur This is definitely something I'd like to incorporate. I think @dreab8 and @Sanne will be interested as well.
> 
> I'll have to migrate this from Maven to Gradle before I can apply it upstream.

Why not use it as separate maven based subproject until you find time for gradle migration?

Sanne:
hi @carstenartur , we've been using JMH since a while for "one off" tests targeted tests - indeed sometimes using Maven. The next step, higher value would be to have them run periodically and automatically, hence integrated in the gradle build.

I actually suspect @sebersole and @dreab8 already adapted your suggestion, there's various similar tests in the ORM6 branch. So maybe this was done already - question for @sebersole and @dreab8 I guess we can close this now, or is there more you'd like to incorporate?

carstenartur:
No, then we can close it. where are the graphical diagrams for the performance comparison of the builds visible that fall out of ci? Where can I find the reports to compare the numbers between the releases?

Sanne:
>  where are the graphical diagrams for the performance comparison of the builds visible that fall out of ci?

No that's what is missing and why it would be cool to have this integrated in the gradle build. As mentioned, today we run benchmarks manually only. We definitely want to improve on that.

Another complexity is that JMH is good for testins a very specific angle (a micro-benchmark), so while it would be interesting to have many tests based on JMH, these won't be giving a single "summary" score; for that we need a full system benchmark; our team regularly refers to SpecJ Enterprise for a full-system sanity check but it's not open and there's many difficulties in getting approval to publish its results.

carstenartur:
This pull request is for exactly providing that. You can just create a separate jenkins build job and there you are..

PR #2809, opened 2019-03-11
HHH-13309 Extended bean managers implementing the new interface are not correctly detected

yrodiere:
https://hibernate.atlassian.net//browse/HHH-13309


yrodiere:
Merged, thanks

PR #2812, opened 2019-03-13
HHH-13241 HHH-13138 (5.3)

gbadner:
https://hibernate.atlassian.net/browse/HHH-13241
https://hibernate.atlassian.net/browse/HHH-13138

The fix for HHH-13138 is included here because it is required in order for all the HHH-13241 tests to pass.


gbadner:
Commits have been added to test/fix regression deleting an entity with a lazy null many-to-one association.

PR #2814, opened 2019-03-14
[HHH-13319]setParameter doesn't work in native Query if Query is already executed with different parameter value for collection type when new collection length is less than the previous	

git-rbanerjee:
When re using a native query with a Collection type param , its not removing the already expanded parameter from parameterBindingMap in expandListValuedParameters. For example first time the Collection had 3 values so it expanded to ids_0 , ids_1, ids_2 . Then next run it has two values so its just updating ids_0 , ids_1 not removing ids_2 from parameterBindingMap which is causing Caused by: org.hibernate.QueryException: Named parameter does not appear in Query: ids_2


gavinking:
The test passes in H7.

PR #2818, opened 2019-03-20
HHH-13326 : Transaction passed to Hibernate Interceptor methods is null when JTA is used

gbadner:
https://hibernate.atlassian.net/browse/HHH-13326

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/interceptor/InterceptorNonNullTransactionTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/interceptor/InterceptorNonNullTransactionTest.java
@@ -1,0 +1,256 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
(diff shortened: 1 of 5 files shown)


hibernate-core/src/test/java/org/hibernate/test/interceptor/InterceptorNonNullTransactionTest.java
```diff
@@ -0,0 +1,293 @@
...
+			@Override
+			public boolean allowTransactionAccess(
+					SessionFactory sessionFactory,
+					JpaComplianceTransactionSetting jpaComplianceTransactionSetting) {
+				// setting is ignored if jpaComplianceTransactionSetting != JpaComplianceTransactionSetting.TRUE
+				return jpaComplianceTransactionSetting != JpaComplianceTransactionSetting.TRUE;
+			}
+		};
+
+		public abstract boolean allowTransactionAccess(
```
yrodiere: As far as I can tell, this method and its implementations are dead code. Bad copy/paste, maybe?

gsmet: I force-pushed an update to fix that. Will wait for CI and merge.

PR #2820, opened 2019-03-20
HHH-13324 Install the security manager check layer once and for all

gsmet:
If the security manager is defined later, the checks won't be installed.

The security manager should be defined when starting the JVM.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/ByteBuddyState.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/ByteBuddyState.java
@@ -57,10 +57,9 @@ public final class ByteBuddyState {
 
 	private final ByteBuddy byteBuddy;
 
-	private final ForDeclaredMethods getDeclaredMethodMemberSubstitution;
-	private final ForDeclaredMethods getMethodMemberSubstitution;
-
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/ByteBuddyState.java
```diff
@@ -211,36 +205,11 @@ void clearState() {
 			}
 		}
 
-		if ( System.getSecurityManager() != null ) {
-			// we authorize the proxy class to access the method lookup dispatcher
-			HibernateMethodLookupDispatcher.registerAuthorizedClass( unloadedClass.getTypeDescription().getName() );
-		}
+		classRewriter.registerAuthorizedClass( unloadedClass.getTypeDescription().getName() );
```
yrodiere: Seems you'll be calling `,getTypeDescription().getName()` even when the security manager is not enabled, while previously you didn't.
I don't know if it matters (performance?), but I thought I'd warn you.

gsmet: That's a good remark indeed. I updated my PR, could you check?

yrodiere: Looks good to me if that's all you changed.

PR #2824, opened 2019-03-28
HHH-9861 Fix wrong mapping of ColumnType FLOAT for H2 database

hTnevS:
H2 has no float datatype. Instead, float is an alias for double. This leads to an error when hibernate is trying to validate the schema (found: double, expected: float).

Source for H2 datatypes:
http://www.h2database.com/html/datatypes.html


hTnevS:
H2 has no float datatype. Instead, float is an alias for double. This leads to an error when hibernate is trying to validate the schema (found: double, expected: float).

Source for H2 datatypes:
http://www.h2database.com/html/datatypes.html

hTnevS:
Build currently failing due to HHH-13357:
https://hibernate.atlassian.net/browse/HHH-13357

PR #2826, opened 2019-04-05
HHH-13343 (5.3) Bytecode enhancement using ByteBuddy fails when the class is not available from the provided ClassLoader

gbadner:
https://hibernate.atlassian.net/browse/HHH-13343


gbadner:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2837

PR #2827, opened 2019-04-12
HHH-13364 Query.getSingleResult and getResultList() throw PessimisticLockException when pessimistic lock fails with timeout

gbadner:
https://hibernate.atlassian.net/browse/HHH-13364


dreab8:
it looks go to me

PR #2828, opened 2019-04-12
Remove final form EntityAction subclasses

DavideD:
They are declared as final making it impossible to extends them and create custom actions.
The ActionQueue only accept them as actions, making it impossible to add custom actions to the queue.

Probably there are others to update.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/action/internal/EntityInsertAction.java
+++ b/hibernate-core/src/main/java/org/hibernate/action/internal/EntityInsertAction.java
@@ -31,7 +31,7 @@ import org.hibernate.metamodel.model.domain.spi.EntityTypeDescriptor;
  *
  * @see EntityIdentityInsertAction
  */
-public final class EntityInsertAction extends AbstractEntityInsertAction {
+public class EntityInsertAction extends AbstractEntityInsertAction {
 	private Object version;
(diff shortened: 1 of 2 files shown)


dreab8:
It seems fine to me, @sebersole is there any reason I'm missing to keep `EntityUpdateAction` and `EntityInsertAction`  final?

sebersole:
Well the simple truth is that custom actions are not supported as evidenced
by the fact that these are final :D

We could remove the final, but that does not change whether custom actions
are supported or not (and if they are not supported I'd argue we should
keep the final).

Put another way, supporting custom actions is about more than just removing
this final.  What are the semantics of these custom actions?  In what uses
are they supported?  How do they work wth auto-flush?  We'd have to sketch
out a lot of things here.

Out of curiosity Davide, can you not just use the before/after transaction
hooks as part of the ActionQueue for your needs?  That is intended and
designed to accept custom "actions".

DavideD:
> Out of curiosity Davide, can you not just use the before/after transaction
> hooks as part of the ActionQueue for your needs?  That is intended and
> designed to accept custom "actions".

Ah, maybe aI can make it work if pre-post insert events. I haven't thought about it. I will check.

DavideD:
Ok, I've found an alternative approach. Thanks

PR #2830, opened 2019-04-14
support for ANSI SQL-style DATE/TIME/TIMESTAMP literals in HQL

gavinking:
This change adds support for literals of form:

    DATE 'YYYY-MM-DD'
    TIME 'HH:MM:SS'
    TIMESTAMP 'YYYY-MM-DD HH:MM:SS'
    TIMESTAMP 'YYYY-MM-DD HH:MM:SS +HH:MM'
    TIMESTAMP 'YYYY-MM-DD HH:MM:SS XXX/Xxxxx'

To the query language. This is the syntax defined by ANSI SQL, and supported by most databases.

Currently, we support the syntax defined by the JPA spec, which looks like the JDBC escape syntax.

    {d 'YYYY-MM-DD'}
    {t 'HH:MM:SS'}
    {ts 'YYYY-MM-DD HH:MM:SS'}

This is very ugly, and does not support offsets/timezones in timestamps.

Original issue here:

<https://hibernate.atlassian.net/browse/HQLPARSER-85>


sebersole:
We'll agree to disagree about it being ugly, but I'm definitely +1 for
supporting this in addition to...  Thanks!

sebersole:
The reason I like the originals btw (aside from JPA requiring them) is that it makes it easier to plug in support for arbitrary literal types.  As an example, `hibernate-spatial` could plug in "parser" for spatial literals.  Not yet implemented, just a plan to come back to.

gavinking:
> it makes it easier to plug in support for arbitrary literal types

That's a good idea.

I'm not arguing that the ODBC/JDBC-style syntax is especially bad, just that the ANSI SQL style is much more consistent with the rest of the language.

gavinking:
I've merged in the changes from #2862, to resolve the conflicts.

gavinking:
I've merged these changes into the big date/time PR #2863. Closing here.

PR #2832, opened 2019-04-15
support for EXTRACT function in HQL

gavinking:
This change adds support for the ANSI-SQL syntax:

    EXTRACT(<field> FROM <expression>)

to HQL.

And fixes something that was almost certainly a bug in the implementation of  `AbstractParameter` (see 428e45f).


gavinking:
Note that I have gone with the grammar that was already there in Hibernate 6. However, I'm inclined to think that the field should just be an identifier, instead of making `YEAR`, `MONTH`, `DAY`, `HOUR`, `MINUTE`, `SECOND`, `TIMEZONE_HOUR`, and `TIMEZONE_MINUTE` keywords, because:

1. That's a *lot* of keywords for one function!
2. Most databases support a bunch of additional fields, for example `MILLISECONDS`, and I don't see why we should prevent you from being able to use those.

If you like, I can change the grammar to make this an arbitrary identifier.

sebersole:
Regarding identifier-versus-keyword...

1) Keyword-as-identifier is not as big of a deal as it was in the past because of how the grammar is structured.  Not sure if that is the basis of your point, but just something to consider.
2) Considering you normalize them all to a single node type (`SqmExtractUnit`), the keyword-versus-identifier distinction is not really important - you could definitely change that to simply use identifier.  Use `identifier` though rather than `IDENTIFIER` if you chose to do that.

I'm not sure I like moving from `SqlExpression` to `SqlAstNode` as the things passed to functions.  It would definitely eliminate a kludge I had to do for a similar case wrt calling out to registered functions to handle `TRIM` calls and passing along the `LEADING`, `TRAILING`, `BOTH` indicator.  So I definitely like it in that sense.  I'll have to think about that part.  @dreab8 - wdyt?

Thanks

gavinking:
> Keyword-as-identifier is not as big of a deal as it was in the past because of how the grammar is structured. Not sure if that is the basis of your point, but just something to consider.

Yes, I understand that these are "soft" keywords that don't hammer the space of identifiers, but there's still no need for them to be listed explicitly in the grammar, since they don't drive any parsing rules.

But this is really a bigger issue that we need to discuss somewhere else: it's perfectly fine that the JPA spec grammar treats things like function names as keywords, since we're not trying to compile a parser from it, but I think it's wrong to take that too literally and actually treat them as keywords in the ANTLR grammar.

People often think they should make a grammar for a parser compiler as strict as possible, detecting all sorts of errors as early as possible as *syntax* errors. But that's not quite right. Really you want to make the grammar as tolerant as possible, and catch errors in things like function names later on on the process, where you can produce a much more meaningful error. Parsers, by nature, produce pretty terribly error messages: 

> unexpected token: 'seconds' 

instead of 

> unknown field as argument to extract: 'seconds' (did you mean 'second'?)

So the only keywords the ANTLR grammar should declare are things it absolutely needs in order to recognize language constructs and drive parser rules. That doesn't include function names.

But anyway, we should discuss this elsewhere.


> Considering you normalize them all to a single node type (`SqmExtractUnit`), the keyword-versus-identifier distinction is not really important - you could definitely change that to simply use identifier. Use `identifier` though rather than `IDENTIFIER` if you chose to do that.

Yes, I realize this.

gavinking:
> I'm not sure I like moving from `SqlExpression` to `SqlAstNode` as the things passed to functions. It would definitely eliminate a kludge I had to do for a similar case wrt calling out to registered functions to handle `TRIM` calls and passing along the `LEADING`, `TRAILING`, `BOTH` indicator. So I definitely like it in that sense.

Well it's not only the `extract()` and `trim()` functions that have arguments which aren't expressions, but also `cast()` (which I'm going to work on next).

But, sure, I could introduce a `SqlFunctionArgument` interface which is a supertype of `SqlExpression` and a couple of other things.

sebersole:
That's a pretty condescending comment.  I've worked with grammars for quite
some time (about as long as you actually).

If you have specific suggestions let's discuss those. But generalized
statements about how I don't understand how to write grammars.. no

gavinking:
Superseded by merged pull request #2858. Closing.

PR #2834, opened 2019-04-16
HHH-11223 (5.3) Initialize bytecode enhanced lazy to-one using a single SQL statement

gbadner:
https://hibernate.atlassian.net/browse/HHH-11223


gbadner:
Closing because HHH-11147 is preferred over HHH-11223.

PR #2835, opened 2019-04-16
HHH-13357 OffsetTimeTest fails using TimeAsTimestampRemappingH2Dialect in non-GMT European time zones

yrodiere:
https://hibernate.atlassian.net//browse/HHH-13357

It turns out the problem was how we pick the offset. We used to set the offset that corresponds to the default JVM timezone *at the time of the timestamp*, while we should have set the offset that corresponds to the default JVM timezone *at the current time*, like we do when converting a Time value.

More details in comments in the code and in the commit message.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/OffsetTimeJavaDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/OffsetTimeJavaDescriptor.java
@@ -12,7 +12,7 @@ import java.time.Instant;
 import java.time.LocalDate;
 import java.time.OffsetDateTime;
 import java.time.OffsetTime;
-import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.Calendar;
 import java.util.Date;
@@ -65,6 +65,12 @@ public class OffsetTimeJavaDescriptor extends AbstractTypeDescriptor<OffsetTime>
 		final ZonedDateTime zonedDateTime = offsetTime.atDate( LocalDate.of( 1970, 1, 1 ) ).toZonedDateTime();
 
 		if ( Timestamp.class.isAssignableFrom( type ) ) {
+			/*
+			 * Workaround for HHH-13266 (JDK-8061577).
+			 * Ideally we'd want to use Timestamp.from( offsetDateTime.toInstant() ), but this won't always work.
+			 * Timestamp.from() assumes the number of milliseconds since the epoch
+			 * means the same thing in Timestamp and Instant, but it doesn't, in particular before 1900.
+			 */
 			return (X) Timestamp.valueOf( zonedDateTime.toLocalDateTime() );
 		}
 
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/type/descriptor/java/OffsetTimeJavaDescriptor.java
```diff
@@ -95,22 +101,44 @@ public OffsetTime fromString(String string) {
...
+		 * since a zone has a varying offset over time,
+		 * thus the zone might have a different offset for the given timezone than it has for the current date/time.
+		 * For example, if the timestamp represents 1970-01-01TXX:YY,
+		 * and the JVM is set to use Europe/Paris as a timezone, and the current time is 2019-04-16-08:53,
+		 * then applying the JVM timezone to the timestamp would result in the offset +01:00,
+		 * but applying the JVM offset would result in the offset +02:00, since DST is in effect at 2019-04-16-08:53.
+		 *
+		 * Of course none of this would be a problem if we just stored the offset in the database,
+		 * but I guess there are historical reasons that explain why we don't.
+		 */
```
gbadner: If the timestamp, 1970-01-01T10:00, is read on 1/1/19 in Paris, then the OffsetTime would represent 11:00 because 1/1/19 would be during Standard Time with an offset of +1.

If the same timestamp is read on 4/16/19 in Paris, then the OffsetTime would represent 12:00 because 4/16/19 would be during Daylight Savings Time with an offset of +2.

It doesn't seem right that a different time would be read, depending on whether DST is in effect. Am I missing something here?

I agree that storing the offset would resolve the ambiguity.

@gsmet, what do you think?

gbadner: The Javadoc for ```java.sql.Time``` says:

"The date components should be set to the "zero epoch" value of January 1, 1970 and should not be accessed."

It makes sense that Hibernate gets the offset for a ```Time``` value using ```OffsetDateTime.now().getOffset()```, because that date for a ```Time``` value should be ignored.

```Timestamp``` is different, because the date is available, and it should be possible to determine the time on a particular date with the correct offset.

I'm not an expert on this by any means, so I could be missing something here.

yrodiere: > Timestamp is different, because the date is available, and it should be possible to determine the time on a particular date with the correct offset.

It's not, though. The date is available but in some (most?) cases is completely artificial and meaningless: we force it to `1970-01-01` when we write.

This is what used to cause the issue: we used to infer the correct offset from the timestamp, which means we used an offset that was not the same as the one we used when writing, and we ended up reading something different than what we wrote.

Anyway... let me rephrase the problem:

* In the database, we have a time and (in the case of timestamps) a date.
* In the entity, we have a time and an offset.
* We need to convert between those two.

Obviously we cannot preserve the offset, since it's not stored. This means we just cannot store an `OffsetDateTime` correctly and any solution will be inherently flawed.

With that in mind, we can chose to either:

1. Store the local time (i.e. `offsetTime.toLocalTime()`) and add an arbitrary offset when reading (e.g. the current offset of the default JVM timezone).
2. Store the UTC time (i.e. `offsetTime.atOffset(ZoneOffset.UTC).toLocalTime()`) and present it with an arbitrary offset when reading (e.g. the current offset of the default JVM timezone).

With solution 1:

* Storing `10:00:00+00:00` (`10:00:00 UTC`) and then reading it with JVM offset = `+01:00` would result in reading `10:00:00+01:00` (`09:00:00 UTC`)
* Storing `10:00:00+00:00` (`10:00:00 UTC`) and then reading it with JVM offset = `-01:00` would result in reading `10:00:00-01:00` (`11:00:00 UTC`)
* Storing `10:00:00+01:00` (`09:00:00 UTC`) and then reading it with JVM offset = `+01:00` would result in reading `10:00:00+01:00` (`09:00:00 UTC`)
* Storing `10:00:00+01:00` (`09:00:00 UTC`) and then reading it with JVM offset = `-01:00` would result in reading `10:00:00-01:00` (`11:00:00 UTC`)

With solution 2:

* Storing `10:00:00+00:00` (`10:00:00 UTC`) and then reading it with JVM offset = `+01:00` would result in reading `11:00:00+01:00` (`10:00:00 UTC`)
* Storing `10:00:00+00:00` (`10:00:00 UTC`) and then reading it with JVM offset = `-01:00` would result in reading `09:00:00-01:00` (`10:00:00 UTC`)
* Storing `10:00:00+01:00` (`09:00:00 UTC`) and then reading it with JVM offset = `+01:00` would result in reading `10:00:00+01:00` (`09:00:00 UTC`)
* Storing `10:00:00+01:00` (`09:00:00 UTC`) and then reading it with JVM offset = `-01:00` would result in reading `08:00:00-01:00` (`09:00:00 UTC`)

Solution 2 seems more correct, indeed. That being said, it's not what we used to implement:

* When storing as `Time`, we used to implement solution 1.
* When storing as `Timestamp`, we used to implement solution 1 too, but with a twist: the offset was determined not from the current JVM date, but from the stored date. Which triggered the bug we can now see.

So yes, my fix leads to reading a different local time depending on whether DST is in effect or not. But that's always been the case, and it's worse than that: we read a different local time depending on the JVM timezone.

We could switch to solution 2, but that would require storing different information, and we would lose backward compatibility. I'm not sure it's a good idea to address this in this PR.

gbadner: Thanks for this explanation. I had completely missed that we force it to 1970-01-01 when writing the value. I understand your reasoning now.

I'm starting to think Hibernate should at least issue a warning, maybe even an exception, since Hibernate changes the date in the timestamp.

I'm not sure how best to deal with this. I think it's really a corner case, and after thinking about it more, maybe we just shouldn't allow it.

I think we should discuss on the hibernate-dev mailing list.

yrodiere: We can discuss it further on the mailing list for sure, but I'd be inclined to merge the fix now, since it only impacts corner cases and makes the code consistent with what we do in every other case.
This would at least fix the build.

sebersole:
Is there any reason to not merge this?

gbadner:
I've pushed a commit to master to temporarily skip the failing tests. We need to discuss this on the mailing list.

xstefank:
Is there any progress on this, please?

yrodiere:
As far as I'm concerned we can merge it. I think I explained the reasons behind these changes, and why it won't solve everything.
@gbadner , if you still have concerns, maybe you could start a thread on the mailing list?

gbadner:
I've reverted the commit that skipped the failing tests (352b029404877822468738e693136e9a5ced6342), rebased, and merged this fix. Thanks @yrodiere, and sorry for the delay!

romainmoreau:
Hello,

Did you notice this [issue](https://hibernate.atlassian.net/browse/HHH-13379) ?

It's related to this one and I think it would make sense to fix both of them in the same version.

PR #2836, opened 2019-04-16
HHH-13152 - managed/cached entities and fetching (proposed fix)

sebersole:
Proposed fix for HHH-13152


hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
```diff
@@ -430,7 +432,19 @@ private Object doLoad(
...
 
 		if ( entity != null ) {
-			return persistenceContextEntry.isManaged() ? entity : null;
+			if ( persistenceContextEntry.isManaged() ) {
+				final LoadQueryInfluencers influencers = event.getSession().getLoadQueryInfluencers();
+				final boolean hasActiveInfluencers = influencers.getEffectiveEntityGraph().getSemantic() != null
+						|| influencers.hasEnabledFetchProfiles();
+				if ( hasActiveInfluencers ) {
+					// we may need to force initialization of some of the entity's attributes
+					persister.initializeLazyProperties( entity, event.getSession() );
```
dreab8: not sure but probably we can pass the `RootGraphImplementor<?>` or the `EffectiveEntityGraph` to the `initializeLazyProperties`, this avoids to call again `session.getLoadQueryInfluencers();` and
`loadQueryInfluencers.getEffectiveEntityGraph()`

sebersole:
Thanks Andrea.

One thing that occurred to me is that I also should be setting the
AttributeNode's graph (if one) as the effective graph for the corresponding
attribute initialization.

sebersole:
Just pushed some changes to apply sub-graphs to the process.  

At the moment it only works for to-one associations within our limited sub-graph support.

PR #2841, opened 2019-04-18
HHH-13374: workaround for JDK bug

coladict:
https://hibernate.atlassian.net/browse/HHH-13374
Complete with testcase. Forced update to fix formatting.


coladict:
It just hit me that once the bug gets fixed, then this will invert it. I know how to fix it. Will do it tomorrow on my workstation and rebase after HHH-13357 is fixed by #2835 or something, so all the tests will pass.

Sanne:
Many thanks @coladict ! Do you know if the JDK bug got assigned an identifier? Would love to track this.

coladict:
I reported it months ago, but never got confirmation of it being assigned. Another bug I reported after it did get a number, and is already fixed in JDK-14.

beikov:
I'll close this as out of date since this appears to be fixed in newer JDK versions already.

PR #2844, opened 2019-04-24
HHH-14393 Respect precision and scale for type DOUBLE

quaff:
```
	@Column(precision = 8, scale = 6)
	private Double latitude;
```
hibernate should create column latitude with double(8,6), but only double.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
@@ -96,7 +96,7 @@ public class MySQLDialect extends Dialect {
 		registerColumnType( Types.INTEGER, "integer" );
 		registerColumnType( Types.CHAR, "char(1)" );
 		registerColumnType( Types.FLOAT, "float" );
-		registerColumnType( Types.DOUBLE, "double precision" );
+		registerColumnType( Types.DOUBLE, "double precision($p,$s)" );
 		registerColumnType( Types.BOOLEAN, "bit" ); // HHH-6935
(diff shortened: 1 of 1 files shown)


quaff:
Reopen it

quaff:
@gbadner Would you check it?

beikov:
We did this for 6.0 already and hit some issues due to the JPA column annotation defaults resulting in `double(0,0)` being used. Not sure how safe it is to do this in 5.4

gavinking:
On what database is precision/scale meaningful for the `double precision` type?

(Agree that we should not do this in H5.)

quaff:
I'm closing it since it's not an obvious fix.

quaff:
> On what database is precision/scale meaningful for the `double precision` type?
> 
> (Agree that we should not do this in H5.)

FYI: https://dev.mysql.com/doc/refman/8.0/en/floating-point-types.html
> MySQL permits a nonstandard syntax: FLOAT(M,D) or REAL(M,D) or DOUBLE PRECISION(M,D). Here, (M,D) means than values can be stored with up to M digits in total, of which D digits may be after the decimal point. For example, a column defined as FLOAT(7,4) is displayed as -999.9999. MySQL performs rounding when storing values, so if you insert 999.00009 into a FLOAT(7,4) column, the approximate result is 999.0001.

gavinking:
OK, thanks @quaff. But the next paragraph is:

> As of MySQL 8.0.17, the nonstandard `FLOAT(M,D)` and `DOUBLE(M,D)` syntax is deprecated and you should expect support for it to be removed in a future version of MySQL.

So I don't think we should add support for this to Hibernate.

PR #2845, opened 2019-04-24
HHH-13378 hibernate-jipijapa

peter-griffiths:
- Updated HibernateStatistics to update fields inherited from HibernateAbstractStatistics directly rather than using the getters that sometimes return an unmodifiable collection, changing the fields' access to protected in the process.

- Added a test case covering the instantiation of HibernateStatistics that fails without the fix.


Sanne:
hi @scottmarlow , looks like we received this patch in the wrong repository. Will you need this patch in WildFly?

scottmarlow:
> hi @scottmarlow , looks like we received this patch in the wrong repository. Will you need this patch in WildFly?

Hi @Sanne , from the https://hibernate.atlassian.net/browse/HHH-13378 description it sounds like no (looks like the change is or was already in WildFly): `The WildFly code this module is based on is not affected by the same issue.`

Sanne:
ok, thanks for checking @scottmarlow

beikov:
AFAIU the jipijapa module is part of Wildfly (again?), so there is nothing we can do on the Hibernate side.

PR #2846, opened 2019-04-24
Hhh 11147

sebersole:
https://hibernate.atlassian.net/browse/HHH-11147

Adds the ability to use enhanced entities as a "proxy", meaning being able to return an entity knowing just its identifier using a new interceptor implementation.


hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/AbstractInterceptor.java
```diff
@@ -0,0 +1,161 @@
...
+ */
+package org.hibernate.bytecode.enhance.spi.interceptor;
+
+import org.hibernate.engine.spi.PersistentAttributeInterceptor;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+
+/**
+ * @author Steve Ebersole
+ */
+public abstract class AbstractInterceptor implements PersistentAttributeInterceptor, SessionAssociableInterceptor {
```
dreab8: just a minor, `PersistentAttributeInterceptor` can be removed, `SessionAssociableInterceptor` already extends `PersistentAttributeInterceptor`

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/BytecodeLazyAttributeInterceptor.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Set;
+
+import org.hibernate.Incubating;
+import org.hibernate.engine.spi.PersistentAttributeInterceptor;
+
+/**
+ * @author Steve Ebersole
+ */
+@Incubating
+public interface BytecodeLazyAttributeInterceptor extends PersistentAttributeInterceptor, SessionAssociableInterceptor {
```
dreab8: as in the previous comment, `PersistentAttributeInterceptor` can be removed.

hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/group/ComplexFetchGroupTest.java
```diff
@@ -25,80 +25,200 @@
...
+@TestForIssue(jiraKey = "HHH-11223")
+@RunWith(BytecodeEnhancerRunner.class)
+@CustomEnhancementContext(EnhancerTestContext.class)
+//@FailureExpected( jiraKey = "HHH-11223" )
 public class ComplexFetchGroupTest extends BaseNonConfigCoreFunctionalTestCase {
 
 	@Test
-	public void testNonOwningOneToOneAccess() {
+	public void testLoadNonOwningOneToOne() {
+		// Test loading D and accessing E
```
dreab8: I think we can remove this comment

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultMergeEventListener.java
```diff
@@ -191,36 +208,54 @@ protected void entityIsTransient(MergeEvent event, Map copyCache) {
 		LOG.trace( "Merging transient instance" );
 
 		final Object entity = event.getEntity();
-		final EventSource source = event.getSession();
+		if ( entity instanceof PersistentAttributeInterceptable ) {
```
dreab8: this check can be removed

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
```diff
@@ -253,17 +253,27 @@ private Object proxyOrLoad(
...
+					return managed;
+				}
 			}
 
 			// we won't create proxies, but if there is already one associated with
 			// the PC we should use it
-			final Object proxy = persistenceContext.getProxy( keyToLoad );
-			if ( proxy != null ) {
-				return proxy;
+			if ( persister.hasProxy() ) {
```
dreab8: inside this block  `persister.getEntityMetamodel().getBytecodeEnhancementMetadata().isEnhancedForLazyLoading()` have to be `true` so `persister.hasProxy()` so if i do no miss something based on the implementation 
`public boolean hasProxy() {
		// skip proxy instantiation if entity is bytecode enhanced
		return entityMetamodel.isLazy() && !entityMetamodel.getBytecodeEnhancementMetadata().isEnhancedForLazyLoading();
	}`
it will be always `false`

hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
```diff
@@ -1171,6 +1181,7 @@ public final Object internalLoad(
 				event.setEntityId( null );
 				event.setInstanceToLoad( null );
 				event.setResult( null );
+				event.setShouldUnwrapProxy( unwrapProxy );
```
dreab8: I think the set should be moved before the `fireLoad( event, type );` at line 1174.

gbadner:
Rebased master with commit that skips failing OffsetTime tests, and force-pushed.

sebersole:
The merge tests seemed wrong to me.  I changed the tests and all 3 pass - I was kind of surprised the "merge detached uninitialized bytecode proxy" passed as-is.

Anyway, fixed tests, squashed and force pushed.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -4279,6 +4307,50 @@ public Object load(Serializable id, Object optionalObject, LockOptions lockOptio
...
+					entity,
+					session,
+					LockOptions.READ
+			);
+
+			final LazyAttributeLoadingInterceptor interceptor = enhancementMetadata.injectInterceptor(
+					entity,
+					proxyInterceptor.getEntityKey().getIdentifier(),
+					session
+			);
```
gbadner: IIUC ```PojoEntityTuplizer#afterInitialize``` has already injected the ```LazyAttributeLoadingInterceptor```. I don't think it hurts anything, but seems redundant.

hibernate-core/src/main/java/org/hibernate/Hibernate.java
```diff
@@ -187,7 +202,10 @@ public static boolean isPropertyInitialized(Object proxy, String propertyName) {
 
 		if ( entity instanceof PersistentAttributeInterceptable ) {
 			PersistentAttributeInterceptor interceptor = ( (PersistentAttributeInterceptable) entity ).$$_hibernate_getInterceptor();
-			if ( interceptor != null && interceptor instanceof LazyAttributeLoadingInterceptor ) {
+			if ( interceptor instanceof EnhancementAsProxyLazinessInterceptor ) {
+				return false;
+			}
+			if ( interceptor instanceof LazyAttributeLoadingInterceptor ) {
 				return ( (LazyAttributeLoadingInterceptor) interceptor ).isAttributeLoaded( propertyName );
```
gbadner: Currently, ```Hibernate.isPropertyInitialized( activity, "instruction" )``` will return ```true``` if the attribute is initialized to an uninitialized proxy.

At the same time, ```Hibernate.isInitialized( activity.getInstruction() )``` will return ```false```.

I find this confusing. IMO, the return values for both should be ```false```.

hibernate-core/src/main/java/org/hibernate/boot/spi/SessionFactoryOptions.java
```diff
@@ -308,4 +308,8 @@ default boolean areJPACallbacksEnabled() {
 		return true;
 	}
 
+	/**
+	 * Can bytecode-enhanced entity classes be used as a "proxy"?
+	 */
+	boolean isEnhancementAsProxyEnabled();
```
gbadner: Since this is a new SPI method, shouldn't this be added as a default method returning ```false```?

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/AbstractInterceptor.java
gbadner: Also, ```AbstractSessionAssociableInterceptor``` might be a better name.

gbadner: Should this class implement ```Serializable```?

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/AbstractInterceptor.java
```diff
@@ -0,0 +1,161 @@
...
+		return session;
+	}
+
+	@Override
+	public void setSession(SharedSessionContractImplementor session) {
+		this.session = session;
+		if ( session != null && !allowLoadOutsideTransaction ) {
+			this.allowLoadOutsideTransaction = session.getFactory().getSessionFactoryOptions().isInitializeLazyStateOutsideTransactionsEnabled();
+			if ( this.allowLoadOutsideTransaction ) {
+				this.sessionFactoryUuid = session.getFactory().getUuid();
```
gbadner: Is it possible that this ```AbstractInterceptor``` will become associated with a ```Session``` from a different ```SessionFactory```? If so, ```sessionFactoryUuid``` won't be updated if ```allowLoadOutsideTransaction``` == ```true```.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/AbstractInterceptor.java
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	@Override
+	public SharedSessionContractImplementor getLinkedSession() {
+		return session;
+	}
+
+	@Override
+	public void setSession(SharedSessionContractImplementor session) {
+		this.session = session;
```
gbadner: Should further checks be done when  ```this.session != null && session != null```, similar to what is done in [```AbstractPersistentCollection#setCurrentSession```](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/collection/internal/AbstractPersistentCollection.java#L678-L698)?

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementAsProxyLazinessInterceptor.java
```diff
@@ -0,0 +1,165 @@
...
+			SharedSessionContractImplementor session) {
+		super( entityName, session );
+
+		this.identifierAttributeNames = identifierAttributeNames;
+		assert identifierAttributeNames != null;
+
+		this.nonAggregatedCidMapper = nonAggregatedCidMapper;
+		assert nonAggregatedCidMapper != null || identifierAttributeNames.size() == 1;
+
+		this.entityKey = entityKey;
```
gbadner: ```entityKey``` must be non-null?

hibernate-core/src/main/java/org/hibernate/Hibernate.java
gbadner: I see that in ```FetchGraphTest#testLoadNonOwningOneToOne```, that this is intended.

It would be helpful to add documentation to this method that a return value of ```true``` could indicate that the property can be initialized to an uninitialized collection or proxy.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -4279,6 +4307,50 @@ public Object load(Serializable id, Object optionalObject, LockOptions lockOptio
...
+			}
+
+			return interceptor.readObject(
+					entity,
+					nameOfAttributeBeingAccessed,
+					value
+			);
+		}
+
+		throw new IllegalStateException(  );
```
gbadner: Probably won't happen, but there should be a description of the exception.

hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
```diff
@@ -798,7 +814,11 @@ Object createProxy(Serializable id, SharedSessionContractImplementor session)
 	EntityTuplizer getEntityTuplizer();
 
 	BytecodeEnhancementMetadata getInstrumentationMetadata();
-	
+
+	default BytecodeEnhancementMetadata getBytecodeEnhancementMetadata() {
+		return getInstrumentationMetadata();
+	}
```
gbadner: ```getBytecodeEnhancementMetadata()``` does not seem to be used anywhere. Should it be removed?

hibernate-core/src/main/java/org/hibernate/Hibernate.java
sebersole: I disagree.  The contract of isPropertyInitialized is whether that value is available on the class

hibernate-core/src/main/java/org/hibernate/boot/spi/SessionFactoryOptions.java
sebersole: Maybe

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/AbstractInterceptor.java
sebersole: Have you run into this before?  This is exactly the code as it happened before this work.  I see what you are saying, but I think that has nothing to do with this work.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementAsProxyLazinessInterceptor.java
sebersole: Ofc

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
sebersole: * `#hasProxy` - can it produce HibernateProxy instances
* `#isEnhancedForLazyLoading` - is the entity enhanced

hibernate-core/src/main/java/org/hibernate/Hibernate.java
sebersole: Unless I am off base, the same should be true in 4.x if an object is enhanced and has a proxy for an association.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/AbstractInterceptor.java
sebersole: It would not hurt I guess.  But again this was taken directly from the existing interceptor code.  Have you seen trouble with that?  I think this is another one of "not specific to this work"

hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
sebersole: No.  Actually the other way around - that is the method we should use.  I got caught not finishing what I intended there.  I have changed that (and some other things) and will push soon

gbadner:
Supersedes https://github.com/hibernate/hibernate-orm/pull/2864

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementAsProxyLazinessInterceptor.java
```diff
@@ -93,62 +160,94 @@ public Object forceInitialize(Object target, String attributeName) {
```
dreab8: can we remove the` BytecodeLogger.LOGGER.tracef`? I think the `forceInitialize` method that is called after does it

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
```diff
@@ -309,7 +309,7 @@ private Object proxyOrLoad(
 			persistenceContext.addEntity(
 					entity,
 					Status.MANAGED,
-					ArrayHelper.filledArray( LazyPropertyInitializer.UNFETCHED_PROPERTY, Object.class, persister.getPropertyTypes().length ),
+					new Object[ persister.getPropertyTypes().length ],
```
gbadner: Why is this change necessary? Is it possible that a loaded property will be mistaken for null?

PR #2851, opened 2019-04-29
Fixed typo

Shanky2304:
Made minor edit to doc, replaced a `,` with `.` when referring to JDBC types.


beikov:
Thanks for the fix

PR #2852, opened 2019-05-01
HHH-13377 - remove lazy loaded attributes from initializedLazyFields for refresh of bytecode enhanced entities

jay-m-paulsen:
Remove lazy loaded attributes from the initializedLazyFields set in the entity's bytecode interceptor so accessing these attributes will reload them after refresh of entity.

I'm not sure if TwoPhaseLoad is the proper place for this fix.  I first tried clearing the initializedLazyFields set in the LazyAttributeLoadingInterceptor in the afterInitialize method of PojoEntityTuplizer as this seemed or more likely place to do this.  However, this caused a handful of existing test cases to fail.

This change is necessary so that an entity with initialized lazy loaded attributes has these attributes marked as uninitialized after the entity is refreshed.


dreab8:
Hi @jay-paulsen,

Thanks for your PR and sorry for taking so long to review it.
I'm going to close it in favour of https://github.com/hibernate/hibernate-orm/pull/9368

PR #2853, opened 2019-05-01
HHH-13380 - bytecode enhanced entities in OneToMany relationship with…

jay-m-paulsen:
… custom equals() in child entity throws LazyInitializationException.

The core of the fix is in PojoEntityInstantiator so that it can set the session in the LazyAttributeLoadingInterceptor.  This allows the equals method in the child entity of the OneToMany to refer to the lazily loaded Parent entity and get it initialized.  Without this change, the lazy loading of the OneToMany throws a LazyInstantiationException because it had no reference to the session in order to load the parent entity.

### DIFF
--- a/documentation/src/test/java/org/hibernate/userguide/proxy/tuplizer/DynamicInstantiator.java
+++ b/documentation/src/test/java/org/hibernate/userguide/proxy/tuplizer/DynamicInstantiator.java
@@ -11,6 +11,7 @@ package org.hibernate.userguide.proxy.tuplizer;
 import java.io.Serializable;
 
 import org.hibernate.HibernateException;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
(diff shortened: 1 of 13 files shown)


Sanne:
Hi @jay-m-paulsen - apologies for the very long delay. Found this now as I was debugging the same issue reported as https://github.com/quarkusio/quarkus/issues/11325 .

I'll review, rebase and apply your patch soon, although maybe I'll need to avoid changing the SPI.

Many thanks!

Sanne:
I've merged this, but using a different solution for the 5.4 and master branches so to not break the current SPI.

I like the solution proposed here better though, and have made a note to apply this idea on ORM6.

Thanks again!

jay-m-paulsen:
@Sanne - HHH-13377 is a similar problem that I reported around the same time as HHH-13380. You may find it relevant for your work. Thanks!

PR #2855, opened 2019-05-03
HHH-9861 Fix wrong mapping of ColumnType FLOAT for H2 database

hTnevS:
H2 has no float datatype. Instead, float is an alias for double. This leads to an error when hibernate is trying to validate the schema (found: double, expected: float).

Source for H2 datatypes:
http://www.h2database.com/html/datatypes.html


hTnevS:
Updated to resolve merge conflicts.

hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
```diff
@@ -118,7 +118,8 @@ public H2Dialect() {
 		registerColumnType( Types.DECIMAL, "decimal($p,$s)" );
 		registerColumnType( Types.NUMERIC, "decimal($p,$s)" );
 		registerColumnType( Types.DOUBLE, "double" );
-		registerColumnType( Types.FLOAT, "float" );
+		// H2 does not have a float type. Instead, it's mapping floats as alias to double
+		registerColumnType( Types.FLOAT, "double" );
```
murdos: I would do things differently, and register h2 "real" database type to java FLOAT: 
`registerColumnType(Types.FLOAT, "real");`

FYI this what has been done for the [JHipster ](https://www.jhipster.tech) generator, and thus what is used by many generated applications: https://github.com/jhipster/jhipster/blob/master/jhipster-framework/src/main/java/io/github/jhipster/domain/util/FixedH2Dialect.java#L36

hTnevS:
Hi murdos, sorry for the late reply. I think I'm fine with that. Could you elaborate tho, why JHipster decided to register it as real and not as double? H2 documentation seems to allow both.

NathanQingyangXu:
> Hi murdos, sorry for the late reply. I think I'm fine with that. Could you elaborate tho, why JHipster decided to register it as real and not as double? H2 documentation seems to allow both.

From the datatype page of h2, it seems `REAL` is more appropriate to map to float:
Screen Shot 2020-09-02 at 11 02 04 AM

PR #2856, opened 2019-05-03
HHH-13389: Fix JPA metamodel generation NPEs for any-to-many access="noop" properties

mikejhill:
JPA metamodel generation fails with NPEs with `access="noop"` properties when these properties use collection associations (any-to-many). This PR replaces the `null` member used in these property accessors with a non-null placeholder (`NoopMember`).


dreab8:
Hi @mikejhill ,

thanks for your PR and sorry for taking so long to review it. hbm has been deprecated and we are planning to remove its support so we decided not to apply any improvement.

PR #2858, opened 2019-05-04
HQL functions

gavinking:
This patch adds working support to the H6 query language for the following list of "standard" functions:

- `cast()` and `extract()`
- null handling functions `coalesce()`, `nullif()`, `ifnull()`
- string functions `concat()`, `locate()`,  `position()`, `substring()`, `trim()`, `lower()`, `upper()`, `replace()`, `length()`
- numeric functions `abs()`, `mod()`, `sqrt()`, `ln()`, `exp()`, `power()`, `sign()`, `ceiling()`, `floor()`, `round()`, `sin()`, `cos()`, `tan()`, `asin()`, `acos()`, `atan()`, `atan2()`
- comparison functions `least()` and `greatest()`
- aggregate functions `avg()`, `min()`, `max()`, `sum()`, `count()`, `count(*)`
- datetime functions `current_date`, `current_time`, `current_timestamp`, `current_instant`
- legacy HQL functions `str()`, `year()`, `month()`, `day()`, `hour()`, `minute()`, `second()`

Many of these functions were already partially implemented, or even partially working, but none were passing the basic tests I've added.

The patch also allows query parameters to occur in the select clause.

Finally, this patch completely rewrites the SQL function registration for the `Dialect`s, cleaning up a pile of unmaintained code with lots of mistakes. I've tried to check everything against the documentation for the various databases, and I've tried to be consistent about which functions we include, and which we ignore.


gavinking:
I'm still missing `power()`, `ceiling()`, and `floor()`.

gavinking:
It's not every day you get to remove almost 6000 LOC while also adding functionality and fixing lots of little bugs ;-)l

PR #2860, opened 2019-05-08
fix non-standard functions in select clause

gavinking:
Function templates are created by the `Dialect`s with refs to `StandardSpiBasicTypes` as return types, which need to be resolved before trying to use the produced function in a select.

This patch makes this JPA syntax work correctly:

    select function('ascii', 'x')


gavinking:
This patch was needed in, and therefore merged into both #2879 and #2863, so I'm going to close it here.

PR #2862, opened 2019-05-09
consistent treatment of literals

gavinking:
This is a minor cleanup to handle all the types of `SqmLiteral` in the same place, with consistent code.


gavinking:
This was merged into #2863.

PR #2863, opened 2019-05-09
datetime operators and functions in HQL

gavinking:
This patch introduces a useful set of tools for working with date and time types in HQL.

## Date/time duration arithmetic

First, the patch introduces a powerful ANSI-like syntax to HQL for doing arithmetic with dates, timestamps, and durations, as an abstraction over:

- `timestampadd()` and `timestampdiff()`,
- `dateadd()` and `datediff()`,
- interval arithmetic in Postgres, Oracle, and Teradata, or
- `xxxs_between()` and `add_xxxs()` functions on DB2

This is a *very useful* feature, since HQL didn't previously have any operations at all for doing arithmetic with dates and times.

The syntax is composed of the following basic elements:

- literal durations, given a unit and magnitude, e.g. `1 day`, `12 year`, `0.03 second`
- conversion of duration to scalar magnitude with `by` operator, given a unit, e.g. `x by day`, `y by second`
- addition and subtraction of durations with `+`, `-`, e.g. `1 day - 3 hour`
- scalar multiplication for durations with `*`, e.g. `2 * 1 month`
- adjustment of a date or time by a duration with `+`, `-`, e.g. `current date + 1 year`
- duration expressed as difference between two dates or two timestamps with `-`, e.g. `current datetime - record.timestamp`

The accepted duration units are a subset of the allowed set of the fields understood by the `extract()` function: `year`, `quarter`, `month`, `week`, `day`, `hour`, `minute`, `second`, and `nanosecond`.

- The difference between two dates is a `day` duration.
- The difference between two timestamps is a `nanosecond` duration.
- Persistent attributes of type `java.time.Duration` mapped to columns of type `BIGINTEGER` or `DECIMAL` by the pre-existing Hibernate built-in type are `nanosecond` durations.

### Kinds of duration

There are two basic families of durations:

- `year`/`month` durations, and
- `week`/`day`/`hour`/`minute`/`second`/`nanosecond` durations.

A `java.time.Duration` is always a `nanosecond` duration. There is no Hibernate nor Java type which represents a `year`/`month` duration.

It's legal to write:

    current datetime + 1 month + 2 hour

It's even legal to write:

    current datetime + (1 month + 2 hour)

(Since this gets rewritten to the first form by the translator.)

But it's *not* legal to write:

    (1 month + 2 hour) by hour

Nor:

    (1 month + 10 week) by month

Since it's conceptually impossible to convert months to hours or weeks to months.

### Differences from ANSI SQL intervals

ANSI SQL features the (misnamed) `INTERVAL` family of types. The duration system here is conceptually very similar, but very-much simplified:

- ANSI SQL has a profusion of different explicit `INTERVAL` types, for example `INTERVAL DAY`, `INTERVAL MONTH`, `INTERVAL HOUR TO SECOND`, `INTERVAL YEAR TO MONTH` capturing the different combination of units that may occur in a duration, whereas the only explicit duration type in HQL is the day-to-nanosecond type `java.time.Duration`.
- Thus, ANSI SQL interval values are always explicitly labeled with their unit(s) when created, always carry their unit(s) along with them, and the unit(s) are preserved by all operations. On the other hand, the Java `Duration` type has no unit field, so HQL durations are sometimes implicitly converted to `nanosecond`s (when persisted, or when projected). Furthermore, the difference between two dates or two datetimes produces a value with an implicit unit of `day` or `nanosecond`.

So HQL is slightly less strongly-typed in its handling of units. However, the unit of a duration *is* preserved during addition, subtraction, and scalar multiplication, though that behavior is almost invisible to the application. (You have to look at the generated SQL to notice it!)

Moreover:

- ANSI SQL features an amazingly complicated system for writing down multi-unit duration literals. For example, one can write `10 year + 2 month` as `INTERVAL '10-02' YEAR TO MONTH`. This appears to be an entirely useless feature, and I don't know what they were thinking. The overwhelmingly common case is a single-unit duration, which looks like `INTERVAL '3' DAY` in ANSI SQL.
- Converting from an `INTERVAL` type to an integer is very complicated in ANSI SQL. First you have to convert your multi-field interval to an interval with just one field using `cast()`, then you have to `cast()` *again* to a numeric type. (Some databases don't even seem to properly support this sequence of operations, leaving you to `extract()` stuff field-by-field and then add them up!) Any HQL duration can be converted to an integer quickly using the `by` operator, specifying the unit.

### Additional examples

Here's some additional examples.

Add a duration in hours to a timestamp:
    
    event.start + 24 hour

Subtract a duration in weeks from a date:

    person.dob - 2 week

Add/subtract two durations:

    2 week + 24 hour
    1 day - 12 minute

Multiply a duration by a scalar:

    2 * x.n day
 
Measure the difference between two timestamps in seconds:
    
    (event.end - event.start) by second

Add twice the difference between two timestamps to another timestamp:
    
    current datetime + 2 * (event.end - event.start)

Convert between duration units:

    (2 week + 1 day) by hour

Mix it all together:

    (current time + 1 hour - event.end) by second

Use persistent attributes of type `java.time.Duration` in duration arithmetic:

    current date + n * thing.periodicity
    thing.periodicity + 1 second
    cal.repeat by minute

## Date/time formatting

Second, the patch introduces a standard `format()` function that allows formatting of a datetime according to the format strings defined by `java.time.format.DateTimeFormatter`, implemented based on:

- `to_char()` in most databases
- `date_format()` in MySQL and Ingres
- `format()` in SQL Server
- `formatdatetime()` in H2
- `varchar_format()` in DB2

Thus, we can write stuff like:

    select format(current datetime as 'EEE, dd/MM/yy HH:mm:ss') from Entity

And have it be portable across databases. This is useful, since the syntax for date/time formatting is quite vendor-specific and unportable.

### Date/time format strings

The recognized format elements are:

- `G`, `GG` the era (does not work everywhere)
- `y`, `yy`, `yyy`, and `yyyy`, the year
- `M`, `MM`, the month as a number
- `MMM`, `MMMM`, the month as a word
- `w`, `ww`, the ISO week of the year (does not work everywhere)
- `Y`, `YY`, `YYY`, `YYYY`, the ISO year for the week of year (does not work everywhere)
- `W`, the week of the month (works hardly anywhere)
- `EEE`, `EEEE`, the day of the week as a word
- `e`, `ee`, the day of the week as a number (does not produce consistent results)
- `d`, `dd`, the day of the month
- `D`, `DD`, `DDD`, the day of the year
- `a`, `aa`, AM or PM
- `h`, `hh`, the clock hour from 1 to 12
- `H`, `HH`, the hour from 0 to 23
- `m`, `mm`, the minute
- `s`, `ss`, the second
-  `S` .. `SSSSSS`, fractional seconds
- `x`, `xx`, `xxx`, the timezone offset (does not work everywhere)
- `Z`, `ZZ`, `ZZZ`, the timezone offset (works hardly anywhere)
- `z`, `zz`, `zzz`, the timezone name (works hardly anywhere)
- punctuation characters including `:`, `/`, `-`, `.`, `()`, `[]`, `@` are interpreted literally
- the content of `'single-quoted strings'` is interpreted literally

## Additional extractable date/time field types

Next, the patch expands the list of `extract()` field types, adding:

- `day of week`
- `day of month` (a synonym for `day`)
- `day of year`
- `week of month` (where the 1st is in week 1, and weeks end on Sat)
- `week of year` (where the 1st of Jan is in week 1, and weeks end on Sat)
- `offset` (i.e. the timezone offset)
- `offset hour`
- `offset minute`
- `time` (the time part of a timestamp)
- `date` (the date part of a timestamp)
- `nanosecond` (the `second` field, multiplied by one billion)

Furthermore: 

- `second` has been standardized to always include fractional seconds, and
- `week` has been standardized to mean the ISO week number. (Note that `week` and `week of year` are *not* synonyms.)

Both of these fields previously exhibited inconsistent semantics across dialects. 

This work enhances the existing support for `year`, `quarter`, `month`, `week`, `day`, `hour`, `minute`, `second`. However:

- the fields `millisecond` and `microsecond` have been removed.

It's important to note that not every field type works on every platform. In particular, very many databases don't have any sort of meaningful support for timezones.

Note that `select extract(offset from thing.timestamp)` returns an instance of `java.time.ZoneOffset`, via a new built-in standard type for mapping the type `ZoneOffset`.

Note also that none of these new field types may be used as units of duration.

## Literal date/times and current date/times for `java.time` types

The patch incorporates support for the following literal syntax for `java.time` date/time types:

- `date yyyy-MM-dd` produces a `LocalDate`
- `time HH:mm:ss` produces a `LocalTime`
- `time HH:mm` produces a `LocalTime`
- `datetime yyyy-MM-dd HH:mm:ss` produces a `LocalDateTime`
- `datetime yyyy-MM-dd HH:mm:ss 'XXX/Xxxxx'` produces a `ZonedDateTime`
- `datetime yyyy-MM-dd HH:mm:ss 'GMT'` produces a `ZonedDateTime`
- `datetime yyyy-MM-dd HH:mm:ss '+HH:mm'` produces a `ZonedDateTime`
- `offset datetime yyyy-MM-dd HH:mm:ss +HH:mm`  produces an `OffsetDateTime`
- `offset datetime yyyy-MM-dd HH:mm:ss +HH`  produces an `OffsetDateTime`

Alternatively, a quoted string may be used, for example, `date 'yyyy-MM-dd'` or `time 'HH:mm:ss'` as in ANSI SQL.

This comes along with the following syntax for `java.time` current date/times:

- `current instant` produces an `Instant`
- `current date` produces a `LocalDate`
- `current time` produces a `LocalTime`
- `current datetime` produces a `LocalDateTime`

This new syntax is more consistent with the more Java-like naming conventions used throughout the rest of the language than the syntax defined by the JPA spec.

## Improved dialect support for SQL date/time functions

Finally, this patch adds additional  datetime-related native SQL function registrations to the `Dialect`s.


gavinking:
I've cherry-picked the work from #2870 onto this branch because I needed stuff from here in order to implement `extract(offset from ...)`, which is now working on this branch, producing a `ZoneOffset`.

gavinking:
I merged in #2830.

gavinking:
Need to properly consider alternatives to the syntax 

    (event.end - event.start) by second

This `by` keyword is conceptually an operator that converts a duration to an integer, given a unit. I think it's cute and reads well, but perhaps it's too cute. I've been trying to think of a decent way to express this using a more function-like syntax (by analogy with `extract()`). The best options I've come up with are:

     measure(second in event.end - event.start)

     measure(event.end - event.start by second)

     duration(second of event.end - event.start)

     duration(event.end - event.start as second)

I'm not sure if any of those options are better.

gavinking:
Actually here are my two current favorite syntax options, which allow one to and subtract durations, and multiply them by a scalar.

This is what I have working right now:

    (current date - e.start) by day
    e.start + 2 week + 1 day

This is very close to ANSI SQL. But I feel it's an abuse of the `+` and `-` notation. A reasonable alternative, and probably more readable is:

    duration(e.start to current date) by day
    shift(e.start by 2 week + 1 day)

Here, the function-like syntax `duration(x to y)` is doing the work of `-`, making clear that we're producing a duration, before casting it to an integer using `by day`. And `shift(x by y)` is doing the work of `+`.

gavinking:
Here's a third option for syntax, that makes a whole lot of sense. Instead of `+`/`-`, or `shift()/`duration()`, we could use three binary operators named `before`, `after`, and `until`:

    e.start until current date by day    // i.e. (e.start until current date) by day
    1 day before e.start
    2 week after current date
    (2 week - 1 day) after e.start
    1 day before 2 week after e.start
    3 minute by millisecond
    (1 hour + 3 minute) by second

I'll have to think a little carefully about the correct precedence for these operators, but I think something like this could work (in order of decreasing precedence)

- postfix unit
- `before`, `after`
- `until`
- `by` unit
- `*`
- `+`, `-`

Note that it's reasonable to use `+`, `-`, `*` on durations, since they form a vector space.

gavinking:
I have merged #2860 into this branch.

gavinking:
I realized I'm making a mistake here. In the following new constructs:

    cast(x as Time)
    cast(x as Date)
    cast(x as Timestamp)
    
    extract(date from x)
    extract(time from x)
    
    current date
    current time
    current timestamp

I've used the legacy `javax.sql` types `Date`, `Time`, and `Timestamp`. These are really bad (times which inherit dates!) and nobody should use them anymore.

So really the type of the above constructs should be `LocalDate`, `LocalTime`, `LocalDateTime` from `java.time`. Which is an easy change to make. **But**, I'm not sure about the naming here. Should it be:

    cast(x as DateTime)
    
    current datetime

or even:

    cast(x as LocalTime)
    cast(x as LocalDate)
    cast(x as LocalDateTime)

But I'm not very keen on:

    current local time
    current local date
    current local date time

That's a really verbose syntax, that might sorta make sense of there was a use for `current offset date time` or `current zoned date time`, but sadly there's almost no database that those things would make sense on.

I guess there's nothing especially wrong with sticking with `timestamp` instead of `datetime`, but I'm just concerned that it's obviously misleading.

gavinking:
I've decided to go with:

    cast(string as LocalTime)
    cast(string as LocalDate)
    cast(string as LocalDateTime)

(At least for now.)

Still weighing up the pros/cons of introducing a `datetime` keyword. I think I've narrowed the options down to:

    current datetime
    datetime '10:30:00 2003-12-30'

vs:

    current date time
    date time '10:30:00 2003-12-30'

I guess I find either to be perfectly acceptable.

gavinking:
I've implemented:

    current time
    time '10:30:00'
    current date
    date '2003-12-30'
    current datetime
    datetime '2003-12-30 10:30:00'

Though, as I was working, a third syntax option occurred to me, that perhaps emphasizes the connection to the Java types:

    local time
    local time '10:30:00'
    local date
    local date '2003-12-30'
    local datetime
    local datetime '2003-12-30 10:30:00'

gavinking:
I'm now considering the following alternative to the `by` operator:

    cast(duration as Integer by hour)

JWGmeligMeyling:
> There is no Hibernate nor Java type which represents a year/month duration.

I this could be achieved through Java's [`Period`](https://docs.oracle.com/javase/8/docs/api/java/time/Period.html) class, a year/month/day based counter part of `Duration` .

gavinking:
> > There is no Hibernate nor Java type which represents a year/month duration.
> 
> I think this could be achieved through Java's [`Period`](https://docs.oracle.com/javase/8/docs/api/java/time/Period.html) class, a year/month/day based counter part of `Duration`.

@JWGmeligMeyling I thought of that but the problem is what would you do with the day part of a `Period`?

JWGmeligMeyling:
I think the `+interval 1 day` handling for most DBMSes will actually be closer to the behaviour described in `Period` than the one in `Interval`. But perhaps it doesn't matter to add the day part for both Period and Interval? But anyway, we agree that an interval with both years or months and hours or minutes etc can be expressed in Java. The parts would have to be separated.

gavinking:
@JWGmeligMeyling what is `Interval` here? do you mean to say `Instant`?

The issue is that a month isn't a well defined number of days. OK, sure, if we want to be super-fussy about this, a day isn't a well-defined number of seconds either, given stuff like leap seconds, but it's close enough for most purposes in business computing.

JWGmeligMeyling:
I ment to say `Duration`, I got confused by the SQL jargon.

I agree on your examples.

Just wondering, `ts + 1 day` will still do a SQL `ts + 1 day` rather than a time-based equivalent, right? So given a long enough interval, `ts + 1 day <> ts + (1 day) as hour`?

gavinking:
> Just wondering, `ts + n day` will still do a SQL `ts + n day` rather than a time-based equivalent, right? 

Well it's all very dependent on the actual database. This is defined by `Dialect.timestampaddPattern()`:

- on Postgres it's `ts + n * interval '1 day'`
- on H2 it's `dateadd(day, n, ts)`
- on Oracle it's `ts + numtodsinterval(n, 'day')`
- on MySQL it's `timestampadd(day, n, ts)`

etc.

> So given a long enough interval, `ts + n day <> ts + (n day) as hour`?

I don't know what you mean.

PR #2866, opened 2019-05-09
add additional target types for cast() function

gavinking:
With this patch, the following additional `cast()` target types are now supported

- `Binary`
- `Character`
- `BigDecimal` and `BigInteger`
- `Boolean` (with caveats)
- `LocalTime`
- `LocalDate`
- `LocalDateTime`

Also, length or precision/scale may be specified for target types.

- `cast(number as String(10))`
- `cast(text as Binary(100)`
- `cast(amount as BigInteger(20))`
- `cast(measurement as BigDecimal(10, 5))`

Of course, the length, precision, or scale might be ignored, depending on the database type defined by the `Dialect`. (In particular, the are *always* ignored on Spanner, which doesn't support length, precision, or scale in `cast()`.)

Also, when any `DECIMAL` type occurs without an explicitly-defined precision, a database-specific default is used. This makes `cast`ing and DDL generation more portable across databases when you're using `BigInteger`s or `BigDecimal`s.

### DIFF
--- a/hibernate-core/src/main/antlr/org/hibernate/query/hql/internal/HqlParser.g4
+++ b/hibernate-core/src/main/antlr/org/hibernate/query/hql/internal/HqlParser.g4
@@ -611,14 +611,7 @@ castFunction
 	;
 
 castTarget
-	// todo (6.0) : should allow either
-	// 		- named cast (IDENTIFIER)
-	//			- JavaTypeDescriptorRegistry (imported) key
-	//			- java.sql.Types field NAME (alias for its value as a coded cast)
-	//			- "pass through"
-	//		- coded cast (INTEGER_LITERAL)
-	//			- SqlTypeDescriptorRegistry key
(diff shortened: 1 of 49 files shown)


gavinking:
Almost by side-effect, the patch also fixes an ancient problem where columns for mapped attributes of type `BigInteger` got a scale of 2 instead of 0. (A `JavaTypeDescriptor` now provides hints for default length/precision/scale instead of always defaulting to 255/19/2.)

gavinking:
I finally got fed up with maintaining this as a separate conflicting branch. So I cleaned up the conflicts, and merged it into #2879 as a single commit fc9f6025fb833564835a12e60d6e202563e44004 (though note that some earlier work had already been merged in 18123e704a1ec15b529e90395d59d793d740bfa2 and 5cbc2efa703c47d0285ada4333f5f55ced51cb16).

Closing.

PR #2870, opened 2019-05-11
 add 'day of week' and 'day of year' as allowed parameters to extract()

gavinking:
With this patch, the `extract()` function accepts `day of week` and `day of year`. Oh and `day of month` just for completeness (a synonym for `day`).

    select extract(day of year from person.dob) from Person person

It works on Ingres, SQL Server, Sybase, H2, HSQLDB, Postgres, DB2, and Cache.

It doesn't work on MySQL, Oracle, SAP, HANA, Informix, Unisys, McKoi, nor TimesTen.

So, exactly *half* of the platforms.

In principle `extract(day of year from ...)` could be made to work on some other platforms by date arithmetic. For example, in MySQL it could map to:

    timestampdiff(day, arg, makedate(extract(year from arg), 0))

But I'm not going to implement that RN.

### DIFF
--- a/hibernate-core/src/main/antlr/org/hibernate/query/hql/internal/HqlParser.g4
+++ b/hibernate-core/src/main/antlr/org/hibernate/query/hql/internal/HqlParser.g4
@@ -795,6 +795,7 @@ extractFunction
 
 extractField
 	: datetimeField
+	| dayField
 	| timeZoneField
 	| secondsField
 	;
@@ -809,6 +810,12 @@ datetimeField
 	| MINUTE
 	| SECOND
 	;
+
+dayField
+	: DAY OF MONTH
+	| DAY OF WEEK
+	| DAY OF YEAR
+    ;
 
 secondsField
 	: MILLISECOND
--- a/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
@@ -566,4 +566,15 @@ public class DB2Dialect extends Dialect {
 	public boolean supportsPartitionBy() {
 		return true;
 	}
+
+	@Override
+	public String translateExtractField(String fieldName) {
+		switch ( fieldName ) {
+			case "dayofmonth": return "day";
+			case "dayofyear": return "doy";
+			case "dayofweek": return "dow";
+			default: return fieldName;
+		}
+	}
+
 }
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -3257,4 +3257,13 @@ public abstract class Dialect implements ConversionContext {
 			return null;
 		}
 	}
+
+	public String translateExtractField(String fieldName) {
+		switch ( fieldName ) {
+			case "dayofmonth": return "dd";
+			case "dayofyear": return "dy";
+			case "dayofweek": return "dw";
+			default: return fieldName;
+		}
+	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/dialect/HSQLDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/HSQLDialect.java
@@ -650,4 +650,14 @@ public class HSQLDialect extends Dialect {
 	public String getCascadeConstraintsString() {
 		return " CASCADE ";
 	}
+
+	@Override
+	public String translateExtractField(String fieldName) {
+		switch ( fieldName ) {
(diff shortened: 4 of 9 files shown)


gavinking:
Hrm. I'm not very happy about the underscores in `timezone_hour` and `timezone_minute`. In most of the query language we use Java-style naming conventions, the unique current exception being that the JPA spec defines `current_date`, `current_time`, `current_timestamp`,  instead of Java-style `currentDate``currentTime`, `currentTimestamp` or `current date`, `current time`, `current timestamp` as supported in some databases.

For example, the argument to `cast()` is `BigInteger` or `BigDecimal`, so it seems like the argument to `extract()` should be `timezoneHour` or `timezoneMinute`. Though I would also be happy with `timezone hour` or `timezone minute`.

I think we should try hard to have a consistent naming convention here.

gavinking:
> Hrm. I'm not very happy about the underscores in `timezone_hour` and `timezone_minute`.

One solution would be to replace the `extract()` function by property access, for example:

    event.startDate.month

instead of

    extract(month from event.startDate)

It's certainly a lot less verbose.

sebersole:
-1 to the "virtual property" appproach

beikov:
What do you think about an "instance method" approach i.e. allowing to define function with a receiver type e.g. `event.startDate.getMonth()`?

sebersole:
I'm just not a fan of syntactically ambiguous paths - they are more
intensive generally speaking to parse.  My preference would be for
something like `month( event.startDate )` which is syntactically
unambiguous.

sebersole:
I also prefer the camelCase approach for the reasons Gavin mentions

beikov:
I'm specifically suggesting the "instance method" approach because instance methods are easier to discover when IDE tooling implements autocompletion, but that aspect is orthogonal to this discussion here.

gavinking:
> -1 to the "virtual property" appproach

Why, precisely?

I mean, the language already has a veneer of OO in it, and we do use OO-style of property access for other things (properties of entities, and of embeddeds).

And `LocalDate`, `LocalTime`, `LocalDateTime` already have almost  *exactly* the properties I need:

Screen Shot 2019-05-16 at 13 58 28

The only discomfort here is that there is no `offset.hours` nor `offset.minutes`, but why not just return the whole `offset` as an instance of `ZoneOffset`?

OK, sure, the `java.sql` subclasses of `Date` don't exactly correspond:

Screen Shot 2019-05-16 at 14 01 23

But I think that people shouldn't be using `java.util.Date`s anymore, so I don't really care about this.

gavinking:
> why not just return the whole offset as an instance of `ZoneOffset`?

On the other hand, that idea also neatly solves my problem with `extract()`. Instead of 

    extract(timezone hours from thing.start)

 I could let you write:

    extract(offset from thing.start)

Or

    extract(timezone from thing.start)

And return a `ZoneOffset` to the program. That should actually be quite straightforward to implement.

sebersole:
I've already explained why.

If you want to offer both options then I can accept that.  But the simple truth is that `month( event.start )` is far more performant to parse compared to `event.start.month` or `event.start.month()` or any other "virtual property" approach.  The parser can recognize `month(...)` syntactically and understand immediately what branch to take whereas in the virtual property approach it needs to route through our "path recognition" logic during semantic analysis.

We can agree to disagree I guess, but given the choice between (what you call) "OO" and performance I'd choose performance - and I'm sure I'm not  the only one.  Options for the end user are good I guess, so like I said, if you want to support both I am fine with that.

So I would suggest for a well-defined set of extractions (mainly around date/time) we offer both.  And that is easy to document in a table layout, e.g.:

extract function | simplified function | virtual property
---------------------| ------------------------ | ---------------------
`extract(month from thing)` | month(thing) | thing.month
... | ... | ...

gavinking:
> If you want to offer both options then I can accept that. But the simple truth is that `month( event.start )` is far more performant to parse compared to `event.start.month` or `event.start.month()` or any other "virtual property" approach. 

I don't think that's true. But even if, for the sake of argument, we assume it is true, how's it any harder or slower to parse and translate `event.start.month` than it is to parse and translate any other property of any regular embeddable? If we're already perfectly capable of interpreting:

    person.address.street

Then there's no problem at all making sense of

    person.dob.year

The only difference is that in the case of `person.address.street`, it maps directly to a column, whereas in the case of `person.dob.year`, it maps to a call to `extract()`. But from a performance/complexity perspective they're not much different at all.

gavinking:
I've cherry-picked the commit in this PR into #2863, because I needed some of the infrastructure I had developed over there to implement `offset`. I'm closing this PR, but we can keep discussing the syntax issue here, or in #2863, whichever you prefer.

PR #2871, opened 2019-05-11
promote left() and right() functions to work on all platforms

gavinking:
The standard `substring()` function defined (in slightly different forms) by ANSI SQL and JPA is useful, but can be quite hard to use for some simple things, especially since the XQLs have no `let` expressions.

For many common tasks, `left()` and `right()` are easier to use, especially since you don't need to compute the length of the string to use `right()`.

Hence, almost every database supports `left()` and `right()`, except Oracle and TimesTen which let you simulate `right()` by passing a negative index to `substring()`.

This patch makes it possible to use `left()` and `right()` on all platforms (including Oracle and TimesTen). Well, all platforms except McKoi, but I don't think we care about McKoi.


gavinking:
This is now part of #2879.

PR #2872, opened 2019-05-11
register insert() function for the Dialects which support it

gavinking:
It's definitely worth adding `insert()` to the lists of platform-specific functions.

I've even added an emulated `insert()` to Postgres, since it's currently impossible to call `overlay()` from HQL.


gavinking:
This is now part of #2879.

PR #2873, opened 2019-05-12
add missing trunc() / truncate() registrations

gavinking:
The first time around I missed some dialect registrations for `trunc()` / `truncate()`.

I also missed a really easy way to emulate it on SQL Server.


gavinking:
This is now part of #2879.

PR #2874, opened 2019-05-12
major improvements to the MimerSQLDialect

gavinking:
This patch makes `MimerSQLDialect` usable.

- Fixed the type mappings which were broken:
    - `BLOB`s are not `VARBINARY` as far as JDBC knows
    - `CLOB`s are not `VARCHAR` as far as JDBC knows
    - there were instances of `$1` instead of `$l`
- Use `CommonFunctionFactory` for SQL function registration
    - and remove registrations for functions that were removed from Mimer in 11.
- Update the sequence syntax for Mimer 11 and support pooled sequences


gavinking:
Closing since I have now merged these changes into a new branch.

PR #2875, opened 2019-05-12
adapt SpannerDialect to use CommonFunctionFactory

gavinking:
`SpannerDialect` wasn't reusing `CommonFunctionFactory` to register all it's SQL functions.

Also it was using upper-cased type and function names, which is inconsistent with the other dialects.


gavinking:
Closing since I have now merged these changes into a new branch..

PR #2876, opened 2019-05-12
delete from Dialect

gavinking:
Hibernate should absolutely take the opportunity of a H6 release to prune the `Dialect` hierarchy.

Here's my strawman proposal as a PR. The removed `Dialect`s were either:

1. deprecated years and years ago, or
2. for unmaintained/unused databases and
3. themselves unmaintained and/or of generally low quality.

For now I've left `FirebirdDialect` and `InterbaseDialect` alone, but frankly I'm itching to add them to the kill list.

In principle we could move the removed dialects to some sort of community-maintained-dialects project on github but frankly the community wasn't actually doing anything at all to maintain them, so I doubt it's worth it.


gavinking:
Closing this, since this didn't spark off any sort of discussion and all and anyway I now have a much clearer idea of which Dialects I think we should remove.

PR #2877, opened 2019-05-12
more SQL aggregate functions + make every()/any() standard

gavinking:
This patch registers more SQL aggregate functions with the `Dialect`s, especially:

- `median()`
- `stdev_pop()` and `var_pop()`
- `stdev_samp()` and `var_samp()`
- `bit_and()` and `bit_or()`

(Where supported.)

And in addition it introduces the ANSI SQL-standard aggregate functions `every()`/`any()` to HQL. These functions accept a *predicate* (not an expression!) On most platforms these are mapped to `sum()`, but some have native functions for this.


gavinking:
I had to pull in commit 67a68659f9fd6b90a42b58c5f6e9407ff207f42a from #2860 in order to make this stuff work correctly.

gavinking:
This is now part of #2879.

PR #2878, opened 2019-05-13
two fixes for boolean attribute handling

gavinking:
This patch makes two somewhat-related changes:

1. fixes rendering of boolean literals 'true' and 'false' to SQL
2. fixes type mapping for boolean attributes in HSQL

1: This is simply a fix for a bug in the new query parser

2: The HSQL documentation states explicitly that BIT is not the right way to
represent booleans in HSQL, and that BOOLEAN should be used instead.
And indeed, I was running in to problems arising from this incorrect
mapping.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/HSQLDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/HSQLDialect.java
@@ -129,7 +129,7 @@ public class HSQLDialect extends Dialect {
 
 		registerColumnType( Types.BIGINT, "bigint" );
 		registerColumnType( Types.BINARY, "binary($l)" );
-		registerColumnType( Types.BIT, "bit" );
+		registerColumnType( Types.BIT, "boolean" );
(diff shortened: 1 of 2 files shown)


gavinking:
Related to #2877.

hibernate-core/src/main/java/org/hibernate/dialect/HSQLDialect.java
```diff
@@ -146,7 +147,6 @@ public HSQLDialect() {
 		registerColumnType( Types.TIMESTAMP, "timestamp" );
 		registerColumnType( Types.VARCHAR, "varchar($l)" );
 		registerColumnType( Types.VARBINARY, "varbinary($l)" );
-		registerColumnType( Types.NCLOB, "clob" );
```
gavinking: Oops, this was a mistake, should not have removed this line.

gavinking:
Closing this PR without merging, because these fixes have been superceded by work on a different branch.

PR #2879, opened 2019-05-14
various Dialect enhancements

gavinking:
This patch incorporates the following changes in the area of `Dialect`s:

## Improvements to SQL type mappings

The patch improves the `Dialect`-specific SQL type name mappings:

- eliminate the massive duplication in type name registration; each dialect now only overrides what is non-standard
- use `timestamp with time zone` where possible, depending on the dialect (for Java types `OffsetDateTime`, `ZonedDateTime`, and `Calendar`)
- use `boolean` instead of `bit` where possible, depending on the dialect (for Java types `Boolean` and `boolean`)
- sort out some confusion surrounding binary string type mappings
- improvements specific to TimesTen and Mimer

## SQL type precision and scale

Most importantly, it fixes a bunch of issues surrounding precision and scale:

- allow `JavaTypeDescriptor`s to contribute a length/precision/scale to their mapped type
- fix the total mess that was `real` vs `float` vs `double` — since none of these type names has anything like a consistent precision across databases, and since `float` vs `double` is definitely *not* a difference of precision (!), explicit `float(24)` and `float(53)` are now used as consistently as possible
- allow user to specify `timestamp(p)` precision using `@Column(precision=p)`
- fix mapping of `BigInteger` to give it scale 0 — fixes a pretty ancient bug
- when any `decimal`, `numeric`, or `timestamp` type has no explicitly-defined precision, use a dialect-specific default — this makes `cast`ing and DDL generation more portable across databases, especially with `BigInteger` and `BigDecimal`
- change default mapping for `Duration` from `bigint` to `numeric(21,0)`

## Additional types in `cast()` function

The following additional `cast()` target types are now supported

- `Binary`
- `Character`
- `BigDecimal` and `BigInteger`
- `Boolean` (with caveats)
- `LocalTime`,  `LocalDate`, and `LocalDateTime`

Also, length or precision/scale may be specified for target types.

- `cast(number as String(10))`
- `cast(text as Binary(100)`
- `cast(amount as BigInteger(20))`
- `cast(measurement as BigDecimal(10, 5))`

Of course, the length, precision, or scale might be ignored, depending on the SQL type defined by the `Dialect`. (In particular, they're *always* ignored on Spanner, which doesn't support length, precision, or scale in `cast()`.)

## Additional "standard" HQL functions

HQL was enhanced with the following new "standard" functions:

- `left()` and `right()`, as simpler alternatives to `substring()`
-  `any()` and `every()` aggregate functions

## Additional dialect-specific function registrations

The following non-standard function registrations were added:

- add missing `trunc()`, `truncate()`, and `insert()` functions on most databases
- add a bunch of dialect-specific SQL aggregate functions, especially statistical functions
- registered some dialect-specific functions similar to `current_date`/`current_time`
- improvements specific to Mimer and Teradata
- migrate `MimerDialect` and `SpannerDialect` to use of `CommonFunctionFactory`

This patch also fixes the use of non-standard functions in the `select` clause.

## Function list

To allow the user to actually see what functions a dialect supports:

- add the `hibernate.dialect.show_sql_functions` property to allow the user to easily log all the registered SQL functions with their signatures to the console
- some misc improvements to SQL function registrations in order to get a more accurate signature list


gavinking:
> fix the total mess that was real vs float vs double .. since none of these type names has anything like a consistent precision across databases, and since float vs double is definitely not a difference of precision (!), explicit `float(24)` and `float(53)` are now used as consistently as possible

The irony here is that the interpretation of the precision  argument of the `FLOAT` type isn't totally consistent either. There are a couple of DBs which interpret it as decimal digits instead of binary digits. 

And then there's Sybase. Even after reading it about 20 times I still don't understand what Sybase interprets it to mean  ¯\_(ツ)_/¯. Somehow single precision is `p<=16`. Perhaps they have an idiosyncratic interpretation of "single precision". 

So anyway I will probably introduce an additional abstraction over float precision.

gavinking:
I've introduced some support for mapping things to `time[stamp] with timezone`. It's disabled for now because these JDBC types just aren't well-supported on most databases for now. But users of Oracle could surely take advantage of it in their user types. (And it's there for the future.)

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -210,21 +210,24 @@ protected Dialect() {
...
 		registerHibernateType( Types.BIT, 1, StandardSpiBasicTypes.BOOLEAN.getJavaTypeDescriptor().getTypeName() );
 		registerHibernateType( Types.BOOLEAN, StandardSpiBasicTypes.BOOLEAN.getJavaTypeDescriptor().getTypeName() );
-		registerHibernateType( Types.CHAR, StandardSpiBasicTypes.CHARACTER.getJavaTypeDescriptor().getTypeName() );
+		registerHibernateType( Types.CHAR, StandardSpiBasicTypes.STRING.getJavaTypeDescriptor().getTypeName() );
 		registerHibernateType( Types.CHAR, 1, StandardSpiBasicTypes.CHARACTER.getJavaTypeDescriptor().getTypeName() );
-		registerHibernateType( Types.CHAR, 255, StandardSpiBasicTypes.STRING.getJavaTypeDescriptor().getTypeName() );
 		registerHibernateType( Types.DATE, StandardSpiBasicTypes.DATE.getJavaTypeDescriptor().getTypeName() );
 		registerHibernateType( Types.DOUBLE, StandardSpiBasicTypes.DOUBLE.getJavaTypeDescriptor().getTypeName() );
-		registerHibernateType( Types.FLOAT, StandardSpiBasicTypes.FLOAT.getJavaTypeDescriptor().getTypeName() );
+		registerHibernateType( Types.FLOAT, StandardSpiBasicTypes.DOUBLE.getJavaTypeDescriptor().getTypeName() );
```
gavinking: I'm not 100% sure about this particular change right here.

gavinking:
> There are a couple of DBs which interpret it as decimal digits instead of binary digits.

I've fixed this issue, at least for Mimer and CUBRID. I don't have access to Sybase to try and figure out what its semantics are.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
gavinking: Well, I think it's right; the difference between `Types.FLOAT` and `Types.DOUBLE` has nothing to do with precision, so we shouldn't pretend it does.

gavinking:
I added the very-useful `hibernate.dialect.show_sql_functions` setting.

gavinking:
Damn, looks like due to some changes in the tests for envers, the build of this branch is now failing. 😟

gavinking:
FTR, I think it's commit @da13a0b401290283ed6fc0b6630d9e3379f19bbe that breaks my branch.

gavinking:
I have merged #2873 and #2872 into this branch.

gavinking:
I have merged #2860 into this branch.

gavinking:
I have merged #2860 into this branch.

gavinking:
I have merged #2871 and #2877 into this branch.

Naros:
@gavinking Thanks for the heads up!

I honestly dislike the fact that a few of the Envers tests operate this way in the first place, essentially they allow hbm2dll to create the schema then the test drops part of it to manually re-create with to allow null values in columns.  I made a note when I ported these to come back to them as a part of 6.0 and see if these can be re-worked not to operate this way in the first place.

The good news is this only impacts 3 test Envers classes overall and they all do precisely the same.  The columns that are actually re-created using this approach don't have any form of precision, length or scale so that's likely why the original author picked the method in the first place.

gavinking:
WTF, how can I have merge conflicts when I just rebased on the most recent revision of `wip/6.0`? I didn't think that was even possible!

JWGmeligMeyling:
> `any()` and `every()` aggregate functions

Any may cause ambiguity for subqueries that return a single boolean valued column. From the PostgreSQL documentation:

>Boolean aggregates `bool_and` and `bool_or` correspond to standard SQL aggregates every and any or some. As for any and some, it seems that there is an ambiguity built into the standard syntax:
>
>```
>SELECT b1 = ANY((SELECT b2 FROM t2 ...)) FROM t1 ...;
>```
>
>Here `ANY` can be considered either as introducing a subquery, or as being an aggregate function, if the subquery returns one row with a Boolean value. Thus the standard name cannot be given to these aggregates.

Are we sure `any` and `every` names are appropriate and non-ambiguous in JPQL/HQL?

gavinking:
> Are we sure any and every names are appropriate and non-ambiguous in JPQL/HQL?

They're unambiguous in context because of how they occur syntactically.

sebersole:
@gavinking I am going to make some time this week to help you get this up to date with the current 6.0 branch if you have the time as well.

gavinking:
> I am going to make some time this week to help you get this up to date with the current 6.0 branch if you have the time as well.

Great, let's do it!

sebersole:
Tomorrow work for you?

gavinking:
Yep, should work fine. You are in central time, right?

sebersole:
Yes.  Around 9 or 10 (my time) works best for me.  @dreab8 may join as well

gavinking:
Yep, 4pm my time looks good.

PR #2880, opened 2019-05-16
HHH-13106 : Fixes schema validation failure with PostgreSQL 10

kedar-joshi:
Few months ago, commit 1db476db included the fixes for [HHH-12973](https://hibernate.atlassian.net/browse/HHH-12973). This commit also added the following change in `PostgreSQL81Dialect.java`

```java
@Override
	// return "select relname from pg_class where relkind='S'"; // Removed
	return "select * from information_schema.sequences"; // Added
}
```

Few months later, commit 76981d9 added support for default generated identity columns introduced in PostgreSQL 10. See [HHH-13202](https://hibernate.atlassian.net/browse/HHH-13202) for details. Unfortunately `select * from information_schema.sequences` doesn't return implicit sequences generated for default identity columns. 

The combined effect of these two changes was that the `SequenceInformationExtractor::extractMetadata` returned incorrect number of sequences. Especially, the implicit sequences were no longer visible to Hibernate. This resulted in [HHH-13106](https://hibernate.atlassian.net/browse/HHH-13106) where setting `hibernate.hbm2ddl.auto` to `validate` would make application to fail with exception **Schema-validation: missing sequence**.

This PR tried to find a common ground between these changes to restore expected behavior.


vellotis:
I'll second kedar-joshi.

> **Is there any plan to fix-merge this?**

kislyakdmitry:
Guys, are you planning to merge this one or fix this issue in another PR?
This issue is still reproducible

beikov:
The test uses an entity with a sequence generator but the DDL uses identity generation. Obviously, there is a mismatch here and I don't see a bug. When you use `@GeneratedValue(strategy = IDENTITY)` on the entity along with `generated always as identity`, the validation will work.

No bug here, so I'll close this.

beikov:
Also see https://github.com/hibernate/hibernate-orm/pull/5530

PR #2881, opened 2019-05-17
HHH-13408 Add support for custom package.

serac:
Allow customizing the package of generated meta models via a new configuration option: `metaModelPackage="org.example.package"`.

See https://hibernate.atlassian.net/browse/HHH-13408

### DIFF
--- a/tooling/metamodel-generator/src/test/java/org/hibernate/jpamodelgen/test/metamodelpackage/MetaModelPackageAnnotationGeneratedTest.java
+++ b/tooling/metamodel-generator/src/test/java/org/hibernate/jpamodelgen/test/metamodelpackage/MetaModelPackageAnnotationGeneratedTest.java
@@ -29,4 +29,12 @@ public class MetaModelPackageAnnotationGeneratedTest extends CompilationTest {
 	public void testMetaModelPackageAnnotationGenerated() {
 		assertMetamodelClassGeneratedFor( TestEntity.class, ALT_PACKAGE );
 	}
+
+	@Test
+	@TestForIssue(jiraKey = "HHH-13408")
+	@WithClasses(ExtendedTestEntity.class)
+	@WithProcessorOption(key = JPAMetaModelEntityProcessor.META_MODEL_PACKAGE, value = ALT_PACKAGE)
+	public void testMetaModelPackageAnnotationGeneratedWithInheritance() {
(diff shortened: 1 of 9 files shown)


tooling/metamodel-generator/src/test/java/org/hibernate/jpamodelgen/test/metamodelpackage/MetaModelPackageAnnotationGeneratedTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.hibernate.jpamodelgen.test.util.WithClasses;
+import org.hibernate.jpamodelgen.test.util.WithProcessorOption;
+import org.junit.Test;
+
+import static org.hibernate.jpamodelgen.test.util.TestUtil.assertMetamodelClassGeneratedFor;
+
+/**
+ * @author Marvin S. Addison
+ */
+public class MetaModelPackageAnnotationGeneratedTest extends CompilationTest {
```
Naros: For completeness could you also test
* Empty string supplied package name
* Not supplying the configuration parameter, therefore assuming default behavior.

serac: Pushed commit with additional test cases as requested.

Naros:
@serac thanks for adding the requested test coverage.

serac:
Our team discovered a shortcoming of this patch set for a presumably uncommon JPA inheritance use case: when `metaModelPackage` is defined, an inherited class that defines no persistent fields is missing an import statement that causes compilation failure.  We have developed a small patch and included additional test coverage, but I wanted to check about process since this PR had already been signed off and scheduled. Should I push the additional commit to this branch, or open another PR?

sebersole:
It's not been merged yet, so pushing to this branch is fine

serac:
I am withdrawing this PR due to problems encountered during testing. I discovered that meta-models generated when `metaModelPackage` pointed to a package other that the default were not subject to having their fields populated as expected, which caused `NullPointerException` when using them to construct criteria queries. This suggests that additional changes are required in the persistence provider in order for this feature to be fully functional. Unfortunately puts the capability squarely beyond the "quick win" I had hoped for.

PR #2886, opened 2019-05-24
HHH-13334 - attribute converter add mapping of Java Byte type to JDBC…

Vinche59:
… TINYINT type.


gavinking:
The test passes in H7.

PR #2888, opened 2019-05-27
HHH-13154 Add support for JPA criteria on stateless sessions

yrodiere:
https://hibernate.atlassian.net//browse/HHH-13154

This is necessary to allow users to prepare for ORM 6 where we will remove Hibernate Criteria in favor of JPA criteria.

I don't think it introduces any binary incompatibility. Adding @gbadner just in case you want to double-check.

This merely pulls up the `createQuery(criteria)` method declarations from `Session` to `QueryProducer` (to also expose it in `StatelessSession`) and implementations from `SessionImpl` to `AbstractSharedSessionContract` (to also implement it in `StatelessSessionImpl`).

I had to leave the method declarations in `Session` in order to resolve the conflict between signatures from `QueryProducer` and from `EntityManager`.

The tests are rather light, but there's no point in re-testing all the different criteria ("between", "eq", "asc", ...), since they are largely independent from the session. So I just backported a basic test from ORM 6.

I also checked that this allows to solve [HSEARCH-3447](https://hibernate.atlassian.net/browse/HSEARCH-3447)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
@@ -17,6 +17,10 @@ import java.util.UUID;
 import javax.persistence.FlushModeType;
 import javax.persistence.TransactionRequiredException;
 import javax.persistence.Tuple;
+import javax.persistence.criteria.CriteriaDelete;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.CriteriaUpdate;
+import javax.persistence.criteria.Selection;
 
 import org.hibernate.AssertionFailure;
 import org.hibernate.CacheMode;
@@ -63,6 +67,8 @@ import org.hibernate.engine.transaction.internal.TransactionImpl;
 import org.hibernate.engine.transaction.spi.TransactionImplementor;
 import org.hibernate.id.uuid.StandardRandomStrategy;
 import org.hibernate.jpa.internal.util.FlushModeTypeHelper;
+import org.hibernate.jpa.spi.CriteriaQueryTupleTransformer;
+import org.hibernate.jpa.spi.HibernateEntityManagerImplementor;
 import org.hibernate.jpa.spi.NativeQueryTupleTransformer;
 import org.hibernate.jpa.spi.TupleBuilderTransformer;
 import org.hibernate.persister.entity.EntityPersister;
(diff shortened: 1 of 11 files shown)


hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
```diff
@@ -730,6 +739,83 @@ public QueryImplementor createQuery(String queryString) {
 		}
 	}
 
+	private CriteriaCompiler criteriaCompiler;
```
gsmet: I'm not a big fan of adding a field in the middle of the code.

hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
```diff
@@ -730,6 +739,83 @@ public QueryImplementor createQuery(String queryString) {
...
+
+	@SuppressWarnings("WeakerAccess")
+	protected CriteriaCompiler criteriaCompiler() {
+		if ( criteriaCompiler == null ) {
```
gsmet: This is never called in a concurrent way?

hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
yrodiere: I just moved this code from another class, so I thought I'd keep the same style. I can move the field to the top if you want?

hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
yrodiere: I'm 100% sure it's okay for `Session`, and for `StatelessSession` I don't think this class is thread-safe anyway: we use this pattern in other places, such as `org.hibernate.internal.AbstractSharedSessionContract#accessTransaction` or `org.hibernate.internal.AbstractSharedSessionContract#getSessionIdentifier`.

Even if this was called in a concurrent way, the worst that could happen is that we create the `CriteriaCompiler` twice. As far as I can tell, this would work just fine, since `CriteriaCompiler` is stateless.

hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
yrodiere: Done

hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
yrodiere: As discussed, `org.hibernate.internal.AbstractSharedSessionContract#accessTransaction` suffers from the same issue and is exposed (indirectly) through the `StatelessSession` interface, so it seems `StatelessSession` is not thread-safe anyway. Plus there's the matter of the connection, which probably cannot be used from multiple threads simultaneously.

yrodiere:
Merged, thanks!

PR #2889, opened 2019-05-27
HHH-13415 Restore build compatibility with JDK11.0.3

yrodiere:
https://hibernate.atlassian.net//browse/HHH-13415

Necessary to work on [HHH-13409](https://hibernate.atlassian.net//browse/HHH-13409) in particular.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,16 @@
 language: java
 
-jdk:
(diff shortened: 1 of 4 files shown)


yrodiere:
@gsmet Just rebased on what you did. Would you mind reviewing this?

PR #2890, opened 2019-05-27
HHH-13409 Hibernate ORM does not detect services provided by libraries in the module path

yrodiere:
https://hibernate.atlassian.net//browse/HHH-13409

~Based on #2889 ([HHH-13415](https://hibernate.atlassian.net//browse/HHH-13415)) and #2887, which should be merged first.~ => Done

### The problem

When a user application runs as a module (with a module-info.java), with a dependency to Hibernate ORM, everything works well. *But* if the application needs to rely on an external component that integrates into ORM through Java services (using the ServiceLoader), this component will not be detected.

The reason is that the `ServiceLoader`, when loading services from Java modules, does not look for `META-INF/services/*` files directly anymore. Instead it looks for a service catalog in a map whose keys are classloaders.

Since we load services by passing the `AggregatedClassLoader` to the `ServiceLoader`, this aggregated class loader is used to look for a service catalog, and since this classloader is completely synthetic (we created it ourselves), there is no service catalog to be found, and the service loader will not find any service.

### The solution

I implemented two solutions:

1. one that is simple, but that may introduce regressions (I'm not sure).
2. one that is complex, but tries harder to preserve the previous behavior when it comes to services loaded from the classpath.

The first solution is implemented in commit `HHH-13409 Make AggregatedServiceLoader compatible with the module path (JDK9+)`. Basically instead of creating one service loader with the aggregated class loader, we create one service loader per individual class loader and aggregate the resulting loaded services, trying to remove duplicates, if any.

The second solution `HHH-13409 Rework AggregatedServiceLoader to minimize the risk of regression`, makes some changes:

1. It detects at bootstrap which JDK is running, and just uses the old behavior for JDK8
2. Even when creating one service loader per class loader, it creates a service loader for the aggregated class loader first. In practice it means services will retrieved as before, except those that can be found in individual class loaders and not in the aggregated class loader.
3. To be extra careful, we use the new `ServiceLoader.Provider` interface introduced in JDK9 to retrieve services: it allows to de-duplicate services *before* instantiating them, by checking metadata ahead of instantiation, instead of instantiating every service and then removing the duplicates. It might help to avoid some unwanted side-effects.

I think the second solution is better and probably the one we should go with when we drop support for JDK8 anyway, but maybe it's too defensive. I could use a second opinion :)

### DIFF
--- a/documentation/documentation.gradle
+++ b/documentation/documentation.gradle
@@ -16,6 +16,7 @@ ext {
 			'hibernate-ehcache',
 			'hibernate-java8',
 			'hibernate-orm-modules',
+			'hibernate-integrationtest-java-modules',
 			'release'
 	]
 }
--- a/gradle/java-module.gradle
+++ b/gradle/java-module.gradle
@@ -81,7 +81,6 @@ dependencies {
 	testRuntime( libraries.log4j )
 	testRuntime( libraries.javassist )
 	testRuntime( libraries.byteBuddy )
-	testRuntime( libraries.woodstox )
 
 	//Databases
 	testRuntime( libraries.h2 )
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -53,9 +53,6 @@ ext {
             jandex:         'org.jboss:jandex:2.0.5.Final',
             classmate:      'com.fasterxml:classmate:1.3.4',
 
-            // Woodstox
-            woodstox:           "org.codehaus.woodstox:woodstox-core-asl:4.3.0",
-
             // Dom4J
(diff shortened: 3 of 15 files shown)


yrodiere:
@Sanne This fixes the problem we talked about during the last Hangout. Would you mind having a look?

Sanne:
hi @yrodiere it's looking great at high level, but I'm trying it out and got confused:

if I just checkout commit `6e387e0` (the integrationt tests) shouldn't it fail? it's not failing for me.

Sanne:
nevermind, was running with Java 8 :-)

Sanne:
merged, many thanks!

PR #2892, opened 2019-05-28
HHH-13418 : Revert log level caching

johnaohara:
Historically some classes have cached the logger level in static fields to prevent calls to the logging framework to reolve logging levels at run time.

Caching the log level at build time can cause a race condition in log levels at runtime, especially when the build system is multi-tiered. This is evident in a Quarkus native binary, as an incorrect log level at runtime;

https://github.com/quarkusio/quarkus/issues/2223

Optimisations in logging frameworks now mean that there is not a performance penalty for checking log level defined in the logging framework, removing the need to cache the log level.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultSaveOrUpdateEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultSaveOrUpdateEventListener.java
@@ -99,8 +99,7 @@ public class DefaultSaveOrUpdateEventListener extends AbstractSaveEventListener 
 	}
 
 	protected Serializable entityIsPersistent(SaveOrUpdateEvent event) throws HibernateException {
-		final boolean traceEnabled = LOG.isTraceEnabled();
-		if ( traceEnabled ) {
+		if ( LOG.isTraceEnabled() ) {
 			LOG.trace( "Ignoring persistent instance" );
 		}
 		EntityEntry entityEntry = event.getEntry();
@@ -137,7 +136,7 @@ public class DefaultSaveOrUpdateEventListener extends AbstractSaveEventListener 
 
 			}
 
-			if ( traceEnabled ) {
+			if ( LOG.isTraceEnabled() ) {
 				LOG.tracev(
 						"Object already associated with session: {0}",
 						MessageHelper.infoString( entityEntry.getPersister(), savedId, factory )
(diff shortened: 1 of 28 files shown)


sebersole:
The performance team has generally told us to do this in fact.  So I am curious why...

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultSaveOrUpdateEventListener.java
```diff
@@ -99,8 +99,7 @@ protected Serializable performSaveOrUpdate(SaveOrUpdateEvent event) {
 	}
 
 	protected Serializable entityIsPersistent(SaveOrUpdateEvent event) throws HibernateException {
-		final boolean traceEnabled = LOG.isTraceEnabled();
```
gsmet: I understand why we did it when there is a static field. I'm curious though on this pattern here?

If it doesn't save us anything, I'm all for consistency but it seems to be a bit different so I thought I might as well ask.

johnaohara:
@sebersole cached log levels cause an issue when we use Hibernate in a native build of Quarkus. The application bootstrap process is split into 2 phases. First phase is at build time where the application is bootstrapped and static fields are initialized. Again at application runtime, where runtime log levels can be redefined.  If the log level is different at runtime from build time, then the incorrect level is cached in the static fields, causing logs to be generated that then get discarded, this creates a memory overhead of logs generated at the wrong level.

I spoke to David Lloyd about the if those log level checks, in jboss logging, should get inlined by the JVM and we came to the conclusion that those method calls should be inlined. I verified that reverting those caching changes do not cause performance degredation in 2 benchmarks.  This would fix a logging issue in Quarkus, and jboss logging now appears to be optimised to allow those calls to be inlined.

sebersole:
Thanks for the explanation @johnaohara - makes sense

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultSaveOrUpdateEventListener.java
johnaohara: I made those changes for consistency. I searched through the code base for all occurrences where we assign to a variable as the call gets inlined there didn't seem much sense to store in a variable. I can revert those changes if needed

johnaohara: I made those changes for consistency. I searched through the code base for all occurrences where we assign to a variable as the call gets inlined there didn't seem much sense to store in a variable. I can revert those changes if needed

johnaohara: I made those changes for consistency. I searched through the code base for all occurrences where we assign to a variable as the call gets inlined there didn't seem much sense to store in a variable. I can revert those changes if needed

PR #2896, opened 2019-06-05
HHH-13426 Criteria on custom user type does not work with applied filter

aonnikov:
https://hibernate.atlassian.net/browse/HHH-13505
This resolves ClassCastException that occurs in QueryParameters class when criteria contains filtering on a composite user type and there is a filter applied.


aonnikov:
CI is green but still displayed as pending in the pull request.

aonnikov:
closed as no one seem to be interested in resoving this bug

gavinking:
> closed as no one seem to be interested in resoving this bug

Well, I mean, in fairness it's a bug relating to (1) an API which was _completely removed_ in H6, and (2) another API, `CompositeUserType`, which was substantially redesigned.

Now, if you could show that the bug still exists in H6, I assume there would be substantially more interest in it.

PR #2897, opened 2019-06-05
HHH-11147 handleWrite does not hit the db

dreab8:
@sebersole @gbadner @Sanne apart the need for code refactoring and for a better name for the new methods, do you think the changes 1c6ccf0 have any sense? Basically what I'm trying to do is to avoid to hit the db for EnhancementAsProxyLazinessInterceptor#handleWrite() (it seems it is what happen for LazyAttributeLoadingInterceptor).

I added a test 4b70c8c but probably I'm missing a lot of cases that can fail with this change.

as a side effect it seems also the failing test i added a couple of days ago 4b70c8c works with the change.


dreab8:
yes you are right avoiding the select will always cause the update (and I haven't think about that), it just seemed to me this was the behaviour with the proxy but I may be wrong, not really an expert on this part :( Tomorrow I will try to doublecheck and add few more tests.

sebersole:
Sorry, that's not what I was trying to imply.  

I had thought you meant that you were seeing unnecessary UPDATE statements.  But I see now you probably just meant that calling a setter does not trigger initializing the proxy.  Yes, that should happen.  Though only least the lazy group (and the base group) to which the "writing attribute" belongs.

dreab8:
@sebersole you are right, my change looks wrong.

sebersole:
Why do you say it is wrong?  What exactly is wrong?

dreab8:
I believed that with an HibernateProxy a call to a setter method should not trigger a query but it seems I was wrong.

sebersole:
Right, it should do the select

PR #2901, opened 2019-06-11
massive suite of enhancements to Dialects, dates/times, query language

gavinking:
This pull request contains everything from #2879 and #2863 along with a bunch of additional work, including bugfixes, some query language enhancements, and cleanups and improvements to the `Dialect`s.

In addition to the many enhancements already documented in #2879 and #2863, this branch adds the following improvements:

## Changes to HQL

- add support for `/* … */` style comments to HQL
- add support for queries with no `from` clause, extremely useful for testing!
- allow `select` clause to come after `from` and before `order by`
- add support for literal enum values, for example `test.hql.Sex.FEMALE` (not sure if this is a bugfix or a whole new feature)
- add “standard” support for ANSI `overlay()` function to HQL
- introduce a “standard” `pad()` function to HQL, with ANSI-`trim()`-like syntax
- make `cast(arg as Boolean)` where `arg` is a string or integer work on the databases I can test on
- make `cast(arg as String)` where `arg` is boolean work on the databases I can test on
- a lot of testing to make sure `cast()` works somewhat reliably

## Changes to `Dialect`s

- rewrite dialects for Derby and Firebird, removing inheritance from parent `Dialect`, and adding support for all new functionality
- added `CoackroachDialect` for CockroachDB
- remove use of `Dialect` subclasses to represent product versions — the version is now detected automatically (the old versioned `Dialect`s still exist in `@Deprecated` form as empty stubs for backward compatibility)
- automatically generate `check` constraints for columns mapped to Java `boolean` and `enum` types
- map `enum`s to columns of type `TINYINT` where possible (`enum`s now limited to 128 values)
- add support for durations and `format()` to `SpannerDialect`
- factor out sequence support from `Dialect` to a new `SequenceSupport` hierarchy (by analogy with what we already have for identity columns)
- huge cleanup of the mess of `LimitHandler`s
- add a new `RowsLimitHandler` for `Firebird`, `Interbase`, and `TimesTen`
- totally rewrite `SQLServer2005LimitHandler`
- Use `offset`/`fetch` on DB2 11.1+
- automatically guess the JDBC driver class from the JDBC URL and attempt to load it (this isn't necessary for modern JDBC drivers which load as services, but it doesn't hurt, and it helps with older drivers)
- improve some startup-time logging and error reporting
- replaced "varargs" function templates with support for variadic parameters in pattern function templates
- for certain important JPA-standard functions, obtain the pattern from the `Dialect` itself, to reduce the number of custom function classes
- clean up handling of `LONGVARCHAR`, `LONGNVARCHAR`, and `LONGVARBINARY`
- handle limit on `VARCHAR` column lengths in Sybase Anywhere
- add pooled sequence support for Firebird
- many bugfixes and minor improvements

ASIDE: I love how there’s no longer any good reason to need any configuration property except for `javax.persistence.jdbc.url` when connecting to the major databases. I use something like:

    <property name="javax.persistence.jdbc.url" 
             value="jdbc:mysql://localhost/test?user=root;password=secret”/>

… and I’m all set, even for old-fashioned JDBC drivers which don’t autoload. This was mostly already working, but it’s working even better now.

## Changes to date/time work

- finish support for mapping `OffsetDateTime` to `timestamp with time zone` (this is the work that required merging my two branches) — most databases still don't have real support for `timestamp with time zone`, but for the couple that do, this is nice
- add `current offset datetime`, which evaluates to an `OffsetDateTime`
- some bugfixes to work done in #2863.


gavinking:
Awesome, with today's refactorings, I removed almost a thousand lines of code. Good day.

gavinking:
Actually I removed *more* than 1k LOC.

sebersole:
I am at a point where I need to pull over SqmFunction handling in the new 6.0 branch I told you about a few weeks ago.  I am trying to stay away from pulling in "type changes" because the new branch mostly reverts back to the Persister, etal model - for sure any type changes here have a high likelihood of conflict.  So I am thinking of pulling in just the work in your PRs related to SqmFunction.  Would a direct merge of  #2879 and #2863 be the best place to do that?

sebersole:
So I'll just pull in what I can from #2879 and #2863 and we'll have to decide later how to integrate any other specifics from here.

gavinking:
Hi Steve, I really want to help with this, but I’m on family vacation in Kyrgyzstan until the 9th, with no laptop.

gavinking:
@sebersole I'm back now, but I think the best thing to do would be to chat on the phone about this. Send me an email.

hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
```diff
@@ -506,10 +507,10 @@ protected void registerDateTimeTypeMappings() {
 
 	protected void registerBinaryTypeMappings() {
 		registerColumnType( Types.BINARY, 2000, "raw($l)" );
-		registerColumnType( Types.BINARY, "long raw" );
+		registerColumnType( Types.BINARY, "blob" );
```
carstenartur: Wasn't this already changed in https://hibernate.atlassian.net/browse/HHH-10345?

gavinking: Could be, but note that this is a patch to the Hibernate 6 branch, and that change was made to Hibernate 5.

jordanlewis:
Hi @gavinking, I noticed that this PR in part adds a CockroachDB dialect to Hibernate. I'm really excited to hear this as an engineer at Cockroach Labs! Our users love Hibernate. I'd love to help you out any way that I can, whether that's just info about CockroachDB, help diagnosing issues, or help testing.

Let me know! You can also email me at jordan@cockroachlabs.com if you prefer.

Again, I am super thrilled to see this work - thank you for taking it on (and creating Hibernate in the first place).

sebersole:
As I understand it this is not updated for the current state of the 6.0 branch.  Correct?

gavinking:
Still not but hoping to get it done real soon.

PR #2902, opened 2019-06-11
HHH-13433 EntityManager.find() should only check for roll-back-only condition if there is an active JTA transaction, otherwise ORM should throw convert( e, lockOptions )

scottmarlow:
https://hibernate.atlassian.net/browse/HHH-13433
This change complements the HHH-12472 change, in the following way.  If there isn't an active JTA transaction, we will not throw "IllegalStateException: JPA compliance dictates throwing IllegalStateException when #getRollbackOnly is called on non-active transaction", instead we will throw the exception returned from call to convert(e, lockOptions).


gbadner:
Superseded  by https://github.com/hibernate/hibernate-orm/pull/3100.

PR #2904, opened 2019-06-13
Improved IN clause parameter padding …

m-tommy:
… for dialects with InExpressionCountLimit

When the number of parameters of an IN expression is limited by the dialect, and this limit is not the power of two (e.g. Oracle 1000),
the current implementation does not optimize statements with parameters between the optimization limit and the upper limit of the dialect. (e.g. Oracle 513-1000)
Over time, it can lead to increased memory consumption.

This change closes this gap.


NathanQingyangXu:
This PR seems duplicated with https://github.com/hibernate/hibernate-orm/pull/3468, which has been merged already.

PR #2911, opened 2019-06-14
HHH-11147 - Add test using IdClass

dreab8:


dreab8:
It seems it was applied

PR #2914, opened 2019-06-19
HHH-13442 - CollectionType#getCollection() method improvements (5.3 branch)

dreab8:
https://hibernate.atlassian.net/browse/HHH-13442
Changes to the CollectionType#getCollection()

The CollectionKey is now created only once
Moved the LOG.tracef( "Created collection wrapper: ...) inside the block where the collection wrapper is created


Sanne:
looks like we applied this already.

PR #2915, opened 2019-06-19
HHH-13443: Build failing to parse *.properties file attributes containing trailing space

ahmedmohiduet:
My accidently putting "spring.jpa.hibernate.ddl-auto=update "(*note one trailing space!) in my application-dev.properties failing build, giving me: "Unrecognized legacy `hibernate.hbm2ddl.auto` value : update" whereas real message should be: "Unrecognized legacy `hibernate.hbm2ddl.auto` value : `update `" which may give us a clue we have malformed attributes/properties or we can have a trim before exception conclude

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/tool/schema/Action.java
+++ b/hibernate-core/src/main/java/org/hibernate/tool/schema/Action.java
@@ -150,7 +150,7 @@ public enum Action {
 			return hbm2ddlSetting( (Action) value );
 		}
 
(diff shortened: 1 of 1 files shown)


Sanne:
good idea, thanks! will merge soon.

Sanne:
Merged and applied the same principle in some more places.

ahmedmohiduet:
Thank you Sanne :tada: !

PR #2916, opened 2019-06-19
HHH-13444 Remove ignored EntityMode field from CollectionKey

Sanne:
https://hibernate.atlassian.net/browse/HHH-13444


dreab8:
It look good to me, thanks @Sanne

PR #2917, opened 2019-06-20
HHH-13447 Minimize number of EventListenerRegistry lookups within a S…

Sanne:
…ession use

This registry lookup happens quite often within the work of a Session. Let's keep it in a field.

- https://hibernate.atlassian.net/browse/HHH-13447


Sanne:
merged

PR #2919, opened 2019-06-20
HHH-13446 - Validate mapped-by values got from annotations in bytecode enhancers

barreiro:


Sanne:
hey @barreiro , thanks! Could you add a test case?

barreiro:
@Sanne I have tried, but unfortunately isn't straight forward. The behavior is different between javassist and bytebuddy enhancers.

Sanne:
Could you add one just for byte-buddy? Javassist is deprecated and will be removed soon.

barreiro:
I could, but why not fix both when that fix is ready !?

Sanne:
> I could, but why not fix both when that fix is ready !?

Because we prefer having some protection against regressions. You mentioned you tried to create a test, so I assume you have something that could be adjusted fairly easily?

But also, there is no rush in merging as we're not releasing today, so you have a little more time to tweak your tests.

PR #2920, opened 2019-06-21
HHH-13442 - CollectionType#getCollection() method improvements

dreab8:
https://hibernate.atlassian.net/browse/HHH-13442

inside the `CollectionType#getCollection()` the `CollectionKey` is created twice, the first time inside the `LoadContext#locateLoadingCollection`.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/loading/internal/LoadContexts.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/loading/internal/LoadContexts.java
@@ -159,20 +159,20 @@ public class LoadContexts {
 	}
 
 	/**
-	 * Attempt to locate the loading collection given the owner's key.  The lookup here
+	 * Attempt to locate the loading collection given the CollectionKey obtained from the owner's key.  The lookup here
 	 * occurs against all result-set contexts...
 	 *
 	 * @param persister The collection persister
-	 * @param ownerKey The owner key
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/engine/loading/internal/LoadContexts.java
```diff
@@ -163,16 +163,16 @@ public CollectionLoadContext getCollectionLoadContext(ResultSet resultSet) {
 	 * occurs against all result-set contexts...
 	 *
 	 * @param persister The collection persister
-	 * @param ownerKey The owner key
+	 * @param key The Collection key
 	 * @return The loading collection, or null if not found.
 	 */
-	public PersistentCollection locateLoadingCollection(CollectionPersister persister, Serializable ownerKey) {
-		final LoadingCollectionEntry lce = locateLoadingCollectionEntry( new CollectionKey( persister, ownerKey ) );
+	public PersistentCollection locateLoadingCollection(CollectionPersister persister, CollectionKey key) {
```
Sanne: the javadoc of the method reads "Attempt to locate the loading collection given the owner's key. "  Is that still valid?

dreab8: Good catch, I'm going to fix it, thanks.

PR #2924, opened 2019-06-25
Remove SessionFactoryImplementor field from Collectionkey

dreab8:
@sebersole @gbadner , based on @Sanne perf tests  seems that this change can improve performances, but I'm not sure it can't cause problems.
it seems to me that for `o.h.type.Type#getHashCode()` and `o.h.type.Type#isEqual` the `SessionFactoryImplementor` is useful only for an `EntityType` and that a `CollectionKey#keyType` cannot be an `EntityType`, but I may be wrong.

If you all think that it is fine I will open a Jira.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/CollectionKey.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/CollectionKey.java
@@ -25,16 +25,10 @@ public final class CollectionKey implements Serializable {
 	private final String role;
 	private final Serializable key;
 	private final Type keyType;
-	private final SessionFactoryImplementor factory;
 	private final int hashCode;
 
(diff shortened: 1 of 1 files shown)


Sanne:
This relies on the assumption that `org.hibernate.type.EntityType#isEqual` will never be invoked from a `CollectionKey`.

The assumption seems valid according to our unit tests.. but is it a solid idea? Would love to hear about this from @sebersole and @gbadner if they have time.

I would suggest we merge this but keep it in master only.

IMO the better solution is to get rid of the method
```isEqual(Object x, Object y, SessionFactoryImplementor factory)``` by refactoring method ```org.hibernate.type.EntityType#getAssociatedEntityPersister``` to not use lazy initialization, but lookup the `associatedEntityPersister` during constructors.

Unfortunately that's changing a bunch of signatures, not a good time to do so.

PR #2925, opened 2019-06-25
HHH-13462 Introduce a fastpath for SessionImpl#fireLoad to be used by…

Sanne:
… internal loops

https://hibernate.atlassian.net/browse/HHH-13462

@gbadner @sebersole @dreab8  this patch introduces a copy of `void SessionImpl#fireLoad(LoadEvent event, LoadType loadType)`, which skips these 3 operations:

     checkOpenOrWaitingForAutoClose();
     checkTransactionSynchStatus();
     delayedAfterCompletion();

I called the new method `void fireLoadNoChecks(LoadEvent event, LoadType loadType)` , this should make it much more effective when invoked from hot loops.

I do believe it's safe to skip these "checks" from `internalLoad` and from `immediateLoad`. Could you have a thought on that please?

Also: if you know of any other method which could be a good candidate to invoke this method safely, please let me know.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
@@ -505,7 +505,7 @@ public abstract class AbstractSharedSessionContract implements SharedSessionCont
 
 	@Override
 	public boolean isConnected() {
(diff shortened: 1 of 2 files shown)


Sanne:
discussed this further with @dreab8 , then merged.

PR #2927, opened 2019-06-26
HHH-13465 Allow inlined access to the PersistenceContext for internal…

Sanne:
… methods

See also branch "HHH-13465-5.3" for the backport.

This patch is rather large; best is to just see I added this new method:
 - https://github.com/hibernate/hibernate-orm/blob/e0ba73e1f7ddf95264831690ebec6b38114227bc/hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java#L514-L526

And then moved most - but not all - usages of internal code to the lighter method.

There are more methods using the new method than the old ones; so it migth be easier to just check that (and if) the remaining use cases are the ones you'd expect to still use the existing method.


Sanne:
rebased and merged.

PR #2929, opened 2019-06-26
Fix JavaModulePathIT when testing with javassist

yrodiere:
Works when executing the tests locally with JDK11 with this command:

```
./gradlew :hibernate-integrationtest-java-modules:test --stacktrace --info -Dhibernate.bytecode.provider=javassist
```


dreab8:
Thanks @yrodiere applied to master

PR #2930, opened 2019-06-27
HHH-11147 : fixes batch fetching bytecode-enhanced proxies

gbadner:
The fix is a little ugly. It tries to avoid adding, removing, and re-adding the same EntityKey to BatchFetchQueue#batchLoadableEntityKeys.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
@@ -394,7 +394,19 @@ public class StatefulPersistenceContext implements PersistenceContext {
 	public void addEntity(EntityKey key, Object entity) {
 		entitiesByKey.put( key, entity );
 		if( batchFetchQueue != null ) {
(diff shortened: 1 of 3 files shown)


gbadner:
I just added a test that shows that Hibernate is not throwing ObjectNotFoundException when a proxy for an entity that does not exist is initialized; instead there's a failure when flushing the session; Hibernate thinks the entity is stale.

I don't think it's related to batching, so it probably doesn't belong in BatchFetchProxyTest. Feel free to move it into a new test.

This really should be fixed either here or in a separate PR though.

dreab8:
I think it is better to create another PR for the failing test. I'm going to do it

dreab8:
@gbadner @Sanne @sebersole 
The 2 PRs resulting from splitting this PR:

- https://github.com/hibernate/hibernate-orm/pull/2931 (just changes related with the batch)

- https://github.com/hibernate/hibernate-orm/pull/2932 ( tests related with the loading a non existing id)

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/2931 and by https://github.com/hibernate/hibernate-orm/pull/2932

PR #2931, opened 2019-06-27
HHH-11147 : fixes batch fetching bytecode-enhanced proxies

dreab8:
This is the Gail's https://github.com/hibernate/hibernate-orm/pull/2930 without the failing tests related with the initialisation of an non existing entity.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
@@ -299,6 +299,21 @@ public class DefaultLoadEventListener implements LoadEventListener {
 					event.getSession()
 			);
 
+			// IMPLEMENTATION NOTE:
+			// The entity's EntityKey may already have been added to the BatchFetchQueue
+			// by ManyToOneType#scheduleBatchLoadIfNeeded.
+			// In the case that the entity is loaded by Session#load or Session#getReference,
+			// it will not be in the BatchFetchQueue, so go ahead and add it.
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
```diff
@@ -299,6 +299,21 @@ private Object proxyOrLoad(
 					event.getSession()
 			);
 
+			// IMPLEMENTATION NOTE:
+			// The entity's EntityKey may already have been added to the BatchFetchQueue
+			// by ManyToOneType#scheduleBatchLoadIfNeeded.
+			// In the case that the entity is loaded by Session#load or Session#getReference,
+			// it will not be in the BatchFetchQueue, so go ahead and add it.
+			persistenceContext.getBatchFetchQueue().addBatchLoadableEntityKey( keyToLoad );
```
Sanne: I suppose this section should also check for `org.hibernate.persister.entity.EntityPersister#isBatchLoadable` ?

sebersole: Why?  Its done in `#addBatchLoadableEntityKey` - no need to do it twice

Sanne: oh right.

I got confused, but I was checking something odd: it seems like this patch is able to reduce the number of SQL statements being executed even when batching is not enabled.

Trying to figure out why.. ideas?

Sanne:
merged #2933 instead

PR #2932, opened 2019-06-27
HHH-11147 : Initializing an entity that does not exists

dreab8:
This is the test @gbadner created that shows that Hibernate is not throwing ObjectNotFoundException when a proxy for an entity that does not exist is initialized plus a second test that show the same issue when trying to set a property ;


Sanne:
thanks @gbadner & @dreab8 ! Rebasing and merging.

@gbadner since these additional fixes are all related to HHH-11147 we'll need to wrap them all up and backport. Let's chat about how you prefer we do that?

PR #2933, opened 2019-06-27
Batch fetching support for bytecode proxies on top of HHH-11147

sebersole:
Proposal for alternative approach to Gail and Andrea's work on this - so intended as an alternative to https://github.com/hibernate/hibernate-orm/pull/2931

At a high-level it centers on the addition of 2 methods - 

- org.hibernate.engine.spi.PersistenceContext#addEnhancedProxy
- org.hibernate.bytecode.spi.BytecodeEnhancementMetadata#createEnhancedProxy

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
@@ -292,31 +292,12 @@ public class DefaultLoadEventListener implements LoadEventListener {
 				}
 			}
 
+			// Potentially add a batch-fetch entry into the queue for this entity
+			persistenceContext.getBatchFetchQueue().addBatchLoadableEntityKey( keyToLoad );
+
 			// This is the crux of HHH-11147
 			// create the (uninitialized) entity instance - has only id set
-			final Object entity = persister.getEntityTuplizer().instantiate(
-					keyToLoad.getIdentifier(),
-					event.getSession()
-			);
-
-			// add the entity instance to the persistence context
-			persistenceContext.addEntity(
-					entity,
-					Status.MANAGED,
-					null,
-					keyToLoad,
-					null,
-					LockMode.NONE,
-					true,
-					persister,
-					true
-			);
-
-			persister.getEntityMetamodel()
-					.getBytecodeEnhancementMetadata()
-					.injectEnhancedEntityAsProxyInterceptor( entity, keyToLoad, event.getSession() );
-
-			return entity;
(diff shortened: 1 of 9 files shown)


Sanne:
@sebersole should we create a new HHH issue to track this one?

sebersole:
IMO its not really necessary.  It's all a continuation of developing a not-yet-released feature

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
```diff
@@ -292,31 +292,12 @@ private Object proxyOrLoad(
 				}
 			}
 
+			// Potentially add a batch-fetch entry into the queue for this entity
+			persistenceContext.getBatchFetchQueue().addBatchLoadableEntityKey( keyToLoad );
```
Sanne: Minor nit-picking: invoking `getBatchFetchQueue()` will create a queue if there isn't one already. This object is quite large, so we should try to avoid it when not necessary.

It's possible that batching was not enabled, so we should check `keyToLoad.isBatchLoadable()` before doing this.

Sanne: I have this change locally, I guess I can just push your PR and then add a commit.

dreab8: ok my comment is unesefull, sorry for the noise

sebersole: So then every single call to this method should be changed to this paradigm as well.

sebersole: I wonder if a better option might be to delay creating the Maps inside `BatchFetchQueue`.  That way we do not have to make the same change everywhere this same current paradigm is used - and if we make the change in one place we need to make it in all imo.  The idea being that it is no longer as heavy to just create the `BatchFetchQueue`.

sebersole: The Travis failure is a test failure in `hibernate-osgi`.  Pretty sure that is not related to these changes :(

sebersole: Re-started the Travis build just to see if it was maybe (hopefully) a transient problem

Sanne: > I wonder if a better option might be to delay creating the Maps inside

I am just doing that.. will push soon.

Sanne:
merged, thanks all!

PR #2934, opened 2019-06-29
HHH-10995 Add support for Firebird 2.5 and 3.0

mrotteveel:
Adds new dialects for Firebird 2.5 and Firebird 3.0 (https://hibernate.atlassian.net/browse/HHH-10995)

Minor change to InterbaseDialect for improved sequence support in Interbase and Firebird 2.1 and earlier


gavinking:
Wait wait wait, I did a whole bunch of work on `FirebirdDialect` on my branch.

mrotteveel:
@gavinking By the looks of it, your changes are based on the wip/6.0 branch. Would it be possible to take this in on master, and reworking them for 6.0 later? Or are you looking to get rid of version-specific dialects?

gavinking:
@mrotteveel  

> your changes are based on the wip/6.0 branch.

Yeah, that's right, but I'm thinking of adapting them to the master branch. But I'm kind of in limbo waiting to hear news on what's going to happen with wip/6.0.

> Would it be possible to take this in on master, and reworking them for 6.0 later?

Well, yeah, it's *possible*, and I can see that your work includes some stuff that mine *doesn't*, though there is quite some overlap. But I was hoping to *not* multiply the number of version-specific dialects right at the point where I think they should all be deprecated.

> Or are you looking to get rid of version-specific dialects?

Yes, absolutely; it was always a bad design, though it was a bit justifiable back in the day when some JDBC drivers didn't support JDBC metadata *at all*.

mrotteveel:
@gavinking 

> > Would it be possible to take this in on master, and reworking them for 6.0 later?
> 
> Well, yeah, it's _possible_, and I can see that your work includes some stuff that mine _doesn't_, though there is quite some overlap. But I was hoping to _not_ multiply the number of version-specific dialects right at the point where I think they should all be deprecated.
> 
> And plus we need to take the bits of your work that are missing from mine and merge them into my `FirebirdDialect`.

That sounds reasonable. In that case it might be better to close this PR, and consider it as a source of inspiration for further changes to `FirebirdDialect`. If you need more information or help, let me know.

> > Or are you looking to get rid of version-specific dialects?
> 
> Yes, absolutely; it was always a bad design, though it was a bit justifiable back in the day when some JDBC drivers didn't support JDBC metadata _at all_.

I'm interested how you will solve the issue with version difference. With the current setup of dialects, version differences are hard (or impossible) to do in a single dialect (eg introducing BOOLEAN in Firebird 3, extended NUMERIC/DECIMAL precision and new decimal floating point type (DECFLOAT) in upcoming Firebird 4, new functions per version, etc).

gavinking:
> That sounds reasonable. In that case it might be better to close this PR, and consider it as a source of inspiration for further changes

Well if you don't mind me just copy/pasting some of your work into my own branch I will go ahead and do that. But let's not close this pull request just yet, until we see what the Hibernate team thinks.

> I'm interested how you will solve the issue with version difference. With the current setup of dialects, version differences are hard (or impossible) to do in a single dialect

Sure, but that's one of the issues that is already fixed on my branch.

sebersole:
Well really that was fixed in the work before your branch

mrotteveel:
> > That sounds reasonable. In that case it might be better to close this PR, and consider it as a source of inspiration for further changes
> 
> Well if you don't mind me just copy/pasting some of your work into my own branch I will go ahead and do that. But let's not close this pull request just yet, until we see what the Hibernate team thinks.

Go ahead, copying some of my code is fine by me.
 
> > I'm interested how you will solve the issue with version difference. With the current setup of dialects, version differences are hard (or impossible) to do in a single dialect
> 
> Sure, but that's one of the issues that is already fixed on my branch.

Good to know, I'll take a closer look in the future.

gavinking:
> Well really that was fixed in the work before your branch

To clarify: the difference is that whereas previous to my work there was a `Dialect` subclass for each version of each database, there's now a single `Dialect` subclass per database, and the old version-specific dialects like `Postgres57Dialect` etc are now just empty classes marked `@Deprecated`. Which is why I'm suggesting it would be best not to add `Firebird25Dialect` and `Firebird30Dialect` at this moment in time.

gavinking:
> Good to know, I'll take a closer look in the future.

The most relevant change is in `DialectFactoryImpl.constructDialect()`, which now passes a `DialectResolutionInfo` to the constructor of the `Dialect` (if there is an appropriate such constructor).

gavinking:
@mrotteveel would you please take a look at this code:

<https://github.com/gavinking/hibernate-orm/blob/MERGED/hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java>

which is supposed to be a merge of your work with mine, and tell me if there's anything I missed, or anything you don't think is quite right?

Thanks!!

mrotteveel:
@gavinking Some notes:

- Line 105: Firebird does have `real` as a datatype (it isn't documented in the Firebird Language Reference though), and is a 32 bit single precision. In theory, Firebird has `float(p)`, but its behavior for the precision parameter is undocumented (and not conforming to the SQL specification), so it is normally only used as an alias for `real` (or in practice: `real` is used as an alias for `float`).
- Line 115: I suggest adding `registerColumnType( Types.CHAR, 8_191, "char($l)" )` as well
- Line 112: `TIMESTAMP_WITH_TIMEZONE` support will be introduced in Firebird 4 (as type `TIMESTAMP WITH TIME ZONE`)
- Lines 123-125: Jaybird prefers to handle Firebird blobs (text and binary) as `LONGVARCHAR`/`LONGVARBINARY` and will report it as type information for blob columns. My pull requests contained definitions for this:

      registerColumnType( Types.LONGVARBINARY, "blob sub_type binary" );
      registerColumnType( Types.LONGVARCHAR, "blob sub_type text" );

- Line 182: I suggest using `(0<>?1)` instead of `(?1<>0)` as Firebird versions 2.5 and earlier have problems with type inference with parameters in the left-hand-side of an expression
- Line 204: Timestamp precision is actually 100 microseconds, so `100_000` nanoseconds
- Line 383: The use of `SEQUENCE` instead of `GENERATOR` in `FirebirdSequenceSupport.LEGACY_INSTANCE` will restrict compatibility to Firebird 2.0 and higher. It is my experience that there is a group of Firebird users that are still using Firebird 1.5 for various reasons. It might be a good idea to add a condition on `getVersion() < 200` to return `InterbaseSequenceSupport .INSTANCE` to provide backwards compatibility there.
- Line 416: Using `for update` will force Firebird to return a single row per fetch (its primary purpose is for positioned updates/deletes with named cursors), the use of `for update of <aliases>` doesn't do anything different from a normal `for update` (it is only provided for syntactic compatibility); it might be better to just use plain `with lock` here as well.

gavinking:
> Firebird does have real as a datatype (it isn't documented in the Firebird Language Reference though)

Hrm. But if it's undocumented, should we be generating DDL that uses it?

> In theory, Firebird has float(p), but its behavior for the precision parameter is undocumented (and not conforming to the SQL specification)

What precisely *is* the behavior? Decimal precision instead of binary precision? I can easily accommodate that.

>  I suggest adding `registerColumnType( Types.CHAR, 8_191, "char($l)" )` as well

Well in other dialects we don't bother with the length limit on `CHAR`, since anyway the only real use of `CHAR` in Hibernate is for columns of type `CHAR(1)`.

> `TIMESTAMP_WITH_TIMEZONE` support will be introduced in Firebird 4 (as type `TIMESTAMP WITH TIME ZONE`)

Ah, OK, I will change that then.

> Jaybird prefers to handle Firebird blobs (text and binary) as `LONGVARCHAR`/`LONGVARBINARY` and will report it as type information for blob columns.

I think what I've got here is correct as far as DDL generation goes. Look at how `LONGVARCHAR`/`LONGVARBINARY` are treated in `Dialect.getTypeName()`.

>  I suggest using `(0<>?1)` instead of `(?1<>0)` as Firebird versions 2.5 and earlier have problems with type inference with parameters in the left-hand-side of an expression

OK, thanks.

> Timestamp precision is actually 100 microseconds, so 100_000 nanoseconds

Yes, I know that but for `dateadd()` and `datediff()` we want to pass the argument in milliseconds. (Well spotted, BTW).

> The use of `SEQUENCE` instead of `GENERATOR`... will restrict compatibility to Firebird 2.0 and higher. It is my experience that there is a group of Firebird users that are still using Firebird 1.5 for various reasons. It might be a good idea to ... provide backwards compatibility there.

Hrm, OK, you might notice that I documented this as supporting 2.5 and above, since I has understood that previous versions were EOL. But if this is the only change we need to make in order to support older versions, I might as well make it.

> it might be better to just use plain `with lock` here as well.

OK. I will make that change.

mrotteveel:
@gavinking 

> > Firebird does have `real` as a datatype (it isn't documented in the Firebird Language Reference though)
> 
> Hrm. But if it's undocumented, should we be generating DDL that uses it?

Good point. I thought it had originally been documented in the InterBase 6 documentation and had been omitted when the Firebird 2.5 language reference was written, but the only trace I have been able to find in the old documentation was for REAL in the list of reserved words.
 
> > In theory, Firebird has `float(p)`, but its behavior for the precision parameter is undocumented (and not conforming to the SQL specification)
> 
> What precisely _is_ the behavior? Decimal precision instead of binary precision? I can easily accommodate that.

Yes, it applies decimal precision (switching from 32 bit precision to 64 bit precision with precision >= 8), but as that behavior has never been documented (that I'm aware of), I feel hesitant to rely on it (especially as it isn't standard compliant to begin with). Using `FLOAT` (or `REAL`) vs `DOUBLE PRECISION` will achieve the same effect.

> > I suggest adding `registerColumnType( Types.CHAR, 8_191, "char($l)" )` as well
> 
> Well in other dialects we don't bother with the length limit on `CHAR`, since anyway the only real use of `CHAR` in Hibernate is for columns of type `CHAR(1)`.

Ok!

[..]

> > Jaybird prefers to handle Firebird blobs (text and binary) as `LONGVARCHAR`/`LONGVARBINARY` and will report it as type information for blob columns.
> 
> I think what I've got here is correct as far as DDL generation goes. Look at how `LONGVARCHAR`/`LONGVARBINARY` are treated in `Dialect.getTypeName()`.

Ok, that will indeed work.

[..]

> > The use of `SEQUENCE` instead of `GENERATOR`... will restrict compatibility to Firebird 2.0 and higher. It is my experience that there is a group of Firebird users that are still using Firebird 1.5 for various reasons. It might be a good idea to ... provide backwards compatibility there.
> 
> Hrm, OK, you might notice that I documented this as supporting 2.5 and above, since I had understood that previous versions were EOL. But if this is the only change we need to make in order to support older versions, I might as well make it.

Hmm, that is a good point. The usage of global temporary table by `getDefaultIdTableStrategy` will limit that to Firebird 2.1 or higher. It might be better to make a clean break here and just stop supporting versions before Firebird 2.5. Firebird 2.5 was just EOL'd last month with the release of Firebird 2.5.9, so people stuck on even older versions may just have to keep using older Hibernate versions or use the InterbaseDialect as workaround. As far as I can tell, the InterbaseDialect should still support InterBase 6 (which was the point where Firebird was forked), so should still work with current Firebird versions with the loss of support of newer features/types.

[..]

gavinking:
> Good point. I thought it had originally been documented in the InterBase 6 documentation and had been omitted when the Firebird 2.5 language reference was written, but the only trace I have been able to find in the old documentation was for REAL in the list of reserved words.

And the only documentation for 3.0 is in Russian (FTR, it defines `FLOAT` as 32 bits, and `DOUBLE PRECISION` as 64 bits, and doesn't mention `REAL`).

> Yes, it applies decimal precision (switching from 32 bit precision to 64 bit precision with precision >= 8), but as that behavior has never been documented (that I'm aware of), I feel hesitant to rely on it (especially as it isn't standard compliant to begin with).

But from what you're saying, Firefox is actually quite a lot closer to the ANSI standard behavior than I had understood from the docs. I'm really not sure what to do here; I mean I've migrated to use `float(n)` on almost all the other databases, since it seems more faithful to what the ANSI standard intends. *Shrug*. Not sure what's best here.

> The usage of global temporary table by `getDefaultIdTableStrategy()` will limit that to Firebird 2.1 or higher.

Well again, if we're *only* talking about changing `getDefaultIdTableStrategy()` and `getSequenceSupport()` in order to support previous versions, that's almost trivial. But is there anything *else* that would need to change?

mrotteveel:
> > Good point. I thought it had originally been documented in the InterBase 6 documentation and had been omitted when the Firebird 2.5 language reference was written, but the only trace I have been able to find in the old documentation was for REAL in the list of reserved words.
> 
> And the only documentation for 3.0 is in Russian (FTR, it defines `FLOAT` as 32 bits, and `DOUBLE PRECISION` as 64 bits, and doesn't mention `REAL`).

Yes, there is a Russian company that sponsors writing that documentation, and the translation to English lags behind. For English it is usual a matter of referencing the 2.5 language reference + the Firebird 3 release notes.
 
> > Yes, it applies decimal precision (switching from 32 bit precision to 64 bit precision with precision >= 8), but as that behavior has never been documented (that I'm aware of), I feel hesitant to rely on it (especially as it isn't standard compliant to begin with).
> 
> But from what you're saying, Firefox is actually quite a lot closer to the ANSI standard behavior than I had understood from the docs. I'm really not sure what to do here; I mean I've migrated to use `float(n)` on almost all the other databases, since it seems more faithful to what the ANSI standard intends. _Shrug_. Not sure what's best here.

I think it is probably better to keep it as it is now, with `FLOAT` mapped to `Types.REAL` and and `DOUBLE PRECISION` to `Types.DOUBLE`. It is what most users are expecting anyway.

> > The usage of global temporary table by `getDefaultIdTableStrategy()` will limit that to Firebird 2.1 or higher.
> 
> Well again, if we're _only_ talking about changing `getDefaultIdTableStrategy()` and `getSequenceSupport()` in order to support previous versions, that's almost trivial. But is there anything _else_ that would need to change?

The support for built-in functions in older versions is rather limited (relying on users defining UDFs backed by a native library), but I'll need to double-check if there is anything else.

Firebird 1.5 to Firebird 2 was a big step in SQL support, but had to break some backwards compatibility (hence why some still haven't moved on), but Firebird 2.0, 2.1 and 2.5 were more incremental so most users upgraded along. Firebird 1.5 doesn't support some SQL features like derived tables, so I actually wonder if anything but the most basic Hibernate use actually works on that version.

gavinking:
> I think it is probably better to keep it as it is now, with `FLOAT` mapped to `Types.REAL` and and `DOUBLE PRECISION` to `Types.DOUBLE`. It is what most users are expecting anyway.

Well the thing is that I changed stuff so that Java `float` and `double` both map to JDBC `Types.FLOAT`, with different precisions. So with what I had previously, they both mapped to SQL `DOUBLE PRECISION` on Firebird, which was imperfect. (Note that historically Hibernate never really used `Types.REAL` for anything, and instead abused `Types.FLOAT` to mean single precision.)

So I have gone and removed those overrides on my branch, and made `FirebirdDialect` use `float(p)` like the other dialects, but where `p` is decimal precision.

But it might be a good idea to get in touch with the Firebird team and ask them to commit to supporting the ANSI-style syntax in Firebird 4, given that the actual database already implements it.

mrotteveel:
> But it might be a good idea to get in touch with the Firebird team and ask them to commit to supporting the ANSI-style syntax in Firebird 4, given that the actual database already implements it.

I'll do that.

I looked at the impact on earlier versions:

1. Sequence behavior for versions < 2.0: same as `InterbaseSequenceSupport`
2. `getDefaultIdTableStrategy()` for versions < 2.1 cannot use global temporary tables
3. `getSelectGUIDString()` requires 2.1 or higher
4. `getNoColumnsInsertString()` requires 2.1 or higher (no alternative available, so no impact) 
5. `DATEADD`/`DATEDIFF`  requires 2.1 or higher (several addXXX UDFs are included with Firebird since 1.0 as an alternative, but those aren't installed by default)
6. `supportsCommentOn()` requires 2.0 or higher

I haven't checked the list of built-in functions, but things will get progressively worse for older versions.

gavinking:
> I'll do that.

Excellent, thanks, appreciated!

> I looked at the impact on earlier versions:

Thanks, I've incorporated those differences into `FirebirdDialect`.

Very much appreciated!

mrotteveel:
> > I looked at the impact on earlier versions:
> 
> Thanks, I've incorporated those differences into `FirebirdDialect`.
> 
> Very much appreciated!

And thank you for taking the time for this!

mrotteveel:
@gavinking 
> But it might be a good idea to get in touch with the Firebird team and ask them to commit to supporting the ANSI-style syntax in Firebird 4, given that the actual database already implements it.

Firebird 4 will support the standard `FLOAT` syntax: https://github.com/FirebirdSQL/firebird/commit/426aeb50c81e7d332868e66a84604a99c18b5e74

beikov:
Hey @mrotteveel!

I'll close this PR for now since wip/6.0 should be mostly up-to-date with what you proposed. If you'd like to work on Firebird 4 support or improve parts of the new `FirebirdDialect` I'm open to review and integrate that on the wip/6.0 branch, but I don't think it's a good idea to add two new Dialects that will be removed/deprecated in 6.0.

For your information, we are currently discussing the way forward for Dialects that we don't maintain https://github.com/hibernate/hibernate-orm/discussions/3783

mrotteveel:
@beikov  No problem. I left a comment on that discussion with my perspective, and I'll take another look at wip/6.0 now that Firebird 4 is nearing completion.

PR #2935, opened 2019-07-03
HHH-13466 : ClassCastException when changing a Collection association to a Set if @PreUpdate listener exists

gbadner:
https://hibernate.atlassian.net/browse/HHH-13466

The bug happens because an entity's reference to a `PersistentCollection` was modified to refer to a `HashSet`. 

During flush, the pre-update listener is executed, `DefaultFlushEntityEventListener#copyState` detects a dirty collection because of [this block](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultFlushEntityEventListener.java#L377-L382), specifically this line:

`( state[index] != newState[index] && !types[index].isEqual( state[index], newState[index] ) )`

`state[index]` refers to the wrapped collection (`PersistentBag`);
`newState[index]` refers to the unwrapped collection (a `Set`), read from the entity after the pre-update listener executes.

The reason why `types[index].isEqual( state[index], newState[index] ) == false ` is because, in `CollectionType#isEqual`,  `( (PersistentCollection) x ).isWrapper( y ) ) == false`. 

`PersistentBag#isWrapper` returns `false` because it doesn't wrap the `Set`; instead it wraps a `List` with the elements that were extracted from the `Set` by the `PersistentBag` constructor.

The fix adds a new field to `PersistentBag`, `providedCollection`. which is explicitly provided to a `PersistentBag` constructor. This is the collection that is directly accessible to the application, so it makes sense that, if `providedCollection` is not `null`, then that is the `Collection` for which `PersistentBag#isWrapper` should return `true`, even if it was a `Set` that was provided.

Without this change, `FlushVisitor#processCollection` throws the CCE when it tries to cast the `Set` to a `PersistentCollection`.

I tried to find a way to look up the `PersistentCollection` from `FlushVisitor#processCollection`. Looking up the collection by `CollectionKey` won't work because it returns the original, unmodified collection. I also looked at iterating `PersistentContext#getCollectionEntries` values, but, its not possible to locate the `PersistentCollection` because `PersistentCollection.isWrapper( collection )` returns `false`, for the same reason as above.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
@@ -32,7 +32,11 @@ import org.hibernate.type.Type;
  */
 public class PersistentBag extends AbstractPersistentCollection implements List {
 
+	// TODO: Why is this.bag protected? Can it be changed to private?
 	protected List bag;
+
+	// The Collection provided to a PersistentBag constructor,
+	private Collection providedCollection;
 
 	/**
 	 * Constructs a PersistentBag.  Needed for SOAP libraries, etc
@@ -70,6 +74,7 @@ public class PersistentBag extends AbstractPersistentCollection implements List 
 	@SuppressWarnings("unchecked")
 	public PersistentBag(SharedSessionContractImplementor session, Collection coll) {
 		super( session );
+		providedCollection = coll;
 		if ( coll instanceof List ) {
 			bag = (List) coll;
 		}
@@ -99,7 +104,12 @@ public class PersistentBag extends AbstractPersistentCollection implements List 
 
 	@Override
 	public boolean isWrapper(Object collection) {
-		return bag==collection;
+		if ( providedCollection == null || providedCollection == bag ) {
+			return bag == collection;
+		}
+		else {
+			return providedCollection == collection;
+		}
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentIdentifierBag.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentIdentifierBag.java
@@ -37,8 +37,12 @@ import org.hibernate.type.Type;
  * @author Gavin King
  */
 public class PersistentIdentifierBag extends AbstractPersistentCollection implements List {
+	// TODO: why are values and identifiers protected? Can they be changed to private?
 	protected List<Object> values;
 	protected Map<Integer, Object> identifiers;
+
+	// The Collection provided to a PersistentIdentifierBag constructor,
+	private Collection providedValues;
 
 	/**
 	 * Constructs a PersistentIdentifierBag.  This form needed for SOAP libraries, etc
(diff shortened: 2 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
```diff
@@ -99,7 +104,12 @@ public PersistentBag(SessionImplementor session, Collection coll) {
 
 	@Override
 	public boolean isWrapper(Object collection) {
-		return bag==collection;
+		if ( providedCollection == null || providedCollection == bag ) {
+			return bag == collection;
+		}
+		else {
+			return providedCollection == collection;
+		}
```
gbadner: Hibernate does not maintain consistency between `bag` and `providedCollection`. I looked at checking to see if the entries were consistent, similar to what is done by `PersistentBag#equalsSnapshot`. As you know, that can be an expensive check. 

In this case `role` is `null`, so in order to do the check, a new defaulted SPI method would need to be added that provides the role. 

I was thinking that if they were not consistent, then `IllegalStateException` should be thrown. 

I've started making these changes, but it got a bit complicated, and I'd like to get your take on this before spending more time on this.

@Steve, WDYT? Should the consistency check be done?

gbadner: In retrospect, ` || providedCollection == bag` can be dropped from the if-statement.
`

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
```diff
@@ -32,8 +32,12 @@
  */
 public class PersistentBag extends AbstractPersistentCollection implements List {
 
+	// TODO: Why is this.bag protected? Can it be changed to private?
 	protected List bag;
 
+	// The Collection provided to a PersistentBag constructor,
```
sebersole: Not a huge fan of storing this - but I guess this gets to the overall solution you are trying.    See my general review comment...

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
sebersole: > Hibernate does not maintain consistency between `bag` and `providedCollection`

You just added `providedCollection`, so yeah Hibernate would not be doing anything with it.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
```diff
@@ -32,8 +32,12 @@
  */
 public class PersistentBag extends AbstractPersistentCollection implements List {
 
+	// TODO: Why is this.bag protected? Can it be changed to private?
```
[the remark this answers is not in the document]
gbadner: OK, I'll remove the TODO. I don't want to break things now by making it private.

gbadner:
@sebersole, I'm going to close this and come up with a better solution.

gbadner:
Reopening because my other solutions were too complicated.

gbadner:
Superseded by https://github.com/hibernate/hibernate-orm/pull/2957

PR #2937, opened 2019-07-07
HHH-11888 : Different Behavior on insertable of @JoinColumn of @OneToMany field between Hibernate4/Hibernate5

sumita-mudgil:
…Many field between Hibernate4/Hibernate5

Different Behavior on "insertable" of "@JoinColumn" of "@OneToMany" field between Hibernate4/Hibernate5
First the 'joincolumn' was added to 'SampleValue' with default insertable/updatable as true and later while linking it with sampleValue, it was added again with overridden insertable and updatable values and hence the exception "Same column is added more than once with different values for isInsertable" was being thrown as it was already added with different values.
With this fix, column is added to sampleValue only once with correct insertable and updatable values

https://hibernate.atlassian.net/browse/HHH-11888


gavinking:
Not entirely sure what is being claimed here but:

- the suggested change does not apparently "fix" the linked issue
- the PR itself does not come with a test case
- the proposed change breaks the ordering of columns in the PK constraint of one of our tests

Closing this,

PR #1521, opened 2016-08-10
HHH-11032 performance improvements of PersistentBag.equalsSnapshot

nguyenfilip:

PR #2938, opened 2019-07-11
HHH-11032 : performance improvements of PersistentBag.equalsSnapshot

gbadner:
https://hibernate.atlassian.net/browse/HHH-11032

I added a commit to @nnguyenfilip and @barreiro's [commit](https://github.com/barreiro/hibernate-orm/commit/43c5f2ddbf7110fa4ca2326b657b663776ce7e64) that may improve performance:

* reduces the number of hashCode calculations;
* reduces the number of ```Map``` lookups;
* adds a check to return ```false``` early, if number of instances with the same hash is different.

I also made some variables final, and renamed some variables for readability.

Supersedes https://github.com/hibernate/hibernate-orm/pull/1521

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
@@ -11,7 +11,6 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
@@ -62,7 +61,7 @@ public class PersistentBag extends AbstractPersistentCollection implements List 
 	 */
 	@Deprecated
 	public PersistentBag(SessionImplementor session) {
-		this( ( SharedSessionContractImplementor) session );
+		this( (SharedSessionContractImplementor) session );
 	}
 	/**
 	 * Constructs a PersistentBag
@@ -144,20 +143,37 @@ public class PersistentBag extends AbstractPersistentCollection implements List 
 
 		// HHH-11032 - Group objects by Type.getHashCode() to reduce the complexity of the search
 		final Map<Integer, List<Object>> hashToInstancesBag = groupByEqualityHash( bag, elementType );
-		final Map<Integer, List<Object>> HashToInstancesSn = groupByEqualityHash( sn, elementType );
-		if ( hashToInstancesBag.size() != HashToInstancesSn.size() ) {
+		final Map<Integer, List<Object>> hashToInstancesSn = groupByEqualityHash( sn, elementType );
+		if ( hashToInstancesBag.size() != hashToInstancesSn.size() ) {
 			return false;
 		}
 
+		// First iterate over the hashToInstancesBag entries to see if the number
+		// of List values is different for any hash value.
 		for ( Map.Entry<Integer, List<Object>> hashToInstancesBagEntry : hashToInstancesBag.entrySet() ) {
-			final int hash = hashToInstancesBagEntry.getKey();
+			final Integer hash = hashToInstancesBagEntry.getKey();
 			final List<Object> instancesBag = hashToInstancesBagEntry.getValue();
-			final List<Object> instancesSn = HashToInstancesSn.getOrDefault( hash, Collections.emptyList() );
-			if ( instancesBag.size() != instancesSn.size() ) {
+			final List<Object> instancesSn = hashToInstancesSn.get( hash );
+			if ( instancesSn == null || ( instancesBag.size() != instancesSn.size() ) ) {
 				return false;
 			}
+		}
+
+		// We already know that both hashToInstancesBag and hashToInstancesSn have:
+		// 1) the same hash values;
+		// 2) the same number of values with the same hash value.
+
+		// Now check if the number of occurrences of each element is the same.
+		for ( Map.Entry<Integer, List<Object>> hashToInstancesBagEntry : hashToInstancesBag.entrySet() ) {
+			final Integer hash = hashToInstancesBagEntry.getKey();
+			final List<Object> instancesBag = hashToInstancesBagEntry.getValue();
+			final List<Object> instancesSn = hashToInstancesSn.get( hash );
 			for ( Object instance : instancesBag ) {
-				if ( countOccurrences( instance, instancesBag, elementType ) != countOccurrences( instance, instancesSn, elementType ) ) {
+				if ( !expectOccurrences(
+						instance,
+						instancesBag,
+						elementType,
+						countOccurrences( instance, instancesSn, elementType ) ) ) {
 					return false;
 				}
 			}
@@ -179,16 +195,26 @@ public class PersistentBag extends AbstractPersistentCollection implements List 
 		return ( (Collection) snapshot ).isEmpty();
 	}
 
-	private int countOccurrences(Object element, List list, Type elementType)
-			throws HibernateException {
-		final Iterator iter = list.iterator();
+	private int countOccurrences(Object element, List<Object> list, Type elementType) {
 		int result = 0;
-		while ( iter.hasNext() ) {
-			if ( elementType.isSame( element, iter.next() ) ) {
+		for ( Object listElement : list ) {
+			if ( elementType.isSame( element, listElement ) ) {
 				result++;
 			}
 		}
 		return result;
+	}
+
+	private boolean expectOccurrences(Object element, List<Object> list, Type elementType, int expected) {
+		int result = 0;
+		for ( Object listElement : list ) {
+			if ( elementType.isSame( element, listElement ) ) {
+				if ( result++ > expected ) {
+					return false;
+				}
+			}
+		}
+		return result == expected;
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
@@ -11,9 +11,13 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.hibernate.HibernateException;
 import org.hibernate.engine.spi.SessionImplementor;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
```diff
@@ -137,34 +136,35 @@ public void beforeInitialize(CollectionPersister persister, int anticipatedSize)
...
 		if ( sn.size() != bag.size() ) {
 			return false;
 		}
 
 		// HHH-11032 - Group objects by Type.getHashCode() to reduce the complexity of the search
-		Map<Integer, List<Object>> instanceCountBag = groupByEqualityHash( bag, elementType );
-		Map<Integer, List<Object>> instanceCountSn = groupByEqualityHash( sn, elementType );
-		if ( instanceCountBag.size() != instanceCountSn.size() ) {
+		final Map<Integer, List<Object>> hashToInstancesBag = groupByEqualityHash( bag, elementType );
+		final Map<Integer, List<Object>> HashToInstancesSn = groupByEqualityHash( sn, elementType );
```
sebersole: Minor - but should be `hashToInstancesSn` rather than `HashToInstancesSn`

Sanne:
The new approach is possibly [1] better for large bags, but it's hugely heavier for smaller bags.

Maybe it should run this version only when `bag.size()` is higher than some figure `N` ?

1 - I say possibly because I didn't test this either - I agree it would be nice to check, and having such a test could help to identify a reasonable figure for `N`.

A question on fundamentals.. are we ok to rely on `:hashCode` on the user types for a bag? I can't think of a specific problem with it, but it feels like the bag so far has been a "safe space" for users in terms of not relying on it.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
```diff
@@ -131,21 +134,46 @@ public void beforeInitialize(CollectionPersister persister, int anticipatedSize)
...
+
+		// HHH-11032 - Group objects by Type.getHashCode() to reduce the complexity of the search
+		final Map<Integer, List<Object>> hashToInstancesBag = groupByEqualityHash( bag, elementType );
+		final Map<Integer, List<Object>> HashToInstancesSn = groupByEqualityHash( sn, elementType );
+		if ( hashToInstancesBag.size() != HashToInstancesSn.size() ) {
+			return false;
+		}
+
+		for ( Map.Entry<Integer, List<Object>> hashToInstancesBagEntry : hashToInstancesBag.entrySet() ) {
+			final int hash = hashToInstancesBagEntry.getKey();
```
Sanne: should use `Integer` rather than `int` so to prevent unboxing and then autoboxing again.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
```diff
@@ -131,21 +134,46 @@ public void beforeInitialize(CollectionPersister persister, int anticipatedSize)
...
+			final List<Object> instancesBag = hashToInstancesBagEntry.getValue();
+			final List<Object> instancesSn = HashToInstancesSn.getOrDefault( hash, Collections.emptyList() );
+			if ( instancesBag.size() != instancesSn.size() ) {
 				return false;
 			}
+			for ( Object instance : instancesBag ) {
+				if ( countOccurrences( instance, instancesBag, elementType ) != countOccurrences( instance, instancesSn, elementType ) ) {
```
Sanne: there's another set of possible optimisations here:

1# don't count on both lists, but rather count on one and then start iterating on the other and see if you can match the same figure. You will be able to quit the loop early, if you reach the figure before ending the loop.

2 # A more tricky to code alternative one would be to not count _neither_ list, and increment the iterator on turns attempting to match the largest figure found so far; essentially you'd also quite early,  but always based on the smallest set of matches - so safe a couple more iterations, but also possibly cheaper to verify each condition.

I'd go with the first proposal as the second is probably too funky: far less readable and I'm not sure it's worth the difference.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
```diff
@@ -131,21 +134,46 @@ public void beforeInitialize(CollectionPersister persister, int anticipatedSize)
...
+		final Map<Integer, List<Object>> hashToInstancesBag = groupByEqualityHash( bag, elementType );
+		final Map<Integer, List<Object>> HashToInstancesSn = groupByEqualityHash( sn, elementType );
+		if ( hashToInstancesBag.size() != HashToInstancesSn.size() ) {
+			return false;
+		}
+
+		for ( Map.Entry<Integer, List<Object>> hashToInstancesBagEntry : hashToInstancesBag.entrySet() ) {
+			final int hash = hashToInstancesBagEntry.getKey();
+			final List<Object> instancesBag = hashToInstancesBagEntry.getValue();
+			final List<Object> instancesSn = HashToInstancesSn.getOrDefault( hash, Collections.emptyList() );
```
Sanne: rather than `getOrDefault` you could just return `false` when the get returns null.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
```diff
@@ -131,21 +134,46 @@ public void beforeInitialize(CollectionPersister persister, int anticipatedSize)
...
-			if ( unequal ) {
+
+		// HHH-11032 - Group objects by Type.getHashCode() to reduce the complexity of the search
+		final Map<Integer, List<Object>> hashToInstancesBag = groupByEqualityHash( bag, elementType );
+		final Map<Integer, List<Object>> HashToInstancesSn = groupByEqualityHash( sn, elementType );
+		if ( hashToInstancesBag.size() != HashToInstancesSn.size() ) {
+			return false;
+		}
+
+		for ( Map.Entry<Integer, List<Object>> hashToInstancesBagEntry : hashToInstancesBag.entrySet() ) {
```
Sanne: it might be worth iterating the entrySet twice: first just check all of `instancesBag.size() != instancesSn.size()` match. The only if they really do match, you do the loop to check the elements.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
gbadner: oops, yes, I'll make that change.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
gbadner: Thanks, I'll make that change.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
gbadner: Oops, yes, I'll make that change.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
gbadner: Yes, I'll make that change.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
gbadner: ```getOrDefault``` was in the @barreiro update to the original commit in ```getItemsWithSameHash```. I thought about changing it, but I kept it, for readability, so I wouldn't need to add a check for ```null``` in the following ```if``` statement. 

I suppose checking for ```null``` would be less expensive than ```size```.

@Sanne, could you respond for future reference? I'm not against changing it.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
gbadner: I'm thinking about this one. The only time this would make a difference is if there are:
 
1) the same element multiple times;
2) many elements with the same hash.

1) could obviously happen for an ```ElementCollection``` or ```@ManyToMany`` `, but I don't know how often it really happens in practice, or if it could result in many or just a few duplicates. For a one-to-many association, it should never happen. This is actually a good area for optimization. I'll address that in a general comment.

2) probably shouldn't happen all that much; if there are few elements with the same hash, and many elements in the bag, it might be costly to do the extra iteration.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
gbadner: For 1#, I assume you meant "You will be able to quit the loop early, if you *exceed* the figure before ending the loop."

That's a good idea. I'll make that change.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
Sanne: No strong opinion but I'd change it:

> I suppose checking for null would be less expensive than size.

+1 and it saves invoking `Collections.emptyList()` as  well.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
Sanne: ok but (2) is actually quite likely if the objects are indeed different: you'd immediately spot some (likely) change in hash distribution.

Of course iterating twice is slightly more expensive than iterating only once, so the case in which all objects are the same would be slightly slower - I suppose not much though as each loop iteration is simpler.

But if there's any difference, you'd spot it much earlier as apparently it's the `isSame` operation which is slow - you'd be skipping _all of them_ which is quite great :)

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
Sanne: exactly! :+1:

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
gbadner: OK, I'll change it.

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
gbadner: ```isSame``` will depend on the element type.

For most types, ```isSame``` delegates to ```isEqual``, which again depends on the type.

For ```EntityType```, ```isSame``` returns ```x == y```.

For ```ComponentType```, ```isSame``` first checks if ```x == y``. If ```false```, ```isSame``` is called for each sub-property.

IOW, calling ```isSame``` for entity elements will be a lot faster than for ```String``` elements.

gbadner: @Sanne, I just want to make sure you see that ```iSame``` performance is highly type-dependent...

Sanne: Sure, I understand and agree.  A consequence is that it might be that `isSame` is very cheap, but it might be expensive.

The comment on https://hibernate.atlassian.net/browse/HHH-11032 seems to suggest that - at least in the use case of the original issue reporter - dodging invoking the `isSame` method as much as possible would be good:

> Since isSame uses reflection, it's quite slow.

(I'm not sure to what reflection specifically he refers, but I guess either way `isSame` is indeed more expensive than the other methods, at least because it's highly type dependent)

gbadner: For associations, ```isSame``` is very cheap; it only involves ```x == y```. I think it is worthwhile to do things differently for associations vs basic/embeddable elements. WDYT?

Sanne: interesting. Ok, send a PR and I'll review it :)

But let's not over think this? At some point we need to balance the fact to reach a quick fix for those who need it, which as far as we know were happy enough already with the original patch. 

We can always improve things in subsequent patches later - and possibly keep a simpler approach for the stable backports, while tinkering with something more aggressive for master?

So far we know:
 - the "old algorythm" allocates less so it should be preferred for small bags (let's say small ~ 10 ?)
 - otherwise use the new strategy which uses buckets by hash, iterate only once (like you said) and yet try to exit the second loop early on (when the count exceeds the expected figure).

I'm not sure how to take advantage of the fact that associations are cheaper, I guess I'll see your proposal :)

gbadner: @Sanne, I'll update the PR tomorrow.

gbadner: @Sanne, I realized that even for non-associations, ```is_same``` will probably mostly be doing reference checks.

I'll make this change as well.

gbadner:
@Sanne, @barreiro, I've pushed a commit with the recommended changes.

Sanne:
Sorry for taking so long to have a look. Looks awesome! Many thanks, merging.

Sanne:
@gbadner I merged in master only. I guess you'll want to backport? Up to you, I'll flag JIRA accordingly.

gbadner:
Thanks @Sanne! I'll create a PR to backport.

PR #2940, opened 2019-07-11
HHH-13496 Performance focused improvements from some static code analysis

Sanne:
hi @sebersole , @dreab8 , this PR is rather large but there's a lot of small performance improvements.

I'd push it already as I'm quite confident about it, but I don't wish it to get in the way of any big refactoring.

Could you just skim through the list of files it touches? I'm happy to apply just a subset of commits if that's more convenient for you.

Thanks

 - https://hibernate.atlassian.net/browse/HHH-13496


hibernate-core/src/main/java/org/hibernate/cfg/CopyIdentifierComponentSecondPass.java
```diff
@@ -79,8 +79,8 @@ public void doSecondPass(Map persistentClasses) throws MappingException {
 		AtomicInteger index = new AtomicInteger( 0 );
 		if ( columnByReferencedName.isEmpty() ) {
 			isExplicitReference = false;
-			for (Ejb3JoinColumn joinColumn : joinColumns) {
-				columnByReferencedName.put( "" + index.get(), joinColumn );
+			for ( Ejb3JoinColumn joinColumn : joinColumns ) {
+				columnByReferencedName.put( "0", joinColumn );
```
dreab8: the index is incremented inside the for loop, if `joinColumns`contains more than 1 `Ejb3JoinColumn` with this change we are not using the correct value for the Map key.

dreab8: I think the change should be `columnByReferencedName.put(  String.valueOf( index.get() ) , joinColumn );`

dreab8:
@Sanne apart for the comment it looks fine to me, not easy to choose only subset of the commits so if for @sebersole it is fine I would apply all the commits.

hibernate-core/src/main/java/org/hibernate/cfg/CopyIdentifierComponentSecondPass.java
Sanne: nice catch. Not sure how I missed the loop.

Sanne: fixed

Sanne:
@sebersole I've been hesitant to push this to not "get in the way" of your larger refactoring, but thinking about it you could just use a "ours/theirs" strategy when rebasing to discard any changes I made which conflicts with any of yours.

So it's easier if I just push this optimistically; if this ever gets in your way let me know I can help rebasing.

Sanne:
merged

PR #2941, opened 2019-07-14
HHH-13133

skis:
Pull request to HHH-13133

### DIFF
--- a/tooling/hibernate-enhance-maven-plugin/src/main/java/org/hibernate/orm/tooling/maven/MavenEnhancePlugin.java
+++ b/tooling/hibernate-enhance-maven-plugin/src/main/java/org/hibernate/orm/tooling/maven/MavenEnhancePlugin.java
@@ -157,7 +157,7 @@ public class MavenEnhancePlugin extends AbstractMojo {
 
 			writeOutEnhancedClass( enhancedBytecode, file );
 
-			getLog().info( "Successfully enhanced class [" + file + "]" );
+			getLog().debug( "Successfully enhanced class [" + file + "]" );
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


tooling/hibernate-enhance-maven-plugin/src/main/java/org/hibernate/orm/tooling/maven/MavenEnhancePlugin.java
```diff
@@ -157,7 +157,7 @@ public boolean doExtendedEnhancement(UnloadedClass classDescriptor) {
 
 			writeOutEnhancedClass( enhancedBytecode, file );
 
-			getLog().info( "Successfully enhanced class [" + file + "]" );
```
Sanne: could you change this to make sure the strings aren't appended when the logger is not set to debug?

Something like

    log = getLog()
   if (log.isDebugEnabled()) { log.debug( .... ) }

Sanne:
Many thanks! Only had a very minor comment, if you could address that I'll merge it right away.

Generally speaking, try to not include unrelated changes in the same commit / same JIRA id.. but it's fine since they are very minor.

Sanne:
merged it. Thanks!

skis:
Thanks for merging Sanne. Strangely I didn't get any notification for the last 4 days.

tooling/hibernate-enhance-maven-plugin/src/main/java/org/hibernate/orm/tooling/maven/MavenEnhancePlugin.java
skis: I will make the change and submit another PR.

Sanne:
@skis no worries, and no need to make another PR as I already changed it!

Thanks again

PR #2942, opened 2019-07-16
HHH-13435 - Able to change tenantId in the same session

sfragata:
I've done one part of this feature

- add tenant identifier to entity key and collection key
- Session can change its tenant identifier (setTenantIdentifier)

It miss

- When using transactions, need be able to, when the object already have this connection, change its schema (connection.setCatalog)
NOTE: I don't think we can do transactions and multi tenant  of type DATABASE (must have one connection or maybe a XA transaction ?? )


Sanne:
hi @sfragata , good job getting this to work, but I'm not convinced that we will want such a feature.

One problem is that I'm not sure if this is a good idea; on JIRA you described using two sessions now and wanting to prefer using the same Session to "benefit from caching" . Could you explain how this would benefit you?

The other problem is that adding more fields to objects like `CollectionKey` is not a good idea; this class is very sensitive as high end systems allocate a huge amount of these so we need to make this object smaller, not larger. Adding a field on it with the current design would make the memory costs much higher.

sebersole:
+1.  This is a change we do not want to make imo

sfragata:
@Sanne with this change, in our project, we need to open a session to one tenant and if I need to use a second one I need to close this one and open a new hibernate session. Cache works by I can't use transactions if I close a session.

This PR is to add the ability to a session change its tenantId when we need, so I can use the same session and the same database connection. The problem is if I only add a new method to change the tenant into session class, the cache won't work, because entity key and collection key don't have the tenant id 
For example a Client table with 2 schemas, S1 and S2, if I start my session with tenant=S1 and I do
`session.get(1)`, I'm going to get `S1.Client where id=1`. If I just change the tenant to S2 (`session.setTenantId("S2")`) when I do session.get(1) it will hit the first level cache instead of execute `S2.Client where id=1`, that's why I've added the tenant to entity and collection keys

Sanne:
@sfragata please clarify some things:

> in our project, we need to open a session to one tenant and if I need to use a second one I need to close this one and open a new hibernate session.

What's the problem to have to open a new Session ? What concerns you?

Also: you don't have to close necessarily close the previous session, it's ok to have as many sessions open as you need. Just make sure they are all closed when you are done with them.

> Cache works...

Caching will only work "within" the same tenant, so I don't understand what is the benefit you hope to obtain to keep the cache across Sessions.

>  ... by I can't use transactions if I close a session.

I don't understand this. A properly setup Transaction Manager should be able to do such things - however it's not clear to me exactly what you are after.

Regarding the caching keys needing the tenant keys: we understand that. We used to have such tenantId fields in the keys - it turned out to be a real problem so we don't want to reintroduce such a feature unless there is a very clear need with no alternatives.

Assuming you do manage to persuade @sebersole and me (unlikely but for completeness) :

one better implementation would be to have two different kinds of keys; the patch would get a bit more complex, especially as you'd need to be very careful in making sure the "key factory" indirection wouldn't also risk becoming a performance liability.

Sanne:
Hi,

if there's nothing else to add I suppose I could close this PR? Did I understand your use case and explained why I don't think it's a good idea?

PR #2945, opened 2019-07-19
HHH-13505 : NullPointerException thrown by StatisticsImpl#getCacheRegionStatistics

gbadner:
https://hibernate.atlassian.net/browse/HHH-13505


scottmarlow:
@gbadner also +1 for back porting this change (to 5.3), as it looks safe.

gbadner:
Thanks @scottmarlow !

PR #2947, opened 2019-07-22
HHH-13500 Fix audit strategy sub-query return incorrect results

Naros:
https://hibernate.atlassian.net/browse/HHH-13500


dreab8:
Thanks @Naros

PR #2950, opened 2019-07-24
HHH-13492 : OptimisticLockException after locking, refreshing, and updating an entity

gbadner:
https://hibernate.atlassian.net/browse/HHH-13492

The problem here is that an ```EntityVerifyVersionProcess``` is registered with the ```EntityEntry``` associated with the entity at the time that the entity is locked.

Later, when the entity is refreshed, that ```EntityEntry``` is replaced by a new one in the ```PersistenceContext```. This makes the ```EntityEntry``` registered with the ```EntityVerifyVersionProcess``` stale.

After the updated entity is pushed to the database by ```EntityUpdateAction#execute```, which calls ```EntityEntry#postUpdate``` to update the version in the entity and the current ```EntityEntry```.

Later, when ```EntityVerifyVersionProcess#doBeforeTransactionCompletion``` executes using the stale ```EntityEntry```, ```OptimisticLockException``` is thrown because the version in the stale ```EntityEntry``` does not match the current version.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/action/internal/EntityIncrementVersionProcess.java
+++ b/hibernate-core/src/main/java/org/hibernate/action/internal/EntityIncrementVersionProcess.java
@@ -19,17 +19,14 @@ import org.hibernate.persister.entity.EntityPersister;
  */
 public class EntityIncrementVersionProcess implements BeforeTransactionCompletionProcess {
 	private final Object object;
-	private final EntityEntry entry;
 
 	/**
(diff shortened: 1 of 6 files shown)


gbadner:
@sebersole, thanks for reviewing.

If this becomes a performance issue, I suppose {{EntityIncrementVersionProcess}}/{{EntityVerifyVersionProcess}} could be updated somehow when a new {{EntityEntry}} is added.

PR #2954, opened 2019-07-25
HHH-13511 Remove interning of aliases in org.hibernate.loader.Default…

Sanne:
…EntityAliases

Being discussed with @dreab8 and @gsmet


Sanne:
fixed and merged

PR #2956, opened 2019-07-25
Hhh 13466 1

gbadner:


hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
```diff
@@ -34,8 +34,12 @@
  */
 public class PersistentBag extends AbstractPersistentCollection implements List {
 
+	// TODO: Why is this.bag protected? Can it be changed to private?
```
sebersole: As we've discussed elsewhere this is a historical pattern for inheritance.  Feel free to change these to use protected getter/setter if you prefer.

gbadner:
I added some code that checks if the "provided" collection is equivalent to the "wrapped" collection. It turns out the code is expensive and pretty worthless, so I'm closing this PR.

Superseded by https://github.com/hibernate/hibernate-orm/pull/2957

PR #2957, opened 2019-07-26
HHH-13466 2

gbadner:
https://hibernate.atlassian.net/browse/HHH-13466

Supersedes: https://github.com/hibernate/hibernate-orm/pull/2956


gbadner:
@sebersole, thanks for reviewing. I agree with what you say; unfortunately, I don't see an alternative.

PR #2958, opened 2019-07-26
HHH-13379 Regression of Instant serialization

yrodiere:
https://hibernate.atlassian.net//browse/HHH-13379

This is for 5.4, but should be backported to 5.3.

See [here](https://hibernate.atlassian.net/browse/HHH-13379?focusedCommentId=105317&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-105317) for an explanation of what went wrong and how this PR fixes the problem.


gbadner:
@yrodiere, I've only gotten as far as running the updated tests without the fix. I found that the following tests don't fail:

- LocalDateTest
- LocalDateTimeTest
- LocalTimeTest
- OffsetTimeTest

I also see that the following were updated in the PR:

- LocalDateJavaDescriptor
- LocalDateTimeJavaDescriptor

Should there be test failures in LocalDateTest and LocalDateTimeTest, or were LocalDateJavaDescriptor and LocalDateTimeJavaDescriptor not affected? If they are not affected, should we keep the current (simpler) code?

The following have 4 failures each.:

- InstantTest
- OffsetDateTimeTest
- ZonedDateTimeTest

yrodiere:
You're right, there's something wrong with `LocalDateTimeTest`. I'm having a look.
`LocalDate` is probably affected by the bug, but only in areas where DST change occurs around midnight. I'll try to come up with some test cases.

yrodiere:
@gbadner I updated the PR:

* I added more tests, in particular for LocalDate and LocalDateTime. I couldn't find a way to reproduce the problem with these types.
* I rephrased comments in tests where HHH-13379 cannot be reproduced, to make it clearer the tests were added "just in case", to check that HHH-13379 was indeed not relevant there.
* I removed the changes to LocalDateJavaDescriptor and LocalDateTimeJavaDescriptor, since the changes were not useful.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ZonedDateTimeJavaDescriptor.java
```diff
@@ -100,12 +111,21 @@ public ZonedDateTime fromString(String string) {
...
+			 * Daylight Saving Time introduces ambiguity in the year/month/day/hour/etc representation once a year
+			 * (on DST end), so conversion must be done using the number of milliseconds since the epoch.
+			 * - around 1905, both methods are equally valid, so we don't really care which one is used.
 			 */
-			return ts.toLocalDateTime().atZone( ZoneId.systemDefault() );
+			if ( ts.getYear() < 5 ) { // Timestamp year 0 is 1900
+				return ts.toLocalDateTime().atZone( ZoneId.systemDefault() );
+			}
+			else {
+				return ts.toInstant().atZone( ZoneId.systemDefault() );
```
gbadner: In [e44b38f716bf397531bbe741efac384f8829aef8](https://github.com/hibernate/hibernate-orm/commit/e44b38f716bf397531bbe741efac384f8829aef8), the following was returned:

```return ZonedDateTime.ofInstant( ts.toInstant(), ZoneId.systemDefault() );```

Is ```ts.toInstant().atZone( ZoneId.systemDefault() )``` equivalent?

gbadner: I see that Javadoc for ```Instant#atZone``` says:

"This method is equivalent to ```ZonedDateTime.ofInstant(this, zone)```."

PR #2960, opened 2019-07-30
HHH-13523 - StatementPreparerImpl should not need to retrieve the JDDCService as often

dreab8:
https://hibernate.atlassian.net/browse/HHH-13523

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/StatementPreparerImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/StatementPreparerImpl.java
@@ -18,6 +18,7 @@ import org.hibernate.boot.spi.SessionFactoryOptions;
 import org.hibernate.engine.jdbc.spi.JdbcServices;
 import org.hibernate.engine.jdbc.spi.SqlExceptionHelper;
 import org.hibernate.engine.jdbc.spi.StatementPreparer;
+import org.hibernate.resource.jdbc.spi.JdbcObserver;
 import org.hibernate.resource.jdbc.spi.LogicalConnectionImplementor;
 
 /**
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/StatementPreparerImpl.java
```diff
@@ -199,11 +205,7 @@ private void setStatementTimeout(PreparedStatement preparedStatement) throws SQL
 	}
 
 	private JdbcServices getJdbcService() {
```
Sanne: do we still need the `getJdbcService()` method ?

dreab8: probably not let me check

hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/StatementPreparerImpl.java
```diff
@@ -37,6 +38,11 @@
 	 */
 	StatementPreparerImpl(JdbcCoordinatorImpl jdbcCoordinator) {
 		this.jdbcCoordinator = jdbcCoordinator;
+		this.jdbcServices = jdbcCoordinator
```
Sanne: this still looks quite "heavy" . I see `JdbcCoordinatorImpl` is the only place invoking this ctor, and it also retrives the `jdbcServices` in its own constructor.. keep it and pass it on directly?

hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/StatementPreparerImpl.java
dreab8: I'll remove it and add another little change

hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/StatementPreparerImpl.java
dreab8: I'll do

Sanne:
rebased & merged

PR #2967, opened 2019-07-31
two optimisations for Statements handling

Sanne:
- https://hibernate.atlassian.net/browse/HHH-13526
 - https://hibernate.atlassian.net/browse/HHH-13528

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
@@ -81,18 +81,16 @@ public class ResourceRegistryStandardImpl implements ResourceRegistry {
 	public void release(Statement statement) {
 		log.tracev( "Releasing statement [{0}]", statement );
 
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
```diff
@@ -81,18 +81,16 @@ public void register(Statement statement, boolean cancelable) {
...
-			final Set<ResultSet> resultSets = xref.get( statement );
-			if ( resultSets != null ) {
-				closeAll( resultSets );
-			}
-			xref.remove( statement );
+			// Keep this at DEBUG level, rather than warn.  Numerous connection pool implementations can return a
+			// proxy/wrapper around the JDBC Statement, causing excessive logging here.  See HHH-8210.
+			log.unregisteredStatement();
 		}
+
```
dreab8: cannot be done the same for `public void release(ResultSet resultSet, Statement statement)` when `statement ! null`

Sanne:
did you have a comment which was deleted? I got a confusing notification

PR #2969, opened 2019-07-31
HHH-13528 Remove call to resourceRegistry.release(ResultSet resultSet…

dreab8:
…, Statement statement) from AbstractLoadPlanBasedLoader#executeLoad()

it should be safe to remove this call in such a case the Resultset should be always associated with the Statement so the next call to release(Statement) should be enough


Sanne:
merged this. Thanks!

PR #2971, opened 2019-07-31
Performance work: candidates for backporting to 5.3

Sanne:
@gbadner @dreab8 @stuartwdouglas FYI these are the commits I'd suggest we backport from master.

Might look large and scary, but if you check individual commits out you'll see they are trivial, so should be a no brainer.

List of JIRAs:
 - https://hibernate.atlassian.net/browse/HHH-13520
 - https://hibernate.atlassian.net/browse/HHH-13521
 - https://hibernate.atlassian.net/browse/HHH-13522
 - https://hibernate.atlassian.net/browse/HHH-13523
 - https://hibernate.atlassian.net/browse/HHH-13524
 - https://hibernate.atlassian.net/browse/HHH-13526
 - https://hibernate.atlassian.net/browse/HHH-13528
 - https://hibernate.atlassian.net/browse/HHH-13529

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/AbstractBatchImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/AbstractBatchImpl.java
@@ -21,6 +21,7 @@ import org.hibernate.engine.jdbc.spi.JdbcServices;
 import org.hibernate.engine.jdbc.spi.SqlExceptionHelper;
 import org.hibernate.engine.jdbc.spi.SqlStatementLogger;
 import org.hibernate.internal.CoreMessageLogger;
+import org.hibernate.resource.jdbc.ResourceRegistry;
 
 /**
  * Convenience base class for implementers of the Batch interface.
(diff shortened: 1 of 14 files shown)


gbadner:
The only change that I really had to do a lot of checking for was this [commit](https://github.com/hibernate/hibernate-orm/pull/2971/commits/eb17563a8326200dbd57d68a7e419081ecc56b6c).

Everything else looked good to me.

My main concern with that commit was that there were 3 possibilities that removing:

```resourceRegistry.release( wrapper.getResultSet(), wrapper.getStatement() );```

could cause a leak:

1. the ```Statement``` was not registered;
2. the ```ResultSet``` was not registered;
3. the ```Statement``` is null.

In all of the above cases, the ```ResultSet``` would not get closed.

By adding AssertionFailure and running unit tests, I've found examples of all 3 of these cases, which are probably bugs.

AFAICT, none of these cases can happen in the one particular place in ```AbstractLoadPlanBasedLoader.java#executeLoad``` where this line is removed.

That's because ```AbstractLoadPlanBasedLoader#getResultSet``` calls ```ResultSetReturnImpl#extract```. which calls ```postExtract``` which registers the ```Statement``` and ```ResultSet```, only if the ```ResultSet``` is not ```null```. AFAICT, ResultSet cannot be ```null``` in this case.

Please be very, very careful about moving it from other places.

gbadner:
@Sanne, I'll approve the PR to this point. Do you plan to add more commits to this?

Sanne:
Thanks @gbadner ! I'll merge then.

There are likely more improvements coming, but let's keep them in their own PR - that should help to separate the commits you already reviewed from any future ones.

PR #2974, opened 2019-08-03
HHH-13249 Log a slow SQL query

batbileg289:
Logs queries which executed slower than specified milliseconds in the configuration. Disabled by default.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
@@ -1729,7 +1729,7 @@ public interface AvailableSettings extends org.hibernate.jpa.AvailableSettings {
 	/**
 	 * Setting that logs query which executed slower than specified milliseconds. Default is 0 (disabled).
 	 */
-	String LOG_SLOW_QUERY = "hibernate.session.events.log.slow_query";
+	String LOG_SLOW_QUERY = "hibernate.session.events.log.LOG_QUERIES_SLOWER_THAN_MS";
 
 	/**
 	 * Defines a default {@link org.hibernate.SessionEventListener} to be applied to opened Sessions.
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
@@ -1725,6 +1725,11 @@ public interface AvailableSettings extends org.hibernate.jpa.AvailableSettings {
 	 * is enabled logging of Session metrics is enabled by default too.
 	 */
 	String LOG_SESSION_METRICS = "hibernate.session.events.log";
+	
+	/**
+	 * Setting that logs query which executed slower than specified milliseconds. Default is 0 (disabled).
+	 */
+	String LOG_SLOW_QUERY = "hibernate.session.events.log.slow_query";
 
 	/**
 	 * Defines a default {@link org.hibernate.SessionEventListener} to be applied to opened Sessions.
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/ResultSetReturnImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/ResultSetReturnImpl.java
@@ -54,8 +54,9 @@ public class ResultSetReturnImpl implements ResultSetReturn {
 	public ResultSet extract(PreparedStatement statement) {
 		// IMPL NOTE : SQL logged by caller
 		long executeStart = 0;
-		if ( this.sqlStatementLogger.getLogSlowQuery() > 0 )
+		if ( this.sqlStatementLogger.getLogSlowQuery() > 0 ) {
 			executeStart = System.currentTimeMillis();
+		}
 		try {
 			final ResultSet rs;
 			try {
@@ -86,8 +87,9 @@ public class ResultSetReturnImpl implements ResultSetReturn {
 	public ResultSet extract(CallableStatement callableStatement) {
 		// IMPL NOTE : SQL logged by caller
 		long executeStart = 0;
-		if ( this.sqlStatementLogger.getLogSlowQuery() > 0 )
+		if ( this.sqlStatementLogger.getLogSlowQuery() > 0 ) {
 			executeStart = System.currentTimeMillis();
+		}
 		try {
 			final ResultSet rs;
 			try {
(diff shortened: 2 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/engine/jdbc/spi/SqlStatementLogger.java
```diff
@@ -20,9 +20,15 @@
...
 	private static final Logger LOG = CoreLogging.logger( "org.hibernate.SQL" );
+	private static final Logger LOG_SLOW = CoreLogging.logger( "org.hibernate.SQL_SLOW" );
 
 	private boolean logToStdout;
 	private boolean format;
+	
+	/**
+	 * Configuration value that indicates slow query. (In milliseconds) 0 - disabled.
+	 */
+	private long logSlowQuery;
```
Sanne: Could you change this to a final field?

See also https://github.com/hibernate/hibernate-orm/commit/3483d7c02276b8d2d57929323b6715831235ce15

hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/ResultSetReturnImpl.java
```diff
@@ -61,6 +65,7 @@ public ResultSet extract(PreparedStatement statement) {
 			}
 			finally {
 				jdbcExecuteStatementEnd();
+				sqlStatementLogger.logSlowQuery( statement.toString(), executeStart );
```
Sanne: could you make sure that this is not executed if the feature is not enabled?

In particular doing statement.toString() is otherwise going to be a performance problem.

(same for the other methods)

hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/ResultSetReturnImpl.java
```diff
@@ -53,6 +53,10 @@ public ResultSetReturnImpl(JdbcCoordinator jdbcCoordinator) {
 	@Override
 	public ResultSet extract(PreparedStatement statement) {
 		// IMPL NOTE : SQL logged by caller
+		long executeStart = 0;
+		if ( this.sqlStatementLogger.getLogSlowQuery() > 0 ) {
+			executeStart = System.currentTimeMillis();
```
Sanne: since you're measuring _elapsed time_ I would suggest to use nanoTime() instead.

Sanne: (which implies the configuration reading code might need to be adapted to deal with nano-seconds precision ?)

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -1725,6 +1725,11 @@
 	 * is enabled logging of Session metrics is enabled by default too.
 	 */
 	String LOG_SESSION_METRICS = "hibernate.session.events.log";
+	
+	/**
+	 * Setting that logs query which executed slower than specified milliseconds. Default is 0 (disabled).
+	 */
+	String LOG_SLOW_QUERY = "hibernate.session.events.log.slow_query";
```
Sanne: The name of this property makes me expect it to be a boolean flag, taking true/false rather than a number.

LOG_QUERIES_SLOWER_THAN_MS ?

hibernate-core/src/main/java/org/hibernate/engine/jdbc/spi/SqlStatementLogger.java
```diff
@@ -105,4 +119,30 @@ public void logStatement(String statement, Formatter formatter) {
...
+	 *
+	 * @param sql The SQL query.
+	 * @param startTime Start time in milliseconds.
+	 */
+	@AllowSysOut
+	public void logSlowQuery(String sql, long startTime) {
+		if ( logSlowQuery < 1 ) {
+			return;
+		}
+		assert startTime > 0 : "startTime is invalid!";
```
Sanne: wouldn't all start times be invalid then?

hibernate-core/src/main/java/org/hibernate/engine/jdbc/spi/SqlStatementLogger.java
```diff
@@ -105,4 +119,30 @@ public void logStatement(String statement, Formatter formatter) {
...
+
+		long spent = System.currentTimeMillis() - startTime;
+
+		assert spent >= 0 : "startTime is invalid!";
+
+		if ( spent > logSlowQuery ) {
+			String logData = "SlowQuery: " + spent + " milliseconds. " + sql;
```
Sanne: should there be something after milliseconds to clarify that what follows is the sql? Should the SQL be quoted?

hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/ResultSetReturnImpl.java
batbileg289: I think milliseconds is enough. Also, thread-safe and more efficient.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/spi/SqlStatementLogger.java
batbileg289: Fixed in 35a2d98abd80e3ed64734e22f69ec6277b08726a

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
batbileg289: Fixed in 35a2d98abd80e3ed64734e22f69ec6277b08726a

hibernate-core/src/main/java/org/hibernate/engine/jdbc/spi/SqlStatementLogger.java
batbileg289: Fixed in 35a2d98abd80e3ed64734e22f69ec6277b08726a

hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/ResultSetReturnImpl.java
batbileg289: The method will be executed, but immediately return if the feature is not enabled. 

statement.toString() execution is fixed in 35a2d98abd80e3ed64734e22f69ec6277b08726a

hibernate-core/src/main/java/org/hibernate/engine/jdbc/spi/SqlStatementLogger.java
batbileg289: if the feature is enabled, startTime must have a valid value.

Sanne:
Looks great, thanks! Testing

Sanne:
merged it :100: 

We should also mention this option on the reference documentation, but that's ok to handle as a separate ticket. Would you like to do it? No pressure :)

 - https://hibernate.atlassian.net/browse/HHH-13539

PR #2975, opened 2019-08-06
HHH-13527 Remove contention from StatisticsImpl

Sanne:
- https://hibernate.atlassian.net/browse/HHH-13527

The backported version for 5.3 branch version needs to be a bit different, sending a separate PR.


Sanne:
don't merge this as it needs to incorporate feedback from review on https://github.com/hibernate/hibernate-orm/pull/2976

PR #2976, opened 2019-08-06
5.3 - HHH-13527 Remove contention from StatisticsImpl

Sanne:
backported version of https://github.com/hibernate/hibernate-orm/pull/2975

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
@@ -731,23 +731,25 @@ public class DefaultLoadEventListener extends AbstractLockUpgradeEventListener i
 		final EntityPersister persister,
 		SessionImplementor source ) {
 		final EntityDataAccess cache = persister.getCacheAccessStrategy();
+		final SessionFactoryImplementor factory = source.getFactory();
 		final Object ck = cache.generateCacheKey(
 				event.getEntityId(),
 				persister,
-				source.getFactory(),
+				factory,
 				source.getTenantIdentifier()
 		);
 
 		final Object ce = CacheHelper.fromSharedCache( source, ck, persister.getCacheAccessStrategy() );
-		if ( source.getFactory().getStatistics().isStatisticsEnabled() ) {
+		final StatisticsImplementor statistics = factory.getStatistics();
+		if ( statistics.isStatisticsEnabled() ) {
 			if ( ce == null ) {
-				source.getFactory().getStatistics().entityCacheMiss(
+				statistics.entityCacheMiss(
 						StatsHelper.INSTANCE.getRootEntityRole( persister ),
 						cache.getRegion().getName()
 				);
 			}
 			else {
-				source.getFactory().getStatistics().entityCacheHit(
+				statistics.entityCacheHit(
 						StatsHelper.INSTANCE.getRootEntityRole( persister ),
 						cache.getRegion().getName()
 				);
--- a/hibernate-core/src/main/java/org/hibernate/stat/internal/StatisticsImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/stat/internal/StatisticsImpl.java
@@ -6,8 +6,6 @@
  */
 package org.hibernate.stat.internal;
 
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.LongAdder;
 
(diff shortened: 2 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
```diff
@@ -673,11 +673,11 @@ private Object loadFromSecondLevelCache(
 			final EntityKey entityKey) {
 
 		final SessionImplementor source = event.getSession();
-		final boolean useCache = persister.canReadFromCache()
-				&& source.getCacheMode().isGetEnabled()
-				&& event.getLockMode().lessThan( LockMode.READ );
+		final boolean skipCache = ( ! persister.canReadFromCache() )
+				|| ( ! source.getCacheMode().isGetEnabled() )
+				|| ( ! event.getLockMode().lessThan( LockMode.READ ) );
```
gbadner: I must be missing something; I don't understand how the above change would make any difference performance-wise.

Was this change made to be more readable, rather than to improve performance?

hibernate-core/src/main/java/org/hibernate/stat/internal/StatsConcurrentHashMap.java
```diff
@@ -0,0 +1,76 @@
...
+ * used correctly for the purpose of Hibernate's statistics.
+ * In particular, we do like the semantics of ConcurrentHashMap#computeIfAbsent
+ * but not how it performs.
+ * See also:
+ *  - http://jsr166-concurrency.10961.n7.nabble.com/Re-ConcurrentHashMap-computeIfAbsent-td11687.html
+ *  - https://hibernate.atlassian.net/browse/HHH-13527
+ *
+ * @author Sanne Grinovero
+ */
+final class StatsConcurrentHashMap<V> {
```
gbadner: The name ```StatsConcurrentHashMap``` makes it sounds like it is a ```HashMap```, but it really isn't (since the class does not extend ```HashMap```).

Sanne: good point. How should I call it?

hibernate-core/src/main/java/org/hibernate/stat/internal/StatsConcurrentHashMap.java
```diff
@@ -0,0 +1,76 @@
...
+	/**
+	 * This method is inherently racy and expensive. Only use on non-hot paths, and
+	 * only to get a recent snapshot.
+	 * @return all keys in string form.
+	 */
+	public String[] keysAsArray() {
+		//We don't allocate an array directly, nor use streams, as its a concurrent map which we don't want to lock:
+		ArrayList<String> collector = new ArrayList<>( map.size() );
+		for ( Map.Entry<String,V> e : map.entrySet() ) {
+			collector.add( e.getKey() );
```
gbadner: Why use ```map.entrySet()```? If ```map.keySet()``` is used instead, then there would be no need to call ```e.getKey()```.

Javadoc for ```ConcurrentHashMap``` doesn't seem to mention why ```entrySet()``` would preferred over ```keySet()```.

Sanne: keySet() creates a fully copy of the key collection into a set: it's less efficient.

Sanne: Looks like what I said is no longer true since java8, I could change it back to keySet and have the same impact.

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
Sanne: you're right. This was meant to have some additional changes but I then changed them back.  I'll remove it as it became unnecessary noise. Thanks!

hibernate-core/src/main/java/org/hibernate/stat/internal/StatsConcurrentHashMap.java
gbadner: I've been trying to think of a good name, but having trouble.

I would prefer to have a separate class for each type of statistic, to be able to encapsulate a description and the computation for the particular type. I suppose it wouldn't perform as well, but I think it would be a lot more clear.

Should I implement my idea?

Sanne:
updated the PR

hibernate-core/src/main/java/org/hibernate/stat/internal/StatsConcurrentHashMap.java
Sanne: I wouldn't do that. The only reason I created a new container class is to have them behave consistently: I just needed to ensure all statistics usage would avoid using computeIfAbsent but rather use the alternative which always starts with an optimistic get.

Having separate container implementations would not help with consistency.

Sanne:
thanks! merged, will update master now.

PR #2977, opened 2019-08-06
Performance related backports

Sanne:
- https://hibernate.atlassian.net/browse/HHH-13494
 - https://hibernate.atlassian.net/browse/HHH-13495
 - https://hibernate.atlassian.net/browse/HHH-13527
 - https://hibernate.atlassian.net/browse/HHH-13531
 - https://hibernate.atlassian.net/browse/HHH-13534
 - https://hibernate.atlassian.net/browse/HHH-13544
 - https://hibernate.atlassian.net/browse/HHH-13512

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/util/StringHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/util/StringHelper.java
@@ -736,15 +736,32 @@ public final class StringHelper {
 	 *
 	 * @return The unquoted versions.
 	 */
-	public static String[] unquote(String[] names, Dialect dialect) {
+	public static String[] unquote(final String[] names, final Dialect dialect) {
 		if ( names == null ) {
 			return null;
 		}
-		String[] unquoted = new String[names.length];
-		for ( int i = 0; i < names.length; i++ ) {
-			unquoted[i] = unquote( names[i], dialect );
+		int failedIndex = -1;
+		final int length = names.length;
+		for ( int i = 0; i < length; i++ ) {
+			if ( isQuoted( names[i], dialect ) ) {
+				failedIndex = i;
+				break;
+			}
 		}
-		return unquoted;
+		if ( failedIndex == -1 ) {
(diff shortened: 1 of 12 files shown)


Sanne:
updated it to incorporate review changes to https://github.com/hibernate/hibernate-orm/pull/2976

Sanne:
Now also including backport of https://hibernate.atlassian.net/browse/HHH-13512

hibernate-core/src/main/java/org/hibernate/internal/util/StringHelper.java
```diff
@@ -708,12 +708,15 @@ public static String unquote(String name) {
...
-				&& name.charAt( name.length() - 1 ) == dialect.closeQuote() ) );
+	public static boolean isQuoted(final String name, final Dialect dialect) {
+		if ( name == null || name.isEmpty() ) {
+			return false;
+		}
+		final char first = name.charAt( 0 );
+		final char last = name.charAt( name.length() - 1 );
+
+		return ( ( first == last ) &&
+				( first == '`' || first == '"' || first == dialect.closeQuote() ) );
```
gbadner: The above is not quite right, because first and last may not be the same.  For example, ```SQLServerDialect.openQuote()``` returns ```'['```, and closeQuote() returns ```']'```.

For example,

gbadner: It should be:

```return first == dialect.openQuote() && last == dialect.closeQuote();```

gbadner: @Sanne, or should the fix avoid calling the dialect methods, if possible?

Sanne: Good catch. There is a different isQuoted method in which a similar optimisations was valid, but not here.

Sanne:
:100:

PR #2982, opened 2019-08-09
HHH-13551 StrategyRegistrationProvider does not properly handle implementations from different classloader

yrodiere:
https://hibernate.atlassian.net/browse/HHH-13551

I'm honestly not sure this is a legitimate bug. My personal opinion is that users should not expose to Hibernate ORM two classloaders that have a different version of the same class. But I suppose one can view things differently, so here is a fix.

The major risk is that we'll be catching exceptions that should have been thrown and noticed by the user, because their service really is misconfigured (a `META-INF/service/*` file that contains invalid class names, for example). I tried to mitigate the problem by logging the exceptions we ignore at the WARN level, but even then I'm really not sure it's a good idea...

Maybe we should add a feature flag? But then the feature wouldn't make much sense if it's disabled by default...


Sanne:
thanks @yrodiere and @snicoll : rebased and merged!

PR #2983, opened 2019-08-10
HHH-13259 Fix StackOverflowError in StringHelper

Legohuman:
Before fix method org.hibernate.internal.util.StringHelper#replace
matched placeholders illegally in case when ordinal parameters list was
expanded. Ex. placeholder ?1 was matched with ?11, ?12, ?13 etc. For
queries with 2 or more IN clauses with different collections there were
a situation when ?1 from the first clause matched with already expanded
placeholders from the second collection. Each match led to recursive
call of replace method. If collection in second clause was very long
then StackOverflowError occurred.

Fix adds check of partial placeholder match for wholeWords mode which
is used in expanding list parameters. Partial matches are skipped
during replace.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/util/StringHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/util/StringHelper.java
@@ -6,21 +6,13 @@
  */
 package org.hibernate.internal.util;
 
-import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.BitSet;
-import java.util.Collection;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Locale;
-import java.util.StringTokenizer;
-import java.util.regex.Pattern;
-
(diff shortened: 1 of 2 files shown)


Sanne:
I made a couple of changes and merged it:
 - rebased & fixed conflicts
 - moved the tests into the existing StringHelperTest
 - changed formatting to be consistent with our style.

Many thanks!

Legohuman:
Great! @Sanne, do you know which version of Hibernate this commit will be included in?

Sanne:
sure, it will be version `5.4.5.Final`. 

We might also backport it to some 5.3.x version, but probably too late for the next as the release process is ongoing.

For all such details, and in case plans change, we always keep the issue metadata up to date:
 - https://hibernate.atlassian.net/browse/HHH-13259

kdudzik:
Hi, when is the 5.4.5.Final release planned?

JWGmeligMeyling:
@kdudzik it has just been released

PR #2984, opened 2019-08-12
Add unit tests for org.hibernate.internal.util.StringHelper

EdgarScott:
I've analysed your codebase and noticed that `org.hibernate.internal.util.StringHelper` is not fully tested.

I've written some tests for the methods in this class with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

Hopefully, these tests will help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes that you consider important in a subsequent PR.


Sanne:
Thanks, I didn't know about DiffBlue - very interesting!

sebersole:
+1 about DiffBlue - very cool.

About StringHelper, we generally do not write unit tests for internal-only
util classes.  But I'll look these over and see.

gavinking:
We don't need to test internal classes. Closing.

PR #2985, opened 2019-08-13
HHH-13565

Sanne:
https://hibernate.atlassian.net/browse/HHH-13565

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
@@ -6,15 +6,19 @@
  */
 package org.hibernate.event.service.internal;
 
+import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.ListIterator;
+import java.util.Map;
 import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.function.Supplier;
 
 import org.hibernate.event.service.spi.DuplicationStrategy;
+import org.hibernate.event.service.spi.EventActionWithParameter;
 import org.hibernate.event.service.spi.EventListenerGroup;
 import org.hibernate.event.service.spi.EventListenerRegistrationException;
 import org.hibernate.event.service.spi.JpaBootstrapSensitive;
@@ -23,6 +27,7 @@ import org.hibernate.jpa.event.spi.CallbackRegistryConsumer;
 
 /**
  * @author Steve Ebersole
+ * @author Sanne Grinovero
  */
 class EventListenerGroupImpl<T> implements EventListenerGroup<T> {
 	private EventType<T> eventType;
@@ -30,9 +35,14 @@ class EventListenerGroupImpl<T> implements EventListenerGroup<T> {
 
 	private final Set<DuplicationStrategy> duplicationStrategies = new LinkedHashSet<>();
 
-	// Performance: make sure a forEach iteration on this type is efficient; in particular we choose ArrayList
-	// so to avoid allocating iterators.
-	private ArrayList<T> listeners;
+	// Performance: make sure iteration on this type is efficient; in particular we do not want to allocate iterators,
+	// not having to capture state in lambdas.
+	// So we keep the listeners in both a List (for convenience) and in an array (for iteration). Make sure
+	// their content stays in synch!
+	private T[] listeners = null;
+
+	//Update both fields when making changes!
+	private List<T> listenersAsList;
 
 	public EventListenerGroupImpl(EventType<T> eventType, EventListenerRegistryImpl listenerRegistry) {
 		this.eventType = eventType;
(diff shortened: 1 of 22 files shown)


hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
```diff
@@ -152,8 +153,12 @@
 
 	private CriteriaCompiler criteriaCompiler;
 
-	public AbstractSharedSessionContract(SessionFactoryImpl factory, SessionCreationOptions options) {
+	public AbstractSharedSessionContract(
+			SessionFactoryImpl factory,
+			SessionCreationOptions options,
+			FastSessionServices fastSessionServices) {
```
dreab8: nitpicking: probably we can avoid changing the constructor, we can get the `FastSessionServices` from the `factory`

Sanne: +1

thanks, cleaned it up and pushed a new version.

Sanne:
I've polished it a bit more, updated, and been re-benchmarking. Awesome results.

Will now push as soon as the JDK matrix test completes.

hibernate-core/src/main/java/org/hibernate/internal/FastSessionServices.java
```diff
@@ -0,0 +1,280 @@
...
+
+	private static Map<String, Object> initializeDefaultSessionProperties(SessionFactoryImpl sf) {
+		HashMap<String,Object> p = new HashMap<>();
+
+		//Static defaults:
+		p.putIfAbsent( AvailableSettings.FLUSH_MODE, FlushMode.AUTO.name() );
+		p.putIfAbsent( JPA_LOCK_SCOPE, PessimisticLockScope.EXTENDED.name() );
+		p.putIfAbsent( JPA_LOCK_TIMEOUT, LockOptions.WAIT_FOREVER );
+		p.putIfAbsent( JPA_SHARED_CACHE_RETRIEVE_MODE, CacheModeHelper.DEFAULT_RETRIEVE_MODE );
+		p.putIfAbsent( JPA_SHARED_CACHE_STORE_MODE, CacheModeHelper.DEFAULT_STORE_MODE );
```
NathanQingyangXu: Seems we don't need 'putIfAbsent' for the keys should be absent for sure

hibernate-core/src/main/java/org/hibernate/internal/FastSessionServices.java
```diff
@@ -0,0 +1,280 @@
...
+				JPA_LOCK_TIMEOUT,
+				AvailableSettings.FLUSH_MODE,
+				JPA_SHARED_CACHE_RETRIEVE_MODE,
+				JPA_SHARED_CACHE_STORE_MODE,
+				QueryHints.SPEC_HINT_TIMEOUT
+		};
+		final Map<String, Object> properties = sf.getProperties();
+		for ( String key : ENTITY_MANAGER_SPECIFIC_PROPERTIES ) {
+			if ( properties.containsKey( key ) ) {
+				p.put( key, properties.get( key ) );
```
NathanQingyangXu: Seems we can improve performance by using 'get()' result to avoid 'containsKey' overhead.
Also, we might consider 'getOrDefault' usage, so we might avoid some unnecessary default value assignment previously.

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
```diff
@@ -22,13 +27,22 @@
...
 	private final EventListenerRegistryImpl listenerRegistry;
 
 	private final Set<DuplicationStrategy> duplicationStrategies = new LinkedHashSet<>();
-	private List<T> listeners;
+
+	// Performance: make sure iteration on this type is efficient; in particular we do not want to allocate iterators,
+	// not having to capture state in lambdas.
+	// So we keep the listeners in both a List (for convenience) and in an array (for iteration). Make sure
+	// their content stays in synch!
+	private T[] listeners = null;
```
NathanQingyangXu: we can save one JVM step by avoiding assignment of default value. JVM has incurred overhead by assigning any reference field to null just after object is allocated.

hibernate-core/src/main/java/org/hibernate/internal/FastSessionServices.java
Sanne: Hi Nathan, these commits have been merged long ago.

If you want to improve code of Hibernate, please send a pull request with further changes associated to a JIRA - but also maybe choose wisely what to work on, your time is precious and limited and there's likely more important things to fix.

In particular the "get" is not safe to refactor as some properties might be set intentionally to `null`.

NathanQingyangXu: thanks for responding. I knew the PR has been merged long time ago, just left some comments to communicate. Yeah, null value in map is a good point. Yeah, your team members' time is pretty limited, but contributors' are not, though reviewing contributor's PR takes team member's cycles. Thanks.

PR #2987, opened 2019-08-13
HHH-13556 Tests doing dynamic fetch scrolling a collection fail on DB2

dreab8:
https://hibernate.atlassian.net/browse/HHH-13556


gbadner:
Thanks @dreab8 !

PR #2990, opened 2019-08-15
HHH-13579 Optimise ResourceRegistryStandardImpl to avoid heavy alloca…

Sanne:
…tion of iterators

> This class is tracking resources in small collections, and nested collections. When a statement is closed we need two levels of iteration; using a for loop on Iterator is too expensive, while a forEach on lambda would be more effective - even though the lambda is more expensive to initialize, but that's a one off cost.

 - https://hibernate.atlassian.net/browse/HHH-13579

I've been benchmarking this and it looks very effective. In a micro-benchmark which was focusing on small and simple operations (simple loads by id from the database) this would reduce the allocation of objects in TLAB by around 12%.

The code looks uglier to maintain, but I suppose this class won't be changing much and I think it's worth it.


Sanne:
merged this one, thanks!

PR #2992, opened 2019-08-17
Refactoring: use NamingHelper inside of Constraint

stokito:
Here is a small refactoring that can improve speed a little bit and removes duplicated code.
There is a lot of code duplication between `Constraint` and `NamingHelper` related to generation of constraint name.
I did step by step refactoring and it turned out that `Constraint` can just reuse the `NamingHelper`.
I kept an API and all tests passed.
There is a tricky part in #23833fd because the `Column` class doesn't have an `Identifier` (IMHO it worth to create one) so we'll create them manually.
Also I worried that the Table.name and Column.name fields can be in uppercase while the NamingHelper always lowercase them before hashing. This may change the generated hash.
But I checked and it turned out that the Table.name and Column.name fields are always in lowercase.

So this PR should be safe to merge.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/naming/NamingHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/naming/NamingHelper.java
@@ -77,9 +77,9 @@ public class NamingHelper {
 		// Use a concatenation that guarantees uniqueness, even if identical names
 		// exist between all table and column identifiers.
 
-		StringBuilder sb = new StringBuilder()
-				.append( "table`" ).append( tableName ).append( "`" )
-				.append( "references`" ).append( referencedTableName ).append( "`" );
+		StringBuilder sb = new StringBuilder(64)
+				.append( "table`" ).append( tableName ).append( '`' )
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/naming/NamingHelper.java
```diff
@@ -174,4 +138,13 @@ public String hashedName(String s) {
 			throw new HibernateException( "Unable to generate a hashed name!", e );
 		}
 	}
+
+	private String generateColumnsHash(String prefix, StringBuilder sb, Identifier[] columnNames) {
+		// Ensure a consistent ordering of columns, regardless of the order
+		// they were bound.
+		Arrays.stream( columnNames )
+				.sorted()
+				.forEachOrdered( columnName -> sb.append( "column`" ).append( columnName ).append( '`' ) );
```
NathanQingyangXu, line 147: The above code snippet might be slower than traditional approach. It would be great to conduct some JMH benchmark to justify the stream usage.

PR #2994, opened 2019-08-18
Replaces simple uses of Iterator with a corresponding for-each loop.

carstenartur:


Sanne:
As mentioned on another PR, I'm not sure if I should merge this as it provides little value and yet will certainly be in the way of the team, who's busy in a 6.0 branch.

IF we decide to go ahead and merge such changes, could you use the right formatting style?

for example

    for (Object array1 : array) {

should be

    for ( Object array1 : array ) {

and 

    if (Hibernate.isPropertyInitialized(array1, mappedBy)

should be

    if ( Hibernate.isPropertyInitialized( array1, mappedBy ) {

sebersole:
Probably a bit out of date, but the code styles can be found at
https://github.com/hibernate/hibernate-ide-codestyles

beikov:
IMO this is bad performance wise as it allocates iterators and shouldn't be merged

Sanne:
> IMO this is bad performance wise as it allocates iterators and shouldn't be merged

Not going to merge it as-is but iterating over an array shouldn't allocate an Iterator instance?

If @carstenartur is willing to fix the code style I'll be happy to review each single use case, polish and merge :)

beikov:
Using addAll(Arrays.asList(...)) will allocate a list and interator. The approach you suggested @sanne is fine, the other isn't.

carstenartur:
Are you aware of https://bugs.openjdk.java.net/browse/JDK-8155600?
I am not convinced that it is the case that it is slower. It might be in some corner cases with old java versions but not in general.. Besides the obvious reason to use it to improve readability (getting rid of helper variables) I think it opens the door to some optimizations in future with the same argument you should use lambdas: use internal iteration instead of external to be able to make use of internal parallelization. I have seen some reports that it improves performance but I think without real proof based on some realistic tests in the hibernate univers it is academic to claim that:)

beikov:
I am aware that the JIT will optimize at *some point* and escape analysis will *most probably* produce the same native code, but the boot code will never be executed as much to even be JIT compiled.

I agree in general that the use of lambdas etc. can help wrt. internal iteration, but when you know the implementation, like it is the case here(it's an array), it's best to just iterate the array by index as is.

carstenartur:
hi @beikov 
I hope you don't care that I add some comments while still leaving the decision to you guys with deepest respect about what to do now.
As far as I know the reason that the foreach loop can even be faster than simple uses of iterator has something to do with the most important advantage that the foreach loop brings to you: it is the type safe access to the loop elements. While the simple iterator needs to process some kind of type checking within the loop the foreach loop can leave that out within the loop completely as the type is known beforehand. I don't remember where I read about a publication going into details. I always recommend to prefer foreach loop to iterator if possible. The reasons are these:
1) it is much easier to read. You really have to check your loop variable  handling exactly to know which elements are processed with simple uses of iterator. It is a very nice thing to get rid of variables only introduced to be able to process a loop.
2) type safety
3) An iterator always will have to be executed sequentially. It is a semantical requirement given by the way the loop syntax has been created. Foreach inherently suggests to have no given order within the collection and because of this it is semantically much nearer to the parallel streams than the iterator. After you changed to foreach loop some ides offer you to change your loop into an lampda foreach in a very simple way.
4) error checking in ide: only after you change to foreach loop the ide is able to detect a problem like a unused loop element within the loop. If you do not access an array element within a simple loop the ide does not care.

Sanne:
@beikov 

> Using addAll(Arrays.asList(...)) will allocate a list and iterator. The approach you suggested @Sanne is fine, the other isn't.

agreed but I guess you're talking about the other PR? I don'tsee any `Arrays.asList` in this one :)

@carstenartur good point about JDK-8155600 but we still have a very strong user base on JDK8 .. some are very sensitive to an increase of allocations so I need to be careful on such. But also that seems not relating to this PR.

Since you're both very interested in performance, you might be interested to know that we've been optimising some very hot loops recently; patch is about to land.

One thing of interest is that I've been benchmarking & profiling each small change individually; regarding a very similar issue with Iteration allocations, I noticed that while we iterate on EventListeners all iterators are "JITted away", except `FlushEventListener` - while their usage all looks the same. I have still no explanation for it, will dig further or pass it on to the JDK team :)

But my take away is that one can't consistently rely on JIT: it's cool when it works but when it matters it might be best to be explicit about it.

Also, Christian makes a good point about non-optimised code performance. Have you all seen the bootstrap performance improvements I developed for https://quarkus.io/? Thanks to those patches one can now boot Hibernate in ~10 ms and indeed one thing I had to get my head around was to assume it would run in interpreted mode: optimise all bootstrap code for interpreted as it's rarely getting hot enough. (Class initialization & classloaders are still slow, so the ~10ms is not feasible at first run, but it's great for the hot-reload feature)

Back to this PR: it looks very good in principle, will just need to find a little time to check them each.

beikov:
@carstenartur sorry about the fuzz, I commented on the wrong PR. The changes you did in this PR are good. A for-each loop on an array compiles down to an indexed loop, so it's just a nicer syntax for the same code 👍

Sanne:
FYI I fixed the formatting, testing now and will merge soon.

Sanne:
merged this. Thanks all for the great care!

PR #2995, opened 2019-08-18
Changing occurrences of manual array copying via for loop

carstenartur:


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElement.java
```diff
@@ -348,9 +349,7 @@ public String getIdentityColumn() {
 					table, propertyNames[i],
 					inSelect
 			);
-			for ( int j = 0; j < propertyNameColumns.length; j++ ) {
-				columns.add( propertyNameColumns[j] );
-			}
+			columns.addAll(Arrays.asList(propertyNameColumns));
```
iherasymenko: There is a better way so that a new Arrays.asList wrapper won't be allocated. java.util.Collections.addAll();

carstenartur: That's right. Feel free to provide an update.

JWGmeligMeyling: Because this change affects code in the HQL parser and entity persisters, I think the extra memory allocation is unacceptable. I am sure there's a reason why these were loops over arrays in the first place 😄  Changing the code to use `java.util.Collections.addAll` however may improve readability.

carstenartur: Problem with a loop is that it can cause repeating acquisition of memory while processing the loop which increases pressure on the garbage collection. In general it is better acquire the needed memory only once.

Sanne: @carstenartur what "repeating acquisition of memory" are you referring to? I guess the Collection needing to be resized?

It sould be possible to hint for the resize just once before looping, and yet avoid allocating an intermediary `List` instance.

carstenartur: @Sanne Yes, I talk about resizing.

Sanne: for example:

     ArrayList<String> columns = new ArrayList<>();
    ...
     columns.ensureCapacity( columns.size() + propertyNameColumns.length );
     Collections.addAll( columns, propertyNameColumns );

Or even better, would be lovely to get rid of the ArrayList and the intermediate `propertyNameColumns` arrays, but that's certainly going to get a bit more complex and is unlikely to be worth is as this is all being rewritte in in 6.
I would suggest the above change so to stick to the intent of this commit: getting rid of manual array copies over a loop, which is a very sensible idea.

Thanks!

beikov: Using ensureCapacity is the way to go here IMO

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElement.java
```diff
@@ -342,16 +342,17 @@ public String getIdentityColumn() {
 		}
 
 		final String[] propertyNames = getIdentifierPropertyNames();
-		List<String> columns = new ArrayList<>();
+		ArrayList<String> columns = new ArrayList<>();
```
beikov: You could pre-size it like

`ArrayList<String> columns = new ArrayList<>( propertyNames.length );`

hibernate-core/src/main/java/org/hibernate/sql/InsertSelect.java
```diff
@@ -45,7 +45,7 @@ public InsertSelect addColumn(String columnName) {
 	}
 
 	public InsertSelect addColumns(String[] columnNames) {
-		this.columnNames.addAll(Arrays.asList(columnNames));
+		Collections.addAll( this.columnNames,columnNames );
```
beikov: You could additionally change the type of `columnNames` to `ArrayList` and use `ensureCapacity` before adding.

beikov:
Except for styling issues, this looks good.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/IntoClause.java
```diff
@@ -160,10 +160,8 @@ private void visitPropertySpecNodes(AST propertyNode, List types) {
 			if ( persister.getIdentifierType() instanceof CompositeType ) {
 				if ( componentIds == null ) {
 					String[] propertyNames = ( (CompositeType) persister.getIdentifierType() ).getPropertyNames();
-					componentIds = new HashSet();
-					for ( String propertyName : propertyNames ) {
-						componentIds.add( propertyName );
-					}
+					componentIds = new HashSet(propertyNames.length);
+					Collections.addAll(componentIds,propertyNames);
```
NathanQingyangXu: I think the above two statements could be combined with
`
componentIds = new HashSet(propertyNames);
`

hibernate-core/src/main/java/org/hibernate/sql/InsertSelect.java
```diff
@@ -20,7 +20,7 @@
 public class InsertSelect {
 	private String tableName;
 	private String comment;
-	private List columnNames = new ArrayList();
+	private ArrayList<String> columnNames = new ArrayList<>();
```
NathanQingyangXu: Do we need ArrayList field type specification? In previous JDK version there is minor performance implication but in recent JDK the difference is eliminated, unless I were wrong.

hibernate-envers/src/test/java/org/hibernate/envers/test/tools/TestTools.java
```diff
@@ -27,11 +27,9 @@
  */
 public class TestTools {
 	public static <T> Set<T> makeSet(T... objects) {
-		final Set<T> ret = new HashSet<T>();
+		final Set<T> ret = new HashSet<T>(objects.length);
```
NathanQingyangXu: btw, the above usage is not totally correct for it doesn't take into consideration of 'load factor'. Suppose load factor is 0.75, you might specify as `new HashSet<>(objects.length / 0.75)`.
Guava library provides util method for this, i.e. `Sets.withExpectedSize(objects.length)`. see https://guava.dev/releases/23.0/api/docs/src-html/com/google/common/collect/Sets.html#line.252

hibernate-envers/src/test/java/org/hibernate/envers/test/integration/strategy/ValidityAuditStrategyManyToManyTest.java
```diff
@@ -155,19 +155,15 @@ public void testHistoryOfEd1() {
 	private SetOwningEntity createOwningEntity(SetOwnedEntity... owned) {
 		SetOwningEntity result = new SetOwningEntity( ing_id, "parent" );
 		result.setReferences( new HashSet<SetOwnedEntity>() );
-		for ( SetOwnedEntity setOwnedEntity : owned ) {
-			result.getReferences().add( setOwnedEntity );
-		}
+		result.getReferences().addAll(Arrays.asList(owned));
```
NathanQingyangXu: Why not use Collections.addAll above?

hibernate-envers/src/test/java/org/hibernate/envers/test/integration/onetomany/BasicSetWithMulId.java
```diff
@@ -94,11 +94,9 @@ public void testRevisionsCounts() {
 	}
 
 	private <T> Set<T> makeSet(T... objects) {
-		Set<T> ret = new HashSet<T>();
+		Set<T> ret = new HashSet<T>(objects.length);
```
NathanQingyangXu: Again, this size spec is usually not what user wants. See https://guava.dev/releases/23.0/api/docs/src-html/com/google/common/collect/Sets.html#line.252

hibernate-core/src/main/java/org/hibernate/sql/InsertSelect.java
```diff
@@ -44,9 +44,8 @@ public InsertSelect addColumn(String columnName) {
 	}
 
 	public InsertSelect addColumns(String[] columnNames) {
-		for ( String columnName : columnNames ) {
-			this.columnNames.add( columnName );
-		}
+		this.columnNames.ensureCapacity(this.columnNames.size()+columnNames.length);
```
NathanQingyangXu: Nice!

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElement.java
```diff
@@ -341,15 +341,14 @@ public String getIdentityColumn() {
...
 		for ( String propertyName : propertyNames ) {
 			String[] propertyNameColumns = toColumns( table, propertyName, inSelect );
-			for ( String propertyNameColumn : propertyNameColumns ) {
-				columns.add( propertyNameColumn );
-			}
+			columns.ensureCapacity( columns.size() + propertyNameColumns.length );
+			Collections.addAll( columns,propertyNameColumns );
 		}
-		return columns.toArray( new String[columns.size()] );
+		return columns.toArray( new String[0] );
```
NathanQingyangXu: Interesting. see https://stackoverflow.com/questions/174093/toarraynew-myclass0-or-toarraynew-myclassmylist-size

hibernate-core/src/main/java/org/hibernate/sql/InsertSelect.java
carstenartur: You prefer List<String> instead of ArrayList<String>? Yes, I think it is better too. We should change it.

hibernate-envers/src/test/java/org/hibernate/envers/test/tools/TestTools.java
carstenartur: Yes, what would be better is the constructor with two parameters including the loadfactor itself.

hibernate-envers/src/test/java/org/hibernate/envers/test/integration/strategy/ValidityAuditStrategyManyToManyTest.java
carstenartur: Yes, why not..

hibernate-envers/src/test/java/org/hibernate/envers/test/integration/onetomany/BasicSetWithMulId.java
carstenartur: Same as above, explicitly setting the loadfactor using the two parameter constructor would make sure that you explicitly create a hashset of exactly the needed size.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElement.java
carstenartur: Yes, in recent java versions using the zero sized constructor is the fastest way.

NathanQingyangXu: it seems after you fix the style issue your PR has good chance tobe merged. why not continue perfecting the PR to have it finalized?

NathanQingyangXu: it seems after you fix the style issue your PR has good chance tobe merged. why not continue perfecting the PR to have it finalized?

hibernate-envers/src/test/java/org/hibernate/envers/test/tools/TestTools.java
NathanQingyangXu: Recently I found Hibernate has provided a util method for this: `org.hibernate.internal.util.collections.CollectionHelper.mapOfSize(int size)` already

Naros:
Great work guys, but one little nitpick.  I noticed a number of the changes do not adhere to the code formatting rules with regards to spacing, e.g. `methodCall( arg1, arg2 );`.  While check style may omit tests from explicitly causing a build failure in this regard, its good practice to use the style in all changes.

PR #3001, opened 2019-08-20
HHH-13592 AutoFlushEvent#isFlushRequired is always false

yrodiere:
https://hibernate.atlassian.net/browse/HHH-13592


yrodiere:
Hi @gbadner , does this look good to you?

yrodiere:
Thanks!

PR #3002, opened 2019-08-22
Convert anonymous inner classes to lambda expressions or member references

carstenartur:
Not sure if this is a good idea. Not in all cases it improves readability a lot. But see yourself..

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/action/internal/BulkOperationCleanupAction.java
+++ b/hibernate-core/src/main/java/org/hibernate/action/internal/BulkOperationCleanupAction.java
@@ -179,25 +179,22 @@ public class BulkOperationCleanupAction implements Executable, Serializable {
 
 	@Override
 	public AfterTransactionCompletionProcess getAfterTransactionCompletionProcess() {
-		return new AfterTransactionCompletionProcess() {
-			@Override
-			public void doAfterTransactionCompletion(boolean success, SharedSessionContractImplementor session) {
-				for ( EntityCleanup cleanup : entityCleanups ) {
-					cleanup.release();
-				}
-				entityCleanups.clear();
-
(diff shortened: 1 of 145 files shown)


Sanne:
Sorry I don't think that's a good idea. I hope you generated this automatically?

The way lambdas are implemented in the JVM, each of them needs to be defined in a new classloader.  That's not a problem for runtime performance, but if you have too many lambdas in a project it takes ages to initialize them all at cold boot, and it consumes more metadata space.

Please see https://developers.redhat.com/blog/2018/02/14/java-class-metadata/

Sanne:
N.B. that's not to say that we should forbid lambdas, just that they should be used sparingly and when there is a good reason. Anonymous classes might look ugly but will have to do for most cases.

carstenartur:
no problem. It is created semiautomatic using Netbeans - we can close it.  I am aware of this startup time and metadata argument. Your position is the same as what is explained at (regarding hibernate-search) 
https://hibernate.atlassian.net/browse/HSEARCH-3564?filter=-3

This is a pointer to someone who prefers using lambdas instead of anonymous classes like in this refactoring but I understand your arguments. So no need to discuss it on my side. Just want to share this to explain why I dared to create this pull request.
https://medium.com/@yury.nino.roa/effective-java-for-lambda-functions-method-references-and-standard-interfaces-6648552f685c
So just close this pull request when you decide that it is not what you want. In case you would like to use parts of it - eg. the member references - just tell me. I could throw out everything else.

Sanne:
the method references are very nice; I would like to have those eventually, but remember the core ORM team still needs to be able to do rebases regularly.

Considering you can generate these automatically, maybe let's close this and then regenerate the refactoring to references on ORM 6 only - when the branches are finally stabilized from the heavy rebasing we have this summer?

carstenartur:
Thats fine, ok!

PR #3004, opened 2019-08-23
HHH-12030 Fix ClassCastException when Embeddable collection uses generic types

Naros:
https://hibernate.atlassian.net/browse/HHH-12030


dreab8:
rebased and applied upstream.

Thanks @Naros

PR #3005, opened 2019-08-23
HHH-13564 - Fix NullPointerException for audited entity with embedded-id in superclass that makes use of generics.

Naros:
https://hibernate.atlassian.net/browse/HHH-13564


Naros:
Merged.

PR #3006, opened 2019-08-24
HHH-13586 : ClassCastException when using a single region name for both entity and query results

gbadner:
https://hibernate.atlassian.net/browse/HHH-13586

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/internal/DisabledCaching.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/internal/DisabledCaching.java
@@ -12,6 +12,7 @@ import java.util.Set;
 
 import org.hibernate.cache.cfg.spi.DomainDataRegionConfig;
 import org.hibernate.cache.spi.CacheImplementor;
+import org.hibernate.cache.spi.DomainDataRegion;
 import org.hibernate.cache.spi.QueryResultsCache;
 import org.hibernate.cache.spi.Region;
 import org.hibernate.cache.spi.RegionFactory;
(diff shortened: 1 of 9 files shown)


gbadner:
@sebersole, I agree with what you wrote.

The only exception is for ```org.hibernate.Cache#evictRegion```. Currently, in this PR, data in both types of regions with the specified name get evicted.

I see that ```org.hibernate.Cache``` already has methods for each type of region, so the application can already use these methods instead.

I'm in favor of deprecating both ```org.hibernate.Cache#evictRegion``` and ```CacheImplementor#getRegion```

WDYT?

gbadner:
@sebersole, while I'm at it, what do you think about deprecating ```CacheImplementor#getCacheRegionNames```, since ```#getDomainDataRegionNames``` and ```getQueryCacheRegionNames``` will be available?

gbadner:
@sebersole, after looking more closely, I see that ```Cache#evictRegion``` really can't be deprecated because the other "evict" methods in ```Cache``` do not evict by region name.

I've made the other SPI changes.

I still need to add some tests.

Please take a look and let me know how it looks to you.

gbadner:
@sebersole, I thought about adding ```Cache#evictDataRegion``` and ```#evictQueryRegion``` but decided against it because ```DomainDataRegion``` and ```QueryResultsRegion``` are both SPIs, not APIs. I preferred to keep ```Cache``` not knowing about the different types of regions.

Currently, calling ```Cache#evictRegion``` will evict all caches with the provided region name.

I realized there was still a little fuzziness to the new methods that I added. I figured it would be best to fix it now.

```#getQueryCacheRegionNames``` included the default ```QueryResultsRegion``` region name, but ```#getQueryResultsCacheStrictly( defaultQueryResultsRegionName )``` returns ```null```.

That makes it a bit clumsy for an application to use. 

I am changing ```#getQueryCacheRegionNames``` to exclude the default ```QueryResultsRegion``` region name, since it is intended for use with ```#getQueryResultsCacheStrictly```.

The deprecated ```getCacheRegionNames``` will continue to include the default ```QueryResultsRegion``` region name, and the deprecated ```getRegion( defaultQueryResultsRegionName )``` will continue return the default ```QueryResultsRegion```.

Going forward, ```getDefaultQueryResultsCache().getRegion()``` should be used for getting the default ```QueryResultsRegion```, not the other methods.

I am clarifying the Javadoc so it is more obvious how the various methods deal with the default ```QueryResultsRegion```/```QueryResultsCache```.

Does this make sense to you in principal?

gbadner:
Added more test cases and made the changes proposed in previous comment.

gbadner:
The first 2 commits have be pushed to master.

The remainder have been squashed and pushed to a new PR for [HHH-13613](https://hibernate.atlassian.net/browse/HHH-13613): https://github.com/hibernate/hibernate-orm/pull/3029

PR #3012, opened 2019-08-29
HHH-12991 Metamodel Generator with Java 11 and Error Prone

hackmann:
Workaround problem with JPA metamodel generator not finding a JAXB
implementation in the bootstrap classpath. Explicitly ask it to look
in the same classpath as generator


beikov:
I'll close this as out of date because in 6.0 we moved to Java 11 as baseline and use Jakarta XML Binding now through an explicit dependency.

PR #3015, opened 2019-08-31
fix typos

KangZhiDong:


KangZhiDong:
> hi, thanks for the help but most of these fixes are actually wrong :)
> 
> Could you please amend the patch to only change the incorrect ones?

“an http、 a url、A unique” ，https://www.a-or-an.com

Sanne:
> > hi, thanks for the help but most of these fixes are actually wrong :)
> > Could you please amend the patch to only change the incorrect ones?
> 
> “an http、 a url、A unique” ，https://www.a-or-an.com

where the sources read `https://javaee.github.io/javaee-spec/javadocs/javax/persistence/TemporalType.html[`@TemporalType`]` that will be rendered as "TmporalType" so it's starting with a T.  The "http" isn't visible in the final document.

thanks!

KangZhiDong:
> > > hi, thanks for the help but most of these fixes are actually wrong :)
> > > Could you please amend the patch to only change the incorrect ones?
> > 
> > 
> > “an http、 a url、A unique” ，https://www.a-or-an.com
> 
> where the sources read `https://javaee.github.io/javaee-spec/javadocs/javax/persistence/TemporalType.html['@TemporalType']` that will be rendered as "TemporalType" so it's starting with a T. The "http" isn't visible in the final document.
> 
> thanks!

I understand

PR #3016, opened 2019-09-02
HHH-13608 Oracle8iDialect should use CASE_INSENSITIVE pattern matching when checking the statement type

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13608


dreab8:
rebased and applied upstream.

Thanks @vladmihalcea

vladmihalcea:
@dreab8 Thanks for integrating it.

dreab8:
Hi @vladmihalcea , I have reverted this commit due to failures of `org.hibernate.test.queryhint.QueryHintTest#testQueryHint()` and `org.hibernate.test.queryhint.QueryHintTest#testQueryHintAndComment()`

dreab8:
sorry @vladmihalcea I think I spotted the issue

vladmihalcea:
I was fixing and testing it now. If you already fixed it, I'll leave it the. It requires a replace for the entity name as the entity uses an explicit name as it should.

dreab8:
yes I have already done, Thanks

PR #3019, opened 2019-09-02
HHH-13606 LocalDateTimeTest fails for HANA on CI

yrodiere:
https://hibernate.atlassian.net/browse/HHH-13606

Turns out this was simply a bug in the HANA JDBC driver that has been fixed since.

Note that we already updated to version 2.4.59 of the driver in `libraries.gradle`, so we probably just forgot to update `hana/matrix.gradle`.

### DIFF
--- a/databases/hana/matrix.gradle
+++ b/databases/hana/matrix.gradle
@@ -5,4 +5,4 @@
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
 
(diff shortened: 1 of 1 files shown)


yrodiere:
Here is a successful HANA build following the update: https://ci.hibernate.org/view/Personal%20jobs/job/hibernate-orm-personal-yoann-hana/10/

yrodiere:
Ok, it's been long enough. Merging.

PR #3021, opened 2019-09-03
HHH-13611 : Restore EntityMetamodel constructor to take SessionFactoryImplementor argument instead of PersisterCreationContext

gbadner:
https://hibernate.atlassian.net/browse/HHH-13611

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
@@ -568,7 +568,7 @@ public abstract class AbstractEntityPersister
 			this.naturalIdRegionAccessStrategy = null;
 		}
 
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -568,7 +568,7 @@ public AbstractEntityPersister(
 			this.naturalIdRegionAccessStrategy = null;
 		}
 
-		this.entityMetamodel = new EntityMetamodel( persistentClass, this, creationContext );
+		this.entityMetamodel = new EntityMetamodel( persistentClass, this, creationContext.getSessionFactory() );
```
dreab8: we can use `factory` instead of calling `creationContext.getSessionFactory`

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/3033

PR #3022, opened 2019-09-04
HHH-9301 - Support select variable refs in group by for DBMS dialects that support it

jwgmeligmeyling:
While not strictly compliant with the SQL specification, the MySQL, PostgreSQL and H2 support the use of select aliases in the GROUP BY clause. An obvious benefit is that produced SQL queries will be easier to read, because complex select expressions from aggregrate queries will have to be included in the group by clause as well. These can now simply reference the aliases of the respective columns for the tuple element. However, there is also a functional difference. For function invocations that have parameterized arguments, the query optimizer can't guarantee the that the function result between the selection projection and grouping process are equal. This results in an error because a value is projected which is not grouped by. An example where this for example becomes relevant, is when parameterizing TimeScaleDB's [`time_bucket_gapfill()`](https://docs.timescale.com/latest/api#time_bucket_gapfill-examples) function.

For example, using `time_bucket_gapfill()` one might want to produce a criteria query that produces the following SQL:

```sql
SELECT
    time_bucket_gapfill(?, time, ?, ?) AS ts,
    avg(tg) as tg
FROM iaqmeasurement
GROUP BY ts
ORDER BY ts;
```

When the alias is not used as grouping value, the query will yield an error:

```sql
SELECT
    time_bucket_gapfill(?, time, ?, ?) AS ts,
    avg(tg) as tg
FROM iaqmeasurement
GROUP BY time_bucket_gapfill(?, time, ?, ?)
ORDER BY ts;
```

Of course the parameter values can just be stored in a CTE as well, but I think we should consider support for variable refs in group by for DBMS dialects that support it. This pull request implements the feature equivalently to how its currently done for order by clauses, and only enables the feature for known supported dialects (H2, PostgreSQL and MySQL - based on [https://stackoverflow.com/a/3841804/2104280](https://stackoverflow.com/a/3841804/2104280)). I am not sure what we should do with dialects that do not support it. We could either throw an error, rely on the database to throw an error, or copy the expression from the select alias ourselves in an attempt to fix the query. 

Jira: https://hibernate.atlassian.net/browse/HHH-9301

### DIFF
--- a/hibernate-core/src/main/antlr/hql-sql.g
+++ b/hibernate-core/src/main/antlr/hql-sql.g
@@ -237,6 +237,10 @@ tokens
 		return false;
 	}
 
+	protected boolean isGroupExpressionResultVariableRef(AST ident) throws SemanticException {
+		return false;
+	}
+
 	protected void handleResultVariableRef(AST resultVariableRef) throws SemanticException {
 	}
 
@@ -394,7 +398,7 @@ resultVariableRef!
 	;
 
 groupClause
-	: #(GROUP { handleClauseStart( GROUP ); } (expr [ null ])+ ( #(HAVING logicalExpr) )? ) {
+	: #(GROUP { handleClauseStart( GROUP ); } ({ isGroupExpressionResultVariableRef( _t ) }? resultVariableRef | expr [ null ])+ ( #(HAVING logicalExpr) )? ) {
 		handleClauseEnd();
 	}
 	;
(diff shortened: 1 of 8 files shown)


sebersole:
Thanks @JWGmeligMeyling !  This is very useful, I agree.

jwgmeligmeyling:
Perhaps something to integrate for 5.4 now we're integrating small non-breaking features again? @Sanne

Sanne:
@jwgmeligmeyling sure, go ahead and merge it in master? Looks like it was reviewed already.

dreab8:
rebased and applied upstream.

Thanks @jwgmeligmeyling

PR #3027, opened 2019-09-06
Use PU Info Classloader

kaja-mohideen:
Have entity classes from different bundles loaded seamlessly.


Sanne:
hi thanks for the patch!
Could you either describe what you're aiming at, or even better add some tests?

kaja-mohideen:
This is to avoid race condition (instantiation of EM) during startup when there are multiple bundles trying to use a common bundle which owns the PersistentUnit.

Sanne:
@kaja-mohideen ah, makes sense - thanks for the explanation!

I'll create a JIRA for this and merge it.

Sanne:
Merged as https://hibernate.atlassian.net/browse/HHH-13696

thanks!

kaja-mohideen:
Shows PR is not merged, but closed !

Sanne:
@kaja-mohideen no worries, that's because we rebase before merging. Github is unable to detect this and leaves the PRs open, so we close them manually.

When in doubt, check the master branch, and search commit logs for the JIRA issues:
 - https://github.com/hibernate/hibernate-orm/commits/master

PR #3029, opened 2019-09-06
HHH-13613 : Improve CacheImplementor SPI to better support a DomainDa…

gbadner:
…taRegion and QueryResultsRegion with the same name

https://hibernate.atlassian.net/browse/HHH-13613

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/internal/EnabledCaching.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/internal/EnabledCaching.java
@@ -56,11 +56,9 @@ public class EnabledCaching implements CacheImplementor, DomainDataRegionBuildin
 	private final SessionFactoryImplementor sessionFactory;
 	private final RegionFactory regionFactory;
 
-	private final Map<String,Region> regionsByName = new ConcurrentHashMap<>();
+	private final Set<String> domainDataAndQueryResultsRegionNames = new HashSet<>();
 
-	// A map by name for QueryResultsRegion instances that have the same name as a Region
-	// in #regionsByName.
-	private final Map<String, QueryResultsRegion> queryResultsRegionsByDuplicateName = new ConcurrentHashMap<>();
+	private final Map<String, DomainDataRegion> domainDataRegionMap = new ConcurrentHashMap<>();
 
 	private final Map<NavigableRole,EntityDataAccess> entityAccessMap = new ConcurrentHashMap<>();
 	private final Map<NavigableRole,NaturalIdDataAccess> naturalIdAccessMap = new ConcurrentHashMap<>();
@@ -95,11 +93,12 @@ public class EnabledCaching implements CacheImplementor, DomainDataRegionBuildin
 					RegionFactory.DEFAULT_QUERY_RESULTS_REGION_UNQUALIFIED_NAME,
 					sessionFactory
 			);
-			regionsByName.put( queryResultsRegion.getName(), queryResultsRegion );
+			domainDataAndQueryResultsRegionNames.add( queryResultsRegion.getName() );
 			defaultQueryResultsCache = new QueryResultsCacheImpl(
 					queryResultsRegion,
 					timestampsCache
 			);
+			// defaultQueryResultsCache is not included in namedQueryResultsCacheMap.
 		}
 		else {
 			timestampsCache = new TimestampsCacheDisabledImpl();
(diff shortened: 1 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/cache/internal/EnabledCaching.java
```diff
@@ -95,11 +93,12 @@ public EnabledCaching(SessionFactoryImplementor sessionFactory) {
 					RegionFactory.DEFAULT_QUERY_RESULTS_REGION_UNQUALIFIED_NAME,
 					sessionFactory
 			);
-			regionsByName.put( queryResultsRegion.getName(), queryResultsRegion );
+			domainDataAndQueryResultsRegionNames.add( queryResultsRegion.getName() );
```
gbadner: The default ```QueryResultsRegion``` is not added to ```#legacySecondLevelCacheNames```. Should it be?

If so, should it be qualified?

`
				legacySecondLevelCacheNames.add(
						StringHelper.qualifyConditionally(
								getSessionFactory().getSessionFactoryOptions().getCacheRegionPrefix(),
								queryResultsRegion.getName()
						)
				);
`

hibernate-core/src/main/java/org/hibernate/cache/internal/EnabledCaching.java
```diff
@@ -428,7 +442,7 @@ public void evictDefaultQueryRegion() {
 
 	@Override
 	public void evictQueryRegion(String regionName) {
-		final QueryResultsCache cache = getQueryResultsCache( regionName );
+		final QueryResultsCache cache = namedQueryResultsCacheMap.get( regionName );
```
gbadner: ```#getQueryResultsCache``` will create a new ```QueryResultsCache``` if one does not exist with the provided ```regionName```. 

I didn't think that it made sense to create a new ```QueryResultsCache``` when evicting, so I changed the line to lookup only an existing.

Should this be reverted?

hibernate-core/src/main/java/org/hibernate/cache/internal/EnabledCaching.java
```diff
@@ -491,32 +506,26 @@ public QueryResultsCache getQueryResultsCacheStrictly(String regionName) {
...
-		else {
-			// There was already a different type of Region with the same name.
-			queryResultsRegion = queryResultsRegionsByDuplicateName.computeIfAbsent(
-					regionName,
-					this::makeQueryResultsRegion
-			);
-		}
-		final QueryResultsCacheImpl regionAccess = new QueryResultsCacheImpl(
+		domainDataAndQueryResultsRegionNames.add( regionName );
+		legacySecondLevelCacheNames.add( regionName );
```
gbadner: The new named ```QueryResultsRegion``` is not qualified when added to #legacySecondLevelCacheNames. 

Should it be qualified as follows?

```
legacySecondLevelCacheNames.add( StringHelper.qualifyConditionally( 
        getSessionFactory().getSessionFactoryOptions().getCacheRegionPrefix(), 
        regionName ) 
);
```

gbadner:
```StatisticsImpl#getCacheRegionStatistics``` and ```getQueryRegionStats``` call ```CacheImplementor#getQueryResultsCache```. 

Should a ```QueryResultsRegion``` really be created if one does not exist yet?

I suppose it makes sense to do this, if the application will ultimately create the region.

WDYT?

hibernate-core/src/main/java/org/hibernate/cache/internal/EnabledCaching.java
sebersole: Practically speaking the default region does have a name, but conceptually that is not the case.  No it should not be added to `legacySecondLevelCacheNames`

hibernate-core/src/main/java/org/hibernate/cache/internal/EnabledCaching.java
sebersole: I think not creating the region on evict is better.  So I'd leave your change

hibernate-core/src/main/java/org/hibernate/cache/internal/EnabledCaching.java
sebersole: Actually, I believe that the legacy behavior was to not expect the user to prefix the names for query regions which is different than its expectations for domain data regions.  Not completely sure, so we should maybe verify that.

sebersole:
>> StatisticsImpl#getCacheRegionStatistics and getQueryRegionStats call CacheImplementor#getQueryResultsCache.
>>
>> Should a QueryResultsRegion really be created if one does not exist yet?

The stats reference itself should be created.  So from there it depends whether you want to have a mutable `CacheRegionStatisticsImpl` to be able to inject the `Region` later if need be or to just create the `Region` up front to pass along to the `CacheRegionStatisticsImpl`.  IMO creating the `Region` when we create the `CacheRegionStatisticsImpl` makes the most sense.

sebersole:
Actually, `#getQueryRegionStats` is used only when we are triggered when the stats are being updated because of the execution of a query - it is only ever called from `StatisticsImpl#queryCacheHit`, `StatisticsImpl#queryCacheMiss`, etc.  Here we fully expect the region to exist.

WRT `StatisticsImpl#getCacheRegionStatistics`, as outlined in its comments it is a deprecated legacy method and its implementation it intended to be consistent with the legacy expectation.  Its replacement is `StatisticsImpl#getQueryRegionStatistics` which correctly uses `CacheImplementor#getQueryResultsCacheStrictly` instead


.  So it should create the region

PR #3037, opened 2019-09-10
HHH-13616 Enable the hibernate-orm-modules test for JDK 11

dreab8:
https://hibernate.atlassian.net/browse/HHH-13616


hibernate-orm-modules/hibernate-orm-modules.gradle
```diff
@@ -108,8 +103,8 @@ provision {
 }
 
 dependencies {
-	testCompile project( ":hibernate-core" )
-	testCompile project( ":hibernate-envers" )
+	compile project( ":hibernate-core" )
```
yrodiere: This, and moving test files to `src/main`, looks wrong. What was the problem with keeping the test classes in the test directory?

dreab8: I may be wrong but it seems the two classes were treated as test classes.
`java.lang.Exception: Test class should have exactly one public zero-argument constructor` moving to the `src` folder seems fixing the issue

yrodiere: Looks like a gradle configuration problem to me. In Maven, we can set a glob pattern to select which classes should be considered as tests. I suppose you can do the same in Gradle, but I'm curious why the default didn't work, since it works fine for other modules...

Did the failure occur when you explicitly filtered tests by pattern, by any chance? I the pattern was too inclusive, that could explain why some non-test classes were included.

dreab8: not sure if it a gradle or an arquillian problem and what it is even more strange to me is that it depends on the JDK version used

hibernate-orm-modules/hibernate-orm-modules.gradle
```diff
@@ -39,11 +39,6 @@ ext {
 	fpackStagingDir = file( "target/featurepack" ) //Target build directory for the Feature Pack
 }
 
-if ( JavaVersion.current().isJava11Compatible() ) {
-	logger.warn( '[WARN] Skipping all tests for hibernate-orm-modules due to Gradle issues with JDK 11' )
-	test.enabled = false
-}
```
gsmet: Does it work with JDK 13? Otherwise this condition should be updated, not removed.

dreab8: no it doesn't

dreab8: it seems I found a solution

dreab8:
@Yoann thanks for the review and for the hint

PR #3040, opened 2019-09-17
HHH-13604 Check APIs used after the SessionFactory has been created

fax4ever:
https://hibernate.atlassian.net/browse/HHH-13604

~~I had to comment some code, since regex pattern compile is still used at runtime.~~ not anymore!

At the very moment the rules are active only for `CheckForbiddenAPIAtRuntimeTest` run.
In order to apply the rules to all tests, please remove or comment:
``` java
// TODO HHH-13604 Temporary loading the rules only for CheckForbiddenAPIAtRuntimeTest.
if ( !"CheckForbiddenAPIAtRuntimeTest".equals( clazz.getSimpleName() ) ) {
  return;
}
```
of `org.hibernate.testing.junit4.runtimecheck.BMRuntimeCheckCustomRunner` class.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/session/runtime/check/Book.java
+++ b/hibernate-core/src/test/java/org/hibernate/session/runtime/check/Book.java
@@ -1,0 +1,80 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.session.runtime.check;
+
+import java.util.Objects;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+@Entity
+public class Book {
+
+	@Id
+	private Integer id;
+	private String isbn;
+	private String title;
+	private String author;
+	private Integer copies;
+
+	public Integer getId() {
+		return id;
+	}
+
+	public void setId(Integer id) {
+		this.id = id;
+	}
+
+	public String getIsbn() {
+		return isbn;
+	}
+
+	public void setIsbn(String isbn) {
+		this.isbn = isbn;
+	}
+
+	public String getTitle() {
+		return title;
+	}
+
+	public void setTitle(String title) {
+		this.title = title;
+	}
+
+	public String getAuthor() {
+		return author;
+	}
+
+	public void setAuthor(String author) {
+		this.author = author;
+	}
+
+	public Integer getCopies() {
+		return copies;
+	}
+
+	public void setCopies(Integer copies) {
+		this.copies = copies;
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if ( this == o ) {
+			return true;
+		}
+		Book book = (Book) o;
+		return Objects.equals( getId(), book.getId() ) &&
+				Objects.equals( getIsbn(), book.getIsbn() ) &&
+				Objects.equals( getTitle(), book.getTitle() ) &&
+				Objects.equals( getAuthor(), book.getAuthor() ) &&
+				Objects.equals( getCopies(), book.getCopies() );
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash( id, isbn, title, author, copies );
+	}
+}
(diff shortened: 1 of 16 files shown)


fax4ever:
It seems that skipping all the cases when the `Pattern#regex` is called from some class initialization method at any level, the test works.

@Sanne thanks for help you've provided so far,

Now I'm going to try to apply the same rules to the whole test phase.At the moment I wouldn't know how to do it, but I can explore the case...

fax4ever:
Hi @Sanne, it seems that our investment has borne some fruits.

In fact, activating the Byteman rules we can catch both some readings of annotations and some compations of new regular expressions after that the session factory has been created.

As a follow up, I created: [HHH-13669](https://hibernate.atlassian.net/browse/HHH-13669)

The rules are not active, in order to keep the tests from failing.
To active those, you can comment some code in `BMRuntimeCheckCustomRunner` class.

Meanwhile is it the case to merge this request and close the related issue? Thanks

hibernate-core/src/test/java/org/hibernate/session/runtime/check/Book.java
```diff
@@ -0,0 +1,80 @@
...
+	public void setCopies(Integer copies) {
+		this.copies = copies;
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if ( this == o ) {
+			return true;
+		}
+		Book book = (Book) o;
```
NathanQingyangXu: maybe we should check parameter's class before down casting

hibernate-core/src/test/java/org/hibernate/session/runtime/check/Book.java
```diff
@@ -0,0 +1,80 @@
...
+				Objects.equals( getIsbn(), book.getIsbn() ) &&
+				Objects.equals( getTitle(), book.getTitle() ) &&
+				Objects.equals( getAuthor(), book.getAuthor() ) &&
+				Objects.equals( getCopies(), book.getCopies() );
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash( id, isbn, title, author, copies );
+	}
```
NathanQingyangXu: Given that id field is not generated, it seems the above two methods should be based on it exclusively.

NathanQingyangXu: maybe using getter method is more appropriate for two reasons:

- be consistent with `equals()` method above
- proxy friendly

hibernate-core/src/test/java/org/hibernate/session/runtime/check/CheckAnnotationReadTest.java
```diff
@@ -0,0 +1,53 @@
...
+import javax.persistence.Entity;
+
+import org.hibernate.testing.TestForIssue;
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.junit.Test;
+
+/**
+ * @author Fabio Massimo Ercoli
+ */
+@TestForIssue(jiraKey = "HHH-13604")
```
NathanQingyangXu: Not a big deal, but using space chars right after '(' and before ')' would be consistent with existing coding format

hibernate-core/src/test/java/org/hibernate/session/runtime/check/CheckClassLookupTest.java
```diff
@@ -0,0 +1,62 @@
...
+	public static class ClassLoadedWithinClassInitialization {
+		private static final Class<?> CLASS;
+
+		static {
+			Class<?> readClass = null;
+			try {
+				readClass = Class.forName( "org.hibernate.session.runtime.check.Book" );
+			}
+			catch (ClassNotFoundException e) {
+				throw new RuntimeException( e );
```
NathanQingyangXu: You might well use `throw new ExceptionInInitializerError( e );`

hibernate-core/src/test/java/org/hibernate/session/runtime/check/CheckForbiddenAPIAtRuntimeTest.java
```diff
@@ -0,0 +1,65 @@
...
+		book.setAuthor( "William Blake" );
+		book.setCopies( 1_000_000 );
+
+		doInHibernate( this::sessionFactory, session -> {
+			session.persist( book );
+		} );
+
+		doInHibernate( this::sessionFactory, session -> {
+			Book loaded = session.load( Book.class, 1 );
+			assertEquals( book, loaded );
```
NathanQingyangXu: Now I understand your implementation of `equals` for sake of testing purpose. We might well use `Assert.assertTrue(EqualsBuilder.reflectionEquals(expected,actual));` to avoid providing a non-intuitive `equals` overridden method.

fax4ever:
I'm closing this pull request, I could reopen it if we need it.

PR #3041, opened 2019-09-18
HHH-10398 Allow MOD column naming by strategy

Naros:
https://hibernate.atlassian.net/browse/HHH-10398

### DIFF
--- a/hibernate-envers/src/main/java/org/hibernate/envers/boot/internal/ImprovedModifiedColumnNamingStrategy.java
+++ b/hibernate-envers/src/main/java/org/hibernate/envers/boot/internal/ImprovedModifiedColumnNamingStrategy.java
@@ -1,0 +1,80 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.envers.boot.internal;
+
+import org.hibernate.envers.boot.spi.ModifiedColumnNamingStrategy;
+import org.hibernate.envers.configuration.internal.GlobalConfiguration;
+import org.hibernate.envers.configuration.internal.metadata.MetadataTools;
+import org.hibernate.envers.configuration.internal.metadata.reader.PropertyAuditingData;
+import org.hibernate.envers.internal.tools.StringTools;
+import org.hibernate.mapping.Column;
+import org.hibernate.mapping.Selectable;
+import org.hibernate.mapping.Value;
+import org.hibernate.type.BasicType;
+import org.hibernate.type.ManyToOneType;
+import org.hibernate.type.OneToOneType;
+
+import org.dom4j.Element;
+
+/**
+ * A {@link ModifiedColumnNamingStrategy} that adds modified columns with the following rules:
+ * <ul>
+ *     <li>For basic types, prioritizes audit annotation naming followed by physical column name appended with suffix.</li>
+ *     <li>For associations with single column foreign keys, behaves like basic types.</li>
+ *     <li>For associations with multiple column foreign keys, prioritizes audit annotation naming followed by using property name.</li>
+ *     <li>For embeddables, behaves like associations with multiple column foreign keys</li>
+ * </ul>
+ *
+ * @author Chris Cranford
+ * @since 5.4.6
+ */
+public class ImprovedModifiedColumnNamingStrategy implements ModifiedColumnNamingStrategy {
+	@Override
+	public void addModifiedColumns(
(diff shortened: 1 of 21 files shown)


hibernate-envers/src/main/java/org/hibernate/envers/boot/internal/ImprovedModifiedColumnNamingStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+
+		boolean basicType = value.getType() instanceof BasicType;
+		boolean toOneType = value.getType() instanceof ManyToOneType || value.getType() instanceof OneToOneType;
+
+		if ( basicType || toOneType ) {
+			if ( value.getColumnSpan() == 1 ) {
+				Selectable selectable = value.getColumnIterator().next();
+				if ( selectable instanceof Column ) {
+					// This should not be applied for formulas
+					String columnName = propertyAuditingData.getModifiedFlagName();
```
dreab8: it can be `String columnName;`

hibernate-envers/src/main/java/org/hibernate/envers/boot/internal/ImprovedModifiedColumnNamingStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+							propertyAuditingData.getModifiedFlagName(),
+							columnName
+					);
+
+					return;
+				}
+			}
+		}
+
+		// Default legacy behavior
```
dreab8: I may be wrong but to fall back to the legacy behaviour I think it is necessary to consider the case of `propertyAuditingData.isModifiedFlagNameExplicitlySpecified()` and use `propertyAuditingData.getExplicitModifiedFlagName()` as a column name

hibernate-envers/src/main/java/org/hibernate/envers/boot/internal/LegacyModifiedColumnNamingStrategy.java
```diff
@@ -0,0 +1,47 @@
...
+	@Override
+	public void addModifiedColumns(
+			GlobalConfiguration globalCfg,
+			Value value,
+			Element parent,
+			PropertyAuditingData propertyAuditingData) {
+		String columnName = propertyAuditingData.getModifiedFlagName();
+		if ( propertyAuditingData.isModifiedFlagNameExplicitlySpecified() ) {
+			columnName = propertyAuditingData.getExplicitModifiedFlagName();
+		}
```
dreab8: it is not this changing the legacy behaviour? it seems that before we only passed `propertyAuditingData.getModifiedFlagName()`.

hibernate-envers/src/main/java/org/hibernate/envers/boot/internal/ImprovedModifiedColumnNamingStrategy.java
dreab8: in order to fall back to the legacy behaviour should't it be checked if `propertyAuditingData.isModifiedFlagNameExplicitlySpecified()` and is such a case use `propertyAuditingData.getExplicitModifiedFlagName();` ?

hibernate-envers/src/main/java/org/hibernate/envers/boot/internal/LegacyModifiedColumnNamingStrategy.java
dreab8: probably it is better to change it to
`
String columnName;
		if ( propertyAuditingData.isModifiedFlagNameExplicitlySpecified() ) {
			columnName = propertyAuditingData.getExplicitModifiedFlagName();
		}
		else {
			columnName = propertyAuditingData.getModifiedFlagName();
		}
`

hibernate-envers/src/main/java/org/hibernate/envers/boot/internal/ImprovedModifiedColumnNamingStrategy.java
Naros: That's a good point.  It might even make more sense to have the _Improved_ strategy extend  from the _Legacy_ strategy and then simply delegate to the super implementation in this use case.

hibernate-envers/src/main/java/org/hibernate/envers/boot/internal/LegacyModifiedColumnNamingStrategy.java
Naros: Changed

hibernate-envers/src/main/java/org/hibernate/envers/boot/internal/ImprovedModifiedColumnNamingStrategy.java
Naros: Changed

PR #3043, opened 2019-09-18
HHH-13130 - Provide Gradle-based bytecode enhancement as a task separ…

sebersole:
…ate from the compileJava task

### DIFF
--- a/tooling/hibernate-gradle-plugin/src/main/groovy/org/hibernate/orm/tooling/gradle/EnhancementHelper.java
+++ b/tooling/hibernate-gradle-plugin/src/main/groovy/org/hibernate/orm/tooling/gradle/EnhancementHelper.java
@@ -1,0 +1,178 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.orm.tooling.gradle;
+
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.gradle.api.GradleException;
+import org.gradle.api.Project;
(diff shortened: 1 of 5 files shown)


sebersole:
Cherry-pick from the commit in master

tooling/hibernate-gradle-plugin/src/main/groovy/org/hibernate/orm/tooling/gradle/EnhancementHelper.java
```diff
@@ -0,0 +1,178 @@
...
+		List<URL> urls = new ArrayList<>();
+		for ( File file : runtimeClasspath ) {
+			try {
+				urls.add( file.toURI().toURL() );
+			}
+			catch (MalformedURLException e) {
+				throw new GradleException( "Unable to resolve classpath entry to URL : " + file.getAbsolutePath(), e );
+			}
+		}
+		return new URLClassLoader( urls.toArray( new URL[0] ), Enhancer.class.getClassLoader() );
```
gbadner: The old line in ```HibernatePlugin``` was:

```		return new URLClassLoader( urls.toArray( new URL[urls.size()] ), Enhancer.class.getClassLoader() );```

Was  ```new URL[urls.size()]``` supposed to be changed to ```urls.toArray( new URL[0] )```?

gbadner:
@sebersole, I'm not fluent in Gradle. The [documentation](https://docs.jboss.org/hibernate/orm/5.3/userguide/html_single/Hibernate_User_Guide.html#BytecodeEnhancement-enhancement-gradle) says that the Gradle plugin can be used for build-time enhancement.

I'm not fluent in Gradlew. Does this change make it possible to enhance classes that are already built?

If so, the documentation should be updated.

tooling/hibernate-gradle-plugin/src/main/groovy/org/hibernate/orm/tooling/gradle/EnhancementHelper.java
sebersole: Building an array of size 0 is the new recommended way to do this.  The JVM can already dynamically resize an array very efficiently.

gbadner: OK, good to know.

sebersole:
Thank you IntelliJ :)

gbadner:
> > I'm not fluent in Gradlew. 

The above was an edit mistake. You must have copied it before I changed it.

I am trying to understand the use case so that I can try it out. I'm discussing with Stephen...

PR #3045, opened 2019-09-20
HHH-13633 (5.3) Bugs join-fetching a collection when scrolling with a stateless session using enhancement as proxy

gbadner:
https://hibernate.atlassian.net/browse/HHH-13633

More testing is needed to cover new cases introduced by the new code.

I would like to get a review to make sure this is on the right track.

This issue is being fixed in 5.3 branch first, due to HHH-13634 on master.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
@@ -46,6 +46,7 @@ import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.persister.entity.OuterJoinLoadable;
 import org.hibernate.pretty.MessageHelper;
 import org.hibernate.proxy.HibernateProxy;
+import org.hibernate.proxy.LazyInitializer;
 import org.hibernate.query.spi.ScrollableResultsImplementor;
 import org.hibernate.tuple.entity.EntityMetamodel;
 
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
```diff
@@ -298,28 +299,61 @@ public Object internalLoad(
 			// caller did not request forceful eager loading, see if we can create
 			// some form of proxy
 
+			final boolean allowBytecodeProxy = getFactory()
+					.getSessionFactoryOptions()
+					.isEnhancementAsProxyEnabled();
+
```
dreab8: `StatelessSessionImpl` has a property `allowBytecodeProxy`, no need to recalculate it.

hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
```diff
@@ -298,28 +299,61 @@ public Object internalLoad(
...
+			final boolean entityHasHibernateProxyFactory = persister.getEntityMetamodel()
+					.getTuplizer()
+					.getProxyFactory() != null;
+
```
dreab8: it can be moved inside the `if` statement

hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
```diff
@@ -298,28 +299,61 @@ public Object internalLoad(
...
+						final LazyInitializer li = ( (HibernateProxy) proxy ).getHibernateLazyInitializer();
+						if ( li.isUnwrap() ) {
+							return li.getImplementation();
+						}
+
+						return persistenceContext.narrowProxy( proxy, persister, entityKey, null );
+					}
+
+					// specialized handling for entities with subclasses with a HibernateProxy factory
+					if ( persister.getEntityMetamodel().hasSubclasses() ) {
```
dreab8: can be `entityMetamodel.hasSubclasses()`

hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
gbadner: oh, ok; thanks, I'll get rid of that

hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
gbadner: OK, I'll do that.

hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
gbadner: OK, I'll change this as well.

gbadner:
Superseded by https://github.com/hibernate/hibernate-orm/pull/3066.

PR #3049, opened 2019-09-27
HHH-13645 : StatsNamedContainer#getOrCompute throws NullPointerException when computed value is null

gbadner:
https://hibernate.atlassian.net/browse/HHH-13645


Sanne:
Thanks Gail! Added a comment so to remember the implementation choices and merged.

PR #3050, opened 2019-09-30
HHH-12965 Avoid creating foreign keys between audit and main tables

Naros:
https://hibernate.atlassian.net/browse/HHH-12965?filter=18160

* Fixes a corner case not addressed by HHH-10667
* Avoids creating foreign-key constraints for any many-to-one


dreab8:
rebased and applied upstream, thanks @Naros

PR #3051, opened 2019-09-30
Fix for HHH-13640

sebersole:


sebersole:
I'm going to go ahead and apply this to master.  Afterwards Andrea will apply his test additions.  

@gbadner - from there we can create a PR for you against 5.3 if you wish or you can just cherry-pick his commit and mine

PR #3052, opened 2019-10-01
HHH-13361 Audited entities that use composite identifiers to support

kbenahmedPalm:
Support audited entities using composite identifiers based on nested relations


Naros:
Thanks for the PR but can you re-open the PR against master rather than 5.1, @kbenahmedPalm ?

Naros:
Superseded by PR #3060

PR #3054, opened 2019-10-02
HHH-13640 - Fix initialization of existing proxy association leaf sub…

dreab8:
…class

not sure this is the right solution to fix the problem with the `testExistingProxyAssociationLeafSubclass`


dreab8:
applied upstream

PR #3058, opened 2019-10-04
HHH-13653 Uninitialized entity does not get initialized when a setter…

dreab8:
… is called with enhancement-as-proxy enabled

https://hibernate.atlassian.net/browse/HHH-13653

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementAsProxyLazinessInterceptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementAsProxyLazinessInterceptor.java
@@ -6,7 +6,6 @@
  */
 package org.hibernate.bytecode.enhance.spi.interceptor;
 
-import java.io.Serializable;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementAsProxyLazinessInterceptor.java
```diff
@@ -59,6 +61,9 @@ public EnhancementAsProxyLazinessInterceptor(
 		final EntityPersister entityPersister = session.getFactory().getMetamodel().entityPersister( entityName );
 		this.inLineDirtyChecking = entityPersister.getEntityMode() == EntityMode.POJO
 				&& SelfDirtinessTracker.class.isAssignableFrom( entityPersister.getMappedClass() );
+		// if self-dirty tracking is enabled but DynamicUpdate is not enabled then we need to initialise the entity
+		// 	because the pre-computed update statement contains even not dirty properties and so we need all the values
+		initializeOnRead = !inLineDirtyChecking || !entityPersister.getEntityMetamodel().isDynamicUpdate();
```
gbadner: IIUC, ```initializeOnRead``` indicates that the entity should be initialized before writing an attribute value to the entity.

If that is the case, the field should be named something like ```initializeBeforeWrite```.

dreab8: thanks @gbadner I have applied your suggestion

dreab8:
thanks @gbadner , applied upstream

PR #3059, opened 2019-10-07
HHH-12858 HHH-13432 fix Oracle failing tests

dreab8:
Hi @sebersole, this should fix the tests failing on Oracle but not sure this is the correct/best way to do it


dreab8:
rebased and applied upstream

PR #3060, opened 2019-10-10
HHH-13361 : Allow audited of enntities having nested identifiers

kbenahmedPalm:


hibernate-envers/src/test/java/org/hibernate/envers/test/embeddedid/TestRunner.java
```diff
@@ -0,0 +1,83 @@
...
+import org.hibernate.boot.registry.StandardServiceRegistry;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.envers.test.BaseEnversFunctionalTestCase;
+import org.hibernate.envers.test.Priority;
+import org.hibernate.testing.TestForIssue;
+import org.junit.Before;
+import org.junit.Test;
+
+@TestForIssue(jiraKey = "HHH-13361")
+public class TestRunner extends BaseEnversFunctionalTestCase {
```
Naros: Please rename this test to something like `NestedEmbeddedIdTest`.  It helps better identify what the test is actually trying to accomplish at a glance rather than having to read the code, comments, or the jira to gather that.

hibernate-envers/src/test/resources/mappings/embeddedComponents/embedded.hbm.xml
```diff
@@ -0,0 +1,20 @@
...
+		<property name="connection.username">sa</property>
+		<property name="connection.password"></property>
+		<property name="hibernate.hbm2ddl.auto">update</property>
+		<property name="show_sql">true</property>
+
+		<mapping class="org.hibernate.envers.test.embeddedid.OwnerOfRelationCode" />
+		<mapping class="org.hibernate.envers.test.embeddedid.CompositeEntity" />
+
+	</session-factory>
+</hibernate-configuration>
```
Naros: As mentioned earlier, this is completely unnecessary, so you should be able to safely remove this entirely.  You expose the classes the test should load by overriding the getter methods and the other hibernate property settings are handled automatically by the test suite.

hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/metadata/BasicMetadataGenerator.java
```diff
@@ -6,20 +6,26 @@
  */
 package org.hibernate.envers.configuration.internal.metadata;
 
+import java.util.Iterator;
 import java.util.Properties;
 
+import org.dom4j.Element;
```
Naros: Please do not re-order imports unnecessarily.  This creates unnecessary diffs in the files.  In general, `org.dom4j.` imports are listed below `org.hibernate` and in a separate section on purpose.  This helps call attention to classes that use this library as it is something we plan to remove in the next major release.

Naros:
I think there is a much simpler way to re-use existing mappers without tainting the `SingleIdMapper` behavior to support any type of nested embeddable class.  Specifically, we can translate the existence of a `ComponentType` by adding the ability to chain `EmbeddedIdMapper` instances together.  I've opened PR #4494 which supersedes this with such a solution.

PR #3061, opened 2019-10-11
HHH-13663 Session#setHibernateFlushMode() method not callable without an active transaction

mimmoz81:
This PR addresses issue HHH-13663, by letting setHibernateFlushMode() run even within an inactive transaction, as it's deprecated counterpart setFlushMode() does.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/flush/TestHibernateFlushModeOnInactiveTransaction.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/flush/TestHibernateFlushModeOnInactiveTransaction.java
@@ -1,0 +1,30 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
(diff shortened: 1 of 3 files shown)


hibernate-core/src/test/java/org/hibernate/test/flush/TestHibernateFlushModeOnInactiveTransaction.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Test for issue https://hibernate.atlassian.net/browse/HHH-13663
+ * 
+ * @author Luca Domenichini
+ */
+@TestForIssue(jiraKey = "HHH-13663")
+public class TestHibernateFlushModeOnInactiveTransaction extends BaseCoreFunctionalTestCase {
+
+	@Test
```
Sanne: So the problem is actually only triggered when using thread-local bound Sessions.

I think the test should enable that as it's not the default?

Sanne:
Merged, thanks again! FYI I did make another little change to the test, as it still wasn't using thread-local bound Sessions so the test wouldn't fail ;)

mimmoz81:
which little change, if you wish? (just to not repeat the same error next time..)

mimmoz81:
thanks for your quick review! oh yes, it must use getCurrentSession()... 👍

PR #3062, opened 2019-10-14
HHH-13640 

gbadner:
https://hibernate.atlassian.net/browse/HHH-13640

Deprecated SharedSessionContractImplementor#internalLoad(String, Serializable, boolean, boolean, Boolean unwrapProxy), and added tests.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
@@ -255,6 +255,10 @@ public interface SharedSessionContractImplementor
 	Object internalLoad(String entityName, Serializable id, boolean eager, boolean nullable)
 			throws HibernateException;
 
+	/**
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
```diff
@@ -255,6 +255,10 @@ void initializeCollection(PersistentCollection collection, boolean writing)
 	Object internalLoad(String entityName, Serializable id, boolean eager, boolean nullable)
 			throws HibernateException;
 
+	/**
+	 * @deprecated Use {@link #internalLoad(String, Serializable, boolean, boolean)} instead.
+	 */
+	@Deprecated
 	default Object internalLoad(
```
gbadner: Should this method be removed instead?

dreab8: I think it can be removed

gbadner:
I've removed {{#internalLoad}}, squashed the commits and pushed upstream.

PR #3063, opened 2019-10-14
Fix for HHH-4235

flozano:
This issue is from 2010 and remains unfixed (https://hibernate.atlassian.net/projects/HHH/issues/HHH-4235)

PS: Please backport to 5.4.X? :)


flozano:
Thanks, will it be back-ported to 5.4?

dreab8:
thanks to you for the fix, I pushed it on master branch so it will be part of the next 5.4.7 release

PR #3064, opened 2019-10-14
HHH-13665 selecting an Entity annotated with @Immutalbe but not with @Cachable causes a NPE when use_reference_entries is enabled

dreab8:
https://hibernate.atlassian.net/browse/HHH-13665


Sanne:
thanks @dreab8 and @JWGmeligMeyling !

PR #3070, opened 2019-10-17
HHH-13672 The temporary PersistenceContext of a StatelessSession is not cleared after a refresh operation 

dreab8:
https://hibernate.atlassian.net/browse/HHH-13672


dreab8:
Hi @gbadner,

I think you are right, I'll try to create a test and what about creating a separate Jira for the PC not been cleared when an exception get thrown ?

PR #3074, opened 2019-10-20
HHH-13638

dennisshirazi:
I  rewrote the recursive implementation to an iterative approach. 

I wrote a test case for this as well. Unfortunately I was unable to recreate the issue with my test case. If I add too many values I receive a different StackoverflowError(i.e. 10000 values):
```java.lang.StackOverflowError
	at antlr.BaseAST.toString(BaseAST.java:333)
	at antlr.BaseAST.toStringList(BaseAST.java:341)
	at antlr.BaseAST.toStringList(BaseAST.java:347)
	at antlr.BaseAST.toStringList(BaseAST.java:347)
        ...
 ```
If I only use 5000 values both the new implementation as well as the old recursive implementation work.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/MetadataSources.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/MetadataSources.java
@@ -14,6 +14,7 @@ import java.io.Serializable;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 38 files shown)


dennisshirazi:
Sorry. I accidentally added changes to my pull request that weren't intended. I create a new pull request once I've fixed the problems.

PR #3075, opened 2019-10-20
HHH-13638

dennisshirazi:
I rewrote the recursive implementation to an iterative approach.

I wrote a test case for this as well. Unfortunately I was unable to recreate the issue with my test case. If I add too many values I receive a different StackoverflowError(i.e. 10000 values):
```
	at antlr.BaseAST.toString(BaseAST.java:333)
	at antlr.BaseAST.toStringList(BaseAST.java:341)
	at antlr.BaseAST.toStringList(BaseAST.java:347)
	at antlr.BaseAST.toStringList(BaseAST.java:347)
        ...
```
If I only use 5000 values both the new implementation as well as the old recursive implementation work.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
@@ -6,8 +6,10 @@
  */
 package org.hibernate.hql.internal.ast.util;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Deque;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -105,20 +107,27 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
+
+			AST nextSibling = currentNode.getNextSibling();
+			if ( nextSibling != null ) {
+				nodeDeque.push( nextSibling );
+			}
+
+			AST firstChild = currentNode.getFirstChild();
+			if ( firstChild != null ) {
+				nodeDeque.push( firstChild );
+			}
```
NathanQingyangXu: Seems there is some order difference after the refactoring. Previously nodes are found depth-first, now are found breadth-first. If we swap the above code block #120-123 with code block #125-128, the order should be the same. Not sure whether the node order found matters or not, though.
Overall, pretty good refactoring.

beikov:
I think this is obsolete and can be closed after https://github.com/hibernate/hibernate-orm/pull/3088 was merged. Do you agree?

beikov:
Should be fixed since 5.4.9

PR #3077, opened 2019-10-22
Optimistic locking issue when entities present in a query with LEFT Join

fssouza:
Hello,
I've created a test case reproducing this issue. 
It happens because both entities on the query have the same @Version column (opt_lock). 
Let me describe the test:

- Create entity App (opt lock=0)
- Create entity Installation (opt lock=0)
- Modify App  (opt lock=1)
- Modify App again  (opt lock=2)
- Execute query with `App LEFT Join Installation`
- Obtain the Installation entity using its ID (opt lock=2) -> Wrong, it should be 0
- Make a change to the installation entity and try to commit

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/lock/App.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/lock/App.java
@@ -1,0 +1,55 @@
+package org.hibernate.jpa.test.lock;
+
+import java.util.UUID;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.Version;
+
+@Entity
+public class App {
+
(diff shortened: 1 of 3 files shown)


fssouza:
@brmeyer Hello, I saw your name on "hibernate-test-case-templates" and decided to ask you how to get someone to review the issue that I found. I follow the route that says it is the best: fork + test case

dreab8:
Hi @fssouza ,

sorry for taking so much time to review your PR and thanks for your PR.

Usually the process is first to create a [Jira](https://hibernate.atlassian.net/jira/dashboards/last-visited) where you describe the issue you are facing and then create a PR with a simple reproducer.

I'm trying to reproduce your issue with Hibernate 6.6.3 but with no result so I'm going to close this PR. If you still face the same problem with the latest Hibernate version please open a Jira and a new PR with the reproducer.

Thanks again

PR #3079, opened 2019-10-23
fix for HHH-13651

nyouna:
I added support for null element in embedded collection snapshot equality calculation as it was before (version < 5.4.4 or version < 5.3.11).
This prevent to get a NPE while calculating the equality because the current code try to compute hashcode for null object.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
@@ -67,6 +67,7 @@ public class PersistentBag extends AbstractPersistentCollection implements List 
 	public PersistentBag(SessionImplementor session) {
 		this( (SharedSessionContractImplementor) session );
 	}
+
 	/**
 	 * Constructs a PersistentBag
 	 *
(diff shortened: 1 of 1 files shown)


Sanne:
thanks @nyouna ! I'd merge it but it needs some more steps:

- [ ] use the Hibernate ORM code style
- [ ] the JIRA includes a test; would be nice toadapt it and include it in our regression tests
- [ ] if you could prefix commits with the JIRA code that would be great, like "HHH-13651
 NPE on flushing when ElementCollection field contains null element" rather than "fix for .."

if you could do this, that would be great!

nyouna:
@Sanne Thanks for the comment, i'll fix the 3 points you mentioned ASAP and will update the PR.
About the test case, I wrote it (sorry for using 2 different users for jira (Google IDP) and for github.
I'll already launched it and verified it passes after the fix.
I'll include it in the regression tests. Do you think it is ok to add a test to org.hibernate.test.collection.bag.PersistentBagTest or do you prefer me to add it as a new test class?

nyouna:
@Sanne : remarks 1 and 3 were pushed. just 1 point (and this is the reason for now that I didn't squashed the 2 commits) applying the codestyle to the modified file (PersistentBag) updated a few other lines that are not related to my change. It seems it wasn't applied before. Is it ok for you?

Sanne:
I've merged this and included an adapted version of the test. Thanks!

nyouna:
@Sanne Thanks!

PR #3078, opened 2019-10-23
HHH-13685 Upgrade to Gradle 5

yrodiere:
https://hibernate.atlassian.net/browse/HHH-13685

Before we merge this, we need to upgrade and release three Gradle plugins:

* ~https://github.com/sebersole/gradle-maven-publish-auth/pull/20~ => Done
* ~https://github.com/hibernate/hibernate-matrix-testing/pull/9~ => Done
* ~https://github.com/sebersole/gradle-xjc-plugin/pull/3~ => Done

I checked:

* publishing to the local Maven repository
* publishing snapshots to nexus
* running the matrix plugin

I'd like to run tests on CI before we merge, but we need to release the Gradle plugins first.

PR #3080, opened 2019-10-25
HHH-13689 Replace uses of the deprecated osgi plugin with the biz.aQute.bnd plugin

yrodiere:
https://hibernate.atlassian.net/browse/HHH-13689

**DO NOT MERGE**: This pull request is based on #3078 (upgrade to Gradle 5), which should be merged first. This means the first few commits are those from #3078 , but they will disappear once #3078 is merged and I rebase the PR on master.

The OSGi plugin [is deprecated](https://docs.gradle.org/5.0-milestone-1/release-notes.html#deprecated-osgi-plugin) in Gradle 5, meaning we will have to switch to its replacement before we can think about upgrading to Gradle 6. And we will have to upgrade to Gradle 6 at some point, for JDK13+ support.

This PR addresses two points:

* Switching to the recommended replacement for the OSGi plugin, the [BND Gradle plugin](https://github.com/bndtools/bnd/tree/master/biz.aQute.bnd.gradle), and adapting the configuration (the syntax is a bit different).
* Adding more configuration to remove unwanted `Import-Packages`, unwanted package versions, etc.

I tried to keep the changes in the `MANIFEST.MF` files to a minimum. Compared to 5.4.7.Final, here is what will change:

* In `hibernate-core`, there are two new imported packages (`Import-Package` header): `com.ibm.websphere.jtaextensions` and `org.objectweb.jonas_tm`. The old plugin didn't detect the dependencies, but the new one does: we load classes from these packages dynamically, if they are present, in the transaction manager code. I made the imports optional and removed the version constraints, since I'm not sure what the version constraints are.
* In `hibernate-core`, there is a new `Private-Package` header that references two packages: `org.hibernate.xsd.cfg` and `org.hibernate.xsd.mapping`. From what I can tell, those packages only contain resources (XSD files). They weren't exported in 5.4.7.Final. The `Private-Package` header is just a byproduct of BND and does not change anything in OSGi runtimes (it's ignored).
* In `hibernate-gradle-plugin`, the `Import-Package`s for the Groovy APIs no longer point to a specific version (the "version" bit was removed). As far as I can tell, we don't have an explicit dependency to a specific version of Groovy in our own gradle files, so I think it makes sense? I can add a version manually, but I would need to know which version we want to target. Or, you know, we just leave it as is: I might be wrong, but running a Gradle plugin in an OSGi environment looks pretty niche.
* All OSGi instructions were removed from `javadoc` and `sources` JARs. Apparently they were copied there by mistake. If we wanted to restore them, it would be quite complicated with the new plugin.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,7 @@ buildscript {
 		classpath 'org.asciidoctor:asciidoctor-gradle-plugin:1.5.7'
 		classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.4'
 		classpath 'de.thetaphi:forbiddenapis:2.5'
+		classpath 'biz.aQute.bnd:biz.aQute.bnd.gradle:4.2.0'
 	}
 }
 
--- a/gradle/java-module.gradle
+++ b/gradle/java-module.gradle
@@ -26,7 +26,7 @@ apply from: rootProject.file( 'gradle/libraries.gradle' )
 apply from: rootProject.file( 'gradle/databases.gradle' )
 
 apply plugin: 'java'
-apply plugin: 'osgi'
+apply plugin: 'biz.aQute.bnd.builder'
 
 apply plugin: 'checkstyle'
 apply plugin: 'build-dashboard'
--- a/gradle/published-java-module.gradle
+++ b/gradle/published-java-module.gradle
@@ -10,69 +10,90 @@ apply from: rootProject.file( 'gradle/java-module.gradle' )
 apply from: rootProject.file( 'gradle/publishing-repos.gradle' )
 apply from: rootProject.file( 'gradle/publishing-pom.gradle' )
 
-
-
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 // Jar
 
 jar {
-	manifest = osgiManifest {
-		// GRADLE-1411: Even if we override Imports and Exports
-		// auto-generation with instructions, classesDir and classpath
-		// need to be here (temporarily).
+	manifest {
+		attributes(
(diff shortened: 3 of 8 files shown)


yrodiere:
I just rebased and force-pushed, and checked the generated manifests again, just to be sure.

The changes mentioned in the description of this PR are still valid:

> * In `hibernate-core`, there are two new imported packages (`Import-Package` header): `com.ibm.websphere.jtaextensions` and `org.objectweb.jonas_tm`. The old plugin didn't detect the dependencies, but the new one does: we load classes from these packages dynamically, if they are present, in the transaction manager code. I made the imports optional and removed the version constraints, since I'm not sure what the version constraints are.
> * In `hibernate-core`, there is a new `Private-Package` header that references two packages: `org.hibernate.xsd.cfg` and `org.hibernate.xsd.mapping`. From what I can tell, those packages only contain resources (XSD files). They weren't exported in 5.4.7.Final. The `Private-Package` header is just a byproduct of BND and does not change anything in OSGi runtimes (it's ignored).
> * In `hibernate-gradle-plugin`, the `Import-Package`s for the Groovy APIs no longer point to a specific version (the "version" bit was removed). As far as I can tell, we don't have an explicit dependency to a specific version of Groovy in our own gradle files, so I think it makes sense? I can add a version manually, but I would need to know which version we want to target. Or, you know, we just leave it as is: I might be wrong, but running a Gradle plugin in an OSGi environment looks pretty niche.
> * All OSGi instructions were removed from `javadoc` and `sources` JARs. Apparently they were copied there by mistake. If we wanted to restore them, it would be quite complicated with the new plugin.

You can find the diff [here](https://gist.github.com/yrodiere/234146423b45654cb3e5266c951f6353).

sebersole:
Yes, that all sounds reasonable to me.

sebersole:
I had already run it locally and it passed for me

PR #3083, opened 2019-10-28
fix a boolean logic error in DefaultSaveOrUpdateEventListener

NathanQingyangXu:
I think the following code snippet has obvious logic error:

`final boolean isEqual = !entityEntry.getPersister().getIdentifierType()
						.isEqual( requestedId, entityEntry.getId(), factory );`

From the logging statement (only `requestedId` is used), it seems the boolean variable name is not misnomer. Seems the `!` operator at the beginning is overlooked.


NathanQingyangXu:
@sebersole Could you take a look? Seems an obvious error and correction.

quaff:
@NathanQingyangXu why close without comment?

NathanQingyangXu:
I thought my PR is not important and has been there for a long time without response. I created other PR targetting existing ticket and thought maybe I can focus on that.

PR #3084, opened 2019-10-28
HHH-13614 - Allow the IntegratorProvider to be supplied via its FQN in the JPA persistence.xml

dreab8:
https://hibernate.atlassian.net/browse/HHH-13614

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
@@ -276,37 +276,11 @@ public class EntityManagerFactoryBuilderImpl implements EntityManagerFactoryBuil
 			return;
 		}
 
-		MetadataBuilderContributor metadataBuilderContributor = null;
-		Class<? extends MetadataBuilderContributor> metadataBuilderContributorImplClass = null;
-
-		if ( metadataBuilderContributorSetting instanceof MetadataBuilderContributor ) {
-			metadataBuilderContributor = (MetadataBuilderContributor) metadataBuilderContributorSetting;
-		}
-		else if ( metadataBuilderContributorSetting instanceof Class ) {
-			metadataBuilderContributorImplClass = (Class<? extends MetadataBuilderContributor>) metadataBuilderContributorSetting;
-		}
-		else if ( metadataBuilderContributorSetting instanceof String ) {
-			final ClassLoaderService classLoaderService = standardServiceRegistry.getService( ClassLoaderService.class );
-
-			metadataBuilderContributorImplClass = classLoaderService.classForName( (String) metadataBuilderContributorSetting );
-		}
-		else {
-			throw new IllegalArgumentException(
-					"The provided " + METADATA_BUILDER_CONTRIBUTOR + " setting value [" + metadataBuilderContributorSetting + "] is not supported!"
-			);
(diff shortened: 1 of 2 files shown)


dreab8:
hi @vladmihalcea the PR applies few changes to your https://github.com/hibernate/hibernate-orm/pull/3032

hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
```diff
@@ -1394,4 +1382,51 @@ private void addCacheRegionDefinition(CacheRegionDefinition cacheRegionDefinitio
 			this.cacheRegionDefinitions.add( cacheRegionDefinition );
 		}
 	}
+
+	private <T> T loadSettingInstance(String settingName, Object settingValue, Class<T> clazz) {
+		T instance = null;
+		Class<? extends T> instanceClass = null;
+
+		if ( clazz.isInstance( settingValue ) ) {
```
dreab8: not sure if this should be `if ( clazz.isInstance( settingValue ) ) `or `if ( clazz.isAssignableFrom( settingValue.getClass() ) )`

vladmihalcea: The `isInstance` check allows us to pass an actual `IntegratorProvider` instance that was build programatically. We could support both actual instances and `Class<? extend IntegratorProvider>` that allow us to instantiate the `IntegratorProvider`.

vladmihalcea:
@dreab8 Looks good. Much better to split the logic into dedicated methods.

Sanne:
I'm failing to see the use case. Integrator Providers can be automatically picked up from classpath, the reason is takes an object instance is to allow for this to  work in modular / managed environments - but that wouldn't work with a FQN as you can't set the classloader.

So what's the point to have this?

vladmihalcea:
For Spring Boot, you could define the IntegratorProvider in the application.properties file. Otherwise, you need to do all sorts of hacks to pass an instance.

Sanne:
ok, thanks @vladmihalcea . Weird that they don't use the automatic integration facility but I guess that's their choice.

vladmihalcea:
@Sanne I wrote [several articles](https://vladmihalcea.com/hibernate-event-listeners/) that show the advantages of using the `IntegratorProvider`, and several people commented that they can't use the functionality easily with Spring Boot.

Spring Boot support

That's when I noticed the opportunity for having this enhancement.

hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
dreab8: @vladmihalcea reverted to` clazz.isAssignableFrom( settingValue.getClass())`

dreab8:
added few more tests, rebased and merged upstream.

Thanks a lot @vladmihalcea

vladmihalcea:
@dreab8 Cool! I'm going to update my articles once this will be released.

Sanne:
Nice, thanks @vladmihalcea . Keep in mind that if you declare your Integrator as a manifest, ORM can pick it up without any configuration being needed. For example Hibernate Validator and Hibernate Search are automatically integrated - no need to configure them, one just needs to add the dependency.

PR #3088, opened 2019-10-31
HHH-13670 - Reproducer Missing from clause in query with joined inheritance, regression in 5.4.5

jwgmeligmeyling:
Reproducer for https://hibernate.atlassian.net/browse/HHH-13670, which is an issue with dereferencing attributes for joined inheritance entities in queries introduced with https://hibernate.atlassian.net/browse/HHH-12993 . 

As a result, the tests pass on releases 5.4.6 and before or with `AvailableSettings.OMIT_JOIN_OF_SUPERCLASS_TABLES` disabled.

Test is derived from my test case here https://github.com/Blazebit/blaze-persistence/pull/885

I don't have time to work on a fix now, but I believe @beikov is interested into looking into this.


beikov:
This should fix the issues that people are having with this new optimization

jwgmeligmeyling:
👍  confirmed the fix works.

This should definitely be integrated into the next 5.4.x release imho.

dreab8:
rebased and pushed upstream.
@jwgmeligmeyling  thanks for the reproducer and @beikov thanks for the fix.

Sanne:
awesome work, thanks all!

PR #3091, opened 2019-11-02
Update preface.doc

marcobehler:
I'm currently revising my guide on database access in Java and was thinking - with the risk of it being rejected for self promotion- that including it in the 'Getting Started' guide would help more people get a clearer picture of the whole Java database landscape and where and how Hibernate fits in. 

At least that's the feedback I got so far. In any case, thanks for consideration.


marcobehler:
@sebersole thanks!

Should I just add another commit, moving it to the next paragraph as first line? Or how would you want it?

sebersole:
I went ahead and made the change myself (it is minor) and pushed.  Thanks

PR #3092, opened 2019-11-05
HHH-13705 Enhancement as Proxy with inline dirty checking - flush of an @ManyToOne with an Embedded value having not null properties causes PropertyValueException

dreab8:
https://hibernate.atlassian.net/browse/HHH-13705

An example is https://github.com/hibernate/hibernate-orm/compare/master...dreab8:HHH-13705?expand=1#diff-3e7d7f46b89f78dfa756348772394254 

when the `User` is loaded the associated `Client` results having  fields that are marked as dirty in the dirtyness tracker, this happens when the `Client` is instantiated.

For `User` the fields that are marked as dirty in the dirtyness tracker are cleared when `PojoEntityTuplizer#afterInitialize` calls `( (SelfDirtinessTracker) entity ).$$_hibernate_clearDirtyAttributes();`

this causes an update during the flush.

I hope my explanation is not too messy ;)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/tuple/entity/BytecodeEnhancementMetadataPojoImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/tuple/entity/BytecodeEnhancementMetadataPojoImpl.java
@@ -20,6 +20,7 @@ import org.hibernate.engine.spi.EntityKey;
 import org.hibernate.engine.spi.PersistenceContext;
 import org.hibernate.engine.spi.PersistentAttributeInterceptable;
 import org.hibernate.engine.spi.PersistentAttributeInterceptor;
+import org.hibernate.engine.spi.SelfDirtinessTracker;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.engine.spi.Status;
 import org.hibernate.mapping.PersistentClass;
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/tuple/entity/BytecodeEnhancementMetadataPojoImpl.java
```diff
@@ -140,8 +141,14 @@ public PersistentAttributeInterceptable createEnhancedProxy(EntityKey entityKey,
...
 
 		// first, instantiate the entity instance to use as the proxy
-		final PersistentAttributeInterceptable entity = (PersistentAttributeInterceptable) persister.getEntityTuplizer().instantiate( identifier, session );
+		final EntityTuplizer entityTuplizer = persister.getEntityTuplizer();
+		final PersistentAttributeInterceptable entity = (PersistentAttributeInterceptable) entityTuplizer
+				.instantiate( identifier, session );
 
+		// clear the fields that are marked as dirty in the dirtyness tracker
+		if ( entity instanceof SelfDirtinessTracker ) {
+			( (SelfDirtinessTracker) entity ).$$_hibernate_clearDirtyAttributes();
```
sebersole: But didn't you just instantiate the entity?

dreab8: what I noticed is that when  
```
@Entity(name = "Client")
public static class Client {
        ....
	@Embedded
	private Log log = new Log();
}
```
is instantiated, because the field `log` has an assigned initial value `new Log()` a call of `$$_hibernate_write_log` and then of `$$_hibernate_trackChange` is triggered so the newly instantiated entity results dirty.

dreab8: probably a better approach is to add  to the enhanced class the ability to reset the tracker on exit of the constructor method

sebersole: No, you explained it.. that's reasonable

sebersole: An alternative (same net effect) would be to disable dirty tracking during construction

dreab8:
rebased and merged upstream.

Thanks @gbadner and @sebersole  for the reviews

PR #3094, opened 2019-11-11
HHH-13704 Make Javassist optional

DavideD:
https://hibernate.atlassian.net/browse/HHH-13704

Not sure if I've used the right approach, but now Javassist should be an optional dependency on the classpath.
I've moved the test requiring Javassist in a separate folder (only one really).

I guess the next step would be to run all the existing tests with javassist enabled.


Sanne:
Thanks! 

Not looked properly yet, but I see a file `.gradletasknamecache ` and a `Task` file included..  I guess that was a mistake?

DavideD:
> Not looked properly yet, but I see a file `.gradletasknamecache ` and a `Task` file included.. I guess that was a mistake?

Ah, yes, let me remove them

DavideD:
> > Not looked properly yet, but I see a file `.gradletasknamecache ` and a `Task` file included.. I guess that was a mistake?
> 
> Ah, yes, let me remove them

Fixed

dreab8:
I think we need to add to the migration doc the fact that current users opting for javassist as bytecode provider will need to add the dependency to the javassist lib

DavideD:
I will sent a new PRI will close this and send a new PR

PR #3096, opened 2019-11-12
HHH-13687 TenantSchemaResolver not called in integration test after u…

Sanne:
…pgrade from

This should fix :
 - https://hibernate.atlassian.net/browse/HHH-13687 TenantSchemaResolver not called in integration test after upgrade from
 - https://hibernate.atlassian.net/browse/HHH-13710 Wrong tenant-identifier in Envers temporary session
 - https://hibernate.atlassian.net/browse/HHH-13690 Multi-tenancy supporting session factories can not be created


Sanne:
thanks @dreab8

dreab8:
@Sanne thanks for fixing the issue :)

PR #3097, opened 2019-11-13
HHH-13728 JPQL for Enum with converter

luvarqpp:
Added new test which shows that JPQL with enum with custom converter fails to build correct query.

https://hibernate.atlassian.net/browse/HHH-13728


gavinking:
This test actually passes in H7.

PR #3098, opened 2019-11-14
HHH-13730 Upgrade to Classmate 1.4.0

Sanne:
See:
 - https://hibernate.atlassian.net/browse/HHH-13730
 - https://github.com/wildfly/wildfly/pull/12796


gbadner:
@Sanne, should this be changed to classmate 1.5.1?

Sanne:
@gbadner I've opened a different JIRA issue for 1.5.1: https://hibernate.atlassian.net/browse/HHH-13731

So far I know for sure that we'll upgrade at least to 1.4.0 so this one is a good start.

It seems we might also upgrade to 1.5.1 but that discussion in WildFly needs feedback from some more teams to confirm - either way, I already upgraded Hibernate ORM _5.4_ to 1.5.1 as in that case we don't have to align to the other components.

PR #3100, opened 2019-11-14
HHH-13433 HHH-13737 (5.3)

gbadner:
https://hibernate.atlassian.net/browse/HHH-13433
https://hibernate.atlassian.net/browse/HHH-13737

Once approved, the same test case and logging will get pushed to master.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
@@ -3531,7 +3531,7 @@ public final class SessionImpl
 			throw exceptionConverter.convert( new IllegalArgumentException( e.getMessage(), e ) );
 		}
 		catch ( JDBCException e ) {
-			if ( accessTransaction().getRollbackOnly() ) {
+			if ( accessTransaction().isActive() && accessTransaction().getRollbackOnly() ) {
 				// assume this is the similar to the WildFly / IronJacamar "feature" described under HHH-12472
 				return null;
 			}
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
```diff
@@ -3531,8 +3531,14 @@ public void remove(Object entity) {
...
 		}
 		catch ( JDBCException e ) {
-			if ( accessTransaction().getRollbackOnly() ) {
-				// assume this is the similar to the WildFly / IronJacamar "feature" described under HHH-12472
+			if ( accessTransaction().isActive() && accessTransaction().getRollbackOnly() ) {
+				// Assume this is the similar to the WildFly / IronJacamar "feature" described under HHH-12472.
+				// Just log the exception and return null.
+				if ( log.isDebugEnabled() ) {
+					log.debug( "JDBCException was thrown for a transaction marked for rollback; " +
+									   "this is probably due to IronJacamar failing fast after " +
```
scottmarlow: Would it be okay to generalize the logged message to not mention IronJacamar, in case IronJacamar is not in use, perhaps something like:

```
log.debug( "JDBCException was thrown for a transaction marked for rollback; " +
									   "this is probably due to an operation failing fast due to " +
									   "transaction marked for rollback.", e );

```

gbadner: @scottmarlow , yes, good idea; I'll do that.

I see there's a checkstyle failure, so I'll fix that as well.

gbadner:
@scottmarlow , I just pushed a commit with the change you requested, and a fix for the checkstyle failure.

scottmarlow:
Looks good! :)

gbadner:
@scottmarlow , thanks Scott! I'll leave this one open until it's approved for 5.3 branch. I'll port the HHH-13737 to master.

PR #3102, opened 2019-11-15
HHH-13722 : add test case

rasenderhase:
Hi,
I added a test case for [HHH-13722 ArrayStoreException in Constraint.generateName](https://hibernate.atlassian.net/browse/HHH-13722).
The error occurs during hibernate initialization. So the test just breaks with the exception reported in JIRA.
I started the test with

    sh gradlew :hibernate-core:test --tests org.hibernate.test.mapping.joinformula* --stacktrace

Then the exception can be viewed in `hibernate-orm/hibernate-core/target/reports/tests/test/classes/org.hibernate.test.mapping.joinformula.JoinFormulaTest.html`

Hope this helps.

Regards
Andy


Sanne:
Thanks! I'll check it out soon

Sanne:
I've adapted this a bit (fixed headers mostly), then fixed the issue and merged it all.

PR #3103, opened 2019-11-16
HHH-13365 : Entities in joined subclass table are not inserted with batch size > 0 using sequence-identity ID generator

gbadner:
https://hibernate.atlassian.net/browse/HHH-13365

```SequenceIdentityGenerator``` is deprecated. I'm submitting a PR because the fix is easy, and ```SequenceIdentityGenerator``` is completely unusable with ```hibernate.jdbc.batch_size > 0``` .


dreab8:
rebased and applied upstream
Thanks @gbadner

PR #3104, opened 2019-11-18
HHH-12210 testcase for session closed issue if JTA,  EM and ENABLE_LAZY_LOAD_NO_TRANS is used.

MarkusLutum:
Adds a testcase to reproduce and issue which we have with wildfly 18 and enable lazy load outside transactions


dreab8:
Thanks @MarkusLutum for the test case, I have incorporated it in https://github.com/hibernate/hibernate-orm/pull/3310 with a solution.

PR #3105, opened 2019-11-18
HHH-9662 Support composite nested identity generated values

fax4ever:
https://hibernate.atlassian.net/browse/HHH-9662
https://hibernate.atlassian.net/browse/HHH-13623 (duplicated but it has a nice test case)

So far it is supported only for HSQL db dialect, we need to test more dialects.
I would prefer to have a feedback before to continue.

There are some redundant tests.
EventMilestoneTest is inspired by HHH-13623.


fax4ever:
I'm closing this pull request, I could reopen it if we need it.

PR #3106, opened 2019-11-19
HHH-13355 : StaleStateException for updates to optional secondary table using saveOrUpdate

gbadner:
https://hibernate.atlassian.net/browse/HHH-13355


dreab8:
rebased and applied upstream
thanks @gbadner

PR #3107, opened 2019-11-21
HHH-13742 - Missing from clause with joined inheritance property in association subquery

jwgmeligmeyling:
Reproducer for https://hibernate.atlassian.net/browse/HHH-13742 , working on a fix.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
@@ -8,6 +8,7 @@ package org.hibernate.hql.internal.ast.util;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -31,6 +32,7 @@ import org.hibernate.hql.internal.ast.tree.FromReferenceNode;
 import org.hibernate.hql.internal.ast.tree.ImpliedFromElement;
 import org.hibernate.hql.internal.ast.tree.ParameterContainer;
 import org.hibernate.hql.internal.ast.tree.QueryNode;
+import org.hibernate.hql.internal.ast.tree.SqlFragment;
 import org.hibernate.hql.internal.classic.ParserHelper;
 import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
@@ -154,9 +156,7 @@ public class JoinProcessor implements SqlTokenTypes {
 				if ( entityPersister != null && entityPersister instanceof AbstractEntityPersister ) {
 					AbstractEntityPersister aep = (AbstractEntityPersister) entityPersister;
 					String[] tables = aep.getTableNames();
-					for ( String table : tables ) {
-						result.add( table );
-					}
+					Collections.addAll(result, tables);
 				}
 			}
 		}
@@ -171,10 +171,25 @@ public class JoinProcessor implements SqlTokenTypes {
 				FromReferenceNode fromReferenceNode = (FromReferenceNode) node;
 				String[] tables = fromReferenceNode.getReferencedTables();
 				if ( tables != null ) {
-					for ( String table : tables ) {
-						result.add( table );
+					Collections.addAll(result, tables);
+				}
+			}
+			else if (node instanceof SqlFragment) {
+				SqlFragment sqlFragment = (SqlFragment) node;
+				FromElement fromElement = sqlFragment.getFromElement();
+
+				if (fromElement != null) {
+					// For joins, we want to add the table where the association key is mapped as well as that could be a supertype that we need to join
+					String role = fromElement.getRole();
+					if ( role != null ) {
+						result.add( fromElement.getOrigin().getPropertyTableName(role.substring(role.lastIndexOf('.') + 1)) );
+					}
+					AST withClauseAst = fromElement.getWithClauseAst();
+					if ( withClauseAst != null ) {
+						collectReferencedTables( new ASTIterator( withClauseAst ), result );
 					}
 				}
+
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


jwgmeligmeyling:
@beikov  I think I have a fix based on what you provided in #3088. The key is that we need to consider `FromElements` introduced in `SqlFragment` AST nodes. Could you review my approach? Perhaps there is a way to combine the logic from our both solutions, but I have not figured out how yet.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -154,9 +157,7 @@ public static JoinType toHibernateJoinType(int astJoinType) {
 				if ( entityPersister != null && entityPersister instanceof AbstractEntityPersister ) {
 					AbstractEntityPersister aep = (AbstractEntityPersister) entityPersister;
 					String[] tables = aep.getTableNames();
-					for ( String table : tables ) {
```
beikov: That was on purpose. I wanted to avoid the creation of the intermediate array list.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -171,9 +172,20 @@ private void collectReferencedTables(ASTIterator iterator, Set<String> result) {
 				FromReferenceNode fromReferenceNode = (FromReferenceNode) node;
 				String[] tables = fromReferenceNode.getReferencedTables();
 				if ( tables != null ) {
-					for ( String table : tables ) {
```
beikov: Again, on purpose to avoid the array list

hibernate-core/src/test/java/org/hibernate/query/hhh13670/HHH13670Test.java
```diff
@@ -101,6 +100,42 @@ public void testSubTypeJoinWithTableGroupJoins() {
         });
     }
 
+    @Test
+    public void testSubTypePropertyReferencedFromEntityJoinInSyntheticSubquery() {
+        doInJPA(this::sessionFactory, em -> {
+            List<Tuple> resultList = em.createQuery(
+                    "SELECT  subB_0.id, subA_0.id, subB_0.id, subA_0.id FROM SubB subB_0 INNER JOIN SubA subA_0 ON 1=1 WHERE (EXISTS (SELECT 1 FROM subB_0.parent _synth_subquery_0 WHERE subA_0.id = _synth_subquery_0.id)) ORDER BY subB_0.id ASC, subA_0.id ASC", Tuple.class)
```
beikov: Please also add a testcase for a query like
```
SELECT  subB_0.id, subA_0.id, subB_0.id, subA_0.id FROM SubB subB_0 INNER JOIN SubA subA_0 ON 1=1 WHERE (EXISTS (SELECT 1 FROM Super s WHERE subA_0.id = s.parent.id)) ORDER BY subB_0.id ASC, subA_0.id ASC
```

beikov:
Also there are NPEs introduced by this change: https://travis-ci.org/hibernate/hibernate-orm/jobs/615099875#L1060

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
jwgmeligmeyling: Switched to `Collections.addAll` which also avoids the creation of the intermediate array list. (Aligned with the transition introduced in #2995 )

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
jwgmeligmeyling: Switched to `Collections.addAll` which also avoids the creation of the intermediate array list. (Aligned with the transition introduced in #2995 )

hibernate-core/src/test/java/org/hibernate/query/hhh13670/HHH13670Test.java
jwgmeligmeyling: Added test case

hibernate-core/src/test/java/org/hibernate/query/hhh13670/HHH13670Test.java
```diff
@@ -101,6 +100,52 @@ public void testSubTypeJoinWithTableGroupJoins() {
...
+
+            assertEquals(1, resultList.size());
+        });
+    }
+
+    @Test
+    public void testSubTypePropertyReferencedFromEntityJoinInSyntheticSubquery2() {
+        doInJPA(this::sessionFactory, em -> {
+            List<Tuple> resultList = em.createQuery(
+                    "SELECT  subB_0.id, subA_0.id, subB_0.id, subA_0.id FROM SubB subB_0 INNER JOIN SubA subA_0 ON 1=1 WHERE (EXISTS (SELECT 1 FROM Super s WHERE subA_0.id = s.parent.id)) ORDER BY subB_0.id ASC, subA_0.id ASC", Tuple.class)
```
beikov: I think what I actually meant was `SELECT  subB_0.id, subA_0.id, subB_0.id, subA_0.id FROM SubB subB_0 INNER JOIN SubA subA_0 ON 1=1 WHERE (EXISTS (SELECT 1 FROM Super s WHERE s.id = subA_0.parent.id)) ORDER BY subB_0.id ASC, subA_0.id ASC` to test if an expression referring to a superclass property appearing just in the subquery WHERE clause also triggers the addition of the superclass table

jwgmeligmeyling: I have ran that query as well and it returns the expected empty row set.

I've also ran `SELECT subB_0.id, subA_0.id, subB_0.id, subA_0.id FROM SubB subB_0 INNER JOIN SubA subA_0 ON 1=1 WHERE (EXISTS (SELECT 1 FROM Super s WHERE s.id = subB_0.parent.id)) ORDER BY subB_0.id ASC, subA_0.id ASC` , which correctly returns 6 rows.

dreab8:
applied upstream.

Thanks @jwgmeligmeyling for the fix and @beikov for the review

PR #3108, opened 2019-11-21
Bring AbstractProductQuery#getSingleResult in line with the spec for some databases

TomyLobo:
Due to the way the elements are compared, the method has come into alignment with the spec by accident, reporting a NonUniqueResultException as prescribed by the spec for cases where the query returns multiple rows.

One exception to this is HSQL.
Under certain circumstances, a HSQL query can return the same (==) object twice.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/internal/AbstractProducedQuery.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/internal/AbstractProducedQuery.java
@@ -1572,10 +1572,16 @@ public abstract class AbstractProducedQuery<R> implements QueryImplementor<R> {
 	public R getSingleResult() {
 		try {
 			final List<R> list = list();
-			if ( list.size() == 0 ) {
-				throw new NoResultException( "No entity found for query" );
+			switch ( list.size() ) {
+				case 0:
+					throw new NoResultException( "No entity found for query" );
+
+				case 1:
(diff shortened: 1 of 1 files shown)


TomyLobo:
Well, whatever database you are using for the test cases was triggering the non-conformity of AbstractProductQuery#getSingleResult that I mentioned in the description.

The easy fix would be to put a DISTINCT in that statement.
But is it the correct fix?

TomyLobo:
I looked into it some more.
If I look at this test case in the debugger:
https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/test/java/org/hibernate/test/collection/map/PersistentMapTest.java#L269-L271
and evaluate s.createQuery(that HQL statement).list(), I get a list with 2 elements.
If I do the same but remove the "JOIN FETCH s.map", I get only one element.

Now from my understanding, JOIN FETCH is supposed to fetch the 2 persisted LocalizedStrings into the "map" field of a *single* instance of the MultilingualString class.
According to the test case, successfully does that, but in addition to that, it returns the same (==) object twice.

I do not yet have an explanation for that.
Might even be a bug in Hibernate for all I know.
If anyone could help me figure out why this is happening, that would be helpful.

gavinking:
The code is out of date, there is no issue, and no test case. Closing.

PR #3110, opened 2019-11-25
HHH-13675 HHH-13651 (5.3)

gbadner:
https://hibernate.atlassian.net/browse/HHH-13675
https://hibernate.atlassian.net/browse/HHH-13651


gbadner:
Created against master by mistake.

PR #3111, opened 2019-11-25
HHH-13675 HHH-13651 (5.3)

gbadner:
https://hibernate.atlassian.net/browse/HHH-13675
https://hibernate.atlassian.net/browse/HHH-13651

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
@@ -202,9 +202,23 @@ public class PersistentBag extends AbstractPersistentCollection implements List 
 		}
 		Map<Integer, List<Object>> map = new HashMap<>();
 		for (Object o : searchedBag) {
-			map.computeIfAbsent( elementType.getHashCode( o ), k -> new ArrayList<>() ).add( o );
+			map.computeIfAbsent( nullableHashCode(o , elementType), k -> new ArrayList<>() ).add( o );
 		}
 		return map;
+	}
+
+	/**
+	 * @param o
+	 * @param elementType
+	 * @return the default elementType hashcode of the object o, or null if the object is null
+	 */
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
```diff
@@ -196,7 +199,28 @@ public boolean equalsSnapshot(CollectionPersister persister) throws HibernateExc
...
+		}
+		return map;
+	}
+
+	/**
+	 * @param o
+	 * @param elementType
+	 * @return the default elementType hashcode of the object o, or null if the object is null
+	 */
+	private Integer nullableHashCode(Object o, Type elementType) {
```
NathanQingyangXu: seems should be a static method

hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentBag.java
```diff
@@ -196,7 +199,28 @@ public boolean equalsSnapshot(CollectionPersister persister) throws HibernateExc
 	 * @return Map of "equality" hashCode to List of objects
 	 */
 	private Map<Integer, List<Object>> groupByEqualityHash(List<Object> searchedBag, Type elementType) {
-		return searchedBag.stream().collect( Collectors.groupingBy( elementType::getHashCode ) );
+		if ( searchedBag.isEmpty() ) {
+			return Collections.emptyMap();
+		}
+		Map<Integer, List<Object>> map = new HashMap<>();
+		for ( Object o : searchedBag ) {
+			map.computeIfAbsent( nullableHashCode( o, elementType ), k -> new ArrayList<>() ).add( o );
```
NathanQingyangXu: using LinkedList seems not a bad idea, btw. It would simulate the hashmap bucket implementation as much as possible and avoid ArrayList sizing dilemma.

PR #2546, opened 2018-09-27
HHH-12996 - Reproducer for missing property paths of IdClass attributes

jwgmeligmeyling:
https://hibernate.atlassian.net/browse/HHH-12996

PR #3114, opened 2019-11-26
HHH-13752 - Test and fix deletion of entities with many-to-many assocations using non-primary keys for join table

beikov:
https://hibernate.atlassian.net/browse/HHH-13752

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/exec/DeleteExecutor.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/exec/DeleteExecutor.java
@@ -7,6 +7,7 @@
 package org.hibernate.hql.internal.ast.exec;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import org.hibernate.HibernateException;
@@ -23,6 +24,7 @@ import org.hibernate.persister.collection.AbstractCollectionPersister;
 import org.hibernate.persister.entity.Queryable;
 import org.hibernate.sql.Delete;
 import org.hibernate.type.CollectionType;
+import org.hibernate.type.ComponentType;
 import org.hibernate.type.Type;
 
 import org.jboss.logging.Logger;
@@ -58,7 +60,8 @@ public class DeleteExecutor extends BasicExecutor {
 				final SqlGenerator gen = new SqlGenerator( factory );
 				gen.whereClause( whereClause );
 				parameterSpecifications = gen.getCollectedParameters();
-				idSubselectWhere = gen.getSQL().length() > 7 ? gen.getSQL() : "";
+				String sql = gen.getSQL();
+				idSubselectWhere = sql.length() > 7 ? sql : "";
 			}
 			else {
 				parameterSpecifications = new ArrayList<>();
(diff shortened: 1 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/exec/DeleteExecutor.java
```diff
@@ -74,20 +77,34 @@ public DeleteExecutor(HqlSqlWalker walker, Queryable persister) {
...
-								&& notSupportingTuplesInSubqueries ) {
+						Type keyType = cPersister.getKeyType();
+						String[] columnNames;
+						if ( keyType.isComponentType() ) {
+							ComponentType componentType = (ComponentType) keyType;
+							List<String> columns = new ArrayList<>( componentType.getPropertyNames().length );
+							for ( String propertyName : componentType.getPropertyNames() ) {
+								Collections.addAll( columns, persister.toColumns( propertyName ) );
+							}
+							columnNames = columns.toArray( new String[columns.size()] );
```
NathanQingyangXu: In modern JVM, usually `columns.toArray(new String[0])` is faster than `columns.toArray(new String[columns.size()]`. See https://stackoverflow.com/questions/174093/toarraynew-myclass0-or-toarraynew-myclassmylist-size

dreab8:
applied @NathanQingyangXu suggested and pushed upstream
Thanks @beikov for the work and @NathanQingyangXu for the review

dreab8:
the PR is causing the failure of `org.hibernate.test.subselect.CompositeIdTypeBindingTest
#testCompositeTypeBinding()` on PostgreSQL

jwgmeligmeyling:
`org.hibernate.test.subselect.CompositeIdTypeBindingTest #testCompositeTypeBinding()` is a test that is excluded for H2, hence it went unnoticed. The failure is "unrelated" to being  ran on PostgreSQL.

The issue is that the logic looks up a property in the EntityPersisters PropertyMapping, which is absent. This is likely due to [HHH-12996](https://hibernate.atlassian.net/browse/HHH-12996) (reproducer in https://github.com/hibernate/hibernate-orm/pull/2546). Due to that bug, properties part op a compound key mapped by an `IdClass` are not properly represented in the metamodel / PropertyMapping, which may cause to this property being missing here as well. It so happens to be an `IdClass` mapped entity that fails this test.

As a workaround, the property can be retrieved as part of the magic "id" attribute, i.e. lookup `id.isbn` rather than `isbn`. However, this requires us to reliably detect whether the referenced side of the association is mapped by a IdClass entity...

I'm not sure how to address this properly. However, it is unrelated to the used dialect. This fix seems inherently not to work for IdClass mappings.

jwgmeligmeyling:
In hindsight, we should be able to work around HHH-12996  by just falling back to the old behaviour (which supports PK based mappings only) if the property is not found in the property mapping due to the poor representation of IdClass attributes in the metamodel (which also implies that we want to reference the PK anyway).

jwgmeligmeyling:
@beikov You missed something while cherry-picking 😉  Jenkins passes fine on my original branch:  https://github.com/hibernate/hibernate-orm/compare/master...jwgmeligmeyling:HHH-13752?expand=1#diff-3a4887cb66f00af2b2788321fa0389e3R94

beikov:
Hey @dreab8 the PR should be fine now for PostgreSQL. Whenever you get to it, would be nice if you could integrate that for the next 5.4 release

dreab8:
Applied upstream

Thanks @beikov

PR #3116, opened 2019-11-27
HHH-13704 Makes Javassist optional

DavideD:
[HHH-13704](https://hibernate.atlassian.net/browse/HHH-13704)
Follows: https://github.com/hibernate/hibernate-orm/pull/3094 

Now javassist is only needed when you want to use it as Bytecode provider.
I've cleaned up the changes I made before and added an integration test to check the module on WildFly.

@Sanne I checked and the existing test suite calls the part of the code in ORM that I've updated.


DavideD:
I would like to remove javassist from the hibernate-core jboss module as well but I couldn't find a way to make the tests work otherwise. I think it's fine for now and we could work to remove javassist in ORM 6 as soon as possible.

hibernate-core/src/test/java/org/hibernate/optional/javassist/OptionalJavassistDependencyTest.java
```diff
@@ -0,0 +1,26 @@
...
+public class OptionalJavassistDependencyTest {
+
+	@Rule
+	public ExpectedException thrown = ExpectedException.none();
+
+	@Test
+	public void shouldNotBeOnTheClasspath() throws Exception {
+		thrown.expect( ClassNotFoundException.class );
+
+		Class<?> aClass = getClass().getClassLoader().loadClass( "javassit.Classpath" );
```
Sanne: this test will always succeed, even when javassist is on classpath as you have a typo in the package name :)

hibernate-orm-modules/src/test/resources/arquillian.xml
```diff
@@ -23,14 +23,9 @@
         <container qualifier="container.active-1" mode="suite" default="true">
             <configuration>
                 <property name="jbossHome">${wildFlyInstallDir}</property>
-                <!-- For Remote debugging of Wildfly add the following line to the 'javaVmArguments' section below: -->
-                <!-- -Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y-->
-                <!-- DO NOT add comments within the property section below as the Arquillian parser can't deal with it -->
```
Sanne: ok for the change but could you keep the warning about the comments?

hibernate-orm-modules/src/test/java/org/hibernate/wildfly/integrationtest/JavassistHibernateModulesOnWildflyTest.java
```diff
@@ -0,0 +1,107 @@
...
+		// EntityManager methods exposed through Session only as of 5.2
+		Kryptonite loaded = session.find( Kryptonite.class, 1L );
+
+		assertThat( loaded.description, equalTo( "Some Kryptonite" ) );
+	}
+
+	@Test
+	public void shouldBeAbleToCreateProxyWithJavassist() {
+		Properties properties = new Properties();
+		properties.setProperty( AvailableSettings.BYTECODE_PROVIDER, Environment.BYTECODE_PROVIDER_NAME_JAVASSIST );
```
Sanne: this relates to my other comment regarding `BYTECODE_PROVIDER` ^

hibernate-orm-modules/src/test/java/org/hibernate/wildfly/integrationtest/JavassistHibernateModulesOnWildflyTest.java
```diff
@@ -0,0 +1,107 @@
...
+					.createPersistenceUnit()
+						.name( "primary" )
+						.transactionType( PersistenceUnitTransactionType._JTA )
+						.jtaDataSource( "java:jboss/datasources/ExampleDS" )
+						.getOrCreateProperties()
+							// We want to use the ORM from this build instead of the one coming with WildFly
+							.createProperty().name( "jboss.as.jpa.providerModule" ).value( "org.hibernate:" + ORM_MINOR_VERSION ).up()
+							.createProperty().name( "hibernate.hbm2ddl.auto" ).value( "create-drop" ).up()
+							.createProperty().name( "hibernate.allow_update_outside_transaction" ).value( "true" ).up()
+							.createProperty().name( AvailableSettings.BYTECODE_PROVIDER ).name( Environment.BYTECODE_PROVIDER_NAME_JAVASSIST ).up()
```
Sanne: Are you sure you can set `BYTECODE_PROVIDER` this way? This property is read only from system properties when the Hibernate ORM classes are first initialized.

You seem to have noticed below, as the test is setting it in a different way.

I'd suggest to use the Arquillian capability to set JVM properties.

migration-guide.adoc
```diff
@@ -9,6 +9,11 @@ earlier versions, see any other pertinent migration guides as well.
 
 == Known changes
 
+=== Optional Javassist
+
+Javassist is not a transitive dependencies of Hibernate ORM core anymore. If you are using it in your
```
Sanne: s/a transitive dependencies/a transitive dependency/

hibernate-orm-modules/src/test/resources/arquillian.xml
DavideD: I will just remove the commit

hibernate-orm-modules/src/test/java/org/hibernate/wildfly/integrationtest/JavassistHibernateModulesOnWildflyTest.java
DavideD: How would you set it on Arquillian?

DavideD: Forget about it, silly question

Sanne: https://github.com/hibernate/hibernate-orm/blob/9b08e6bc290d5e35dfa2344dead19f1918ce11c9/hibernate-orm-modules/src/test/resources/arquillian.xml#L28

DavideD:
@Sanne I think I've fixed all the issue

hibernate-orm-modules/src/test/java/org/hibernate/wildfly/integrationtest/JavassistHibernateModulesOnWildflyTest.java
DavideD: I'm not going to update the arquillian.xml because I don't want the change to affect the other tests as well. I've also added a coment in the test.

dreab8:
@Sanne @DavideD I do not think this change can be applied to a minor release, it will break app using javassist for enhancement

DavideD:
> @Sanne @DavideD I do not think this change can be applied to a minor release, it will break app using javassist for enhancement

It makes sense. I will let you gyu decide when to include it, ping me when you need it rebased

DavideD:
Thanks to @dreab8, I think we  fixed the problem with the build. The build was failing because karaf has a transitive dependency to Javassit. Add ing the bundle to the OSGi test solved the issue.

DavideD:
@dreab8 maybe update the [issue](https://hibernate.atlassian.net/browse/HHH-13704) to `Fix version: 5.5.0`

Sanne:
thanks the OSGi tests seem fixed now.

@DavideD no we're not planning a 5.5 release. Let's adapt this to make it better suited for 5.4, proposal:

I'd like to take the changes as, with one exception: Javassist still listed as an actual runtime dependency in the produced pom files.

- We know this code works in case of an exclusion
- We could keep the nice tests to proctect from regressionsby having an exclusion during the integration tests?
 - I can add an exclusion in the Quarkus pom...
 - marking the module dependency as "optional" is a change we can keep: Wildfly can choose to keep the module if they prefer.

WDYT? I'm going to needs this as it's a blocker for:
 - https://github.com/quarkusio/quarkus/issues/6020#issue-534537964

DavideD:
@Sanne I'm checking if it's doable

DavideD:
I've added a commit making Javassist a compile dependency again.  We should create an issue to revert the commit for the next Minor release.

Sanne:
merged! thanks all

PR #3117, opened 2019-11-27
HHH-13755 Update Hibernate Gradle plugin in the documentation

DavideD:
[HHH-13755](https://hibernate.atlassian.net/browse/HHH-13755)

A small update to the example in the documentation to make it clearer what you need and which options you can set when using the plugin


Sanne:
merged. thanks!

PR #3119, opened 2019-11-29
HHH-13726 Extract 'prepareEventListeners' method out of SessionFactor…

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13726

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataImpl.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.regex.Pattern;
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataImpl.java
```diff
@@ -355,6 +360,39 @@ public void validate() throws MappingException {
...
+
+	private Object instantiate(String listenerImpl, ClassLoaderService classLoaderService) {
+		try {
+			return classLoaderService.classForName( listenerImpl ).newInstance();
+		}
+		catch (Exception e) {
+			throw new HibernateException( "Could not instantiate requested listener [" + listenerImpl + "]", e );
+		}
+	}
+
```
NathanQingyangXu: The above code was moved from SessionFactoryImpl without any code change.

PR #3121, opened 2019-11-29
HHH-13758 - Limit Handler for SQL server doesn't work with CTE queries with strings literals

jasonex7:
Fixes [HHH-13758](https://hibernate.atlassian.net/browse/HHH-13758)


Sanne:
nice catch. Thanks!

jasonex7:
> nice catch. Thanks!

your welcome. 

How could I contribute with a backport to 5.2 branch? This is my first contribution to this project.

Sanne:
We're very conservative with older branches, and 5.2 is not even longer supported so that would be a bit unusual. Can't you upgrade?

In case you really want to get it backported, the first step of the process is to ask nicely to @gbadner as she is responsible for all the previous branches.

jasonex7:
Thanks for the information. It's OK for me. I found a workaround for previous versions with a custom dialect based on the previous one but with the fix.

PR #3124, opened 2019-12-03
HHH-13759 InlineDirtyChecking causes exception when embedded field is private in a superclass

DavideD:
https://hibernate.atlassian.net/browse/HHH-13759
Original issue: https://github.com/quarkusio/quarkus/issues/5756

This patch will make sure that there is no expcetion and it will behave the same way as when we put the `@Embedded` annotation on the getter/setter.

This lead me to create another issue for that scenario: https://hibernate.atlassian.net/browse/HHH-13764

Maybe somebody with more experince with ByteBuddy can suggest a solution. I think we should intercept the call to the setter but I haven't figure out how to do that yet.

### DIFF
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -27,6 +27,8 @@ ext {
 
     agroalVersion = '1.7'
 
+    assertjVersion = '3.14.0'
+
     geolatteVersion = '1.4.0'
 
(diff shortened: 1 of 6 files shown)


DavideD:
Pushed force: The tests are now under `org.hibernate.orm.test.bytecode.enhance.internal.bytebuddy`

DavideD:
I realized that this solution doesn't work. I've sent another PR for HHH-13764: https://github.com/hibernate/hibernate-orm/pull/3127

I thought they were related but the are not.

PR #3125, opened 2019-12-03
Wip/6.0 -  Avoid extra join for ToOne with JoinTable

dreab8:
Hi @sebersole, short description of the PR.

A the method `forceTableReferenceJoinRendering`  is added to `Predicate` and `Expression`

This method is called when a predicate or  an expression is added to the `QuerySpec` (selection,where clause, limit...) and by `SingularAssociationAttributeMapping#createTableGroupJoin` on the join predicate (we need to render the join in such a case).

A `TableReference` field is added to `ColumnReference` so that when `ColumnReference#forceTableReferenceJoinRendering` is called  the new `Tablereference#forceRendering` field is set to true.

This field `Tablereference#forceRendering` is used by the `AbstractSqlAstWalker#renderTableReferenceJoins` method to check if `tableJoin.getJoinedTableReference()` should be rendered in the generated sql.

Any suggestion is more than welcome :)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
@@ -419,7 +419,9 @@ public class LoaderSelectBuilder {
 		referencedMappingContainer.visitFetchables( processor, null );
 
 		return fetches;
-	}	private SelectStatement generateSelect(SubselectFetch subselect) {
+	}
+
+	private SelectStatement generateSelect(SubselectFetch subselect) {
 		// todo (6.0) : i think we may even be able to convert this to a join by piecing together
 		//		parts from the subselect-fetch sql-ast..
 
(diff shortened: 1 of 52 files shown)


sebersole:
>> when a predicate or an expression is added to the QuerySpec (selection,where clause, limit...) 

What about `select max( a.zip ) from Address a`?  Or `select trim( a.zip ) from ...`?  Or `select case when a.zip = '12345' then "Xanadu" else "Neverland" end from ...`?  Or ...

The problem is its not always the ColumnReference (expression) itself that is added as a selection.  Sometimes the ColumnReference is a "sub expression" of an expression that is added as a selection.  True, this could be handled by "propagation", but that's pretty fugly.  I think we ought to investigate the option we've discussed before where we lazily create the `TableReferenceJoin` references *as they are needed* - meaning when we ask the `TableGroup` to resolve a `TableReference`.  I think that is an accurate "trigger" for knowing whether we need to render a `TableReference(Join)` when we render the SQL.  Its more "natural" too imo - it better matches the fact that whether to render a particular `TableReference(Join)` really comes down to whether that table was referenced.  The TableReference is requested from the TableGroup, that is an accurate definition of it being referenced I think.  At least I cannot think of a moment where we get a TableReference from the TableGroup where we are not looking to find or create a ColumnReference.  And the join-column references are different - they would not be created unless the particular TableReference(Join) needs to be created during `TableGroup#resolveTableReference`.

dreab8:
The subExpression case is managed by the PR by "propagation" but the case of functions is not managed.
Anyway I agree we have to explore your idea

dreab8:
rejected

PR #3130, opened 2019-12-04
HHH-13766 Add test for issue

dreab8:
Test case for https://hibernate.atlassian.net/browse/HHH-13766

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/inlinedirtychecking/AttributeConverterTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/inlinedirtychecking/AttributeConverterTest.java
@@ -1,0 +1,183 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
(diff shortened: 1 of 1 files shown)


hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/inlinedirtychecking/AttributeConverterTest.java
```diff
@@ -0,0 +1,183 @@
...
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.validation.constraints.NotNull;
+
+import org.hibernate.boot.MetadataSources;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.cfg.AvailableSettings;
+
```
NathanQingyangXu: Do we need the above empty line?

dreab8: not a real need, added by IJ when formatting the code.

yrodiere:
I rebased on `main`, and it seems the test no longer fails.

Let's see how the build goes, and if that's still true, let's merge the test and consider the problem fixed in 5.6? It was probably fixed in a previous version, but I'd rather work on fixing other bugs than waste my time looking for the exact version that fixed this one :)

PR #3131, opened 2019-12-05
Remove mention of Oracle and DB2 not being in MC

aguibert:
Both Oracle and DB2 JDBC drivers are now available in MC so the statement is no longer accurate


aguibert:
Jira issue: https://hibernate.atlassian.net/browse/HHH-13767

PR #3133, opened 2019-12-06
HHH-10844 Resolve columnDefinition to appropriate sql-type for audit mappings

Naros:
https://hibernate.atlassian.net/browse/HHH-10844

### DIFF
--- a/hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/metadata/MetadataTools.java
+++ b/hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/metadata/MetadataTools.java
@@ -9,6 +9,10 @@ package org.hibernate.envers.configuration.internal.metadata;
 import java.util.Iterator;
 import javax.persistence.JoinColumn;
 
+import org.hibernate.boot.Metadata;
+import org.hibernate.dialect.Dialect;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.envers.internal.EnversMessageLogger;
 import org.hibernate.envers.internal.tools.StringTools;
 import org.hibernate.mapping.Column;
 import org.hibernate.mapping.Formula;
@@ -18,12 +22,20 @@ import org.dom4j.Attribute;
 import org.dom4j.Document;
 import org.dom4j.Element;
 
+import org.jboss.logging.Logger;
+
 /**
  * @author Adam Warski (adam at warski dot org)
  * @author Lukasz Antoniak (lukasz dot antoniak at gmail dot com)
  * @author Michal Skowronek (mskowr at o2 dot pl)
  */
 public final class MetadataTools {
+
+	private static final EnversMessageLogger LOG = Logger.getMessageLogger(
+			EnversMessageLogger.class,
+			MetadataTools.class.getName()
+	);
+
 	private MetadataTools() {
 	}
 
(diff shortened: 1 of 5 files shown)


hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/metadata/MetadataTools.java
```diff
@@ -285,38 +297,68 @@ public static Element createJoin(
 		return joinMapping;
 	}
 
-	public static void addColumns(Element anyMapping, Iterator selectables) {
+	public static void addColumns(Element anyMapping, Iterator<?> selectables, Metadata metadata) {
```
NathanQingyangXu: I checked the codebase and found we can improve our generics usage here by changing the second parameter type from `Iterator` to `Iterator<Selectable>`.

hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/metadata/MetadataTools.java
```diff
@@ -285,38 +297,68 @@ public static Element createJoin(
...
+		addColumns( anyMapping, selectables, metadata, metadata.getDatabase().getDialect() );
+	}
+
+	public static void addColumns(Element anyMapping, Iterator<?> selectables, Mapping mapping, Dialect dialect) {
 		while ( selectables.hasNext() ) {
 			final Selectable selectable = (Selectable) selectables.next();
 			if ( selectable.isFormula() ) {
 				throw new FormulaNotSupportedException();
```
NathanQingyangXu: I checked the usage of Selectable and found there is a hidden logic that if `isFormula` is false, it would be of `Column` type (including subtype), but this seems a little bit weird and deciding whether it is `Column` or not would be more straightforward. Ignore my message if my concern is invalid for knowledge gap reason.

hibernate-envers/src/test/java/org/hibernate/envers/test/integration/basic/BasicTypeColumnDefinitionTest.java
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	@Test
+	public void testRevisionHistory() {
+		assertEquals( 2, getAuditReader().getRevisions( BasicTypeContainer.class, 1 ).size() );
+
+		final BasicTypeContainer rev1 = getAuditReader().find( BasicTypeContainer.class, 1, 1 );
+		assertEquals( "test", rev1.getData() );
+		assertEquals( "1234567890", rev1.getColumnWithDefinition() );
+		assertEquals( Integer.valueOf( 1 ), rev1.getCaseNumber() );
```
NathanQingyangXu: could we rely on autoboxing above without explicitly creating a boxing type?

Naros: Doing so would lead to an ambiguous call to `assertEquals` as the last argument is `Integer` and the former would be an `int.

hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/metadata/MetadataTools.java
Naros: That leads to a compilation failure inside `AuditMetadataGenerator` at line 636 where we pass in an explicit `Iterator<Column>` argument rather than `Iterator<Selectable>`; hence the `?`.

hibernate-envers/src/test/java/org/hibernate/envers/test/integration/basic/BasicTypeColumnDefinitionTest.java
NathanQingyangXu: thanks for letting me know. Yeah, it does make sense.

hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/metadata/MetadataTools.java
NathanQingyangXu: Thanks for the response

hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/metadata/MetadataTools.java
Naros: We actually use the `#isFormula()` pattern in quite a number of places across the code base to differentiate between whether a `Selectable` is a formula or column sub-type.  I don't believe there is any reason to change whats here unless I'm misunderstanding your point here.

NathanQingyangXu: You didn't misunderstand my point. Thanks for the explanation.

dreab8:
applied upstream.

Thanks @Naros

PR #3134, opened 2019-12-07
HHH-13644 NullPointerException when calling StoredProcedureQuery.getResultStream() instead of StoredProcedureQuery.getResultList()

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13644
The root cause is ProcedureCallImpl's getResultStream() method should inherit **javax.persistence.Query**'s default method, rather than **org.hibernate.query.Query**'s default method which is only meant for HQL.

An effective testing case was included. Let me know if we need other db dialects.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/procedure/internal/ProcedureCallImplTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/procedure/internal/ProcedureCallImplTest.java
@@ -1,0 +1,39 @@
+package org.hibernate.procedure.internal;
+
+import java.util.stream.Stream;
+import javax.persistence.EntityManager;
+import javax.persistence.Query;
+import org.hibernate.dialect.H2Dialect;
+import org.hibernate.jpa.test.BaseEntityManagerFunctionalTestCase;
+import org.hibernate.testing.RequiresDialect;
+import org.hibernate.testing.TestForIssue;
+import org.junit.Assert;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/procedure/internal/ProcedureCallImplTest.java
```diff
@@ -0,0 +1,39 @@
...
+	@Test
+	@TestForIssue( jiraKey = "HHH-13644" )
+	@RequiresDialect( H2Dialect.class )
+	public void testNoNullPointerExceptionThrown() {
+
+		EntityManager em = getOrCreateEntityManager();
+
+		em.getTransaction().begin();
+
+		em.createNativeQuery("CREATE ALIAS GET_RANDOM_VALUE FOR \"java.lang.Math.random\";").executeUpdate();
```
NathanQingyangXu: The stored procedure native SQL is vendor specific so I specify H2 as the required dialect. However, it seems the bug and fixing is not vendor specific so any one should be enough.

NathanQingyangXu: @Sanne I solved another critical bug. Enjoy!

sebersole:
Applied upstream.  Thanks Nathan!

PR #3135, opened 2019-12-09
Fix SQLException: Connection is closed

jsuper:
Fix SQLException: Connection is closed

In some case, the physicalConnection was not null , but it was closed by connection pool. All follow operations will throw SQLException if we not check the connection state.


sebersole:
The build is failing with these changes...

I'm not sure this approach will work without JTA.  When using JDBC transactions, the transaction is associated with the Connection.  You now have a new Connection.  You'll need to add a test for that.

Also, I'm not sure about the performance impact of this change (let alone the correctness concern above).  Everytime Hibernate grabs the Connection you have now added overhead with the `#isClosed` call. That has a performance impact.  **If the correctness aspect is verified**, I still think this should be protected by a (disabled by default) setting

PR #3137, opened 2019-12-09
HHH-13724 Add testing configs and dialect for CockroachDB

rafiss:
This adds a CockroachDB dialect and build configuration for testing with CockroachDB.

Tests that use features that CockroachDB does not support yet are annotated with SkipForDialect.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
@@ -145,4 +145,8 @@ public class CockroachDB1920Dialect extends PostgreSQL95Dialect {
 
     @Override
     public boolean supportsComputedIndexes() { return false; }
+
+    @Override
+    public boolean supportsNoWait() { return false; }
+
 }
--- a/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
@@ -22,97 +22,97 @@ import org.hibernate.type.descriptor.sql.VarcharTypeDescriptor;
  */
 public class CockroachDB1920Dialect extends PostgreSQL95Dialect {
 
-    public CockroachDB1920Dialect() {
-        super();
-        registerColumnType( Types.BLOB, "bytea" );
-    }
+	public CockroachDB1920Dialect() {
+		super();
+		registerColumnType(Types.BLOB, "bytea");
+	}
 
-    @Override
-    public IdentityColumnSupport getIdentityColumnSupport() {
-        return new CockroachDB1920IdentityColumnSupport();
-    }
+	@Override
+	public IdentityColumnSupport getIdentityColumnSupport() {
+		return new CockroachDB1920IdentityColumnSupport();
+	}
 
-    @Override
-    public MultiTableBulkIdStrategy getDefaultMultiTableBulkIdStrategy() {
-        // CockroachDB 19.2.0 does not support temporary tables, so we must override the Postgres behavior.
-        return new InlineIdsInClauseBulkIdStrategy();
-    }
+	@Override
+	public MultiTableBulkIdStrategy getDefaultMultiTableBulkIdStrategy() {
+		// CockroachDB 19.2.0 does not support temporary tables, so we must override the Postgres behavior.
+		return new InlineIdsInClauseBulkIdStrategy();
+	}
 
-    public boolean doesReadCommittedCauseWritersToBlockReaders() {
-        return true;
-    }
+	public boolean doesReadCommittedCauseWritersToBlockReaders() {
+		return true;
+	}
 
-    @Override
-    public boolean supportsExpectedLobUsagePattern() {
-        return false;
-    }
+	@Override
+	public boolean supportsExpectedLobUsagePattern() {
+		return false;
+	}
 
-    @Override
-    public SqlTypeDescriptor getSqlTypeDescriptorOverride(int sqlCode) {
-        SqlTypeDescriptor descriptor;
-        switch (sqlCode) {
-            case Types.BLOB: {
-                // Make BLOBs use byte[] storage.
-                descriptor = VarbinaryTypeDescriptor.INSTANCE;
-                break;
-            }
-            case Types.CLOB: {
-                // Make CLOBs use string storage.
-                descriptor = VarcharTypeDescriptor.INSTANCE;
-                break;
-            }
-            default: {
-                descriptor = super.getSqlTypeDescriptorOverride(sqlCode);
-                break;
-            }
-        }
-        return descriptor;
-    }
+	@Override
+	public SqlTypeDescriptor getSqlTypeDescriptorOverride(int sqlCode) {
+		SqlTypeDescriptor descriptor;
+		switch (sqlCode) {
+			case Types.BLOB: {
+				// Make BLOBs use byte[] storage.
+				descriptor = VarbinaryTypeDescriptor.INSTANCE;
+				break;
+			}
+			case Types.CLOB: {
+				// Make CLOBs use string storage.
+				descriptor = VarcharTypeDescriptor.INSTANCE;
+				break;
+			}
+			default: {
+				descriptor = super.getSqlTypeDescriptorOverride(sqlCode);
+				break;
+			}
+		}
+		return descriptor;
+	}
 
-    @Override
-    public boolean supportsJdbcConnectionLobCreation(DatabaseMetaData databaseMetaData) {
-        return false;
-    }
+	@Override
+	public boolean supportsJdbcConnectionLobCreation(DatabaseMetaData databaseMetaData) {
+		return false;
+	}
 
-    @Override
-    public boolean supportsLockTimeouts() {
-        return false;
-    }
+	@Override
+	public boolean supportsLockTimeouts() {
+		return false;
+	}
 
-    @Override
-    public boolean supportsSkipLocked() {
-        // CockroachDB 19.2.0 doesn't support this: https://github.com/cockroachdb/cockroach/issues/40476
-        return false;
-    }
+	@Override
+	public boolean supportsSkipLocked() {
+		// CockroachDB 19.2.0 doesn't support this: https://github.com/cockroachdb/cockroach/issues/40476
+		return false;
+	}
 
-    @Override
-    public boolean supportsNoWait() {
-        // CockroachDB 19.2.0 doesn't support this: https://github.com/cockroachdb/cockroach/issues/40476
-        return false;
-    }
+	@Override
+	public boolean supportsNoWait() {
+		// CockroachDB 19.2.0 doesn't support this: https://github.com/cockroachdb/cockroach/issues/40476
+		return false;
+	}
 
-    @Override
-    public boolean supportsMixedTypeArithmetic() {
-        return false;
-    }
+	@Override
+	public boolean supportsMixedTypeArithmetic() {
+		return false;
+	}
 
-    @Override
-    public boolean canCreateSchema() {
-        return false;
-    }
+	@Override
+	public boolean canCreateSchema() {
+		return false;
+	}
 
-    @Override
-    public boolean supportsStoredProcedures() {
-        return false;
-    }
+	@Override
+	public boolean supportsStoredProcedures() {
+		return false;
+	}
 
-    @Override
-    public boolean supportsComputedIndexes() {
-        return false;
-    }
+	@Override
+	public boolean supportsComputedIndexes() {
+		return false;
+	}
 
-    @Override
-    public boolean supportsLoFunctions() {
-        return false;
-    }
+	@Override
+	public boolean supportsLoFunctions() {
+		return false;
+	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
@@ -24,9 +24,9 @@ public class CockroachDB1920Dialect extends PostgreSQL95Dialect {
 
     public CockroachDB1920Dialect() {
         super();
-        registerColumnType(Types.INTEGER, "int8");
-        registerColumnType(Types.FLOAT, "float8");
-        registerColumnType(Types.BLOB, "bytea");
+        registerColumnType( Types.INTEGER, "int8" );
+        registerColumnType( Types.FLOAT, "float8" );
+        registerColumnType( Types.BLOB, "bytea" );
     }
 
     @Override
@@ -56,11 +56,12 @@ public class CockroachDB1920Dialect extends PostgreSQL95Dialect {
         switch (sqlCode) {
             case Types.BLOB: {
                 // Make BLOBs use byte[] storage.
-                descriptor = VarbinaryTypeDescriptor.INSTANCE; // CockroachDBBlobTypeDescriptor.INSTANCE;
+                descriptor = VarbinaryTypeDescriptor.INSTANCE;
                 break;
             }
             case Types.CLOB: {
-                descriptor = VarcharTypeDescriptor.INSTANCE; // CockroachDBClobTypeDescriptor.INSTANCE;
+                // Make CLOBs use string storage.
+                descriptor = VarcharTypeDescriptor.INSTANCE;
                 break;
             }
             default: {
@@ -90,7 +91,9 @@ public class CockroachDB1920Dialect extends PostgreSQL95Dialect {
     }
 
     @Override
-    public boolean supportsMixedTypeArithmetic() { return false; }
+    public boolean supportsMixedTypeArithmetic() {
+        return false;
+    }
 
     @Override
     public boolean canCreateSchema() { return false; }
--- a/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
@@ -1,14 +1,15 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
 package org.hibernate.dialect;
 
 import org.hibernate.dialect.identity.CockroachDB1920IdentityColumnSupport;
 import org.hibernate.dialect.identity.IdentityColumnSupport;
-import org.hibernate.dialect.identity.PostgreSQL10IdentityColumnSupport;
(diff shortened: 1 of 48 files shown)


rafiss:
There are multiple commits right now just for ease of reviewing. I can squash these together before this gets merged, or whenever you like.

hibernate-core/src/test/java/org/hibernate/jpa/test/lock/NativeSQLQueryTimeoutTest.java
```diff
@@ -46,9 +49,10 @@ public void test(){
 
 				fail("Should have thrown lock timeout exception!");
 			} catch (Exception expected) {
+				expected.printStackTrace();
 				assertTrue(
 					ExceptionUtil.rootCause(expected)
-						.getMessage().contains("canceling statement due to user request")
+						.getMessage().contains("csanceling statement due to user request")
```
NathanQingyangXu: So the message is expected to contain the misspelled `csanceling`, right?

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/PostgreSQLMultipleSchemaSequenceTest.java
```diff
@@ -48,6 +50,7 @@
  * @author Vlad Mihalcea
  */
 @RequiresDialect(PostgreSQL81Dialect.class)
+@RequiresDialectFeature(value = DialectChecks.SupportSchemaCreation.class)
```
NathanQingyangXu: Not a big deal, but sometimes you omit the `value` spec elsewhere. A little bit inconsistent, but I think we don't have coding style guide on this.

hibernate-core/src/test/java/org/hibernate/test/where/annotations/LazyManyToManyNonUniqueIdWhereTest.java
```diff
@@ -53,10 +53,15 @@ public void setup() {
...
 				this::sessionFactory, session -> {
 
-					session.createSQLQuery( "drop table MATERIAL_RATINGS" ).executeUpdate();
-					session.createSQLQuery( "drop table BUILDING_RATINGS" ).executeUpdate();
-					session.createSQLQuery( "drop table ASSOCIATION_TABLE" ).executeUpdate();
-					session.createSQLQuery( "drop table MAIN_TABLE" ).executeUpdate();
+					session.createSQLQuery("drop table if exists MATERIAL_RATINGS cascade").executeUpdate();
+					session.createSQLQuery("drop table if exists BUILDING_RATINGS cascade").executeUpdate();
+					session.createSQLQuery("drop table if exists ASSOCIATION_TABLE cascade").executeUpdate();
+					session.createSQLQuery("drop table if exists MAIN_TABLE cascade").executeUpdate();
```
NathanQingyangXu: it would be consistent to retain the empty chars after `(` and before `)`.

hibernate-core/src/test/java/org/hibernate/jpa/test/lock/NativeSQLQueryTimeoutTest.java
rafiss: This was a typo. Thanks for catching it!

hibernate-core/src/test/java/org/hibernate/test/where/annotations/LazyManyToManyNonUniqueIdWhereTest.java
rafiss: I agree -- I'll make this consistent.

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/PostgreSQLMultipleSchemaSequenceTest.java
rafiss: I removed `value` when there is only one argument.

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
```diff
@@ -0,0 +1,118 @@
...
+
+import org.hibernate.dialect.identity.CockroachDB1920IdentityColumnSupport;
+import org.hibernate.dialect.identity.IdentityColumnSupport;
+import org.hibernate.hql.spi.id.MultiTableBulkIdStrategy;
+import org.hibernate.hql.spi.id.inline.InlineIdsInClauseBulkIdStrategy;
+import org.hibernate.type.descriptor.sql.*;
+
+import java.sql.*;
+
+import java.sql.DatabaseMetaData;
```
NathanQingyangXu: Seems the above import statement has been included in the previous one. Also, should we order imports by some way? I usually put static imports after non-static ones and order statements in each group by string ASC order. I might consider whether to use wildcard, though I am not sure about the convention in Hibernate on this.

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * An SQL dialect for CockroachDB 19.2 and later. This is the first dialect for CockroachDB. It extends
```
NathanQingyangXu: should be 'A SQL dialect', right?

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public IdentityColumnSupport getIdentityColumnSupport() {
+        return new CockroachDB1920IdentityColumnSupport();
+    }
+
+    @Override
+    public MultiTableBulkIdStrategy getDefaultMultiTableBulkIdStrategy() {
+        // CockroachDB 19.2.0 does not support temporary tables, so we must override the Postgres behavior.
```
NathanQingyangXu: if we aim not only for one single version, we'd better change the verbiage to something like 'CockroachDB 19.2.0 or later ...'.

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
NathanQingyangXu: Should we change ' for CockroachDB 19.2 and later' to ' for CockroachDB 19.2'? From the classname it implies so.

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public boolean supportsLockTimeouts() {
+        return false;
+    }
+
+    @Override
+    public boolean supportsSkipLocked() {
+        // CockroachDB doesn't support this: https://github.com/cockroachdb/cockroach/issues/40476
```
NathanQingyangXu: I would add version to the above comment to be exact. It might be solved by later version and this class is only for specific version, right?

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -2976,6 +2976,22 @@ public boolean supportsNoWait() {
...
+	 * Does this dialect/database support LO_CREATE and related functions.
+	 *
+	 * @return {@code true} if LO_CREATE and related functions are supported
+	 */
+	public boolean supportsLoFunctions() {
+		return false;
+	}
+
+	/**
+	 * Does this dialect/database support mixed type arithmetic (e.g. <int> + <float>).
```
NathanQingyangXu: Given '<' and '>' are special HTML chars, I saw only "(e.g. +)" shows up in javadoc. You might need html escaped text for '<' and '>'.

hibernate-core/src/main/java/org/hibernate/dialect/identity/CockroachDB1920IdentityColumnSupport.java
```diff
@@ -0,0 +1,34 @@
...
+public class CockroachDB1920IdentityColumnSupport extends IdentityColumnSupportImpl {
+    @Override
+    public boolean supportsIdentityColumns() {
+        // Full support requires setting the sql.defaults.serial_normalization=sql_sequence in CockroachDB.
+        return false;
+    }
+
+    @Override
+    public String getIdentitySelectString(String table, String column, int type) {
+        return "select currval('" + table + '_' + column + "_seq')";
```
NathanQingyangXu: Do we need some escaping mechanism to guard against the possibility that 'table', 'column' names might contain special characters (like "(", ")", "'", etc.)?

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ClobTypeDescriptor.java
```diff
@@ -129,6 +139,9 @@ else if ( Reader.class.isAssignableFrom( value.getClass() ) ) {
 			Reader reader = (Reader) value;
 			return options.getLobCreator().createClob( DataHelper.extractString( reader ) );
 		}
+		else if ( String.class.isAssignableFrom( value.getClass() ) ) {
```
NathanQingyangXu: Should we use `String.class.isAssignableFrom ( value )` instead?

hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/basic/PredicateTest.java
```diff
@@ -246,6 +246,7 @@ public void testByteArray() {
 
 	@Test
 	@TestForIssue( jiraKey = "HHH-5803" )
+	@RequiresDialectFeature(DialectChecks.SupportsMixedTypeArithmetic.class)
```
NathanQingyangXu: You'd better leave empty chars after '(' and before ')' as elsewhere.

hibernate-core/src/test/java/org/hibernate/test/optlock/OptimisticLockTest.java
```diff
@@ -161,20 +162,23 @@ private void testDeleteOptimisticLockFailure(String entityName) {
...
+				// CockroachDB always runs in SERIALIZABLE isolation, and uses SQL state 40001 to indicate
+				// serialization failure.
+			}
+			else {
+				throw e;
+			}
 		}
 		else if ( !(cause instanceof StaleObjectStateException) && !(cause instanceof StaleStateException) ) {
-			fail( "expectd StaleObjectStateException or StaleStateException exception but is" + cause );
+			fail( "expected StaleObjectStateException or StaleStateException exception but is" + cause );
```
NathanQingyangXu: We'd better leave an empty char after " but is", right?

hibernate-core/src/test/java/org/hibernate/test/typeoverride/TypeOverrideTest.java
```diff
@@ -47,7 +39,13 @@ public void testStandardBasicSqlTypeDescriptor() {
 		assertSame( IntegerTypeDescriptor.INSTANCE, remapSqlTypeDescriptor( IntegerTypeDescriptor.INSTANCE ) );
 
 		// A few dialects explicitly override BlobTypeDescriptor.DEFAULT
-		if ( PostgreSQL81Dialect.class.isInstance( getDialect() ) || PostgreSQLDialect.class.isInstance( getDialect() ) )  {
+		if ( CockroachDB1920Dialect.class.isInstance( getDialect() ))  {
+			assertSame(
```
NathanQingyangXu: It seems previously somewhere you use 'assertEquals' and here you are relying on references being the same. Are you sure 'assertSame' is the correct one, btw?

hibernate-core/src/test/java/org/hibernate/test/where/hbm/LazyManyToManyNonUniqueIdNotFoundWhereTest.java
```diff
@@ -37,10 +38,15 @@ public void setup() {
...
 				this::sessionFactory, session -> {
 
-					session.createSQLQuery( "drop table MATERIAL_RATINGS" ).executeUpdate();
-					session.createSQLQuery( "drop table BUILDING_RATINGS" ).executeUpdate();
-					session.createSQLQuery( "drop table ASSOCIATION_TABLE" ).executeUpdate();
-					session.createSQLQuery( "drop table MAIN_TABLE" ).executeUpdate();
+					session.createSQLQuery("drop table if exists MATERIAL_RATINGS cascade").executeUpdate();
+					session.createSQLQuery("drop table if exists BUILDING_RATINGS cascade").executeUpdate();
+					session.createSQLQuery("drop table if exists ASSOCIATION_TABLE cascade").executeUpdate();
+					session.createSQLQuery("drop table if exists MAIN_TABLE cascade").executeUpdate();
```
NathanQingyangXu: Could you add empty char delimiters above and elsewhere?

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
rafiss: I just used "An SQL" because that's what other dialect javadocs have (e.g. MySQL55Dialect or PostgreSQL82Dialect). I think the reasoning is that "SQL" can be pronounced by saying each letter, rather than sounding like "sequel."

rafiss: The dialect would still work with later versions of CockroachDB. Also, I used the same language that other dialects have (e.g. PostgreSQL82Dialect).

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
rafiss: I will eliminate the duplicate.
It seems like wildcard imports are used elsewhere in Hibernate.
As far as ordering goes, the convention I was used to is to to have intra-project imports first, then imports from other projects, then static imports (with each section being alphabetical).

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDB1920Dialect.java
rafiss: Good point! added.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
rafiss: I removed the '<' and '>' characters since I don't think they are needed.

hibernate-core/src/main/java/org/hibernate/dialect/identity/CockroachDB1920IdentityColumnSupport.java
rafiss: This implementation is the same as PostgreSQL81IdentityColumnSupport

However, since the CockroachDB dialect does not support identity columns, it would be more clear if I just removed these method overrides.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ClobTypeDescriptor.java
rafiss: No, the `isAssignableFrom` method requires an argument of type `Class<?>`. See lines 134 and 138 also.

hibernate-core/src/test/java/org/hibernate/test/optlock/OptimisticLockTest.java
rafiss: Yes, nice catch

hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/basic/PredicateTest.java
rafiss: I will change this, but this does not seem very consistent throughout the Hibernate project (especially for annotations).

hibernate-core/src/test/java/org/hibernate/test/where/hbm/LazyManyToManyNonUniqueIdNotFoundWhereTest.java
rafiss: Fixed

NathanQingyangXu:
> This adds a CockroachDB dialect and build configuration for testing with CockroachDB.
> 
> Tests that use features that CockroachDB does not support yet are annotated with SkipForDialect.

It would be ideal to update 'documentation' relevant part, i.e. db dialect in user guide.

gavinking:
Folks, note that there is already a [dialect for Cockroach in Hibernate 6](https://github.com/hibernate/hibernate-orm/blob/wip/6.0/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java), that I wrote back in June, but that has only just been reintegrated with the new Hibernate 6 stream.

Having eyeballed this PR, I think it might be better to take my version as a starting point and see what it's missing,

rafiss:
@gavinking Thanks for taking a look! We had discussed our work with some of the other Hibernate maintainers, and they had recommended this approach. There were two main reasons: (1) that earlier PR seems specific to Hibernate 6, and we wanted to include this dialect in a general release as soon as possible, and (2) we want our dialect to match the Postgres one as closely as possible, so we thought it would be a good approach to inherit from the existing Postgres dialect. I believe @Sanne was planning to review this PR soon.

I have been looking through your PR -- do you have anything in mind from it that you'd like me to incorporate here? I really appreciate any guidance you have, and I'm grateful for your feedback. I'm excited to enhance the support for CockroachDB!

gavinking:
> we want our dialect to match the Postgres one as closely as possible, so we thought it would be a good approach to inherit from the existing Postgres dialect.

Well the problem is that Cockroach *isn't* Postgres, even though I can see that a lot of work has been done to achieve a reasonable level of compatibility with it, so there is stuff in the Postgres dialect which (at least from what I can see in the docs) just isn't correct for Cockroach, for example, SQL function registrations that Cockroach doesn't have, and mappings for types that just don't exist in Cockroach.

And so it makes it *really* hard from looking at the `CockroachDialect` to understand exactly what things are really there in and supposed to work on Cockroach itself, and which things are just baggage from the dialect it inherits.

There are really good reasons for us trying to get away from the use of inheritance between different product's `Dialects`. Yes, it's OK for products so closely related as MySQL and MariaDB, but that just isn't the case here.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -3053,4 +3069,12 @@ public boolean supportsSelectAliasInGroupByClause() {
...
 	}
 
+	public boolean supportsStoredProcedures() {
+		return true;
+	}
+
+	public boolean supportsComputedIndexes() {
+		return true;
+	}
+
```
dreab8: in my opinion it is better just to use the `@SkipForDialect(value = CockroachDB1920Dialect.class,...)` annotation and avoid all these new methods.

rafiss:
Thanks for the feedback on this, everyone. I put up a new PR #3280 that does not inherit from the Postgres dialect.

PR #3138, opened 2019-12-10
HHH-13759 - Support for private Embedded in MappedSuperclass

barreiro:


hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/MappedSuperclassWithEmbeddableTest.java
```diff
@@ -0,0 +1,148 @@
...
+		@Override
+		public boolean equals(Object o) {
+			if ( this == o ) {
+				return true;
+			}
+			if ( o == null || getClass() != o.getClass() ) {
+				return false;
+			}
+			EmbeddedValue that = (EmbeddedValue) o;
+			return superField.equals( that.superField );
```
NathanQingyangXu: I think `Objects.equals(superField, that.superField)` would be more robust and align well with the `hashCode` implementation, though it would be no concern if we are sure `superField` is never null.

DavideD: This is a test case :)
We know it's not going to be null

DavideD:
@barreiro I'm going to close this PR and keep only this one: https://github.com/hibernate/hibernate-orm/pull/3141
Your commit is included there, maybe you can have a look at the refactoring I made to check that it doesn't brake anything

PR #3139, opened 2019-12-11
HHH-13077 Optimize query plan call count

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13077

I know Vlad created a new PR after the first PR was reverted: https://github.com/hibernate/hibernate-orm/pull/2634.

Compared to his PR, my PR is less intrusive and pretty easy to confirm.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
@@ -109,7 +109,27 @@ import org.hibernate.type.descriptor.sql.SqlTypeDescriptor;
 @SuppressWarnings("WeakerAccess")
 public abstract class AbstractSharedSessionContract implements SharedSessionContractImplementor {
 	private static final EntityManagerMessageLogger log = HEMLogging.messageLogger( SessionImpl.class );
-
+	
+	/**
+	 * @see this#doCreateQuery(String)
+	 */
+	private static class PlanAwareQueryImpl<T> extends QueryImpl<T> {
+		private final HQLQueryPlan plan;
+		
+		public PlanAwareQueryImpl(
+				SharedSessionContractImplementor producer,
+				ParameterMetadata parameterMetadata,
+				String queryString,
+				HQLQueryPlan plan) {
+			super( producer, parameterMetadata, queryString );
+			this.plan = plan;
+		}
+		
+		public HQLQueryPlan getPlan() {
+			return plan;
+		}
+	}
+	
 	private transient SessionFactoryImpl factory;
 	private final String tenantIdentifier;
 	protected transient FastSessionServices fastSessionServices;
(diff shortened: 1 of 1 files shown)


NathanQingyangXu:
@vladmihalcea I implemented the feature as per your suggestion (modifying QueryImpl by adding a new field to store QueryPlan). Could you take a look at it?
I think we can achieve the goal to pass the plan to the other method without touching QueryImpl, but I wanna get your opinion on this PR first. Thanks.

hibernate-core/src/main/java/org/hibernate/query/internal/QueryImpl.java
```diff
@@ -18,16 +19,19 @@
 public class QueryImpl<R> extends AbstractProducedQuery<R> implements Query<R> {
 	private final String queryString;
 
+	private final HQLQueryPlan queryPlan;
```
NathanQingyangXu: My other concern for the above field is it might delay GC for I read somewhere it is memory intensive. Maybe a `WeakReference` is more appropriate?

jwgmeligmeyling:
I think the separation between query and its HQL plan is a matter of separation of concern. Its not sensible in a sense for a query to hold onto its own plan. The cache should be a fairly optimal look up and prevent unnecessary recomputation of the HQL plan. Do we have any actual indications that this is a performance bottleneck?

NathanQingyangXu:
> I think the separation between query and its HQL plan is a matter of separation of concern. Its not sensible in a sense for a query to hold onto its own plan. The cache should be a fairly optimal look up and prevent unnecessary recomputation of the HQL plan. Do we have any actual indications that this is a performance bottleneck?

I have no idea, TBH. If you are okay with the current implementation, it would be great. Just wanna raise some awareness, which might not be valid.

NathanQingyangXu:
> I think the separation between query and its HQL plan is a matter of separation of concern. Its not sensible in a sense for a query to hold onto its own plan. The cache should be a fairly optimal look up and prevent unnecessary recomputation of the HQL plan. Do we have any actual indications that this is a performance bottleneck?

I chose this ticket simply because it was labeled to be released in next version (i.e. 5.4.11) and I think if I took over this ticket, my PR has good chance to be reviewed and merged.
As you can see, I implemented as suggested by the ticket creator. However, I am not a fan of the implementation plan to store query plan within query. IMHO, we have much easier way to pass the query plan between methods in the same class.

NathanQingyangXu:
> I think the separation between query and its HQL plan is a matter of separation of concern. Its not sensible in a sense for a query to hold onto its own plan. The cache should be a fairly optimal look up and prevent unnecessary recomputation of the HQL plan. Do we have any actual indications that this is a performance bottleneck?

I agree with you and refactored the implementation without complicating Query. It is simple refactoring to pass additional info between two methods in the same class.

hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
```diff
@@ -818,11 +835,17 @@ protected void applyQuerySettingsAndHints(Query query) {
...
+		if ( hqlQuery instanceof PlanAwareQueryImpl ) {
+			queryPlan = ( (PlanAwareQueryImpl) hqlQuery).plan;
+		}
+		if ( queryPlan == null ) {
+			queryPlan = getFactory().getQueryPlanCache().getHQLQueryPlan(
+					hqlQuery.getQueryString(),
+					false,
+					getLoadQueryInfluencers().getEnabledFilters()
+			);
+		}
```
NathanQingyangXu: I retained the above code for the method is `protected` and it might have been used by our user. Let me know if it is overkill. Thanks.

NathanQingyangXu:
I closed this PR in favour of Vlad's: https://github.com/hibernate/hibernate-orm/pull/2634

PR #3140, opened 2019-12-11
HHH-13777 UnsupportedOperationException is thrown for MappedSuperclas…

dreab8:
…s if the id is declared on subclasses

https://hibernate.atlassian.net/browse/HHH-13777

@sebersole any advice for a better name than `isNotDeclaredIdentifierDescriptorLegal()` is really appreciated.


sebersole:
`#isMissingIdentifierLegal` ?

Anyway, I think the solution looks good

sebersole:
or how about flipping the logic and have `#isIdMappingRequired`

dreab8:
I like it, I'll make the change and push it. Thanks

dreab8:
applied Steve suggestion and pushed

PR #3141, opened 2019-12-11
HHH-13759 Error on enhanced field with embedded and @MappedSuperclass

DavideD:
For issue [HHH-13759](https://hibernate.atlassian.net/browse/HHH-13759)
These are some additional commits to add on top of https://github.com/hibernate/hibernate-orm/pull/3138

This PR adds:
* test cases to check that the classes have the additional fields created during enhancements
* a simple refactoring of the existing code to make it clearer that we skip some logic when dealing with `@MappedSuperclass`.
* a comment in InlineDirtyCheckingHandler

I guess they are all optional but I would like to have them included


DavideD:
A link to the decompiled classes with ByteBuddy:
https://gist.github.com/DavideD/476e75b2b862173c148e1fd87220cc52

Sanne:
works! thanks all

PR #3144, opened 2019-12-13
HHH-13677 Make org.hibernate.flushMode config take effect

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13677

We ignored setting flush mode during session building. This PR simply grabs the setting from `fastSessionService`. Testing case added.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
@@ -256,6 +256,8 @@ public final class SessionImpl
 		// NOTE : pulse() already handles auto-join-ability correctly
 		getTransactionCoordinator().pulse();
 
+		getSession().setHibernateFlushMode( ConfigurationHelper.getFlushMode( getSessionProperty( AvailableSettings.FLUSH_MODE ), FlushMode.AUTO ) );
+
 		if ( log.isTraceEnabled() ) {
 			log.tracef( "Opened Session [%s] at timestamp: %s", getSessionIdentifier(), getTimestamp() );
 		}
--- a/hibernate-core/src/test/java/org/hibernate/internal/FlushModeConfigTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/internal/FlushModeConfigTest.java
@@ -1,0 +1,60 @@
+package org.hibernate.internal;
+
+import org.hibernate.FlushMode;
+import org.hibernate.Session;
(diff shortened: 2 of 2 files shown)


NathanQingyangXu:
Thank you, guys!

hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
```diff
@@ -256,6 +256,8 @@ public SessionImpl(SessionFactoryImpl factory, SessionCreationOptions options) {
 		// NOTE : pulse() already handles auto-join-ability correctly
 		getTransactionCoordinator().pulse();
 
+		getSession().setHibernateFlushMode( ConfigurationHelper.getFlushMode( getSessionProperty( AvailableSettings.FLUSH_MODE ), FlushMode.AUTO ) );
+
```
dreab8: @Sanne should we add the `FlushMode` to the `FastSessionServices`?

NathanQingyangXu: Actually, `FastSessionServices` contains `FlushMode` default value. If you drill down the `getSessionProperty()` you can see it will look up in `FastSessionServices` already:
Screen Shot 2020-01-27 at 5 34 12 PM

Sanne: Right, looks good. Is something not working as expected @dreab8 ? Feel free to add tests :)

dreab8: I haven't noticed that `getSessionProperty` takes the value from the `FastSessionServices`

NathanQingyangXu:
That is understandable. Btw, why not make FastSessionServices to store
default settings using array rather than map? We could refactor
AvailableSettings into enum and move default settings to be stored in
FastSessionServices to the top of the rnum. We might refactor this util
method. Then we can make the FastSessionServices even faster. Just a
brainstorming excercise but seems doable.

dreab8:
Applied upstream.
Thanks @NathanQingyangXu

hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
Sanne: hum both `ConfigurationHelper#getFlushMode` and the `getSessionProperty` methods are rather slow though. I see what @dreab8 meant now, I will add an additional fix to this.

PR #3147, opened 2019-12-14
HHH-13373, HHH-13472 fix the bug in SequenceInformationExtractorMariaDBDatabaseImpl

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13373
https://hibernate.atlassian.net/browse/HHH-13472

I think this is a serious bug which manifests itself in face of multiple dbs. Suppose some db has some sequence whose name is not shared in another db, SequenceInformationExtractorMariaDBDatabaseImpl will be broken for internally it relies on 'UNIION ALL' to fetch all sequence info as per the names fetched from all dbs, regardless of whether the sequence exists in the current db or not.

However, testing is difficult, unless the 'hibernate_orm_test' user has privilege to create other db and sequence there.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/MariaDB103Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/MariaDB103Dialect.java
@@ -14,7 +14,7 @@ import org.hibernate.type.StandardBasicTypes;
 
 /**
  * An SQL dialect for MariaDB 10.3 and later, provides sequence support, lock-timeouts, etc.
- * 
+ *
  * @author Philippe Marschall
  */
 public class MariaDB103Dialect extends MariaDB102Dialect {
@@ -57,7 +57,7 @@ public class MariaDB103Dialect extends MariaDB102Dialect {
 
 	@Override
 	public String getQuerySequencesString() {
-		return "select table_name from information_schema.TABLES where table_type='SEQUENCE'";
+		return "select table_name from information_schema.TABLES where table_schema = database() and table_type = 'SEQUENCE'";
 	}
 
 	@Override
--- a/hibernate-core/src/test/java/org/hibernate/test/dialect/functional/MariaDBExtractSequenceMatadataTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/dialect/functional/MariaDBExtractSequenceMatadataTest.java
@@ -1,16 +1,14 @@
 package org.hibernate.test.dialect.functional;
 
+import java.sql.*;
+
+import org.hibernate.cfg.Environment;
 import org.hibernate.dialect.MariaDB103Dialect;
-import org.hibernate.testing.FailureExpected;
-import org.hibernate.testing.RequiresDialect;
-import org.hibernate.testing.TestForIssue;
+import org.hibernate.engine.jdbc.env.spi.JdbcEnvironment;
+import org.hibernate.testing.*;
 import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.junit.Assert;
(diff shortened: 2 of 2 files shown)


dreab8:
Hi @NathanQingyangXu , thanks for the PR can you please add also a test case?

NathanQingyangXu:
The test involves multiple dbs. I found multitenancy testing only supports h2. Could you shed some light on how to test it? Thanks.

dreab8:
I may be wrong but I think the issue with testing multitenancy on other dbs than h2 is that we have to be sure that the`hibernate_orm_test` user has the privileges to create a new database and at the moment I'm not sure this is the case for our CI instance, I'll try to verify it.

NathanQingyangXu:
thanks. yeah, in this case I need the privilege to create another db, then create a sequence there with different name than ones in current db. Then we are set.

NathanQingyangXu:
> Hi @NathanQingyangXu , thanks for the PR can you please add also a test case?

I added one annotated with `@FailureExpected`. Thanks.

dreab8:
Thanks @NathanQingyangXu ,
pushed it upstream.
I checked and it seems the users has privileges to create a new db so I removed the `@FailureExpected` annotation, I also changed the annotation of the `tearDownDBs` method to `AfterClass` in order to be sure it is executed even if an exception is raised  in the `setUpDBs` method.

NathanQingyangXu:
Thanks. I used to worry this PR most because of testing concern. Now it
turns out to be merged quickly. I appreciate your help so much.

hibernate-core/src/main/java/org/hibernate/dialect/MariaDB103Dialect.java
```diff
@@ -57,7 +57,7 @@ public String getSelectSequenceNextValString(String sequenceName) {
 
 	@Override
 	public String getQuerySequencesString() {
-		return "select table_name from information_schema.TABLES where table_type='SEQUENCE'";
+		return "select table_name from information_schema.TABLES where table_schema = database() and table_type = 'SEQUENCE'";
```
gavinking: @NathanQingyangXu so is this syntax really specific to MariaDB 10.3, or is it also correct on earlier versions?

NathanQingyangXu: I tried to get the answer from https://mariadb.com/kb/en/database/ to no avail. But given the page was created 9 years ago, it seems to have a long history.

NathanQingyangXu: I checked the MySQL info for v5.5 and found this:https://dev.mysql.com/doc/refman/5.5/en/information-functions.html#function_database.
So the function exists from the beginning of Mariadb.

NathanQingyangXu: Would you mind my asking why you asked the question?x

NathanQingyangXu: If your concern is more about sequence, then yeah, it is only in this version that mariadb supports sequence, because of some Chinese engineers open source contribution, to my knowledge.

gavinking: OK thanks that’s what I needed to know (I need to replicate this change on the Hibernate 6 branch).

gavinking: OK thanks that’s what I needed to know (I need to replicate this change on the Hibernate 6 branch).

PR #3148, opened 2019-12-14
HHH-13310, HHH-13321 Make getParameterValue(), isBound() work for HQL collection parameter

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13310
https://hibernate.atlassian.net/browse/HHH-13321

This PR aims to fix the above two defects

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/spi/QueryParameterBindings.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/spi/QueryParameterBindings.java
@@ -21,6 +21,10 @@ import org.hibernate.type.Type;
 public interface QueryParameterBindings {
 	boolean isBound(QueryParameter parameter);
 
+	<T> T getQueryParameterValue(QueryParameter<T> parameter);
+	Object getQueryParameterValue(String name);
+	Object getQueryParameterValue(int position);
+
 	<T> QueryParameterBinding<T> getBinding(QueryParameter<T> parameter);
 	<T> QueryParameterBinding<T> getBinding(String name);
 	<T> QueryParameterBinding<T> getBinding(int position);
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/query/spi/QueryParameterBindings.java
```diff
@@ -21,6 +21,10 @@
 public interface QueryParameterBindings {
 	boolean isBound(QueryParameter parameter);
 
+	<T> T getQueryParameterValue(QueryParameter<T> parameter);
```
jwgmeligmeyling: This is an SPI which IMO shouln't be extended in a minor release.

Also, this should be embedded inside `QueryParameterBinding` and `QueryParameterListBinding`. Why doesn't [`getBinding(parameter).getBindValue()`](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/query/spi/QueryParameterBinding.java#L51) work for you? It should even return a type `T`.

NathanQingyangXu: The reason I have to introduce new method is tricky and subtle. Internally getBinding() is not a pure query method and it involves creation side effect (invoking 'makeBinding' if no binding is found).

NathanQingyangXu: I added default implementation of these 'pure' query methods to maintain SPI backward compatibility. Sadly the original `getBinding()` method are used for both query and insertion and that is why the bugs show up because of it. Providing 'side effect free' query methods solve the issue. Not sure the efforts here are still necessary in Hibernate v6, though.

PR #3151, opened 2019-12-17
HHH-13788 Schema update try to recreate existing tables

dreab8:
https://hibernate.atlassian.net/browse/HHH-13788

It should fix https://github.com/quarkusio/quarkus/issues/5883

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/MariaDBDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/MariaDBDialect.java
@@ -6,6 +6,14 @@
  */
 package org.hibernate.dialect;
 
+import java.sql.DatabaseMetaData;
+import java.sql.SQLException;
+
+import org.hibernate.engine.jdbc.env.spi.AnsiSqlKeywords;
+import org.hibernate.engine.jdbc.env.spi.IdentifierCaseStrategy;
+import org.hibernate.engine.jdbc.env.spi.IdentifierHelper;
+import org.hibernate.engine.jdbc.env.spi.IdentifierHelperBuilder;
+
 /**
  * @author Vlad Mihalcea
  */
@@ -21,5 +29,16 @@ public class MariaDBDialect extends MySQL5Dialect {
 	@Override
 	protected MySQLStorageEngine getDefaultMySQLStorageEngine() {
 		return InnoDBStorageEngine.INSTANCE;
+	}
+
+	@Override
+	public IdentifierHelper buildIdentifierHelper(IdentifierHelperBuilder builder, DatabaseMetaData dbMetaData)
+			throws SQLException {
+
+		// some MariaDB drivers does not return case strategy info
+		builder.setUnquotedCaseStrategy( IdentifierCaseStrategy.MIXED );
+		builder.setQuotedCaseStrategy( IdentifierCaseStrategy.MIXED );
+
+		return super.buildIdentifierHelper( builder, dbMetaData );
 	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
@@ -7,6 +7,7 @@
 package org.hibernate.dialect;
 
 import java.sql.CallableStatement;
+import java.sql.DatabaseMetaData;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
@@ -27,6 +28,9 @@ import org.hibernate.dialect.pagination.LimitHandler;
 import org.hibernate.dialect.pagination.LimitHelper;
 import org.hibernate.dialect.unique.MySQLUniqueDelegate;
 import org.hibernate.dialect.unique.UniqueDelegate;
+import org.hibernate.engine.jdbc.env.spi.IdentifierCaseStrategy;
+import org.hibernate.engine.jdbc.env.spi.IdentifierHelper;
+import org.hibernate.engine.jdbc.env.spi.IdentifierHelperBuilder;
 import org.hibernate.engine.spi.RowSelection;
 import org.hibernate.exception.LockAcquisitionException;
 import org.hibernate.exception.LockTimeoutException;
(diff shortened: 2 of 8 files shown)


hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaUpdateTestWithUseJdbcMetadataDeaultsTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+		new SchemaValidator().validate( metadata );
+
+		new SchemaUpdate().setHaltOnError( true )
+				.setOutputFile( output.getAbsolutePath() )
+				.setFormat( false )
+				.execute( EnumSet.of( TargetType.DATABASE, TargetType.SCRIPT ), metadata );
+
+		final String fileContent = new String( Files.readAllBytes( output.toPath() ) );
+		assertThat( "The update output file should be empty", fileContent, is( "" ) );
```
Sanne: I'm not sure I understand this test. Why does it test that a schema update script is _empty_ ?

Shouldn't we rather test that the schema update is correctly applied?

BTW I tried to revert the fix, and the test will still pass - so there's something probably not working as expected?

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaUpdateTestWithUseJdbcMetadataDeaultsTest.java
```diff
@@ -0,0 +1,122 @@
...
+				.execute( EnumSet.of( TargetType.DATABASE ), metadata );
+
+		new SchemaValidator().validate( metadata );
+
+		new SchemaUpdate().setHaltOnError( true )
+				.setOutputFile( output.getAbsolutePath() )
+				.setFormat( false )
+				.execute( EnumSet.of( TargetType.DATABASE, TargetType.SCRIPT ), metadata );
+
+		final String fileContent = new String( Files.readAllBytes( output.toPath() ) );
```
Sanne: should you use `getAbsolutePath` here as well?

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaUpdateTestWithUseJdbcMetadataDeaultsTest.java
dreab8: > Shouldn't we rather test that the schema update is correctly? 

depends what you want to test, the issue is the that the update tries to recreate existing tables so we have to check that the update script is empty.

>  I tried to revert the fix, and the test will still pass  
You have to try it with MySql

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaUpdateTestWithUseJdbcMetadataDeaultsTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+		final MetadataSources metadataSources = new MetadataSources( ssr );
+		metadataSources.addAnnotatedClass( TestEntity.class );
+
+		metadata = (MetadataImplementor) metadataSources.buildMetadata();
+		metadata.validate();
+	}
+
+	@After
+	public void tearsDown() {
```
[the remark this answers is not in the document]
dreab8: a typo

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaUpdateTestWithUseJdbcMetadataDeaultsTest.java
dreab8: `getAbsolutePath()` returns a `String` while `Files.readAllBytes` requires a `Path`

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaUpdateTestWithUseJdbcMetadataDeaultsTest.java
```diff
@@ -0,0 +1,122 @@
...
+import static org.junit.Assert.assertThat;
+
+/**
+ * @author Andrea Boriero
+ */
+@RunWith(Parameterized.class)
+public class SchemaUpdateTestWithUseJdbcMetadataDeaultsTest {
+
+	@Parameterized.Parameters
+	public static Collection<String> parameters() {
```
NathanQingyangXu: btw, it seems you can use `String[]` directly

dreab8: Thank @NathanQingyangXu I applied your suggestion

dreab8:
@Sanne thanks for the review, I pushed a commit with some changes to the test and any further suggestion to improve it is more than welcome.

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaUpdateTestWithUseJdbcMetadataDeaultsTest.java
NathanQingyangXu: should we specify mysql dialect for the testing, then?

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaUpdateTestWithUseJdbcMetadataDeaultsTest.java
NathanQingyangXu: It should be me who say thank you. I borrowed the testing methodology in this PR and applied it in my PR: https://github.com/hibernate/hibernate-orm/pull/3144. During the process, I found we can use array directly. Glad I can help.

Sanne:
hi @dreab8 , I'm back to trying this out and still have some issues.

When I run:

    ./gradlew :hibernate-core:test -Pdb=pgsql --tests SchemaUpdateWithUse*

The two new tests fail: `org.hibernate.tool.schema.spi.SchemaManagementException: Schema-validation: missing table [TEST_ENTITY]`

dreab8:
Hi @Sanne, I checked and it seems the changes fix the issue for MySql but not for PostgreSQL. I'm investigating.

Thanks for testing with PostgreSQL.

Sanne:
AFAIR this relates to the fact that we didn't have the right metadata - could possibly be solved now since we actually load it? Not sure if the timing is right in relation with schema gen.

But yes even if this is still not fixed, it might be easier to solve now.

dreab8:
@yrodiere @Sanne  I locally rebased my branch on main and run the `SchemaUpdateWithUseJdbcMetadataDefaultsSettingToFalseTest` test with PostgreSQL and it passed, I pushed force... let's see if all the tests pass on Github :crossed_fingers:

dreab8:
the SchemaUpdateWithUseJdbcMetadataDefaultsSettingToFalseTest  set `hibernate.temp.use_jdbc_metadata_defaults` to false so for the identifier the default UPPER case strategy is used, the but this is not correct for all dbs and so the schema validation process fails.

dreab8:
@Sanne @yrodiere locally I have a fix for the default strategy issue, I need to test it a little more and not sure if it belongs to the same Jira  or if we need to create a separate Jira for it

PR #3152, opened 2019-12-21
HHH-13792 evict entity cache after transaction has committed for read…

dohertyfjatl:
…-write cache access strategy

- changed ``EntityReadWriteAccess`` to remove the no-op ``unlockRegion`` method (now uses the method inherited from ``AbstractCachedDomainDataAccess``, which calls ``evictAll``)

- changed ``AbstractReadWriteAccess`` to add a no-op ``removeAll`` method (as this is called by the constructor in ``BulkOperationCleanupAction.EntityCleanup``, during the transaction)

- added new file ``ReadWriteCacheTest`` with some test scenarios:
  - ``testDeleteHQL``/``testDeleteNativeQuery``/``testUpdateHQL``/``testUpdateNativeQuery`` which confirm that the fix corrects the scenario where stale entities could be loaded into the cache and remain there (because the eviction was happening before the transaction had committed)
  - ``testDelete``/``testUpdate`` show that entity updates/deletions were not affected by this issue (only HQL/native queries)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/spi/support/AbstractReadWriteAccess.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/spi/support/AbstractReadWriteAccess.java
@@ -198,11 +198,16 @@ public abstract class AbstractReadWriteAccess extends AbstractCachedDomainDataAc
 	public void remove(SharedSessionContractImplementor session, Object key) {
 		if ( getStorageAccess().getFromCache( key, session ) instanceof SoftLock ) {
 			log.debugf( "Skipping #remove call in read-write access to maintain SoftLock : %s", key );
-			// don'tm do anything... we want the SoftLock to remain in place
+			// don't do anything... we want the SoftLock to remain in place
 		}
 		else {
 			super.remove( session, key );
 		}
+	}
+
+	@Override
+	public void removeAll(SharedSessionContractImplementor session) {
+		// A no-op
 	}
 
 	/**
--- a/hibernate-core/src/main/java/org/hibernate/cache/spi/support/EntityReadWriteAccess.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/spi/support/EntityReadWriteAccess.java
@@ -151,9 +151,4 @@ public class EntityReadWriteAccess extends AbstractReadWriteAccess implements En
 	public SoftLock lockRegion() {
 		return null;
 	}
-
-	@Override
-	public void unlockRegion(SoftLock lock) {
-
-	}
 }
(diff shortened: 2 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/cache/spi/support/EntityReadWriteAccess.java
```diff
@@ -151,9 +151,4 @@ public boolean afterUpdate(
 	public SoftLock lockRegion() {
 		return null;
 	}
-
-	@Override
-	public void unlockRegion(SoftLock lock) {
```
dohertyfjatl: Removing this no-op method here means it will inherit the method from ``AbstractCachedDomainDataAccess``, which calls ``evictAll``.

This is invoked by the ``release`` method in ``BulkOperationCleanupAction.EntityCleanup``, which is called as part of ``doAfterTransactionCompletion``.

hibernate-core/src/main/java/org/hibernate/cache/spi/support/AbstractReadWriteAccess.java
```diff
@@ -198,13 +198,18 @@ protected void handleLockExpiry(SharedSessionContractImplementor session, Object
...
-			// don'tm do anything... we want the SoftLock to remain in place
+			// don't do anything... we want the SoftLock to remain in place
 		}
 		else {
 			super.remove( session, key );
 		}
 	}
 
+	@Override
+	public void removeAll(SharedSessionContractImplementor session) {
```
dohertyfjatl: Adding this no-op method here overrides the method in ``AbstractCachedDomainDataAccess`` (which calls ``clearCache``).

 This is invoked by the constructor in ``BulkOperationCleanupAction.EntityCleanup``, which is called during the transaction (so the no-op here avoids the cache being cleared prior to the transaction committing).

hibernate-core/src/test/java/org/hibernate/cache/spi/ReadWriteCacheTest.java
```diff
@@ -0,0 +1,310 @@
...
+
+		public String toString() {
+			return "Book[id=" + id + ",title=" + title + "]";
+		}
+	}
+
+	private final class TransactionInterceptor extends EmptyInterceptor {
+		@Override
+		public void beforeTransactionCompletion(Transaction tx) {
+			if ( interceptTransaction.get() ) {
```
dohertyfjatl: If this value is ``true`` then a separate session will load the entity here:
- (_without the fix_) the cache was already cleared by this point, so the old value gets added to the cache and will remain after the transaction is committed
- (_with the fix_) the cache is cleared after the transaction is committed - so even though the old value is loaded here it will be evicted after this method completes

dohertyfjatl:
@sebersole if you get a chance, would you (or someone else) be able to review this bug fix?
We're quite keen to get a fix for this issue, so if you can help it would be much appreciated!

sebersole:
Looks reasonable to me.  Thanks!

PR #3153, opened 2019-12-21
HHH-11958 Make EntityManager.find() support QueryHints.HINT_READONLY

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-11958

Lots of refactoring code, but the implementation is still straightforward. 

The gist is to rely on `loadQueryInfluencers` to pass the hint, from LoadEvent all the way down to `QueryParameters`.

Testing case added.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
@@ -52,12 +52,19 @@ public class LoadQueryInfluencers implements Serializable {
 
 	private final EffectiveEntityGraph effectiveEntityGraph = new EffectiveEntityGraph();
 
+	private Boolean readOnly;
+
 	public LoadQueryInfluencers() {
-		this( null );
+		this( null, null );
 	}
 
 	public LoadQueryInfluencers(SessionFactoryImplementor sessionFactory) {
+		this(sessionFactory, null);
+	}
+
+	public LoadQueryInfluencers(SessionFactoryImplementor sessionFactory, Boolean readOnly) {
 		this.sessionFactory = sessionFactory;
+		this.readOnly = readOnly;
 	}
 
 	public SessionFactoryImplementor getSessionFactory() {
@@ -282,4 +289,11 @@ public class LoadQueryInfluencers implements Serializable {
 		effectiveEntityGraph.applyGraph( (RootGraphImplementor<?>) loadGraph, GraphSemantic.LOAD );
 	}
 
+	public Boolean getReadOnly() {
+		return readOnly;
+	}
+
+	public void setReadOnly(Boolean readOnly) {
+		this.readOnly = readOnly;
+	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
@@ -376,11 +376,21 @@ public interface EntityPersister extends EntityDefinition {
 	Object load(Serializable id, Object optionalObject, LockMode lockMode, SharedSessionContractImplementor session)
 	throws HibernateException;
 
+	default Object load(Serializable id, Object optionalObject, LockMode lockMode, SharedSessionContractImplementor session, Boolean readOnly)
+	throws HibernateException {
+		return load( id, optionalObject, lockMode, session );
+	}
+
 	/**
 	 * Load an instance of the persistent class.
 	 */
 	Object load(Serializable id, Object optionalObject, LockOptions lockOptions, SharedSessionContractImplementor session)
 	throws HibernateException;
+
+	default Object load(Serializable id, Object optionalObject, LockOptions lockOptions, SharedSessionContractImplementor session, Boolean readOnly)
+			throws HibernateException {
+		return load( id, optionalObject, lockOptions, session );
+	}
 
 	/**
 	 * Performs a load of multiple entities (of this type) by identifier simultaneously.
@@ -546,7 +556,7 @@ public interface EntityPersister extends EntityDefinition {
 	 * Does this class have a natural id cache
 	 */
 	boolean hasNaturalIdCache();
-	
+
 	/**
 	 * Get the NaturalId cache (optional operation)
 	 */
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/NamedQueryLoader.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/NamedQueryLoader.java
@@ -48,15 +48,25 @@ public final class NamedQueryLoader implements UniqueEntityLoader {
 
 	@Override
 	public Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session, LockOptions lockOptions) {
+		return load( id, optionalObject, session, (Boolean) null );
+	}
+
+	@Override
+	public Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session, LockOptions lockOptions, Boolean readOnly) {
 		if ( lockOptions != null ) {
 			LOG.debug( "Ignoring lock-options passed to named query loader" );
 		}
-		return load( id, optionalObject, session );
+		return load( id, optionalObject, session, readOnly );
 	}
 
 	@Override
 	public Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session) {
-		LOG.debugf( "Loading entity: %s using named query: %s", persister.getEntityName(), queryName );
+		return load( id, optionalObject, session, (Boolean) null );
+	}
+
+	@Override
+	public Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session, Boolean readOnly) {
+		LOG.debugf("Loading entity: %s using named query: %s", persister.getEntityName(), queryName);
 
 		// IMPL NOTE: essentially we perform the named query (which loads the entity into the PC), and then
 		// do an internal lookup of the entity from the PC.
@@ -73,6 +83,9 @@ public final class NamedQueryLoader implements UniqueEntityLoader {
 		query.setOptionalEntityName( persister.getEntityName() );
 		query.setOptionalObject( optionalObject );
 		query.setFlushMode( FlushMode.MANUAL );
+		if ( readOnly != null ) {
+			query.setReadOnly( readOnly );
+		}
 		query.list();
 
 		// now look up the object we are really interested in!
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/ejb3configuration/PersisterClassProviderTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/ejb3configuration/PersisterClassProviderTest.java
@@ -264,7 +264,7 @@ public class PersisterClassProviderTest {
 		public boolean hasNaturalIdentifier() {
 			return false;
 		}
-		
+
         @Override
 		public int[] getNaturalIdentifierProperties() {
 			return new int[0];
@@ -280,7 +280,7 @@ public class PersisterClassProviderTest {
 				SharedSessionContractImplementor session) {
 			return null;
 		}
-		
+
 		@Override
         public boolean hasNaturalIdCache() {
             return false;
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
@@ -570,7 +570,8 @@ public class DefaultLoadEventListener implements LoadEventListener {
 				event.getEntityId(),
 				event.getInstanceToLoad(),
 				event.getLockOptions(),
-				event.getSession()
+				event.getSession(),
+				event.getReadOnly()
 		);
 
 		final StatisticsImplementor statistics = event.getSession().getFactory().getStatistics();
--- a/hibernate-core/src/main/java/org/hibernate/event/spi/LoadEvent.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/spi/LoadEvent.java
@@ -46,23 +46,24 @@ public class LoadEvent extends AbstractEvent {
 	private boolean isAssociationFetch;
 	private Object result;
 	private PostLoadEvent postLoadEvent;
+	private Boolean readOnly;
 
-	public LoadEvent(Serializable entityId, Object instanceToLoad, EventSource source) {
-		this( entityId, null, instanceToLoad, DEFAULT_LOCK_OPTIONS, false, source );
+	public LoadEvent(Serializable entityId, Object instanceToLoad, EventSource source, Boolean readOnly) {
+		this( entityId, null, instanceToLoad, DEFAULT_LOCK_OPTIONS, false, source, readOnly );
 	}
 
-	public LoadEvent(Serializable entityId, String entityClassName, LockMode lockMode, EventSource source) {
-		this( entityId, entityClassName, null, lockMode, false, source );
+	public LoadEvent(Serializable entityId, String entityClassName, LockMode lockMode, EventSource source, Boolean readOnly) {
+		this( entityId, entityClassName, null, lockMode, false, source, readOnly );
 	}
 
-	public LoadEvent(Serializable entityId, String entityClassName, LockOptions lockOptions, EventSource source) {
-		this( entityId, entityClassName, null, lockOptions, false, source );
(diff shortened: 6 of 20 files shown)


hibernate-core/src/main/java/org/hibernate/persister/entity/NamedQueryLoader.java
```diff
@@ -47,15 +47,15 @@ public NamedQueryLoader(String queryName, EntityPersister persister) {
 	}
 
 	@Override
-	public Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session, LockOptions lockOptions) {
+	public Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session, LockOptions lockOptions, Boolean readOnly) {
```
jwgmeligmeyling: Probably needs an overloaded variant with the previous signature:

```java
public Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session, LockOptions lockOptions) {
```

hibernate-core/src/main/java/org/hibernate/persister/entity/NamedQueryLoader.java
```diff
@@ -47,15 +47,15 @@ public NamedQueryLoader(String queryName, EntityPersister persister) {
...
 		if ( lockOptions != null ) {
 			LOG.debug( "Ignoring lock-options passed to named query loader" );
 		}
-		return load( id, optionalObject, session );
+		return load( id, optionalObject, session, readOnly );
 	}
 
 	@Override
-	public Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session) {
+	public Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session, Boolean readOnly) {
```
jwgmeligmeyling: Probably needs an overloaded variant with the previous signature:

```java
public Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session) {
```

hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
```diff
@@ -373,15 +373,23 @@ Serializable loadEntityIdByNaturalId(
...
+	}
+
 	/**
 	 * Load an instance of the persistent class.
 	 */
-	Object load(Serializable id, Object optionalObject, LockOptions lockOptions, SharedSessionContractImplementor session)
+	Object load(Serializable id, Object optionalObject, LockOptions lockOptions, SharedSessionContractImplementor session, Boolean readOnly)
 	throws HibernateException;
 
+	default Object load(Serializable id, Object optionalObject, LockOptions lockOptions, SharedSessionContractImplementor session) {
```
jwgmeligmeyling: The default implementation should actually be the one with the new signature in order to remain compatibility with custom persisters.

hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
```diff
@@ -373,15 +373,23 @@ Serializable loadEntityIdByNaturalId(
 	/**
 	 * Load an instance of the persistent class.
 	 */
-	Object load(Serializable id, Object optionalObject, LockMode lockMode, SharedSessionContractImplementor session)
+	Object load(Serializable id, Object optionalObject, LockMode lockMode, SharedSessionContractImplementor session, Boolean readOnly)
 	throws HibernateException;
 
+	default Object load(Serializable id, Object optionalObject, LockMode lockMode, SharedSessionContractImplementor session) {
```
jwgmeligmeyling: The default implementation should actually be the one with the new signature in order to remain compatibility with custom persisters.

hibernate-core/src/test/java/org/hibernate/test/legacy/CustomPersister.java
```diff
@@ -309,9 +309,10 @@ public Object load(
 		Serializable id,
 		Object optionalObject,
 		LockOptions lockOptions,
-		SharedSessionContractImplementor session
+		SharedSessionContractImplementor session,
+		Boolean readOnly
```
jwgmeligmeyling: IMO this class should compile without changes

hibernate-core/src/test/java/org/hibernate/test/cfg/persister/GoofyPersisterClassProvider.java
```diff
@@ -265,12 +265,12 @@ public boolean hasLazyProperties() {
 		}
 
 		@Override
-		public Object load(Serializable id, Object optionalObject, LockMode lockMode, SharedSessionContractImplementor session) {
+		public Object load(Serializable id, Object optionalObject, LockMode lockMode, SharedSessionContractImplementor session, Boolean readOnly) {
```
jwgmeligmeyling: IMO this class should compile without changes

jwgmeligmeyling: IMO this class should compile without changes

hibernate-core/src/test/java/org/hibernate/jpa/test/ejb3configuration/PersisterClassProviderTest.java
```diff
@@ -302,12 +302,12 @@ public boolean hasLazyProperties() {
 		}
 
 		@Override
-		public Object load(Serializable id, Object optionalObject, LockMode lockMode, SharedSessionContractImplementor session) {
+		public Object load(Serializable id, Object optionalObject, LockMode lockMode, SharedSessionContractImplementor session, Boolean readOnly) {
```
jwgmeligmeyling: IMO this class should compile without changes

NathanQingyangXu:
> More thought into the required API changes is needed. Existing persisters should remain compatible without any signature change IMO. Unless this targets 6.x

Thanks for excellennt comments. Will think carefully again from scratch. Sure, this PR is targetting v5, not v6. Thanks

jwgmeligmeyling:
👍 

Usually with persister changes i've seen it done as follows:

- Add a new method with the desired signature
- Replace occurences of the old method to reference the method with the new signature
- Override the old method (potentially with a default if its an interface) to point to the new method with a sensible fallback for the missing attribute (usually the old behaviour).

hibernate-core/src/main/java/org/hibernate/persister/entity/NamedQueryLoader.java
NathanQingyangXu: If you inspect the parent interface, i.e. UniqueEntityLoader, I've already added the following default method as following:
`
	default Object load(
			Serializable id,
			Object optionalObject,
			SharedSessionContractImplementor session,
			LockOptions lockOptions) {
		return load(id, optionalObject, session, lockOptions, null);
	}
`
So it seems I don't need to overload the variant any more.

hibernate-core/src/main/java/org/hibernate/persister/entity/NamedQueryLoader.java
NathanQingyangXu: thanks. The same reason as the above comment.

hibernate-core/src/main/java/org/hibernate/persister/entity/NamedQueryLoader.java
NathanQingyangXu: The previous signature has been implemented as default interface method in its parent interface to route to the new method which accepts a new parameter and hence generalize the old one and is supposed to be the one to be overridden.

hibernate-core/src/main/java/org/hibernate/persister/entity/NamedQueryLoader.java
NathanQingyangXu: again, the previous signature is a deprecated one and has been implemented as default interface method.

hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
NathanQingyangXu: Well, in this case the new method generalizes the old one for it accepts a new parameter while the old one doesn't. For that reason, it seems hiding the old one (it is still overridable for sure) and exposing the new one is better. Maybe I need to add `@Deprecated` to the old method to make it explicit.

hibernate-core/src/main/java/org/hibernate/persister/entity/NamedQueryLoader.java
jwgmeligmeyling: Although deprecated, doing the change this way causes existing persisters and loaders to not compile, as shown by your very own changes in tests. As custom persisters may also exist in user code, this is undesirable.

If the new method is non-abstract, the old method may reuse it. However if the new method is abstract, it should be default and fall back to the old method unless overridden.

Just undo your changes in the unrelated tests and ensure it compiles.

NathanQingyangXu: The new method accepts a new parameter (i.e. readOnly boolean value) which is not accessible from old method. So it seems old method can reuse the new method, but not vice versa. The compiling seems fine for as you can see, this PR is green. The old method is still provided (though a default one is provided to pass `null` value as the new parameter to the new method) and existing client code relying on it won't be broken, unless I were wrong.

NathanQingyangXu: The old method is always there (by default it will route to the new method by passing 'null' as 'readOnly' parameter), so old code won't be broken and compiling should be fine. They simply lack the feature to pass 'readOnly' value.

hibernate-core/src/test/java/org/hibernate/test/cfg/persister/GoofyPersisterClassProvider.java
NathanQingyangXu: Again, the old method is still there (though default implementation is provided):

see the following code changes in EntityPersister:
`	/**
	 * Load an instance of the persistent class.
	 */
	Object load(Serializable id, Object optionalObject, LockMode lockMode, SharedSessionContractImplementor session, Boolean readOnly)
	throws HibernateException;

	default Object load(Serializable id, Object optionalObject, LockMode lockMode, SharedSessionContractImplementor session) {
		return load( id, optionalObject, lockMode, session, null );
	}`

hibernate-core/src/main/java/org/hibernate/persister/entity/NamedQueryLoader.java
NathanQingyangXu: you are right. Updated.

hibernate-core/src/main/java/org/hibernate/persister/entity/NamedQueryLoader.java
NathanQingyangXu: Correct. I updated. Thanks a lot for your precious feedback. I really appreciated it.

hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
NathanQingyangXu: Exactly! Thanks.

hibernate-core/src/test/java/org/hibernate/test/legacy/CustomPersister.java
NathanQingyangXu: Reverted back. Thanks.

hibernate-core/src/test/java/org/hibernate/jpa/test/ejb3configuration/PersisterClassProviderTest.java
NathanQingyangXu: Reverted back. Thanks.

hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
NathanQingyangXu: Exactly! Thanks.

hibernate-core/src/test/java/org/hibernate/test/cfg/persister/GoofyPersisterClassProvider.java
NathanQingyangXu: Reverted it back. Thanks.

NathanQingyangXu:
> 👍
> 
> Usually with persister changes i've seen it done as follows:
> 
> * Add a new method with the desired signature
> * Replace occurences of the old method to reference the method with the new signature
> * Override the old method (potentially with a default if its an interface) to point to the new method with a sensible fallback for the missing attribute (usually the old behaviour).

Tons of thanks. I learned a lot during the process. Please take a new look to see whether something is still wrong.

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
```diff
@@ -52,12 +52,19 @@
 
 	private final EffectiveEntityGraph effectiveEntityGraph = new EffectiveEntityGraph();
 
+	private Boolean readOnly;
+
 	public LoadQueryInfluencers() {
```
jwgmeligmeyling: Perhaps add  `@Deprecated`

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
```diff
@@ -52,12 +52,19 @@
...
-		this( null );
+		this( null, null );
 	}
 
 	public LoadQueryInfluencers(SessionFactoryImplementor sessionFactory) {
```
jwgmeligmeyling: Perhaps add  `@Deprecated`

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
```diff
@@ -52,12 +52,19 @@
 
 	private final EffectiveEntityGraph effectiveEntityGraph = new EffectiveEntityGraph();
 
+	private Boolean readOnly;
```
jwgmeligmeyling: Isn't this really just a `boolean` with default `false`? Probably simplifies some `null` checks elsewhere.

NathanQingyangXu: I chose Boolean type intentionally for sometimes (most of the times) its value is unknown and I wanna rely on some default value to take effect. That being said, I am gonna reconsider its design. Thanks for your comment.

NathanQingyangXu: I think we cannot assume 'false' is the default value for 'readOnly'. For instance, **StatefulPersistenceContext** class has **setDefaultReadOnly**() method and it is likely the default value has been true already. Using Boolean type enjoys the flexibility to delay the decision at the last step to comply with the default value.
Also see QueryParameters's following method:
`
	public boolean isReadOnly(SharedSessionContractImplementor session) {
		return isReadOnlyInitialized
				? isReadOnly()
				: session.getPersistenceContextInternal().isDefaultReadOnly();
	}
`

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
NathanQingyangXu: Again, we have layered design in that null value is absolutely necessary if we don't wanna overwrite the default readOnly value for StatefulPersistenceContext. For that reason, I suggest we don't use `@Deprecated` annotation here. It simply means 'applying readonly as per the StatefulPersistenceContext', though we also have the option (because of the new method) to overwrite the default value.
See the above explanation why Boolean type is chosen.

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
NathanQingyangXu: woa, you work on Sunday?

hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
NathanQingyangXu: same reason as explained below. Thanks. For the same reason I didn't change the reference of old method to new method for both make sense.

sebersole:
Could you do me a favor and squash these commits into 1?  

There are quite a few changes here that will conflict as we apply it to 6.0 work and it is easiest to get them applied if we consolidate those conflicts to as few commits as possible.

Thanks

NathanQingyangXu:
> Could you do me a favor and squash these commits into 1?
> 
> There are quite a few changes here that will conflict as we apply it to 6.0 work and it is easiest to get them applied if we consolidate those conflicts to as few commits as possible.
> 
> Thanks

Done. Let me know what I need to do additionally. Thanks.

PR #3156, opened 2019-12-23
HHH-13796 - Missing from clause in query from BinaryLogicOperatorNode row value constructor translation

jwgmeligmeyling:
https://hibernate.atlassian.net/browse/HHH-13796

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
@@ -377,7 +377,12 @@ class FromElementType {
 		checkInitialized();
 		if ( this.persister != null ) {
 			AbstractEntityPersister aep = (AbstractEntityPersister) this.persister;
-			return aep.getPropertyTableName( propertyName );
+			try {
+				return aep.getSubclassTableName( aep.getSubclassPropertyTableNumber( propertyName ) );
+			}
+			catch (QueryException e) {
+				return null;
+			}
 		}
 		return null;
 	}
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
@@ -33,6 +33,7 @@ import org.hibernate.hql.internal.ast.tree.ImpliedFromElement;
 import org.hibernate.hql.internal.ast.tree.ParameterContainer;
 import org.hibernate.hql.internal.ast.tree.QueryNode;
 import org.hibernate.hql.internal.ast.tree.SqlFragment;
+import org.hibernate.hql.internal.ast.tree.TableReferenceNode;
 import org.hibernate.hql.internal.classic.ParserHelper;
 import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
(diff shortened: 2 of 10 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -167,14 +168,14 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
-				FromReferenceNode fromReferenceNode = (FromReferenceNode) node;
+			if ( node instanceof TableReferenceNode) {
+				TableReferenceNode fromReferenceNode = (TableReferenceNode) node;
 				String[] tables = fromReferenceNode.getReferencedTables();
 				if ( tables != null ) {
 					Collections.addAll(result, tables);
 				}
 			}
-			else if (node instanceof SqlFragment) {
+			if (node instanceof SqlFragment) {
```
NathanQingyangXu: Would be better to have empty char delimiter above as elsewhere, btw

hibernate-core/src/test/java/org/hibernate/jpa/test/jointable/ManyToOneJoinTableTest.java
```diff
@@ -24,11 +45,19 @@
...
+				ResourceImpl.class,
+				IssuerImpl.class
 		};
 	}
 
-	@Test
+	@Override
+	protected void configure(Configuration configuration) {
+		super.configure(configuration);
+//		configuration.setProperty(AvailableSettings.OMIT_JOIN_OF_SUPERCLASS_TABLES, Boolean.FALSE.toString());
```
NathanQingyangXu: Do we have any reason not to delete the above statement which is commented out?

jwgmeligmeyling: Not really, prior the to fix, it can be enabled to show that the fix is in fact a regression caused by the functionality which can be disabled by this property.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
jwgmeligmeyling: Kept the formatting here as is (not my line originally), but yes, we could make the formatting here consistent.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementType.java
```diff
@@ -377,7 +377,12 @@ public String getPropertyTableName(String propertyName) {
 		checkInitialized();
 		if ( this.persister != null ) {
 			AbstractEntityPersister aep = (AbstractEntityPersister) this.persister;
-			return aep.getPropertyTableName( propertyName );
+			try {
+				return aep.getSubclassTableName( aep.getSubclassPropertyTableNumber( propertyName ) );
+			}
+			catch (QueryException e) {
```
beikov: If this could be avoided, that would be great. Otherwise, please document why this is done.

jwgmeligmeyling: Unfortunate decision in the `AbstractEntityPersister`, it should just have returned `null` and not throw any exception, let alone a completely irrelevant `QueryException`... Its just an indication of property not present in the property mapping / metamodel. This happens for many reasons:  for example "nullable" (not optional, but not-proxyable) association properties are always are missing (require a join), nested id class properties are sometimes missing (Hibernate's metamodel is faulty here - known bug). This try-catch statement is done at several places in Hibernate, and can only avoided by extending the AEP with a method that does precisely this. I think having the try-catch block here and there is preferable to extending the already fairly sophisticated API of the AEP, for remaining compatibility with AEP extensions as much as possible.

hibernate-core/src/test/java/org/hibernate/jpa/test/jointable/ManyToOneJoinTableTest.java
```diff
@@ -24,11 +45,19 @@
...
+	}
+
+		@Test
```
NathanQingyangXu: It seems the `@Test` annotation is not vertically aligned with `public`

dreab8:
applied upstream.

Thanks @jwgmeligmeyling for the PR and @beikov and @jwgmeligmeyling for the review.

jwgmeligmeyling:
Thanks!

PR #3157, opened 2019-12-23
HHH-13795 EntityGraph ignored if entity found from cache

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13795


NathanQingyangXu:
Seems pretty difficult to achieve the goal without fundamental code refactoring. Closed for the time being and any contributor is more than welcome to open it in the future.

PR #3158, opened 2019-12-25
HHH-13797 Improve ActionQueue.InsertActionSorter's sort() by topological sorting algorithm

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13797

The benefit of the new implementation compared to previous bubble-sorting based approach:

- time complexity is dropped from quadratic complexity to linear complexity;
- easy way to detect dependency cycle(s);
- elegant implementation translates to better code readability and henceforth maintainability;

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
@@ -105,7 +105,7 @@ public class ActionQueue {
 	 */
 	private static final LinkedHashMap<Class<? extends Executable>,ListProvider> EXECUTABLE_LISTS_MAP;
 	static {
-		EXECUTABLE_LISTS_MAP = new LinkedHashMap<Class<? extends Executable>,ListProvider>( 8 );
+		EXECUTABLE_LISTS_MAP = new LinkedHashMap<>();
 
 		EXECUTABLE_LISTS_MAP.put(
 				OrphanRemovalAction.class,
@@ -1039,23 +1039,24 @@ public class ActionQueue {
 			private final String entityName;
 			private final String rootEntityName;
 
-			private Set<String> parentEntityNames = new HashSet<>( );
+			private Set<String> parentEntityNames = new HashSet<>();
 
-			private Set<String> childEntityNames = new HashSet<>( );
+			private Set<String> childEntityNames = new HashSet<>();
 
-			private BatchIdentifier parent;
+			private int parentCount;
+			private List<BatchIdentifier> children = new LinkedList<>( );
 
 			BatchIdentifier(String entityName, String rootEntityName) {
 				this.entityName = entityName;
 				this.rootEntityName = rootEntityName;
 			}
 
-			public BatchIdentifier getParent() {
-				return parent;
+			public int getParentCount() {
+				return parentCount;
 			}
 
-			public void setParent(BatchIdentifier parent) {
-				this.parent = parent;
+			public List<BatchIdentifier> getChildren() {
+				return children;
 			}
 
 			@Override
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -1170,82 +1151,58 @@ public void sort(List<AbstractEntityInsertAction> insertions) {
...
 				}
-				sorted = true;
 			}
-			while ( !sorted && iterations <= maxIterations );
 
-			if ( iterations > maxIterations ) {
-				LOG.warn( "The batch containing " + latestBatches.size() + " statements could not be sorted after " + maxIterations + " iterations. " +
-								"This might indicate a circular entity relationship." );
+			if ( sortedBatches.size() < latestBatches.size() ) {
+				LOG.warn( "The batch containing " + latestBatches.size() + " statements could not be sorted because of circular dependencies");
```
NathanQingyangXu: we can easily get the cycle participants by computing the two collection diff, if needed.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -1099,34 +1108,6 @@ boolean hasAnyParentEntityNames(BatchIdentifier batchIdentifier) {
...
-			 * @return This {@link BatchIdentifier} has a parent matching the given {@link BatchIdentifier reference
-			 */
-			boolean hasParent(BatchIdentifier batchIdentifier) {
-				return (
-					parent == batchIdentifier
-					|| parentEntityNames.contains( batchIdentifier.getEntityName() )
-					|| ( parentEntityNames.contains( batchIdentifier.getRootEntityName() ) && !this.getEntityName().equals( batchIdentifier.getRootEntityName() ) )
-					|| parent != null && parent.hasParent( batchIdentifier, new ArrayList<>() )
-				);
-			}
```
NathanQingyangXu: The above logic is difficult to comprehend because of various corner cases. It would be excellent to get rid of them if a simpler alternative exists.

NathanQingyangXu: I think the root cause why we still need to rely `parentEntityNames` in face of `parent` is we need multiple `parents`, not single one. It is a graph structure, not a tree structure. Topological sorting algorithm is working fine and accommodate multiple parents.

NathanQingyangXu:
> Change looks good and is easier to read.
> 
> Do you happen to have any performance metrics for large action queues?

That is a good point. I am considering to validate my approach further. Thanks. This PR is supposed to take a long time and I am gonna perfect it further.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -105,7 +105,7 @@
 	 */
 	private static final LinkedHashMap<Class<? extends Executable>,ListProvider> EXECUTABLE_LISTS_MAP;
 	static {
-		EXECUTABLE_LISTS_MAP = new LinkedHashMap<Class<? extends Executable>,ListProvider>( 8 );
```
NathanQingyangXu: The above is a common misuse of `HashMap` or `LinkedHashMap`'s constructor. It doesn't take into consideration of `load refactor`. It is likely the map will be rehashed as long as the buckets number used surpasses 8 * 0.75. Seems the default constructor should be enough here.

sebersole:
On initial glance, the change seem good.  I'm with Jan though - this code needs to be fast.  If the changes are less performant then that is something we need to look at.

Have you done any benchmarking yet?

jwgmeligmeyling:
It actually should be faster now, but I am eager to find out how much it differs in practical terms.

Would be useful to add one or two benchmarks with the pr :) Probably doing a bulk insert with a cascade on an association already fills up the action queue quite enough. Or perhaps create a deep nested tree in a loop and try to remove the root at last. @NathanQingyangXu

NathanQingyangXu:
I believe the elegance is of more concern (previously coding is full of hacks and compromising and so ends up with maintaining roadblock), as long as the performance is not significantly worse.

I would be interested to test it from performance perspective, but currently I am busy with my own stuff (leaving my current company and seeking the next one, :(), so I might not be as ready to attend to the PRs I created in a batch during the long Christmas and New Year holidays, :(.

NathanQingyangXu:
> On initial glance, the change seem good. I'm with Jan though - this code needs to be fast. If the changes are less performant then that is something we need to look at.
> 
> Have you done any benchmarking yet?

Not yet. I need focus and especially I lack many expertise which are mandatory for the performance testing. I am optimistic about the performance improvement (at least we avoided the terrible N**2 worse case which is really really nasty).
Thanks for positive feedback. I do think this PR is the most brilliant one I've ever contributed up to now.

NathanQingyangXu:
@jwgmeligmeyling I would love cooperating with you on perf testing. Feel free to take the initiative if you want. The code is crystal clear in logic and maintainability is improved enormously. I do think this PR is unique in terms of CS concept clearness and algorithm improvement.

jwgmeligmeyling:
I just sent you a message on Zulip.

I am happy with the changes, I am sure they will be more performant, I am interested in how much more performant they are (in particular because besides that, this is untested).

I will approve the PR whenever I had the chance to check it out locally and give it a test spin, just hadn't have the time yet to review it as properly as I want to.

NathanQingyangXu:
No rush. This PR was from the first ticket I created and it is my fav PR up to now. I won't object to see it survives in long testing. Thanks..

NathanQingyangXu:
@gavinking here is an interesting PR. You might be interested, so I added you. Definitively I need more perf testing, but recently I am pretty busy.

PR #3159, opened 2019-12-26
HHH-13799 : Criteria API support for Hibernate Spatial

daniel-shuy:
Currently, Hibernate Spatial functions can be called via Hibernate Criteria API (which is deprecated) or JPQL. To use them in the JPA Criteria API, they have to be manually mapped using `CriteriaBuilder#function(String, Class, Expression...)`, which is tedious and error prone.

This PR provides a wrapper to provide a convenient way to use Hibernate Spatial functions from the JPA Criteria API.

### DIFF
--- a/hibernate-spatial/src/main/java/org/hibernate/spatial/SpatialPredicates.java
+++ b/hibernate-spatial/src/main/java/org/hibernate/spatial/SpatialPredicates.java
@@ -28,6 +28,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially equal" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry expression
 	 *
@@ -49,6 +50,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially equal" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry value
 	 *
@@ -67,6 +69,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially within" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry expression
 	 *
@@ -88,6 +91,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially within" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry value
 	 *
@@ -106,6 +110,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially contains" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry expression
 	 *
@@ -127,6 +132,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially contains" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry value
 	 *
@@ -145,6 +151,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially crosses" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry expression
 	 *
@@ -166,6 +173,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially crosses" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry value
 	 *
@@ -184,6 +192,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially disjoint" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry expression
 	 *
@@ -205,6 +214,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially disjoint" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry value
 	 *
@@ -223,6 +233,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially intersects" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry expression
 	 *
@@ -244,6 +255,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially intersects" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry value
 	 *
@@ -262,6 +274,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially overlaps" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry expression
 	 *
@@ -283,6 +296,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially overlaps" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry value
 	 *
@@ -301,6 +315,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially touches" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry expression
 	 *
@@ -322,6 +337,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "spatially touches" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry value
 	 *
@@ -340,6 +356,7 @@ public class SpatialPredicates {
 	/**
 	 * Create a predicate for testing the arguments for "distance within" constraint.
 	 *
+	 * @param criteriaBuilder CriteriaBuilder
 	 * @param geometry1 geometry expression
 	 * @param geometry2 geometry expression
 	 * @param distance distance expression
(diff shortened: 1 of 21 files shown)


daniel-shuy:
It seems like the `SpatialPredicates` tests I added (and the `SpatialRestrictions` tests it was based on) are being skipped in the CI build. When I tried to run them locally using Gradle they were also being skipped. How can I run the tests?

hibernate-spatial/src/main/java/org/hibernate/spatial/SpatialPredicates.java
```diff
@@ -0,0 +1,492 @@
...
+public class SpatialPredicates {
+
+	private SpatialPredicates() {
+	}
+
+	/**
+	 * Create a predicate for testing the arguments for "spatially equal" constraint.
+	 *
+	 * @param geometry1 geometry expression
+	 * @param geometry2 geometry expression
```
NathanQingyangXu: seems the first parameter is missing above

hibernate-spatial/src/main/java/org/hibernate/spatial/SpatialPredicates.java
```diff
@@ -0,0 +1,492 @@
...
+import org.hibernate.spatial.criterion.SpatialRestrictions;
+
+import org.locationtech.jts.geom.Geometry;
+
+/**
+ * JPA Criteria API support for hibernate-spatial.
+ *
+ * @author Daniel Shuy
+ */
+public class SpatialPredicates {
```
NathanQingyangXu: It might be considerable to specify the class as `final` if it is meant only for static method container.

hibernate-spatial/src/test/java/org/hibernate/spatial/integration/TestSpatialPredicates.java
```diff
@@ -0,0 +1,250 @@
...
+import java.util.Map;
+import java.util.function.BiFunction;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+/**
+ * @see SpatialRestrictions
+ */
+@Skip(condition = SpatialDialectMatcher.class, message = "No Spatial Dialect")
```
NathanQingyangXu: IMHO,  the above annotation explains why the testing is skipped. "No Spatial Dialect" to test against.

hibernate-spatial/src/main/java/org/hibernate/spatial/SpatialPredicates.java
daniel-shuy: Good point. I'm too spoiled by Lombok's `@UtilityClass` nowadays :P

NathanQingyangXu:
> It seems like the `SpatialPredicates` tests I added (and the `SpatialRestrictions` tests it was based on) are being skipped in the CI build. When I tried to run them locally using Gradle they were also being skipped. How can I run the tests?

I guess you might need to come up with a dialect supporting spatial feature. Currently it seems there is no such one from the `@Skip` annotation comment of your testing class.

daniel-shuy:
> 
> 
> > It seems like the `SpatialPredicates` tests I added (and the `SpatialRestrictions` tests it was based on) are being skipped in the CI build. When I tried to run them locally using Gradle they were also being skipped. How can I run the tests?
> 
> I guess you might need to come up with a dialect supporting spatial feature. Currently it seems there is no such one from the `@Skip` annotation comment of your testing class.

I copied the `@Skip` annotation from [TestSpatialRestrictions](https://github.com/hibernate/hibernate-orm/blob/5.4.10/hibernate-spatial/src/test/java/org/hibernate/spatial/integration/TestSpatialRestrictions.java). I assumed the CI build is supposed to run a matrix of all the SQL Dialects, so it should skip the test for non-spatial dialects, but not skip for spatial dialects, eg. Postgis (see the source for [SpatialDialectMatcher](https://github.com/hibernate/hibernate-orm/blob/5.4.10/hibernate-spatial/src/test/java/org/hibernate/spatial/testing/SpatialDialectMatcher.java)), but that doesn't seem to be the case.

How can I run the tests with a spatial dialect?

NathanQingyangXu:
> > > It seems like the `SpatialPredicates` tests I added (and the `SpatialRestrictions` tests it was based on) are being skipped in the CI build. When I tried to run them locally using Gradle they were also being skipped. How can I run the tests?
> > 
> > 
> > I guess you might need to come up with a dialect supporting spatial feature. Currently it seems there is no such one from the `@Skip` annotation comment of your testing class.
> 
> I copied the `@Skip` annotation from [TestSpatialRestrictions](https://github.com/hibernate/hibernate-orm/blob/5.4.10/hibernate-spatial/src/test/java/org/hibernate/spatial/integration/TestSpatialRestrictions.java). I assumed the CI build is supposed to run a matrix of all the SQL Dialects, so it should skip the test for non-spatial dialects, but not skip for spatial dialects, eg. Postgis (see the source for [SpatialDialectMatcher](https://github.com/hibernate/hibernate-orm/blob/5.4.10/hibernate-spatial/src/test/java/org/hibernate/spatial/testing/SpatialDialectMatcher.java)), but that doesn't seem to be the case.
> 
> How can I run the tests with a spatial dialect?

You simply use `@RequireDialect` or `@RequiresDialectFeature` (which is much better).

daniel-shuy:
> 
> 
> > > > It seems like the `SpatialPredicates` tests I added (and the `SpatialRestrictions` tests it was based on) are being skipped in the CI build. When I tried to run them locally using Gradle they were also being skipped. How can I run the tests?
> > > 
> > > 
> > > I guess you might need to come up with a dialect supporting spatial feature. Currently it seems there is no such one from the `@Skip` annotation comment of your testing class.
> > 
> > 
> > I copied the `@Skip` annotation from [TestSpatialRestrictions](https://github.com/hibernate/hibernate-orm/blob/5.4.10/hibernate-spatial/src/test/java/org/hibernate/spatial/integration/TestSpatialRestrictions.java). I assumed the CI build is supposed to run a matrix of all the SQL Dialects, so it should skip the test for non-spatial dialects, but not skip for spatial dialects, eg. Postgis (see the source for [SpatialDialectMatcher](https://github.com/hibernate/hibernate-orm/blob/5.4.10/hibernate-spatial/src/test/java/org/hibernate/spatial/testing/SpatialDialectMatcher.java)), but that doesn't seem to be the case.
> > How can I run the tests with a spatial dialect?
> 
> You simply use `@RequireDialect` or `@RequiresDialectFeature` (which is much better).

That makes no difference. Even if I restrict the test to run with a spatial dialect, when running `gradlew check` (the command the Travis CI job is running) it only runs using the H2 dialect, which is not a spatial dialect.

Is the only way to explicitly specify every spatial dialect for the test?

NathanQingyangXu:
i remember the matrix testing pipeline will test your case against the dialect you specified, unless I am wrong. As a contributor,  I test locally by overwrite some url placeholders after I set up a local testing db.

daniel-shuy:
> 
> 
> i remember the matrix testing pipeline will test your case against the dialect you specified, unless I am wrong. As a contributor, I test locally by overwrite some url placeholders after I set up a local testing db.

I tried with `@RequiresDialect(PostgisDialect.class)`, and even then it was still using the H2 dialect.

NathanQingyangXu:
that is surprising to me. hopefully some team member can help you. you can also sign up in zulip. All team members use it and I am sure you will get answer there.

NathanQingyangXu:
maybe their pipeline is not ready for spatial testing and that explains the @skip comment you encouteted.

hibernate-spatial/src/test/java/org/hibernate/spatial/integration/TestSpatialPredicates.java
```diff
@@ -0,0 +1,248 @@
...
+		for ( Map.Entry<Integer, Boolean> entry : dbexpected.entrySet() ) {
+			if ( entry.getValue() ) {
+				cnt++;
+				if ( !findInList( entry.getKey(), list ) ) {
+					fail( String.format( "Expected object with id= %d, but not found in result", entry.getKey() ) );
+				}
+			}
+		}
+		assertEquals( cnt, list.size() );
+		LOG.info( String.format( "Found %d objects within testsuite-suite polygon.", cnt ) );
```
NathanQingyangXu: Seems we might use `LOG.infof("Found %d objects within testsuite-suite polygon.", cnt )` as well.

daniel-shuy: This was also something I copied from `TestSpatialRestrictions`, but I guess I'll change it and also improve other parts of the tests that were copied from `TestSpatialRestrictions`

hibernate-spatial/src/main/java/org/hibernate/spatial/predicate/FilterPredicate.java
```diff
@@ -0,0 +1,82 @@
...
+	@Override
+	public String render(boolean isNegated, RenderingContext renderingContext) {
+		String geometryParameter = ( (Renderable) geometry ).render( renderingContext );
+		String filterParameter = ( (Renderable) filter ).render( renderingContext );
+		Dialect dialect = renderingContext.getDialect();
+		if ( dialect instanceof SpatialDialect ) {
+			final SpatialDialect seDialect = (SpatialDialect) dialect;
+			String sql = seDialect.getSpatialFilterExpression( geometryParameter );
+			return REGEX_PATTERN_PARAMETER_PLACEHOLDER.matcher( sql )
+					.replaceFirst( filterParameter );
```
daniel-shuy: Anyone knows if there is a better way to replace unnamed parameters (`?`) in the rendered SQL String in an `AbstractSimplePredicate`/`ExpressionImpl` implementation, other than regex replace?

daniel-shuy:
@NathanQingyangXu you're right, the CI pipeline does not support Spatial dialects.

When running locally, do I have to manually install the databases locally? Are there specific versions of each database vendor to test against? Is there a docker-compose or Kubernetes file to set them up?

NathanQingyangXu:
> @NathanQingyangXu you're right, the CI pipeline does not support Spatial dialects.
> 
> When running locally, do I have to manually install the databases locally? Are there specific versions of each database vendor to test against? Is there a docker-compose or Kubernetes file to set them up?

I don't think there is a docker-compose or Kubernetes file to use out-of-the-box. You need to set up a db locally in whatever way and the configuration to test locally is not difficult. I take 'mariadb' as an example:

- locate the hibernate.properties in databases/mariadb/resources folder
- tweak the various db connection parameters if necessary as per its content
- launch testing by specifying './gradlew -Pdb=mariadb xxxx' to specify the db; without '-Pdb=xxx', h2 db will be used for testing (as long as your testing does't dictate other dialect)

However, I am not sure how to test other db which doesn't exist in 'databases' folder. Guess you can create one to explore. Should be straightforward, I guess.

IMHO, testing successfully locally is not difficult, how to persuade code reviewers that your code doesn't work might be challenging, provided CI pipeline doesn't support. Obviously you cannot expect them to setup a local db as what you have done. You might need to talk with them further on how to validate your PR further, if needed.
Good luck!

daniel-shuy:
I managed to test with:
- MySQL
- Postgis
- SQL Server

and resolved all the issues.

I couldn't get Oracle to work, but it was failing even the existing Spatial tests (eg. [TestSpatialRestrictions](https://github.com/hibernate/hibernate-orm/blob/5.4.10/hibernate-spatial/src/test/java/org/hibernate/spatial/integration/TestSpatialRestrictions.java)). See https://hibernate.zulipchat.com/#narrow/stream/132096-hibernate-user/topic/Spatial.20tests.20with.20Oracle/near/186297797.

I couldn't get H2 + GeoDB to work, due to the project being unable to load the GeoDB JDBC driver (`org.opengeo:geodb`). Versions below `0.9` wouldn't work because they depended on the old version of JTS (`com.vividsolutions:jts`), while the project depends on the new version of JTS (`org.locationtech.jts:jts-core`). However, even using version `0.9` I received a `ClassNotFoundException` due to the ClassLoader not finding `net.sourceforge.hatbox.tools.CmdLine`. I tried adding the [Boundless Maven repository](https://repo.boundlessgeo.com/main/) to the Gradle build, and also tried adding the `net.sourceforge.hatbox:hatbox` JAR dependency directly to the project, but both approaches didn't work.

I couldn't get HANA (HANA Express on Docker) to work. It would consistently fail at the `hdbnsutil: creating persistence` step, but with different reasons each time.

NathanQingyangXu:
btw, you can find the ide code style file here https://github.com/hibernate/hibernate-ide-codestyles

daniel-shuy:
Thanks @maesenka!

I already added the checkstyle configuration to my IDE, but maybe I missed out formatting a few files.

daniel-shuy:
Thanks @NathanQingyangXu and @maesenka!

PR #3160, opened 2019-12-31
HHH-13800 Fix typos in Javadoc

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13800

Tons of stuff, but not difficult to review. Typos elsewhere (like in the line comment) are ignored. The scope of Javadoc has been 
big enough. Maybe I should have narrowed the scope to exclude 'internal' packages.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/transform/AliasToBeanConstructorResultTransformer.java
+++ b/hibernate-core/src/main/java/org/hibernate/transform/AliasToBeanConstructorResultTransformer.java
@@ -22,12 +22,12 @@ public class AliasToBeanConstructorResultTransformer implements ResultTransforme
 	/**
 	 * Instantiates a AliasToBeanConstructorResultTransformer.
 	 *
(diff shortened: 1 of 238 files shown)


hibernate-core/src/main/java/org/hibernate/transform/AliasToBeanConstructorResultTransformer.java
```diff
@@ -1,74 +1,74 @@
-/*
```
jwgmeligmeyling: In this file also the indentation or line endings changed. Intentionally?

NathanQingyangXu: Not intentionally. It seems github abandoned the diff effort when too many files are included. Maybe after some timeout?

Sanne:
Awesome, thanks I've merged it.

PR #3161, opened 2020-01-05
HHH-13802 Fix javadoc warnings

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13802

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/Type.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/Type.java
@@ -117,7 +117,7 @@ public interface Type extends Serializable {
 	 * NOTE: The number of elements in this array matches the return from {@link #getColumnSpan}.
 	 *
 	 * @param mapping The mapping object :/
-	 * TODO: Would be much much better to have this aware of Dialect once the service/metamodel split is done
+	 * @todo Would be much much better to have this aware of Dialect once the service/metamodel split is done
 	 *
 	 * @return The dictated sizes.
 	 *
@@ -132,7 +132,7 @@ public interface Type extends Serializable {
 	 * NOTE: The number of elements in this array matches the return from {@link #getColumnSpan}.
 	 *
 	 * @param mapping The mapping object :/
-	 * TODO: Would be much much better to have this aware of Dialect once the service/metamodel split is done
+	 * @todo Would be much much better to have this aware of Dialect once the service/metamodel split is done
 	 *
 	 * @return The default sizes.
 	 *
--- a/hibernate-core/src/main/java/org/hibernate/type/Type.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/Type.java
@@ -117,7 +117,7 @@ public interface Type extends Serializable {
 	 * NOTE: The number of elements in this array matches the return from {@link #getColumnSpan}.
 	 *
 	 * @param mapping The mapping object :/
-	 * @todo Would be much much better to have this aware of Dialect once the service/metamodel split is done
+	 * TODO: Would be much much better to have this aware of Dialect once the service/metamodel split is done
 	 *
 	 * @return The dictated sizes.
 	 *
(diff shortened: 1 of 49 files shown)


hibernate-core/src/main/java/org/hibernate/type/Type.java
```diff
@@ -117,7 +117,7 @@
 	 * NOTE: The number of elements in this array matches the return from {@link #getColumnSpan}.
 	 *
 	 * @param mapping The mapping object :/
-	 * @todo Would be much much better to have this aware of Dialect once the service/metamodel split is done
+	 * TODO: Would be much much better to have this aware of Dialect once the service/metamodel split is done
```
jwgmeligmeyling: I think one of the reasons to use an annotation here is so that it doesn't end up in the generated Javadoc (because there is no doclet registered for the @todo annotation).

I don't think TODO's should end up in the Javadoc, so perhaps we should move this to a regular comment below instead.

hibernate-core/src/main/java/org/hibernate/type/Type.java
```diff
@@ -132,7 +132,7 @@
...
```
jwgmeligmeyling: I think one of the reasons to use an annotation here is so that it doesn't end up in the generated Javadoc (because there is no doclet registered for the `@todo` annotation).

I don't think TODO's should end up in the Javadoc, so perhaps we should move this to a regular comment below instead.

hibernate-core/src/main/java/org/hibernate/type/Type.java
sebersole: The Javadoc tool has the ability to recognize taglets like "@todo".  I wonder if there is a way to have Javadoc recognize it, but not render them?

sebersole:
Awesome that you did this - it is very annoying.

Overall it looks great, although I do agree with @jwgmeligmeyling wrt `@todo`

hibernate-core/src/main/java/org/hibernate/type/Type.java
NathanQingyangXu: I used to struggle with the `@todo` a lot initially. Later on I reverted back all the changes. This is a leftover I forgot.

NathanQingyangXu:
> Awesome that you did this - it is very annoying.
> 
> Overall it looks great, although I do agree with @jwgmeligmeyling wrt `@todo`

Thanks. One thing bothering me a lot is sometimes we use `TODO:` and sometimes we use `@todo` in Javadoc. So I should use `@todo` consistently?

hibernate-core/src/main/java/org/hibernate/type/Type.java
NathanQingyangXu: I reverted back the old `@todo`. Thanks.

hibernate-core/src/main/java/org/hibernate/type/Type.java
NathanQingyangXu: The old `@todo` is reverted back. Thanks.

NathanQingyangXu:
> Awesome that you did this - it is very annoying.
> 
> Overall it looks great, although I do agree with @jwgmeligmeyling wrt `@todo`

I tried to fix as many warnings as possible, but there are still some left unsolved. I even read something about JDK8's bug in terms of Javadoc: https://stackoverflow.com/questions/40747275/what-is-the-proper-link-or-see-javadoc-tag-for-inner-nested-class-construc

sebersole:
> > Awesome that you did this - it is very annoying.
> > Overall it looks great, although I do agree with @jwgmeligmeyling wrt `@todo`
> 
> Thanks. One thing bothering me a lot is sometimes we use `TODO:` and sometimes we use `@todo` in Javadoc. So I should use `@todo` consistently?

IMO we'd ideally use `@todo` anytime the todo appears in Javadoc.  Is it worth the effort to find all non-tagged todos in Javadoc and replace them?  I'd say probably not.  Though I'm not stopping anyone who wants to take that on ;)

sebersole:
> I tried to fix as many warnings as possible, but there are still some left unsolved. I even read something about JDK8's bug in terms of Javadoc: https://stackoverflow.com/questions/40747275/what-is-the-proper-link-or-see-javadoc-tag-for-inner-nested-class-construc

Agreed that is an annoying bug.  The only option there would be to use Java 9 in the build, which I'm not sure we want to do overall.  Let me think about that

sebersole:
Pushed upstream.  Thanks Nathan and Jan!

PR #3162, opened 2020-01-05
HHH-5508 Detect setFetchMode() ignore incorrect path

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-5508

This is the oldest unresolved ticket I can find.


NathanQingyangXu:
The implementation in this PR is incomplete and not totally correct (e.g. retrieval by 'natural id' would fail). Need to find better way to achieve the goal.

PR #3163, opened 2020-01-08
HHH-8776 make Hibernate respect JPA's FETCH entity graph semantic

NathanQingyangXu:
We did respect JPA's FETCH semantic when generating SQL, but at second phase loading stage, we ignored the information and load association purely based on its type. The fixing includes the following steps:

1. pass **HydratedEntityRegistration**'s **entityReference** from AbstractRowReader to **TwoPhaseLoad**'s **initializeEntity** method;
2. in **TwoPhaseLoad** class, grab the info to access to **FetchStrategy** for relevant property as a default value (could be null) which can be still overridden by fetching profile.

The relevant content in user guide doc is updated (actually it is the motivation to fix this PR in the first place). Testing case attached.


NathanQingyangXu:
This PR has ripple effect in that it brings about 45 broken testing cases. Some of them might be expected, while others might not. For that reason, I closed this PR to investigate them.

PR #3164, opened 2020-01-09
HHH-8776 make Hibernate comply with JPA's FETCH graph semantic

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-8776

This is an important PR for it aims to make Hibernate adhere to JPA's entity graph feature (FETCH graph semantic dictates that loading should be lazy by default invariably).

We did right for load event handling and SQL generation, but went out of sync in later preload event processing or TwoPhaseLoader. 

The implementation is straightforward:

- enrich existing **getOverridingEager()** method in the same class to add the logic to return entity graph's default value, right after existing **FetchProfile** overriding logic.

Testing case is attached.


NathanQingyangXu:
This is a complicated PR and more thorough testing is needed. Maybe dramatic refactoring is needed.

PR #3165, opened 2020-01-10
HHH-13808: Fix String format in log

selliera:
Use %s in the log string format, to avoid java.util.UnknownFormatConversionException: Conversion = ']' when enabling debug logs.

Fixes HHH-13808


dreab8:
applied upstream.
Thanks @selliera

PR #3169, opened 2020-01-13
HHH-13694 fix numeric overflow exception for large sequence min values

thokuest:
Fixes a numeric overflow exception for Oracle sequence min values that
exceed `Long.minValue()`. In that case `Long.minValue()` is returned
instead.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/dialect/functional/OracleDialectSequenceInformationTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/dialect/functional/OracleDialectSequenceInformationTest.java
@@ -1,0 +1,98 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.test.dialect.functional;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.Optional;
+import java.util.stream.StreamSupport;
+
+import org.hibernate.boot.registry.StandardServiceRegistry;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.dialect.Dialect;
+import org.hibernate.dialect.Oracle12cDialect;
+import org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentImpl;
+import org.hibernate.engine.jdbc.env.spi.JdbcEnvironment;
+import org.hibernate.engine.jdbc.spi.JdbcServices;
+import org.hibernate.testing.RequiresDialect;
+import org.hibernate.testing.TestForIssue;
+import org.hibernate.testing.junit4.BaseUnitTestCase;
+import org.hibernate.tool.schema.extract.internal.SequenceInformationExtractorOracleDatabaseImpl;
+import org.hibernate.tool.schema.extract.spi.ExtractionContext;
+import org.hibernate.tool.schema.extract.spi.SequenceInformation;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInAutoCommit;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+@RequiresDialect(value = { Oracle12cDialect.class })
+@TestForIssue(jiraKey = "HHH-13694")
+public class OracleDialectSequenceInformationTest extends BaseUnitTestCase {
+
+	private final String SEQUENCE_NAME = "SEQ_MIN_TEST";
+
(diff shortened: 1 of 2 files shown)


thokuest:
First time contributor here! 

`BigDecimal.longValue()` documentation reads

> Note that this conversion can lose information about the overall magnitude and precision of this BigDecimal value as wellas return a result with the opposite sign.

That's why `Long.minValue()` is returned if the value too large. I wonder, should I do this for `resultSetMaxValue()` as well?

hibernate-core/src/test/java/org/hibernate/test/dialect/functional/OracleDialectSequenceInformationTest.java
```diff
@@ -0,0 +1,98 @@
...
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.test.dialect.functional;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInAutoCommit;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
```
NathanQingyangXu: Hibernate's code convention is to put static imports at the end of list, not at the beginning. See other class code for example.

thokuest: Hm, strange. I imported [hibernate-java-formatting.xml](https://raw.githubusercontent.com/hibernate/hibernate-ide-codestyles/master/eclipse/hibernate-java-formatting.xml) into Eclipse but this doesn't seem to do the trick. However, I'll apply your suggestion. Thank you!

thokuest: Ah, I see, [Bug 304415 - [organize imports] Organization of static imports in Java is not configurable](https://bugs.eclipse.org/bugs/show_bug.cgi?id=304415).

NathanQingyangXu: thanks. other than that, your PR seems pretty nice.

thokuest: Thanks! What about the issue in `resultSetMaxValue()`? Should this be a different PR/issue?

NathanQingyangXu: TBH, I am not sure. Sooner or later some team member (I am not) will review for this ticket is labelled as next release version. He will answer your question, I guess.

sebersole:
Welcome Thomas!

I'm actually tempted to just drop these 2 methods from `SequenceInformation`.  We never use them in any meaningful way.  Or change the return type for these 2 methods from `Long` to `BigInteger` - it could be `BigDecimal` instead, but the value is implicitly an integer (in the whole number sense).

I guess at this point it is too late in the game to do either of these, so something like your change is fine - like I said, we never use these values anyway.  We should definitely deprecate these 2 methods IMO.

To your specific question - IMO, if we do it for one we should do it for the other as well

thokuest:
@sebersole, regarding

> To your specific question - IMO, if we do it for one we should do it for the other as well

I added a commit that does the same for `resultSetMaxValue()`. Regarding the question whether to deprecate and eventually remove the two methods from the `SequenceInformationExtractor` and `SequenceInformation` implementation, or to change return type, can I leave it with you to or shall I file an issue with JIRA?

Axinet:
Any plans of including this PR? That issue blocks us from upgrading Hibernate to newer versions.

hibernate-core/src/test/java/org/hibernate/test/dialect/functional/OracleDialectSequenceInformationTest.java
josevicentecruz: Hi guys. Any chances we can get this PR merged soon?

MattCockman:
Hi. To echo the above comment  - are there any plans to merge this?

cuneytcalishkan:
> Any plans for merging this PR? That issue blocks us from upgrading Hibernate to newer versions.

+1 here for the same reason that we cannot upgrade to the latest version.

beikov:
I personally don't think this fix is appropriate. IMO we should introduce new methods that return `BigDecimal`, `BigInteger` or `Number` with default implementations that just wrap the result of these older methods. The Oracle implementation should provide these new methods. In that process, one would also have to change all uses to the new methods. The conversion to the user type e.g. `Long` should IMO be dependent on the actual user type of a generated field/property and should still throw an overflow exception in some cases to signal, that the user chose a too small data type.

beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/4887

PR #3176, opened 2020-01-18
Changes for Hibernate RX

gavinking:
This patch contains a bunch of very minor changes that help simplify the implementation of hibernate-rx, allowing us to do away with a whole lot of copy/paste and other hacks.

There's nothing very interesting here, mostly just changes to access modifiers and removal of `final` from a handful of classes.

EDIT: I've added the following additional changes:

- Make Hibernate always generate lowercase column names in SQL `select`s. This was needed due to the behavior of the SmallRye postgres driver.
- Allow instantiation of `Delete`/`Update`/`Insert` objects to be delegated to the subclass of a persister, allowing hibernate-rx to more easily override the bind variable syntax.

EDIT 2: Today I've made some additional visibility changes, to `MergeContext`, `EntityIdentityInsertAction`, and `SessionImpl`.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/EntityState.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/EntityState.java
@@ -1,0 +1,71 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.event.internal;
+
+import org.hibernate.engine.internal.ForeignKeys;
+import org.hibernate.engine.spi.EntityEntry;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.engine.spi.Status;
+import org.hibernate.internal.CoreLogging;
+import org.hibernate.internal.CoreMessageLogger;
+
+public enum EntityState {
+	PERSISTENT, TRANSIENT, DETACHED, DELETED;
+
+	static final CoreMessageLogger LOG = CoreLogging.messageLogger( EntityState.class );
+
+	/**
+	 * Determine whether the entity is persistent, detached, or transient
+	 *
+	 * @param entity The entity to check
+	 * @param entityName The name of the entity
+	 * @param entry The entity's entry in the persistence context
+	 * @param source The originating session.
+	 *
+	 * @return The state.
+	 */
+	public static EntityState getEntityState(
+			Object entity,
+			String entityName,
+			EntityEntry entry, //pass this as an argument only to avoid double looking
+			SessionImplementor source,
+			Boolean assumedUnsaved) {
+
+		if ( entry != null ) { // the object is persistent
+
+			//the entity is associated with the session, so check its status
+			if ( entry.getStatus() != Status.DELETED ) {
+				// do nothing for persistent instances
+				if ( LOG.isTraceEnabled() ) {
+					LOG.tracev( "Persistent instance of: {0}", EventUtil.getLoggableName( entityName, entity ) );
+				}
+				return PERSISTENT;
+			}
+			// ie. e.status==DELETED
+			if ( LOG.isTraceEnabled() ) {
+				LOG.tracev( "Deleted instance of: {0}", EventUtil.getLoggableName( entityName, entity ) );
+			}
+			return DELETED;
+		}
+		// the object is transient or detached
+
+		// the entity is not associated with the session, so
+		// try interceptor and unsaved-value
+
+		if ( ForeignKeys.isTransient( entityName, entity, assumedUnsaved, source ) ) {
+			if ( LOG.isTraceEnabled() ) {
+				LOG.tracev( "Transient instance of: {0}", EventUtil.getLoggableName( entityName, entity ) );
+			}
+			return TRANSIENT;
+		}
+		if ( LOG.isTraceEnabled() ) {
+			LOG.tracev( "Detached instance of: {0}", EventUtil.getLoggableName( entityName, entity ) );
+		}
+		return DETACHED;
+	}
+
+}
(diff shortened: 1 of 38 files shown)


Sanne:
@dreab8 FYI: have a quick look as it might make your next rebase a bit bumpy

dreab8:
@Sanne no problem for the merge with 6.0, I'll manage it just not sure if the change of visibility of few methods in  the`AbstractEntityPersister` can be an issue for a micro release, if you think it is not then I think you can merge it :)

hibernate-core/src/main/java/org/hibernate/event/internal/EntityState.java
```diff
@@ -0,0 +1,71 @@
...
+
+	static final CoreMessageLogger LOG = CoreLogging.messageLogger( EntityState.class );
+
+	/**
+	 * Determine whether the entity is persistent, detached, or transient
+	 *
+	 * @param entity The entity to check
+	 * @param entityName The name of the entity
+	 * @param entry The entity's entry in the persistence context
+	 * @param source The originating session.
```
NathanQingyangXu: Seems inconsistent period in the above `@param` usages. Also, should we include 'deleted' into the first sentence?

hibernate-core/src/main/java/org/hibernate/event/internal/EntityState.java
```diff
@@ -0,0 +1,71 @@
...
+
+			//the entity is associated with the session, so check its status
+			if ( entry.getStatus() != Status.DELETED ) {
+				// do nothing for persistent instances
+				if ( LOG.isTraceEnabled() ) {
+					LOG.tracev( "Persistent instance of: {0}", EventUtil.getLoggableName( entityName, entity ) );
+				}
+				return PERSISTENT;
+			}
+			// ie. e.status==DELETED
```
NathanQingyangXu: given we are using `entity`, maybe changing it to
```
// i.e., entity.getStatus() == Status.DELETED
```

hibernate-core/src/main/java/org/hibernate/event/internal/EntityState.java
```diff
@@ -0,0 +1,71 @@
...
+	public static EntityState getEntityState(
+			Object entity,
+			String entityName,
+			EntityEntry entry, //pass this as an argument only to avoid double looking
+			SessionImplementor source,
+			Boolean assumedUnsaved) {
+
+		if ( entry != null ) { // the object is persistent
+
+			//the entity is associated with the session, so check its status
```
NathanQingyangXu: It all boils down to personal preference, but I would leave a space char after `//` to be consistent and avoid my eyes taking some used to it. Super minor off course.

hibernate-core/src/main/java/org/hibernate/event/internal/EntityState.java
```diff
@@ -0,0 +1,71 @@
...
+import org.hibernate.internal.CoreLogging;
+import org.hibernate.internal.CoreMessageLogger;
+
+public enum EntityState {
+	PERSISTENT, TRANSIENT, DETACHED, DELETED;
+
+	static final CoreMessageLogger LOG = CoreLogging.messageLogger( EntityState.class );
+
+	/**
+	 * Determine whether the entity is persistent, detached, or transient
```
NathanQingyangXu: `Determines` would be more consistent with existing Javadoc pattern, though not a big deal for sure.

gavinking:
@NathanQingyangXu `EntityState` is not a new class, it is a class that was previously an inner class or `AbstractSaveListener`, that I moved to the package level in order to be able to access it in Hibernate RX. Those are the same code formatting and Javadoc issues that it has had for who knows how many years.

NathanQingyangXu:
> @NathanQingyangXu `EntityState` is not a new class, it is a class that was previously an inner class or `AbstractSaveListener`, that I moved to the package level in order to be able to access it in Hibernate RX. Those are the same code formatting and Javadoc issues that it has had for who knows how many years.

Thanks. I noticed that after my comments are left.

Sanne:
> This may break any custom `EntityPersister` that extends `AbstractEntityPersister` with `protected` overrides for methods for which the visibility has been widened. Extending `AbstractEntityPersister` is the go-to method for implementing custom `EntityPersisters` - due to the complexity - and is not a completely odd thing to do (there even is the `org.hibernate.annotations.Persister` annotation for it, and it is documented in the user guide as well). Some Hibernate integrations also rely on the persisters. As a result, I think `AbstractEntityPersister` should be considered SPI/API.
> 
> Provided that only a few methods changed visibility in the `AbstractEntityPersister`, that these are not very likely to be extended and that even when they are extended, the issue is relatively easy worked around by changing the visiblity in the affected code - this should be okay for integration in 5.4. With this I also take into account that 5.4 is the development version of Hibernate and hasn't made it into any product yet.
> 
> Nonetheless, I think semver-wise this would really require a minor version bump at least. I am personally OK with this even in 5.4.x, but it should be merged with some caution :)

Good points @jwgmeligmeyling ; you almost got me worried on the widening of access of protected methods, but then I remembered this is actually not a breaking change in terms of binary compability: 
if people have already compiled their extensions _before_ upgrading they will be fine as this is ok in terms of bytecode.

On the other hand if they do upgrade before building, they'll have a compilation error, which is kinda nice as it points out a trivial fix for them to apply.

So I think we're good, but thanks a lot for checking!

hibernate-core/src/main/java/org/hibernate/event/internal/EntityState.java
Sanne: @NathanQingyangXu I agree with you :) but since Gavin is very focused on RX progress, I think we'll have to take this as is for now to not distract him, we can cleanup details later.

Sanne:
@gavinking merged as https://hibernate.atlassian.net/browse/HHH-13823

jwgmeligmeyling:
> On the other hand if they do upgrade before building, they'll have a compilation error, which is kinda nice as it points out a trivial fix for them to apply.

Yes, unless the extension comes via a dependency. But I consider this acceptible as well. The very few rare extensions that are affected are easily recompiled with the new visibility.

gavinking:
Thanks, @Sanne!

sebersole:
These all looked great.  We discussed most of them already.  Just getting back from traveling

PR #3177, opened 2020-01-19
HHH-12806 Eliminate all Javadoc warnings

NathanQingyangXu:
This PR aims to eliminate the remaining 46 Javadoc warnings left from last similar endeavour.
The following ticket was borrowed for the purpose:

https://hibernate.atlassian.net/browse/HHH-12806

The procedure to check the Javadoc from codebase is:

1. `./gradlew :hibernate-core:javadoc`;
2. open `hibernate-core/target/docs/javadoc/index.html` in browser.

Usually Javadoc warnings are because of some misuse of Javadoc and thus detrimental to the quality of produced Javadoc. So eliminating all warnings (we have no way to disable some single warning, do we?) is worthwhile. Especially new warning won't be buried among peers.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/SessionFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/SessionFactory.java
@@ -70,7 +70,7 @@ public interface SessionFactory extends EntityManagerFactory, HibernateEntityMan
 	 * for use.
 	 * <p/>
 	 * Note that for backwards compatibility, if a {@link org.hibernate.context.spi.CurrentSessionContext}
-	 * is not configured but JTA is configured this will default to the {@link org.hibernate.context.internal.JTASessionContext}
+	 * is not configured but JTA is configured this will default to the {@code org.hibernate.context.internal.JTASessionContext}
 	 * impl.
 	 *
 	 * @return The current session.
--- a/hibernate-core/src/main/java/org/hibernate/annotations/CacheConcurrencyStrategy.java
+++ b/hibernate-core/src/main/java/org/hibernate/annotations/CacheConcurrencyStrategy.java
@@ -21,25 +21,25 @@ public enum CacheConcurrencyStrategy {
 	/**
 	 * Indicates that read-only strategy should be applied.
 	 *
-	 * @see AccessType#READ_ONLY
+	 * @see org.hibernate.cache.spi.access.AccessType#READ_ONLY
 	 */
 	READ_ONLY( AccessType.READ_ONLY ),
 	/**
 	 * Indicates that the non-strict read-write strategy should be applied.
 	 *
-	 * @see AccessType#NONSTRICT_READ_WRITE
+	 * @see org.hibernate.cache.spi.access.AccessType#NONSTRICT_READ_WRITE
 	 */
 	NONSTRICT_READ_WRITE( AccessType.NONSTRICT_READ_WRITE ),
 	/**
 	 * Indicates that the read-write strategy should be applied.
 	 *
-	 * @see AccessType#READ_WRITE
+	 * @see org.hibernate.cache.spi.access.AccessType#READ_WRITE
 	 */
 	READ_WRITE( AccessType.READ_WRITE ),
 	/**
 	 * Indicates that the transaction strategy should be applied.
 	 *
-	 * @see AccessType#TRANSACTIONAL
+	 * @see org.hibernate.cache.spi.access.AccessType#TRANSACTIONAL
 	 */
 	TRANSACTIONAL( AccessType.TRANSACTIONAL );
 
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/naming/ObjectNameNormalizer.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/naming/ObjectNameNormalizer.java
@@ -51,7 +51,7 @@ public abstract class ObjectNameNormalizer {
 
 	/**
 	 * Normalizes the quoting of identifiers.  This form returns a String rather than an Identifier
-	 * to better work with the legacy code in {@link org.hibernate.mapping}
+	 * to better work with the legacy code in {@code org.hibernate.mapping} package.
 	 *
 	 * @param identifierText The identifier to be quoting-normalized.
 	 * @return The identifier accounting for any quoting that need be applied.
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/source/spi/PluralAttributeMapKeySource.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/source/spi/PluralAttributeMapKeySource.java
@@ -8,18 +8,14 @@ package org.hibernate.boot.model.source.spi;
 
 /**
  * Describes source information about the key of a persistent map.  At high
- * level this broken down further into 2 categories:<ul>
- *     <li>{@link PluralAttributeMapKeySourceEntityAttribute}</li>
- *     <li>
- *         <ul>
- *             <li>{@link PluralAttributeMapKeySourceBasic}</li>
- *             <li>{@link PluralAttributeMapKeySourceEmbedded}</li>
- *             <li>{@link PluralAttributeMapKeyManyToManySource}</li>
- *         </ul>
- *     </li>
+ * level this broken down further into 3 categories:<ul>
+ * <ul>
+ *  <li>{@link PluralAttributeMapKeySourceBasic}</li>
+ *  <li>{@link PluralAttributeMapKeySourceEmbedded}</li>
+ *  <li>{@link PluralAttributeMapKeyManyToManySource}</li>
  * </ul>
  * <p/>
- * {@link PluralAttributeMapKeySourceEntityAttribute} is only relevant from
+ * {@link PluralAttributeMapKeySource} is only relevant from
  * annotations when using {@link javax.persistence.MapKey}.
  *
  * @author Steve Ebersole
@@ -39,7 +35,7 @@ public interface PluralAttributeMapKeySource extends PluralAttributeIndexSource 
 	 * (relevant only for one-to-many and many-to-many associations)?
 	 *
 	 * If this method returns {@code true}, then this object can safely
-	 * be cast to {@link PluralAttributeMapKeySourceEntityAttribute}.
+	 * be cast to {@link PluralAttributeMapKeyManyToManySource}.
 	 *
 	 * @return true, if this plural attribute index source for an attribute of the referenced
 	 * entity; false, otherwise.
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/package.html
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/package.html
@@ -28,9 +28,9 @@
 			</ol>
 		</p>
 		<p>
-			Note that for field-level interception, simply plugging in a new {@link BytecodeProvider}
+			Note that for field-level interception, simply plugging in a new {@link org.hibernate.bytecode.spi.BytecodeProvider}
 			is not enough for Hibernate to be able to recognize new providers.  You would additionally
-			need to make appropriate code changes to the {@link org.hibernate.intercept.Helper}
+			need to make appropriate code changes to the {@link org.hibernate.bytecode.spi.ClassLoadingStrategyHelper}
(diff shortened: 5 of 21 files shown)


hibernate-core/src/main/java/org/hibernate/SessionFactory.java
```diff
@@ -70,7 +70,7 @@
 	 * for use.
 	 * <p/>
 	 * Note that for backwards compatibility, if a {@link org.hibernate.context.spi.CurrentSessionContext}
-	 * is not configured but JTA is configured this will default to the {@link org.hibernate.context.internal.JTASessionContext}
+	 * is not configured but JTA is configured this will default to the {@code org.hibernate.context.internal.JTASessionContext}
```
NathanQingyangXu: `internal` is excluded from javadoc, so I replace `@link` with `@code`. Let me know if you have better idea.

hibernate-core/src/main/java/org/hibernate/annotations/CacheConcurrencyStrategy.java
```diff
@@ -21,25 +21,25 @@
...
 	 *
-	 * @see AccessType#READ_WRITE
+	 * @see org.hibernate.cache.spi.access.AccessType#READ_WRITE
 	 */
 	READ_WRITE( AccessType.READ_WRITE ),
 	/**
 	 * Indicates that the transaction strategy should be applied.
 	 *
-	 * @see AccessType#TRANSACTIONAL
+	 * @see org.hibernate.cache.spi.access.AccessType#TRANSACTIONAL
```
NathanQingyangXu: Javadoc link targeting enum entry has to provide full package name for the enum class to avoid warning.

hibernate-core/src/main/java/org/hibernate/boot/model/naming/ObjectNameNormalizer.java
```diff
@@ -51,7 +51,7 @@ public Identifier normalizeIdentifierQuoting(Identifier identifier) {
 
 	/**
 	 * Normalizes the quoting of identifiers.  This form returns a String rather than an Identifier
-	 * to better work with the legacy code in {@link org.hibernate.mapping}
+	 * to better work with the legacy code in {@code org.hibernate.mapping} package.
```
NathanQingyangXu: There is no possibility to link to package in Javadoc, so I replace `@link` with `@code`. Again, let me know if you have better idea.

hibernate-core/src/main/java/org/hibernate/boot/model/source/spi/PluralAttributeMapKeySource.java
```diff
@@ -39,7 +35,7 @@
 	 * (relevant only for one-to-many and many-to-many associations)?
 	 *
 	 * If this method returns {@code true}, then this object can safely
-	 * be cast to {@link PluralAttributeMapKeySourceEntityAttribute}.
+	 * be cast to {@link PluralAttributeMapKeyManyToManySource}.
```
NathanQingyangXu: This class is one of the most difficult ones to resolve the Javadoc warnings for semantic understanding is required. Let me know whether my understanding is correct or not. I would not be surprised if my understanding is wrong.

hibernate-core/src/main/java/org/hibernate/bytecode/package.html
```diff
@@ -28,9 +28,9 @@
...
 		</p>
 		<p>
-			Note that for field-level interception, simply plugging in a new {@link BytecodeProvider}
+			Note that for field-level interception, simply plugging in a new {@link org.hibernate.bytecode.spi.BytecodeProvider}
 			is not enough for Hibernate to be able to recognize new providers.  You would additionally
-			need to make appropriate code changes to the {@link org.hibernate.intercept.Helper}
+			need to make appropriate code changes to the {@link org.hibernate.bytecode.spi.ClassLoadingStrategyHelper}
 			class.  This is because the detection of these enhanced classes is needed in a static
 			environment (i.e. outside the scope of any {@link org.hibernate.SessionFactory}.
 		</p>
```
NathanQingyangXu: Again, difficult one for semantic understanding is required. Shed light on this, thanks.

hibernate-core/src/main/java/org/hibernate/cache/spi/access/package.html
```diff
@@ -33,10 +33,10 @@
...
 <p>
     Note that, for an <i>asynchronous</i> cache, cache invalidation must be a two step process (lock->unlock or
-    lock->afterUpdate), since this is the only way to guarantee consistency with the database for a nontransactional
+    lock->afterUpdate), since this is the only way to guarantee consistency with the database for a non-transactional
     cache implementation. For a <i>synchronous</i> cache, cache invalidation is a single step process (evict or update).
-    Hence, these contracts ({@link org.hibernate.cache.spi.access.EntityRegionAccessStrategy} and
-    {@link org.hibernate.cache.spi.access.CollectionRegionAccessStrategy}) define a three step process to cater for both
+    Hence, these contracts ({@link org.hibernate.cache.spi.access.EntityDataAccess} and
+    {@link org.hibernate.cache.spi.access.CollectionDataAccess}) define a three step process to cater for both
     models (see the individual contracts for details).
```
NathanQingyangXu: Needs input. But it seems my understanding is not wrong.

hibernate-core/src/main/java/org/hibernate/engine/spi/SessionImplementor.java
```diff
@@ -36,7 +36,7 @@
  *         {@link org.hibernate.resource.jdbc.spi.JdbcSessionContext} delegate
  *     </li>
  *     <li>
- *         {@link TransactionCoordinatorBuilder.Options}
+ *         {@code Options} interface in {@link TransactionCoordinatorBuilder}
```
NathanQingyangXu: There is no possibility to link to inner interface within parent interface in current Javadoc, thus the change.

hibernate-core/src/main/java/org/hibernate/jpa/boot/spi/ProviderChecker.java
```diff
@@ -66,7 +66,7 @@ public static boolean hibernateProviderNamesContain(String requestedProviderName
 
 	/**
 	 * Extract the requested persistence provider name using the algorithm Hibernate uses.  Namely, a provider named
-	 * in the 'integration' map (under the key '{@value AvailableSettings#JPA_PERSISTENCE_PROVIDER}') is preferred, as per-spec, over
+	 * in the 'integration' map (under the key '<b>javax.persistence.provider</b>') is preferred, as per-spec, over
```
NathanQingyangXu: No possibility to refer to interface string static field. Changed it to explicit string. Guess the key should be stable, though not perfect.

jwgmeligmeyling: `@value` should have worked. What do you use for javadoc generation?

hibernate-core/src/main/java/org/hibernate/tool/schema/SourceType.java
```diff
@@ -26,8 +26,8 @@
 	/**
 	 * "script" - External DDL script(s) are used as the exclusive source for generation.  The scripts for schema
 	 * creation and dropping come from different sources.  The creation DDL script is identified by the
-	 * {@value AvailableSettings#HBM2DDL_CREATE_SCRIPT_SOURCE} setting; the drop DDL script is identified by the
```
jwgmeligmeyling: Again, `@value` should have worked here.

hibernate-core/src/main/java/org/hibernate/SessionFactory.java
jwgmeligmeyling: Seems fine. Although we might want to ignore this warning, because one can decide to generate javadoc including the internal package I believe.

hibernate-core/src/main/java/org/hibernate/bytecode/package.html
jwgmeligmeyling: Seems fine

hibernate-core/src/main/java/org/hibernate/cache/spi/access/package.html
jwgmeligmeyling: Seems fine.

hibernate-core/src/main/java/org/hibernate/boot/model/naming/ObjectNameNormalizer.java
jwgmeligmeyling: I think its possible when you add a label to it: `{@link org.hibernate.mapping mapping}

hibernate-core/src/main/java/org/hibernate/SessionFactory.java
NathanQingyangXu: Yeah, developers like you and me knows how to generate, but the main concern is the public user who looks up in publicly hosted Javadocs to look up some tech details.

hibernate-core/src/main/java/org/hibernate/boot/model/naming/ObjectNameNormalizer.java
NathanQingyangXu: I tried it again to no avail. The testing is via command line `./gradlew :hibernate-core:javadoc`. It shows up as warning. However, in Intellij IDE most of the javadoc issues in this PR won't be an issue.

hibernate-core/src/main/java/org/hibernate/jpa/boot/spi/ProviderChecker.java
NathanQingyangXu:
```
./gradlew :hibernate-core:javadoc
```

hibernate-core/src/main/java/org/hibernate/tool/schema/SourceType.java
NathanQingyangXu: Which command did you use, btw? It would be too good to have it work.

hibernate-core/src/main/java/org/hibernate/SessionFactory.java
jwgmeligmeyling: But then it already falls back to `@code`, it just generates the warning right?

NathanQingyangXu: Yeah, I checked the public Javadoc and found it simply display the link text, so you are right.
The gist of this PR is to eliminate all possible Javadoc warning (most of them are because of misuse of Javadoc and therefore are detrimental to users). If it were possible, I would disable this warning for it seems this warning is not harmful like others. But it seems no such possibility exists?

hibernate-core/src/main/java/org/hibernate/tool/schema/SourceType.java
sebersole: IMO we should report this to the Gradle team and see what they say.  `{@value}` is a perfectly legal inline doclet - https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#value.

I'd revert the changes related to this and let's wait to hear back from them.

NathanQingyangXu: I finally solved the issue by adding package to the class in `@value` reference. Not sure why but it seems there is some subtle difference between `@value` and `@see`.

NathanQingyangXu: > I finally solved the issue by adding package to the class in `@value` reference. Not sure why but it seems there is some subtle difference between `@value` and `@see`. Also, 
https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#value dictates the presence of package before class name as well.

hibernate-core/src/main/java/org/hibernate/SessionFactory.java
sebersole: Yeah, I'm torn on this one

NathanQingyangXu: It seems the only way to disable specific Javadoc warning is `@SuppressWarnings("javadoc")`, but it will hide any potential warnings.

PR #3178, opened 2020-01-19
HHH-13780 - add QueryHints.PASS_DISTINCT_THROUGH to NamedQuery

MarekChodak:
This is a fix for issue: 
https://hibernate.atlassian.net/browse/HHH-13780
QueryHints.PASS_DISTINCT_THROUGH doesn't work when specified in @NamedQuery hints

This is my first PR here, sorry for any mistakes

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/annotations/QueryBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/annotations/QueryBinder.java
@@ -67,6 +67,7 @@ public abstract class QueryBinder {
 				.setReadOnly( hints.getBoolean( queryName, QueryHints.READ_ONLY ) )
 				.setComment( hints.getString( queryName, QueryHints.COMMENT ) )
 				.setParameterTypes( null )
+				.setPassDistinctThrough(hints.getPassDistinctThrough(queryName, QueryHints.PASS_DISTINCT_THROUGH))
 				.createNamedQueryDefinition();
 
 		if ( isDefault ) {
@@ -108,8 +109,9 @@ public abstract class QueryBinder {
 				.setReadOnly( hints.getBoolean( queryName, QueryHints.READ_ONLY ) )
 				.setComment( hints.getString( queryName, QueryHints.COMMENT ) )
 				.setParameterTypes( null )
-				.setCallable( hints.getBoolean( queryName, QueryHints.CALLABLE ) );
-		
+				.setCallable( hints.getBoolean( queryName, QueryHints.CALLABLE ) )
+				.setPassDistinctThrough(hints.getPassDistinctThrough(queryName, QueryHints.PASS_DISTINCT_THROUGH));
+
 		if ( !BinderHelper.isEmptyAnnotationValue( resultSetMapping ) ) {
 			//sql result set usage
 			builder.setResultSetRef( resultSetMapping )
--- a/hibernate-core/src/main/java/org/hibernate/cfg/annotations/QueryBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/annotations/QueryBinder.java
@@ -67,7 +67,7 @@ public abstract class QueryBinder {
 				.setReadOnly( hints.getBoolean( queryName, QueryHints.READ_ONLY ) )
 				.setComment( hints.getString( queryName, QueryHints.COMMENT ) )
 				.setParameterTypes( null )
-				.setPassDistinctThrough(hints.getPassDistinctThrough(queryName, QueryHints.PASS_DISTINCT_THROUGH))
+				.setPassDistinctThrough( hints.getPassDistinctThrough( queryName ) )
 				.createNamedQueryDefinition();
 
 		if ( isDefault ) {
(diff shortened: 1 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/annotations/QueryBinder.java
```diff
@@ -67,6 +67,7 @@ public static void bindQuery(
 				.setReadOnly( hints.getBoolean( queryName, QueryHints.READ_ONLY ) )
 				.setComment( hints.getString( queryName, QueryHints.COMMENT ) )
 				.setParameterTypes( null )
+				.setPassDistinctThrough(hints.getPassDistinctThrough(queryName, QueryHints.PASS_DISTINCT_THROUGH))
```
NathanQingyangXu: Hibernate code format has the convention to use space char delimiters aggressively( two exceptions I know are: 1)method declaration; 2) method invocation with empty parameter). So the above statement should be changed as:
```
.setPassDistinctThrough( hints.getPassDistinctThrough( queryName, QueryHints.PASS_DISTINCT_THROUGH ) )
```

hibernate-core/src/main/java/org/hibernate/cfg/annotations/QueryHintDefinition.java
```diff
@@ -84,6 +84,14 @@ public LockMode getLockMode(String query) {
...
 	}
 
+	public Boolean getPassDistinctThrough(String query, String hintName) {
+		String value =(String)  hintsMap.get( hintName );
+		if ( value == null ) {
+			return null;
+		}
+		return getBoolean(query, hintName);
+	}
+
```
NathanQingyangXu: Seems the only additional value of the above method is the possibility to return `null`, compared with the following one. Seems the method is a generic one (not restricted to 'passDistinctThrough' for the hintName has been one parameter, not hardcoded).

hibernate-core/src/main/java/org/hibernate/engine/spi/NamedQueryDefinition.java
```diff
@@ -133,7 +134,8 @@ public NamedQueryDefinition(
 				comment,
 				parameterTypes,
 				null,		// firstResult
-				null		// maxResults
+				null,		// maxResults,
+				null 		//passDistinctThrough
```
NathanQingyangXu: seems we don't need comma for `// maxResults` and we need a space char for `//passDistinctThrough`.

hibernate-core/src/main/java/org/hibernate/cfg/annotations/QueryHintDefinition.java
NathanQingyangXu: Invoking `boolean getBoolean()` brings about code duplication. There might be other more elegant way to achieve the same goal.

hibernate-core/src/main/java/org/hibernate/cfg/annotations/QueryBinder.java
MarekChodak: Thanks, updated my code style and I think I fixed all the places

hibernate-core/src/main/java/org/hibernate/cfg/annotations/QueryHintDefinition.java
MarekChodak: Thanks for the comment. passDistinctThrough method indeed looked like generic, refactor mistake. I refactored the other part to with Optionals, did not see any use of them though in the project. Is it forbidden to use them ? If so, could you explain why ?

hibernate-core/src/main/java/org/hibernate/engine/spi/NamedQueryDefinition.java
MarekChodak: fixed :)

hibernate-core/src/main/java/org/hibernate/cfg/annotations/QueryHintDefinition.java
NathanQingyangXu: Yeah, I have no idea why `Optional` is seldom used. 
Guess you can ask at https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev?

As long as you use `Optional` in private method, it seems safe to use it as you have done.  Maybe the method name is not consistent with existing stuff and even a little bit confusing. Maybe `doGetBoolean`?

NathanQingyangXu: I asked the question for you there. BTW, it is always a good idea to use zulip for that is the IM tool they used.

MarekChodak: ok thanks a lot, renamed the method

NathanQingyangXu: Thanks. BTW, I am a contributor just like you, not team member. Given this PR is not in the PR list of the next release, guess it will take a while for it to be reviewed by team member and merged. Just FYI.

Sanne:
Excellent patch. Many thanks @MarekChodak and @NathanQingyangXu !

I'll combine the three commits and apply.

Sanne:
this was merged manually, github didn't detect it: closing. Thanks again!

PR #3179, opened 2020-01-21
HHH-13811 - Test and fix for H2 boolean literal issue

beikov:
https://hibernate.atlassian.net/browse/HHH-13811

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
@@ -424,6 +424,11 @@ public class H2Dialect extends Dialect {
 	}
 
 	@Override
+	public String toBooleanValueString(boolean bool) {
+		return bool ? "true" : "false";
+	}
+
+	@Override
 	public boolean doesReadCommittedCauseWritersToBlockReaders() {
 		// see http://groups.google.com/group/h2-database/browse_thread/thread/562d8a49e2dabe99?hl=en
 		return true;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
```diff
@@ -423,6 +423,11 @@ public boolean requiresParensForTupleDistinctCounts() {
 		return true;
 	}
 
+	@Override
+	public String toBooleanValueString(boolean bool) {
+		return bool ? "true" : "false";
```
NathanQingyangXu: So I checked other dialect class implementing this method. It seems the following pattern is followed invariably:
```
return String.valueOf( bool );
```
so maybe we might do that as well?

beikov: I actually copied this from the PostgreSQL81Dialect(https://github.com/hibernate/hibernate-orm/blob/e896db2ce2cdcaf2ed08e7333343b79cc228dc59/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java#L410), so it's not consistent in the code base. I don't see any advantage of using one over the other to be honest.

NathanQingyangXu: okay, if so, that is fine.

dreab8:
Hi @beikov ,

it seems that the test does not fail even when I remove the `H2Dialect` impl of the `toBooleanValueString(boolean bool)` method.

jwgmeligmeyling:
@beikov perhaps it depends on the H2 version used? Weve seen before that older versions of H2 are pretty limited sometimes

beikov:
Forget the issue and the PR. The problem is that H2 simply doesn't support comparing a predicate like `EXISTS` with a _booleanish_ literal, so no matter what, you always have to use a case when expression.

PR #3181, opened 2020-01-23
HHH-8776 Fix 'fetch graph' implementation

NathanQingyangXu:
I found our 'fetch graph' implementation is almost perfect (e.g. load plan, SQL generation are perfectly in line with 'fetch graph' JPA spec), the only missing part is in TwoPhaseLoad. We have added 'fetch profile' overwriting (w.r.t meta model) logic there. Seems we are missing the 'fetch graph' overwriting counterpart there.

Testing code for both 'find' and 'query' are added.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
@@ -457,6 +457,7 @@ public final class TwoPhaseLoad {
 			AttributeNodeImplementor<Object> attributeNode = context.findAttributeNode( associationName );
 			if ( attributeNode != null ) {
 				if ( associationType.isCollectionType() ) {
+					// to do: deal with Map's key and value
 					session.setFetchGraphLoadContext( null );
 				}
 				else {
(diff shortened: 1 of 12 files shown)


sebersole:
I do not believe your comment is correct.  This is becoming difficult to discuss.  I'll just look over your PR and push a new branch with the corrections I believe are needed.

NathanQingyangXu:
> I do not believe your comment is correct. This is becoming difficult to discuss. I'll just look over your PR and push a new branch with the corrections I believe are needed.

Go ahead. I think I still have difficulty fully understand your proposal so it would be great if you work on top of my PR. As long as 'fetch graph' is implemented as JPA spec, I would be really proud I ever helped. Thanks.

hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
```diff
@@ -367,38 +376,62 @@ public static void afterInitialize(
...
+		if ( associationType.isCollectionType() || associationType.isAssociationType()  ) {
+
+			// check 'fetch graph' first; skip 'fetch profile' if 'fetch graph' takes effect
+			Boolean overridingEager = isEagerFetchGraph( session, associationName, associationType );
 
-			//This method is very hot, and private so let's piggy back on the fact that the caller already knows the debugging state.
-			if ( isDebugEnabled ) {
-				if ( overridingEager != null ) {
+			if ( overridingEager != null ) {
+				//This method is very hot, and private so let's piggy back on the fact that the caller already knows the debugging state.
```
sebersole: I don't understand this comment.  Did you mean to remove the debugf parameters maybe?

sebersole: Oh, nm.. you  did not add that

PR #3183, opened 2020-01-26
HHH-6686 Fix JQL exception in face of 'is (not) empty'

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-6686

This 'ancient' bug aroused my attention. However, the fixing seems straightforward.
Testing case attached.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/HqlParser.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/HqlParser.java
@@ -18,7 +18,6 @@ import java.util.Set;
 import org.hibernate.QueryException;
 import org.hibernate.hql.internal.antlr.HqlBaseParser;
 import org.hibernate.hql.internal.antlr.HqlTokenTypes;
-import org.hibernate.hql.internal.ast.util.ASTPrinter;
 import org.hibernate.hql.internal.ast.util.ASTUtil;
 import org.hibernate.hql.internal.ast.util.TokenPrinters;
 import org.hibernate.internal.CoreLogging;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/HqlParser.java
```diff
@@ -348,6 +347,10 @@ private AST processIsEmpty(AST node, boolean negated) {
 	private AST createSubquery(AST node) {
 		AST ast = ASTUtil.createParent( astFactory, RANGE, "RANGE", node );
 		ast = ASTUtil.createParent( astFactory, FROM, "from", ast );
+
+		AST alias = ASTUtil.createSibling( astFactory, ALIAS, "_", node );
+		ASTUtil.insertChild( ASTUtil.createSibling( astFactory, SELECT, "select", ast ), astFactory.create(IDENT, alias.getText() ) );
+
```
NathanQingyangXu: The root cause of the bug is for 'exists' sub-query, we didn't specify 'SELECT', though we specified 'FROM'. Adding 'SELECT' fixed the issue.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/HqlParser.java
```diff
@@ -512,11 +515,11 @@ private String toPathText(AST node) {
...
 	public Map<String, Set<String>> getTreatMap() {
-		return treatMap == null ? Collections.<String, Set<String>>emptyMap() : treatMap;
+		return treatMap == null ? Collections.emptyMap() : treatMap;
 	}
 
 	public static void panic() {
-		//overriden to avoid System.exit
+		//overridden to avoid System.exit
 		throw new QueryException( "Parser: panic" );
 	}
```
NathanQingyangXu: I tried to fix some cosmetic legacy issues along the way. Not sure it is a good idea though.

dreab8:
applied upstream.

Thanks @NathanQingyangXu

sebersole:
Nice!

Benschio:
Very nice, thank you very much!

PR #3185, opened 2020-01-28
HHH-12153 Fix customized simple type 'serializable' detection issue

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-12153

Simple fix. Just fix the logic loophole that we only check customized simple type per se without navigating its ancestor types.

### DIFF
--- a/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/annotation/MetaAttributeGenerationVisitor.java
+++ b/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/annotation/MetaAttributeGenerationVisitor.java
@@ -305,12 +305,17 @@ class BasicAttributeVisitor extends SimpleTypeVisitor6<Boolean, Element> {
 			if ( TypeUtils.containsAnnotation( element, Constants.EMBEDDABLE ) ) {
 				return Boolean.TRUE;
 			}
-			for ( TypeMirror mirror : typeElement.getInterfaces() ) {
-				TypeElement interfaceElement = (TypeElement) context.getTypeUtils().asElement( mirror );
-				if ( "java.io.Serializable".equals( interfaceElement.getQualifiedName().toString() ) ) {
-					return Boolean.TRUE;
+			
+			do {
+				for ( TypeMirror mirror : typeElement.getInterfaces() ) {
+					TypeElement interfaceElement = (TypeElement) context.getTypeUtils().asElement( mirror );
+					if ( "java.io.Serializable".equals( interfaceElement.getQualifiedName().toString() ) ) {
+						return Boolean.TRUE;
+					}
 				}
-			}
+				typeElement = TypeUtils.getSuperclassTypeElement( typeElement );
+			} 
+			while ( typeElement != null );
 		}
 		return Boolean.FALSE;
 	}
--- a/tooling/metamodel-generator/src/test/java/org/hibernate/jpamodelgen/test/usertype/jpa/AbstractSerializableSimpleType.java
+++ b/tooling/metamodel-generator/src/test/java/org/hibernate/jpamodelgen/test/usertype/jpa/AbstractSerializableSimpleType.java
@@ -1,0 +1,18 @@
+package org.hibernate.jpamodelgen.test.usertype.jpa;
+
+import java.io.Serializable;
+
+public class AbstractSerializableSimpleType implements Serializable {
+	
+	private Integer value;
+	
+	public AbstractSerializableSimpleType(Integer value) {
+		this.value = value;
+	}
+	
+	@Override
+	public String toString() {
+		return value == null ? "" : value.toString();
+	}
+	
+}
--- a/tooling/metamodel-generator/src/test/java/org/hibernate/jpamodelgen/test/usertype/jpa/AbstractSimpleType.java
+++ b/tooling/metamodel-generator/src/test/java/org/hibernate/jpamodelgen/test/usertype/jpa/AbstractSimpleType.java
@@ -1,0 +1,16 @@
+package org.hibernate.jpamodelgen.test.usertype.jpa;
+
+public abstract class AbstractSimpleType {
+
+	private Integer value;
+		
+	public AbstractSimpleType(Integer value) {
+		this.value = value;
+	}
+
+	@Override
+	public String toString() {
+		return value == null ? "" : value.toString();
+	}
+		
+}
--- a/tooling/metamodel-generator/src/test/java/org/hibernate/jpamodelgen/test/usertype/jpa/ConcreteSerializableSimpleType.java
+++ b/tooling/metamodel-generator/src/test/java/org/hibernate/jpamodelgen/test/usertype/jpa/ConcreteSerializableSimpleType.java
@@ -1,0 +1,9 @@
+package org.hibernate.jpamodelgen.test.usertype.jpa;
+
+public class ConcreteSerializableSimpleType extends AbstractSerializableSimpleType {
+	
+	public ConcreteSerializableSimpleType(Integer value) {
+		super( value );
+	}
+	
+}
--- a/tooling/metamodel-generator/src/test/java/org/hibernate/jpamodelgen/test/usertype/jpa/ConcreteSerializableSimpleTypeImplementingSerializable.java
+++ b/tooling/metamodel-generator/src/test/java/org/hibernate/jpamodelgen/test/usertype/jpa/ConcreteSerializableSimpleTypeImplementingSerializable.java
@@ -1,0 +1,10 @@
+package org.hibernate.jpamodelgen.test.usertype.jpa;
+
+import java.io.Serializable;
+
+public class ConcreteSerializableSimpleTypeImplementingSerializable extends ConcreteSerializableSimpleType implements Serializable {
+
+	public ConcreteSerializableSimpleTypeImplementingSerializable(Integer value) {
+		super( value );
+	}
+}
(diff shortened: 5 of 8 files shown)


Sanne:
Hi @NathanQingyangXu , I don't think this is correct: "Serializable" is a very special interface, it generally doesn't apply to sub-types: the recommendation is that each concrete type should explicitly advertise if it's Serializable (or not).

I'm not familiar with that code though; it might be interesting to check why this needs to be Serializable at all?  The need of that check might be out of date.

NathanQingyangXu:
> Hi @NathanQingyangXu , I don't think this is correct: "Serializable" is a very special interface, it generally doesn't apply to sub-types: the recommendation is that each concrete type should explicitly advertise if it's Serializable (or not).
> 
> I'm not familiar with that code though; it might be interesting to check why this needs to be Serializable at all? The need of that check might be out of date.

Thanks for the feedback. I guess the requirement comes from JPA spec as demonstrated in the screenshot:

Screen Shot 2020-01-28 at 11 16 49 AM

TBH I am not very familiar with the relevant part but from ticket description, it seems to make some sense.
However, I have no idea that `Serializable` needs to be specified explicitly even though its parent class has specified.

I'll follow up by more research, however. Thanks again.

jwgmeligmeyling:
From the `Serializable` Javadoc:

> All subtypes of a serializable class are themselves serializable. 

I am not sure what the convention is. Serializable in inheritance hierarchies is by itself discouraged (Effective Java).

I am interested whether we currently support custom simple types using the Serializable interface if its present on a super class. If we do, then I think the JPA Metamodel generator should be changed accordingly for consistency. I think we can verify that by adding a test case that uses queries this property for the added domain classes.

I really hate `Serializable` columns by the way 🤣 Only ever encountered them in accidental metamodel errors 😋

Sanne:
Right the Serializable javadoc doesn't match with the "best practice" in this case. Both are kinda raising important points: the first one stresses that subclasses have to be serializable according to the language rules as interfaces are inherited, but it's not recommended as it easily leads to mistakes.

I can't presume to rule who is right :)  But yes we should be consistent, that's why I commented about this not being _necessarily_ a correct fix.

Before making such a change, we should check at least:

 - do we need this check still at all? In ORM6 there's less emphasis on serialization across the board.
 - what do other components do?

My suggestion would be to avoid making such changes at all in ORM 5, they are likely going to have unintended consequences. I'd rather propose them as changes for the ORM 6 branch.

@sebersole is the lead authority on such subjects BTW: no need to ping @emmanuelbernard , I'll change the review request.

NathanQingyangXu:
Thanks. Let us listen to @sebersole 's opinion.

pasfilip:
IMO if a super class implements serializable any sub-class is serializable, and must respect serializable contract, and thus it’s redudant to declare serializable again in the sub-class.
I agree that Serializable is not as relevant as it used to be and you can make custom value types work perfectly without them having to be Serializable.
But you should respect the Serializable contract anyway.
It works to implement custom types that are not serializable but the model generator checks this explicitly. The check could be removed that a custom value type must be serializable.

NathanQingyangXu:
I don't think any class which is serializable (from ancestor class) needs to specify the marker interface. A good example is the `PersistentMap` in Hibernate which didn't specify it.

Screen Shot 2020-01-29 at 3 16 56 PM

Also, here is the snippet from the Javadoc of `Serializable`:
Screen Shot 2020-01-29 at 3 18 25 PM

These are literally my two cents, :).

pasfilip:
I vote to remove the need of a value type to be Serializable in order to be included in the generated meta model. 
Most users will be fine with this imo. 
I can't think of any reason why to keep the requirement that in order for a field to be included in the meta model it must implement Serializable. 
Perhaps distributed caches needs this in order to be able to serialize the entity state across nodes? 
If your application relies on rmi you'll run into issues.
These use cases seem somewhat uncommon.
When you do run into problems because Serializable is missing you can still make your custom types implement Serializable unless you're using a third party type.

If the choice is made to keep the requirement as then the meta model should fail when it detects a value type that isn't Serializable.
You could also argue that hibernate core should throw or at least warn at startup that an entity has a field that isn't Serializable since technically the spec seems to require it.

NathanQingyangXu:
TBH, the spec simply says 'implements Serializable' without dictating how to achieve that, IMHO.

jwgmeligmeyling:
@NathanQingyangXu PersistentMap is not a really a good example, as it doesn't use its serialized form for persistence (as it is a PersistentCollection and thus not a simple type). Its probably just serializable because many other Java collections are, and its necessary for i.e. entity caching and passivation of the entity manager in extended contexts. Long story.

That said, I agree that it probably makes more sense to accept this PR as-is , than to change the current behaviour of handling Serializable simple types in Hibernate (that users apparently already rely on).

NathanQingyangXu:
Thanks. I mentioned `PersistentMap` as an example that many class implementing `Serializable` doesn't necessarily specify it if its ancestor class has specified it. I didn't mean it is a simple type. JPA Spec is not Java code, so when it says 'implements Serializable`, it doesn't necessarily means `class xxxx implements Serializable`.

Again, let us listen to what @sebersole will say.

Interesting discussion, though.

jwgmeligmeyling:
A non-primitive and non-String type can only be persisted as simple type in its serialized form, otherwise it requires a custom type or a type transformer.

We could throw a mapping error if a simple type is tried to be constructed for a serializable type, albeit not one that implements it directly.

However, this will prevent anyone from storing serializable entities that come from dependencies and other sources of unmanaged code.

I think there is no harm in applying Serializable transitively here even though its not very conventional.

Sanne:
good points all.

To be clear I didn't mean to express a strong opinion, just meant to point out that such changes aren't as straight forward as it might have looked at first glance - there are various implications to consider, and that makes it a better candidate for 6 than including such a change in a micro release.

Please don't underestimate the impact such changes might have, even if it could be considered a fix :)

sebersole:
I think that technically the PR is right.  

I think the better solution would be to render any attributes that we have deemed to be persistent regardless of the type.  But there are a lot of hurdles to implementing that properly

sebersole:
I think merging this on to master is ok - it *is* a bug imo.  But I would not back port it any further

gavinking:
I think the best way to conceptualize `Serializable` is that it’s an annotation, created before Java had support for annotations. Now, an annotation can be `@Inherited` or not, but if it is, then there should sorta be some mechanism to enforce that this isn’t just a lie and that the semantics of the annotation are truly inherited, and that if they impose a constraint upon the subtypes, then that this constraint is actually enforced by someone.

(For a real interface, which `Serializable` is not, the mechanism for enforcing the constraint is that the subtypes have to implement the abstract operations.)


And that’s the problem with this “annotation”: it imposes constraints on its subtypes that it has no way at all to enforce. And this explains why Sanne thinks that all serializable types should redeclare their serializability, even if that’s contrary to the Javadoc.

So anyway my point is there’s probably not a right answer here.

🤷‍♂️

NathanQingyangXu:
I did some more research on this topic, especially after @gavinking shared his wonderful thoughts on this topic.
I think there is a special Java interface group called 'marker interface', including: 1)Cloneable; 2)Serializiable; 3)Remote (https://www.geeksforgeeks.org/marker-interface-java/)
I think @Sanne 's original comment applies to `Cloneable`, but not on the other two.
The Javadoc on `Remote` seems in line with `Serializable` as well, unless I misunderstood:
Screen Shot 2020-02-01 at 12 22 13 PM

Actually, `Serializable` and `Remote` marker interfaces share many things in common. As long as Javadoc specified, we should support the feature for `instanceof Serializable` returns true for class object inheriting the interface from its ancestor class.

Obviously this PR is technically correct.

NathanQingyangXu:
As always the case, whenever I dig deeper, I got something new that surprised me. It seems all marker interfaces take effect directly or indirectly, including `Cloneable`(https://programming.guide/java/clone-and-cloneable.html):
Screen Shot 2020-02-02 at 10 04 51 AM

I was also surprised by the following failing testing case:
```
@Test
public void test () {
    class ParentCloneable implements Cloneable {}
    class Child extends ParentCloneable {}
    assert.assertFalse( new Child() instanceof Cloneable );
}
```

pasfilip:
@NathanQingyangXu Serializable, Cloneable may be classified as "Marker" interfaces they are still regular interfaces.
Any interface you implement comes with a contract that must be respected.
If you extend a Collection class, the sub-class still must respect the contract and behave appropriately.
I assume you wouldn't be surprised that the subclass, of a List for example, is still an instance of a Collection.
Same goes with any marker interface.
This ties in with the Liskov substituion principle.
For me the fix is definately correct. 
I think the debate should be if a value type that doesn't implement Serializable, as expected by the spec, unless it's marked as a transient field or annotated with @Transient, should still be included in the generated meta model.

NathanQingyangXu:
> @NathanQingyangXu Serializable, Cloneable may be classified as "Marker" interfaces they are still regular interfaces.
> Any interface you implement comes with a contract that must be respected.
> If you extend a Collection class, the sub-class still must respect the contract and behave appropriately.
> I assume you wouldn't be surprised that the subclass, of a List for example, is still an instance of a Collection.
> Same goes with any marker interface.
> This ties in with the Liskov substituion principle.
> For me the fix is definately correct.
> I think the debate should be if a value type that doesn't implement Serializable, as expected by the spec, unless it's marked as a transient field or annotated with @transient, should still be included in the generated meta model.

yeah, I totally agree with you. But the spec doesn't dictate we need the fix, right? For it requires `implements Serializable`, it doesn't exclude 'indirectly implements`, right?

pasfilip:
@NathanQingyangXu Yes I think you're right. From the perspective of the spec it says implements Serializable interface. No matter if it's declared in a super class or re-declared in a sub-class.
From a language perspective it only matters if instanceof Serialzable is true thus including sub-classes of a class that declares Serializable.

NathanQingyangXu:
Yeah, so it seems there is no necessity to discuss further, unless there is additional concern.

PR #3186, opened 2020-01-28
HHH-13783: fix test failure in HiLoSequenceMismatchStrategyTest

breglerj:


dreab8:
applied upstream.
Thanks @breglerj

PR #3187, opened 2020-01-28
Register JPA embeddable types properly, allow access to functions in registry and fix AttributeClassifications

beikov:


sebersole:
Thanks!  Especially helping with my silly mistakes :smile:

PR #3188, opened 2020-01-29
Replace OpCodes.ASM5 with OpCodes.ASM7

gwenneg:
This might help fixing https://github.com/quarkusio/quarkus/issues/6800.

I'm submitting this as a draft for now since I'm not familiar with the hibernate-orm build or its CI.

cc @gsmet

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/BiDirectionalAssociationHandler.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/BiDirectionalAssociationHandler.java
@@ -296,7 +296,7 @@ final class BiDirectionalAssociationHandler implements Implementation {
 		@Override
 		public Size apply(
 				MethodVisitor methodVisitor, Context implementationContext, MethodDescription instrumentedMethod) {
-			return delegate.apply( new MethodVisitor( Opcodes.ASM5, methodVisitor ) {
+			return delegate.apply( new MethodVisitor( Opcodes.ASM7, methodVisitor ) {
 
 				@Override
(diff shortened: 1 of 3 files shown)


Sanne:
thanks @gwenneg !

Sanne:
This looks correct to me and passes all tests, but I'm not entirely sure of why a specific ASM version is there.

I wonder if I shouldn't have updated all of these while I've been keeping Byte Buddy updated.

hi @raphw , would you mind to comment please?

Would it have sense to have a "OpCodes.ASM_LATEST_SUPPORTED" or something similar since we'll (normally) target the ASM version that Byte Buddy is including?

gsmet:
So I'm +1 for such a chance but I agree it's a pain we don't have a constant in ASM pointing to the latest supported version.

We have a lot of these versions in Quarkus. I'm thinking of having a constant in Gizmo for this. But that won't solve the Hibernate ORM issue.

Sanne:
tracked by: https://hibernate.atlassian.net/browse/HHH-13833

Sanne:
I've done some more tests, then (manually) rebased it and merged. Thanks all again!

Still wondering what @raphw would think of such use of constants, but we can defer such improvements.

raphw:
Byte Buddy defines a constant in [`OpenedClassReader`](https://github.com/raphw/byte-buddy/blob/0c8bc15266125a5a8ad5bd54da69a68c0e49d660/byte-buddy-dep/src/main/java/net/bytebuddy/utility/OpenedClassReader.java#L43) which references the latest API version that Byte Buddy supports.

Sanne:
Many thanks @raphw ! I'll add a follow up to refer to that constant.

PR #3190, opened 2020-01-29
HHH-13831 Replaced listener not called when event is fired

DavideD:
When replacing a listener the array in `EventListenerGroupImpl`  is not refereshed.

Is this a bug or am I missing something?

I've created a JIRA: https://hibernate.atlassian.net/browse/HHH-13831

Thanks

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
@@ -222,6 +222,7 @@ class EventListenerGroupImpl<T> implements EventListenerGroup<T> {
 							checkAgainstBaseInterface( listener );
 							performInjections( listener );
 							itr.set( listener );
+							listeners = null; //Marks it for refreshing
 							doAdd = false;
 							break strategy_loop;
 						}
--- a/hibernate-core/src/test/java/org/hibernate/event/service/internal/EventListenerReplacementStrategyTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/event/service/internal/EventListenerReplacementStrategyTest.java
@@ -1,0 +1,99 @@
+package org.hibernate.event.service.internal;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.hibernate.event.service.spi.DuplicationStrategy;
+import org.hibernate.event.service.spi.EventListenerGroup;
+import org.hibernate.event.spi.ClearEvent;
+import org.hibernate.event.spi.ClearEventListener;
+import org.hibernate.event.spi.EventType;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test that a listener replacing the original one is actually called when the right event is fired.
+ * <p>
+ *     Note: I'm using ClearEvent for the tests because it the simpler one I've found.
+ * </p>
+ */
+public class EventListenerReplacementStrategyTest {
+
+	@Test
+	public void testListenerReplacement() {
+		Tracker tracker = new Tracker();
+		EventListenerRegistryImpl listenerRegistry = null;
+		EventListenerGroup<ClearEventListener> listenerGroup = new EventListenerGroupImpl( EventType.CLEAR, listenerRegistry );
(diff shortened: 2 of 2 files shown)


jwgmeligmeyling:
Probably an oversight / it just wasn't necessary before. As it is in the internal package, I wouldn't consider this a bug. Whats the use case? The unit test seems very specific.

Besides that, I am all for integrating this if it solves any issue for you.

DavideD:
The replaced listener is not called if the event is fired using `EventListenerGroup#fireEventOnEachListener`
We didn't notice before because most of the time `EventListenerGroup#listeners()` is called.

This happend because the listener is replaced in internal field `listenersAsList` but it stays out of sync with the other array field `listeners`. The patch mark the `listeners` field for refresh when we replace one of the listeners.

hibernate-core/src/test/java/org/hibernate/event/service/internal/EventListenerReplacementStrategyTest.java
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+
+	private static class ReplaceOriginalStrategy implements DuplicationStrategy {
+
+		static final ReplaceOriginalStrategy INSTANCE = new ReplaceOriginalStrategy();
+
+		@Override
+		public boolean areMatch(Object listener, Object original) {
+			return original.getClass().equals( OriginalListener.class )
```
sebersole: I'm a little nervous about this.  Why not just check for each being an `instanceof ClearEventListener`?  The intention of your test is clearly that there be only one.

I'd also consider asserting `org.hibernate.event.service.spi.EventListenerGroup#listeners` to make sure that the listener actually got replaced/appended.  Coupled with the point above, its possible that your code itself is not actually appending/replacing the listener

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
```diff
@@ -222,6 +222,7 @@ private boolean listenerShouldGetAdded(T listener) {
 							checkAgainstBaseInterface( listener );
 							performInjections( listener );
 							itr.set( listener );
+							listeners = null; //Marks it for refreshing
```
dreab8: not 100% sure but I think the correct thing to do is to remove the original listener from `listeners` .

DavideD: That does not seem the approach used in the class, though

DavideD:
I'm going to create an issue

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
dreab8: forget my comment, I saw how `listerner` it is used

DavideD:
By the way, the end effect is that some listeners cannot be replaced. For example: `DefaultInitializeCollectionEventListener`

sebersole:
The clarify my comment... the real problem in the code block you highlight is both fields are not updated.  Only `#listenersAsList` is updated.  But its `#listeners` that is iterated to handle the events.  So that code block gets the 2 out of sync

hibernate-core/src/test/java/org/hibernate/event/service/internal/EventListenerReplacementStrategyTest.java
DavideD: > I'm a little nervous about this. Why not just check for each being an `instanceof ClearEventListener`? The intention of your test is clearly that there be only one.

OK, I will change it

> I'd also consider asserting `org.hibernate.event.service.spi.EventListenerGroup#listeners` to make sure that the listener actually got replaced/appended. Coupled with the point above, its possible that your code itself is not actually appending/replacing the listener

OK,  will also test listeners()

DavideD:
> But its `#listeners` that is iterated to handle the events. So that code block gets the 2 out of sync

Yes, I was surprised we didn't notice this before though.
It seems that the code uses the iterator from `#listenerAsList` more often than the listeners array.

DavideD:
I will close this PR and have a look to see what happens if I only keep the array field.

PR #3192, opened 2020-01-29
Topic/listener test

sebersole:
@DavideD what do you think?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
@@ -7,14 +7,12 @@
 package org.hibernate.event.service.internal;
 
 import java.lang.reflect.Array;
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.ListIterator;
-import java.util.Map;
 import java.util.Set;
 import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 
 import org.hibernate.event.service.spi.DuplicationStrategy;
@@ -25,24 +23,21 @@ import org.hibernate.event.service.spi.JpaBootstrapSensitive;
 import org.hibernate.event.spi.EventType;
 import org.hibernate.jpa.event.spi.CallbackRegistryConsumer;
 
+import org.jboss.logging.Logger;
+
 /**
  * @author Steve Ebersole
  * @author Sanne Grinovero
  */
 class EventListenerGroupImpl<T> implements EventListenerGroup<T> {
-	private EventType<T> eventType;
+	private static final Logger log = Logger.getLogger( EventListenerGroupImpl.class );
+
+	private final EventType<T> eventType;
 	private final EventListenerRegistryImpl listenerRegistry;
 
 	private final Set<DuplicationStrategy> duplicationStrategies = new LinkedHashSet<>();
 
-	// Performance: make sure iteration on this type is efficient; in particular we do not want to allocate iterators,
-	// not having to capture state in lambdas.
-	// So we keep the listeners in both a List (for convenience) and in an array (for iteration). Make sure
-	// their content stays in synch!
 	private T[] listeners = null;
-
-	//Update both fields when making changes!
-	private List<T> listenersAsList;
 
 	public EventListenerGroupImpl(EventType<T> eventType, EventListenerRegistryImpl listenerRegistry) {
 		this.eventType = eventType;
@@ -82,11 +77,8 @@ class EventListenerGroupImpl<T> implements EventListenerGroup<T> {
 
 	@Override
 	public void clear() {
-		if ( duplicationStrategies != null ) {
-			duplicationStrategies.clear();
-		}
+		duplicationStrategies.clear();
 		listeners = null;
-		listenersAsList = null;
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


sebersole:
Sorry, pushed this a bit early.  Made a silly mistake

DavideD:
I've found another error in the test case I created. I didn't realise I was using two different trackers to keep track of the calls. This commit should fix the problem: https://github.com/DavideD/hibernate-orm/commit/e8a1625af2c99ebf7240eaea062a33cc536a5786

DavideD:
It looks good, but considering the amount of silly mistakes I made on the test, I think I will have a second look tomorrow morning :)

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
```diff
@@ -125,118 +120,141 @@ public void addDuplicationStrategy(DuplicationStrategy strategy) {
...
+							log.debugf( "Skipping listener registration (%s) : `%s`", strategy.getAction(), listener );
+							return;
 						}
 						case REPLACE_ORIGINAL: {
-							checkAgainstBaseInterface( listener );
-							performInjections( listener );
-							itr.set( listener );
-							doAdd = false;
-							break strategy_loop;
+							log.debugf( "Replacing listener registration (%s) : `%s` -> ", strategy.getAction(), existingListener, listener );
```
NathanQingyangXu: Is there a `%s` missing after ` -> ` above?

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
```diff
@@ -25,25 +23,22 @@
...
 
-	// Performance: make sure iteration on this type is efficient; in particular we do not want to allocate iterators,
-	// not having to capture state in lambdas.
-	// So we keep the listeners in both a List (for convenience) and in an array (for iteration). Make sure
-	// their content stays in synch!
 	private T[] listeners = null;
 
-	//Update both fields when making changes!
-	private List<T> listenersAsList;
-
```
NathanQingyangXu: I love the code change. The manual sync up is so difficult to maintain.

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
```diff
@@ -82,20 +77,18 @@ public int count() {
 
 	@Override
 	public void clear() {
-		if ( duplicationStrategies != null ) {
-			duplicationStrategies.clear();
-		}
+		duplicationStrategies.clear();
```
NathanQingyangXu: good catch!

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
```diff
@@ -125,118 +120,141 @@ public void addDuplicationStrategy(DuplicationStrategy strategy) {
...
+			//noinspection unchecked
+			this.listeners = (T[]) Array.newInstance( eventType.baseListenerInterface(), 1 );
+			this.listeners[0] = listener;
 		}
-	}
+		else {
+			final int size = this.listeners.length;
+
+			//noinspection unchecked
+			final T[] newCopy = (T[]) Array.newInstance( eventType.baseListenerInterface(), size+1 );
```
NathanQingyangXu: could we add space chars around `+` like `size + 1`?

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
```diff
@@ -125,118 +120,141 @@ public void addDuplicationStrategy(DuplicationStrategy strategy) {
...
+
+			// first copy the existing listeners
+			System.arraycopy( this.listeners, 0, newCopy, 0, size );
+
+			// and then put the new one after them
+			newCopy[size] = listener;
```
NathanQingyangXu: Could we utilize some util method in JDK as following
```
final T[] newCopy = java.util.Arrays.copyOf( this.listeners, size + 1 );
newCopy[size] = listener;
```
Seems less error-prone and easier to understand.

NathanQingyangXu: After second thought, I think your approach is more consistent with other code snippet (like inserting new element in front).

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
sebersole: Yes :(  @DavideD since you started a new branch/PR can you just fix this there...

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
sebersole: I prefer this style for simple arith

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
sebersole: Right, we could for the append case - not for the prepend case.  Comes down to a question of consistency

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerGroupImpl.java
DavideD: sure

DavideD:
I've created a new PR: https://github.com/hibernate/hibernate-orm/pull/3194

PR #3194, opened 2020-01-30
HHH-13831 Fix EventListenerGroup implementation + Tests for replacement strategies

DavideD:
Follows: https://github.com/hibernate/hibernate-orm/pull/3192
https://hibernate.atlassian.net/browse/HHH-13831

I've added some tests, fixed the existing ones and fix added the misisng placeholder in the log message.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/event/service/internal/EventListenerReplacementStrategyTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/event/service/internal/EventListenerReplacementStrategyTest.java
@@ -1,0 +1,148 @@
+package org.hibernate.event.service.internal;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.hibernate.event.service.spi.DuplicationStrategy;
+import org.hibernate.event.service.spi.EventListenerGroup;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/test/java/org/hibernate/event/service/internal/EventListenerReplacementStrategyTest.java
```diff
@@ -0,0 +1,298 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test that a listener replacing the original one is actually called when the event is fired for each listener.
+ * <p>
+ *     Note: I'm using ClearEvent for the tests because it's the simpler one I've found.
+ * </p>
+ */
+@TestForIssue(jiraKey = "HHH-13831")
+public class EventListenerReplacementStrategyTest {
```
sebersole: Can we rename this `EventListenerDuplicationStrategyTest`?  It now tests more than just the replace strategy

DavideD: Done!

Sanne:
nice! thanks and sorry for the bug :)

PR #3196, opened 2020-01-30
Enhancements to Dialects, Date/Time functionality, and HQL

gavinking:
This pull request contains most of my work which had been done against the older, now abandoned Hibernate 6 branch. The work consists of many closely-interrelated changes in the area of:

- SQL dialects,
- HQL functions, and
- dates and times.

A major goal of this work is to make HQL queries, DDL generation, and date/time handling much more portable between databases. A secondary goal was to clean up the `Dialect` hierarchy and make it easier to maintain in future.

## Changes to SQL dialects

The whole SQL `Dialect` hierarchy has been refactored, with the following goals:

### One `Dialect` class per database

Previously there was an ad-hoc approach where inheritance was used to define version-specific dialect variations. The disadvantages of this approach included:

- the user needed to be careful that they were using the right versioned dialect class, especially when the explicitly specified the dialect in their code, and needed to be aware of the different classes, which might not correspond exactly the database version they were using,
- since the cost of introducing a new `Dialect` class was relatively high, very often the version-specific variations were introduced in a `Dialect` class which did not exactly correspond to the database version which introduced the variation, and
- the actual inheritance hierarchy reflected the whole history of Hibernate development, including several deprecated classes with duplicated code, and inconsistent use of inheritance between products.

The new approach is to use a single `Dialect` class per product, and handle variations in conditional blocks of code which check the database version directly based on the JDBC metadata.

### Improvements to DDL generation

SQL datatypes in generated `CREATE TABLE` statements are improved in several ways:

- The Java field type mapped to a SQL column type can now specify a preferred length, precision, and scale via its `JavaTypeDescriptor`. In certain cases these preferred lengths are dialect-specific, and are delegated back to the `Dialect`. 
- Thus, we no longer generate DDL with wasteful column lengths, precision, or scales, for example, `BigInteger` previously mapped to a decimal type with non-zero scale.
- On most dialects, Java `double` and `float` now both map to SQL `float(p)`, usually to `float(53)` and `float(24)` respectively (though there are some databases which incorrectly interpret the precision of the ANSI `float` type).
- When a `numeric`, `decimal`, or `timestamp` type has no explicitly-specified precision, use a `Dialect`-defined default. (This makes DDL generation more portable across DBs, especially for `BigInteger` and `BigDecimal`.)
- The precision for a SQL `timestamp` type may now be specified using `@Column(precision=p)`.
- Java `Duration` now maps to `numeric(21,0)` instead of `bigint`.
- Java `enum`s now map to `tinyint`.
- Column-level check constraints are now automatically generated for Java `enum` mappings, and for Java `boolean`s mapped to `bit` types.
- The SQL `boolean` type is now used on more platforms, reflecting that fact that more databases have added support for it.
- Handling of the SQL `bit` and `char` types has been cleaned up. (It was previously assumed that `bit` and `char` types were always one bit long.)
- Use `timestamp with time zone` for `OffsetDateTime` and `ZonedDateTime`, where supported.
- Registrations for `longvarchar` and `longvarbinary` are removed since there are no such types in ANSI SQL. (Nor on most databases.)

### Complete clean up of SQL type mappings

The entire monstrosity of SQL type registrations has been fully audited for correctness against the documentation for the various databases, and retested where possible. Unfortunately I was unable to test every database we support. (In particular, I was unable to test on any dialect of Sybase.) 

The approach taken by the new `Dialect`s is that each dialect *only* overrides type names that differ from ANSI SQL, and *only* when the ANSI SQL synonym is unavailable. Thus, there are fewer dialect-specific type registrations.

For example:

- in Hibernate 5, `H2Dialect` overrides 22 SQL type names; `PostgreSQL81Dialect` overrides 20; `DB2Dialect` overrides 23, but
- after this patch, `H2Dialect` has no SQL type mappings; `PostgreSQLDialect` overrides 9; `DB2Dialect` overrides 10.

### Introduction of `SequenceSupport`

By analogy with `IdentityColumnSupport` and `LimitHandler`, support for sequences is now defined by the `SequenceSupport` interface and dialect-specific implementations in `org.hibernate.dialect.sequence`.

### Cleanup of `LimitHandler`s

The `LimitHandler` hierarchy has been significantly refactored. This section of the system has not been well-maintained and had lots of thing that weren’t quite right.

Notable changes:

- We use the ANSI SQL-standard `offset` / `fetch` syntax where possible now that more databases support it. (DB2, Postgres, H2, HSQL, Firebird, Mimer.)
- The order of limit causes with respect to the `for update` clause has been fixed on Postgres, H2, and DB2.
- There is a new `RowsLimitHandler` for Firebird, Interbase, and TimesTen.
- `SQLServer2005LimitHandler` had to be totally rewritten.

### Dialects for CockroachDB and Spanner

I have written a dialect for CockroachDB, which seems like a nice product. I’ve also integrated the dialect for Spanner which previously existed on the old Hibernate 6 branch.

I’ve also much improved the dialects for certain products which just happen to run nicely on my Mac: Mimer, Derby, and Firebird.

## Portable HQL functions

Previously, HQL supported the 20 JPQL-defined standard functions, to a somewhat flaky extent (for example, the date/time functions were not supported on SQL server), along with a handful of additional “portable” functions defined back in ancient times, to an even flakier extent. Until now, the lack of a really well-defined set of portable functions has been a hurdle for writing portable HQL.

Partly as a result of this, we make little real attempt to check the types of function arguments, and generally let these sort of problems be detected by the database. This makes writing HQL rather less pleasant than it should be. This current patch does not solve this issue, but it lays the foundation for fixing the problem which is my next task.

An eventual goal of this work is that we would like to see the JPA spec adopt at least some of these new functions into the standard.

### List of portable functions

This patch introduces the following set of functions, which, while not 100% portable across every single platform we have a `Dialect` for, is *much* more portable than anything we have had until now.

#### JPQL functions:
```
avg(arg)
count([distinct ]arg)
max(arg)
min(arg)
sum(arg)

coalesce(arg0, arg1, ...)
nullif(arg0, arg1)

lower(arg)
upper(arg)
length(arg)
concat(arg0, arg1, ...)
locate(pattern, string[, start])
substring(string, start[, length])
trim([[spec ][character ]from] string)

abs(arg)
mod(arg0, arg1)
sqrt(arg)

current date / current_date
current time / current_time
current timestamp / current_timestamp
```
#### ANSI SQL functions:
```
any(arg)
every(arg)

cast(arg as Type)
extract(field from arg)

ln(arg)
exp(arg)
power(arg0, arg1)
floor(arg)
ceiling(arg)

position(pattern in string)
substring(string from start[ for length])
overlay(string placing replacement from start[ for length])
```
#### HQL functions:

```
local date / local_date
local time / local_time
local datetime / local_datatime
offset datetime / offset_datetime
instant

left(string, length)
right(string, length)
replace(string, pattern, replacement)
pad(string with length spec[ character])

sign(arg)
sin(arg)
cos(arg)
tan(arg)
asin(arg)
acos(arg)
atan(arg)
atan2(arg0, arg1)
round(arg0, arg1)
least(arg0, arg1, ...)
greatest(arg0, arg1, ...)

format(datetime as pattern)
str(arg)
ifnull(arg0, arg1)

timestampadd(field, arg1, arg2)
timestampdiff(field, arg1, arg2)

second(arg)	
minute(arg)	
hour(arg)
day(arg)
month(arg)
year(arg)
```
### Setting `hibernate.dialect.show_hql_functions`

Setting the configuration property `hibernate.dialect.show_hql_functions` to `true` will force Hibernate to print out a list of all standard HQL functions with their signatures, together with any additional native SQL functions that have been registered by the dialect.

### Implementation notes

A lot of effort has gone into reducing the overhead involved in implementing all these functions for every SQL dialect, so you’ll see that a couple of new techniques have been used:

- some rather fancy functions are defined in terms of other functions together with direct construction of SQM nodes: the `extract()` function is the best and most complex example of this, and
- a couple of functions delegate back to the `Dialect` in order to obtain a pattern string: `timestampadd()`, `cast()`, and `trim()` are examples of this.

While neither of these approaches are perfect, I think they’re a whole lot better than having lots and lots of dialect-specific subclasses of `AbstractSqmSelfRenderingFunctionDescriptor`.

## Dates and times

Doing computations with dates and times in HQL has previously meant directly calling database-specific SQL functions. This work introduces the following suite of facilities for working with dates and times:

- Portable `timestampadd()` and `timestampdiff()` functions.
- Layered over these functions, a whole system of durations and a datetime/duration algebra.
- A very powerful `extract()` functions with a long list of extractable field types.
- A very fancy `format()` function for formatting dates and times, according to the syntax defined by `java.time.format.DateTimeFormatter`.
- Better support for timezones, to the very limited extent that the databases themselves support the ANSI SQL type `timestamp with time zone`.
- New built-in type for `ZonedDateTime`, for use with all those databases which *don’t* have good support for timezones.
- Date time literals with a more pleasing syntax than the `{d …}`, `{t …}`, `{dt …}` JDBC-escape-style-syntax that is specified by JPA.
- Dialects may now customize the rendering of date time literals. For example, on SQL Server, use the `cast()` function instead of JDBC escapes, and on Postgres use ANSI-style literals. (Which works better.)

HQL durations are a simplified version of ANSI SQL `interval`s. Both the syntax and semantics have been simplified, without real loss of expressivity. As with the work on HQL functions described above, we would like to see this work used as a model for extensions to the JPA standard.

You can read more [here](https://github.com/hibernate/hibernate-orm/pull/2863).

## Other changes

Inevitably, in the process of all this development, I ran across a few other sources of discomfort.

### HQL improvements

- Allow the `select` clause to come after the `from`/`where`/`group by`/`having` clauses. It really occurs in the wrong order in SQL, and this is a totally trivial change.
- Allow `/* ... */` delimited comments in HQL.
- Make the `from` clause optional, which is very useful for testing, automatically adding `from dual` to the generated SQL if necessary.
- I had to remove support for octal literals in HQL because (a) they're completely useless, and (b) they conflicted with the need to have integers with leading zeros in date/time literals.

### Improvements to automatic dialect resolution

The enumeration of dialects in `org.hibernate.dialect.Database` has been improved to include information about the JDBC URL format and JDBC driver class name for each database. Thus, we can now determine the correct `Dialect` to use given the JDBC URL alone, and even automatically load the JDBC driver class (for databases which still don’t support self-loading drivers), even in testing environments.

In addition, the logging and error reporting in this area has been improved. Previously Hibernate produced a rather confusing error message in some extremely common cases of misconfigured systems.

### Fixes to HQL implementation

I found and fixed some bugs in the new HQL translator:

- The operator precedence rules for expressions and predicates were wrong. The list in the ANTLR grammar must be from highest to lowest precedence.
- Parenthesized expressions in HQL were broken.
- Unary `+`/`-` operators were unimplemented.
- The function call `count(e)` when applied to an entity reference `e` was broken.
- The JPA `function()` syntax was broken.
- The grammar for the `atan2()` function was incorrect.
- Generated SQL would include the `as` keyword on databases which don’t recognize it (Oracle).

### `setFirstResult()` for databases with no `offset` clause

For dialects which don't support an offset I've reverted to the old pre-Hibernate-5 behavior where `setFirstResult()` didn't throw an exception. The old behavior was perfectly fine and correct. If I have `setMaxResults(20)` and `setFirstResult(40)`, i.e. the 3rd page of 20 results each, I get a perfectly well-performing
    
        select top 60 ....
    
It's wrong to complain that this query has bad performance, since (a) it is perfectly fast enough, and (b) it is the fastest thing that you can possibly achieve in a DB with no support for offsets or row numbers. This change should not have been made in Hibernate 5.

### Miscellaneous

- Automatically log a list of self-registered JDBC drivers when no JDBC driver is set explicitly.
- Don’t swallow connection errors in `BasicConnectionCreator`, that caused a lot of problems resolving issues with rejected connections.


gavinking:
OK, now a couple of comments about problems I encountered during this (extremely painful) merge.

## `Type`s vs `MappingModelExpressable`s

First, there are several places where I was forced to jump through hoops because I couldn't find any good way to get a `MappingModelExpressable` from a `Type`, or a `Type` from a `MappingModelExpressable`.

This has cost me hours and hours of work and hair-pulling, and I sure hope this is not just a matter of me "not getting" something about the new design.

Examples of this are:

```
				MappingModelExpressable type = adjustedTimestamp.getExpressionType();
				if (type instanceof SqmExpressable) {
					adjustedTimestampType = (SqmExpressable) type;
				}
				else if (type instanceof BasicValuedMapping) {
					adjustedTimestampType = ((BasicValuedMapping) type).getBasicType();
				}
				else {
					// else we know it has not been transformed
					adjustedTimestampType = expression.getLeftHandOperand().getNodeType();
				}
```

That's nasty as hell, and doesn't even work properly: it's a cause of at least 2 bugs in the duration arithmetic stuff.

A second, even uglier, example:

```
	private MappingModelExpressable<?> getMappingModelExpressable(
			SqmToSqlAstConverter walker,
			AllowableFunctionReturnType<?> resultType) {
		MappingModelExpressable<?> mapping;
		if ( resultType instanceof MappingModelExpressable ) {
			// here we have a BasicType, which can be cast
			// directly to BasicValuedMapping
			mapping = (MappingModelExpressable<?>) resultType;
		}
		else {
			// here we have something that is not a BasicType,
			// and we have no way to get a BasicValuedMapping
			// from it directly
			mapping = returnTypeResolver.resolveFunctionReturnType(
					() -> {
						try {
							// I think it's supposed to be this, but
							// resolveMappingExpressable() looks to
							// be unfinished, and throws
							return (BasicValuedMapping)
									walker.getCreationContext().getDomainModel()
											.resolveMappingExpressable( getNodeType() );
						}
						catch (Exception e) {
							return null; // this works at least approximately
						}
					},
					resolveSqlAstArguments( arguments, walker )
			);
		}
		return mapping;
	}
```

The code above seems to approximately work, but it meant I had to do disgusting violence to the `FunctionReturnTypeResolver`s.

Now I'm pretty sure I noticed things in unrelated code which also don't work completely, so I think what's going on here is just that something is unfinished here, is that right @sebersole?

Anyway, whether that's the case or not, there needs to be a well-defined way to convert between `Type` and `MappingModelExpressable` and vice-versa. I do not understand this code well enough to do it myself.

## JDBC parameter bindings

In the previous Hibernate 6 branch, there was an indirection between HQL parameter indices and SQL parameter indices that let me quite easily have a parameter occurring as a HQL function argument map to two different SQL parameters. This was useful for certain function transformations I do, namely in `overlay()` on some platforms.

This indirection doesn't seem to exist on the new branch, but I imagine it's not too hard to reimplement it, and I'll try to find time to do that myself.

## Query literal rendering

Previously there was a setting that allowed the `Dialect`s to choose how literals occurring in HQL queries would be translated. That setting is gone, and it means that literals occurring in the `select` clause get translated to SQL parameters, which breaks on some platforms.

I think this should be easy to fix, and, again, I'll try to find time to fix it myself.

## enum literals

On the other branch I had support for Java enum literals in HQL working, for example:

    from Person p where p.sex = test.hql.Sex.FEMALE

I have not yet tried to bring this over to the new branch.

sebersole:
I just resolved the conflicts and pushed this upstream.

sebersole:
Regarding what you labeled "Type vs MappingModelExpressables"... its really the conversion from the JPA "de-mapped" model to Hibernate's mapping model (persisters, etc) as part of the interpretation of SQM to SQL.  And you are right, that is not particularly well defined yet.

sebersole:
Regarding "JDBC parameter bindings".. 

You want what `org.hibernate.query.sqm.sql.BaseSqmToSqlAstConverter#consumeSqmParameter` does when its type is an embeddable but in cases where the `SqmParameter` is not an embeddable.  

In other words, iiuc, you want some way to influence the type used to transform the SqmParameter in `#consumeSqmParameter` - beyond just the `SqmParameter` type.  Correct?

sebersole:
"Query literal rendering"  - 

TBH I thought `AS_PARAM_OUTSIDE_SELECT` was the default.  But anyway, I don't think we explicitly removed the ability for the Dialect to define the default so much as it probably just did not get moved from the "old 6" branch to the new

gavinking:
> I just resolved the conflicts and pushed this upstream.

Yaaay, thankyou @sebersole, that's a massive load off my mind! 🙏

> its really the conversion from the JPA "de-mapped" model to Hibernate's mapping model

Yeah, TBH, I didn't understand what those abstractions represented and some day I'll need you guys to explain them to me.

> You want what `org.hibernate.query.sqm.sql.BaseSqmToSqlAstConverter#consumeSqmParameter` does when its type is an embeddable but in cases where the `SqmParameter` is not an embeddable.  

I don't have it in front of me but if I understand correctly, that sounds maybe about right.

> In other words, iiuc, you want some way to influence the type used to transform the SqmParameter in `#consumeSqmParameter` - beyond just the `SqmParameter` type.  Correct?

Well it's not a type transformation exactly it's something like this:

    overlay('xxxxxx' placing ?1 from 3)

Gets transformed to:

    insert('xxxxxx', 3, character_length(?), ?)

And then we need to bind parameter 1 *twice*.

There's a `@FailureExpected` test for this: `FunctionTests.testOverlayFunctionParameters()`.

> TBH I thought `AS_PARAM_OUTSIDE_SELECT` was the default.  But anyway, I don't think we explicitly removed the ability for the Dialect to define the default so much as it probably just did not get moved from the "old 6" branch to the new

Oh well all I did was look for `AvailableSettings.QUERY_LITERAL_RENDERING` and I didn't find it, but perhaps the actual functionality is still there somewhere ... I will investigate further tomorrow. I didn't look into it very deeply.

gavinking:
P.S. bear in mind that I have some additional tests that I will clean up and send as a second pull request once we get this `MappingModelExpressable` stuff sorted. Or maybe I will just send them anyway with the broken ones marked `@FailureExpected`.

gavinking:
> Oh well all I did was look for `AvailableSettings.QUERY_LITERAL_RENDERING` and I didn't find it,

Aaaah, I see now, it is `CRITERIA_LITERAL_HANDLING_MODE` on this branch. OK then, I can use that in the `Dialect`s, I suppose.

sebersole:
Yes, unfortunately it seems like Vlad did not listen when I had said to
make this more query general rather than criteria-specific.

NathanQingyangXu:
BTW, there is a PR to create CockroachDB dialect from one developer of that company: https://github.com/hibernate/hibernate-orm/pull/3137

gavinking:
> there is a PR to create CockroachDB dialect from one developer of that company

Interesting, they submitted that recently by the looks (I wrote mine back in June).

I like mine a whole lot better, however.

NathanQingyangXu:
Yeah, I saw you left a comment in that PR. Thanks. Not sure when he will take a look at it, though.

PR #3197, opened 2020-01-31
Prevent overflow when persisting Durations

gavinking:
`Duration`s should be converted to `BigDecimal`s for storage in `NUMERIC` columns, since a `Duration` might be too big to fit in a `long`.

(I missed this change when I merged my work from the previous branch.)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaDescriptor.java
@@ -6,13 +6,28 @@
  */
 package org.hibernate.type.descriptor.java;
 
+import java.math.BigDecimal;
 import java.time.Duration;
 
 import org.hibernate.dialect.Dialect;
+import org.hibernate.internal.util.StringHelper;
 import org.hibernate.type.descriptor.WrapperOptions;
 
 /**
+ * Descriptor for {@link Duration}, which is represented internally
+ * as ({@code long seconds}, {@code int nanoseconds}), approximately
+ * 28 decimal digits of precision. This quantity must be stored in
+ * the database as a single integer with units of nanoseconds, since
+ * the ANSI SQL {@code interval} type is not well-supported.
+ *
+ * In practice, the 19 decimal digits of a SQL {@code bigint} are
+ * capable of representing six centuries in nanoseconds and are
+ * sufficient for many applications. However, by default, we map
+ * Java {@link Duration} to SQL {@code numeric(21)} here, which
+ * can comfortably represent 60 millenia of nanos.
+ *
  * @author Steve Ebersole
+ * @author Gavin King
  */
 public class DurationJavaDescriptor extends AbstractTypeDescriptor<Duration> {
 	/**
@@ -30,7 +45,10 @@ public class DurationJavaDescriptor extends AbstractTypeDescriptor<Duration> {
 		if ( value == null ) {
 			return null;
 		}
-		return String.valueOf( value.toNanos() );
+		String seconds = String.valueOf( value.getSeconds() );
+		String nanos = String.valueOf( value.getNano() );
+		String zeros = StringHelper.repeat( '0', 9-nanos.length() );
+		return seconds + zeros + nanos;
 	}
 
 	@Override
@@ -38,7 +56,11 @@ public class DurationJavaDescriptor extends AbstractTypeDescriptor<Duration> {
 		if ( string == null ) {
 			return null;
 		}
-		return Duration.ofNanos( Long.parseLong( string ) );
+		int cutoff = string.length() - 9;
+		return Duration.ofSeconds(
+				Long.parseLong( string.substring( 0, cutoff ) ),
+				Long.parseLong( string.substring( cutoff ) )
+		);
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


gavinking:
So actually this is a bug fix. Commit 9593082 actually broke `DurationType`, and I didn't notice because there were no test failures. I'm trying to figure out why the tests didn't fail.

gavinking:
> I'm trying to figure out why the tests didn't fail.

OK, so actually there's nowhere in the test suite where we actually instantiate a `Duration` and try to persist it to the database.

gavinking:
Folks I would like to add a test for `DurationType`. Where should it go?

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaDescriptor.java
```diff
@@ -30,15 +44,22 @@ public String toString(Duration value) {
 		if ( value == null ) {
 			return null;
 		}
-		return String.valueOf( value.toNanos() );
+		return new StringBuilder()
```
NathanQingyangXu: It seems we can specify the capacity of `StringBuilder` to avoid relocating new char array and copying (it would definitively happen for the internal char array's length is 16 by default).

gavinking: I don't think `toString()` is used for anything important, much less for anything performance sensitive, but there's a *much* worse problem right in on the next line that I just fixed, so I fixed your objection while I was at it.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaDescriptor.java
```diff
@@ -30,15 +44,22 @@ public String toString(Duration value) {
 		if ( value == null ) {
 			return null;
 		}
-		return String.valueOf( value.toNanos() );
+		return new StringBuilder(22)
```
NathanQingyangXu: thanks for the change. Really exciting to interact with you.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaDescriptor.java
NathanQingyangXu: no objection at all. As a purely newbie, currently I can only find some minor programming imperfection. It is a great honour to review your code, :). Because of the same reason, I am afraid I am not fully qualified to approve it, unless I am pretty sure. Thanks for the quick response. It is exciting to interact with you here.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaDescriptor.java
```diff
@@ -30,15 +44,22 @@ public String toString(Duration value) {
...
+				.append( value.getSeconds() )
+				.append( value.getNano() )
+				.toString();
```
NathanQingyangXu: To my understanding, the `nano` part is the fraction of second. Its value could be any value from [0, 999_999_999]. Suppose we specify the Duration as `new Duration(24 * 3600, 0)`, this part would be zero. It seems the encoding and decoding would not work as expected (assuming the nano part is always of length of 9).

If I were the author, I would introduce a delimiter (seems '.' is a natural choice) instead.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaDescriptor.java
```diff
@@ -30,15 +44,22 @@ public String toString(Duration value) {
...
 	@Override
 	public Duration fromString(String string) {
 		if ( string == null ) {
 			return null;
 		}
-		return Duration.ofNanos( Long.parseLong( string ) );
+		int cutoff = string.length() - 9;
+		return Duration.ofSeconds(
+				Long.parseLong( string.substring( 0, cutoff ) ),
+				Long.parseLong( string.substring( cutoff ) )
```
NathanQingyangXu: The above code logic seems based on the assumption that nano part's length is of 9, which is not necessarily true.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaDescriptor.java
gavinking: Yes, thanks, of course you're right. I'll fix it.

I'm 99% sure this code is never ever called (it's not used for rendering literal durations), but I'll fix it anyway :-)

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaDescriptor.java
gavinking: That bit looks perfectly fine to me.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaDescriptor.java
gavinking: Fixed, I believe.

NathanQingyangXu:
I am new to v6 and someone just told me we only run tests in org.hibernate.orm package. Do you use Zulip, btw? You can get quick answer there for everybody is using it.

NathanQingyangXu:
https://hibernate.zulipchat.com/#narrow/stream/132096-hibernate-user, if you happen to never use it. You can see there is a dev channel and everybody would love to respond to gavin king there.

dreab8:
at the moment there is a filter active on 6.0 and only tests belonging to the `org.hibernate.orm.test` package are executed by gradle

gavinking:
> at the moment there is a filter active on 6.0

OK then, so I'll assume that there are tests for properties of type`Duration` that are currently disabled by the filter.

So then if nobody objects, I'll go ahead and apply this patch, since it's a fix to something I broke in #3196.

gavinking:
Applied manually

PR #3198, opened 2020-01-31
Dialect.getFallbackSqmMutationStrategy()

gavinking:
This method threw lots of `NotImplementedFor6Exception`s on this branch, so I've tried to figure out what the right implementations are, going by what was there in the other branches. But I'm not at all sure that I got everything here right, so please review this @sebersole @dreab8.

This commit also gives `TeradataDialect` back its `IndexExporter`.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -44,6 +44,7 @@ import org.hibernate.mapping.Constraint;
 import org.hibernate.mapping.ForeignKey;
 import org.hibernate.mapping.Index;
 import org.hibernate.mapping.Table;
+import org.hibernate.metamodel.mapping.CompositeIdentifierMapping;
 import org.hibernate.metamodel.mapping.EntityMappingType;
 import org.hibernate.metamodel.mapping.SqlExpressable;
 import org.hibernate.metamodel.spi.RuntimeModelCreationContext;
@@ -56,6 +57,11 @@ import org.hibernate.query.TrimSpec;
 import org.hibernate.query.hql.HqlTranslator;
 import org.hibernate.query.spi.QueryEngine;
 import org.hibernate.query.spi.QueryOptions;
+import org.hibernate.query.sqm.mutation.internal.idtable.AfterUseAction;
+import org.hibernate.query.sqm.mutation.internal.idtable.IdTable;
+import org.hibernate.query.sqm.mutation.internal.idtable.PersistentTableStrategy;
+import org.hibernate.query.sqm.mutation.internal.idtable.PhysicalIdTableExporter;
+import org.hibernate.query.sqm.mutation.internal.inline.InlineStrategy;
 import org.hibernate.query.sqm.mutation.spi.SqmMultiTableMutationStrategy;
 import org.hibernate.query.sqm.sql.SqmTranslatorFactory;
 import org.hibernate.service.ServiceRegistry;
@@ -1573,7 +1579,18 @@ public abstract class Dialect implements ConversionContext {
 	public SqmMultiTableMutationStrategy getFallbackSqmMutationStrategy(
 			EntityMappingType entityDescriptor,
 			RuntimeModelCreationContext runtimeModelCreationContext) {
-		throw new NotYetImplementedFor6Exception( getClass() );
+		if ( entityDescriptor.getIdentifierMapping() instanceof CompositeIdentifierMapping) {
+			if ( !supportsTuplesInSubqueries() ) {
+				return new InlineStrategy( this );
+			}
+		}
+		
+		return new PersistentTableStrategy(
+				new IdTable( entityDescriptor, name -> name ),
+				AfterUseAction.CLEAN,
+				PhysicalIdTableExporter::new,
+				runtimeModelCreationContext.getSessionFactory()
+		);
 	}
 
 	// callable statement support ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -1573,7 +1579,18 @@ public String getCreateMultisetTableString() {
 	public SqmMultiTableMutationStrategy getFallbackSqmMutationStrategy(
 			EntityMappingType entityDescriptor,
 			RuntimeModelCreationContext runtimeModelCreationContext) {
-		throw new NotYetImplementedFor6Exception( getClass() );
+		if ( entityDescriptor.getIdentifierMapping() instanceof CompositeIdentifierMapping) {
+			if ( !supportsTuplesInSubqueries() ) {
+				return new InlineStrategy( this );
```
gavinking: Doesn't work for now, because `InlineStrategy` itself is unfinished and throws.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -1573,7 +1579,18 @@ public String getCreateMultisetTableString() {
...
+			}
+		}
+		
+		return new PersistentTableStrategy(
+				new IdTable( entityDescriptor, name -> name ),
```
gavinking: Is this right? Should it add `"HT_"` to the table name?

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
```diff
@@ -594,4 +601,20 @@ public JDBCException convert(SQLException sqlException, String message, String s
...
+				? super.getFallbackSqmMutationStrategy( entityDescriptor, runtimeModelCreationContext )
+				: new GlobalTemporaryTableStrategy(
+						new IdTable( entityDescriptor, name -> "HT_" + name ),
+						() -> new TempIdTableExporter( false, this::getTypeName ) {
+							@Override
+							protected String getCreateOptions() {
+								return "on commit delete rows";
+							}
+						},
+						AfterUseAction.CLEAN,
```
gavinking: Is this right?

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -1573,7 +1579,18 @@ public String getCreateMultisetTableString() {
...
+				AfterUseAction.CLEAN,
```
gavinking: Is this right?

hibernate-core/src/main/java/org/hibernate/dialect/TimesTenDialect.java
```diff
@@ -216,28 +219,18 @@ public boolean isCurrentTimestampSelectStringCallable() {
...
+		return new GlobalTemporaryTableStrategy(
+				new IdTable( rootEntityDescriptor,
+						name -> name.length() > 30 ? name.substring( 0, 30 ) : name ),
+				() -> new TempIdTableExporter( false, this::getTypeName ) {
+					@Override
+					protected String getCreateOptions() {
+						return "on commit delete rows";
+					}
+				},
+				AfterUseAction.CLEAN,
```
gavinking: Is this correct?

hibernate-core/src/main/java/org/hibernate/dialect/SAPDBDialect.java
```diff
@@ -195,23 +198,18 @@ public CaseFragment createCaseFragment() {
...
-//					public String getCreateIdTableStatementOptions() {
-//						return "ignore rollback";
-//					}
-//				},
-//				AfterUseAction.DROP,
-//				null
-//		);
+		return new LocalTemporaryTableStrategy(
+				new IdTable( rootEntityDescriptor, name -> "temp." + name ),
+				() -> new PhysicalIdTableExporter() {
```
gavinking: Is it correct to use `PhysicalIdTableExporter` here, even though it's really a temp table?

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -1573,7 +1579,18 @@ public String getCreateMultisetTableString() {
...
+				PhysicalIdTableExporter::new,
```
gavinking: Is this right?

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
sebersole: I believe you are correct

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
sebersole: Yes, we want to clean the rows after we use them

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
sebersole: Yes, we want to export the physical table

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
sebersole: Not sure how firebird implements their global-temp-table support.  Cleaning it should be fine, its more a question of whether its needed.  I'd leave it as is

hibernate-core/src/main/java/org/hibernate/dialect/SAPDBDialect.java
sebersole: No, it should use `TempIdTableExporter`

hibernate-core/src/main/java/org/hibernate/dialect/TimesTenDialect.java
sebersole: See earlier

hibernate-core/src/main/java/org/hibernate/dialect/SAPDBDialect.java
gavinking: OK thanks I will fix that.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
gavinking: Thanks.

hibernate-core/src/main/java/org/hibernate/dialect/SAPDBDialect.java
gavinking: Done.

gavinking:
@sebersole WDYT? Can we merge this now?

PR #3200, opened 2020-01-31
Remove dialect for Mckoi SQL

gavinking:
The last release of this database was in 2012, and anyway [nobody is using it](https://stackoverflow.com/search?tab=newest&q=mckoi).


sebersole:
Applied it manually

PR #3201, opened 2020-01-31
Remove dialects for Interbase, Pointbase, and Progress

gavinking:
The case for removing these three is not quite as clear as for McKoi or JDataStore. However, none of these dialects seem to be actively maintained, and the databases themselves look to be moribund. My recommendation is we don't need them in Hibernate 6.

A user can always maintain their own dialect if they really need it.


sebersole:
Applied manually

PR #3202, opened 2020-01-31
Set CRITERIA_LITERAL_HANDLING_MODE in Dialects

gavinking:
On the abandoned branch, some dialects set `QUERY_LITERAL_RENDERING`, but these settings got lost when I merged the code, because I could not find the setting.

Does this look right to you, @sebersole?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
@@ -115,6 +115,7 @@ public class DerbyDialect extends Dialect {
 				: new DerbyLimitHandler( getVersion() >= 1060 );
 
 		getDefaultProperties().setProperty( Environment.STATEMENT_BATCH_SIZE, NO_BATCH );
+		getDefaultProperties().setProperty( Environment.CRITERIA_LITERAL_HANDLING_MODE, "literal" );
 	}
 
 	public int getDefaultDecimalPrecision() {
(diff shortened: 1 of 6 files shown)


sebersole:
I think it makes better sense to define an explicit API for this.  Never been a fan of the "augment the properties" approach (chicken-egg).

Something like `org.hibernate.query.QueryLiteralRendering Dialect#getFallbackQueryLiteralRendering`?

gavinking:
> I think it makes better sense to define an explicit API for this.

Sure, I don't like it either, but that applies equally to the other properties like `STATEMENT_BATCH_SIZE`, `MAX_FETCH_DEPTH`, `NON_CONTEXTUAL_LOB_CREATION`, and `USE_STREAMS_FOR_BINARY`.

So I'm not sure how you would like to do it. The easy thing to do would simply be t add typesafe setters for these things to `Dialect`, that just hide the `getDefaultProperties()` from subclasses.

sebersole:
> Sure, I don't like it either, but that applies equally to the other properties like `STATEMENT_BATCH_SIZE`, `MAX_FETCH_DEPTH`, `NON_CONTEXTUAL_LOB_CREATION`, and `USE_STREAMS_FOR_BINARY`.

+100.  I dislike it for those as well ;)

But I don't think its critical for a 6.0 Final

sebersole:
Could we apply the discussed changes?

gavinking:
@sebersole sorry, I missed that you had requested changes to this ... and I can't seem to see what they were ....

sebersole:
No worries.  I'm just going to hold off on this and I'll come back and do what I think is right in terms of these various "Dialect default" settings

gavinking:
@sebersole OK, no problem at all.

beikov:
We introduced `CRITERIA_VALUE_HANDLING_MODE` to handle this. You can read the documentation about the property, but I think you will like the new default in 6.0

PR #3204, opened 2020-01-31
Log usage of deprecated Dialect classes

gavinking:
The database-version-specific dialects are now `@Deprecated`. So produce a log message to this effect.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -83,6 +83,7 @@ import java.time.temporal.TemporalAccessor;
 import java.util.Date;
 import java.util.*;
 
+import static org.hibernate.internal.log.DeprecationLogger.DEPRECATION_LOGGER;
 import static org.hibernate.type.descriptor.DateTimeUtils.*;
 
 /**
@@ -136,7 +137,7 @@ public abstract class Dialect implements ConversionContext {
 	// constructors and factory methods ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 	protected Dialect() {
-		LOG.usingDialect( this );
+		logSelectedDialect();
 
 		registerColumnType( Types.BIT, 1, "bit" );
 		registerColumnType( Types.BIT, "bit($l)" );
@@ -223,6 +224,22 @@ public abstract class Dialect implements ConversionContext {
 
 		uniqueDelegate = new DefaultUniqueDelegate( this );
 		defaultSizeStrategy = new DefaultSizeStrategyImpl();
+	}
+
+	private void logSelectedDialect() {
+		LOG.usingDialect( this );
+
+		Class<? extends Dialect> dialect = getClass();
+		if ( dialect.isAnnotationPresent(Deprecated.class) ) {
+			Class<?> superDialect = dialect.getSuperclass();
+			if ( !superDialect.isAnnotationPresent(Deprecated.class)
+				&& !superDialect.equals(Dialect.class) ) {
+				DEPRECATION_LOGGER.deprecatedDialect( dialect.getSimpleName(), superDialect.getName() );
+			}
+			else {
+				DEPRECATION_LOGGER.deprecatedDialect( dialect.getSimpleName() );
+			}
+		}
 	}
 
 	/**
--- a/hibernate-core/src/main/java/org/hibernate/internal/log/DeprecationLogger.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/log/DeprecationLogger.java
@@ -249,4 +249,15 @@ public interface DeprecationLogger extends BasicLogger {
 					"5.3 in upgrading.  It will be removed in a later version."
 	)
 	void logUseOfDeprecatedZeroBasedJdbcStyleParams();
+
+	@LogMessage(level = WARN)
+	@Message(value = "%s has been deprecated",
+			id = 90000025)
+	void deprecatedDialect(String dialect);
+
+	@LogMessage(level = WARN)
+	@Message(value = "%s has been deprecated; use %s instead",
+			id = 90000025)
+	void deprecatedDialect(String dialect, String replacement);
+
 }
--- a/hibernate-core/src/main/java/org/hibernate/internal/CoreMessageLogger.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/CoreMessageLogger.java
@@ -182,15 +182,6 @@ public interface CoreMessageLogger extends BasicLogger {
 	void deprecatedForceDescriminatorAnnotation();
 
 	@LogMessage(level = WARN)
-	@Message(value = "The Oracle9Dialect dialect has been deprecated; use either Oracle9iDialect or Oracle10gDialect instead",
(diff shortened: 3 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -225,6 +225,22 @@ protected Dialect() {
 		defaultSizeStrategy = new DefaultSizeStrategyImpl();
 	}
 
+	private void logDeprecation() {
+		LOG.usingDialect( this );
+
+		Class<? extends Dialect> dialect = getClass();
+		if ( dialect.isAnnotationPresent(Deprecated.class) ) {
```
sebersole: This is against the Quarkus mandate that thou shall not read annotations at runtime.  Just saying

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -136,7 +136,7 @@
 	// constructors and factory methods ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 	protected Dialect() {
-		LOG.usingDialect( this );
+		logDeprecation();
```
sebersole: I don't think the separate method does anything here; just move that code inline.  Or at least name the method better - it does not just log deprecations

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
gavinking: Yeah I wondered about that... but this warning isn't necessary in Quarkus, so perhaps it doesn't matter?

gavinking:
@sebersole is that better?

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
sebersole: The warning is not, but the isAnnotationPresent would still happen.  I don't necessarily care, just pointing it out

hibernate-core/src/main/java/org/hibernate/internal/log/DeprecationLogger.java
```diff
@@ -249,4 +249,15 @@ void connectionProviderClassDeprecated(
...
 	void logUseOfDeprecatedZeroBasedJdbcStyleParams();
+
+	@LogMessage(level = WARN)
+	@Message(value = "%s has been deprecated",
+			id = 90000025)
+	void deprecatedDialect(String dialect);
+
+	@LogMessage(level = WARN)
+	@Message(value = "%s has been deprecated; use %s instead",
+			id = 90000025)
```
sebersole: I'm surprised this even compiles for you. These ids need to be unique.  The JBoss Logging AP checks that during compile

gavinking: Haha OK sorry I pushed super quickly as I was running out the door.

sebersole:
NP :)

hibernate-core/src/main/java/org/hibernate/internal/log/DeprecationLogger.java
gavinking: Hrm, strange, for me it compiles and no tests fail.

Perhaps the ids don't need to be unique when you have an overloaded method name like what I have here?

Or perhaps I'm doing something wrong?

Investigating...

gavinking: Seems to work perfectly:

    WARN: HHH90000025: MySQL8Dialect has been deprecated; use org.hibernate.dialect.MySQLDialect instead

Anyway, @sebersole I'm perfectly happy to give them two different `id`s, I gave them the same `id` only because I figured that the represented the same problem (use of a deprecated `Dialect`).

Your call.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
gavinking: I renamed it.

hibernate-core/src/main/java/org/hibernate/internal/log/DeprecationLogger.java
sebersole: Maybe it does account for overloaded names, not sure.  I agree it is the same underlying issue and would be nice to reuse the same message id.  Just thought that was not allowed (I know its not in other cases).

@dmlloyd Any thoughts on this?  Thanks

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
sebersole: @Sanne what do you think here?

hibernate-core/src/main/java/org/hibernate/internal/log/DeprecationLogger.java
dmlloyd: Overloading is OK if both messages pertain to the "same" thing.  The definition of "same" is a bit hazy but IMO this case would qualify.

IIRC overloading the same ID is allowed by the tooling as long as the method name matches as well.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
Sanne: Currently the constructor of Dialect is somethign which is not invoked at runtime, so this shouldn't be a problem.

I do wonder though if this method isn't better off to be invoked explicitly from the Dialect factory, rather than being a side effect of the constructor - I believe that would give us some more flexibility regarding wanting such a check to happen.

hibernate-core/src/main/java/org/hibernate/internal/log/DeprecationLogger.java
sebersole: Awesome to know.  Thanks @dmlloyd !

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
sebersole: Personally I like the "do this in `DialectFactory`" approach better, but if its not a problem for Quarkus then its not a big deal to me.

hibernate-core/src/main/java/org/hibernate/internal/log/DeprecationLogger.java
sebersole: Awesome to know.  Thanks @dmlloyd

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
gavinking: I'm perfectly happy to do it in `DialectFactory`, my only minor concern would be is that if some people have code where the instantiate their `Dialect` directly, outside of the `DialectFactory` infrastructure, then they might not get the warning. On the other hand, if they're doing that, then I guess they would get a compilation warning in their code. So that's probably good enough.

gavinking: Let me know what you guys think and I'll make the change if that's what you prefer.

Sanne:
If you can make the change that would be great. Thanks!

PR #3205, opened 2020-01-31
HHH-13838 Allow extension of PersistenceXmlParser

Sanne:
hi @sebersole , would you be ok with these changes? Opening up visibility a little bit.

https://hibernate.atlassian.net/browse/HHH-13838


sebersole:
Sure

PR #3206, opened 2020-01-31
Determine type of self rendering function lazily during type access

beikov:


gavinking:
@beikov Interesting. I will try this out.

gavinking:
Ugh, sorry, my apologies, @beivok, for not responding, it kinda just slipped my mind.

I actually *did* try this out, hoping that it would fix a bug I had pending with the date/time functions stuff, and sadly it didn't seem to fix the particular bug I had in mind, but in principle it looks like a good fix to me, the only reason I didn't have any particular feedback was I wasn't sure precisely which cases it fixes.

But by all means go ahead and merge.

PR #3208, opened 2020-01-31
H2 supports 'true' and 'false' boolean literals

gavinking:
H2 supports (ANSI-standard) [boolean literals](https://h2database.com/html/grammar.html#boolean), i.e. `true` and `false`. For some reason, probably oversight, our `H2Dialect` doesn't use 'em.

Does anyone know of any reason we shouldn't make the attached change?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
@@ -184,6 +184,11 @@ public class H2Dialect extends Dialect {
 	}
 
 	@Override
+	public String toBooleanValueString(boolean bool) {
+		return String.valueOf( bool );
+	}
+
+	@Override
 	public LimitHandler getLimitHandler() {
 		return limitHandler;
 	}


NathanQingyangXu:
Not sure about details, but there is a closed PR which seems pretty similar: https://github.com/hibernate/hibernate-orm/pull/3179

gavinking:
Yeah it looks like that PR was trying to use the change to fix a bug. I don’t think it’s a bug fix because I guess H2 converts freely between Booleans and truthy integers. 

So to be clear I’m not trying to fix any bugs, just trying to produce more elegant SQL.

gavinking:
@beikov WDYT?

beikov:
I didn't merge it because I didn't want some possible backwards incompatibility, but if you are certain that this is non-problematic, I'm +1 for integrating this as well.
My use case was that I wanted to compare the EXISTS predicate with a boolean literal and thought H2 could compare a predicate with a boolean literal but not with an integral literal. Unfortunately, H2 doesn't support that, so the fix would not have fixed my original issue.

gavinking:
@beikov 

> I didn't merge it because I didn't want some possible backwards incompatibility

This PR is for Hibernate 6, where strict backwards compatibility is less of a critical requirement. But I doubt that this change will break anything.

gavinking:
Applied manually.

PR #3209, opened 2020-02-01
On MariaDB limit queried sequences to those for the current database

gavinking:
This change reproduces the [fix](https://github.com/hibernate/hibernate-orm/pull/3147) by @NathanQingyangXu in the new Hibernate 6 `MariaDBDialect`.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/MariaDBDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/MariaDBDialect.java
@@ -80,7 +80,7 @@ public class MariaDBDialect extends MySQLDialect {
 	@Override
 	public String getQuerySequencesString() {
 		return getSequenceSupport().supportsSequences()
-				? "select table_name from information_schema.TABLES where table_type='SEQUENCE'"
+				? "select table_name from information_schema.TABLES where table_schema = database() and table_type='SEQUENCE'"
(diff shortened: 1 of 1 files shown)


NathanQingyangXu:
I am honored. Thanks.

hibernate-core/src/main/java/org/hibernate/dialect/MariaDBDialect.java
```diff
@@ -80,7 +80,7 @@ public SequenceSupport getSequenceSupport() {
 	@Override
 	public String getQuerySequencesString() {
 		return getSequenceSupport().supportsSequences()
-				? "select table_name from information_schema.TABLES where table_type='SEQUENCE'"
+				? "select table_name from information_schema.TABLES where table_schema = database() and table_type='SEQUENCE'"
```
NathanQingyangXu: After second thought, I think the SQL statement above might only apply for 10.3 and later for sequence was only introduced since 10.3: https://mariadb.com/kb/en/create-sequence/

For early versions of Mariadb, not sure whether this is correct, however. Could we rely on db version to make the decision? I think this is an example why previous version specific dialect makes sense to some extent, IMHO.

dreab8:
Thanks @gavinking,

I'm closing this PR, I have just finished merging 6.0 with master and this change has been applied.

PR #3210, opened 2020-02-01
Rename SAPDBDialect to MaxDBDialect

gavinking:
The product was renamed a long time ago, and so I think we should rename the `Dialect` too.


gavinking:
Be a little careful with merging this one because it's going to conflict with #3153

gavinking:
Does anyone object to this change?

sebersole:
Applied this manually - https://github.com/hibernate/hibernate-orm/commit/7267b418c508756a6bdccf92e8429bdd8582cc94

PR #3213, opened 2020-02-02
Binary literals

gavinking:
This patch adds support to HQL for hexadecimal-format binary literals, using the standard SQL syntax `X'deadbeef'`.


hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ByteArrayTypeDescriptor.java
```diff
@@ -67,7 +67,7 @@ public String toString(Byte[] bytes) {
 		Byte[] bytes = new Byte[string.length() / 2];
 		for ( int i = 0; i < bytes.length; i++ ) {
 			final String hexStr = string.substring( i * 2, (i + 1) * 2 );
-			bytes[i] = (byte) ( Integer.parseInt( hexStr, 16 ) + Byte.MIN_VALUE );
+			bytes[i] = (byte) Integer.parseInt( hexStr, 16 );
```
gavinking: This old code wasn't consistent with how JDBC drivers handle the conversion from unsigned to signed bytes.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/PrimitiveByteArrayTypeDescriptor.java
```diff
@@ -73,7 +73,7 @@ public String extractLoggableRepresentation(byte[] value) {
 		byte[] bytes = new byte[string.length() / 2];
 		for ( int i = 0; i < bytes.length; i++ ) {
 			final String hexStr = string.substring( i * 2, (i + 1) * 2 );
-			bytes[i] = (byte) (Integer.parseInt(hexStr, 16) + Byte.MIN_VALUE);
+			bytes[i] = (byte) Integer.parseInt( hexStr, 16 );
```
gavinking: This old code wasn't consistent with how JDBC drivers handle the conversion from unsigned to signed bytes.

hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQLDialect.java
```diff
@@ -691,6 +691,11 @@ public String translateExtractField(TemporalUnit unit) {
 		}
 	}
 
+	@Override
+	public String formatBinaryliteral(byte[] bytes) {
+		return "bytea '\\x" + StandardBasicTypes.BINARY.toString( bytes ) + "'";
```
gavinking: I *think* this is the best thing to do on PostgreSQL, though I'm not 100% certain.

gavinking:
An alternative syntax choice for this feature would be, instead of the SQL-style `x'DEAD01BEEF'`, one of:

    { 0xDE, 0xAD, 0x01, 0xBE, 0xEEF }

or even:

        [ 0xDE, 0xAD, 0x01, 0xBE, 0xEEF ]

Reflecting that in Java this is a `byte[]` array.

It's a lot more verbose, however.

gavinking:
Does nobody have any feedback on this?

dreab8:
the PR  looks good to me, about the alternative syntax I slightly prefer `{ 0xDE, 0xAD, 0x01, 0xBE, 0xEF }`

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlLexer.g4
```diff
@@ -93,6 +93,10 @@ UNICODE_ESCAPE
 	: 'u' HEX_DIGIT HEX_DIGIT HEX_DIGIT HEX_DIGIT
 	;
 
+BIN_LITERAL
```
sebersole: Can we use `BINARY_LITERAL`?  It's 3 extra characters and so much more immediately understandable.

gavinking: Sure.

gavinking:
> the PR looks good to me, about the alternative syntax I slightly prefer `{ 0xDE, 0xAD, 0x01, 0xBE, 0xEF }`

I pretty much don't care at all so if other people also prefer that, I'm very happy to change it.

Speak up!

sebersole:
Personally I dislike the `byte[] { 0xDE, 0xAD, 0x01, 0xBE, 0xEF }` syntax, but I think it would be awesome to support both of the others.  Since ya asked ;)  But I don't have particularly string opinions here

gavinking:
>  I think it would be awesome to support both of the others

We can easily do both.

sebersole:
Yep, Antlr ftw

gavinking:
> Antlr ftw

Done. WDYT?

PR #3216, opened 2020-02-03
Disallow FQN of entity name in JPA compliance mode

NathanQingyangXu:


hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
```diff
@@ -836,7 +836,9 @@ private EntityDomainType resolveEntityReference(String entityName) {
 		log.debugf( "Attempting to resolve path [%s] as entity reference...", entityName );
 		EntityDomainType reference = null;
 		try {
-			entityName = creationContext.getJpaMetamodel().qualifyImportableName( entityName );
+			if ( !creationOptions.useStrictJpaCompliance() ) {
```
sebersole: The intention was more to verify that a FQN was not used.  While this change does protect that from happening, it does not (probably) give a very help exception

NathanQingyangXu: thanks. This is a draft PR for I have not enough time for this. Definitively will spend more efforts.

sebersole: The "correct" thing to do would be to add this check after we have `reference`.  From there, if compliance is requested, we would check its `javax.persistence.metamodel.EntityType#getName` (as opposed to `org.hibernate.metamodel.model.domain.EntityDomainType#getHibernateEntityName`) to make sure it matches the incoming `entityName`.  HTH

sebersole:
TBH this is the first time I have seen this "draft PR".  Neat.

NathanQingyangXu:
Closed this PR because of the confusion of "draft PR". Will create a new one based on @sebersole 's comments.

NathanQingyangXu:
> TBH this is the first time I have seen this "draft PR". Neat.

TBH this is the first time I ever tried. When creating a PR, there is another option in dropdown, but guess people seldom noticed it.

PR #3217, opened 2020-02-04
HHH-13829 Support lifecycle events for embeddable classes in collections

gkatiforis:
So far lifecycle events are called only for a single embeddable object in entities. This PR adds the functionality in order to fire lifecycle events for embeddable objects which are in collections. 

Upon callback registration, collections with embeddable are detected and registered in CallbackRegistryImpl in order to be available to EmbeddableCallback upon lifecycle events. EmbeddableCallback performs the event for each embeddable inside the collection.

Also, metadata.initSessionFactory( this ); inside SessionFactoryImpl constructor is moved after the assignment of metamodel property because metamodel property is required to get the element type of the collection in EventListenerRegistryImpl.prepare.


hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
```diff
@@ -149,7 +151,17 @@ public void prepare(MetadataImplementor metadata) {
...
-				if ( property.getType().isComponentType() ) {
+				boolean mustBuildCallbacksForEmbeddable;
+				if(property.getType().isCollectionType()){
+					CollectionType collectionType = (CollectionType) property.getType();
+					final Type collectionElementType = collectionType.getElementType(getSessionFactory());
+					mustBuildCallbacksForEmbeddable = collectionElementType.isComponentType();
+				}
+				else {
+					mustBuildCallbacksForEmbeddable = property.getType().isComponentType();
+				}
```
NathanQingyangXu: There seems to be some room for code logic improvement. For instance
```
Type type = property.getType();
if ( type.isCollectionType() ) {
    type = ( ( CollectionType ) type ).getElementType( getSessionFactory() );
}

if ( type.isComponentType() ) { // instead of 'mustBuildCallbackForEmbeddable'
    ...
}
```

Also, it is highly recommended to use empty char delimiter consistently.
Feel free to ignore my comment. As usual, I am notorious to suggest generic coding improvement without approving for lack of enough domain knowledge, :).

hibernate-core/src/main/java/org/hibernate/jpa/event/internal/CallbackBuilderLegacyImpl.java
```diff
@@ -20,10 +20,8 @@
 import javax.persistence.PersistenceException;
 
 import org.hibernate.MappingException;
-import org.hibernate.annotations.common.reflection.ClassLoadingException;
-import org.hibernate.annotations.common.reflection.ReflectionManager;
-import org.hibernate.annotations.common.reflection.XClass;
-import org.hibernate.annotations.common.reflection.XMethod;
+import org.hibernate.annotations.common.reflection.*;
```
NathanQingyangXu: btw, do we have style rule regarding the `*` usage, btw?

hibernate-core/src/main/java/org/hibernate/jpa/event/internal/CallbackBuilderLegacyImpl.java
```diff
@@ -239,9 +237,22 @@ public void release() {
...
+					embeddableXClass = prop.getElementClass();
+					break;
+				}
+			}
+		}
+		//else get class from a single embeddable
+		if(embeddableXClass == null){
+			final String embeddableClassName = embeddableProperty.getType().getReturnedClass().getName();
+			embeddableXClass = reflectionManager.classForName( embeddableClassName );
+		}
```
NathanQingyangXu: again, we need more aggressive delimiter usage. Please refer to existing code for the pattern.

hibernate-core/src/main/java/org/hibernate/jpa/event/internal/CallbackBuilderLegacyImpl.java
```diff
@@ -239,9 +237,22 @@ public void release() {
...
 	public Callback[] resolveEmbeddableCallbacks(Class entityClass, Property embeddableProperty, CallbackType callbackType, ReflectionManager reflectionManager) {
-
-		final String embeddableClassName = embeddableProperty.getType().getReturnedClass().getName();
-		final XClass embeddableXClass = reflectionManager.classForName( embeddableClassName );
+		XClass embeddableXClass = null;
+		//First lookup for Collection of embeddable and get the embeddable class
+		XClass entityXClass = reflectionManager.classForName(entityClass.getName());
+		for (XProperty prop : entityXClass.getDeclaredProperties(AccessType.FIELD.getType())) {
+			if (prop.getName().equals(embeddableProperty.getName())) {
+				if(prop.isCollection()){
```
NathanQingyangXu: Seems we can combine the above two if together

hibernate-core/src/test/java/org/hibernate/event/EmbeddableCallbackTest.java
```diff
@@ -67,6 +67,65 @@ public void testNullEmbeddable() {
...
+		} );
+		user.get().contactAddresses.forEach( e -> assertEquals(1, e.postRemove));
+	}
+
+	@Test
+	@TestForIssue(jiraKey = "HHH-13829")
+	public void testNullCollectionOfEmbeddable() {
+		doInJPA( this::entityManagerFactory, entityManager -> {
+			User user = new User();
+		    user.id = 1;
```
NathanQingyangXu: seems some indentation inconsistency because of mixing of TAB and SPACE

hibernate-core/src/test/java/org/hibernate/event/EmbeddableCallbackTest.java
```diff
@@ -67,6 +67,65 @@ public void testNullEmbeddable() {
 		} );
 	}
 
+	@Test
+	@TestForIssue(jiraKey = "HHH-13829")
+	public void testCollectionOfEmbeddable() {
+		AtomicReference<User> user  = new AtomicReference<>(new User());
```
NathanQingyangXu: I am curious why `AtomicReference` was used. Seems everything is happening in one thread.

gkatiforis: I use `AtomicReference`  in order to access `user `object both inside and outside lambda expression. Without `AtomicReference` this is not possible.

gkatiforis:
@NathanQingyangXu thank you very much for your suggestions. Everything has been resolved.

hibernate-core/src/test/java/org/hibernate/event/EmbeddableCallbackTest.java
NathanQingyangXu: I gained some new knowledge, Thanks. I searched in the testing folder and found quite some precedence. Guess testing class field is also viable, but this seems enjoy better code readability, though my initial doubt would be pretty common.

NathanQingyangXu: IMHO, org.apache.commons.lang3.mutable.MutableObject (which seems to have been in our class path) seems a better alternative. Sort of misuse of AtomicReference for we are not using it for concurrency reason.

gkatiforis: MutableObject from apache commons seems more appropriate in this case but does not exist in classpath. Do you think it's better to add it or is it ok with `AtomicReference`?

sebersole: @NathanQingyangXu we are not going to add any apache commons dependencies.

Lambdas in Java often do require the use of `AtomicReference`, `AtomicInteger`, etc.  We run into this alot in 6.0 development where we do use lambdas as part of the code base.  It's an unfortunate necessity with lambdas.

However, @gkatiforis, here you do not really have to do it this way.  You are using `org.hibernate.testing.transaction.TransactionUtil#doInJPA`, but could also use `org.hibernate.testing.transaction.TransactionUtil2#fromTransaction` instead which returns a value.  

Personally I prefer `TransactionUtil2` over `TransactionUtil`.  Yes I wrote it, but the preference is more than that.  I like the naming and structure better in the `TransactionUtil2` contract

sebersole: E.g.,

````
final User user = fromTransaction(
        this::entityManagerFactory,
        em -> {
            final User u = ...;
            ...
            em.persist( u );
            ...
            return u;
        }
);
````

To be fair, this is a test and the performance is not so critical.  But still, good to be conscious of better performance opportunities regardless.  Lambdas in Java have quite a few performance considerations.

hibernate-core/src/main/java/org/hibernate/jpa/event/internal/CallbackBuilderLegacyImpl.java
sebersole: It should be avoided.

hibernate-core/src/main/java/org/hibernate/jpa/event/internal/CallbackBuilderLegacyImpl.java
sebersole: Also, I'd make the `prop.isCollection` check the primary - perf.  Its a check against a boolean field versus char-by-char comparison of a String.

hibernate-core/src/main/java/org/hibernate/jpa/event/internal/CallbackBuilderLegacyImpl.java
sebersole:
````
if(embeddableXClass == null){
````

is fugly in an readability way.  Please use spaces appropriately.  

````
if ( embeddableXClass == null ) {
````

would be more consistent with the rest of the codebase, although 

````
if (embeddableXClass == null) {
````

is fine as well.

Is this what you meant by "delimiter usage" @NathanQingyangXu ?

NathanQingyangXu: Yeah. Such format issue often troubles new contributor (including me). Is it an idea to add the link of https://github.com/hibernate/hibernate-ide-codestyles into our wiki article (i.e. https://github.com/hibernate/hibernate-orm/wiki/Contributing-Code)?

hibernate-core/src/main/java/org/hibernate/jpa/event/internal/CallbackBuilderLegacyImpl.java
NathanQingyangXu: Good point. Thanks. Will put it in my toolbox for my future PRs.

hibernate-core/src/test/java/org/hibernate/event/EmbeddableCallbackTest.java
NathanQingyangXu: To my knowledge, Lambdas (and previous inner instance class method) requires 'final' or 'effective final' (the rationale behind the language design is because of concurrency, but that is another story). The reason it is used in this PR is simply because it is a object wrapper. It seems this PR is not using its concurrency feature for everything is running in one thread (I agree if we are involving SessionFactory or other thread-safe object we need them). Also, we only use its 'getter' and 'setter' and any object wrapper (regardless of whether it is thead-safe or not) will work in this specific scenario. That is why I said I think it is sort of misuse.

But for sure, we should use your method. Good to learn of its necessity.

NathanQingyangXu:
> @NathanQingyangXu thank you very much for your suggestions. Everything has been resolved.

Thanks. As a contributor as you, I love to focus my code review on generic coding issue to improve code quality. I think you can expect approval from @sebersole , who knows all tech details and is the literal oracle on such PR.

hibernate-core/src/test/java/org/hibernate/event/EmbeddableCallbackTest.java
gkatiforis: `org.hibernate.testing.transaction.TransactionUtil2#fromTransaction` seems ideal for this case, so i  will remove `AtomicReference`. Thank you for pointing this out.

hibernate-core/src/main/java/org/hibernate/jpa/event/internal/CallbackBuilderLegacyImpl.java
NathanQingyangXu: After testing, I found the style file cannot be imported into the latest Intellij IDEA (2019.3), :(

gkatiforis:
All the above mentioned suggestions have been resolved. I'm always at your disposal for any additional comment. Thank you.

gkatiforis:
Conflicts has been resolved. 

Moved the building of callbacks to separate method `EventEngine.buildCallbacks` and call it after the initialization of `MetamodelImplementor`  because it is required to get the element type of the `CollectionType` using `CollectionType.getElementType(SessionFactoryImplementor factory)`. 

Please suggest a better way, if any.

gavinking:
Collections have a completely different lifecycle that's somewhat independent of the owning entity, so what's proposed here doesn't make much sense, and would only result in performance problems.

PR #3219, opened 2020-02-05
Fix typos in AbstractEntityPersister's generateSQLDeletStrings() method

hyeonisism:
Fix typo.


sebersole:
Will cause conflicts integrating into 6.0 branch though.  I'll go ahead and apply it, but we really should start considering whether such minor changes should get applied to master any longer.  I'll bring this up in our next discussion

PR #3220, opened 2020-02-06
Wip/6.0 Disallow FQN entity name in JPA compliance mode

NathanQingyangXu:
Testing case added

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
@@ -167,7 +167,6 @@ import static java.util.Collections.singletonList;
 import static org.hibernate.grammars.hql.HqlParser.IDENTIFIER;
 import static org.hibernate.query.TemporalUnit.*;
 import static org.hibernate.type.descriptor.DateTimeUtils.DATE_TIME;
-import static org.hibernate.type.descriptor.DateTimeUtils.OFFSET_DATE_TIME;
 import static org.hibernate.type.spi.TypeConfiguration.isJdbcTemporalType;
 
 /**
@@ -264,7 +263,7 @@ public class SemanticQueryBuilder extends HqlParserBaseVisitor implements SqmCre
 
 
 	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-	// To-level statements
+	// Top-level statements
 
 	@Override
 	public SqmSelectStatement visitSelectStatement(HqlParser.SelectStatementContext ctx) {
@@ -828,7 +827,7 @@ public class SemanticQueryBuilder extends HqlParserBaseVisitor implements SqmCre
 		if ( entityReference == null ) {
 			throw new UnknownEntityException( "Could not resolve entity name [" + entityName + "] as DML target", entityName );
 		}
-
+		checkFQNEntityNameJpaComplianceViolationIfNeeded( entityName, entityReference );
 		return entityReference;
 	}
 
@@ -894,6 +893,8 @@ public class SemanticQueryBuilder extends HqlParserBaseVisitor implements SqmCre
 		final EntityDomainType entityDescriptor = getCreationContext().getJpaMetamodel().resolveHqlEntityReference(
 				name
 		);
+
+		checkFQNEntityNameJpaComplianceViolationIfNeeded( name, entityDescriptor );
 
 		if ( entityDescriptor instanceof SqmPolymorphicRootDescriptor ) {
 			if ( getCreationOptions().useStrictJpaCompliance() ) {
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
```diff
@@ -167,7 +167,6 @@
 import static org.hibernate.grammars.hql.HqlParser.IDENTIFIER;
 import static org.hibernate.query.TemporalUnit.*;
 import static org.hibernate.type.descriptor.DateTimeUtils.DATE_TIME;
-import static org.hibernate.type.descriptor.DateTimeUtils.OFFSET_DATE_TIME;
```
NathanQingyangXu: found unused import. maybe we can include associated rule into check style.

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
```diff
@@ -264,7 +263,7 @@ else if ( ctx.deleteStatement() != null ) {
 
 
 	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-	// To-level statements
+	// Top-level statements
```
NathanQingyangXu: not too sure, however

hibernate-core/src/main/java/org/hibernate/query/sqm/StrictJpaComplianceViolation.java
```diff
@@ -26,6 +26,7 @@
 		SUBQUERY_ORDER_BY( "use of ORDER BY clause in subquery" ),
 		LIMIT_OFFSET_CLAUSE( "use of LIMIT/OFFSET clause" ),
 		IDENTIFICATION_VARIABLE_NOT_DECLARED_IN_FROM_CLAUSE( "use of an alias not declared in the FROM clause" ),
+		FQN_ENTITY_NAME( "use of fully-qualified class name for entity name" ),
```
NathanQingyangXu: open to verbiage suggestion

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
```diff
@@ -832,12 +831,18 @@ private SortOrder interpretSortOrder(String value) {
...
-			entityName = creationContext.getJpaMetamodel().qualifyImportableName( entityName );
-			reference = creationContext.getJpaMetamodel().entity( entityName );
+			String importName = creationContext.getJpaMetamodel().qualifyImportableName( entityName );
+			if ( entityName.equals( importName ) && getCreationOptions().useStrictJpaCompliance() ) {
+				throw new StrictJpaComplianceViolation(
+						"Encountered FQN entity name [" + entityName + "], but strict JPQL compliance was requested",
+						StrictJpaComplianceViolation.Type.FQN_ENTITY_NAME
+				);
+			}
+			reference = creationContext.getJpaMetamodel().entity( importName );
```
sebersole: The check here should be based on `reference.getName()` like I said on the other one.  

Why?  Well conceptually it is more correct.  And practically it would allow the exception message to include the entity name they should use in "compliance mode".  Right now, your error message does not tell them the corrective action.

Also, `getCreationOptions().useStrictJpaCompliance()` is a more performant check compared to the String comparison - it should come first.

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
```diff
@@ -894,6 +899,13 @@ public SqmRoot visitPathRoot(HqlParser.PathRootContext ctx) {
 		final EntityDomainType entityDescriptor = getCreationContext().getJpaMetamodel().resolveHqlEntityReference(
 				name
 		);
+		
+		if ( name.equals( entityDescriptor.getHibernateEntityName() ) && getCreationOptions().useStrictJpaCompliance() ) {
+			throw new StrictJpaComplianceViolation(
+					"Encountered FQN entity name [" + name + "], but strict JPQL compliance was requested",
+					StrictJpaComplianceViolation.Type.FQN_ENTITY_NAME
+			);
+		}
```
sebersole: This is close, but the check really should be:

````
if ( getCreationOptions().useStrictJpaCompliance()
        && ! name.equals( entityDescriptor.getName() ) {
    ...
}
````

**Comment by NathanQingyangXu** [comment]
thanks for the guidance. I modified accordingly and I think you can take a new look. I will add testing cases when I have more time.

File: hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/JpaComplianceDisallowFQNTests.java [hunk 6]
```diff
@@ -0,0 +1,43 @@
...
+		interpretSelect( String.format( "update %s set numberOfToes = 0", Person.class.getName() ) );
+	}
+	
+	@Test
+	@ExpectedException(StrictJpaComplianceViolation.class)
+	public void testDelete() {
+		interpretSelect( String.format( "delete %s", Person.class.getName() ) );
+	}
+	
+}
```
NathanQingyangXu: A little bit surprised `@Nested` is not allowed (seems `@Test` is the only expected annotation on top level)

sebersole:
Thanks @NathanQingyangXu

PR #3222, opened 2020-02-06
HHH-13843: make schema migration faster

Francoisvandelft:
A slowdown was noticed after upgrading to hibernate 5.4 during investigation a lot of calls for retrieving  foreignkeys from the database which caused a lot of separate calls for retrieving column-information from to the JDBC-driver.

By retrieving the DatabaseInformation for the whole nameSpace during schemaMigration, a nice speedup (and reduction in SQL-calls) can be achieved.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/DatabaseInformationCachedImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/DatabaseInformationCachedImpl.java
@@ -1,0 +1,60 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.tool.schema.extract.internal;
+
+import java.sql.SQLException;
+
+import org.hibernate.boot.model.relational.Namespace;
+import org.hibernate.boot.model.relational.QualifiedTableName;
+import org.hibernate.engine.jdbc.env.spi.JdbcEnvironment;
+import org.hibernate.resource.transaction.spi.DdlTransactionIsolator;
+import org.hibernate.service.ServiceRegistry;
+import org.hibernate.tool.schema.extract.spi.NameSpaceTablesInformation;
+import org.hibernate.tool.schema.extract.spi.TableInformation;
+
+/**
+ * Will cache all database information by reading the tableinformation from the name space in one call.
+ * Schema migration becomes much faster when this object is used.
+ * NOTE: superclass already caches the sequence information, so perhaps cache can also be coded in superlcass
+ * @author francois
+ */
+public class DatabaseInformationCachedImpl extends DatabaseInformationImpl {
+	private NameSpaceTablesInformation defaultNameSpaceTablesInformation = null;
+	private final Namespace defaultNamespace;
+
+	/**
+	 *
+	 * @param serviceRegistry        ServiceRegistry
+	 * @param jdbcEnvironment        JdbcEnvironment
+	 * @param ddlTransactionIsolator DdlTransactionIsolator
+	 * @param defaultNamespace       NameSpace
+	 */
+	public DatabaseInformationCachedImpl(
+		ServiceRegistry serviceRegistry,
+		JdbcEnvironment jdbcEnvironment,
+		DdlTransactionIsolator ddlTransactionIsolator, Namespace defaultNamespace)
+		throws SQLException {
+		super(serviceRegistry, jdbcEnvironment, ddlTransactionIsolator, defaultNamespace.getName());
+		this.defaultNamespace = defaultNamespace;
+	}
+
+	@Override
+	public TableInformation getTableInformation(QualifiedTableName qualifiedTableName) {
+
+		if (defaultNameSpaceTablesInformation == null) {
+			// Load table information from the whole space in one go (expected to be used in case of schemaMigration, so (almost) all
+			// tables are likely to be retrieved)
+			defaultNameSpaceTablesInformation = getTablesInformation(defaultNamespace);
+		}
+		TableInformation tableInformation = defaultNameSpaceTablesInformation.getTableInformation(qualifiedTableName.getTableName().getText());
+		if (tableInformation != null) {
+			return tableInformation;
+		}
+		return super.getTableInformation(qualifiedTableName);  // result of this call can of course also be cached, does not seem to be in scope now
+	}
+
+}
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/DatabaseInformationCachedImpl.java
```diff
@@ -0,0 +1,60 @@
...
+import org.hibernate.tool.schema.extract.spi.TableInformation;
+
+/**
+ * Will cache all database information by reading the tableinformation from the name space in one call.
+ * Schema migration becomes much faster when this object is used.
+ * NOTE: superclass already caches the sequence information, so perhaps cache can also be coded in superlcass
+ * @author francois
+ */
+public class DatabaseInformationCachedImpl extends DatabaseInformationImpl {
+	private NameSpaceTablesInformation defaultNameSpaceTablesInformation = null;
```
NathanQingyangXu: I am pretty sure we don't need to set `null` above. It has been set during object creation by JVM already.

hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/DatabaseInformationCachedImpl.java
```diff
@@ -0,0 +1,60 @@
...
+	 * @param ddlTransactionIsolator DdlTransactionIsolator
+	 * @param defaultNamespace       NameSpace
+	 */
+	public DatabaseInformationCachedImpl(
+		ServiceRegistry serviceRegistry,
+		JdbcEnvironment jdbcEnvironment,
+		DdlTransactionIsolator ddlTransactionIsolator, Namespace defaultNamespace)
+		throws SQLException {
+		super(serviceRegistry, jdbcEnvironment, ddlTransactionIsolator, defaultNamespace.getName());
+		this.defaultNamespace = defaultNamespace;
```
NathanQingyangXu: Hibernate has specific code style, especially we use empty char aggressively. We have a ide style repo even: https://github.com/hibernate/hibernate-ide-codestyles.

You might need to refer to existing code for the style issue, especially you need to insert more space chars.

hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/DatabaseInformationCachedImpl.java
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	@Override
+	public TableInformation getTableInformation(QualifiedTableName qualifiedTableName) {
+
+		if (defaultNameSpaceTablesInformation == null) {
+			// Load table information from the whole space in one go (expected to be used in case of schemaMigration, so (almost) all
+			// tables are likely to be retrieved)
+			defaultNameSpaceTablesInformation = getTablesInformation(defaultNamespace);
+		}
```
NathanQingyangXu: Wanna confirm the method won't be run in multiple threads, otherwise the above code would be subject to well-known race condition as explained in Effective Java's chapter on 'lazy initialization'.

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/AbstractSchemaMigrator.java
```diff
@@ -139,6 +135,8 @@ public void doMigration(Metadata metadata, ExecutionOptions options, TargetDescr
 		}
 	}
 
+	protected abstract DatabaseInformation getDatabaseInformation(DdlTransactionIsolator ddlTransactionIsolator, Namespace namespace);
```
NathanQingyangXu: Hibernate strives to maintain backward compatibility so new method is supposed to be introduced via JDK8's default interface method mechanism (our client might be compiling based on the public methods we have provided and this new abstract method might break all their existing code).

NathanQingyangXu: Rather than introducing a non-compatible abstract method, is it possible to add caching mechanism in `Helper.buildDatabaseInformation()` internally? Seems more straightforward and safe.

NathanQingyangXu:
Seems a good idea overall. It would be great to add some internal cache mechanism without breaking backward compatibility.

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/AbstractSchemaMigrator.java
dreab8: it is an internal so it is not a problem

Francoisvandelft: I will update the pull request, and make this method none-abstract, which will implement the "current" behavior. "				final DatabaseInformation databaseInformation = Helper.buildDatabaseInformation(
						tool.getServiceRegistry(),
						ddlTransactionIsolator,
						metadata.getDatabase().getDefaultNamespace().getName()
				);"

and I will overwrite this method in the "GroupedSchemaMigratorImpl"

As a result the implementation of "IndividuallySchemaMigratorImpl" and "CheckForExistingForeignKeyTest" are no longer affected by this pull request.

hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/DatabaseInformationCachedImpl.java
Francoisvandelft: fixed

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/AbstractSchemaMigrator.java
NathanQingyangXu: > it is an internal so it is not a problem

Thanks. I forgot it is in 'internal' package.

Sanne:
Very interesting, thanks! @dreab8 could you review this please, looks like the previous comments have been addressed?

gsmet:
My recollection is that back in the time some databases didn't support that. And also, for some one approach was faster and for others slower.

So we need to be a bit careful about that.

Francoisvandelft:
Regarding previous remark, in essence all the code does is calling a method that is already present in the existing DatabaseInformation, so.... me expectation is that the databases already support this, and if not that there is code in place that implements the behavior.
This also means that if the call in the parent-class is not efficient, the performance gain will be smaller.

sebersole:
Yes, originally we had tried what is suggested here.  The problem is large databases.  Pulling all this data back at once can improve performance for smallish, but there is a tipping point at which point the performance gets worse.

Francoisvandelft:
Another approach can be that programmers can replace the default DatabaseInformation-object with their own DatabaseInformation-object. If we can specify that in the AvailableSettings-object the programmers have an option to tweak the code.

beikov:
Thanks for the PR. I expanded on the idea in https://github.com/hibernate/hibernate-orm/pull/11042

PR #3223, opened 2020-02-06
HHH-13847 - Work for the removal of duplicate method

thegeekyasian:
Constraint.java currently has two duplicate methods. Both the methods return the same value i.e. columns.iterator(). It has been marked as a TODO and has been resolved in this commit.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/mapping/Constraint.java
+++ b/hibernate-core/src/main/java/org/hibernate/mapping/Constraint.java
@@ -151,12 +151,8 @@ public abstract class Constraint implements RelationalModel, Exportable, Seriali
 	public Column getColumn(int i) {
 		return  columns.get( i );
 	}
-	//todo duplicated method, remove one
-	public Iterator<Column> getColumnIterator() {
-		return columns.iterator();
-	}
 
-	public Iterator<Column> columnIterator() {
+	public Iterator<Column> getColumnIterator() {
 		return columns.iterator();
 	}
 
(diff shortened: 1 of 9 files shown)


hibernate-core/src/main/java/org/hibernate/mapping/Constraint.java
```diff
@@ -151,12 +151,8 @@ public int getColumnSpan() {
 	public Column getColumn(int i) {
 		return  columns.get( i );
 	}
-	//todo duplicated method, remove one
-	public Iterator<Column> getColumnIterator() {
-		return columns.iterator();
-	}
```
NathanQingyangXu: As in another PR you created, we need to consider backward compatibility. Again, it might be a good idea to merge into wip/v6.0.
The above deletion might break our client's code, IMHO.

thegeekyasian: Yes, this was done as idea to include it in wip/v6.0.

NathanQingyangXu: Then you might need to change the merging destination accordingly.

thegeekyasian: My bad, that has been done. Thank you for pointing it out.

thegeekyasian:
Closing this PR, as it has been done in a separate PR with other issues. Link attached.

https://github.com/hibernate/hibernate-orm/pull/3229

PR #3225, opened 2020-02-07
HHH-13846 Fix TemplateRenderer's param index parsing bug

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13846

Initially I thought this was an easy fix, but ...

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/function/TemplateRenderer.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/function/TemplateRenderer.java
@@ -37,8 +37,8 @@ public class TemplateRenderer {
 	public TemplateRenderer(String template) {
 		this.template = template;
 
-		final List<String> chunkList = new ArrayList<String>();
-		final List<Integer> paramList = new ArrayList<Integer>();
+		final List<String> chunkList = new ArrayList<>();
+		final List<Integer> paramList = new ArrayList<>();
 		final StringBuilder chunk = new StringBuilder( 10 );
 		final StringBuilder index = new StringBuilder( 2 );
 
@@ -47,22 +47,28 @@ public class TemplateRenderer {
 		while ( i < len ) {
 			char c = template.charAt( i );
 			if ( c == '?' ) {
-				chunkList.add( chunk.toString() );
-				chunk.delete( 0, chunk.length() );
-
 				while ( ++i < template.length() ) {
 					c = template.charAt( i );
 					if ( Character.isDigit( c ) ) {
+						if ( chunk.length() > 0 ) {
+							chunkList.add( chunk.toString() );
+							chunk.setLength( 0 );
+						}
 						index.append( c );
 					}
 					else {
+						if ( index.length() == 0 ) {
+							chunk.append( '?' );
+						}
 						chunk.append( c );
 						break;
 					}
 				}
 
-				paramList.add( Integer.valueOf( index.toString() ) );
-				index.delete( 0, index.length() );
+				if ( index.length() > 0 ) {
+					paramList.add( Integer.valueOf( index.toString() ) );
+					index.setLength( 0 );
+				}
 			}
 			else {
 				chunk.append( c );
@@ -74,7 +80,7 @@ public class TemplateRenderer {
 			chunkList.add( chunk.toString() );
 		}
 
-		chunks = chunkList.toArray( new String[chunkList.size()] );
+		chunks = chunkList.toArray( new String[0] );
 		paramIndexes = new int[paramList.size()];
 		for ( i = 0; i < paramIndexes.length; ++i ) {
 			paramIndexes[i] = paramList.get( i );
--- a/hibernate-core/src/main/java/org/hibernate/dialect/function/TemplateRenderer.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/function/TemplateRenderer.java
@@ -76,6 +76,9 @@ public class TemplateRenderer {
 			i++;
 		}
 
+		if ( template.endsWith( "?" ) ) {
+			chunk.append( "?" );
+		}
 		if ( chunk.length() > 0 ) {
 			chunkList.add( chunk.toString() );
 		}
--- a/hibernate-core/src/test/java/org/hibernate/test/dialect/function/TemplateRendererTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/dialect/function/TemplateRendererTest.java
@@ -1,0 +1,37 @@
+package org.hibernate.test.dialect.function;
+
+import java.util.Collections;
+
+import org.hibernate.dialect.function.TemplateRenderer;
+
+import org.hibernate.testing.TestForIssue;
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TemplateRendererTest extends BaseCoreFunctionalTestCase {
+	
+	@Test
+	@TestForIssue( jiraKey = "HHH-13846" )
+	public void testParamIndexWithoutDigit() {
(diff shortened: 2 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/function/TemplateRenderer.java
```diff
@@ -74,7 +76,7 @@ public TemplateRenderer(String template) {
 			chunkList.add( chunk.toString() );
 		}
 
-		chunks = chunkList.toArray( new String[chunkList.size()] );
+		chunks = chunkList.toArray( new String[0] );
```
NathanQingyangXu: As per https://stackoverflow.com/questions/174093/toarraynew-myclass0-or-toarraynew-myclassmylist-size, seems empty array approach performs better, let alone its better maintainability.

hibernate-core/src/main/java/org/hibernate/dialect/function/TemplateRenderer.java
```diff
@@ -61,8 +61,10 @@ public TemplateRenderer(String template) {
 					}
 				}
 
-				paramList.add( Integer.valueOf( index.toString() ) );
-				index.delete( 0, index.length() );
+				if ( index.length() > 0 ) {
+					paramList.add( Integer.valueOf( index.toString() ) );
+				}
+				index.setLength( 0 );
```
NathanQingyangXu: I think 'setLength()' is a little bit better in terms of both performance and code readability. See https://stackoverflow.com/questions/5192512/how-can-i-clear-or-empty-a-stringbuilder

gavinking:
@NathanQingyangXu but does this really completely fix the problem? With this change don't you get a mismatch between the length of `chunkList` and the length of the `paramList`?

gavinking:
TBH, I'm not even sure I think this is a bug. If there's a strong need to allow literal `?` in patterns, I would introduce a well-defined syntax for it (`??` perhaps) but really I would say the main problem here is just the exception:

    java.lang.NumberFormatException: For input string: ""

If we caught that and produced a better message, I think we would be good...

NathanQingyangXu:
> @NathanQingyangXu but does this really completely fix the problem? With this change don't you get a mismatch between the length of `chunkList` and the length of the `paramList`?

Good catch! Yeah, it is much more complicated than I've thought initially. If '?' doesn't follow digit, it should be included in some chunk. I updated the code to make that happen, but I feel it is becoming more and more complicated...

NathanQingyangXu:
> TBH, I'm not even sure I think this is a bug. If there's a strong need to allow literal `?` in patterns, I would introduce a well-defined syntax for it (`??` perhaps) but really I would say the main problem here is just the exception:
> 
> ```
> java.lang.NumberFormatException: For input string: ""
> ```
> 
> If we caught that and produced a better message, I think we would be good...

Good point! Yeah, this seems a really edgy case and it seems not worthwhile to fix it ending up with making the code complicated (there is no foolproof solution following the existing direction).

I enriched the testing case but I am afraid it might not be worthwhile. @sebersole , what do you think?

hibernate-core/src/main/java/org/hibernate/dialect/function/TemplateRenderer.java
```diff
@@ -47,22 +47,28 @@ public TemplateRenderer(String template) {
...
 						break;
 					}
 				}
 
-				paramList.add( Integer.valueOf( index.toString() ) );
-				index.delete( 0, index.length() );
+				if ( index.length() > 0 ) {
+					paramList.add( Integer.valueOf( index.toString() ) );
+					index.setLength( 0 );
+				}
```
NathanQingyangXu: The basic idea is simple. If '?' is not followed by digit(s), it should be teated as normal chunk char. However, the code seems unnecessarily complicated and we might refactor the logic by simple RE processing instead.

hibernate-core/src/test/java/org/hibernate/test/dialect/function/TemplateRendererTest.java
```diff
@@ -0,0 +1,37 @@
...
+				+ "    REPLACE( "
+				+ "        TRANSLATE( "
+				+ "            LOWER(?1), "
+				+ "            'áàãâäéèêëíìîïóòõôöúùûüç', "
+				+ "            'aaaaaeeeeiiiiooooouuuuc' "
+				+ "        ), "
+				+ "        ' ', "
+				+ "        '' "
+				+ "    ), "
+				+ "    '[\\u2000-\\u206F\\u2E00-\\u2E7F\\\\''!\"#$%&()*+,\\-.\\/:;<=>?@\\[\\]^_`{|}~]', " // -- '?'' causes the problem
```
NathanQingyangXu: if the above '?' is followed by digit, it should not be treated as param, so I doubt whether the fixing is worthwhile in the first place.

NathanQingyangXu:
> TBH, I'm not even sure I think this is a bug. If there's a strong need to allow literal `?` in patterns, I would introduce a well-defined syntax for it (`??` perhaps) but really I would say the main problem here is just the exception:
> 
> ```
> java.lang.NumberFormatException: For input string: ""
> ```
> 
> If we caught that and produced a better message, I think we would be good...

Yeah, I have to admit you have a good point. I found I am stuck somewhere. @sebersole is providing more meaningful exception in constructor is acceptable here?

NathanQingyangXu:
This bugfix is more complicated than I originally thought. For that reason, I decided to close the PR and create a new one when I am ready. Thanks for excellent comments from @gavinking

mhagnumdw:
@NathanQingyangXu ,  I think this can solve:

```java
public TemplateRenderer(String template) {
    this.template = template;

    final List<String> chunkList = new ArrayList<String>();
    final List<Integer> paramList = new ArrayList<Integer>();
    final StringBuilder chunk = new StringBuilder( 10 );
    final StringBuilder index = new StringBuilder( 2 );

    int i = 0;
    final int len = template.length();
    while ( i < len ) {
        char c = template.charAt( i );
        if ( c == '?' ) { // Is it a parameter or literal? We will see ...

            while ( ++i < template.length() ) {
                c = template.charAt( i );
                if ( Character.isDigit( c ) ) {
                    index.append( c );
                }
                else {
                    if (index.length() > 0) { // the '?' represent an argument
                        chunkList.add( chunk.toString() );
                        chunk.delete( 0, chunk.length() );
                        
                        paramList.add( Integer.valueOf( index.toString() ) );
                        index.delete( 0, index.length() );
                    } else { // the '?' represent a literal
                        chunk.append( '?' );
                    }
                    chunk.append( c );
                    break;
                }
            }
            
        }
        else {
            chunk.append( c );
        }
        i++;
    }

    if ( chunk.length() > 0 ) {
        chunkList.add( chunk.toString() );
    }

    chunks = chunkList.toArray( new String[chunkList.size()] );
    paramIndexes = new int[paramList.size()];
    for ( i = 0; i < paramIndexes.length; ++i ) {
        paramIndexes[i] = paramList.get( i );
    }
}

```

Diff (before / after)

image

- I believe that if the '?' is followed by a digit should always be treated as a parameter
- I tested it for the test case you created and it worked as expected
- Unfortunately my resources are limited at the moment and therefore I have no way to run the Hibernate tests

What do you think?

NathanQingyangXu:
Thanks for following up. TBH, I think the second comment from Gavin King
makes sense and I don't think there is fool-proof fix following the current
direction. For that reason, I closed my PR. Is the issue really serious
without bypassing?
As to the code change, I think it is becoming unnecessarily complicated.
For example, if the template ends with '?', seems it will be ignored. I
made similar code changes in my closed PR, but I doubt it is correct
direction. Suppose there is a digit after the '?', it seems we cannot say
it is intended param. Maybe a more complete syntax design is needed.
That being said, you are welcome to create PR on top of my code. But I
would think more on the best solution.

PR #3226, opened 2020-02-07
HHH-13852 - Work for method renaming in BaseCoreFunctionalTestCase

thegeekyasian:
Currently, in BaseCoreFunctionalTestCase a method named getXmlFiles()has a TODO placed that suggests the method name to be changed to getOrmXmlFiles().

Since the method is a protected methods, there is a possibility it is being used by other clients. Hence, changing the method name in minor versions might break the code for clients.

Although, this long-awaited change can be done as an improvement in wip/v6.0


thegeekyasian:
Closing this PR as it has been done in a separate PR with other issues. Link attached,

https://github.com/hibernate/hibernate-orm/pull/3229

PR #3227, opened 2020-02-07
HHH-13853 - Pass the merged Integration settings and Persistence Unit properties to buildBootstrapServiceRegistry

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13853

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
@@ -196,8 +196,19 @@ public class EntityManagerFactoryBuilderImpl implements EntityManagerFactoryBuil
 			integrationSettings = Collections.emptyMap();
 		}
 
+		Map mergedIntegrationSettings = null;
+		Properties properties = persistenceUnit.getProperties();
+		if ( properties != null ) {
+			mergedIntegrationSettings = new HashMap( persistenceUnit.getProperties() );
+			mergedIntegrationSettings.putAll( integrationSettings );
+		}
+
 		// Build the boot-strap service registry, which mainly handles class loader interactions
-		final BootstrapServiceRegistry bsr = buildBootstrapServiceRegistry( integrationSettings, providedClassLoader, providedClassLoaderService);
+		final BootstrapServiceRegistry bsr = buildBootstrapServiceRegistry(
+				mergedIntegrationSettings != null ? mergedIntegrationSettings : integrationSettings,
+				providedClassLoader,
+				providedClassLoaderService
+		);
 
 		// merge configuration sources and build the "standard" service registry
 		final StandardServiceRegistryBuilder ssrBuilder = StandardServiceRegistryBuilder.forJpa( bsr );
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/BaseEntityManagerFunctionalTestCase.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/BaseEntityManagerFunctionalTestCase.java
@@ -80,7 +80,7 @@ public abstract class BaseEntityManagerFunctionalTestCase extends BaseUnitTestCa
 		afterEntityManagerFactoryBuilt();
 	}
 
-	private PersistenceUnitDescriptor buildPersistenceUnitDescriptor() {
+	protected PersistenceUnitDescriptor buildPersistenceUnitDescriptor() {
 		return new TestingPersistenceUnitDescriptorImpl( getClass().getSimpleName() );
 	}
 
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/integrationprovider/IntegrationProviderSettingByClassUsingPropertiesTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/integrationprovider/IntegrationProviderSettingByClassUsingPropertiesTest.java
@@ -1,0 +1,57 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.jpa.test.integrationprovider;
+
+import java.util.List;
+import java.util.Properties;
+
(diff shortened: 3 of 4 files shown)


vladmihalcea:
This is a follow-up for the [HHH-13614](https://hibernate.atlassian.net/browse/HHH-13614) issue. I noticed we can only pass an `IntegrationProvider` via the integration settings when bootstrapping the JPA provider.

However, if the user supplies the `hibernate.integrator_provider` setting via the Persistence Unit properties (e.g., `persistence.xml`), the property will be ignored. Hence, this PR which fixes the issue.

vladmihalcea:
@dreab8 If you have time, can you review this issue? Thanks.

nandorholozsnyak:
Is there any progress on the review? :)

hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
```diff
@@ -196,8 +196,19 @@ private EntityManagerFactoryBuilderImpl(
 			integrationSettings = Collections.emptyMap();
 		}
 
+		Map mergedIntegrationSettings = null;
+		Properties properties = persistenceUnit.getProperties();
+		if ( properties != null ) {
+			mergedIntegrationSettings = new HashMap( integrationSettings );
+			mergedIntegrationSettings.putAll( persistenceUnit.getProperties() );
+		}
+
```
dreab8: I do not think it is so simple as adding to a map the `integrationSettings` and then `persistenceUnit.getProperties()`, see `MergedSettings` creation at https://github.com/hibernate/hibernate-orm/pull/3227/files#diff-c1662f25d34183e3651ada18824bbd27R216

vladmihalcea: @dreab8 The goal of this change is to allow the Persistence Unit properties to be passed to the `BootstrapServiceRegistry` creation process.

The `MergedSettings` object requires the `BootstrapServiceRegistry` instance, hence this change. So, you cannot create a `MergedSettings` without a `BootstrapServiceRegistry`. 

But the goal of this PR is not to provide all possible settings to the `BootstrapServiceRegistry`, as the  `MergedSettings does. You only need the JPA properties as they are the typical way of configuring a JPA application.

Also, this change does not impact how the `MergedSettings` is used. It just provides more info to the `BootstrapServiceRegistry`, and the temporary `mergedIntegrationSettings` is never used after the `BootstrapServiceRegistry` instance gets created.

All in all, I think this is safe to be integrated, having a test case that proves the fix, and all other tests have run successfully too.

dreab8: I didn;t want to say that you have to use `MergedSettings` but that it is an example showing that the logic is not so easy in case of values that are presents in both `integrationSettings` and `persistenceUnit.getProperties()` .

In your solution the properties of `PersistenceUnitDescriptor` override the `integrationSettings` ones and I think this is not correct.

vladmihalcea: That's a good observation. I could initialize the new `Map` using the `persistenceUnit.getProperties()` and call `putAll` using the `integrationSettings`, as these are provided at bootstrap time, and should take precedence.

Should I refactor it like this?

vladmihalcea: I rebased the changes according to your suggestion.

sebersole: Awesome!

dreab8: thanks @vladmihalcea

sebersole:
Going to apply this.

hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
```diff
@@ -196,8 +196,19 @@ private EntityManagerFactoryBuilderImpl(
 			integrationSettings = Collections.emptyMap();
 		}
 
+		Map mergedIntegrationSettings = null;
+		Properties properties = persistenceUnit.getProperties();
+		if ( properties != null ) {
+			mergedIntegrationSettings = new HashMap( persistenceUnit.getProperties() );
```
NathanQingyangXu: I think at least we can reuse `properties` at the above statement to improve code readability a little bit. 
```
if (properties != null ) {
    mergedIntegrationSettings = new HashMap( properties );
    ... ...
}
```

hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
sebersole: The more complex merging that Andrea is talking about (I think) is more about, e.g., JDBC config from integrationSettings ought to override not just JDBC config from PU properties - it should also override datasource config from the PU properties.  I can't say whether those more complete overriding is needed here - probably not

hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
sebersole: +1 good catch

sebersole: @NathanQingyangXu do you want to create a PR for that since it was your catch?

NathanQingyangXu: There is another subtle bug that if the passed 'integrationSettings' is null the entries from PU might be added to `Collections.emptyMap()` ending up with `UnsupportedOperationException`.
Yeah, I am gonna create a PR to fix all such issues.

NathanQingyangXu: PR created at: https://github.com/hibernate/hibernate-orm/pull/3246. It might be a little bit aggressive, but I do think we can improve code readability besides fixing a subtle potential bug.

PR #3228, opened 2020-02-07
HHH-13855 - Remove unnecessary delared variable JtaManager in HibernatePersistenceProviderAdaptor

thegeekyasian:
In the HibernatePersistenceProviderAdaptor, there is an unnecessary declaration of JtaManager. The sole purpose of having this variable is only to use in the overridden method named InjectJtaManager.

This was be improved by removing the JtaManager as there is no use of the variable, as of now at least.


Sanne:
ok, thanks!

PR #3229, opened 2020-02-09
HHH-13856 - Long-awaited TODOs done as an improvement

thegeekyasian:
1. TODO: Remove duplicate method from ConstraintConstaint.java currently has two duplicate methods. 1. getColumnIterator()2. columnIterator()

Both the methods return the same value i.e. column.iterator().

One of them needs to be removed in order to reduce and clean the duplication. 

2. TODO: Change method name for getXmlFiles in BaseCoreFunctionalTestCaseCurrently, in BaseCoreFunctionalTestCase a method named getXmlFiles()has a TODO placed that suggests the method name to be changed to getOrmXmlFiles().

Since the method has a protected scope, there is a possibility it is being used by other clients. Hence, changing the method name in minor versions might break the code for clients.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/unique/DB2UniqueDelegate.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/unique/DB2UniqueDelegate.java
@@ -35,7 +35,7 @@ public class DB2UniqueDelegate extends DefaultUniqueDelegate {
 					dialect,
 					uniqueKey.getName(),
 					uniqueKey.getTable(),
-					uniqueKey.columnIterator(),
+					uniqueKey.getColumnIterator(),
 					uniqueKey.getColumnOrderMap(),
 					true,
 					metadata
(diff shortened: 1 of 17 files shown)


sebersole:
6.0 is not a minor release, so no worries there.  

Would you be able to also create a PR against master that deprecates these 2 methods you dropped?

FWIW, on `Constraint`, I'd actually be more inclined to keep `#columnIterator` rather than `#getColumnIterator` -

1. `#columnIterator` is the original one
2. minor design thing, but I try to avoid using getter in the Java Bean sense when the thing should not be used as a bean attribute.

thegeekyasian:
Thank you for reviewing this. 

Sure, I'll create the PR against the master deprecating these methods. 

Regarding `getColumnIterator`, I kept it in order to keep it consistent with other methods in the class. If that's a constraint, should I bring back the `columnIterator` and remove the `getColumnIterator` then?

sebersole:
> Regarding `getColumnIterator`, I kept it in order to keep it consistent with other methods in the class. If that's a constraint, should I bring back the `columnIterator` and remove the `getColumnIterator` then?

It's not a big deal.

PR #3230, opened 2020-02-10
HHH-13857 - Avoid lazy initialization when obtaining persistent class

flozano:
Current utility code in Hibernate class forces initialization of the
proxies just to obtain the target persistent class that is used, because
it uses `getImplementation()` method of LazyInitializer.

There is a method in LazyInitializer that can do the same without
initialization (`getPersistentClass()`), so I think it should be used instead.

I created HHH-13857 to track this.


beikov:
Such improvements won't be merged into 5.6 anymore. Let's focus on 6.3/main with https://github.com/hibernate/hibernate-orm/pull/6808

PR #3231, opened 2020-02-10
Wip/6.0 Remove an unnoticed negation operator in DefaultSaveOrUpdateEventListener

NathanQingyangXu:
This is an issue I've found long time ago. I risked exposing my ignorance to create a PR targeting v6 to have it fixed. Seems it is not noticed simply because it is not used often.


sebersole:
I'm inclined to find out why this code is there if it is never triggered.  But the change is safe as it goes.  To verify I made the same change in master and the tests passed.

NathanQingyangXu:
> I'm inclined to find out why this code is there if it is never triggered. But the change is safe as it goes. To verify I made the same change in master and the tests passed.

Thanks. Actually this is the first bug I've found and I even created my first PR for it without ticket. Later on I know better how to create PR to have some issue fixed, but now I am happy this bug is gone forever.

PR #3233, opened 2020-02-12
Wip/6.0 HHH-13877 Make`@SortNatural` default for SortedSet and SortedMap

NathanQingyangXu:
https://trello.com/c/up8I5oIi/58-support-for-sorted-set-and-sorted-map-without-sortnatural-or-sortcomparator

https://hibernate.atlassian.net/browse/HHH-13877

testing case added

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
@@ -663,14 +663,6 @@ public abstract class CollectionBinder {
 			}
 		}
 
-		if ( isSortedCollection ) {
-			if ( ! hadExplicitSort && !hadOrderBy ) {
-				throw new AnnotationException(
-						"A sorted collection must define an ordering or sorting : " + safeCollectionRole()
-				);
-			}
-		}
-
 		collection.setSorted( isSortedCollection || hadExplicitSort );
 
 		if ( comparatorClass != null ) {
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/metamodel/mapping/collections/MapOperationTests.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/metamodel/mapping/collections/MapOperationTests.java
@@ -47,11 +47,14 @@ public class MapOperationTests {
 					entityContainingMaps.addBasicByBasic( "someKey", "someValue" );
 					entityContainingMaps.addBasicByBasic( "anotherKey", "anotherValue" );
 
-					entityContainingMaps.addSortedBasicByBasic( "key1", "value1" );
 					entityContainingMaps.addSortedBasicByBasic( "key2", "value2" );
+					entityContainingMaps.addSortedBasicByBasic( "key1", "value1" );
 
 					entityContainingMaps.addSortedBasicByBasicWithComparator( "kEy1", "value1" );
 					entityContainingMaps.addSortedBasicByBasicWithComparator( "KeY2", "value2" );
+
+					entityContainingMaps.addSortedBasicByBasicWithSortNaturalByDefault( "key2", "value2" );
+					entityContainingMaps.addSortedBasicByBasicWithSortNaturalByDefault( "key1", "value1" );
 
 					entityContainingMaps.addBasicByEnum( EnumValue.ONE, "one" );
 					entityContainingMaps.addBasicByEnum( EnumValue.TWO, "two" );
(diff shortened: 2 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
```diff
@@ -663,14 +663,6 @@ else if ( comparatorSort != null ) {
...
 		}
 
-		if ( isSortedCollection ) {
-			if ( ! hadExplicitSort && !hadOrderBy ) {
-				throw new AnnotationException(
-						"A sorted collection must define and ordering or sorting : " + safeCollectionRole()
-				);
-			}
-		}
-
```
NathanQingyangXu: I manually compared to explicit `@SortNatural` case and found everything is the same(only `hadExplicitSort` values are different but it is used internally in this method and its main usage is the above code snippet), so removing the above code snippet seems enough to achieve the goal of this PR, unless I were wrong.

hibernate-core/src/test/java/org/hibernate/orm/test/metamodel/mapping/collections/SortNaturalByDefaultTests.java
```diff
@@ -0,0 +1,166 @@
...
+@SessionFactory
+public class SortNaturalByDefaultTests {
+	
+	@Test
+	void test(SessionFactoryScope scope) {
+		scope.inTransaction(
+			session -> {
+				final Person person = session.createQuery( "select p from Person p", Person.class ).getSingleResult();
+				final SortedSet<Phone> phones = person.getPhones();
+				assertThat( phones.size(), is( 3 ) );
```
sebersole: We should also assert that the phones are in fact sorted in proper order.

sebersole:
I need to look through this some more.  

However, this would be a new feature or improvement since we did not previously support this.  New features/improvements ought to have a corresponding Jira - the reason being it will be easier to consolidate all new features and improvements when we get to CRs and eventually Final

NathanQingyangXu:
> I need to look through this some more.
> 
> However, this would be a new feature or improvement since we did not previously support this. New features/improvements ought to have a corresponding Jira - the reason being it will be easier to consolidate all new features and improvements when we get to CRs and eventually Final

hibernate-core/src/test/java/org/hibernate/orm/test/metamodel/mapping/collections/SortNaturalByDefaultTests.java
NathanQingyangXu: Updated. Originally I had thought we might end up with testing SortedSet or SortedMap. Testing case enriched enormously.

hibernate-core/src/test/java/org/hibernate/orm/test/metamodel/mapping/collections/MapOperationTests.java
```diff
@@ -47,8 +47,11 @@ public void createData(SessionFactoryScope scope) {
 					entityContainingMaps.addBasicByBasic( "someKey", "someValue" );
 					entityContainingMaps.addBasicByBasic( "anotherKey", "anotherValue" );
 
-					entityContainingMaps.addSortedBasicByBasic( "key1", "value1" );
 					entityContainingMaps.addSortedBasicByBasic( "key2", "value2" );
+					entityContainingMaps.addSortedBasicByBasic( "key1", "value1" );
```
NathanQingyangXu: It doesn't make sense to insert testing data with the natural order. Ideally we are supposed to populate big number of data to ensure the order is not maintained not by accident.

NathanQingyangXu:
> I need to look through this some more.
> 
> However, this would be a new feature or improvement since we did not previously support this. New features/improvements ought to have a corresponding Jira - the reason being it will be easier to consolidate all new features and improvements when we get to CRs and eventually Final

I created a ticket for this: https://hibernate.atlassian.net/browse/HHH-13877 and updated the PR accordingly.

NathanQingyangXu:
> Looked through this some more and approve the changes.
> 
> However because of other "code cleanup" type changes a few files here now have completely unnecessary conflicts which is the problem I was pointing out a few weeks ago.
> 
> Anyway, please rebase these changes on top of those cosmetic cleanups. Thanks

A lesson learned. Conflict resolved (mainly from another PR, but yeah, cosmetic code changes break other PRs easily) and force pushed.

sebersole:
Thanks @NathanQingyangXu 

I'll apply after the checks complete

PR #3234, opened 2020-02-12
HHH-13861: Expose the doWork() and doReturningWork() APIs on StatelessSession as well

dreab8:
https://hibernate.atlassian.net/browse/HHH-13861

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/stateless/StatelessDoWorkTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/stateless/StatelessDoWorkTest.java
@@ -1,0 +1,148 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.test.stateless;
+
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.util.List;
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+import org.hibernate.StatelessSession;
+import org.hibernate.dialect.H2Dialect;
+
+import org.hibernate.testing.RequiresDialect;
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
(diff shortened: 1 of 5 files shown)


dreab8:
Thanks @gbadner , I applied your suggestion

Sanne:
testing...

Sanne:
I think in this case we might not want to have those methods with a "default" implementation:

- we don't support custom implementation of `org.hibernate.StatelessSession` so adding the methods to the interface isn't going to break code.
- the defualt implementation isn't functional

hibernate-core/src/test/java/org/hibernate/test/stateless/StatelessDoWorkTest.java
```diff
@@ -0,0 +1,146 @@
...
+					(connection) -> {
+						PreparedStatement preparedStatement = connection.prepareStatement(
+								"SELECT NAME FROM TEST_ENTITY WHERE ID = ?" );
+						preparedStatement.setInt( 1, PERSISTED_TEST_ENTITY_ID );
+						ResultSet resultSet = preparedStatement.executeQuery();
+						String name = null;
+						if ( resultSet.next() ) {
+							name = resultSet.getString( 1 );
+						}
+						return name;
```
NathanQingyangXu: I know `PreparedStatement`, `ResultSet` will be closed automatically when `Connection` is closed, but it seems it is not difficult to make the code more robust as following
```
try ( PreparedStatement preparedStatement = connection.prepareStatement(
								"SELECT NAME FROM TEST_ENTITY WHERE ID = ?" ) ) {
       preparedStatement.setInt( 1, PERSISTED_TEST_ENTITY_ID );
						
      try (ResultSet resultSet = preparedStatement.executeQuery() ) {
          String name = null;
	  if ( resultSet.next() ) {
		name = resultSet.getString( 1 );
	  }
	  return name;
    }
}
```
ignore the comment if you think it is too trivial. TBH, I hesitated a lot to comment.

hibernate-core/src/test/java/org/hibernate/test/stateless/StatelessDoWorkTest.java
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	@Test
+	public void testDoWork() {
+		try (StatelessSession statelessSession = sessionFactory().openStatelessSession()) {
+			statelessSession.doWork(
+					(connection) -> {
+						PreparedStatement preparedStatement = connection.prepareStatement(
+								"DELETE FROM TEST_ENTITY " );
+						preparedStatement.execute();
```
NathanQingyangXu: Ignore this comment if you think it is too trivial:
```
try ( PreparedStatement preparedStatement = connection.prepareStatement(
								"DELETE FROM TEST_ENTITY " ) ) {
       preparedStatement.execute();
}
```

dreab8:
@NathanQingyangXu I applied you suggestion.

@Sanne I changed the 2 default methods as we discussed, now they throw an `UnsupportedOperationException` exception.

Thanks for the reviews.

Sanne:
merged. thanks all!

PR #3236, opened 2020-02-14
HHH-13865 : Change some private methods/classes in org.hibernate.id.enhanced.Optimizer implementations to be protected

gbadner:
https://hibernate.atlassian.net/browse/HHH-13865

This change makes it easier to add support for HiLoOptimizer in hibernate-rx.

I've asked Sanne to review this because I am a little concerned that this change could hurt performance.

Once we work out changes that are acceptable for extending HiLoOptimizer, I will create a different PR for the other "pooled" optimizers.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/id/enhanced/HiLoOptimizer.java
+++ b/hibernate-core/src/main/java/org/hibernate/id/enhanced/HiLoOptimizer.java
@@ -53,12 +53,78 @@ import org.jboss.logging.Logger;
 public class HiLoOptimizer extends AbstractOptimizer {
 	private static final Logger log = Logger.getLogger( HiLoOptimizer.class );
 
-	private static class GenerationState {
+	protected static class GenerationState {
+		private final int incrementSize;
 		private IntegralDataTypeHolder lastSourceValue;
 		private IntegralDataTypeHolder upperLimit;
 		private IntegralDataTypeHolder value;
-	}
 
+		/**
+		 * @param incrementSize - the increment size.
+		 * @throws HibernateException if {@code incrementSize < 1}.
+		 */
+		private GenerationState(int incrementSize) {
+			if ( incrementSize < 1 ) {
+				throw new HibernateException( "increment size cannot be less than 1" );
+			}
+			this.incrementSize = incrementSize;
+		}
+
+		/**
+		 * Has this {@link GenerationState} been initialized with a
+		 * "source" value (e.g., from the database).
+		 *
+		 * @return true, if it is initialized; false, otherwise.
+		 */
+		public final boolean isInitialized() {
+			return lastSourceValue != null;
+		}
+
+		/**
+		 * Is a new "source" value (e.g., from the database) required
+		 * in order to obtain the next value? A new source value is
+		 * required when {@link #isInitialized()} is {@code false},
+		 * or the current value has already reached the upper limit.
+		 * When {@code true} is returned, {@link #updateFromNewSourceValue(IntegralDataTypeHolder)}
+		 * must be called in order to obtain the next value.
+		 *
+		 * @return true, if a new source value is required in order
+		 *               to obtain the next value; false, otherwise.
+		 */
+		public final boolean requiresNewSourceValue() {
+			return ! isInitialized() || ! upperLimit.gt( value );
+		}
+
+		/**
+		 * Returns the next value, and increments the internal value
+		 * in preparation for the following value to be requested.
+		 *
+		 * @return the next value.
+		 * @throws HibernateException if {@link #requiresNewSourceValue()} is {@code true}.
+		 */
+		public final Number makeValueThenIncrement() {
+			if ( requiresNewSourceValue() ) {
+				throw new HibernateException( "Value is either uninitialized or has already reached the upper limit. GenerationState#updateFromNewSourceValue must be called before obtaining the next value." );
+			}
+			return value.makeValueThenIncrement();
+		}
+
+		/**
+		 * Update this GenerationState with a new source value (e.g., from the database).
+		 *
+		 * @param sourceValue - the value from the "source" of values (e.g., from the database);
+		 * @throws HibernateException if sourceValue < 1.
+		 */
+		public final void updateFromNewSourceValue(IntegralDataTypeHolder sourceValue) {
+			// Copy sourceValue, so caller cannot modify the internal state.
+			lastSourceValue = sourceValue.copy();
+
+			// upperLimit defines the upper end of the bucket values
+			upperLimit = lastSourceValue.copy().multiplyBy( incrementSize ).increment();
+			// initialize value to the low end of the bucket
+			value = upperLimit.copy().subtract( incrementSize );
+		}
+	}
 
 	/**
 	 * Constructs a HiLoOptimizer
@@ -80,33 +146,29 @@ public class HiLoOptimizer extends AbstractOptimizer {
 	public synchronized Serializable generate(AccessCallback callback) {
 		final GenerationState generationState = locateGenerationState( callback.getTenantIdentifier() );
 
-		if ( generationState.lastSourceValue == null ) {
+		if ( !generationState.isInitialized() ) {
 			// first call, so initialize ourselves.  we need to read the database
 			// value and set up the 'bucket' boundaries
-			generationState.lastSourceValue = callback.getNextValue();
-			while ( generationState.lastSourceValue.lt( 1 ) ) {
-				generationState.lastSourceValue = callback.getNextValue();
+
+			IntegralDataTypeHolder firstSourceValue = callback.getNextValue();
+			while ( firstSourceValue.lt( 1 ) ) {
+				firstSourceValue = callback.getNextValue();
 			}
-			// upperLimit defines the upper end of the bucket values
-			generationState.upperLimit = generationState.lastSourceValue.copy().multiplyBy( incrementSize ).increment();
-			// initialize value to the low end of the bucket
-			generationState.value = generationState.upperLimit.copy().subtract( incrementSize );
+			generationState.updateFromNewSourceValue( firstSourceValue );
 		}
-		else if ( ! generationState.upperLimit.gt( generationState.value ) ) {
-			generationState.lastSourceValue = callback.getNextValue();
-			generationState.upperLimit = generationState.lastSourceValue.copy().multiplyBy( incrementSize ).increment();
-			generationState.value = generationState.upperLimit.copy().subtract( incrementSize );
+		else if ( generationState.requiresNewSourceValue() ) {
+			generationState.updateFromNewSourceValue( callback.getNextValue() );
 		}
-		return generationState.value.makeValueThenIncrement();
+		return generationState.makeValueThenIncrement();
 	}
 
 	private GenerationState noTenantState;
 	private Map<String,GenerationState> tenantSpecificState;
 
-	private GenerationState locateGenerationState(String tenantIdentifier) {
+	protected GenerationState locateGenerationState(String tenantIdentifier) {
 		if ( tenantIdentifier == null ) {
 			if ( noTenantState == null ) {
-				noTenantState = new GenerationState();
+				noTenantState = new GenerationState( incrementSize );
 			}
 			return noTenantState;
 		}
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/id/enhanced/HiLoOptimizer.java
```diff
@@ -53,12 +53,78 @@
 public class HiLoOptimizer extends AbstractOptimizer {
 	private static final Logger log = Logger.getLogger( HiLoOptimizer.class );
 
-	private static class GenerationState {
+	protected static class GenerationState {
+		private final int incrementSize;
 		private IntegralDataTypeHolder lastSourceValue;
 		private IntegralDataTypeHolder upperLimit;
 		private IntegralDataTypeHolder value;
-	}
```
gbadner: The following changes involve callers referencing the new methods added to GenerationState, rather than accessing the fields directly.

An alternative is make the fields in GenerationState protected, instead of adding the methods. I really don't like that alternative though.

hibernate-core/src/main/java/org/hibernate/id/enhanced/HiLoOptimizer.java
```diff
@@ -53,12 +53,78 @@
...
+		 *
+		 * @param sourceValue - the value from the "source" of values (e.g., from the database);
+		 * @throws HibernateException if sourceValue < 1.
+		 */
+		public final void updateFromNewSourceValue(IntegralDataTypeHolder sourceValue) {
+			// Copy sourceValue, so caller cannot modify the internal state.
+			lastSourceValue = sourceValue.copy();
+
+			// upperLimit defines the upper end of the bucket values
+			upperLimit = lastSourceValue.copy().multiplyBy( incrementSize ).increment();
```
dreab8: may be I'm missing something but cannot just be `upperLimit = sourceValue.copy().multiplyBy( incrementSize ).increment();` removing the ` lastSourceValue = sourceValue.copy();` ?

hibernate-core/src/main/java/org/hibernate/id/enhanced/HiLoOptimizer.java
```diff
@@ -53,12 +53,78 @@
 public class HiLoOptimizer extends AbstractOptimizer {
 	private static final Logger log = Logger.getLogger( HiLoOptimizer.class );
 
-	private static class GenerationState {
+	protected static class GenerationState {
+		private final int incrementSize;
```
dreab8: not sure about having the `incrementSize` here and a copy in the `AbstractOptimizer`, plus the check done in the `GenerationState` constructor
 `if ( incrementSize < 1 ) {
	throw new HibernateException( "increment size cannot be less than 1" );
}`
has been already done in the `HiLoOptimizer` one. 

What about adding an `incrementSize`  parameter to the `updateFromNewSourceValue` method

hibernate-core/src/main/java/org/hibernate/id/enhanced/HiLoOptimizer.java
```diff
@@ -53,12 +53,78 @@
...
+		 * required when {@link #isInitialized()} is {@code false},
+		 * or the current value has already reached the upper limit.
+		 * When {@code true} is returned, {@link #updateFromNewSourceValue(IntegralDataTypeHolder)}
+		 * must be called in order to obtain the next value.
+		 *
+		 * @return true, if a new source value is required in order
+		 *               to obtain the next value; false, otherwise.
+		 */
+		public final boolean requiresNewSourceValue() {
+			return ! isInitialized() || ! upperLimit.gt( value );
```
dreab8: cannot be enough `return  ! upperLimit.gt( value );` ?

hibernate-core/src/main/java/org/hibernate/id/enhanced/HiLoOptimizer.java
```diff
@@ -80,47 +146,43 @@ public HiLoOptimizer(Class returnClass, int incrementSize) {
 	public synchronized Serializable generate(AccessCallback callback) {
 		final GenerationState generationState = locateGenerationState( callback.getTenantIdentifier() );
 
-		if ( generationState.lastSourceValue == null ) {
+		if ( !generationState.isInitialized() ) {
```
dreab8: what about having a `isNotInitialized()` method?

Sanne:
This is tricky; let's have a brainstorming first to explore if it's worth going into this direction.

Personally I don't think I would have done this, for the following reasons - but feel free to give me more context:

1. I would expect that RX would possibly use a different design altogether? Especially the locking / synchronization strategies might not apply. A self-filling queue with a background worker is something I had envisioned for ORM too, but was tricky to implement - I believe it could be more efficient, and actually simpler for RX.

2. MVP of Hibernate RX : are we that far to need advanced optimisers? It's great that we start looking but I feel it's one of the last things to polish.

gbadner:
@Sanne , thanks for chiming in with your concerns, I definitely wanted to get feedback before going too far down this road.

The hibernate-rx issue is https://github.com/hibernate/hibernate-rx/issues/47. We picked this one as a first issue only because it's marked "good first issue". I'm fine with dropping this, and moving to a different issue that's more urgent.

Here is the PR I came up with for the hibernate-rx issue: https://github.com/hibernate/hibernate-rx/pull/57

I had originally copied a lot of this code into hibernate-rx, but then I realized I could extend HiLoOptimizer instead. I thought that extending it would be more desirable, provided it didn't adversely affect performance.

I also wondered about whether synchronization would work in hibernate-rx. Davide suggested running it under stress, similar to what is done for OGM: https://github.com/hibernate/hibernate-ogm/blob/master/core/src/test/java/org/hibernate/ogm/backendtck/id/SequenceNextValueGenerationTest.java

I haven't had a chance to do that yet. We (obviously) don't want to do this if synchronizing ID generation won't work.

gavinking:
> I also wondered about whether synchronization would work in hibernate-rx.

I actually don't understand @Sanne's comment on this point. AFAICT, there's no problem *per se* with calling a synchronized method. All we're trying to do here is avoid any sort of blocking IO, so we obviously can't call a synchronized method that goes off and calls the database or whatever, but there the problem is the blocking IO, not the synchronization.

For example, there's no problem at all, from my point of view with Hibernate RX calling a *local* cache using synchronization on the cache. It would even be OK if it were a distributed cache that performed replication via asynchronous IO. On the other hand, there *would* be a problem if the cache were a distributed cache that performed replication synchronously.

Now with respect to the optimizers, I don't understand how they work at all. But if they do all their work in memory, I don't see what is the problem with reusing them in Hibernate RX. OTOH if they query the database, then I guess we can't reuse them.

> I had originally copied a lot of this code into hibernate-rx, but then I realized I could extend HiLoOptimizer instead. I thought that extending it would be more desirable, provided it didn't adversely affect performance.

Obviously I would have *much* preferred to reuse Hibernate's `IdentifierGenerator` contract in Hibernate RX, and get all the sophisticated functionality there for free, but of course the problem is that this contract is fundamentally unsuitable (`generate()` doesn't return a `CompletionStage`), so I was forced to build something basic from scratch instead.

I would be *very happy* if it were reasonable to integrate Hibernate's existing optimizers with this, because I don't like the idea of having to maintain two different hierarchies of identifier generators that do essentially the exact same thing.

gavinking:
@Sanne 

> I'd copy the code in RX and then make aggressive changes

I disagree. We should do everything reasonable to avoid unnecessary copy/paste. The more copy/pasted code we have in Hibernate RX, the harder it is to understand what Hibernate RX is *actually doing*.

In particular: almost anything that never touches the database doesn't belong in Hibernate RX.

NathanQingyangXu:
@gavinking I happened to fix a related bug so I got exposed to some basics of id optimizer. Here is a good artcile to explain the basics: https://vladmihalcea.com/the-hilo-algorithm/. Thanks.

gavinking:
@NathanQingyangXu I understand what they do *conceptually*. I'm saying I'm not quite clear on the precise division of responsibilities in the current design of Hibernate.

Sanne:
hey all, allow me to clarify a thing about my language: when I said "Personally I don't think I would have done this" :  I didn't mean that in any authoritative way - I apologise if it came through that way. 

Please consider I've not been able to check the latest progress and evaluate this in detail - that's precisely why I ask questions :)


@gavinking :

> I actually don't understand @Sanne's comment on this point. AFAICT, there's no problem _per se_ with calling a synchronized method. All we're trying to do here is avoid any sort of blocking IO, so we obviously can't call a synchronized method that goes off and calls the database or whatever, but there the problem is the blocking IO, not the synchronization.

Right. I don't mean to pick on the use of the "synchronization" keyword, I'm referring to the inherent design of the current optimisers.  As you say, there's nothing _per se_ wrong in the semantics, but I do wonder for sake of efficiency I'd be surprised we could reuse the same algorhytms. That said, I couldn't study the solution and I guess it's ok to not provide the most efficient solution at this time.

Which is why I'm surprised we want to implement the optimisers? They are meant as optimisations so I assumed they would be low priority.

> For example, there's no problem at all, from my point of view, with Hibernate RX calling a _local_ cache using synchronization on the cache [...]

Agreed, but assuming that the threads would be waiting for some purely in-memory operations. AFAIK the current optimisers are designed with the assumption that it's fine to perform blocking IO within the critical section. I might be wrong, or it might  be that I'm thinking about a specific implementation, but my fear is you'd need to actually check all implementations.

> Now with respect to the optimizers, I don't understand how they work at all. But if they do all their work in memory, I don't see what is the problem with reusing them in Hibernate RX. OTOH if they query the database (and block the calling thread while they're doing it), then I guess we can't reuse them.

That's what I mean.  Although, we don't need all the advanced optimisers: I'd say implement a single one, hardcode its usage.

> > I had originally copied a lot of this code into hibernate-rx, but then I realized I could extend HiLoOptimizer instead. I thought that extending it would be more desirable, provided it didn't adversely affect performance.

If you can that's great. But how do you invoke the AccessCallback to increase the higher limit without blocking all other threads?

> Obviously I would have _much_ preferred to reuse Hibernate's `IdentifierGenerator` contract in Hibernate RX, and get all the sophisticated functionality there for free, but of course the problem is that this contract is fundamentally unsuitable (`generate()` doesn't return a `CompletionStage`), so I was forced to build something basic from scratch instead.
> 
> I would be _very happy_ if it were reasonable to integrate Hibernate's existing optimizers with this, because I don't like the idea of having to maintain two different hierarchies of identifier generators that do essentially the exact same thing.

+1 

If that's possible, I'd totally agree. I'm just skeptical on that being possible, given how the current ones work. Surprise me :)

Quoting my previous comment: "A self-filling queue with a background worker ..." perhaps we could have both projects share the same implementations, provided we improve the SPI in ORM/blocking first?

But  that's me thinking longer-term. If you need this PR merged to not get stuck, I have no objections.

Sanne:
thanks @gbadner for the link to the RX pr.

So this is what I feared:
 - https://github.com/hibernate/hibernate-rx/pull/57/files#r380264356

Sanne:
So my suggestion would be to split this in two tasks:

1. Implement a very simple optimiser urgently
2. Implement a smarter optimiser - optional.

Regarding 2# ; I think I'd like doing it as it's quite interesting but I won't have time soon - it's more likely to be a fun challenge for the spring holidays :)

If anyone wants to do it, I'd suggest to run indeed some tests like we did for OGM; but with a twist:

- to verify that there's strictly no duplicates, no matter the DB isolation level and the level of parallel threads
- to also verify that we're not incrementing the database value far more then necessary. By "far more" I'd say the threshold should depend on the block size: rounding errors are fine, but we shouldn't let entire blocks go to waste.

Sanne:
@NathanQingyangXu when Gavin sais he doesn't understand something about Hibernate, best to take it in very relative terms :)

gavinking:
OK, @Sanne, now I understand: you're saying that the optimizers *do* hit the DB from a `synchronized` function. Let's continue the discussion in the other issue.

NathanQingyangXu:
> @NathanQingyangXu when Gavin sais he doesn't understand something about Hibernate, best to take it in very relative terms :)

Yeah, lesson learned. It is funny, like teaching Paul Mccartney how to play guitar or Picasso how to draw, :).

Sanne:
@gbadner , @gavinking do we still need this PR in ORM ?

gavinking:
> @gbadner , @gavinking do we still need this PR in ORM ?

My conclusion from the discussion above is that you convinced us that we need to go down a different path.

Which is, reimplement block optimization from scratch in HR. Which is fine.

So I guess we should close it?

Sanne:
yes, thanks for confirming @gavinking . closing

PR #3238, opened 2020-02-14
Wip/6.0 circularity detection

dreab8:
This PR tries to` fix :

- circularity detection 
  - https://github.com/hibernate/hibernate-orm/commit/01c2cc082f151a439acebb595fdf396d438e9618#diff-cb8a1a8f5a0e41bddbc848061ad794feR199  as discussed with Steve  may be we can do this check inside the CollectionLoader.
- hql query with a where clause checking for Embeddable is null generates the wrong sql (commit https://github.com/hibernate/hibernate-orm/commit/18a3a5804daca510ac7e00c8e05730420ffc4ba9)
-  FromClauseIndex#findFetchedJoinByPath not resolving FetchedJoin during EntityResultImpl creation (see Trello related card)  (commit https://github.com/hibernate/hibernate-orm/commit/4479e5517f971e394ea628a2be3e5cb6eeb32109)


design/working/circular-fetching.adoc
```diff
@@ -0,0 +1,208 @@
...
+```
+from LineItem l
+    join fetch l.order o
+    join fetch l.order o2
+```
+
+This should be illegal.  It is possible to handle it specially, but that would be very complicated.
+
+`FromClauseIndex#findJoinFetch(NavigablePath parentPath, String fetchableName)` - if this does not return null we
+know that the association is already join fetched and we should throw and exception
```
NathanQingyangXu: throw an exception?

design/working/circular-fetching.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+// we need to be able to ultimately be able to resolve the "identifying columns" for a given path.  E.g.
+
+```
+interface DomainResultCreationState {
+    ...
+
+    Fetchable resolveFetchable(NavigablePath navigablePath) {
+        // the path passed in here would be `pathToParent` (i.e. `Root(r).leaf1(l)`)
+        // we'd used that to determine the lhs's identifying columns via
```
NathanQingyangXu: we'd use that?

design/working/circular-fetching.adoc
```diff
@@ -0,0 +1,208 @@
...
+}
+
+@Entity
+@Table(name="leaf")
+class LeafEntity {
+    ...
+
+	@ManyToOne
+	@JoinColumn(name="root_id)
+    RootEntity rootEntity;
```
NathanQingyangXu: guess there is some indentation issue here

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ForeignKeyDescriptor.java
```diff
@@ -61,6 +64,10 @@ default void visitColumns(ColumnConsumer consumer) {
 
 	void visitColumnMappings(FkColumnMappingConsumer consumer);
 
+	<T> T visitColumnMapping(FkColumnMappingFunction<T> function);
+
+	// todo (6.0): the 2 interfaces does not take into account composite keys, referringColumn
```
NathanQingyangXu: do not take?

PR #3239, opened 2020-02-16
HQL 'insert select', 'insert values', and 'update' queries

gavinking:
This pull request:

- fixes some bugs in the grammar and implementation for HQL `update` queries
- fills in the implementation of HQL `insert ... select ...` queries
- adds `insert ... values ...` queries to HQL

Note that `insert ... values ...` allows insertion of multiple rows at once. This is a very useful feature for setting up test data.

Note that this is a rough draft and I'm looking for feedback from @sebersole on a number of implementation points.


hibernate-core/src/main/java/org/hibernate/query/hql/internal/SqmPathRegistryImpl.java
```diff
@@ -170,7 +170,7 @@ public SqmFrom findFromExposing(String navigableName) {
 		for ( SqmFrom fromElement : sqmFromByPath.values() ) {
 			if ( definesAttribute( fromElement.getReferencedPathSource(), navigableName ) ) {
 				if ( found != null ) {
-					throw new IllegalStateException( "Multiple from-elements expose unqualified attribute : " + navigableName );
+//					throw new IllegalStateException( "Multiple from-elements expose unqualified attribute : " + navigableName );
```
gavinking: @sebersole I had to remove this exception in order to make the following query work:

    insert into Foo (bar) select f.bar from Foo f

However, I feel like it should not have been necessary to make this change, and that there is some deeper problem here. The subquery should have its own inner-scoped namespace.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -564,6 +570,22 @@ private NonSelectQueryPlan buildUpdateQueryPlan() {
...
+//		final String entityNameToUpdate = sqmInsert.getTarget().getReferencedPathSource().getHibernateEntityName();
+//		final EntityPersister entityDescriptor = getSessionFactory().getDomainModel().findEntityDescriptor( entityNameToUpdate );
+
+//		final SqmMultiTableMutationStrategy multiTableStrategy = entityDescriptor.getSqmMultiTableMutationStrategy();
+//		if ( multiTableStrategy == null ) {
+			return new SimpleInsertQueryPlan( sqmInsert, domainParameterXref );
+//		}
+//		else {
+			//TODO:
+//			return new MultiTableUpdateQueryPlan( sqmInsert, domainParameterXref, multiTableStrategy );
```
gavinking: @sebersole do we need a `MultiTableInsertQueryPlan`?

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/StandardSqmInsertTranslator.java
```diff
@@ -0,0 +1,233 @@
...
+	@Override
+	public SqmInsertTranslation translate(SqmInsertStatement sqmStatement) {
+		InsertStatement sqlAst;
+		if ( sqmStatement instanceof SqmInsertSelectStatement ) {
+			sqlAst = visitInsertSelectStatement( (SqmInsertSelectStatement) sqmStatement );
+		}
+		else {
+			sqlAst = visitInsertValuesStatement( (SqmInsertValuesStatement) sqmStatement );
+		}
+		return new SqmInsertTranslation( sqlAst, getJdbcParamsBySqmParam() );
```
gavinking: @sebersole if you absolutely insist, we could have separate `StandardSqmInsertValuesTranslator` and `StandardSqmInsertSelectTranslator`, but it just seems all a bit like overkill.

To be honest I even really don't see the need for both `SqmInsertSelectStatement` and `SqmInsertValuesStatement`. I would make them one class.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/StandardSqmUpdateTranslator.java
```diff
@@ -263,13 +262,10 @@ public Expression resolveSqlExpression(
...
-						assert valueExprJdbcCount == 1;
-
-						assignments.add( new Assignment( (ColumnReference) valueExpression, valueExpression ) );
-					}
-					else {
-						throw new NotYetImplementedFor6Exception( "Support for composite valued assignments in an UPDATE query is not yet implemented" );
+					for (ColumnReference columnReference : targetColumnReferences) {
+						assignments.add(
+								new Assignment( columnReference, valueExpression )
+						);
```
gavinking: @sebersole I'm not sure that this is 100% correct, but it's certainly better than what was there before.

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
```diff
@@ -301,42 +303,81 @@ public SqmSelectStatement visitSelectStatement(HqlParser.SelectStatementContext
...
+		if ( ctx.querySpec()!=null ) {
+			final SqmInsertSelectStatement<?> insertStatement = new SqmInsertSelectStatement<>( root, creationContext.getNodeBuilder() );
+			parameterCollector = insertStatement;
+			final SqmDmlCreationProcessingState processingState = new SqmDmlCreationProcessingState(
+					insertStatement,
+					this
+			);
+
+			processingStateStack.push( processingState );
+			processingState.getPathRegistry().register( root );
```
NathanQingyangXu: is it better to move the last statement into the try block below so we can ensure the `processingState` is always popped even when the registering process throws some exception (I checked its internal implementation and found it seems possible)?
ignore if irrelevant. Just a little bit confusing from casual code reader's eyes.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -531,6 +533,10 @@ private NonSelectQueryPlan buildNonSelectQueryPlan() {
 			return buildUpdateQueryPlan();
 		}
 
+		if ( getSqmStatement() instanceof SqmInsertStatement ) {
+			return buildInsertQueryPlan();
+		}
+
 		throw new NotYetImplementedException( "Query#executeUpdate for Statements of type [" + getSqmStatement() + "not yet supported" );
```
NathanQingyangXu: Just noticed the above exception message lacks ']' somewhere, though outside the scope of this PR.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/cte/package-info.java
```diff
@@ -24,7 +24,7 @@
  * 			* select - `QuerySpec`
  * 			* delete - `DeleteStatement`
  * 			* update - `UpdateStatement`
- * 			* insert-select - `InsertSelectStatement
+ * 			* insert-select - `InsertStatement
```
NathanQingyangXu: there is a missing `\`` at the end

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/insert/InsertStatement.java
```diff
@@ -7,60 +7,73 @@
...
+	private static final Logger log = Logger.getLogger( InsertStatement.class );
+
 	private TableReference targetTable;
 	private List<ColumnReference> targetColumnReferences;
-	private List<Expression> values;
-
-	public InsertStatement(TableReference targetTable) {
-		this.targetTable = targetTable;
-	}
+	private QuerySpec sourceSelectStatement = null;
```
NathanQingyangXu: We never need to set `null` value for JVM has done that for us already, right?

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/insert/InsertStatement.java
```diff
@@ -7,60 +7,73 @@
...
-		if ( values == null ) {
-			values = new ArrayList<>();
-		}
-		values.add( expression );
+	public List<Values> getValuesList() {
+		return valuesList;
+	}
+
+	public void setValuesList(List<Values> valuesList) {
+		this.valuesList = valuesList;
```
NathanQingyangXu: Saw the g4 rule dictates the list shoud not be empty. Do we need to guard against empty list to avoid producing invalid SQL?

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
```diff
@@ -35,38 +35,40 @@ subQuery
...
-	: INTO entityName
+valuesList
+	: VALUES values (COMMA values)*
 	;
 
-targetFieldsSpec
-	:
-	LEFT_PAREN dotIdentifierSequence (COMMA dotIdentifierSequence)* RIGHT_PAREN
+values
+	: LEFT_PAREN expression (COMMA expression)* RIGHT_PAREN
```
NathanQingyangXu: I know very little about ANTLR. Is it possible to dicate the above two list share the same number of elements? Sorry for my ignorance.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/StandardSqlAstInsertTranslator.java
```diff
@@ -42,37 +43,62 @@ public JdbcInsert translate(InsertSelectStatement sqlAst) {
...
 
 		appendSql( ") " );
 
-		visitQuerySpec( sqlAst.getSourceSelectStatement() );
+		if ( sqlAst.getSourceSelectStatement()!=null ) {
+			visitQuerySpec( sqlAst.getSourceSelectStatement() );
+		}
+		else {
+			appendSql("values");
+			boolean firstTuple = true;
```
NathanQingyangXu: Do we need to check the values and the columns share the same length here? Seems better to initiate a SQL roundtrip which is destined to fail.

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
gavinking: No, it's not possible, and you generally don't want to do as little validation as possible in the parser because parsers don't usually give good error messages.

However, you're right that we should have code to detect user error here. However that is a problem that predates my work (it also affects `insert ... select ...`).

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/StandardSqlAstInsertTranslator.java
gavinking: Definitely don't want to do it here in the code that renders SQL, no.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
sebersole: Depends how we implement it.  This is different from delete and update in that those need special handling for the matching ids whereas insert would not.  But maybe it makes the impl easier.  The other option would be to simply have a single plan for insert handling and have it manage the multi-table aspect

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/StandardSqmInsertTranslator.java
sebersole: I agree a single translator is enough.

We could consolidate the 2 statement types but that would mean having a commonality between SqmQuerySpec and whatever you use to model the values constructor.  So I'm not against this part, though its not a big deal to me

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/StandardSqmUpdateTranslator.java
sebersole: Yes, generally implemented > !implemented ;)

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SqmPathRegistryImpl.java
sebersole: Not sure.  Ideally add some *specific* tests for scoping and see.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
gavinking: OK, I will have a close a look at that next and do whatever is simpler.

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
gavinking: Perhaps, but that's not how it was written before this commit, and it's not what happens in `visitUpdateStatement()`, so it's not related to this PR. Anyway, If `register()` throws, then AFAIK the whole query translation process is hosed, so I doubt the try/finally block is even really useful for anything.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/cte/package-info.java
gavinking: I'm not going to force-push to fix a missing quite in a comment that occurs in code that's not in any way related to this PR. Sorry.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/insert/InsertStatement.java
gavinking: I fixed it.

gavinking:
I've updated this PR to accommodate a change on the `wip/6.0` branch, and also to remove an unused class.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
gavinking: Alright, so I had a look into this, and noticed that `update` is broken for attributes mapped to `@SecondaryTable`s. Which makes me question whether I can really be bothered doing the work to make `insert` work for `@SecondaryTable`s at this point, rather than doing it later...

sebersole:
Applied upstream manually

PR #3241, opened 2020-02-16
HQL 'any'/'some', 'every'/'all', and 'exists' subquery operators

gavinking:
This PR adds support to Hibernate 6 for the following subquery operators:

- `exists` and `not exists` predicates
- `any`/`some` and `every`/`all` expressions used on RHS of comparison operators

Note that `any`/`some` and `every`/`all` collide with the HQL aggregate functions of the same name, but can be distinguished by the syntactic form of their arguments.

UPDATE: 

- also allows subqueries to occur in the `select` list.

### DIFF
--- a/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
+++ b/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
@@ -372,6 +372,7 @@ predicate
 	| expression (NOT)? BETWEEN expression AND expression	# BetweenPredicate
 	| expression (NOT)? LIKE expression (likeEscape)?		# LikePredicate
 	| expression comparisonOperator expression				# ComparisonPredicate
+	| EXISTS expression										# ExistsPredicate
 	| MEMBER OF path										# MemberOfPredicate
 	| NOT predicate											# NegatedPredicate
 	| predicate AND predicate								# AndPredicate
@@ -657,14 +658,6 @@ sumFunction
 	: SUM LEFT_PAREN DISTINCT? expression RIGHT_PAREN
 	;
 
-everyFunction
-	: (EVERY|ALL) LEFT_PAREN DISTINCT? predicate RIGHT_PAREN
-	;
-
-anyFunction
-	: (ANY|SOME) LEFT_PAREN DISTINCT? predicate RIGHT_PAREN
-	;
-
 minFunction
 	: MIN LEFT_PAREN DISTINCT? expression RIGHT_PAREN
 	;
@@ -675,6 +668,14 @@ maxFunction
 
 countFunction
 	: COUNT LEFT_PAREN DISTINCT? (expression | ASTERISK) RIGHT_PAREN
+	;
+
+everyFunction
+	: (EVERY|ALL) LEFT_PAREN (predicate | subQuery) RIGHT_PAREN
+	;
+
+anyFunction
+	: (ANY|SOME) LEFT_PAREN (predicate | subQuery) RIGHT_PAREN
 	;
 
 standardFunction
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -528,7 +528,9 @@ public abstract class BaseSqmToSqlAstConverter
 	protected void consumeFromClauseRoot(SqmRoot<?> sqmRoot) {
 		log.tracef( "Resolving SqmRoot [%s] to TableGroup", sqmRoot );
 
-		assert ! fromClauseIndex.isResolved( sqmRoot );
+		if ( fromClauseIndex.isResolved( sqmRoot ) ) {
+			log.tracef( "Already resolved SqmRoot [%s] to TableGroup", sqmRoot );
+		}
 
 		final EntityPersister entityDescriptor = resolveEntityPersister( sqmRoot.getReferencedPathSource() );
 
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -66,6 +66,7 @@ import org.hibernate.query.sqm.tree.domain.SqmPluralValuedSimplePath;
 import org.hibernate.query.sqm.tree.domain.SqmTreatedPath;
 import org.hibernate.query.sqm.tree.expression.Conversion;
 import org.hibernate.query.sqm.tree.expression.JpaCriteriaParameter;
+import org.hibernate.query.sqm.tree.expression.SqmAny;
 import org.hibernate.query.sqm.tree.expression.SqmBinaryArithmetic;
 import org.hibernate.query.sqm.tree.expression.SqmByUnit;
 import org.hibernate.query.sqm.tree.expression.SqmCaseSearched;
@@ -74,6 +75,7 @@ import org.hibernate.query.sqm.tree.expression.SqmCastTarget;
 import org.hibernate.query.sqm.tree.expression.SqmDistinct;
 import org.hibernate.query.sqm.tree.expression.SqmDurationUnit;
 import org.hibernate.query.sqm.tree.expression.SqmEnumLiteral;
+import org.hibernate.query.sqm.tree.expression.SqmEvery;
 import org.hibernate.query.sqm.tree.expression.SqmExpression;
 import org.hibernate.query.sqm.tree.expression.SqmExtractUnit;
 import org.hibernate.query.sqm.tree.expression.SqmFieldLiteral;
@@ -100,6 +102,7 @@ import org.hibernate.query.sqm.tree.insert.SqmInsertSelectStatement;
 import org.hibernate.query.sqm.tree.predicate.SqmAndPredicate;
 import org.hibernate.query.sqm.tree.predicate.SqmBetweenPredicate;
 import org.hibernate.query.sqm.tree.predicate.SqmComparisonPredicate;
+import org.hibernate.query.sqm.tree.predicate.SqmExistsPredicate;
 import org.hibernate.query.sqm.tree.predicate.SqmGroupedPredicate;
 import org.hibernate.query.sqm.tree.predicate.SqmInListPredicate;
 import org.hibernate.query.sqm.tree.predicate.SqmInSubQueryPredicate;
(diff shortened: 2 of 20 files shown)


gavinking:
I have pushed a second commit which allows subqueries to occur in the HQL `select` list, and be referenced by alias in the `order by` clause.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -522,7 +528,9 @@ public Void visitFromClause(SqmFromClause sqmFromClause) {
 	protected void consumeFromClauseRoot(SqmRoot<?> sqmRoot) {
 		log.tracef( "Resolving SqmRoot [%s] to TableGroup", sqmRoot );
 
-		assert ! fromClauseIndex.isResolved( sqmRoot );
+		if ( fromClauseIndex.isResolved( sqmRoot ) ) {
+			log.tracef( "Already resolved SqmRoot [%s] to TableGroup", sqmRoot );
+		}
```
gavinking: @sebersole I had to suppress this assertion to get `select` clause subqueries referenced in the `order by` clause to work. But I suppose there must be some deeper cause to this that is perhaps worth fixing.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/select/QuerySpec.java
```diff
@@ -112,6 +121,41 @@ public void accept(SqlAstWalker sqlTreeWalker) {
 
 	@Override
 	public MappingModelExpressable getExpressionType() {
-		return null;
+		SqlSelection first = selectClause.getSqlSelections().get(0);
+		return ( (SqlSelectionImpl) first ).getWrappedSqlExpression().getExpressionType();
```
gavinking: @sebersole I needed this typecast in three places. I guess either:

- `getWrappedSqlExpression()` should be pulled up to the interface, or
- `we should add a `getExpressionType()` method to the interface.

Or something else?

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
```diff
@@ -677,6 +670,14 @@ countFunction
...
 	;
 
+everyFunction
+	: (EVERY|ALL) LEFT_PAREN (predicate | subQuery) RIGHT_PAREN
+	;
+
+anyFunction
+	: (ANY|SOME) LEFT_PAREN (predicate | subQuery) RIGHT_PAREN
+	;
+
```
gavinking: Note that `any`/`some` and `every`/`all` have quite different semantics depending upon whether their argument is a predicate or a subquery. However, I failed to convince myself that we really needed to distinguish these two roles in the parser.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
sebersole: No idea why that would be necessary.  It would seem like you are resolving the from-clause twice or perhaps related to "scoping"

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/select/QuerySpec.java
sebersole: Probably the second makes more sense

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
gavinking: @sebersole Yeah it confused me too, so I went digging the other day and I believe it's related to the fact that by default a select expression referenced by alias in the `order by` clause gets copied into the `order by`.  Now this behavior surprised me, because I had been under the impression that most databases allow you to reference a selected expression by alias in `order by`, but somebody wrote explicit code to do this so there must be a reason for it.

Anyway it means that `select` clause expression can end up being walked twice.

gavinking: > but somebody wrote explicit code to do this so there must be a reason for it

FTR, it looks like "somebody" was you haha.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/select/QuerySpec.java
gavinking: OK, I will make that change.

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
sebersole: I'm having a hard time envisioning a situation for the predicate-form.  For example, what does `... where someLhsExpr > all (1=1)`.  Yes, I had predicate there initially but that was just wrong from the first drafts of the grammar because these qualifiers had not yet been tested.

Not specific to this PR, but I do not like the categorization of these as functions.  They are really more a qualifier or restriction of the sub-query-based predicate.  In terms of modeling I think it makes the most sense to wrap the sub-query in a ANY/SOME expression wrapper

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/select/QuerySpec.java
gavinking: > OK, I will make that change.

Done.

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
gavinking: It's used as an aggregate function:

    select every(person.age>10) from Person p
    select any(order.status = Status.UNPAID) from Customer p where p.id = :id

gavinking: > Not specific to this PR, but I do not like the categorization of these as functions. They are really more a qualifier or restriction of the sub-query-based predicate.

Yeah well that's what I was alluding to when I said I failed to convince myself that we need to separate them in the *grammar*. The aggregate functions `every()` and `any()` are certainly functions and belong alongside `sum()` and friends.

But the subquery *quantifiers* aren't functions at all. (Indeed, I find them to be a rather strange syntax.)

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
sebersole: You mean something like HQL `select p from Person p order by p`?

But the way this is meant to work is that `#consumeFromClauseRoot` (this method) only happens from `#visitFromClause`.  All other cases route through `#visitRootPath`.

Basically, when walking the from-clause nodes we do a specialized "visitation".  Other usages of a `SqmFrom` in other parts of the query-spec are handled by the normal visitation.

Do you have a stack trace?

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
sebersole: I've never seen that form before.  Nice!

Ok, so you mean to handle the "aggregate function" form via SQM functions and the sub-query qualifier form via `SqmAny`, etc?

sebersole: Ok, I see that is what you do in `SemanticQueryBuilder#visitAnyFunction`.  That's fine.  Personally I'd be inclined to define distinct rules just to be clear about the distinction, but I'm certainly not going to fight over it ;)

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/select/QuerySpec.java
sebersole: Nice!

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
sebersole: oh, i see what you mean...  For curiosity, if you can still produce that, could you switch `org.hibernate.dialect.Dialect#replaceResultVariableInOrderByClauseWithPosition` to true and see what happens?

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
gavinking: > I've never seen that form before. Nice!

So it's actually part of ANSI SQL, if I'm reading the spec correctly, but I only found one database that supported them. However, it's possible to emulate them on most databases, so that's what I did (because I think they're useful).

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
gavinking: I mean like this HQL:

    select (select 1) as one, (select 'foo') as foo order by one, foo

Currently produces this SQL:

    select
        (select
            1 
        from
            dual),
        (select
            'foo' 
        from
            dual) 
    from
        dual 
    order by
        (select
            1 
        from
            dual),
        (select
            'foo' 
        from
            dual)

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
gavinking: > Ok, so you mean to handle the "aggregate function" form via SQM functions and the sub-query qualifier form via `SqmAny`, etc?

Precisely.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
gavinking: @sebersole if I'm not mistaken, `replaceResultVariableInOrderByClauseWithPosition()` appears to currently have no effect.

sebersole: It should affect how `AbstractSqlAstWalker#visitSqlSelectionExpression` works:

```
	@Override
	public void visitSqlSelectionExpression(SqlSelectionExpression expression) {
		final boolean useSelectionPosition = dialect.replaceResultVariableInOrderByClauseWithPosition();

		if ( useSelectionPosition ) {
			appendSql( Integer.toString( expression.getSelection().getJdbcResultSetIndex() ) );
		}
		else {
			expression.getExpression().accept( this );
		}
	}
```

beikov: I just saw this discussion and wanted to mention that some DBMS i.e. PostgreSQL and I also think MySQL also allow to use a select alias in the GROUP BY clause. Support for that was merged into 5.4 recently.

gavinking: @beikov Yeah, I think that's ideal.

gavinking: @sebersole that method never gets called in my test.

sebersole:
Pushed this manually after fixing the conflict

gavinking:
Thanks 👍

PR #3244, opened 2020-02-17
HHH-13179 Subclass 2nd level caching now works for XML mappings

larkar:
Fixes HHH-13179 and HHH-13173, regressions in 5.3 and forward.

Possibly originates from HHH-12146 - Support enabling caching at any level within a mapped hierarchy.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
@@ -523,6 +523,7 @@ public class ModelBinder {
 			PersistentClass superEntityDescriptor) {
 		for ( IdentifiableTypeSource subType : entitySource.getSubTypes() ) {
 			final SingleTableSubclass subEntityDescriptor = new SingleTableSubclass( superEntityDescriptor, metadataBuildingContext );
+			subEntityDescriptor.setCached(superEntityDescriptor.isCached());
 			bindDiscriminatorSubclassEntity( (SubclassEntitySourceImpl) subType, subEntityDescriptor );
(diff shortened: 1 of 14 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
```diff
@@ -523,6 +523,7 @@ private void bindDiscriminatorSubclassEntities(
 			PersistentClass superEntityDescriptor) {
 		for ( IdentifiableTypeSource subType : entitySource.getSubTypes() ) {
 			final SingleTableSubclass subEntityDescriptor = new SingleTableSubclass( superEntityDescriptor, metadataBuildingContext );
+			subEntityDescriptor.setCached(superEntityDescriptor.isCached());
```
NathanQingyangXu: We have code format requirement (see https://github.com/hibernate/hibernate-ide-codestyles). So it would be appreciated if you add space chars as following to the above statement
```
subEntityDescriptor.setCached( superEntityDescriptor.isCached() );
```
The same applies to the other code changes below.

NathanQingyangXu: You can require @sebersole to review for he implemented HHH-12146. However, he is super busy  with v6 and patience would be appreciated.

larkar: Thanks Nathan for your input! 

Fixed code formatting issue. I requested @sebersole as reviewer as suggested, and I totally understand if this is not a top priority on the core teams list! :)

sebersole:
Thanks @larkar

PR #3245, opened 2020-02-17
HHH-8051 - Add graceful not-found support for to-one associations.

Naros:
https://hibernate.atlassian.net/browse/HHH-8051

This is the HHH-8051 fix based on master rather than the legacy 6.0 branch.

### DIFF
--- a/hibernate-envers/src/main/java/org/hibernate/envers/Audited.java
+++ b/hibernate-envers/src/main/java/org/hibernate/envers/Audited.java
@@ -11,6 +11,8 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+import org.hibernate.Incubating;
+
 /**
  * When applied to a class, indicates that all of its properties should be audited.
  * When applied to a field, indicates that this field should be audited.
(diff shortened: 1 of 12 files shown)


hibernate-envers/src/main/java/org/hibernate/envers/configuration/EnversSettings.java
```diff
@@ -142,4 +143,17 @@
...
 	 */
 	String MODIFIED_COLUMN_NAMING_STRATEGY = "org.hibernate.envers.modified_column_naming_strategy";
+
+	/**
+	 * Globally defines whether legacy relation not-found behavior should be used or not.
+	 * Defaults to {@code true}.
+	 *
+	 * By specifying {@code true}, any {@code EntityNotFoundException} will be thrown unless the containing
+	 * class or property explicitly specifies that use case to be ignored.  Conversely, when specifying the
+	 * value {@code false}, the inverse applies an requires explicitly specifying the use case as error so
```
NathanQingyangXu: seems 'the inverse applies an requires explicity specifiying ...' is confusing. 'an' -> 'and'?

hibernate-envers/src/main/java/org/hibernate/envers/Audited.java
```diff
@@ -37,6 +39,25 @@
...
+	 *
+	 * When the configuration property is {@code true}, this is to use the legacy behavior which
+	 * implies that the system should throw the {@code EntityNotFoundException} errors unless
+	 * the user has explicitly specified the value {@link RelationTargetNotFoundAction#IGNORE}.
+	 *
+	 * When the configuration property is {@code false}, this is to use the new behavior which
+	 * implies that the system should ignore the {@code EntityNotFoundException} errors unless
+	 * the user has explicitly specified the value {@link RelationTargetNotFoundAction#ERROR}.
+	 *
+	 * @since 6.0
```
gbadner: `6.0` is not the correct version. I would like to remove the above, rather than correcting it, since it will need to be corrected again if/when backported to 5.3 branch.

gbadner:
This new feature should be added to documentation as "experimental".

hibernate-envers/src/main/java/org/hibernate/envers/Audited.java
Naros: @gbadner That seems reasonable.  I completely missed that when I ported the code to master.

Naros:
Closed in favor of PR #4478

PR #3246, opened 2020-02-18
HHH-13853 fix minor issue

NathanQingyangXu:
a followup of https://github.com/hibernate/hibernate-orm/pull/3227.
Fixing an issue that if 'integrationSettings' is null (will be assigned Collections.emptyMap(), which is immutable) and new properties is added to it, exception will be thrown.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
@@ -181,7 +181,7 @@ public class EntityManagerFactoryBuilderImpl implements EntityManagerFactoryBuil
 			ClassLoaderService providedClassLoaderService ) {
 		this( persistenceUnit, integrationSettings, null, providedClassLoaderService);
 	}
-	
+
 	private EntityManagerFactoryBuilderImpl(
 			PersistenceUnitDescriptor persistenceUnit,
 			Map integrationSettings,
@@ -193,13 +193,14 @@ public class EntityManagerFactoryBuilderImpl implements EntityManagerFactoryBuil
 		this.persistenceUnit = persistenceUnit;
 
 		if ( integrationSettings == null ) {
-			integrationSettings = Collections.emptyMap();
+			integrationSettings = new HashMap();
 		}
 
 		Map mergedIntegrationSettings = null;
 		Properties properties = persistenceUnit.getProperties();
 		if ( properties != null ) {
-			mergedIntegrationSettings = new HashMap( persistenceUnit.getProperties() );
+			// original integratin setting entries take precedence
+			mergedIntegrationSettings = new HashMap( properties );
 			mergedIntegrationSettings.putAll( integrationSettings );
 		}
 


hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
```diff
@@ -192,20 +192,17 @@ private EntityManagerFactoryBuilderImpl(
...
-		}
+		final Map integtrationSettingsCopy = integrationSettings == null ? new HashMap() : new HashMap( integrationSettings );
 
-		Map mergedIntegrationSettings = null;
 		Properties properties = persistenceUnit.getProperties();
 		if ( properties != null ) {
-			mergedIntegrationSettings = new HashMap( persistenceUnit.getProperties() );
-			mergedIntegrationSettings.putAll( integrationSettings );
+			// original integration settings take precedence
+			properties.forEach( integtrationSettingsCopy::putIfAbsent );
```
NathanQingyangXu: I do think the above code makes the subtle logic explicit ending up with better code readability and maintainability. Maybe it is worthwhile to log something is some property is ignored (which might be pretty troublesome to investigate why something is not working as expected).

hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
```diff
@@ -192,20 +192,17 @@ private EntityManagerFactoryBuilderImpl(
 
 		this.persistenceUnit = persistenceUnit;
 
-		if ( integrationSettings == null ) {
-			integrationSettings = Collections.emptyMap();
-		}
+		final Map integtrationSettingsCopy = integrationSettings == null ? new HashMap() : new HashMap( integrationSettings );
```
NathanQingyangXu: We should avoid `Collections.emptyMap` if there is possibility to modify it later.
Also, I do think defensive copying never hurts (a good practise as per 'Effective Java') so I create one up front. That also simplified the logic later on.

hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java
```diff
@@ -192,20 +192,17 @@ private EntityManagerFactoryBuilderImpl(
...
-			mergedIntegrationSettings = new HashMap( persistenceUnit.getProperties() );
-			mergedIntegrationSettings.putAll( integrationSettings );
+			// original integration settings take precedence
+			properties.forEach( integtrationSettingsCopy::putIfAbsent );
 		}
 
 		// Build the boot-strap service registry, which mainly handles class loader interactions
 		final BootstrapServiceRegistry bsr = buildBootstrapServiceRegistry(
-				mergedIntegrationSettings != null ? mergedIntegrationSettings : integrationSettings,
+				integtrationSettingsCopy,
```
NathanQingyangXu: The original implementation doesn't enjoy crystal clear logic for it is tightly coupled with previous logic. Centralizing the logic in one single location up front is better, IMO.

Sanne:
good catch. Thanks @NathanQingyangXu !

PR #3247, opened 2020-02-19
HHH-13870 - Extract task action in Gradle plugin to fix up-to-date check

dreis2211:
See https://hibernate.atlassian.net/browse/HHH-13870

### DIFF
--- a/tooling/hibernate-gradle-plugin/src/main/groovy/org/hibernate/orm/tooling/gradle/HibernatePlugin.java
+++ b/tooling/hibernate-gradle-plugin/src/main/groovy/org/hibernate/orm/tooling/gradle/HibernatePlugin.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.orm.tooling.gradle;
 
+import org.gradle.api.Action;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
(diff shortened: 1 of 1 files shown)


tooling/hibernate-gradle-plugin/src/main/groovy/org/hibernate/orm/tooling/gradle/HibernatePlugin.java
```diff
@@ -43,10 +44,29 @@ private void applyEnhancement(final Project project, final HibernateExtension hi
 
 			final Task compileTask = project.getTasks().findByName( sourceSet.getCompileJavaTaskName() );
 			assert compileTask != null;
-			compileTask.doLast(
-					task -> EnhancementHelper.enhance( sourceSet, hibernateExtension.enhance, project )
-			);
+			compileTask.doLast(new EnhancerAction( sourceSet, hibernateExtension, project ));
```
NathanQingyangXu: I would suggest adding some comment explaining why lambda is not a viable solution (including the gradle bug link) so we can avoid confusion and raise awareness of when the lambda can be reverted back.

dreis2211: Added a comment

sebersole:
Thanks @dreis2211

PR #3250, opened 2020-02-19
Wip/6.0 Port 'refresh' cascade testing case

NathanQingyangXu:
Part of https://trello.com/c/OVdBw7qD/27-verify-cascading
Other types of cascade testing has been taken over by Andrea.

Fix a generic issue when loading entity by id providing existing entity (like in cascade refresh case).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/SingleIdEntityLoaderDynamicBatch.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/SingleIdEntityLoaderDynamicBatch.java
@@ -34,7 +34,6 @@ import org.hibernate.sql.exec.spi.JdbcParameterBinding;
 import org.hibernate.sql.exec.spi.JdbcParameterBindings;
 import org.hibernate.sql.exec.spi.JdbcSelect;
 import org.hibernate.sql.results.internal.RowTransformerPassThruImpl;
-
 import org.jboss.logging.Logger;
 
 /**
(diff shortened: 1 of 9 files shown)


hibernate-core/src/main/java/org/hibernate/loader/ast/internal/SingleIdEntityLoaderDynamicBatch.java
```diff
@@ -187,6 +182,14 @@ public T load(
 			Object entityInstance,
 			LockOptions lockOptions,
 			SharedSessionContractImplementor session) {
+		initializeSingleIdLoaderIfNeeded( session );
 		return singleIdLoader.load( pkValue, entityInstance, lockOptions, session );
```
NathanQingyangXu: Fix a potential NPE issue above

hibernate-core/src/main/java/org/hibernate/sql/exec/internal/JdbcSelectExecutorStandardImpl.java
```diff
@@ -158,7 +158,7 @@ public String getEffectiveOptionalEntityName() {
 
 			@Override
 			public Object getEffectiveOptionalId() {
-				return null;
+				return executionContext.getEntityId();
```
NathanQingyangXu: We seems to need the non-null id value here so the only usage of 'getEffectiveOptionalInstance()' in `AbstractEntityInitilizer` (see below) could work because of the id comparison precondition

sebersole:
@NathanQingyangXu Why change this from annotations to XML?

NathanQingyangXu:
> @NathanQingyangXu Why change this from annotations to XML (which is cause of the conflict)?

What did you mean? I saw no usage of XML in this PR. Am I missing something?

NathanQingyangXu:
> @NathanQingyangXu Why change this from annotations to XML (which is cause of the conflict)?

The conflict is because Andrea's additional `@FailureExpected`. Now conflict is resolved.

PR #3174, opened 2020-01-17
HHH-13809 Fix some obvious issues in user guide

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13809

Hibernate's user guide is amazing. I learned a lot during creating this PR. Issues fixed include:

- grammatical errors (typos seem to have been eliminated to great extent, though not completely)
- UI CSS tweaking (make image positioned in the centre, etc.)
- update outdated references (basic type mapping table, etc.)
- many more

example code snippet in test folder is not included to make the code review manageable. Might create other ticket to improve the format issue (e.g. mixing of tab and spaces make the indentation inconsistent).

For anyone interested at testing, here is the procedure to test it locally:

1. change directory to the root folder of 'hibernate-orm' repo;
2. issue command: `./gradlew :documentation:renderUserGuide`;
3. open the user guide at `documentation/target/asciidoc/userguide/html_single/Hibernate_User_Guide.html`;

PR #3251, opened 2020-02-20
HHH-13872 - Make the Java Stream close the underlying ScrollableResultsIterator upon calling a terminal operation

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13872

### DIFF
--- a/documentation/src/main/asciidoc/userguide/chapters/query/hql/HQL.adoc
+++ b/documentation/src/main/asciidoc/userguide/chapters/query/hql/HQL.adoc
@@ -165,9 +165,10 @@ include::{sourcedir}/HQLTest.java[tags=jpql-api-positional-parameter-example]
 It's good practice not to mix parameter binding forms in a given query.
 ====
 
-In terms of execution, JPA `Query` offers 2 different methods for retrieving a result set.
+In terms of execution, JPA `Query` offers 3 different methods for retrieving a result set.
 
 * `Query#getResultList()` - executes the select query and returns back the list of results.
+* `Query#getResultStream()` - executes the select query and returns back a `Stream` over the results.
 * `Query#getSingleResult()` - executes the select query and returns a single result. If there were more than one result an exception is thrown.
 
 [[jpql-api-list-example]]
@@ -176,6 +177,15 @@ In terms of execution, JPA `Query` offers 2 different methods for retrieving a r
 [source, JAVA, indent=0]
 ----
 include::{sourcedir}/HQLTest.java[tags=jpql-api-list-example]
+----
+====
+
+[[jpql-api-stream-example]]
+.JPA `getResultStream()` result
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/HQLTest.java[tags=jpql-api-stream-example]
 ----
 ====
 
@@ -413,6 +423,29 @@ include::{sourcedir}/HQLTest.java[tags=hql-api-stream-example]
 ====
 Just like with `ScrollableResults`, you should always close a Hibernate `Stream` either explicitly or using a https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html[try-with-resources] block.
 ====
+
+[[jpql-api-stream]]
+==== Query streaming
+
+Since version 2.2, the JPA `Query` interface offers support for returning a `Stream` via the `getResultStream` method.
+
+Just like the `scroll` method, you can use a try-with-resources block to close the `Stream`
+prior to closing the currently running Persistence Context.
+
+Since Hibernate 5.4, the `Stream` is also closed when calling a terminal operation,
+as illustrated by the following example.
+
+[[jpql-api-stream-terminal-operation]]
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/HQLTest.java[tags=jpql-api-stream-terminal-operation]
+----
+====
+
+The `Stream` is closed automatically after calling the `collect` method,
+since there is no reason to keep the underlying JDBC `ResultSet` open
+if the `Stream` cannot be reused.
 
 [[hql-case-sensitivity]]
 === Case Sensitivity
--- a/documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
+++ b/documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
@@ -10,7 +10,6 @@ import java.math.BigDecimal;
 import java.sql.Timestamp;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
@@ -43,7 +42,6 @@ import org.hibernate.userguide.model.PhoneType;
 import org.hibernate.userguide.model.WireTransferPayment;
 
 import org.hibernate.testing.DialectChecks;
-import org.hibernate.testing.FailureExpected;
 import org.hibernate.testing.RequiresDialect;
 import org.hibernate.testing.RequiresDialectFeature;
 import org.hibernate.testing.SkipForDialect;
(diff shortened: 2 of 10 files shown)


vladmihalcea:
@NathanQingyangXu @dreab8 @sebersole Would you mind taking a look at this PR?

This PR adds two benefits:

1. It allows us to pass the `closeHandler` when creating a derived `Stream`, `IntStream`, `LongStream`, `DoubleStream` as a result of a `mapToInt`, `mapToLong`, etc method on the original `Stream`
2. We can close the underlying `ResultSet` after calling a terminal operation since the `Stream` is not reusable afterward. If we don't do that, we the `ResultSet` will be open until the `Session` gets closed or the or the currently running transaction is committed or rolled back.

hibernate-core/src/main/java/org/hibernate/query/spi/StreamDecorator.java
```diff
@@ -0,0 +1,308 @@
...
+ * The {@link StreamDecorator} wraps a Java {@link Stream} and registers a {@code closeHandler}
+ * which is passed further to any resulting {@link Stream}.
+ *
+ * The goal of the {@link StreamDecorator} is to close the underlying {@link Stream} upon
+ * calling a terminal operation.
+ *
+ * @author Vlad Mihalcea
+ * @since 5.4
+ */
+public class StreamDecorator<R> implements Stream<R> {
```
NathanQingyangXu: Seems another exciting PR from Vlad. I just wanna raise some awareness of the maintenance cost of the current implementation. So suppose next JVM adds some new method to Stream, do we need to modify the class accordingly to sync up? Seems not a concern for JVM maintains backward compatibility by interface default method?

Other than delegator, Spring's AOP might be a good implementation alternative provided JDK exposes some mean to know whether the current method is 'terminal' one or not.

hibernate-core/src/test/java/org/hibernate/test/stream/basic/JpaStreamTest.java
```diff
@@ -70,12 +79,100 @@ public void testQueryStream() {
...
+			assertFalse( resourceRegistry.hasRegisteredResources() );
+		} );
+	}
+
+	private ResourceRegistry resourceRegistry(Session session) {
+		SharedSessionContractImplementor sharedSessionContractImplementor = (SharedSessionContractImplementor) session;
+		JdbcCoordinator jdbcCoordinator = sharedSessionContractImplementor.getJdbcCoordinator();
+		return jdbcCoordinator.getLogicalConnection().getResourceRegistry();
+	}
+
```
NathanQingyangXu: One good edge case for this PR is to verify the following scenario:

- some stream client code has closed stream already (to adhere to best practice, maybe via 'try' block);
- this new feature was added without client knowing and the stream might be closed twice;

It would be excellent to add a testing case to ensure the above common scenario won't throw some exception (not only for now, but for the long future).

hibernate-core/src/main/java/org/hibernate/query/spi/StreamDecorator.java
```diff
@@ -0,0 +1,308 @@
...
+import java.util.function.ToDoubleFunction;
+import java.util.function.ToIntFunction;
+import java.util.function.ToLongFunction;
+import java.util.stream.Collector;
+import java.util.stream.DoubleStream;
+import java.util.stream.IntStream;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+
+import org.hibernate.Incubating;
```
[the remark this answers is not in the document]
vladmihalcea: I used it at first, but then I took it out because I didn't know which features qualify for `@Incubating`. Now, it's clear.

hibernate-core/src/main/java/org/hibernate/query/spi/StreamDecorator.java
vladmihalcea: @sebersole @NathanQingyangXu I thought of that too, but then I run the tests with Java 13 and there was no issue. I've been using this approach in {FlexyPool](https://github.com/vladmihalcea/flexy-pool) as an alternative to dynamic proxies for the JDBC `Connection`. If there's a method added in a new JDK version, and we need to compile with JDK 8, we can just add the new methods without `@Override` and use Reflection to invoke the method on the delegate.

Another approach is to use Byte Buddy, but there will be many if/else checks to determine which methods we need to call `close` and which require creating a new `StreamDecorator`.

hibernate-core/src/test/java/org/hibernate/test/stream/basic/JpaStreamTest.java
vladmihalcea: Yes, we should totally add that test case. Luckily, the `AsbtractScrollableResults` returns immediately from `close` if the RS was already closed:

```
if ( this.closed ) {
	// noop if already closed
	return;
}
```

NathanQingyangXu: Yeah, I noticed that, but adding testing case would make us sleep well if the above code was refactored.

hibernate-core/src/main/java/org/hibernate/query/spi/StreamDecorator.java
vladmihalcea: @sebersole @NathanQingyangXu Cool. I found why it works on Java 13. It looks like the JDK folks decided to mark the new `Stream` methods using `default`, so that's why the test run. However, we still need to add those so that we can either decorate the `Stream` or call `close` if it's a terminator operation. I'll issue a new commit based on these comments.

NathanQingyangXu:
Do we have the concern that this feature will break some existing client code which assumes it is their responsibility to close the stream whenever they want? Seems not.
It might be a good idea to put some thing in our wonderful user guide to explain this cool feature to avoid unnecessary bug reporting or suchlike.

vladmihalcea:
@NathanQingyangXu I added a test case for the close method and it works just fine. I think that in case this feature gets merged, it's a good idea to update the User Guide to explain how the JPA 2.2 Stream results work in Hibernate, and why it's a good idea to call a terminal operation sooner than later.

vladmihalcea:
@NathanQingyangXu @sebersole I pushed the new commit based on our discussion.

I'm now thinking of adding a new test case that will run on Java 9+ only so that I can verify the outcome of the `takeWhile` and `dropWhile` methods. I want to make sure the `Stream` is closed for those methods as well when executing the terminal operation.

sebersole:
There are some check-style failures.  Other than that, +1

vladmihalcea:
I run the `check` phase to fix those issues as well.

vladmihalcea:
I also added test cases for `takeWhile` and `doWhile`.

vladmihalcea:
I updated the User Guide as well. I'll rebase and merge it tomorrow.

vladmihalcea:
@NathanQingyangXu I rebased the branch from master and squashed all previous commits. If you can merge it, I'd really appreciate it. Thanks.

NathanQingyangXu:
> @NathanQingyangXu I rebased the branch from master and squashed all previous commits. If you can merge it, I'd really appreciate it. Thanks.

I have no such rights for I am a community contributor only. But I guess Steve will do that soon.

NathanQingyangXu:
> I updated the User Guide as well. I'll rebase and merge it tomorrow.

Thanks for the User Guide update. I used to finish a big PR to improve on it: https://github.com/hibernate/hibernate-orm/pull/3174. 
One thing I really love it is the testing code and easy verification. That is brilliant.
Seems lots of work needs to be done after v6 is released.

documentation/src/main/asciidoc/userguide/chapters/query/hql/HQL.adoc
```diff
@@ -414,6 +424,29 @@ include::{sourcedir}/HQLTest.java[tags=hql-api-stream-example]
...
 
+[[jpql-api-stream]]
+==== Query streaming
+
+Since version 2.2, the JPA `Query` interface offers support for returning a `Stream` via the `getResultStream` method.
+
+Just like the `scroll` method, you can use a try-with-resources block to close the `Stream`
+prior to closing the currently running Persistence Context.
+
+Since Hibernate 5.4, the `Stream` is also closed when calling a terminal operation,
```
NathanQingyangXu: technically speaking, this feature will appear in next version, i.e. 5.4.13, not ever since 5.4 was released.

documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -673,6 +671,38 @@ public void test_jpql_api_list_example() {
...
+			// tag::jpql-api-stream-terminal-operation[]
+			List<Person> persons = entityManager.createQuery(
+					"select p " +
+					"from Person p " +
+					"where p.name like :name", Person.class )
+			.setParameter( "name", "J%" )
+			.getResultStream()
+			.skip( 5 )
+			.limit( 5 )
+			.collect( Collectors.toList() );
```
NathanQingyangXu: It would be great to make consistent usage of code style. I even created a ticket to fix the various format issue (sometimes indentation produces really bad result because of mixed usage of TAB and space, e.g.).
Reformatting our testing code with the https://github.com/hibernate/hibernate-ide-codestyles will solve the issue easily.
So I am okay with this (there is some minor format issue above like Line 678. There should be a space after `try` and before `{`).
Many thanks for updating the User Guide, however.

PR #3253, opened 2020-02-20
HHH-13874 - Deprecating methods that will be removed soon

thegeekyasian:
Two methods that are dropped in v6.0 are now marked as deprecated in this commit.

As discussed in the removal PR https://github.com/hibernate/hibernate-orm/pull/3229

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/mapping/Constraint.java
+++ b/hibernate-core/src/main/java/org/hibernate/mapping/Constraint.java
@@ -156,6 +156,7 @@ public abstract class Constraint implements RelationalModel, Exportable, Seriali
 		return columns.iterator();
 	}
 
+	@Deprecated
 	public Iterator<Column> columnIterator() {
 		return columns.iterator();
 	}
--- a/hibernate-core/src/main/java/org/hibernate/mapping/Constraint.java
+++ b/hibernate-core/src/main/java/org/hibernate/mapping/Constraint.java
@@ -156,6 +156,14 @@ public abstract class Constraint implements RelationalModel, Exportable, Seriali
 		return columns.iterator();
 	}
 
+	/**
+	 * This is a duplicate method, that has been removed in v.6.0.
+	 * {@link #getColumnIterator()} is the one that will stay.
+	 *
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/mapping/Constraint.java
```diff
@@ -156,6 +156,7 @@ public Column getColumn(int i) {
 		return columns.iterator();
 	}
 
+	@Deprecated
```
sebersole: You'll have to add a deprecation message

sebersole: Otherwise the build will fail with checkstyle failures

thegeekyasian: Ah, my bad. I've added the deprecation messages.

Thanks for the correction.

hibernate-core/src/main/java/org/hibernate/mapping/Constraint.java
```diff
@@ -156,6 +156,15 @@ public Column getColumn(int i) {
 		return columns.iterator();
 	}
 
+	/**
+	 * This is a duplicate method, that has been removed in v.6.0.
+	 * {@link #getColumnIterator()} is the one that will stay.
+	 *
+	 * @author TheGeekyAsian
```
NathanQingyangXu: technically speaking, you were not the original author of the method, were you?

thegeekyasian: No, I wasn’t. 

I assumed that’s the deprecation template where one has to add their names while adding a message. 

Should I remove that?

NathanQingyangXu: Javadoc `@author` should only be reserved for the original author of the method IMO.

thegeekyasian: I agree with you on that.

I'll add a new commit that drops the `@author` added.

PR #3255, opened 2020-02-21
Wip/6.0 Wrap up verification of 'sorted-map' and 'sorted-set'

NathanQingyangXu:
Wrapped up existing testing cases for 'sorted-map' and 'sorted-set'.

Also ported 'org.hibernate.test.sorted' along the way. Fixed a serious Criteria bug when fixing some regression error.


hibernate-core/src/test/java/org/hibernate/orm/test/metamodel/mapping/collections/SetOperationTests.java
```diff
@@ -119,8 +124,7 @@ public void testSqmFetch(SessionFactoryScope scope) {
 					).getSingleResult();
 
 					assertThat( entity.getSetOfBasics(), InitializationCheckMatcher.isInitialized() );
-
-					assert  entity.getSetOfBasics().size() == 2;
+					assertThat( entity.getSetOfBasics(), IsCollectionWithSize.hasSize( 2 ) );
```
NathanQingyangXu: I think reusing existing hamcrest stuff enjoys many benefits, one of which is the detailed error message.

hibernate-core/src/test/java/org/hibernate/orm/test/metamodel/mapping/collections/SetOperationTests.java
```diff
@@ -164,14 +168,40 @@ public void testSortedSetAccess(SessionFactoryScope scope) {
...
+					assertThat( entity.getSortedSetOfBasics(), IsCollectionWithSize.hasSize( 5 ) );
+					assertThat( entity.getSetOfEnums(), InitializationCheckMatcher.isNotInitialized() );
+
+					assertThat( entity.getSortedSetOfBasics(), IsIterableContainingInOrder.contains(
+							"abc",
+							"bcd",
+							"cde",
+							"def",
+							"efg"
+					) );
```
NathanQingyangXu: An example error message I encountered went like this:
```
Expected: iterable containing ["aBC", "bcD", "cDe", "DeF", "Efg"]
     but: item 0: was "abc"
```

hibernate-core/src/test/java/org/hibernate/orm/test/sorted/SortTest.java
```diff
@@ -0,0 +1,169 @@
...
+						"jboss.com"
+				) );
+
+				sess.clear();
+				criteria = criteriaBuilder.createQuery( Search.class );
+				root = criteria.from( Search.class );
+				root.fetch( "searchResults", JoinType.LEFT );
+				criteria.select( root );
+				s = sess.createQuery( criteria ).uniqueResult();
+				assertTrue( Hibernate.isInitialized( s.searchResults ) );
```
NathanQingyangXu: Actually the above statement is the only reason `@FailureExpected` is used. Seems some complicated reason because of the query cache for Criteria. 
Given it is a good testing case for Criteria, I retained it though it seems laziness has nothing to do with the testing intention for this method.

hibernate-core/src/test/java/org/hibernate/orm/test/sql/exec/onetoone/OneToOneWithDerivedIdentityTest.java
```diff
@@ -52,7 +52,7 @@ public void testGet() {
 		inTransaction(
 				session -> {
 					Person person = session.get( Person.class, PERSON_ID );
-					assertEquals( person.getName(), "Alfio" );
+					assertEquals( "Alfio", person.getName() );
```
NathanQingyangXu: A common erroneous usage of `assertEquals` in our codebase.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/AbstractSqmFrom.java
```diff
@@ -409,55 +409,55 @@ public SqmMapJoin joinMap(String attributeName, JoinType jt) {
...
+					String.format(
+							Locale.ROOT,
+							"Passed attribute [%s] did not correspond to a joinable reference [%s] relative to %s",
+							joinedPathSource.getPathName(),
+							joinedPathSource,
+							getNavigablePath().getFullPath()
+					)
+			);
+		}
+		addSqmJoin( sqmJoin );
```
NathanQingyangXu: So the obvious bug here is we did build `join` but didn't add it to 'joins' field. Simple fixing.

hibernate-core/src/test/java/org/hibernate/orm/test/sql/exec/onetoone/OneToOneWithDerivedIdentityTest.java
sebersole: Yes, this is once of the great things about hamcrest which we use in newer tests

PR #3256, opened 2020-02-22
HHH-13875 : Optional one-to-one does not always join the associated entity table when querying

gbadner:
https://hibernate.atlassian.net/browse/HHH-13875

@jwgmeligmeyling, could you please take a look at this? It is related to https://hibernate.atlassian.net/browse/HHH-12775.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
@@ -394,11 +394,14 @@ public class DotNode extends FromReferenceNode implements DisplayableNode, Selec
 
 		if ( isDotNode( parent ) ) {
 			// our parent is another dot node, meaning we are being further dereferenced.
-			// thus we need to generate a join unless the parent refers to the associated
-			// entity's PK (because 'our' table would know the FK).
+			// thus we need to generate a join unless the association is non-nullable and
+			// parent refers to the associated entity's PK (because 'our' table would know the FK).
 			parentAsDotNode = (DotNode) parent;
 			property = parentAsDotNode.propertyName;
-			joinIsNeeded = generateJoin && !isPropertyEmbeddedInJoinProperties( parentAsDotNode.propertyName );
+			joinIsNeeded = generateJoin && (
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
```diff
@@ -394,11 +394,14 @@ private void dereferenceEntity(
...
-			// thus we need to generate a join unless the parent refers to the associated
-			// entity's PK (because 'our' table would know the FK).
+			// thus we need to generate a join unless the association is non-nullable and
+			// parent refers to the associated entity's PK (because 'our' table would know the FK).
 			parentAsDotNode = (DotNode) parent;
 			property = parentAsDotNode.propertyName;
-			joinIsNeeded = generateJoin && !isPropertyEmbeddedInJoinProperties( parentAsDotNode.propertyName );
+			joinIsNeeded = generateJoin && (
+					entityType.isNullable() ||
+					!isPropertyEmbeddedInJoinProperties( parentAsDotNode.propertyName )
```
jwgmeligmeyling: Interesting, I thought `entityType.isNullable()` would imply `!isPropertyEmbeddedInJoinProperties( parentAsDotNode.propertyName )`.

Obviously the API of the `PropertyMapping` gets a little bit vague here. Most code assumes that properties in the `PropertyMapping` are actually deducible from just the association owners subclass column span, which nullable properties are inherently not. In that sense, it doesn't make sense for the property to be considered here. As far as I remember, this was at least how these were handled at the time I wrote this originally. It might have changed since, or I might just as well have simply overlooked this case.

Anyways, I expected `joinIsNeeded` for `generateJoin && entityType.isNullable` - so nothing worries me here. The tests that the two of us introduced with HHH-12775 and others are also still passing. The tests for this PR are very extensive.  So I am sure this change is fine!

gbadner: @gbadner, thanks for your comments. I think the the issue here is that a nullable one-to-one association was not taken into account. When a querying using a path for an optional one-to-one with an implicit join, a join must be included in the query.

PR #3257, opened 2020-02-22
HHH-13876 Fix an obvious bug in StandardStack implementation

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13876

The bug is obvious. However, it is never used in v5 and its implementation has been refactored in v6.


Sanne:
Thanks for spotting this @NathanQingyangXu :+1: 

This method seems not used at all. Did you consider deleteding the method instead? Since it's in the package `org.hibernate.internal.` we can consider making such non-backwards compatible changes.

Sanne:
Hi @NathanQingyangXu , sorry I pushed an alternative fix to delete the bug and some more dead code. I hope you don't mind, thanks for spotting this!

NathanQingyangXu:
off course I don't mind. thanks.

PR #3258, opened 2020-02-24
Changes for Rx

DavideD:
If possible, I woud like to relax some of the scopes of some methods so that I can reuse them or override them in Hibernate ORM. Let me know if I need to create a JIRA for this.

Thanks

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
@@ -1005,7 +1005,7 @@ public class SessionImpl
 	}
 
 	@Override
-	public final Object internalLoad(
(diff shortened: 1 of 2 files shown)


Sanne:
changes look ok. I'd prefer to have a JIRA to refer to as that's a rule in ORM.

Could you create one, mention it in the commits as per the usual template, and then merge? Thanks!

DavideD:
Jira: https://hibernate.atlassian.net/browse/HHH-13878

sebersole:
We need to be careful about these changes.  Making them accessible may help
y'all, but it also means some user and is going to use it as well

Sanne:
agreed @sebersole - we're trying to be careful generically speaking.  Specifically, these changes aren't very concerning, would you agree?

PR #3260, opened 2020-02-25
HHH-13879 - Slow query log should use System#nanoTime not System#currentTimeMillis

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13879

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/ResultSetReturnImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/ResultSetReturnImpl.java
@@ -46,9 +46,9 @@ public class ResultSetReturnImpl implements ResultSetReturn {
 	@Override
 	public ResultSet extract(PreparedStatement statement) {
 		// IMPL NOTE : SQL logged by caller
-		long executeStart = 0;
+		long executeStartNanos = 0;
 		if ( this.sqlStatementLogger.getLogSlowQuery() > 0 ) {
-			executeStart = System.currentTimeMillis();
+			executeStartNanos = System.nanoTime();
 		}
 		try {
 			final ResultSet rs;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/ResultSetReturnImpl.java
```diff
@@ -46,9 +46,9 @@ public ResultSetReturnImpl(JdbcCoordinator jdbcCoordinator, JdbcServices jdbcSer
 	@Override
 	public ResultSet extract(PreparedStatement statement) {
 		// IMPL NOTE : SQL logged by caller
-		long executeStart = 0;
+		long executeStartNanos= 0;
```
kortov: I guess all the `long executeStartNanos= 0` lines here are missing a space after `executeStartNanos`

vladmihalcea: Thanks for spotting it. I fixed it and rebased it.

NathanQingyangXu: Saw @Sanne 's comment on your ticket. I don't know too much about it, but it seems another exciting PR from Vlad.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/spi/SqlStatementLogger.java
```diff
@@ -134,34 +135,34 @@ public void logStatement(String statement, Formatter formatter) {
...
-	public void logSlowQuery(String sql, long startTime) {
+	public void logSlowQuery(String sql, long startTimeNanos) {
 		if ( logSlowQuery < 1 ) {
 			return;
 		}
-		assert startTime > 0 : "startTime is invalid!";
-
-		long spent = System.currentTimeMillis() - startTime;
+		if ( startTimeNanos <= 0 ) {
+			throw new IllegalArgumentException( "startTimeNanos should be greater than 0!" );
```
NathanQingyangXu: I would suggest providing more clue about  the exception like
```
throw new IllegalArgumentException( "startTimeNanos [" + startTimeNanos + "] should be greater than 0!" );
```
to help debugger a little bit.

vladmihalcea:
@sebersole Thanks for merging it.

PR #3264, opened 2020-02-26
JDBC optimisations: skip warnings reset, entrust the Connection Pool to do it

Sanne:
see  - https://hibernate.atlassian.net/browse/HHH-13880


Sanne:
hi @sebersole, did you have any time to look?

I'm open to suggestions and could extract a simpler form from this; I believe even just doing what's essentially in 76b1d0d63ee16e3e860a3c7a026413421bd22275 it would help for the most common cases, and yet be a less invasive patch. Might like it better?

yrodiere:
Hey @Sanne , do you think this is still relevant? I suspect that after all this time, rebasing would amount to completely rewriting the PR...

Sanne:
I know it's still a performance issue - but also I suspect nobody was quite happy with this mess.
Let's close the particular PR but would be nice to resolve the underlying problem.

PR #3269, opened 2020-02-26
Wip/6.0 Fix HQL 'setClause' ANTLR grammar error

NathanQingyangXu:
Seems an obvious bug. I found this during the 'column quoting' verification developing.

The original grammar would be broken for the following HQL statements involving multiple set clauses:
```
update Product p set p.name = ?1, p.description = ?2 were p.id = 1
```
Obviously we need `COMMA` here.


gavinking:
Note that I already caught this error and fixed it in https://github.com/hibernate/hibernate-orm/pull/3239.

NathanQingyangXu:
> Note that I already caught this error and fixed it in #3239.

Thanks for letting me know. For that reason, I closed this PR.

PR #3270, opened 2020-02-27
HHH-13322 Fix Oracle dialect's 'getQuerySequencesString()' issue

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13322

Very similar to my previous PR: https://github.com/hibernate/hibernate-orm/pull/3147 (for https://hibernate.atlassian.net/browse/HHH-13373). The root cause is we didn't limit the sequence information within current schema (in Oracle, user/schema is the same thing).

Changing 'all_sequences' table to 'user_sequences' should solve the issue.
Testing case added after careful local testing based on (Mac + Docker + Oracle).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
@@ -420,7 +420,7 @@ public class Oracle8iDialect extends Dialect {
 		else {
 			return
 				String.format(
-						"%s start with %d increment by  %d",
+						"%s start with %d increment by %d",
 						getCreateSequenceString( sequenceName ),
 						initialValue,
 						incrementSize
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
```diff
@@ -490,7 +490,7 @@ public boolean forUpdateOfColumns() {
 
 	@Override
 	public String getQuerySequencesString() {
-		return "select * from all_sequences";
+		return "select * from user_sequences";
```
NathanQingyangXu: In Oracle, 'all_sequences' is a view to list everything regardless of schema, whereas 'user_sequences' will limit within current schema. User and schema is the same thing in Oracle.

dreab8:
applied upstream, I made only a little change to the test, on the CI Oracle instance we have an issue with the schema name you originally used in the test.

Thanks @NathanQingyangXu

PR #3271, opened 2020-02-28
HHH-13884 Order.reverse() contract

seregamorph:
https://hibernate.atlassian.net/browse/HHH-13884

backport for #3272

[javax.persistence.criteria.Order.reverse()](https://docs.jboss.org/hibernate/jpa/2.2/api/javax/persistence/criteria/Order.html#reverse--) method javadoc explicitly declares that `reverse()` should return new instance.


NathanQingyangXu:
Seems a good catch. JPA does require to return a new Order instance. The implementation is based on fluent API design. One of the concern is this code change might break our client's code, but if JPA dictates it, it might be the way we should go.

seregamorph:
@NathanQingyangXu  any objections to merge it?

NathanQingyangXu:
I have approved but I have no rights to merge it for I am not a team member.

seregamorph:
@beikov any chance to accept this? If no, let's close it

PR #3272, opened 2020-02-28
HHH-13884 Order.reverse() contract

seregamorph:
https://hibernate.atlassian.net/browse/HHH-13884

backported in #3271

[javax.persistence.criteria.Order.reverse()](https://docs.jboss.org/hibernate/jpa/2.2/api/javax/persistence/criteria/Order.html#reverse--) method javadoc explicitly declares that `reverse()` should return new instance.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/select/SqmSortSpecification.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/select/SqmSortSpecification.java
@@ -18,7 +18,8 @@ import org.hibernate.query.sqm.tree.expression.SqmExpression;
 public class SqmSortSpecification implements JpaOrder {
 	private final SqmExpression sortExpression;
 	private final String collation;
-	private SortOrder sortOrder;
+	private final SortOrder sortOrder;
+
 	private NullPrecedence nullPrecedence;
 
 	public SqmSortSpecification(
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/tree/select/SqmSortSpecification.java
```diff
@@ -16,22 +16,21 @@
  * @author Steve Ebersole
  */
 public class SqmSortSpecification implements JpaOrder {
-	private SqmExpression sortExpression;
-	private String collation;
-	private SortOrder sortOrder;
-	private final NullPrecedence nullPrecedence;
```
seregamorph: please note: `nullPrecenence` duplicates `precedence` field.
`nullPrecendence` was assigned only in constructor and **never used**, all other access is done to `precedence` field.
I looks like a defect. I fixed it, because otherwise I could not write a test to validate all assignments in this class.

sebersole: I prefer the name `nullPrecedence` - its more consistent and self-documenting

seregamorph: fixed

seregamorph:
@sebersole requested change addressed. Still any objections to merge it?

seregamorph:
@sebersole @beikov please make a decision on it. It was created half a year ago. If you don't like it, lets close.

beikov:
Could you please rebase it, I can't merge it right now.

seregamorph:
@beikov rebased on top of `wip/6.0` and squashed into a single commit without changes.

PR #3275, opened 2020-03-01
Wip/6.0 Fix a double-checked locking implementation imperfection

NathanQingyangXu:
Double checked locking is subtle and error-prone. There is some imperfection as per https://en.wikipedia.org/wiki/Double-checked_locking. See code review self-comment for details.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
@@ -753,7 +753,7 @@ public abstract class AbstractCollectionPersister
 	}
 
 	// lazily initialize instance field via 'double-checked locking'
-	// see https://en.wikipedia.org/wiki/Double-checked_locking
+	// see https://en.wikipedia.org/wiki/Double-checked_locking on why 'volatile' and local copy is used
 	protected CollectionLoader getStandardCollectionLoader() {
 		CollectionLoader localCopy = standardCollectionLoader;
 		if ( localCopy == null ) {
--- a/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
@@ -744,7 +744,7 @@ public abstract class AbstractCollectionPersister
 		}
 	}
 
-	private CollectionLoader standardCollectionLoader;
+	private volatile CollectionLoader standardCollectionLoader;
 
 	@Override
 	public void initialize(Object key, SharedSessionContractImplementor session) throws HibernateException {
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
```diff
@@ -744,18 +744,21 @@ protected void logStaticSQL() {
...
 		determineLoaderToUse( key, session ).load( key, session );
 	}
 
+	// lazily initialize instance field via 'double-checked locking'
+	// see https://en.wikipedia.org/wiki/Double-checked_locking on why 'volatile' and local copy is used
 	protected CollectionLoader getStandardCollectionLoader() {
 		CollectionLoader localCopy = standardCollectionLoader;
 		if ( localCopy == null ) {
 			synchronized (this) {
+				localCopy = standardCollectionLoader;
```
NathanQingyangXu: Without the above statement, there is no reason to check again at the next statement for `localCopy` would always be null.

hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
```diff
@@ -744,18 +744,21 @@ protected void logStaticSQL() {
 		}
 	}
 
-	private CollectionLoader standardCollectionLoader;
+	private volatile CollectionLoader standardCollectionLoader;
```
NathanQingyangXu: The `volatile` keyword is mandatory to benefit from the 'local copy' paradigm later.

hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
```diff
@@ -744,18 +744,21 @@ protected void logStaticSQL() {
...
+	// lazily initialize instance field via 'double-checked locking'
+	// see https://en.wikipedia.org/wiki/Double-checked_locking on why 'volatile' and local copy is used
 	protected CollectionLoader getStandardCollectionLoader() {
 		CollectionLoader localCopy = standardCollectionLoader;
 		if ( localCopy == null ) {
 			synchronized (this) {
+				localCopy = standardCollectionLoader;
 				if ( localCopy == null ) {
 					localCopy = createCollectionLoader( LoadQueryInfluencers.NONE );
 					standardCollectionLoader  = localCopy;
```
NathanQingyangXu: see the explanation in screenshot to fully understand the tech details. It is subtle and error-prone.

Screen Shot 2020-02-29 at 10 31 24 PM

sebersole:
Double-check locking is also sometimes perfectly valid.  

@barreiro if you have a chance would you be willing to look this over?  I'd love to get your thoughts.

NathanQingyangXu:
btw, I created the PR as per the java version (after v1.5) in https://en.wikipedia.org/wiki/Double-checked_locking. The currently implementation boils down to single checking. Also, it is the advanced version (as per the wiki, it can bring 25% performance improvement) 
Screen Shot 2020-03-16 at 9 44 50 AM
and relies on heavily on `volatile` keyword.

NathanQingyangXu:
> Double-check locking is also sometimes perfectly valid.
> 
> @barreiro if you have a chance would you be willing to look this over? I'd love to get your thoughts.

btw, I created the PR as per the java version (after v1.5) in https://en.wikipedia.org/wiki/Double-checked_locking. The current implementation boils down to single checking (the gist of double-checked locking is the second checking within `synchronized` block, which is missing currently). Also, it is the advanced version (as per the wiki, it can bring 25% performance improvement) based on `local copy`
and relies on heavily on `volatile` keyword.

barreiro:
Looks good to me.

sebersole:
Thanks @NathanQingyangXu && @barreiro

PR #3278, opened 2020-03-01
Wip/6.0 Make List.toArray() usage consistent

NathanQingyangXu:
It is counter-intuitive that the following typical pattern
```
list.toArray(new T[list.size()]);
```
is slower than
```
list.toArray(new T[0]);
```
see the good article of https://shipilev.net/blog/2016/arrays-wisdom-ancients/ for details

A even better approach is to migrate away from Array usage in favour of Collection. Array puts performance cost at runtime whereas Collection moves that cost to compiling phase. See another good article on this: https://eclipsesource.com/blogs/2014/04/11/3-good-reasons-to-avoid-arrays-in-java-interfaces/. Array is the most ugly compromise during Java's evolution.

This PR tries to make the common coding pattern consistent (improving performance along the way).

This PR also fixes a naming error (from org.hibernate.internal.util.streams.**StingArrayCollector** -> org.hibernate.internal.util.streams.**StringArrayCollector**)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/action/internal/BulkOperationCleanupAction.java
+++ b/hibernate-core/src/main/java/org/hibernate/action/internal/BulkOperationCleanupAction.java
@@ -91,7 +91,7 @@ public class BulkOperationCleanupAction implements Executable, Serializable {
 			}
 		}
 
-		this.affectedTableSpaces = spacesList.toArray( new String[ spacesList.size() ] );
+		this.affectedTableSpaces = spacesList.toArray( new String[ 0 ] );
 	}
 
 	/**
@@ -138,7 +138,7 @@ public class BulkOperationCleanupAction implements Executable, Serializable {
 			}
 		}
 
-		this.affectedTableSpaces = spacesList.toArray( new String[ spacesList.size() ] );
+		this.affectedTableSpaces = spacesList.toArray( new String[ 0 ] );
 	}
 
 
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
@@ -889,7 +889,7 @@ public class SessionFactoryOptionsBuilder implements SessionFactoryOptions {
 
 	@Override
 	public SessionFactoryObserver[] getSessionFactoryObservers() {
-		return sessionFactoryObserverList.toArray( new SessionFactoryObserver[ sessionFactoryObserverList.size() ] );
+		return sessionFactoryObserverList.toArray( new SessionFactoryObserver[ 0 ] );
 	}
 
 	@Override
@@ -1109,7 +1109,7 @@ public class SessionFactoryOptionsBuilder implements SessionFactoryOptions {
 
 	@Override
 	public EntityNameResolver[] getEntityNameResolvers() {
-		return entityNameResolvers.toArray( new EntityNameResolver[ entityNameResolvers.size() ] );
+		return entityNameResolvers.toArray( new EntityNameResolver[ 0 ] );
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/naming/NamingHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/naming/NamingHelper.java
@@ -54,7 +54,7 @@ public class NamingHelper {
 			columnNamesArray = new Identifier[0];
 		}
 		else {
-			columnNamesArray = columnNames.toArray( new Identifier[ columnNames.size() ] );
+			columnNamesArray = columnNames.toArray( new Identifier[ 0 ] );
 		}
 
 		return generateHashedFkName(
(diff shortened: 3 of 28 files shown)


gavinking:
> A even better approach is to migrate away from Array usage in favour of Collection. Array puts performance cost at runtime whereas Collection moves that cost to compiling phase.

I would be very surprised to find that this is really the case. From what I've seen (while benchmarking during the design of the Ceylon compiler) it is still the case that for truly performance-critical code, you still can't beat arrays. Which makes total sense, when you consider what they are.

In particular, arrays of primitives blow everything else away in terms of performance and memory usage (at least until the JDK gets value types). Which should be obvious.

Now, of course the performance advantages of an array are obviously far bigger for *primitive* arrays than arrays of strings, for example, but *even for object arrays* I saw a difference. (JDK8 FTR.)

NathanQingyangXu:
> > A even better approach is to migrate away from Array usage in favour of Collection. Array puts performance cost at runtime whereas Collection moves that cost to compiling phase.
> 
> I would be very surprised to find that this is really the case. From what I've seen (while benchmarking during the design of the Ceylon compiler) it is still the case that for truly performance-critical code, you just can't beat arrays. Which makes total sense, when you consider what they are.
> 
> * An `ArrayList` _can't possibly_ be faster than an array, because under the covers it uses an array, and therefore its performance is hard-limited to the performance of an array.
> * In particular, arrays of primitives blow everything else away in terms of performance and memory usage (at least until the JDK gets value types). Which should be obvious, since they eliminate oodles of little wrapper objects.
> 
> Now, of course the performance advantages of an array are obviously far bigger for _primitive_ arrays than arrays of strings, for example, but _even for object arrays_ I saw some difference. (JDK8 FTR.)

As always, you made a good point. Yeah, I think primitive collection is special for its boxing/unboxing cost. But I am almost sure for reference type, JDK's Collection is a much better alternative. Basically array is special for it goes about type checking (it maintains element type info in the compiled code whereas Java generics erase it and only check type during compiling phase) and that involves runtime cost. On top of that, our common pattern to transform it to array adds new cost. You can read the good article (https://eclipsesource.com/blogs/2014/04/11/3-good-reasons-to-avoid-arrays-in-java-interfaces/) for details or if you were into it, there is a good book on this I enjoy pretty much: https://www.amazon.ca/Java-Generics-Collections-Development-Process/dp/0596527756/ref=sr_1_1?keywords=java+generics+collection&qid=1583154432&sr=8-1

For codebase with long history, modernizing code is beneficial to enjoy new language features or platform improvement (previously `toArray(new T[list.size]` has better performance than `toArray(new T[0])` but JDK's optimization changes so fast). I made some effort on this line, but I do think it is worthwhile to inspect old coding pattern with fresh eyes in context of Java evolution).

sebersole:
For primitive arrays I agree with Gavin here.  

However for Object arrays I disagree, especially when we end up with multiple "related" primitive arrays versus combining those individual primitives into a class and have a List of those.  My testing shows that equivalent or often even better performance - not to mention it is significantly more readable.

So I think it depends on context.

gavinking:
> However for Object arrays I disagree, especially when we end up with multiple "related" primitive arrays versus combining those individual primitives into a class and have a List of those.

Sure, I certainly agree that we could do a better job of sharing memory, and that's an error (my mistake) that dates back to the very earliest days of Hibernate, but it's not really an issue of arrays vs lists as an issue of "flattening" things into lists/arrays of strings or using complex objects. Ultimately the question of whether you use a list or an array to hold the complex object references isn't going to make much difference at all; under the covers an `ArrayList` is just an array after all.

But to reiterate my real point, when we investigated the issue of arrays vs lists in the context of micro-optimizations for the Ceylon compiler, we concluded that the very fastest way to iterate a list of things was a good old fashioned `for` loop with an `int` loop counter and array element access. It's true that `ArrayList` was not *much* slower (indeed my results lead me to suspect that the JVM has some sort of specific optimizations for `List` iteration), and certainly for most applications, the difference is so slight as to be immaterial.

But to say that arrays are *slower* than lists is simply wrong. They're actually faster, even when you're talking about arrays of objects. For most applications the difference won't matter, but for very performance sensitive code it might.

NathanQingyangXu:
It is my fault to raise a controversial topic and especially I forgot to include primitive array. Let us focus on the goal of this PR. It seems consistently true that new JVM makes `list.toArray(new T[0])` faster than `list.toArray(new T[list.size()])`.

NathanQingyangXu:
given that our usage has been consistent, seems this PR is not that necessary. I dislike the overuse of array of reference type in our codebase and performance difference is not in top priority in recent JVM (as long as it warmed up). What I am concerned is the Java Generics type checking, for array doesn't play well with it. Anyway, I closed this PR for I think it is too trivial.

sebersole:
In our testing, accessing Lists is exactly what you saw - access using a simple for-i loop and `List.get(int)` was incredibly better than using enhanced-for-loops.  I agree it was very disappointing.

I'm not pointing fingers - I've made choices I wish I had handled differently in retrospect.  It happens

I was just bringing up a closely related consideration specific to our codebase for @NathanQingyangXu as a new contributor.

PR #3279, opened 2020-03-01
Wip/6.0 Fix HashMap constructor with expected size usage error

NathanQingyangXu:
There are quite some coding statements with incorrect usage of `new HashMap<>(n)` in our codebase. When we aim to build a map with expected size, we should use another util class in our code base (i.e. `CollectionHelper.mapOfSize( n )`). Using `new HashMap<>(n)` takes no consideration of `load factor`  and we often end up with rehashing unless there are many hashing conflicts.

Also enrich `CollectionHelper`by providing a new `setOfSize()` method and fixed a bug in `concurrentMap()` util method (thus making it not useful any more).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
@@ -58,6 +58,7 @@ import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.engine.spi.Status;
 import org.hibernate.event.spi.EventSource;
 import org.hibernate.internal.CoreMessageLogger;
+import org.hibernate.internal.util.collections.CollectionHelper;
 import org.hibernate.internal.util.collections.ConcurrentReferenceHashMap;
 import org.hibernate.internal.util.collections.IdentityMap;
 import org.hibernate.metamodel.spi.MetamodelImplementor;
@@ -204,7 +205,7 @@ public class StatefulPersistenceContext implements PersistenceContext {
 	@Override
 	public void addUnownedCollection(CollectionKey key, PersistentCollection collection) {
 		if ( unownedCollections == null ) {
-			unownedCollections = new HashMap<>( INIT_COLL_SIZE );
+			unownedCollections = CollectionHelper.mapOfSize( INIT_COLL_SIZE );
 		}
 		unownedCollections.put( key, collection );
 	}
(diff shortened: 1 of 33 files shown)


hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
```diff
@@ -108,17 +109,17 @@
 	 */
 
 	// Loaded entity instances, by EntityKey
-	private HashMap<EntityKey, Object> entitiesByKey;
+	private Map<EntityKey, Object> entitiesByKey;
```
NathanQingyangXu: In previous JDK, specifying `HashMap` enjoys some performance benefit, but recent JDK eliminates the difference and we have no reason not to use interface instead for we seldom rely on HashMap specific feature in our code.

hibernate-core/src/main/java/org/hibernate/internal/util/collections/CollectionHelper.java
```diff
@@ -175,8 +188,7 @@ public static int determineProperSizing(int numberOfElements) {
 	 * @return The created map.
 	 */
 	public static <K, V> ConcurrentHashMap<K, V> concurrentMap(int expectedNumberOfElements, float loadFactor) {
-		final int size = expectedNumberOfElements + 1 + (int) ( expectedNumberOfElements * loadFactor );
-		return new ConcurrentHashMap<>( size, loadFactor );
+		return new ConcurrentHashMap<>( expectedNumberOfElements, loadFactor );
```
NathanQingyangXu: There is a bug in old implementation. Different than `HashMap` or `HashSet`, `ConcurrentHashMap` will perform sizing computation internally so we don't need to compute on top of that.
Screen Shot 2020-03-01 at 3 31 38 PM

hibernate-core/src/main/java/org/hibernate/metamodel/internal/AbstractEmbeddableRepresentationStrategy.java
```diff
@@ -37,7 +38,7 @@ public AbstractEmbeddableRepresentationStrategy(
 		this.embeddableJavaTypeDescriptor = embeddableJavaTypeDescriptor;
 
 		this.propertyAccesses = new PropertyAccess[ propertySpan ];
-		this.attributeNameToPositionMap = CollectionHelper.concurrentMap( propertySpan );
+		this.attributeNameToPositionMap = new ConcurrentHashMap<>( propertySpan );
```
NathanQingyangXu: No need to use the util method in CollectionHelper for `ConcurrentHashMap` constructor is enough for it will compute the size internally.

PR #3282, opened 2020-03-04
HHH-13619

sebersole:
Proposed solution introduces new CollectionSizeNode and CollectionPathNode types and hooks them in to grammars and walking

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/CollectionPathNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/CollectionPathNode.java
@@ -1,0 +1,248 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.hql.internal.ast.tree;
+
+import java.util.List;
+
+import org.hibernate.QueryException;
+import org.hibernate.hql.internal.ast.HqlSqlWalker;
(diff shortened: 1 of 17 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/CollectionPathNode.java
```diff
@@ -0,0 +1,281 @@
...
+			HqlSqlWalker walker) {
+
+		final String referenceName = reference.getText();
+		final String qualifierQueryPath = qualifier == null
+				? ""
+				: ( (FromReferenceNode) qualifier ).getPath();
+		final String referencePath = qualifier == null
+				? referenceName
+				: qualifierQueryPath + "." + reference;
+
```
dreab8: really a minor but probably we can initialize `qualifierQueryPath` and `referencePath` later
```
if ( qualifier == null ) {
   referencePath = referenceName;
....}else{
 final String qualifierQueryPath = ( (FromReferenceNode) qualifier ).getPath();
 referencePath = qualifierQueryPath + "." + reference;
}

File: hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/ImpliedFromElement.java [hunk 2]
```diff
@@ -23,6 +23,21 @@
...
+		super();
+	}
+
+	/**
+	 * Here to add debug breakpoints
+	 */
+	public ImpliedFromElement(FromClause fromClause, FromElement origin, String alias) {
+		super( fromClause, origin, alias );
+	}
+
```
dreab8: another minor, the `ImpliedFromElement(FromClause fromClause, FromElement origin, String alias) ` constructor is never used, may be it can be deleted

sebersole:
Made and pushed some changes based on the review.  Thanks @dreab8

PR #3283, opened 2020-03-04
Wip/6.0 Verify array (draft PR)

NathanQingyangXu:
https://trello.com/c/9NiLgPRJ/16-verify-complete-array-support

This is a draft PR, so Andrea can inspect the progress and interact incrementally. 
Currently only basic testing passed, but with one issue exposed: when one root entity has an array of 1000 elements, eager fetching will end up with 1000 rows sharing the same root entity. Currently the root entity will be initialized 1000 times. This is not directly related to the scope, however.

@dreab8 feel free to keep your eyes on this PR. The commits won't be merged until necessary so you can check its progress easily. Each step might end up with each new issue and you can stop us whenever you found I am heading in wrong direction, :).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
@@ -1295,6 +1295,9 @@ public abstract class AbstractCollectionPersister
 
 	@Override
 	public boolean isLazy() {
+		if ( isArray ) {
+			return false; // array is always eagerly fetched
+		}
 		return isLazy;
 	}
 
@@ -2183,6 +2186,9 @@ public abstract class AbstractCollectionPersister
 
 	@Override
 	public boolean isExtraLazy() {
+		if ( isArray ) {
+			return false; // array is always eagerly fetched
+		}
 		return isExtraLazy;
 	}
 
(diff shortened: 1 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
```diff
@@ -2183,6 +2186,9 @@ public Object getElementByIndex(Object key, Object index, SharedSessionContractI
 
 	@Override
 	public boolean isExtraLazy() {
+		if ( isArray ) {
+			return false;
+		}
 		return isExtraLazy;
 	}
```
NathanQingyangXu: Maybe creating an `AbstractArrayPersister` is a better idea. Effective Java warns the bad practice to rely on such flag field to  create different logic.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -5167,34 +5168,36 @@ public void setPropertyValues(Object object, Object[] values) {
...
-
-						},
+						fetchable -> setPropertyValueInternally( (AttributeMapping) fetchable, object, values ),
 						null
 				);
 			}
 		}
 	}
 
+	private void setPropertyValueInternally(AttributeMapping attribute, Object object, Object[] values) {
```
NathanQingyangXu: intentionally avoid naming of 'setPropertyValue' for it will end up with method overloading with the public method just below.

dreab8:
Hi @NathanQingyangXu here a [PR](https://github.com/hibernate/hibernate-orm/pull/3284) with a different approach to solve the issue

NathanQingyangXu:
closed because of better approach by Andrea.

PR #3285, opened 2020-03-06
HHH-13765 - Missing Hierarchy subclass in query clause when using treat

lelmarir:
With an entity hierarchy as such:

**A**  
  **B** extends **A**    
    **C1** extendes **B**    
    **C2** extends **B**     
      **D** extends **C1**

creating a query using treat:

`select a from A where treat(a as B).property = 'something';`

the generated sql will be:

```
select ... 
from A
where DTYPE in ('C2', 'D') and ...
```

**C1** and **B** are missing, but should be there, since they are not Abstract.
The subclass should be excluded from the query only if the entity is Abstract

https://hibernate.atlassian.net/browse/HHH-13765

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
@@ -622,8 +622,8 @@ public class SingleTableEntityPersister extends AbstractEntityPersister {
 			if ( !queryable.isAbstract() ) {
 				values.add( queryable.getDiscriminatorSQLValue() );
 			}
-			else if ( queryable.hasSubclasses() ) {
-				// if the treat is an abstract class, add the concrete implementations to values if any
+			if ( queryable.hasSubclasses() ) {
+				// add the concrete implementations to values if any
 				Set<String> actualSubClasses = queryable.getEntityMetamodel().getSubclassEntityNames();
 
 				for ( String actualSubClass : actualSubClasses ) {
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
@@ -632,9 +632,7 @@ public class SingleTableEntityPersister extends AbstractEntityPersister {
 					}
 
 					Queryable actualQueryable = (Queryable) getFactory().getEntityPersister( actualSubClass );
-					if ( !actualQueryable.hasSubclasses() ) {
-						values.add( actualQueryable.getDiscriminatorSQLValue() );
-					}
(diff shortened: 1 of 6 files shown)


NathanQingyangXu:
as a general rule, a compelling test case is needed to showcase your code changes do fix some bug, which would get exposed by your testing case without your new code. refer to existing testing code on how to do that. PR with good testing case will get much more chances to be reviewed and ultimately merged.

lelmarir:
> as a general rule, a compelling test case is needed to showcase your code changes do fix some bug, which would get exposed by your testing case without your new code changes. Refer to existing testing testing cases in ongoing or closed PRs on how to do that. PR with good testing case will get much more chances to be reviewed and ultimately merged.

You're right! :) 
Here the test, and meanwhile i've found my initial fix was not quite right

I feel that still now it is not right:
if i do `select ... from VEICLE where DTYPE in ('Car', 'SUV') and (name is not null)`
It will select only Cars and subclasses. Shouldn't it also select all other items that arent Cars (Boats) ignoring the _treat_ condition?
`select ... from VEICLE where (DTYPE in ('Car', 'SUV') and (name is not null)) or (DTYPE in ('Veicle', 'Boat'))` (the or should contains all DTYPE that are not Cars)

lelmarir:
any news on this?

beikov:
A treat expression like e.g. `treat(v as Car).name is not null` should be translated to a case when expression e.g. in SQL to `(case when DTYPE in ('Car', 'SUV') then name else null end) is not null`. This is, AFAICS, what the JPA specificaiton demands, but Hibernate fails to properly implement. Fortunately, Hibernate supports "implicit subtype property access" which allows you to access subtype properties even without the treat operator i.e. you could implement this logic yourself by using the following JPQL/HQL `(case when TYPE(car) = Car OR TYPE(car) = SUV then car.name else null end) is not null`.

I didn't check what your implementation does exactly, but I know that the current treat handling in Hibernate will lead to various issues. I'm not sure if this can be properly fixed in 5.x but I will make sure it will be fixed in 6.0

dreab8:
Hi @lelmarir ,

Thanks for your PR, since the issue seem resolved see 
 [PR](https://github.com/hibernate/hibernate-orm/pull/9384) I'm going to close this.

dreab8:
Hi @lelmarir , can you please add the following as comment to this PR?

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt)
and can be relicensed under the terms of the [LGPL v2.1 license](https://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt) in the future at the maintainers' discretion.
For more information on licensing, please check [here](https://github.com/hibernate/hibernate-orm/blob/main/CONTRIBUTING.md#legal).

PR #3286, opened 2020-03-07
HHH-13891 : ProxyFactory should not be built if any ID or property getter/setter methods are final

gbadner:
https://hibernate.atlassian.net/browse/HHH-13891

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/proxy/pojo/ProxyFactoryHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/proxy/pojo/ProxyFactoryHelper.java
@@ -11,6 +11,7 @@ import java.lang.reflect.Modifier;
 import java.util.Iterator;
 import java.util.Set;
 
+import org.hibernate.HibernateException;
 import org.hibernate.MappingException;
 import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
@@ -84,14 +85,21 @@ public final class ProxyFactoryHelper {
 		Class clazz = persistentClass.getMappedClass();
 		while ( properties.hasNext() ) {
 			Property property = (Property) properties.next();
-			Method method = property.getGetter( clazz ).getMethod();
-			if ( method != null && Modifier.isFinal( method.getModifiers() ) ) {
-				LOG.gettersOfLazyClassesCannotBeFinal( persistentClass.getEntityName(), property.getName() );
-			}
-			method = property.getSetter( clazz ).getMethod();
-			if ( method != null && Modifier.isFinal( method.getModifiers() ) ) {
-				LOG.settersOfLazyClassesCannotBeFinal( persistentClass.getEntityName(), property.getName() );
-			}
+			validateGetterSetterMethodProxyability( "Getter", property.getGetter( clazz ).getMethod() );
+			validateGetterSetterMethodProxyability( "Setter", property.getSetter( clazz ).getMethod() );
+		}
+	}
+
+	public static void validateGetterSetterMethodProxyability(String getterOrSetter, Method method ) {
+		if ( method != null && Modifier.isFinal( method.getModifiers() ) ) {
+			throw new HibernateException(
+					String.format(
+							"%s methods of lazy classes cannot be final: %s#%s",
+							getterOrSetter,
+							method.getDeclaringClass().getName(),
+							method.getName()
+					)
+			);
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/proxy/pojo/ProxyFactoryHelper.java
```diff
@@ -84,14 +85,21 @@ public static void validateProxyability(final PersistentClass persistentClass) {
...
+	public static void validateGetterSetterMethodProxyability(String getterOrSetter, Method method ) {
+		if ( method != null && Modifier.isFinal( method.getModifiers() ) ) {
+			throw new HibernateException(
+					String.format(
+							"%s methods of lazy classes cannot be final: %s#%s",
+							getterOrSetter,
+							method.getDeclaringClass().getName(),
+							method.getName()
+					)
+			);
```
NathanQingyangXu: It seems `getterOrSetter` parameter is not necessary for we can get that information if we refactor the second parameter to accept either `Getter` or `Setter`. For instance
```
private static void validateGetterSetterMethodProxyability(Object getterOrSetter ) {
                String methodCategory = getterOrSetter instanceof Getter ? "Getter" : "Setter";
                Method method = getterOrSetter instanceof Getter ? Getter.class.cast(getterOrSetter).getMethod)(): Setter.class.cast(getterOrSetter).getMethod();
		if ( method != null && Modifier.isFinal( method.getModifiers() ) ) {
			throw new HibernateException(
					String.format(
							"%s methods of lazy classes cannot be final: %s#%s",
							methodCategory,
							method.getDeclaringClass().getName(),
							method.getName()
					)
			);
```
I also changed the visibility from public to private above. If `Getter` and `Setter` shared some common interface (like `methodAware`), it would be more elegant and straightforward.
not a big deal for sure and feel free to ignore my comment.

gbadner: @NathanQingyangXu , `PojoEntityTuplizer#buildProxyFactory` also calls `ProxyFactoryHelper#validateGetterSetterMethodProxyability```, so it needs to be public.

Of course, `ProxyFactoryHelper#validateProxyability` could have repeated the call to `ProxyFactoryHelper#extractProxyGetIdentifierMethod`, but then I would have to change the signature of the method to accept a `Setter` and a `Getter`. 

Since `ProxyFactoryHelper` is an SPI, I would need to deprecate the original method. It just didn't seem worthwhile.

I just did what I thought would have the smallest impact.

NathanQingyangXu: thanks for the explanation

sebersole:
One thing I should have mentioned... we should handle specially the case where the user has explicitly configured laziness (`@Proxy`, e.g.).  In those cases imo it is not enough to simply not build the ProxyFactory - there we really ought to throw an exception; or at the *very* lest log an error

Sanne:
> One thing I should have mentioned... we should handle specially the case where the user has explicitly configured laziness (`@Proxy`, e.g.). In those cases imo it is not enough to simply not build the ProxyFactory - there we really ought to throw an exception; or at the _very_ lest log an error

+1

I just happened to hit a related case: we fail to generate the proxy if the class is final; I believe it should fail in a similar way: a warning if it's acceptable, an explicit error if there is an explicit intent to mark it lazy.

gbadner:
@sebersole, I don't understand what you are saying:

> One thing I should have mentioned... we should handle specially the case where the user has explicitly configured laziness (`@Proxy`, e.g.). In those cases imo it is not enough to simply not build the ProxyFactory - there we really ought to throw an exception; or at the _very_ lest log an error

The [user guide](https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#entity-proxy) says:

"However, if the entity class is final, Javassist will not create a proxy and you will get a POJO even when you only need a proxy reference. In this case, you could proxy an interface that this particular entity implements, as illustrated by the following example."

Why would an exception be thrown or logged?

gbadner:
@sebersole, I just realized that `Proxy#proxyClass` is optional.

Are specifically saying that an exception should be thrown if `@Proxy` or `@Proxy(lazy=true)` is mapped for a final class without mapping `proxyClass`?

gbadner:
> I just happened to hit a related case: we fail to generate the proxy if the class is final; I believe it should fail in a similar way: a warning if it's acceptable, an explicit error if there is an explicit intent to mark it lazy.

I'm confused about the case that is "acceptable".

@sebersole, @Sanne, I would like to get HHH-13891 fixed, then follow that jira on with a new jira for to deal with final classes.

@sebersole, does this PR look ok to you to fix the limited case described in HHH-13891? If so, please approve this PR, and I'll create a new jira for the rest of the work.

sebersole:
> Why would an exception be thrown or logged?

Because `@Proxy` is the user specifically asking for laziness.  So they have explicitly asked for something that is not possible. Sounds like an exception to me...

gbadner:
Merged.

@sebersole, @Sanne, https://hibernate.atlassian.net/browse/HHH-13909 follows up on this.

PR #3288, opened 2020-03-07
Hhh 13893

cigaly:
Code change to handle problem described in https://hibernate.atlassian.net/browse/HHH-13893

PR #3289, opened 2020-03-07
Hhh 13893.v2

cigaly:
[Corrected mess in https://github.com/hibernate/hibernate-orm/pull/3288]

Code change to handle problem described in https://hibernate.atlassian.net/browse/HHH-13893


hibernate-core/src/main/java/org/hibernate/internal/util/config/ConfigurationHelper.java
```diff
@@ -220,8 +220,8 @@ public static long getLong(String name, Map values, int defaultValue) {
 		if ( value == null ) {
 			return defaultValue;
 		}
-		if ( Long.class.isInstance( value ) ) {
-			return (Long) value;
+		if ( Number.class.isInstance( value ) ) {
+			return ( (Number) value ).longValue();
```
NathanQingyangXu: This is a generic code change in configuration. unsure whether there is some unexpected implication. @gavinking maybe you would be interested to take a look.

hibernate-core/src/test/java/org/hibernate/test/cfg/NumberPropertyConversionTest.java
```diff
@@ -0,0 +1,24 @@
...
+import org.hibernate.testing.boot.ServiceRegistryTestingImpl;
+import org.hibernate.testing.junit4.BaseUnitTestCase;
+import org.junit.Test;
+
+public class NumberPropertyConversionTest extends BaseUnitTestCase {
+
+	@Test
+	public void testIntegerToLongConversion() {
+		ServiceRegistry sr = ServiceRegistryTestingImpl.forUnitTesting(
+				Collections.singletonMap( Environment.LOG_SLOW_QUERY, 25 )
```
NathanQingyangXu: Technically, using `Environment` is okay for it inherits `AvailableSettings`, but using `AvailableSettings.LOG_SLOW_QUERY` seems still better.
Also from testing perspective, you might go further than ensuring launching is not broken by adding the following assertions:

1. assert the config's original intended type is 'long', not 'int';
2. assert that the value after conversion is as intended (i.e. we received `25L`).

You also need to add `@TestForIssue( jiraKey = "HHH-13893" )` for this method.

hibernate-core/src/main/java/org/hibernate/internal/util/config/ConfigurationHelper.java
cigaly: Additionally, I will suggest to change type of defaultValue parameter from int to long.

hibernate-core/src/test/java/org/hibernate/test/cfg/NumberPropertyConversionTest.java
cigaly: This test is simplified "replication" of problem that I've met in one real situation. I think that it can be completely removed since suggested changes can be simply tested by directly testing ConfigurationHelper. I've created test case in PropertiesHelperTest that will, hopefully, better handle this issue.

gbadner:
@cigaly, this PR has a number of merge commits, and we don't allow merges commit in the Hibernate ORM repository.

@NathanQingyangXu, thank you for raising your concern about changing to use {{Number}} instead of {{Integer}}/{{Long}}. This change could have unintended consequences.

Javadoc for {{Number#intval}}  says:

"public abstract int intValue()
Returns the value of the specified number as an int, which may involve rounding or truncation.
Returns:
the numeric value represented by this object after conversion to type int."

[5.1.3 Narrowing Primitive Conversions](https://docs.oracle.com/javase/specs/jls/se7/html/jls-5.html#jls-5.1.3) applies, for example, when {{Number#intValue}} is called on a long value that is greater than {{Integer.MAX_VALUE}}. Narrowing such a value would result in a completely different integer value, without throwing an exception.

Hibernate should not be rounding, truncating, or narrowing configuration values.

cigaly:
@gbadner  : Although your comments are correct, the fact is that this is a bug that it is still present, and it will not go away by simply pretending that it does not exists. Therefore, I will create a new branch and new pull request and hope that it will pass less than ten months and fifteen versions before someone will look into it.

gbadner:
@cigaly, the jira is still open because it is a bug; no one is pretending that it does not exist. The proposed fix is not appropriate. That is why this PR was closed and the jira is still open.

A new PR would be appreciated. I cannot promise that this will be addressed quickly though, because there are many other issues that are of a higher priority.

PR #3291, opened 2020-03-09
Wip/6.0 SelectEagerCollectionInitializer

dreab8:
Given an `Order` with a eager collection of `Item`s
the `session.createQuery( "from Order", Order.class ).uniqueResult();`
should not create one Sql query with a join but 2 different queries.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/onetomany/OneToManyEmptyCollectionTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/onetomany/OneToManyEmptyCollectionTest.java
@@ -1,0 +1,136 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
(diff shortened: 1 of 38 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/onetomany/OneToManyEmptyCollectionTest.java
```diff
@@ -0,0 +1,136 @@
...
+		final StatisticsImplementor statistics = scope.getSessionFactory().getStatistics();
+		statistics.clear();
+		scope.inTransaction(
+				session -> {
+					Order order = session.createQuery( "from Order", Order.class ).uniqueResult();
+					assertThat( statistics.getPrepareStatementCount(), is( 2L ) );
+					List<Item> items = order.getEagerItems();
+					assertThat( items.size(), is( 0 ) );
+					assertThat( items, instanceOf( PersistentBag.class ) );
+					assertThat( items, instanceOf( PersistentBag.class ) );
```
NathanQingyangXu: seems duplicated testing statement above

sebersole:
What do you think about removing that `selected` variable and handling that concept in a different way?

dreab8:
@sebersole I'm open to suggestion, it would be great to completely remove the `selected` variable

dreab8:
merged

PR #3287, opened 2020-03-07
HHH-13888 Allow custom SQL loggers by subclassing SqlStatementLogger

ibre5041:
[https://hibernate.atlassian.net/browse/HHH-13888](https://hibernate.atlassian.net/browse/HHH-13888)

PR #3292, opened 2020-03-10
HHH-13894 - SQL query printing on WildFly

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13894

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/resource/jdbc/ResourceRegistry.java
+++ b/hibernate-core/src/main/java/org/hibernate/resource/jdbc/ResourceRegistry.java
@@ -37,19 +37,20 @@ public interface ResourceRegistry {
 	void register(Statement statement, boolean cancelable);
 
 	/**
-	 * Set the provided SQL {@link String} value to the given {@link Statement} that was registered previously
+	 * Set the provided SQL string value to the given {@link Statement} that was previously registered.
 	 *
 	 * @param statement JDBC Statement
-	 * @param sql SQL {@link String} value
+	 * @param sql SQL string value
 	 */
 	void setStatementSql(Statement statement, String sql);
 
 	/**
-	 * Get the SQL {@link String} value associated with the {@link Statement}, if any
+	 * Get the SQL string value associated with the given {@link Statement} or null if
+	 * the statement did not have the SQL string value registered as well.
 	 *
 	 * @param statement JDBC Statement
 	 *
-	 * @return SQL {@link String} value
+	 * @return SQL string value
 	 */
 	String getStatementSql(Statement statement);
 
--- a/hibernate-core/src/main/java/org/hibernate/resource/jdbc/ResourceRegistry.java
+++ b/hibernate-core/src/main/java/org/hibernate/resource/jdbc/ResourceRegistry.java
@@ -37,6 +37,23 @@ public interface ResourceRegistry {
 	void register(Statement statement, boolean cancelable);
 
 	/**
+	 * Set the provided SQL {@link String} value to the given {@link Statement} that was registered previously
+	 *
+	 * @param statement JDBC Statement
+	 * @param sql SQL {@link String} value
+	 */
+	void setStatementSql(Statement statement, String sql);
+
+	/**
+	 * Get the SQL {@link String} value associated with the {@link Statement}, if any
+	 *
+	 * @param statement JDBC Statement
+	 *
+	 * @return SQL {@link String} value
+	 */
+	String getStatementSql(Statement statement);
+
+	/**
 	 * Release a previously registered statement.
 	 *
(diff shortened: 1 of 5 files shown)


vladmihalcea:
@sebersole If you have some time, let me know if you agree with this approach of storing the SQL string value next to the `Statement` we are registering.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
```diff
@@ -370,4 +386,21 @@ private boolean hasRegistered(final HashMap resource) {
 	private boolean hasRegistered(final ArrayList resource) {
 		return resource != null && !resource.isEmpty();
 	}
+	
+	private class StatementRegistryEntry {
```
NathanQingyangXu: Seems the inner class should be static one for seemingly it doesn't need access to enclosing object.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/ResourceRegistry.java
```diff
@@ -36,6 +36,23 @@
...
+	 * @param sql SQL {@link String} value
+	 */
+	void setStatementSql(Statement statement, String sql);
+
+	/**
+	 * Get the SQL {@link String} value associated with the {@link Statement}, if any
+	 *
+	 * @param statement JDBC Statement
+	 *
+	 * @return SQL {@link String} value
```
NathanQingyangXu: Do we really need the `{@link String}` above? Seems it is so basic to deserve one.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
```diff
@@ -219,17 +233,19 @@ public void register(ResultSet resultSet, Statement statement) {
...
 		if ( statement != null ) {
-			HashMap<ResultSet,Object> resultSets = xref.get( statement );
+			final StatementRegistryEntry statementRegistryEntry = xref.remove( statement );
+			HashMap<ResultSet,Object> resultSets = statementRegistryEntry != null ?
+					statementRegistryEntry.resultSets : null;
 
 			// Keep this at DEBUG level, rather than warn.  Numerous connection pool implementations can return a
 			// proxy/wrapper around the JDBC Statement, causing excessive logging here.  See HHH-8210.
-			if ( resultSets == null ) {
+			if ( resultSets == null) {
```
NathanQingyangXu: Seems the above code change is not better than before. We do need a space char here.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
vladmihalcea: Good catch. It's totally better to be a nested one, especially for performance and its impact on GC.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/ResourceRegistry.java
vladmihalcea: True. I could simply replace it with a `string` instead.

hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
vladmihalcea: That's weird. I thought I ran a CTRL ALT L on those bocks, and the `check` command didn't complain either.

I'll fix all issues and push a new commit.

vladmihalcea:
@NathanQingyangXu I pushed a new commit to address the issues you found. Thanks for taking the time to review it.

NathanQingyangXu:
> @NathanQingyangXu I pushed a new commit to address the issues you found. Thanks for taking the time to review it.

Thanks. I have no further concern, but I guess Steve is the final reviewer so it won't matter whether I approve or not.

NathanQingyangXu:
btw, there is another related PR recently: https://github.com/hibernate/hibernate-orm/pull/3287. Seems the 'slowQuery' feature is really popular.

nicolas-tg-ch:
Any plans to merge this soon in Hibernate 5.4/5.5? I have the same problem on Websphere Liberty, see https://stackoverflow.com/questions/68815550/hibernates-slow-sql-queries-in-websphere-liberty

(I had many problems to login/sign up on hibernate.atlassian.net that's why I post here.)

gavinking:
This problem was already fixed in a different PR by @NathanQingyangXu.

PR #3295, opened 2020-03-10
HHH-13619 - Support for JPA's `size` function as a select expression

gbadner:
- Fix to work on Oracle by removing "as" between table name and alias

@sebersole or @dreab8, this works for Oracle. I thought there was some `Dialect` method that indicates if "as" is needed between table name and alias, but I don't see it.

Is there one that I missed?

Do you know if this will break for any dialects?


sebersole:
Wow, Oracle still doesn't support AS for defining a table alias?  Crazy.  The crazy thing is it treats AS differently for select expressions versus table expressions.  Consistency FTW.

ANSI SQL defines the AS as optional and all the databases I know of support *not* using it.  It is a shame because it is generally considered more readable.

gavinking:
> Wow, Oracle still doesn't support AS for defining a table alias?

Nope. Amazing, huh?

PR #3298, opened 2020-03-12
Wip/6.0 HHH 13756 Entity graph implementation

NathanQingyangXu:
https://trello.com/c/d9dirY4C/59-entitygraph-support
https://hibernate.atlassian.net/browse/HHH-13756

Testing cases for loader, HQL and Criteria are added. To avoid getting lost in ANTLR tree forest, many util method was created. Because of our testing mechanism, JUnit5's `@Nest` and `parameterized testing` features cannot be tapped into to avoid duplication, however.

One bug flushed out (not directly related to entity graph) and a PR has been created: https://github.com/hibernate/hibernate-orm/pull/3297


hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -366,36 +382,33 @@ private void applyOrdering(TableGroup tableGroup, PluralAttributeMapping pluralA
...
 		final List<Fetch> fetches = new ArrayList<>();
 
-		final Consumer<Fetchable> processor = createFetchableConsumer( fetchParent, querySpec, creationState, fetches );
+		final BiConsumer<Fetchable, Boolean> processor = createFetchableBiConsumer( fetchParent, querySpec, creationState, fetches );
 
 		final FetchableContainer referencedMappingContainer = fetchParent.getReferencedMappingContainer();
-		referencedMappingContainer.visitKeyFetchables( processor, null );
-		referencedMappingContainer.visitFetchables( processor, null );
+		referencedMappingContainer.visitKeyFetchables( fetchable -> processor.accept( fetchable, true ), null );
+		referencedMappingContainer.visitFetchables( fetchable -> processor.accept( fetchable, false ), null );
```
NathanQingyangXu: The boolean parameter to denote whether we are exploring 'keySubGraph' is important to disambiguate in entity graph implementing, so I changed `Consumer` to `BiConsumer`.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/StandardSqmSelectTranslator.java
```diff
@@ -86,7 +85,9 @@
 	// prepare for 10 root selections to avoid list growth in most cases
 	private final List<DomainResult> domainResults = CollectionHelper.arrayList( 10 );
 
-	private GraphImplementor<?> currentJpaGraphNode;
+	private final EntityGraphNavigator entityGraphNavigator;
+
+	private int fetchDepth;
```
NathanQingyangXu: I moved the field here for best practice. Also, it seems we don't need to set default value here.

NathanQingyangXu:
> Unless I miss something, `EntityGraphSemanticTraverser` is not really traversing anything - it is more a helper for things that do traversals. Am I wrong? Not a major deal, just comes down to a question of "good" naming. Anyway, going to apply - we can always adjust naming later

I struggled a lot about the naming. The last version is 'EntityGraphSemanticsEnforcer', but I like the `backtrack` method name and it easily leads to `traversal` naming.
TBH, I found some imperfection in the current implementation but I can fix it later. Thanks for approval.

sebersole:
`EntityGraphTraversalState`?

sebersole:
`EntityGraphTraversalContext`?

NathanQingyangXu:
I will adopt `EntityGraphTraversalContext` later when I need to fix some imperfection I found recently. Thanks for the suggestion.

PR #3300, opened 2020-03-14
HHH-13882 Fix bug in EntityGraphQueryHint

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13882

The bug manifests itself with the following preconditions:

1. entity graph is used on collection
2. LockMode is specified for query

The root cause is we set 'classAlias' of LHS to RHS in EntityGraphQueryHint. Luckily EntityGraphQueryHint will be refactored in v6 so this PR only applies for v5.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/query/spi/EntityGraphQueryHint.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/query/spi/EntityGraphQueryHint.java
@@ -18,7 +18,6 @@ import javax.persistence.Subgraph;
 import org.hibernate.QueryException;
 import org.hibernate.engine.internal.JoinSequence;
 import org.hibernate.graph.GraphSemantic;
-import org.hibernate.graph.RootGraph;
 import org.hibernate.graph.spi.AppliedGraph;
 import org.hibernate.graph.spi.RootGraphImplementor;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/engine/query/spi/EntityGraphQueryHint.java
```diff
@@ -143,7 +141,7 @@ else if ( propertyType.isCollectionType() ) {
 
 						final FromElementFactory fromElementFactory = new FromElementFactory(
 								fromClause, origin,
-								attributeName, classAlias, columns, false
+								attributeName, null, columns, false
```
NathanQingyangXu: Obviously, we should not use `classAlias` here for it is the joining lhs and using it only ends up with screwing up internal `formElementByClassAlias` map in `FromClause` by overwriting lhs by a collection from element without persister, thus the NPE bug when `QueryLoader#applyLocks()` is invoked.

wgtmpeters:
Will this ever be merged or is there a workaround?

NathanQingyangXu:
> Will this ever be merged or is there a workaround?

This is a simple and effective fix. @beikov do you have bandwidth to take a look?

beikov:
Could you please rebase this PR @NathanQingyangXu ?

PR #3301, opened 2020-03-14
HHH-13889 Fix a literal escaping issue in LiteralExpression

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13889

Seems an easy fix; the testing case provided in the ticket was commited


dreab8:
Thanks @NathanQingyangXu , applied upstream

PR #3302, opened 2020-03-15
Eliminate some raw types

marschall:
Eliminate some raw types and add some generics.


marschall:
The code base as quite a few raw types, this PR fixes some of them. I payed attention not to change any signatures of any public non-internal members.

I don't have a JIRA for this, I can open one if required.

NathanQingyangXu:
> The code base as quite a few raw types, this PR fixes some of them. I payed attention not to change any signatures of any public non-internal members.
> 
> I don't have a JIRA for this, I can open one if required.

I did consider the similar thing. Yeah, the codebase has tons of generics issue because of its long history.

PR #3304, opened 2020-03-19
Follow up on PR #2461

sebersole:
Follow up on PR #2461 moving tests to new test package (effectively enabling them as part of the build).  ATM these tests all fail due to `NativeQuery` support and `import.sql` support not yet being implemented.

NOTE : atm this is not yet ready for merging until support for `NativeQuery` and `import.sql` handling is implemented


sebersole:
Major conflicts

PR #3305, opened 2020-03-20
Wip/6.0 Entity Graph implementation followup

NathanQingyangXu:
As Steve suggested, `EntityGraphSemanticsTraverser` is renamed to `EntityGraphTraversalState` (I think it is more in line with existing naming convention compared to `EntityGraphTraversalContext`).

Another improvement is to update user guide, including:

1. remove outdated content only applicable when `fetch graph` is not fully implemented;
2. add explanation about the difference between `fetch graph` and `load graph`;
3. correct an erroneous explanation of `SELECT` fetch mode;
4. revert back incompatible v5 fetch graph code changes;
4. minor verbiage tweaking.

Also, I cleaned up some duplicated code found during the testing.

### DIFF
--- a/documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc
+++ b/documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc
@@ -45,6 +45,12 @@ _dynamic_ (sometimes referred to as runtime)::
     HQL / JPQL::: both Hibernate and JPA Criteria queries have the ability to specify fetching, specific to said query.
     entity graphs::: starting in Hibernate 4.2 (JPA 2.1), this is also an option.
 
+[NOTE]
+====
+`fetch profile` and `entity graph` are mutually exclusive. When both are present, `entity graph` would take effect
+and `fetch profile` would be ignored.
+====
+
 [[fetching-direct-vs-query]]
 === Direct fetching vs. entity queries
 
@@ -77,7 +83,7 @@ include::{extrasdir}/fetching-direct-vs-query-direct-fetching-example.sql[]
 ----
 ====
 
-The `LEFT JOIN` clause is added to the generated SQL query because this association is required to be fetched eagerly.
+The `LEFT OUTER JOIN` clause is added to the generated SQL query because this association is required to be fetched eagerly.
 
 On the other hand, if you are using an entity query that does not contain a `JOIN FETCH` directive to the `Department` association:
 
@@ -101,7 +107,7 @@ so Hibernate requires a secondary select to ensure that the EAGER association is
 [IMPORTANT]
 ====
 If you forget to JOIN FETCH all EAGER associations, Hibernate is going to issue a secondary select for each and every one of those
-which, in turn, can lead to N+1 query issues.
+which, in turn, can lead to N + 1 query issue.
 
 For this reason, you should prefer LAZY associations.
 ====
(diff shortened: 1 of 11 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/PluralAttributeMappingImpl.java
```diff
@@ -382,14 +382,6 @@ public Fetch generateFetch(
...
 									creationState.getSqlAstCreationState().getCreationContext()
 							);
-
-							lhsTableGroup.addTableGroupJoin( tableGroupJoin );
-
-							sqlAstCreationState.getFromClauseAccess().registerTableGroup(
-									fetchablePath,
-									tableGroupJoin.getJoinedGroup()
-							);
-
```
NathanQingyangXu: @dreab8 the above code seems duplicated. The first one has been done in the constructor above, and the second one will be done in the external method. What do you think? Duplicating concern will confuse code reader and end up with potential maintenance difficulty.

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
```diff
@@ -916,7 +916,7 @@ private static Ejb3DiscriminatorColumn processJoinedDiscriminatorProperties(
 			final boolean generateDiscriminatorColumn;
 			if ( discriminatorColumnAnnotation != null ) {
 				if ( context.getBuildingOptions().ignoreExplicitDiscriminatorsForJoinedInheritance() ) {
-					LOG.debugf( "Ignoring explicit DiscriminatorColumn annotation on ", clazzToProcess.getName() );
+					LOG.debugf( "Ignoring explicit DiscriminatorColumn annotation on: %s", clazzToProcess.getName() );
```
NathanQingyangXu: This bug was found by accident

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/PluralAttributeMappingImpl.java
dreab8: the code should be removed, it was a mistake to add this code, good catch

documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc
```diff
@@ -204,14 +217,6 @@ include::{sourcedir}/GraphFetchingTest.java[tags=fetching-strategies-dynamic-fet
...
 ====
 
-[NOTE]
-====
-When executing a JPQL query, if an EAGER association is omitted, Hibernate will issue a secondary select for every association needed to be fetched eagerly,
-which can lead to N+1 query issues.
-
-For this reason, it's better to use LAZY associations, and only fetch them eagerly on a per-query basis.
-====
-
```
NathanQingyangXu: The above content is outdated for it only applies when `fetch graph` is not correctly implemented.

documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc
```diff
@@ -447,12 +452,11 @@ Besides the `FetchType.LAZY` or `FetchType.EAGER` JPA annotations,
 you can also use the Hibernate-specific `@Fetch` annotation that accepts one of the following ``FetchMode``s:
 
 SELECT::
-	The association is going to be fetched lazily using a secondary select for each individual entity,
-	collection, or join load.
-	It's equivalent to JPA `FetchType.LAZY` fetching strategy.
+	The association is going to be fetched using a secondary select for each individual entity,
+	collection, or join load. This mode can be used for either `FetchType.EAGER` or `FetchType.LAZY`.
```
NathanQingyangXu: obviously the original content is wrong for `SELECT` is mostly used for `EAGER` loading, not restrictive to `LAZY` loading.

NathanQingyangXu:
found new things undone. closed temporarily.

dreab8:
applied upstream, Thanks @NathanQingyangXu.

PR #3306, opened 2020-03-20
Re-enable hibernate-osgi, hibernate-orm-modules

sebersole:
Really does 3 things:

1. re-enables the hibernate-osgi module
2. re-enabled the hibernate-orm-modules module
3. disables the hibernate-envers module


sebersole:
@Naros I added you here as a heads up because I ended up completely disabling the hibernate-envers module.  The build "works" but it produces an artifact that simply does not work atm.

sebersole:
Had to removal addditional references to hibernate-envers (hibernate-orm-modules)

PR #3307, opened 2020-03-20
Wip/6.0 Fix ToOne @Fetch(FetchMode.SELECT) 

dreab8:
Querying the parent a select was executed to select the parent but not a second one for the association annotated with `@Fetch(FetchMode.SELECT), the issue was that a `EntityFetchDelayedInitializer` was created for the association instead of a `EntityInitializerSelectFetch` because the` fetchTiming` was equal to `FetchTiming.DELAYED`

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
@@ -34,6 +34,7 @@ import org.hibernate.mapping.Component;
 import org.hibernate.mapping.IndexedCollection;
 import org.hibernate.mapping.KeyValue;
 import org.hibernate.mapping.OneToMany;
+import org.hibernate.mapping.OneToOne;
 import org.hibernate.mapping.PersistentClass;
 import org.hibernate.mapping.Property;
 import org.hibernate.mapping.Selectable;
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
```diff
@@ -1061,8 +1061,14 @@ public static SingularAssociationAttributeMapping buildSingularAssociationAttrib
...
 
-		final FetchTiming fetchTiming = FetchStrategyHelper.determineFetchTiming( fetchStyle, type, sessionFactory );
+		final FetchTiming fetchTiming;
 
+		if ( !( (ToOne) bootProperty.getValue() ).isLazy() ) {
+			fetchTiming = FetchTiming.IMMEDIATE;
+		}
+		else {
+			fetchTiming = FetchStrategyHelper.determineFetchTiming( fetchStyle, type, sessionFactory );
+		}
```
NathanQingyangXu: Feel free to ignore my comment. It would be great to centralize all the fetch timing decision logic into `FetchStrategyHelper` for code maintaining purpose.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
```diff
@@ -1061,8 +1061,14 @@ public static SingularAssociationAttributeMapping buildSingularAssociationAttrib
 						sessionFactory
 				);
 
-		final FetchTiming fetchTiming = FetchStrategyHelper.determineFetchTiming( fetchStyle, type, sessionFactory );
+		final FetchTiming fetchTiming;
 
+		if ( !( (ToOne) bootProperty.getValue() ).isLazy() ) {
```
NathanQingyangXu: I think we cannot take it for granted that the downcast is always valid. Seems `ToOne` is only one child class type for `AssociationType`. We might well fix the issue by `instanceof` checking first but it would be great to fix loophole for other association type.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
```diff
@@ -1061,8 +1061,14 @@ public static SingularAssociationAttributeMapping buildSingularAssociationAttrib
...
+			fetchTiming = FetchTiming.IMMEDIATE;
```
NathanQingyangXu: There is a subtlety here. Previously when fetch mode is `JOIN`, IMMEDIATE is always returned; now if `FetchType.LAZY` is specified together with `FetchMode.JOIN`, IMMEDIATE won't be returned.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
sebersole: Yes that is always a safe bet *now*.  To be determined what happens when we implement ANY support

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
sebersole: That is true.  Each can affect the other.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
sebersole: I don't think that is really possible.  Happy to be proven wrong because, as you say, it would make it better in all the ways that centralizing code does.

sebersole: To be clear, anything is *possible*, sure.  The question is whether centralizing this make the code more simple or more complex. Specifically does centralizing it mean passing more arguments?  How many more?  There is a tipping point.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
sebersole: Oh, wait...

This is while determining the *mapped* fetching strategy.  I missed that part when I replied.  I *think* this is actually safe here.  @NathanQingyangXu do you have a specific case you are thinking of where this would happen?

@dreab8 Another consideration would be enhanced proxies.  I forget how we ended up implementing that wrt `#isLazy`.  Perhaps you could add a test for that as well.

dreab8: I realised the change I made cause for sure an issue in case of `OneToOne` when the value is nullable in sucha a case the `fetchTiming` should be Immediate no matter if the fetch type is set lazy.
Probably the code should be
```
if ( ( value instanceof OneToOne && value.isNullable() ) || !( value ).isLazy() ) {
			fetchTiming = FetchTiming.IMMEDIATE;
		}
...
```

dreab8:
@sebersole I agree with you about exploring a bit more the "subtlety" as I realised that it is not completely clear to me how the fetch mode should influence the fetch timing.

dreab8:
Added more tests and fixed the code

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
NathanQingyangXu: I am thinking of the following scenario:
```
@Fetch(FetchMode.JOIN)
private SimpleEntity manyToOne;
```
The code change in this PR will treat it as 'lazy', but from user guide, the `join mode' will lead to 'eager'. I didn't explore it fully, but I am not sure this scenario will lead to some invariance violation or even runtime exception?

NathanQingyangXu: however, it would be great to test the above scenario behaves the same between v5 and v6. Maybe I am wrong, but...

dreab8: I added a test also for this case, basically I run all the tests on v5 and then ported them on v6

dreab8:
Thanks @sebersole and @NathanQingyangXu , applied upstream

PR #3308, opened 2020-03-23
wip/6.0 Loader 'readOnly' support

NathanQingyangXu:
The motivation of this PR is the fact that https://hibernate.atlassian.net/browse/HHH-11958 is not working in v6 because of code refactoring.
However, it is not difficult to fix it in v6. This PR doesn't include CollectionLoader into the scope, however.
A testing case is included.


NathanQingyangXu:
good point. I will try to change as per your suggestion. Thanks.

sebersole:
Not necessarily saying that is the right thing.  I'm going to go ahead and apply this.  We can come back to the parameter-object discussion later if we want

PR #3309, opened 2020-03-24
HHH-13900: added keepReference option to SQLDelete

richardbizik:
keepReference options allows hibernate to keep references for objects
that are being deleted. This behavior is wanted for soft delete
operations.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultDeleteEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultDeleteEventListener.java
@@ -281,7 +281,9 @@ public class DefaultDeleteEventListener implements DeleteEventListener,	Callback
 
 		new ForeignKeys.Nullifier(  entity, true, false, session, persister ).nullifyTransientReferences( entityEntry.getDeletedState() );
 		new Nullability( session ).checkNullability( entityEntry.getDeletedState(), persister, Nullability.NullabilityCheckType.DELETE );
-		persistenceContext.registerNullifiableEntityKey( key );
+		if( persister.getClassMetadata() != null && !persister.getClassMetadata().keepReference() ) {
+			persistenceContext.registerNullifiableEntityKey(key);
+		}
 
 		if ( isOrphanRemovalBeforeUpdates ) {
 			// TODO: The removeOrphan concept is a temporary "hack" for HHH-6484.  This should be removed once action/task
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultDeleteEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultDeleteEventListener.java
@@ -281,8 +281,8 @@ public class DefaultDeleteEventListener implements DeleteEventListener,	Callback
 
 		new ForeignKeys.Nullifier(  entity, true, false, session, persister ).nullifyTransientReferences( entityEntry.getDeletedState() );
 		new Nullability( session ).checkNullability( entityEntry.getDeletedState(), persister, Nullability.NullabilityCheckType.DELETE );
-		if( persister.getClassMetadata() != null && !persister.getClassMetadata().keepReference() ) {
-			persistenceContext.registerNullifiableEntityKey(key);
+		if ( persister.getClassMetadata() != null && !persister.getClassMetadata().keepReference() ) {
+			persistenceContext.registerNullifiableEntityKey( key );
 		}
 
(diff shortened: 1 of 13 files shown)


hibernate-core/src/main/java/org/hibernate/event/internal/DefaultDeleteEventListener.java
```diff
@@ -281,7 +281,9 @@ protected final void deleteEntity(
 
 		new ForeignKeys.Nullifier(  entity, true, false, session, persister ).nullifyTransientReferences( entityEntry.getDeletedState() );
 		new Nullability( session ).checkNullability( entityEntry.getDeletedState(), persister, Nullability.NullabilityCheckType.DELETE );
-		persistenceContext.registerNullifiableEntityKey( key );
+		if( persister.getClassMetadata() != null && !persister.getClassMetadata().keepReference() ) {
```
NathanQingyangXu: minor. A space should be inserted between `if` and `(`

hibernate-core/src/test/java/org/hibernate/test/annotations/delete/keepreference/DeathStar.java
```diff
@@ -0,0 +1,56 @@
...
+public class DeathStar extends BaseEntity {
+
+    @NotNull
+	@OneToOne(optional = false, cascade = CascadeType.ALL, fetch = FetchType.EAGER)
+	private Vader vader;
+	@OneToMany(cascade = CascadeType.MERGE, orphanRemoval = true, mappedBy = "deathStar",fetch = FetchType.EAGER)
+	private Set<Trooper> troppers;
+    @NotNull
+    @OneToOne(optional = false, fetch = FetchType.EAGER)
+	private Universe universe;
```
NathanQingyangXu: there is vertical alignment issue above. Not a big deal though.

hibernate-core/src/test/java/org/hibernate/test/annotations/delete/keepreference/KeepReferenceTest.java
```diff
@@ -0,0 +1,129 @@
...
+import static org.junit.Assert.*;
+
+/**
+ * @author Richard Bizik
+ */
+public class KeepReferenceTest extends BaseCoreFunctionalTestCase {
+
+
+	@Test
+	@TestForIssue( jiraKey = "HHH-XXXXX")//TODO: assign jira key
```
NathanQingyangXu: I think we can eliminate this TODO for you have created the ticket

hibernate-core/src/test/java/org/hibernate/test/annotations/delete/keepreference/DeathStar.java
richardbizik: I replaced all spaces by tabs in test package

sebersole (changes requested):
I reverted your change, specifically in the delete listener, and re-ran your test - it passes.  Really need a test that shows this failing before your change and then passing with the change.

I also thought the test was confusing.  It may even have been styled after some of our own legacy tests, but those are confusing as well ;)

I also think we should just call this "soft delete" rather than "keep references" or "nullifiable".  At the very least, we should pick one or the other and use it consistently.  

I created 3659 on top of yours.  There are 2 commits...

The first reverts the delete listener change, so you can verify its not the test changes from my second commit.

The second commit (1) standardizes on calling this "soft delete" and (2) re-organizes the tests to be more clear (imo).

But we still need a test reproducing the the error without your change.

richardbizik:
Hi, thanks for getting back to me. I tried your changes and you are right. Unfortunately its been quite some time and I cannot remember what was the exact setup when this happened to us and I could not reproduce it in tests. I only remember that this was the method that set the reference to null org.hibernate.engine.internal.ForeignKeys.Nullifier#nullifyTransientReferences
I hope I will have time to reproduce the issue in near future.

hibernate-core/src/main/java/org/hibernate/tuple/entity/EntityMetamodel.java
```diff
@@ -139,6 +140,7 @@ public EntityMetamodel(
 		);
 
 		versioned = persistentClass.isVersioned();
+		keepReference = persistentClass.isCustomSQLDeleteNullifiable();
```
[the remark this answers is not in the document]
richardbizik: Nullifiable was just a naming taken from persistence context. `soft=true` sounds better

gavinking:
Hibernate now has dedicated support for soft delete.

PR #3312, opened 2020-03-24
import.sql handling

sebersole:
fixed grammar and script parsing.  still need to test inclusion in create/drop handling.  

marking draft for now


hibernate-core/src/test/java/org/hibernate/orm/test/grammar/importsql/SchemaManagementScriptTests.java
```diff
@@ -0,0 +1,95 @@
...
+		assertThat( extractedCommands[0], is( FIRST ) );
+		assertThat( extractedCommands[1], is( SECOND ) );
+		assertThat( extractedCommands[2], is( THIRD ) );
+	}
+
+	@Test
+	public void testMultiLineExtraction() {
+		final String firstCommand = "first \nstatement";
+		final String secondCommand = "second \nstatement";
+		final String thirdCommand = "third \nstatement";
```
NathanQingyangXu: It would be more robust to replace `\n` with `System.lineSeparator()`, so we can ensure Windows system is covered.

sebersole:
Good point, even though no one cares about Windows :laughing:

hibernate-core/src/test/java/org/hibernate/orm/test/grammar/importsql/SchemaManagementScriptTests.java
sebersole: Good point, even though no one cares about Windows :laughing:

NathanQingyangXu: It might not be a big deal for testing, but it does pose a big problem elsewhere. Windows user would see very weird sqm tree logging result for sure (see the usage of `\n` in (e.g. SqmTreePrinter)). I just thought it is worthwhile to raise awareness here.

PR #3316, opened 2020-03-27
wip/6.0 Refactor Dialect to include version field

NathanQingyangXu:
Hibernate is an ORM library and should be a role model of OO design. 
Seems the Dialect hierarchy can benefit a lot from moving the common 'version' field to the topmost class, i.e. Dialect.

Feel free to postpone merging it if it conflicts with other PRs. Just another cosmetic improvement but I think it is worthwhile.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -23,6 +23,7 @@ import org.hibernate.dialect.unique.DefaultUniqueDelegate;
 import org.hibernate.dialect.unique.UniqueDelegate;
 import org.hibernate.engine.jdbc.LobCreator;
 import org.hibernate.engine.jdbc.Size;
+import org.hibernate.engine.jdbc.dialect.spi.DialectResolutionInfo;
 import org.hibernate.engine.jdbc.env.internal.DefaultSchemaNameResolver;
 import org.hibernate.engine.jdbc.env.spi.*;
 import org.hibernate.engine.jdbc.spi.JdbcServices;
@@ -135,10 +136,12 @@ public abstract class Dialect implements ConversionContext {
 
 	private DefaultSizeStrategy defaultSizeStrategy;
 
+	private final Integer version;
+
 
 	// constructors and factory methods ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-	protected Dialect() {
+	protected Dialect(Integer version) {
 
 		registerColumnType( Types.BIT, 1, "bit" );
 		registerColumnType( Types.BIT, "bit($l)" );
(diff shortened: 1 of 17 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/CUBRIDDialect.java
```diff
@@ -36,7 +36,6 @@
 	 * Constructs a CUBRIDDialect
 	 */
 	public CUBRIDDialect() {
-		super();
```
NathanQingyangXu: The above statement is unnecessary for JVM has done that for us automatically

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -135,10 +136,12 @@
 
 	private DefaultSizeStrategy defaultSizeStrategy;
 
+	private final Integer version;
```
NathanQingyangXu: wrapper type is chosen for some dialect doesn't support 'version' and we store `null` value to denote that.

hibernate-core/src/main/java/org/hibernate/dialect/CUBRIDDialect.java
sebersole: It is done automatically, yes.  All of us have developed Java for more than a day ;)

But:
1) imo this is exactly the kind of change that cause unnecessary conflicts (right?)
2) often these are nice for debug breakpoints
3) there is no reason to not have them.  there is no performance penalty

sebersole:
Why are we making this change?  With 6.0 the point is to make Dialects non-version specific.  Yes we will still have (deprecated) version specific impls, but, well, they are deprecated.

Can you explain why you think the change is needed or even worthwhile?

NathanQingyangXu:
> Why are we making this change? With 6.0 the point is to make Dialects non-version specific. Yes we will still have (deprecated) version specific impls, but, well, they are deprecated.
> 
> Can you explain why you think the change is needed or even worthwhile?

It simply avoids code duplication for most of dialects need version info. The goal of 'version agnostic' is based on maintaining 'version' field in Dialect, so the refactoring doesn't conflict with the goal.
However, if you persist, I am okay to close this one. I agree it is not mandatory. Repeating version related code in every Dialect seems not a big deal.

hibernate-core/src/main/java/org/hibernate/dialect/CUBRIDDialect.java
NathanQingyangXu: Thanks for the explanation. At least point 2 makes sense. I reverted back the unnecessary changes.

NathanQingyangXu:
Given the refactoring is not mandatory (duplicating version related code in each child class of Dialect seems not a big deal) and it might easily lead to conflict, I closed this PR.

PR #3317, opened 2020-03-29
wip/6.0 Fix some obvious logging bugs

NathanQingyangXu:
some easily verified logging bugs found recently

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchingBatch.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchingBatch.java
@@ -78,7 +78,7 @@ public class BatchingBatch extends AbstractBatchImpl {
 		}
 		catch ( SQLException e ) {
 			abortBatch();
-			LOG.debugf( "SQLException escaped proxy", e );
+			LOG.debug( "SQLException escaped proxy", e );
 			throw sqlExceptionHelper().convert( e, "could not perform addBatch", currentStatementSql );
 		}
 		statementPosition++;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchingBatch.java
```diff
@@ -78,7 +78,7 @@ public void addToBatch() {
 		}
 		catch ( SQLException e ) {
 			abortBatch();
-			LOG.debugf( "SQLException escaped proxy", e );
+			LOG.debug( "SQLException escaped proxy", e );
```
NathanQingyangXu: The bug here is the exception will end up as format parameter, not as root cause. Fixed it to align with other logging statements in the same class.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/DefaultSchemaNameResolver.java
```diff
@@ -61,8 +61,8 @@ private SchemaNameResolver determineAppropriateResolverDelegate(Connection conne
 		}
 		catch (Exception ignore) {
 			log.debugf(
-					"Unable to use Java 1.7 Connection#getSchema : An error occurred trying to resolve the connection default schema resolver: "
-							+ ignore.getMessage() );
+					"Unable to use Java 1.7 Connection#getSchema : An error occurred trying to resolve the connection default schema resolver: %s",
+					ignore.getMessage() );
```
NathanQingyangXu: This bug is more subtle. But imagine what would happen when the `ignore.getMessage()` contains `%s`. So it is still sort of bomb to explode at runtime.

dreab8:
applied upstream, Thanks @NathanQingyangXu

PR #3319, opened 2020-03-30
follow-up on HHH-12950

sebersole:
- fix annotation-processor set up
- does not work atm - we need Gradle 5.2 and upgrading causes some issues with the plugins we use which need addressed (e.g. the matrix-testing plugin)

See https://stackoverflow.com/questions/42441844/annotation-processor-in-intellij-and-gradle/54611475#54611475

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ buildscript {
 }
 
 plugins {
-	id 'com.gradle.build-scan' version '1.16'
+	id 'com.gradle.build-scan' version '2.0.2'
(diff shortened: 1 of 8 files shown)


sebersole:
Marked as draft until I can move past the plugin Gradle versions issue

sebersole:
I'm going to close this.  Let's wait until after IntelliJ 2020.1.1 is released[1] and after we convert to use of `java-library` plugin rather than the more basic `java` plugin

[1] https://youtrack.jetbrains.com/issue/IDEA-235948

PR #3321, opened 2020-03-30
HHH-13915 : Prevent state sharing in basic proxies.

bighenry:
Basic proxies were being created with a shared interceptor.  As the interceptor is stateful, this lead to cross-thread data leakage and thus intermittently broken persistence.

This mod creates an interceptor per proxy, preventing leaking of state.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BasicProxyFactoryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BasicProxyFactoryImpl.java
@@ -24,10 +24,11 @@ import net.bytebuddy.dynamic.scaffold.subclass.ConstructorStrategy;
 public class BasicProxyFactoryImpl implements BasicProxyFactory {
 
 	private static final Class[] NO_INTERFACES = new Class[0];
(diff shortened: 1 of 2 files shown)


bighenry:
This PR is for discussion purposes only.  The fix addresses the issue described in my application, but I've not yet been able to get a test configuration to mimic the break.

The issue is that the test code does not lead to the use of a `BasicProxyFactoryImpl` for the composite unique constraint, whereas in my application this is the case.

Digging through debug stack traces, I suspect this is because in my application Hibernate is representing the constraint in the `entityMetamodel` as an:
-   `EntityBasedCompositionAttribute` : `Attribute(name=unique-props, type=component[guid,customer] [non-identifier,composition])` 
Whereas in my test code it is a:
- `EntityBasedAssociationAttribute` : Attribute(name=discriminator, type=org.hibernate.test.uniqueproperties.UniquePropertiesTest$Discriminator [non-identifier,association])`

Any thoughts on how to change the test configuration to have it exercise a `BasicProxyFactoryImpl` would be greatly appreciated.

(Note: the commented out test was an initial PoC on test setup - I will clean this up having resolved the above)

bighenry:
Note: formatting changes were the result of applying the Hibernate formatter in IntelliJ.

jrenaat:
Superseded by https://github.com/hibernate/hibernate-orm/pull/9598, closing this.

PR #3322, opened 2020-03-31
HHH-13916 : Pending-put should not be created by Infinispan when chec…

gbadner:
…king if an entity is transient

https://github.com/gbadner/hibernate-core/pull/new/HHH-13916

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
@@ -4848,10 +4848,20 @@ public abstract class AbstractEntityPersister
 
 		// check to see if it is in the second-level cache
 		if ( session.getCacheMode().isGetEnabled() && canReadFromCache() ) {
+
+			// We don't care about the cached value here; we only want to know
+			// if the entity exists in the cache. If it does, then we know it is
+			// detached.
+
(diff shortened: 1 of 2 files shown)


gbadner:
An alternative is to have `AbstractEntityPersister#isTransient` call a new method,  `CacheHelper#existsInCache`, which would call `CachedDomainDataAccess#get`.

gbadner:
I'll look into the failures tomorrow...

gbadner:
The test failures appeared to be due to the test not cleaning up after itself properly. I pushed a commit to fix the test.

dreab8:
What It is strange is that the tests were not failing before your change.

dreab8:
@gbadner  I tried to remove the `deleteData()` you added and I changed the `init()` method adding at the beginning the checks
```
	inTransaction(
				session -> {
					List<Node> node = session.createQuery( "from Node").list();
					assertTrue( node.isEmpty() );
				}
		);
.....

```
the assertions is always satisfied but when i run all the tests  the `init` methos of `testOneToManyMoveElementWithDirtySimpleProperty` fails with `javax.persistence.OptimisticLockException: Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect) : [org.hibernate.test.onetomany.Node#2] `

gbadner:
@dreab8, I was finally able to dig into this again, and I see that this approach will not work. Thanks for pushing me to investigate further. I'll send an email with more details.

PR #3323, opened 2020-03-31
HHH-13917: Add support for HANA Cloud

breglerj:
- Add new HANA Cloud dialect
- Adapt Databases enum
- Adapt failing tests

### DIFF
--- a/documentation/src/test/java/org/hibernate/userguide/mapping/basic/SubselectTest.java
+++ b/documentation/src/test/java/org/hibernate/userguide/mapping/basic/SubselectTest.java
@@ -236,10 +236,10 @@ public class SubselectTest extends BaseEntityManagerFunctionalTestCase {
 		"select " +
 		"	a.id as id, " +
 		"	concat(concat(c.first_name, ' '), c.last_name) as clientName, " +
-		"	sum(at.cents) as balance " +
+		"	sum(atr.cents) as balance " +
 		"from account a " +
 		"join client c on c.id = a.client_id " +
-		"join account_transaction at on a.id = at.account_id " +
+		"join account_transaction atr on a.id = atr.account_id " +
 		"group by a.id, concat(concat(c.first_name, ' '), c.last_name)"
 	)
 	@Synchronize( {"client", "account", "account_transaction"} )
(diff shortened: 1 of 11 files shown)


breglerj:
@Sanne: there is no automatic upgrade from HANA to HANA Cloud, because of some feature incompatibilities. Users will either have to start fresh with HANA Cloud or they will have to migrate their existing persistence manually before switching to HANA Cloud. In that sense the added dialect is more like a new database type instead of a new database version. I'd be fine with including it in a minor version so users can get started quickly.

Sanne:
ok, great!

Sanne:
Hi @breglerj,
I've merged the PR. Could you check if there's any need to update documentation?

Regarding versions: this was merged in master, which implies it will be included in `5.5.0.Beta1` as next release. This version will be JPA 3.0, so I expect some resistance from people: many will stay on older versions for a while as they won't want to upgrade right away.
We can of course backport this to 5.4 and or 5.3 - up to you.

But most importantly, we will need this ported to the 6 branch as well. The Dialect(s) changed quite a bit, so we will need your help for that as cherry-picking this commit on the 6 branch won't be trivial. Could you coordinate with @dreab8 to make that happen?

Thanks again!

breglerj:
Hi @Sanne,

Thanks for merging the PR.

I think it would be worth backporting it to 5.4 to have an alternative for people not willing or able to use JPA 3.0. Do you want me to create another PR for the backport, or will you do it?

I'm happy to help with getting the dialect into the 6 branch. @dreab8: let me know what you need from me.

Sanne:
> I think it would be worth backporting it to 5.4 to have an alternative for people not willing or able to use JPA 3.0. Do you want me to create another PR for the backport, or will you do it?

+1
No worries, it's trivial enough. I'll run the tests once and push it already.

breglerj:
@Sanne: thanks.

BTW, I've created a new PR (#3351) to update the documentation.

Sanne:
thanks for the doc changes. Merged the dialect in 5.4 too, I'll do the same with the new PR later.

dreab8:
Thanks @breglerj, I have already ported the changes to 6 branch, it was not too complex :)

breglerj:
@dreab8: great, thanks.

dreab8:
@breglerj thanks to you for the PR :+1:

PR #3296, opened 2020-03-11
HHH-13895 Upgrade the PostgreSQL JDBC driver used for testing to v. 4…

Sanne:
…2.2.11

It's trivial, but prefer CI testing it first.
 - https://hibernate.atlassian.net/browse/HHH-13895

PR #3324, opened 2020-04-01
HHH-13895 Upgrade the PostgreSQL JDBC driver used for testing to v. 42.2.11

dreab8:
https://hibernate.atlassian.net/browse/HHH-13895

It seems the new PostgreSql driver has changed the behaviour if a commit is attempted after an error occurred, in such a case the `org.postgresql.util.PSQLException: The database returned ROLLBACK` exception is thrown.

This behaviour causes the failure of tests where we try to check failures, in such a case a call to `transaction.setRollbackOnly();` in the try/catch avoids the  to call a commit in the `TransactionUtil#doInJPA()` method.

For tests using the `Session#doWork` where we cause an exception on purpose the solution I found is to add the following try/catch block
```
try {
	conn.setAutoCommit( true );
}
catch (Exception e) {
	conn.rollback();
	conn.setAutoCommit( true );
}
```
in the `DriverManagerConnectionProviderImpl$PooledConnections#add()` methos

### DIFF
--- a/databases/pgsql/matrix.gradle
+++ b/databases/pgsql/matrix.gradle
@@ -4,4 +4,4 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
-jdbcDependency 'org.postgresql:postgresql:42.2.2'
+jdbcDependency 'org.postgresql:postgresql:42.2.11'
--- a/databases/pgsql/resources/hibernate.properties
+++ b/databases/pgsql/resources/hibernate.properties
@@ -5,7 +5,7 @@
 # See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
 #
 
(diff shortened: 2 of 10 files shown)


Sanne:
Right, that's expected as the PostgreSQL team just fixed this bug - which some considered critical, so that's why I updated.

I was expecting to see such errors reported on the CI build for #3296 .. remind me please, so we don't run anything other than H2 on pull request checks?

Sanne:
Did you check if your changes to tests have any impact on the other databases?

dreab8:
at the moment I think (but not 100% sure) for PR we are just running tests against h2

dreab8:
running test against Oracle and after I'll try MySQL

dreab8:
the tests passed with both Oracle and MySQL but what I'm not 100% sure about the fix i proposed is the change I made to the `DriverManagerConnectionProviderImpl$PooledConnections#add()` method, not sure this is the right approach.

Sanne:
Hi @dreab8 , interestingly the pgsql reverted that breaking fix, there's a new update of the driver available now.

Sanne:
Replaced by #3576 .. the fix should no longer be neccessary.

PR #3326, opened 2020-04-01
HHH-13920 Make JarFileBasedArchiveDescriptor ignore module-info.class…

PavelKastornyy:
When `persistence.xml` is located in JPMS module then Hibernate throws NullPointerException when it scans jar archive and finds `module-info.class` as I noted in the [issue](https://hibernate.atlassian.net/browse/HHH-13920). 

So, in this PR I check if the entry is a `module-info.class` and if it is this entry is ignored.


NathanQingyangXu:
There seems to be a related ticket for JANDEX: https://issues.redhat.com/browse/JANDEX-41, btw.

PR #3327, opened 2020-04-02
wip/6.0 Implement @Filter for loader

NathanQingyangXu:
https://trello.com/c/RQxAqQaT/104-loaderselectbuilder-implement-use-of-loadqueryinfluencer

The only missing task for this card is the implementation of `collection filter` feature for filtering should not be enforced for entity filtering in loader.

The gist of this PR is to avoid parsing filter `condition`. The main difficulty is how to pass filter parameter into `JdbcSelect` via `JdbcParameterBindings`.

4 legacy bugs flushed out during the implementing and testing. See self-comments in the PR for details.

After this card is done, I will work on implementing `@Filter` for HQL or Criteria (card created at https://trello.com/c/KLZjWiX0/118-filter-support-for-hql-criteria). However, this PR will lay solid ground for later endeavour.


1797890817:
In the Hibernate user manual: https://docs.jboss.org/hibernate/orm/6.0/userguide/html_single/Hibernate_User_Guide.html#pc-filter mentioned [Filters apply to entity queries, but not to direct fetching.] , But I very much hope to support it in a future version. I now want to try to develop a multi-tenant model based on data table columns, where I have encountered a bottleneck. https://github.com/vip-efactory/ejpa-example/tree/br_tenant_discriminator, I can only give up temporarily.

NathanQingyangXu:
> In the Hibernate user manual: https://docs.jboss.org/hibernate/orm/6.0/userguide/html_single/Hibernate_User_Guide.html#pc-filter mentioned [Filters apply to entity queries, but not to direct fetching.] , But I very much hope to support it in a future version. I now want to try to develop a multi-tenant model based on data table columns, where I have encountered a bottleneck. https://github.com/vip-efactory/ejpa-example/tree/br_tenant_discriminator, I can only give up temporarily.

Could you work around the limitation by using HQL instead? Guess there is good reason to ignore filter for loader, but HQL works.

1797890817:
> > In the Hibernate user manual: https://docs.jboss.org/hibernate/orm/6.0/userguide/html_single/Hibernate_User_Guide.html#pc-filter mentioned [Filters apply to entity queries, but not to direct fetching.] , But I very much hope to support it in a future version. I now want to try to develop a multi-tenant model based on data table columns, where I have encountered a bottleneck. https://github.com/vip-efactory/ejpa-example/tree/br_tenant_discriminator, I can only give up temporarily.
> 
> Could you work around the limitation by using HQL instead? Guess there is good reason to ignore filter for loader, but HQL works.

Okay, thank you!

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/SqlAstTreeHelper.java
```diff
@@ -49,6 +49,7 @@ public static Predicate combinePredicates(Predicate baseRestriction, Predicate i
 		}
 		else {
 			combinedPredicate = new Junction( Junction.Nature.CONJUNCTION );
+			combinedPredicate.add( baseRestriction );
```
NathanQingyangXu, line 52: Third bug flushed out

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -4568,8 +4576,29 @@ public boolean isAffectedByEnabledFetchProfiles(LoadQueryInfluencers loadQueryIn
...
+							if ( collectionPersister.getFetchMode() == FetchMode.JOIN
+									&& collectionPersister.isAffectedByEnabledFilters( loadQueryInfluencers ) ) {
+								return true;
+							}
+						}
+					}
+				}
+			}
+		}
+		return false;
```
NathanQingyangXu, line 4601: The second bug flushed out. The original implementation doesn't take into consideration the following edge case:
1. enabled filter is not for the owner entity;
2. there is a collection field to be loaded eagerly by `join`;
3. enabled filter is for this collection's element entity;

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -478,7 +524,8 @@ else if ( fetchDepth > maximumFetchDepth ) {
 				);
 				fetches.add( fetch );
 
-				if ( fetchable instanceof PluralAttributeMapping && fetchTiming == FetchTiming.IMMEDIATE ) {
+				if ( fetchTiming == FetchTiming.IMMEDIATE && joined && fetchable instanceof PluralAttributeMapping ) {
+					applyFiltering( querySpec, loadQueryInfluencers, (PluralAttributeMapping) fetchable );
```
NathanQingyangXu, line 528: The first bug flushed out. We only wanna apply `filtering` and `ordering` for `join` type, not for `select` type.

NathanQingyangXu:
Just realized I overlooked `ManyToMany` filtering scenario. Will open again when it is implemented.

hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
```diff
@@ -990,7 +990,7 @@ private void bindFilters(boolean hasAssociationTable) {
 		if ( filterJoinTables != null ) {
 			for (FilterJoinTable filter : filterJoinTables.value()) {
 				if ( hasAssociationTable ) {
-					collection.addFilter(filter.name(), filter.condition(),
+					collection.addManyToManyFilter(filter.name(), filter.condition(),
 							filter.deduceAliasInjectionPoints(),
 							toAliasTableMap(filter.aliases()), toAliasEntityMap(filter.aliases()));
 				}
```
NathanQingyangXu, line 996: Not 100% sure about the above code change. Seems a catch, though.

dreab8:
applied upstream, Thanks @NathanQingyangXu

PR #3328, opened 2020-04-02
HHH-13925 Upgrade to Gradle 6.3

yrodiere:
https://hibernate.atlassian.net/browse/HHH-13925

~This is a preview: I ran the build locally to some extent, but I'd like a build from the CI. Also, I still need to run other builds, especially on JDK13 and 14, and to test publishing.~ => Done (see comments)

PR #3329, opened 2020-04-03
HHH-13682 Generate Java 11/13/14 bytecode for tests when building with JDK11/13/14

yrodiere:
https://hibernate.atlassian.net/browse/HHH-13682

~This PR is based on #3328 (HHH-13925 - upgrade to Gradle 6.3) . #3328 should be merged first. This PR includes commits from #3328, so it should be rebased after #3328 has been merged.~ => Done

The goal here is simply to compile tests to Java 11/13/14 when we build with JDK 11/13/14. That way, not only do we check that Hibernate ORM can be run within JDK 11/13/14, but we also check that it will correctly interact with Java 11/13/14 bytecode, which is especially important for bytecode enhancement.

Hopefully this will allow us to avoid the kind of problems we've had before, where bytebuddy needed an update and we didn't notice it because we were only ever manipulating Java 8 bytecode in our tests.

Build against JDK13: https://ci.hibernate.org/view/Personal%20jobs/job/hibernate-orm-personal-yoann-jdk13/8/
Build against JDK14: https://ci.hibernate.org/view/Personal%20jobs/job/hibernate-orm-personal-yoann-jdk14/4/

How to test that Java 11/13/14 is actually generated for tests, and not for the main code:

```
# Build with JDK13
$ ...
# Check main code: bytecode version 52, i.e. Java 8
$ od -A d -t 'u2' -N 8 --endian big hibernate-core/target/classes/java/main/org/hibernate/Criteria.class
0000000 51966 47806     0    52
0000008
# Check test code: bytecode version 57, i.e. Java 13
$ od -A d -t 'u2' -N 8 --endian big hibernate-core/target/classes/java/test/org/hibernate/jpa/test/callbacks/PreUpdateBytecodeEnhancementTest.class
0000000 51966 47806     0    57
0000008
```

### DIFF
--- a/gradle/java-module.gradle
+++ b/gradle/java-module.gradle
@@ -208,18 +208,6 @@ processTestResources {
 	}
 }
 
-// Enable the experimental features of ByteBuddy with JDK 12+
-test {
-	//Only safe to attempt to parse the version as an integer since JDK11
-	if ( JavaVersion.current().isJava11Compatible() ) {
-		int majorJVMVersionInt = Integer.valueOf(JavaVersion.current().toString());
-		//Set the -Dnet.bytebuddy.experimental=true property only when we need it:
-		if (majorJVMVersionInt >= 12) {
-			systemProperty 'net.bytebuddy.experimental', true
-		}
-	}
-}
-
 test {
 	if ( project.findProperty( 'log-test-progress' )?.toString()?.toBoolean() ) {
 		// Log a statement for each test.
--- a/gradle/java-module.gradle
+++ b/gradle/java-module.gradle
@@ -10,7 +10,7 @@ buildscript {
 		mavenCentral()
 	}
 	dependencies {
-		classpath 'de.thetaphi:forbiddenapis:2.6'
+		classpath 'de.thetaphi:forbiddenapis:2.7'
 	}
 }
 
--- a/gradle/java-module.gradle
+++ b/gradle/java-module.gradle
@@ -38,10 +38,6 @@ ext {
 	forbiddenAPITargetJDKCompatibility = '11'
 }
 
-
-sourceCompatibility = project.baselineJavaVersion
-targetCompatibility = project.baselineJavaVersion
-
 if ( !project.description ) {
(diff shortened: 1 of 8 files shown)


gradle/java-module.gradle
```diff
@@ -200,18 +208,6 @@ processTestResources {
...
 }
 
-// Enable the experimental features of ByteBuddy with JDK 12+
-test {
-	//Only safe to attempt to parse the version as an integer since JDK11
-	if ( JavaVersion.current().isJava11Compatible() ) {
-		int majorJVMVersionInt = Integer.valueOf(JavaVersion.current().toString());
-		//Set the -Dnet.bytebuddy.experimental=true property only when we need it:
-		if (majorJVMVersionInt >= 12) {
-			systemProperty 'net.bytebuddy.experimental', true
```
Sanne, line 210: We no longer need this? Perhaps the check should be bumped to `>= 15` ? It's kinda a pattern for ByteBuddy to only support the "work in progress" JDKs when this flag is on. I didn't check if the strategy changed.

yrodiere:
> are we still able to check that Hibernate ORM - as built with JDK8 - runs fine in JDK11 ?

We never had that ability in the first place. We were always building with JDK11 when testing in JDK11. It's just that we were setting the target to Java 8, so we were generating Java 8 bytecode... through JDK11.

Also, this patch does not change the bytecode generated for ORM in any way. It's just about the tests.

To sum up:

* Before, we were checking that ORM built in JDK11 with a Java 8 target (Java 8 bytecode) runs fine in JDK11 with tests built in JDK11 with  a Java 8 target (Java 8 bytecode)
* Now, we are checking that ORM built in JDK11 with a Java 8 target (Java 8 bytecode) runs fine in JDK11 with tests built in JDK11 with  a Java 11 target (Java 11 bytecode)
* Optionally, by setting project properties on the command line, we can run a build checking that ORM built in JDK11 with a Java 8 target (Java 8 bytecode) runs fine in JDK11 with tests built in JDK11 with  a Java 8 target (Java 8 bytecode)

PR #3331, opened 2020-04-05
HHH-13928 Allow listening to query execution and execution times

marschall:
https://hibernate.atlassian.net/browse/HHH-13928

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/SessionBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/SessionBuilder.java
@@ -10,6 +10,7 @@ import java.sql.Connection;
 import java.util.TimeZone;
 
 import org.hibernate.resource.jdbc.spi.PhysicalConnectionHandlingMode;
+import org.hibernate.resource.jdbc.spi.StatementExecutionListener;
 import org.hibernate.resource.jdbc.spi.StatementInspector;
 
 /**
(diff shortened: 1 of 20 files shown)


marschall:
A couple of notes about the implementation:

* It is modeled after `StatementInspector` as advised by @sebersole https://hibernate.zulipchat.com/#narrow/stream/132096-hibernate-user/topic/custom.20SqlStatementLogger
* The serialized form of `JdbcCoordinatorImpl` changes.
* There currently is a limitation that the listener is called after `java.sql.PreparedStatement#executeQuery()`. Total query execution can be longer when result set processing incurs additional network round trips for example loading additional data using `java.sql.ResultSet#next()`, `java.sql.ResultSet#getBlob(String)` or similar. `SqlStatementLogger` has the same limitations.
* This benefits from #3292.
* This is an alternative to #3287 [HHH-13888](https://hibernate.atlassian.net/browse/HHH-13888)

ibre5041:
When method parameters are 
```
void statementExecuted(String sql, long startTimeNanos);
```
then it would not be possible get diagnostic infromation from JDBC Statement implementation.
Also it is not guarateed that calling toString on JDBC Statement will return actual SQL statement.

marschall:
> then it would not be possible get diagnostic infromation from JDBC Statement implementation.

True.

> Also it is not guarateed that calling toString on JDBC Statement will return actual SQL statement.

True, but it's what is currently used by `SqlStatementLogger` so it's consistent with this.

My hope is that #3292 gets merged and then I can update this PR so can be updated accordingly. If we want the `Statement` as well we could do any the following:

* Add the `Statement` as a second parameter
* Add a second method
* Add a second interface

I'm happy to go with whatever the Hibernate team prefers.

marschall:
Fixed in 6.4 in a different MR

PR #3332, opened 2020-04-06
HHH-13890 : Add support for custom EventType and listeners

gbadner:
https://hibernate.atlassian.net/browse/HHH-13890

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
@@ -97,7 +97,7 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 
 	private final SessionFactoryImplementor sessionFactory;
 	private final CallbackRegistryImplementor callbackRegistry;
-	private final EventListenerGroupImpl[] registeredEventListeners;
+	private EventListenerGroupImpl[] registeredEventListeners;
 	private CallbackBuilder callbackBuilder;
 
 	/**
@@ -160,6 +160,43 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 		}
 	}
 
+	@SuppressWarnings({ "unchecked", "rawtypes" })
+	private synchronized <T> EventListenerGroupImpl<T> getOrCreateEventListenerGroup(EventType<T> eventType) {
+		final int sizeOriginal = this.registeredEventListeners.length;
+		if ( eventType.ordinal() < sizeOriginal ) {
+			final EventListenerGroupImpl registeredEventListener = registeredEventListeners[ eventType.ordinal() ];
+			if ( registeredEventListener != null ) {
+				// eventType has already been registered;
+				return registeredEventListener; // EARLY RETURN
+			}
+			// eventType has not been registered yet.
+			// Its EventListenerGroupImpl will be created and added to registeredEventListeners below.
+		}
+		else {
+			// eventType is a custom EventType, and there is not enough space in
+			// registeredEventListeners to accommodate it.
+
+			// Allocate a new array to hold listener groups for *all* EventType values that currently exist.
+			// This way an existing, unregistered EventType with a larger ordinal will not require another
+			// allocation when it gets registered in the future.
+			final int sizeNew = EventType.values().size();
+			final EventListenerGroupImpl[] registeredEventListenersNew = new EventListenerGroupImpl[sizeNew];
+
+			// First copy the existing listeners to registeredEventListenersNew.
+			System.arraycopy( this.registeredEventListeners, 0, registeredEventListenersNew, 0, sizeOriginal );
+
+			// Now update the reference.
+			this.registeredEventListeners = registeredEventListenersNew;
+		}
+
+		final EventListenerGroupImpl listenerGroup = new EventListenerGroupImpl(
+				eventType,
+				EventListenerRegistryImpl.this
+		);
+		registeredEventListeners[eventType.ordinal()] = listenerGroup;
+		return listenerGroup;
+	}
+
 	@SuppressWarnings({ "unchecked" })
 	public <T> EventListenerGroupImpl<T> getEventListenerGroup(EventType<T> eventType) {
 		EventListenerGroupImpl<T> listeners = registeredEventListeners[ eventType.ordinal() ];
@@ -218,7 +255,7 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 	@Override
 	@SafeVarargs
 	public final <T> void setListeners(EventType<T> type, T... listeners) {
-		EventListenerGroupImpl<T> registeredListeners = getEventListenerGroup( type );
+		EventListenerGroupImpl<T> registeredListeners = getOrCreateEventListenerGroup( type );
 		registeredListeners.clear();
 		if ( listeners != null ) {
 			for ( T listener : listeners ) {
@@ -236,7 +273,7 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 	@Override
 	@SafeVarargs
 	public final <T> void appendListeners(EventType<T> type, T... listeners) {
-		getEventListenerGroup( type ).appendListeners( listeners );
+		getOrCreateEventListenerGroup( type ).appendListeners( listeners );
 	}
 
 	@Override
@@ -248,7 +285,7 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 	@Override
 	@SafeVarargs
 	public final <T> void prependListeners(EventType<T> type, T... listeners) {
-		getEventListenerGroup( type ).prependListeners( listeners );
+		getOrCreateEventListenerGroup( type ).prependListeners( listeners );
 	}
 
 	private EventListenerGroupImpl[] buildListenerGroups() {
--- a/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
@@ -163,6 +163,7 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 	@SuppressWarnings({ "unchecked", "rawtypes" })
 	private synchronized <T> EventListenerGroupImpl<T> getOrCreateEventListenerGroup(EventType<T> eventType) {
 		final int sizeOriginal = this.registeredEventListeners.length;
+		final EventListenerGroupImpl[] registeredEventListenersNew;
 		if ( eventType.ordinal() < sizeOriginal ) {
 			final EventListenerGroupImpl registeredEventListener = registeredEventListeners[ eventType.ordinal() ];
 			if ( registeredEventListener != null ) {
@@ -171,6 +172,8 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 			}
 			// eventType has not been registered yet.
 			// Its EventListenerGroupImpl will be created and added to registeredEventListeners below.
+			// There is already space for the new EventType in this.registeredEventListeners.
+			registeredEventListenersNew = this.registeredEventListeners;
 		}
 		else {
 			// eventType is a custom EventType, and there is not enough space in
@@ -180,20 +183,21 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 			// This way an existing, unregistered EventType with a larger ordinal will not require another
 			// allocation when it gets registered in the future.
 			final int sizeNew = Math.max( eventType.ordinal() + 1, EventType.values().size() );
-			final EventListenerGroupImpl[] registeredEventListenersNew = new EventListenerGroupImpl[sizeNew];
+			registeredEventListenersNew = new EventListenerGroupImpl[sizeNew];
 
 			// First copy the existing listeners to registeredEventListenersNew.
 			System.arraycopy( this.registeredEventListeners, 0, registeredEventListenersNew, 0, sizeOriginal );
-
-			// Now update the reference.
-			this.registeredEventListeners = registeredEventListenersNew;
 		}
 
 		final EventListenerGroupImpl listenerGroup = new EventListenerGroupImpl(
 				eventType,
 				EventListenerRegistryImpl.this
 		);
-		registeredEventListeners[eventType.ordinal()] = listenerGroup;
+		registeredEventListenersNew[eventType.ordinal()] = listenerGroup;
+
+		// Now update the reference.
+		this.registeredEventListeners = registeredEventListenersNew;
+
 		return listenerGroup;
 	}
 
--- a/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
@@ -97,7 +97,7 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 
 	private final SessionFactoryImplementor sessionFactory;
 	private final CallbackRegistryImplementor callbackRegistry;
-	private EventListenerGroupImpl[] registeredEventListeners;
+	private volatile EventListenerGroupImpl[] registeredEventListeners;
 	private CallbackBuilder callbackBuilder;
 
 	/**
@@ -179,7 +179,7 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 			// Allocate a new array to hold listener groups for *all* EventType values that currently exist.
 			// This way an existing, unregistered EventType with a larger ordinal will not require another
 			// allocation when it gets registered in the future.
-			final int sizeNew = EventType.values().size();
+			final int sizeNew = Math.max( eventType.ordinal() + 1, EventType.values().size() );
 			final EventListenerGroupImpl[] registeredEventListenersNew = new EventListenerGroupImpl[sizeNew];
 
 			// First copy the existing listeners to registeredEventListenersNew.
--- a/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
@@ -199,7 +199,13 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 
 	@SuppressWarnings({ "unchecked" })
 	public <T> EventListenerGroupImpl<T> getEventListenerGroup(EventType<T> eventType) {
-		EventListenerGroupImpl<T> listeners = registeredEventListeners[ eventType.ordinal() ];
+		if ( registeredEventListeners.length < eventType.ordinal() + 1 ) {
+			// eventTpe is a custom EventType that has not been registered.
+			// registeredEventListeners array was not allocated enough space to
+			// accommodate it.
+			throw new HibernateException( "Unable to find listeners for type [" + eventType.eventName() + "]" );
+		}
+		final EventListenerGroupImpl<T> listeners = registeredEventListeners[ eventType.ordinal() ];
 		if ( listeners == null ) {
 			throw new HibernateException( "Unable to find listeners for type [" + eventType.eventName() + "]" );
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
```diff
@@ -76,6 +78,41 @@
...
 
+	/**
+	 * Add a new event type.
+	 *
+	 * @param name - name of the custom event
+	 * @param listenerClass - the base listener class or interface associated with the entity type
+	 * @param <T> - listenerClass
+	 * @return the custom {@link EventType}
+	 */
+	public synchronized static <T> EventType<T> addCustomEventType(String name, Class<T> listenerClass) {
```
Sanne, line 89: The `EVENT_TYPE_BY_NAME_MAP` is not safe for concurrent access; I see you added a `synchronized` on this method, which is correct as it performs write operations on the map.

But this implies that we'll need to have the same lock on reads as well; perhaps the map should be changed from `HashMap` to a `ConcurrentHashMap`?

Alternatively, I wonder if the map could be statically initialized by discovering a service. If you define a "CustomEventTypeContributor" interface, then Hibernate RX could implement it and you could load them when this class is initialized.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
```diff
@@ -76,6 +78,41 @@
...
+		if ( name == null || listenerClass == null ) {
+			throw new HibernateException( "Custom EventType name and associated class must be non-null." );
+		}
+		final EventType existingEventType = EVENT_TYPE_BY_NAME_MAP.get( name );
+		if ( existingEventType != null ) {
+			if ( listenerClass.equals( existingEventType.baseListenerInterface ) ) {
+				LOG.warn( "EventType [" + name + "] with listener Class ["
```
Sanne, line 96: If we log this as a warning, how can we prevent Hibernate RX to log such warnings if it's started multiple times?

We would need to either have a way to check if such types have not been registered already, or register them statically - so making sure it happens once and once only.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
```diff
@@ -76,6 +78,41 @@
...
+								+ "]. An EventType with that name already exists with listener ["
+								+ listenerClass.getName()
+								+ "]."
+				);
+			}
+		}
+		else {
+			final EventType<T> eventType = create( name, listenerClass );
+			EVENT_TYPE_BY_NAME_MAP.put( name, eventType );
+			LOG.info( "Added custom EventType:  [" + name + "], ordinal=[" +eventType.ordinal + "], listener=[" + listenerClass + "]." );
```
Sanne, line 112: Is "INFO" level not a bit too noisy? I'd use "DEBUG", as most users won't need to know how this works behind the scenes.

gbadner:
I fixed the checkstyle failure. I needed to rebase before pushing to fix another failure.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
gbadner: I original made this method `synchronized` because I had added a `List` for resolving an `EventType` by `ordinal` for an earlier incarnation of the fix. This was necessary to make sure that both `EVENT_TYPE_BY_NAME_MAP` and the `List` were kept in sync. I ended up not needing the `List`, so I deleted it. 

I thought it was still a good idea to keep this method method `synchronized`. I didn't really see a reason for changing the `HashMap` to `ConcurrentHashMap`. It seemed very unlikely that an application would look up an `EventType` by name before creating the custom `EventType` via `EventType#addCustomEventType`. I suppose it is possible though.

I'll do as you suggest: 1) remove `synchronized`; 2) change `HashMap` to  `ConcurrentHashMap`.

gbadner: Regarding:

> Alternatively, I wonder if the map could be statically initialized by discovering a service. If you define a "CustomEventTypeContributor" interface, then Hibernate RX could implement it and you could load them when this class is initialized.

It could be done that way, but I'm not sure it's worthwhile. hibernate-rx already has an implementation of `Integrator` for registering listeners: [org.hibernate.rx.service.RxIntegrator](https://github.com/hibernate/hibernate-rx/blob/master/hibernate-rx-core/src/main/java/org/hibernate/rx/service/RxIntegrator.java). 

It would be simple for `Integrator#integrate` to add the custom `EventType` before adding its corresponding listeners.

I could go either way on this. 

@sebersole, do you have an opinion on this?

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
gbadner: I didn't think there would be all that many custom `EventType`s added, so I didn't think much about noisiness. I do see your point though about hibernate-rx users.

I'll make this change.

NathanQingyangXu: It seems a good example to guard against unnecessary logging performance cost. The string concatenation cost is always incurred regardless of whether the logging statement is to be run or not. IMHO, we should either
```
if ( LOG.isInfoEnabled() ) {
    LOG.info( "Added custom EventType:  [" + name + "], ordinal=[" +eventType.ordinal + "], listener=[" + listenerClass + "]." );
}
```
or
```
LOG.info( "Added custom EventType:  [%s], ordinal=["%d"], listener=["%s"].",  name, eventType.ordinal, listenerClass );
```
I prefer the latter for its much better readability.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
Sanne: > It seemed very unlikely that an application would look up an EventType by name before creating the custom EventType via EventType#addCustomEventType. I suppose it is possible though.

Keep in mind that the problem is not strictly about the order in which such operations might happen: the problem lies in the visibility rules. Even read operations need to ensure that any state change which has happened (at any point in the past) is actually "made visible" to them.

> I'll do as you suggest: 1) remove synchronized; 2) change HashMap to ConcurrentHashMap.

+1 to use a ConcurrentMap, but please keep the `synchronized` : like @NathanQingyangXu mentions, otherwise there's a possible race condition.

Thanks!

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
Sanne: > I didn't think there would be all that many custom EventTypes added, so I didn't think much about noisiness. 

I agree that I don't expect many custom types to be added.

But even assuming that RX would add a single type, if it's started twice you'd get this log message - unless the code in RX can somehow figure out that registering its own type is not necessary.

It's not unlikely to start a framework twice in the same JVM: one might have multiple applications in a server, or multiple persistence units in a single application, or running integration tests which start a new Hibernate instance for each test.  Since it's a static map, we need to take such possibilities into account.

A possible alternative would be not use a static map for this purpose.

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
```diff
@@ -160,6 +160,43 @@ public void prepare(MetadataImplementor metadata) {
...
+			this.registeredEventListeners = registeredEventListenersNew;
+		}
+
+		final EventListenerGroupImpl listenerGroup = new EventListenerGroupImpl(
+				eventType,
+				EventListenerRegistryImpl.this
+		);
+		registeredEventListeners[eventType.ordinal()] = listenerGroup;
+		return listenerGroup;
+	}
```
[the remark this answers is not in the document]
Sanne: I'm glad you take performance into account, but this whole method is only used during bootstrap if I understood correctly?
If that's correct, synchronization would not worry me - as long as it's limited to this specific point.

But @NathanQingyangXu is right that `registeredEventListeners` would technically need the `volatile` keyword; would be great to keep the `final` if it was possible to rethink the approach.

gbadner:
@Sanne, @NathanQingyangXu, I've pushed a commit with most of the suggested changes.

I have not added `EventTypeContributor` because I'm still not sure that this is necessary. I'd like to hear from @sebersole before adding it.

Also, since `EventType#addCustomEventType` is public, so it could still be called directly, so I think we would still need to deal with the same issues.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
gbadner: @NathanQingyangXu, I believe I have dealt with your first point (`f (!map.containsKey(key)) {...}` in my most recent commit.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
gbadner: @Sanne, this check was removed in my most recent commit.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
gbadner: I missed this in my last commit. I'll get it in the next round.

FWIW, I don't think this is going to have much of an impact on performance. I don't imagine there would be a lot of custom `EventType`s that would be added.

Also, I believe adding custom `EventType`s and listener registration should only happen when the `SessionFactory` is being built. This should have no effect on an application's operation.

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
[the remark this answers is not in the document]
gbadner: This code should only execute while `SessionFactory` is being built. It should have no effect on an application's operation. I don't really see the point of making this more complicated.

@Sanne, what do you think?

I did add `volatile` for `registeredEventListeners`.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
gbadner: @Sanne, I changed it to {{LOG.debug}} in my most recent commit. I'll change the message as @NathanQingyangXu suggests.

gbadner:
@Sanne, @NathanQingyangXu, I believe I have covered everything that has been mentioned, except for the addition of a `EventTypeContributor`.

As I mentioned earlier, there is not much point in adding `EventTypeContributor` if `EventType#addCustomEventType` is `public`. I suppose it might be possible to change it to be package-protected, but I think it's more trouble than it's worth.

IMHO, adding `EventTypeContributor` is unnecessary. 

A custom `EventType` must be registered using `EventListenerRegistry#setListeners`, `EventListenerRegistry#appendListeners`, or `EventListenerRegistry#prependListeners`; otherwise, `HibernateException` will be thrown. 

I don't see the point of separating add the addition of custom `EventType`s from code that registers listeners. It will just add an extra thing for an application to implement.

@sebersole approved the changes without `EventTypeContributor`, and he has not chimed in here.

@Sanne, @NathanQingyangXu, if everything looks OK with you, and you are OK not having `EventTypeContributor`, then please approve this PR.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
```diff
@@ -76,6 +79,46 @@
...
+
+		final EventType eventType = EVENT_TYPE_BY_NAME_MAP.computeIfAbsent(
+				name,
+				( e -> {
+					final EventType eventTypeNew = EventType.create( name, listenerClass );
+					LOG.debug( String.format(
+							"Added custom EventType: [%s], ordinal=[%d], listener=[%s].",
+							name,
+							eventTypeNew.ordinal,
+							listenerClass.toString()
```
NathanQingyangXu, line 104: We'd replace the above with
```
LOG.debugf( 
							"Added custom EventType: [%s], ordinal=[%d], listener=[%s].",
							name,
							eventTypeNew.ordinal,
							listenerClass.toString()
);
```
`String.format()` is terribly slow and it would incur unnecessary runtime performance overhead regardless of whether the debug level is enabled or not.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
```diff
@@ -76,6 +79,46 @@
...
+	 * @param <T> - listenerClass
+	 * @return the custom {@link EventType}
+	 */
+	@SuppressWarnings({ "unchecked", "rawtypes" })
+	public static synchronized <T> EventType<T> addCustomEventType(String name, Class<T> listenerClass) {
+		if ( name == null || listenerClass == null ) {
+			throw new HibernateException( "Custom EventType name and associated class must be non-null." );
+		}
+
+		final EventType eventType = EVENT_TYPE_BY_NAME_MAP.computeIfAbsent(
```
NathanQingyangXu, line 96: Have we changed the implementation of `EVENT_TYPE_BY_NAME_MAP` to `ConcurrentHashMap`? If so it seems we don't need `synchronized` keyword for this method any more.

gbadner:
@NathanQingyangXu, thank you for approving this. 

@Sanne and I agree that this fix should wait until after 5.4 is branched. We will revisit this after that is done.

NathanQingyangXu:
> @NathanQingyangXu, thank you for approving this.
> 
> @Sanne and I agree that this fix should wait until after 5.4 is branched. We will revisit this after that is done.

Thanks for letting me know. Anyway, I have approved it.

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
```diff
@@ -160,9 +160,52 @@ public void prepare(MetadataImplementor metadata) {
...
+			// Its EventListenerGroupImpl will be created and added to registeredEventListeners below.
+		}
+		else {
+			// eventType is a custom EventType, and there is not enough space in
+			// registeredEventListeners to accommodate it.
+
+			// Allocate a new array to hold listener groups for *all* EventType values that currently exist.
+			// This way an existing, unregistered EventType with a larger ordinal will not require another
+			// allocation when it gets registered in the future.
+			final int sizeNew = Math.max( eventType.ordinal() + 1, EventType.values().size() );
```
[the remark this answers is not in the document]
gbadner: @Sanne, @NathanQingyangXu, before I address your questions, I want to make clear that making `EventType#addCustomEventType` `synchronized` and the line Sanne mentioned are related

When I removed `synchronized` from `EventType#addCustomEventType`, and ran the performance test, I found that there were cases where `eventType.ordinal() + 1` was greater than `EventType.values().size()`. As a result there were multiple `IndexOutOfBoundsException` thrown. 

That is why I changed:

`final int sizeNew = EventType.values().size();`

to: 

`final int sizeNew = Math.max( eventType.ordinal() + 1, EventType.values().size() );`

When I added `synchronized` back to `EventType#addCustomEventType`, that was no longer the case. I could have reverted back to what I had originally, but decided to keep it in case someone decided to removed `synchronized` from `EventType#addCustomEventType`.

Should I remove `synchronized` from `EventType#addCustomEventType`?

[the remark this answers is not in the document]
gbadner: @Sanne, I don't understand why your example would fail.

First off, each `EventType` should have at least one listener registered; otherwise, there's no reason to have an `EventType`. If no listener is registered for a particular `EventType`, then `EventListenerRegistryImpl#getEventListenerGroup` will throw `HibernateException`.

I opted to allocate space for all existing `EventType`s in `#registeredEventListeners` to reduce the number of times it needs to be re-allocated.

Obviously,  the number of allocations will depend on how the application creates `EventType`s and registers its listeners. If it creates all the custom `EventType`s first, then registers listeners, then only one re-allocation will be necessary.  If the application creates a custom `EventType`, then registers its listeners before creating the next custom `EventType`, then the array will need to be re-allocated for each custom `EventType`.

As I mentioned above, if `EventType#addCustomEventType` is not `synchronized`, then `eventType.ordinal() + 1` may be greater than `EventType.values().size()`. If it is synchronized, `eventType.ordinal() + 1` will always be <= `EventType.values().size()`.

AFAICT, the case that you mention should never happen. 

For 1), the array is allocated to accommodate an ordinal of 100.
For 2),  `sizeOriginal` is already 100, so the `if` condition is satisfied. `registeredEventListeners[ eventType.ordinal() ] == null`, so execution falls through to after the `else` clause (i.e., no re-allocation is done).
 
I'm not sure what you mean by "There's no need to literally "loop" among all of them, ...".

Could you add a failing test case to clarify?

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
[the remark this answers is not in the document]
gbadner: @Sanne, in the most recent commit, the reference gets updated after a value is added to the array.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
gbadner: @NathanQingyangXu, fixed in the most recent comment.

hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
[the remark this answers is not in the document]
Sanne: ok thanks Gail, I might have missed something. I'll merge it as is, if there's a problem we can follow up.

Sanne:
thanks @gbadner , I'll squash the commits into one for clarity and merge it.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
sebersole: > @sebersole, do you have an opinion on this?

@gbadner I prefer dedicated integration contracts (`TypeContributor`, e.g.) rather than generic ones (`Integrator`).  Here, I'd actually even consider defining a contract for listener registration as well rather than relying on the generic `Integrator`.  Something like:

```
interface EventSystemContributions {
    <T> void registerEventType(EventType<T> type);

    <T> void prependListener(EventType<T> type, T listener);
    <T> void appendListener(EventType<T> type, T listener);

    void registerDuplicationStrategy(DuplicationStrategy strategy);
    void registerDuplicationStrategy(EventType type, DuplicationStrategy strategy);
}

interface EventSystemContributor {
    void contribute(EventSystemContributions target);
}
```

gbadner:
@Sanne, @NathanQingyangXu, I see that @sebersole has suggested a change:
https://github.com/hibernate/hibernate-orm/pull/3332#discussion_r409576346

If I can get it fixed before it gets released, I'll use the same jira.

hibernate-core/src/main/java/org/hibernate/event/spi/EventType.java
sebersole: I mentioned to Gail that I started a PR of my own on top of hers to look at my suggestion.  I wonder what y'all think about some related clean-up...

The main problem I had is with keeping track of custom event types on EventType itself.  Ugly chicken/egg situation that requires way too much static state.  At a super high-level I'd suggest instead to keep `EventType` simple - just the class definition and the static standard type declarations.  In fact, in retrospect, knowing that customer event-types were in the future I would have split even those out into a different class `StandardEventTypes` e.g.  But specifically here, I want to get rid of anything related to custom types.

Any customization is logically part of the registry, not the EventType class.  As such, imo, it belongs as part of the registry.  

All together this has the benefit of removing the worries about static wrt entries "bleeding" across VMs.

Anyway, I'm going to take that approach in my PR.  The "downside" is that certain methods like `EventType#values` no longer has the same meaning.  Well, it does... it has the same meaning it had before any of these "custom type" changes - it would simply be the list of standard types.

Sanne:
@sebersole I agree with your suggestions, but bear in mind that I had already merged @gbadner 's PR in master (exclusively). I had forgotten to close this PR.

So +1 for such improvements, I just hope that having this already merged isn't too inconvenient. I'll finally close this PR now.

sebersole:
As I mentioned on Zulip, I need some guidance on Rx's expected usage
pattern.  Not registering, but how will it use these custom even types at
runtime?

Sanne:
> As I mentioned on Zulip, I need some guidance on Rx's expected usage
pattern.  Not registering, but how will it use these custom even types at
runtime?

@DavideD could you show / explain the plan to Steve?

PR #3335, opened 2020-04-07
HHH-13936: attempt to join transaction in doFlush

elmuerte:
Call pulseTransactionCoordinator before checking for transaction.


elmuerte:
In the added test case `testIsConnectedFlush` works in the current release, `testFlush` does not.
With this fixed, both cases are successful.

dreab8:
Hi @elmuerte , 
thanks for the PR it looks fine to me.

@Sanne if you think it is ok I'll merge it

Sanne:
Many thanks @elmuerte and @dreab8 ! It's merged now in both master and `5.4`.

PR #3336, opened 2020-04-08
Wip/6.0 Embeddable ForeignKeys

dreab8:
This PR contains the implementation of associations involving Embedded FK, see commits:

- https://github.com/hibernate/hibernate-orm/pull/3336/commits/9d51e9b6c0f1d87a3bb78db350518a03ccafc254

- https://github.com/hibernate/hibernate-orm/pull/3336/commits/034e5b7bfeb56d62db54debde5d76b8a445d5c26

and the removal of the direction attribute from `ForeignKeyDescriptor`

- https://github.com/hibernate/hibernate-orm/pull/3336/commits/cc4df49f63c7faf5a9fe3867349e62d7616b4a33

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/util/MutableInteger.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/util/MutableInteger.java
@@ -1,0 +1,35 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.internal.util;
+
+public class MutableInteger {
+	private int value;
+
+	public MutableInteger(int value) {
+		this.value = value;
+	}
+
+	public MutableInteger deepCopy() {
+		return new MutableInteger( value );
+	}
+
+	public int getAndIncrement() {
+		return value++;
+	}
+
+	public int get() {
+		return value;
+	}
+
+	public void set(int value) {
+		this.value = value;
+	}
+
+	public void increase() {
+		++value;
+	}
+}
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EmbeddedForeignKeyDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EmbeddedForeignKeyDescriptor.java
@@ -1,0 +1,571 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.metamodel.mapping.internal;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.function.Consumer;
+
+import org.hibernate.HibernateException;
+import org.hibernate.LockMode;
+import org.hibernate.engine.FetchStrategy;
+import org.hibernate.engine.FetchTiming;
+import org.hibernate.internal.util.collections.CollectionHelper;
+import org.hibernate.metamodel.mapping.ColumnConsumer;
+import org.hibernate.metamodel.mapping.EmbeddableMappingType;
+import org.hibernate.metamodel.mapping.EmbeddableValuedModelPart;
+import org.hibernate.metamodel.mapping.EntityMappingType;
+import org.hibernate.metamodel.mapping.ForeignKeyDescriptor;
+import org.hibernate.metamodel.mapping.JdbcMapping;
+import org.hibernate.metamodel.mapping.MappingType;
+import org.hibernate.metamodel.mapping.ModelPart;
+import org.hibernate.metamodel.mapping.SingularAttributeMapping;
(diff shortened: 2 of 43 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EmbeddedForeignKeyDescriptor.java
```diff
@@ -0,0 +1,370 @@
...
+										,
+										null,
+										typeConfiguration
+								);
+					}
+					else {
+						attribute.visitJdbcTypes(
+								jdbcMapping -> {
+									jdbcMappings.add( jdbcMapping );
+								},
```
NathanQingyangXu, line 87: seems a  good candidate of method reference usage above
```
attribute.visitJdbcTypes( jdbcMappings::add, ... );
```

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
```diff
@@ -814,6 +859,61 @@ public static void interpretKeyDescriptor(
...
+		final List<String> keyColumnExpressions = new ArrayList<>();
+		bootValueMapping.getColumnIterator().forEachRemaining(
+				column ->
+						keyColumnExpressions.add( column.getText( dialect ) ) );
+
+		final List<String> targetColumnExpressions = new ArrayList<>();
+		fkTarget.getMappedColumnExpressions().forEach(
+				column ->
+						targetColumnExpressions.add( column ) );
+
```
NathanQingyangXu, line 877: seems we can specify ArrayList size in constructor above to avoid performance and memory overhead, though not a big deal.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/embeddable/internal/EmbeddableForeignKeyResultImpl.java
```diff
@@ -0,0 +1,140 @@
...
+
+	public EmbeddableForeignKeyResultImpl(
+			List<SqlSelection> sqlSelections,
+			NavigablePath navigablePath,
+			EmbeddableValuedModelPart embeddableValuedModelPart,
+			String resultVariable,
+			DomainResultCreationState creationState) {
+		super( embeddableValuedModelPart.getEmbeddableTypeDescriptor(), navigablePath );
+		this.resultVariable = resultVariable;
+		AtomicInteger atomicInteger = new AtomicInteger( 0 );
```
NathanQingyangXu, line 51: We might well save some CPU cycles by using default constructor instead. Super minor.

NathanQingyangXu: Wondering whether we need `AtomicInteger` here for it seems we don't need to worry about concurrency issue here. Maybe the sole reason is to have a integer wrapper which is mutable? Using `AtomicInteger` might incur unnecessary performance cost. I saw there is a much light-weight alternative in `org.hibernate.envers.internal.tools.MutableInteger`.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
```diff
@@ -715,23 +744,36 @@ private static ForeignKeyDescriptor interpretKeyDescriptor(
...
+					);
 
-		throw new NotYetImplementedFor6Exception(
-				"Support for composite foreign-keys not yet implemented: " + bootValueMapping.getRole()
-		);
+			attributeMapping.setForeignKeyDescriptor( embeddedForeignKeyDescriptor );
+		}
+		else {
+			throw new NotYetImplementedFor6Exception(
+					"Support for composite foreign-keys not yet implemented: " + bootValueMapping.getRole()
```
NathanQingyangXu, line 771: I saw you used `foreign keys` elsewhere in this PR. Seems `foreign keys` is good enough without `-`. Again, super minor.

dreab8:
Thanks @NathanQingyangXu , I applied your suggestions.

hibernate-testing/src/main/java/org/hibernate/testing/jdbc/SQLStatementInspector.java
```diff
@@ -0,0 +1,54 @@
...
+		assertTrue( sqlQueries.contains( expected ) );
+	}
+
+	public void assertExecutedCount(int expected) {
+		assertEquals( expected, sqlQueries.size() );
+	}
+
+	public void assertNumberOfOccurrenceInQuery(int queryNumber, String toCheck, int expectedNumberOfOccurrences) {
+		String query = sqlQueries.get( queryNumber );
+		int actual = query.split( toCheck, -1 ).length - 1;
```
NathanQingyangXu, line 51: Great improvement! There is one caveat though. So suppose I wanna check 'join' and the SQL contains `select * from joinable`, the above method won't differentiate between 'substring' matching or 'whole world' matching. it would be great to make it fool-proof for this generic method and it would be incredibly useful.

hibernate-core/src/main/java/org/hibernate/internal/util/MutableInteger.java
```diff
@@ -0,0 +1,35 @@
...
+		return value;
+	}
+
+	public void set(int value) {
+		this.value = value;
+	}
+
+	public void increase() {
+		++value;
+	}
```
NathanQingyangXu, line 34: Thanks. This util class would be incredibly useful.

sebersole:
I applied upstream manually to fix a minor thing

PR #3338, opened 2020-04-09
Bug Fix - Escape Strings for entity name after JOIN keyword.

arulsmv:
Thie commit intended for all the organizations having domains ending in "in".
            eg musicmaster.in
    Since the entity packages names will be starting with  "in.*"  any of JOIN query fails as "in"
    is considered as a keyword.
    This fix intended to escape the package names with QUOTE having fully qualified
    'class' name of the entity.


arulsmv:
> seems a good catch. it would be great to add a testing case to showcase your code change fixed the issue. creating a new JIRA ticket would be desirable as well.

Thanks, Nathan Xu for the quick review.

Will add the test cases Since this is my first opensource pull request not sure about the process followed here.
if you can share me quick checklist will follow it 
Created the case https://hibernate.atlassian.net/browse/HHH-13943

NathanQingyangXu:
> https://hibernate.atlassian.net/browse/HHH-13943

Thanks for the testing case. As a matter of fact, I am a community contributor just like you. Seems decent testing cases added. Hopefully some team member can review your PR, but please be patient for they are busy with v6 developing right now.

arulsmv:
> > https://hibernate.atlassian.net/browse/HHH-13943
> 
> Thanks for the testing case. As a matter of fact, I am a community contributor just like you. Seems decent testing cases added. Hopefully some team member can review your PR, but please be patient for they are busy with v6 developing right now.
Thank you, hope it will be pulled soon.

arulsmv:
Can anyone review this, please?
I have another/similar fix for from query with COMMA

nealeu:
@sebersole What gives with such a long delay?

gbadner:
@arulsmv, IIUC, your fix requires the application to put quotes around the package name to get around this bug. Is that correct? If that is correct, then I don't think this is the correct fix. From your test, it looks like the same issue does not exist after {{FROM}} with a package name starts with "in." after {{FROM}}. IMO, the bug should be fixed in Hibernate, without requiring the application to put quotes around the package name.

@arulsmv, please confirm if I am understanding your proposed fix correctly. I have not had a chance to really dig into this, so I could be missing something.

dreab8:
Hi @arulsmv ,

thanks for you PR and sorry for taking so long to reply, I created a [PR](https://github.com/hibernate/hibernate-orm/pull/9383) with test showing the issue has been resolved on recent versions of hibernate, so I'm going to close this.

Thanks again

PR #3340, opened 2020-04-15
Various changes around how basic values are handled in terms of mappi…

sebersole:
…ng consumption for annotations and how value conversions happen

* Split BasicType "resolution" into 2 - one used for reading (mapping model) versus one used from writing (legacy persister model)
* @SqlTypeCode, @SqlType, @SqlTypeRegistration
* @JavaType, @JavaTypeRegistration

Still need support for @MutabilityPlan


hibernate-core/src/main/java/org/hibernate/annotations/SqlTypeCode.java
```diff
@@ -0,0 +1,68 @@
...
+ * or a custom code.
+ *
+ * ````
+ * @Entity
+ * class User {
+ *     ...
+ *     // By default Hibernate maps Java's Integer to JDBC's INTEGER
+ *     // but here we want to use JDBC's TINYINT instead.
+ *     @SqlType( Types.TINYINT )
+ *     int getAge() { ... }
```
dreab8, line 33: I noticed that after the latest gradle upgrade in order to avoid checkstyle errors a space is needed between `@SqlType` and the `(` in the javadoc

hibernate-core/src/main/java/org/hibernate/annotations/SqlTypeCode.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ *     // By default Hibernate maps Java's String to JDBC's VARCHAR
+ *     // but here we want to use JDBC's NVARCHAR instead.
+ *     @SqlType( Types.NVARCHAR )
```
dreab8, line 37: same here

hibernate-core/src/main/java/org/hibernate/annotations/SqlTypeCode.java
dreab8: may be I'm wrong but it seems this doc should belong to `hibernate-core/src/main/java/org/hibernate/annotations/SqlType.java`

sebersole:
Thanks :)

I doubt anyone else gives this a serious look so I'm going to plan on incorporating this along with a few additional changes mainly around the effect of the new annotations on the legacy-resolution.  `@SqlType`, `@SqlTypeCode`, `@JavaType` and `@Mutability` should all adjust the legacy-resolution.

I'm curious @dreab8 - what are your thoughts on how `@MapKeySql...` is implemented by wrapping the matching `@Sql...` variant.

hibernate-core/src/main/java/org/hibernate/annotations/SqlTypeCode.java
sebersole: Well actually, these do belong here.  I just used the wrong annotation in the comment.  Still you are correct that comments are not correct atm.

dreab8:
may be it is a little verbose probably I would go for  `@MapKeySqlTypeCode( Types.TINYINT )`

sebersole:
I should have mentioned before that I followed the existing pattern used in `org.hibernate.annotations.MapKeyType` and a few other places.

I agree that this pattern is more strange and verbose.  But more I was wondering about consistency.

sebersole:
Actually I we use that pattern quite a bit already in our existing annotations.  I think we should keep it as is for the moment for consistency.

Applied upstream manually with some changes

PR #3341, opened 2020-04-16
HHH-13916 : Add a unique Session "token"

gbadner:
https://hibernate.atlassian.net/browse/HHH-13916

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
@@ -119,6 +119,15 @@ public interface SharedSessionContractImplementor
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
```diff
@@ -118,6 +118,15 @@ default SharedSessionContractImplementor getSession() {
...
 
+	/**
+	 * A "token" that is unique to this Session.
+	 *
+	 * @return The token
+	 */
+	default Object getSessionToken() {
+		return this;
+	}
+
```
NathanQingyangXu, line 129: If this is v6 PR, maybe we can drop `default` here for we don't need to maintain backward compatibility and child class can choose to override it to customize easily.

sebersole:
Looks good.

PR #3342, opened 2020-04-16
HHH-13682 Follow-up: restore the "net.bytebuddy.experimental" system property for JDK15+

yrodiere:
Follow-up on #3329 ([HHH-13682](https://hibernate.atlassian.net/browse/HHH-13682))

Couldn't test for now,  because Gradle 6.3 doesn't work with JDK15+ (a problem with Groovy reflection, apparently). But @Sanne suggested this, and I think he's right: we will need to set the system property.

### DIFF
--- a/gradle/java-module.gradle
+++ b/gradle/java-module.gradle
@@ -208,6 +208,15 @@ processTestResources {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Sanne:
thanks!

PR #3348, opened 2020-04-17
HHH-13214 - fix delete re-firing SQL from previous calls

liningone:


Sanne:
Many thanks for the attempt @liningone !

Closing as this is superseded by #3457

PR #3353, opened 2020-04-17
Pr/hhh 13890

sebersole:
Shows conflicts atm.  I based this on the original PR from @gbadner  I'll look at the conflicts later, just wanted to post this for people to look at.  A few notes...

A quick point about concurrency since that was a lot of the discussion regarding the original work.  In this approach, the collections are all built up front so we are not having contention issues.

Also notice that this makes `EventListenerRegistry` to (effectively) be no longer an event.  `EventEngine` follows a pattern we started in 6 as well.

I also wanted to specifically point out a series of changes that I made to make it easier to test - namely the changes to `SessionFactoryBuilder` and adding `SessionFactoryImpl.ExplicitContributorAccess`.  The beauty and the curse of Integrator is that it is open ended - its not specific to any specific thing any specific integrator would need.  More and more I've been moving to specific contracts because they make it easier on the person implementing them (IMO).  Here I added a `EventEngineContributor` contract in this speciailized vein.  The trouble is in registering and applying them in tests.  Generally these "contributors" are discoverable in the normal Java service sense which works great in real-world use.  However, it is tricky for testing.  Its not really possible to apply a Java service so that it only affects a certain test using normal means.  So for the time being I went with this approach.  The alternative I saw was to instead define a way to augment the Java services ultimately returned from the ClassLoaderService.  I'm open to making that change if anyone is dead set against this `ExplicitContributorAccess` approach.

Anyway, I hope I pass the audition :)


sebersole:
@Sanne can you please at least tell me the branch you merged Gail's original work in to?  I need that info to be able to correct the merge conflicts

Sanne:
hi @sebersole , couldn't look at the code here yet but let me share some initial thoughts already:

regarding Gail's work, it was merged in master only. I didn't make any changes to it except squashis all the commits she had, as there was some back-and-forth. I would expect that if you were to cherry-pick your one commit on top of current master it should apply cleanly without any need for conflict resolution?

> A quick point about concurrency since that was a lot of the discussion regarding the original work. In this approach, the collections are all built up front so we are not having contention issues.

That sounds great, and was my preference as well. We shouldn't collect such mutable state in constants, especially when they are potentially shared by multiple persistence units / applications / classloaders.

sebersole:
> regarding Gail's work, it was merged in master only.
> 
> I would expect that if you were to cherry-pick your one commit on top of current master it should apply cleanly without any need for conflict resolution?

Awesome, thanks.

> That sounds great, and was my preference as well. We shouldn't collect such mutable state in constants, especially when they are potentially shared by multiple persistence units / applications / classloaders.

:+1:

dreab8:
this has been merged

PR #3354, opened 2020-04-18
wip/6.0 Implement filter for HQL/Criteria

NathanQingyangXu:
https://trello.com/c/KLZjWiX0/118-filter-support-for-hql-criteria

Also a followup for https://trello.com/c/RQxAqQaT/104-loaderselectbuilder-implement-use-of-loadqueryinfluencer, which was mainly focused on `loader` filter implementation.
 
The idiosyncrasy of `@Filter` and `@FilterJointable` feature is that they bring about additional JDBC parameter bindings. In last PR I registered these filter parameter bindings via `QuerySpec` which is tightly coupled with `loader` and not workable for `HQL/Criteria`. To support both `loader` and `HQL/Criteria`, this PR refactored previous implementation to store filter parameter to `JdbcSelect` instead, which accommodates both scenarios well. However, better idea is more than welcome!

As in last PR, extensive testing cases were added.

### DIFF
--- a/documentation/src/test/java/org/hibernate/userguide/pc/FilterJoinTableTest.java
+++ b/documentation/src/test/java/org/hibernate/userguide/pc/FilterJoinTableTest.java
@@ -18,15 +18,12 @@ import javax.persistence.OneToMany;
 import javax.persistence.OrderColumn;
 
 import org.hibernate.Session;
-import org.hibernate.annotations.Filter;
 import org.hibernate.annotations.FilterDef;
 import org.hibernate.annotations.FilterJoinTable;
 import org.hibernate.annotations.ParamDef;
 import org.hibernate.jpa.test.BaseEntityManagerFunctionalTestCase;
 
 import org.junit.Test;
-
-import org.jboss.logging.Logger;
 
 import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
 import static org.junit.Assert.assertEquals;
@@ -117,10 +114,6 @@ public class FilterJoinTableTest extends BaseEntityManagerFunctionalTestCase {
             name="maxOrderId",
             type="int"
         )
-    )
-    @Filter(
-        name="firstAccounts",
-        condition="order_id <= :maxOrderId"
     )
     public static class Client {
 
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/StandardSqmSelectTranslator.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/StandardSqmSelectTranslator.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.query.sqm.sql.internal;
 
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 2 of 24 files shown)


hibernate-core/src/main/java/org/hibernate/sql/ast/SqlTreePrinter.java
```diff
@@ -215,7 +215,7 @@ private void logNode(String text, Runnable subTreeHandler, boolean indentContinu
 
 	private void logWithIndentation(Object line) {
 		pad( depth );
-		buffer.append( line ).append( '\n' );
+		buffer.append( line ).append( String.format( "%n" ) );
 	}
```
NathanQingyangXu, line 219: Given that we need to support SQL Server, it might be a good idea to use `%n` instead of `\n` to accommodate Windows OS or others.

documentation/src/test/java/org/hibernate/userguide/pc/FilterJoinTableTest.java
```diff
@@ -118,10 +115,6 @@ public void testLifecycle() {
             type="int"
         )
     )
-    @Filter(
-        name="firstAccounts",
-        condition="order_id <= :maxOrderId"
-    )
```
NathanQingyangXu, line 124: The above filter is not used for `Client` but for its `accounts` collection field. Seems an overlooking and might be misleading.

NathanQingyangXu:
need more polishing

NathanQingyangXu:
The PR is ready for review.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/StandardSqmSelectTranslator.java
```diff
@@ -187,6 +195,24 @@ protected void prepareQuerySpec(QuerySpec sqlQuerySpec) {
 
 	@Override
 	protected void postProcessQuerySpec(QuerySpec sqlQuerySpec) {
+		final List<TableGroup> roots = sqlQuerySpec.getFromClause().getRoots();
+		if ( roots != null && roots.size() == 1 ) {
+			final TableGroup root = roots.get( 0 );
+			final ModelPartContainer modelPartContainer = root.getModelPart();
+			final EntityPersister entityPersister = modelPartContainer.findContainingEntityMapping().getEntityPersister();
+			assert entityPersister instanceof AbstractEntityPersister;
+			final String primaryTableAlias = root.getPrimaryTableReference().getIdentificationVariable();
```
sebersole, line 204: An enhancement here would be to support filtering on non-primary-table columns

NathanQingyangXu: I will implement this in the PR for https://trello.com/c/lF9R13J3/115-where-and-wherejointable, which will be created pretty soon.

NathanQingyangXu: Initially I thought this was easy to implement but then I found my code needs to invoke the following method from `org.hibernate.persister.entity.Joinable`.
```
	/**
	 * Get the where clause filter, given a query alias and considering enabled session filters
	 */
	public String filterFragment(String alias, Map<String, Filter> enabledFilters) throws MappingException;
```
seems only primary table alias is needed here. For that reason, I would be cautious to maintain the status quo and keep the code changes out of the new PR which has been created: https://github.com/hibernate/hibernate-orm/pull/3366 as I originally promised.

Plan to investigate further and will create separate PR if needed.

PR #3355, opened 2020-04-18
HHH-13932: Check for non existing query parameter in CriteriaQueryTypeQueryAdapter and throw an exception with the corresponding message

gkatiforis:
Jira ticket: [HHH-13932](https://hibernate.atlassian.net/browse/HHH-13932)

**Issue:** 
When the user attempts to set non existing query parameter using` Query::setParameter` a `NullPointerException` is thrown. This is due to `CriteriaQueryTypeQueryAdapter.explicitParameterInfoMap` not having a key for the parameter, as it was never used in the query.  

**Solution:**
Check for non existing query parameter and throw an `IllegalArgumentException` with the corresponding message.


gavinking:
Bug already fixed. Adding test in #9341.

PR #3356, opened 2020-04-20
HHH-13960 : Add SAXReader sec features to match the defaults

panossot:
HHH-13960 :  https://hibernate.atlassian.net/browse/HHH-13960

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/util/xml/XMLHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/util/xml/XMLHelper.java
@@ -59,13 +59,14 @@ public final class XMLHelper {
 
 	public SAXReader createSAXReader(ErrorLogger errorLogger, EntityResolver entityResolver) {
 		SAXReader saxReader = new SAXReader();
-                try {
-                    saxReader.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
-                    saxReader.setFeature("http://xml.org/sax/features/external-general-entities", false);
-                    saxReader.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
-                }catch(Exception e) {
-                    throw new RuntimeException( e );
-                }
+		try {
+			saxReader.setFeature( "http://apache.org/xml/features/nonvalidating/load-external-dtd", false );
+			saxReader.setFeature( "http://xml.org/sax/features/external-general-entities", false );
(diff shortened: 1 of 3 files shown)


gbadner:
@panossot, it looks like there's a checkstyle failure. It's probably because spaces were used instead of tabs. I'll fix it and push a new commit.

hibernate-core/src/main/java/org/hibernate/internal/util/xml/XMLHelper.java
```diff
@@ -59,6 +59,14 @@ public DocumentFactory getDocumentFactory() {
 
 	public SAXReader createSAXReader(ErrorLogger errorLogger, EntityResolver entityResolver) {
 		SAXReader saxReader = new SAXReader();
+		try {
+			saxReader.setFeature( "http://apache.org/xml/features/nonvalidating/load-external-dtd", false );
```
scottmarlow, line 63: Should we add a comment, perhaps copying description from [Xerces doc](http://xerces.apache.org/xerces-j/features.html)?
Could be something like:
`saxReader.setFeature( "http://apache.org/xml/features/nonvalidating/load-external-dtd", false );  // Ignore the external DTD completely
`

hibernate-core/src/main/java/org/hibernate/internal/util/xml/XMLHelper.java
```diff
@@ -59,6 +59,14 @@ public DocumentFactory getDocumentFactory() {
...
+			saxReader.setFeature( "http://xml.org/sax/features/external-general-entities", false );
```
scottmarlow, line 64: Similar for this one:
`saxReader.setFeature( "http://xml.org/sax/features/external-general-entities", false );  // Do not include external general entities.  `

hibernate-core/src/main/java/org/hibernate/internal/util/xml/XMLHelper.java
```diff
@@ -59,6 +59,14 @@ public DocumentFactory getDocumentFactory() {
...
+			saxReader.setFeature( "http://xml.org/sax/features/external-parameter-entities", false );
```
scottmarlow, line 65: please add line here also:
`saxReader.setFeature( "http://xml.org/sax/features/external-parameter-entities", false );  // Do not include external parameter entities or the external DTD subset.  `

Suggestion is because the comment gives me an idea of what the property settings accomplish.

Sanne:
Thanks! merged

PR #3358, opened 2020-04-20
HHH-13959 - Add nullability and uniqueness for @OneToOne with @JoinTable 

dreab8:
https://hibernate.atlassian.net/browse/HHH-13959

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
@@ -3113,7 +3113,14 @@ public final class AnnotationBinder {
 		propertyBinder.setCascade( cascadeStrategy );
 		propertyBinder.setProperty( property );
 		propertyBinder.setXToMany( true );
-		propertyBinder.makePropertyAndBind();
+
+		final Property boundProperty = propertyBinder.makePropertyAndBind();
+		if ( joinColumn != null ) {
+			boundProperty.setOptional( joinColumn.nullable() && optional );
+		}
+		else {
+			boundProperty.setOptional( optional );
+		}
 	}
 
 	protected static void defineFetchingStrategy(ToOne toOne, XProperty property) {
--- a/hibernate-core/src/test/java/org/hibernate/test/annotations/onetoone/OneToOneJoinTableUniquenessTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/annotations/onetoone/OneToOneJoinTableUniquenessTest.java
@@ -1,0 +1,236 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.test.annotations.onetoone;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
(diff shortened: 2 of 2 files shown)


pdinc-oss:
much, much nicer than our mess from 4AM.

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
```diff
@@ -3113,7 +3113,12 @@ else if (hasSpecjManyToOne) {
 		propertyBinder.setCascade( cascadeStrategy );
 		propertyBinder.setProperty( property );
 		propertyBinder.setXToMany( true );
-		propertyBinder.makePropertyAndBind();
+		if ( property.isAnnotationPresent( JoinTable.class ) ) {
+			propertyBinder.makePropertyAndBind().setOptional( optional );
+		}
+		else {
+			propertyBinder.makePropertyAndBind();
+		}
```
NathanQingyangXu, line 3121: I understand you wanna avoid a temporary variable for `makePropertyAndBind()` result, but it is still a little bit confusing why `propertyBinder.makePropertyAndBind()` remains the same across the two if blocks.

Maybe the following refactoring is a little bit better in terms of code readability
```
final Property boundProperty = propertyBinder.makePropertyAndBind();
if ( property.isAnnotationPresent( JoinTable.class ) ) {
    boundProperty.setOptional( optional );
}
```
Not a big deal and feel free to ignore my comment.

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
```diff
@@ -3113,7 +3113,12 @@ else if (hasSpecjManyToOne) {
 		propertyBinder.setCascade( cascadeStrategy );
 		propertyBinder.setProperty( property );
 		propertyBinder.setXToMany( true );
-		propertyBinder.makePropertyAndBind();
+		if ( property.isAnnotationPresent( JoinTable.class ) ) {
+			propertyBinder.makePropertyAndBind().setOptional( optional );
+		}
+		else {
+			propertyBinder.makePropertyAndBind();
```
jwgmeligmeyling, line 3120: Any particular reason why this clause differs from the case with the `JoinTable`? It would seem only sensible that the optionality is always relevant for the bound property.  I trust that it is not, but then I am somewhat surprised that this is an appropriate fix 😉

pdinc-oss: This patch is a bit different than the one I wrote, i'll look at it.

pdinc-oss: A quick look at it, the bindManyToOne method may be used outside of the JoinTable situation, so setting optional would be disallowed. Also not all usage of JoinTable gets you to this code, there has to be a OneToOne.

dreab8: Hi @jwgmeligmeyling I'm looking at the issue again, and I tend to agree with you. 

The problem with just setting `propertyBinder.makePropertyAndBind().setOptional( optional );` is for cases like the one in `NonNullableCircularDependencyCascadeTest` where we have 
```
@Entity
public class Child
{
   ...

   @ManyToOne(cascade=CascadeType.PERSIST)
   @JoinColumn(name="parentId", nullable=false)
	Parent parent; 
    ...
}
```

I'm pushed a slightly different solution.

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
dreab8: Thanks @NathanQingyangXu , I have applied your suggestion.

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
jwgmeligmeyling: Looks good to me!

dreab8: Thanks @jwgmeligmeyling for the review

dreab8:
applied upstream

PR #3359, opened 2020-04-20
HHH-13960 (5.4)

gbadner:
https://hibernate.atlassian.net/browse/HHH-13960


gbadner:
Created against wrong branch.

PR #3363, opened 2020-04-21
HHH-12997 Specific exception thrown when trying to @JoinFormula along with @OneToMany

Clopma:
Added test with [this](https://hibernate.atlassian.net/browse/HHH-12997?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel) expected faliure in order to track it and fix it at some point.

_Edit: Now the expected faliure is the custom exception._

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/annotations/formula/JoinFormulaWithOneToMany.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/annotations/formula/JoinFormulaWithOneToMany.java
@@ -1,0 +1,147 @@
+package org.hibernate.test.annotations.formula;
+
+import org.hibernate.annotations.JoinFormula;
+import org.hibernate.jpa.boot.internal.EntityManagerFactoryBuilderImpl;
+import org.hibernate.jpa.boot.internal.PersistenceUnitInfoDescriptor;
+import org.hibernate.testing.FailureExpected;
+import org.hibernate.testing.TestForIssue;
+import org.hibernate.testing.junit4.BaseUnitTestCase;
+import org.junit.Test;
+
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.ManyToOne;
+import javax.persistence.OneToMany;
+import javax.persistence.SharedCacheMode;
+import javax.persistence.Table;
+import javax.persistence.ValidationMode;
+import javax.persistence.spi.ClassTransformer;
+import javax.persistence.spi.PersistenceUnitInfo;
+import javax.persistence.spi.PersistenceUnitTransactionType;
+import javax.sql.DataSource;
+import java.net.URL;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Properties;
+import java.util.Set;
+
+public class JoinFormulaWithOneToMany extends BaseUnitTestCase {
+
+    @Test
+    @TestForIssue(jiraKey = "HHH-12997")
+    @FailureExpected(jiraKey = "HHH-12997")
+    public void testJoinFormulaWithOneToMany() {
+
+        PersistenceUnitInfo info = new PersistenceUnitInfo() {
+            @Override
+            public String getPersistenceUnitName() {
+                return "test";
+            }
+
+            @Override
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/test/annotations/formula/JoinFormulaWithOneToMany.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    @Entity @Table(name = "parent") class Parent {
+
+        @Id @GeneratedValue(strategy = GenerationType.AUTO) private Long id;
+
+        @OneToMany(mappedBy = "parent") @JoinFormula(value = "(SELECT * FROM child)") private Set<Child> childs;
+
+    }
+
```
NathanQingyangXu, line 146: From https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#associations-JoinFormula, it seems `@JoinFormula` should be used on `Child`, not on `Parent`, for basically it is a customized foreign key.

Clopma: It has the exact same behaviour both ways. But you are right, It should be the other way so let me change the test.

Clopma:
Updated the commit to match @NathanQingyangXu suggestion.

hibernate-core/src/test/java/org/hibernate/test/annotations/formula/JoinFormulaWithOneToMany.java
NathanQingyangXu: Why are you sure it has the same behaviour both ways? If it is not mentioned in the comprehensive user guide, we can assume it is not supported.

NathanQingyangXu:
> Updated the commit to match @NathanQingyangXu's suggestion.

I still don't think you correctly understand the usage. It is sort of customized foreign key. In your testing case, I failed to see the point to use it. There is no customization necessity here. It is simply a common parent child relationship. If you studied the user guide link I provided above, you would gain a better understanding of its usage.

Clopma:
> > Updated the commit to match @NathanQingyangXu's suggestion.
> 
> I still don't think you correctly understand the usage. It is sort of customized foreign key. In your testing case, I failed to see the point to use it. There is no customization necessity here. It is simply a common parent child relationship. If you studied the user guide link I provided above, you would gain a better understanding of its usage.

This is a simplified case, but let's say I want to make a more complex join than just a `SELECT * FROM child;` In those cases I need the `@JoinFormula`.

hibernate-core/src/test/java/org/hibernate/test/annotations/formula/JoinFormulaWithOneToMany.java
Clopma: It is not supported. But some people on the staff agree this is an interestig feature to add. Anyway, if It is not going to be done in a short term, in my opinion It should be throwing a custom exception saying this is not supported.

NathanQingyangXu:
> > > Updated the commit to match @NathanQingyangXu's suggestion.
> > 
> > 
> > I still don't think you correctly understand the usage. It is sort of customized foreign key. In your testing case, I failed to see the point to use it. There is no customization necessity here. It is simply a common parent child relationship. If you studied the user guide link I provided above, you would gain a better understanding of its usage.
> 
> This is a simplified case, but let's say I want to make a more complex `JOIN ` than just a `SELECT * FROM child;` In those cases I would need the `@JoinFormula`.

Yeah, but then there would not be testing failure any more.

Clopma:
> > > > Updated the commit to match @NathanQingyangXu's suggestion.
> > > 
> > > 
> > > I still don't think you correctly understand the usage. It is sort of customized foreign key. In your testing case, I failed to see the point to use it. There is no customization necessity here. It is simply a common parent child relationship. If you studied the user guide link I provided above, you would gain a better understanding of its usage.
> > 
> > 
> > This is a simplified case, but let's say I want to make a more complex `JOIN ` than just a `SELECT * FROM child;` In those cases I would need the `@JoinFormula`.
> 
> Yeah, but then there would not be testing failure any more.

I don't get what you are saying. This faliure occurs whatever you put in the custom `@JoinFormula`. It doesn't matter if it is a simple `SELECT` or a complex one.

PR #3364, opened 2020-04-22
Add rollup() and cube() for use in group by clause

gavinking:
The following SQL syntax is useful in reporting:

     select avg(price), store, item, cust from sales group by rollup(store, item, cust)

This syntax is supported on at least DB2, Oracle, SQL Server, and Postgres. It's not supported on MySQL.

Note that the ANSI SQL standard syntax for this is, apparently, the following:

     select avg(price), store, item, cust from sales group by store, item, cust with rollup

However, I don't find this to be better. MySQL has partial support for this syntax.

This patch adds support for `rollup()` and `cube()` to HQL.

Note that this is just a proposal and has not been tested, since the query parser does not yet support `group by`.

### DIFF
--- a/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlLexer.g4
+++ b/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlLexer.g4
@@ -157,6 +157,7 @@ COLLATE				: [cC] [oO] [lL] [lL] [aA] [tT] [eE];
 CONCAT				: [cC] [oO] [nN] [cC] [aA] [tT];
 COUNT				: [cC] [oO] [uU] [nN] [tT];
 CROSS				: [cC] [rR] [oO] [sS] [sS];
+CUBE				: [cC] [uU] [bB] [eE];
 CURRENT				: [cC] [uU] [rR] [rR] [eE] [nN] [tT];
 CURRENT_DATE		: [cC] [uU] [rR] [rR] [eE] [nN] [tT] '_' [dD] [aA] [tT] [eE];
 CURRENT_INSTANT		: [cC] [uU] [rR] [rR] [eE] [nN] [tT] '_' [iI] [nN] [sS] [tT] [aA] [nN] [tT]; //deprecated legacy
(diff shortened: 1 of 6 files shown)


sebersole:
Syntax wise I'm not sure the benefit of the parens aside from defining the grouping terms as a tuple.  Meaning I'm not sure that syntax is any better than 

```
select avg(price), store, item, cust from sales group by rollup store, item, cust
```
not the SQL variant.

Just a quick take.  I'll look at the PR more in depth today/tomorrow

gavinking:
@sebersole well it seems to me that the advantage of the parens is that:

- you can have additional grouped columns that *aren't* in the `cube()` or `rollup()`, which seems very useful to me, and
- it seems to be more easily extensible to the `grouping sets` syntax which is also supported by Postgres, DB2, and SQL Server (though I don't plan to add support for that at this stage)

sebersole:
Ah, I did not realize only part of the grouping could be rolled-up.  So then that part makes sense.  But...

Although, then I am not sure what the syntax for that particular part looks like for either syntax you mentioned.

```
select avg(price), store, item, cust, non_rolled_value from sales group by rollup(store, item, cust), non_rolled_value
```

Especially wrt the SQL /  MySQL syntax

gavinking:
@sebersole I checked that this works on Postgres:

    select avg(price), store, item, cust from sales group by cube(store, item), cust

I don't know how to do that on MySQ, and I strongly suspect it's impossible.

gavinking:
By the way, let me leave some references here:

<https://www.postgresql.org/docs/12/queries-table-expressions.html#QUERIES-GROUPING-SETS>

<https://www.ibm.com/support/knowledgecenter/en/SSEPGG_10.5.0/com.ibm.db2.luw.sql.ref.doc/doc/r0059215.html>

<https://docs.microsoft.com/en-us/sql/t-sql/queries/select-group-by-transact-sql?view=sql-server-ver15>

<https://docs.oracle.com/cd/F49540_01/DOC/server.815/a68003/rollup_c.htm#31838>

<https://dev.mysql.com/doc/refman/5.7/en/group-by-modifiers.html>

sebersole:
BTW, I wanted to point out Gavin since you are contributing a lot of new types of SQM nodes.  Because SQM is also the impl for our JPA Criteria tree, adding support for these should be as simple as adding methods for them on our Hibernate-specific CriteriaBuilder contract : `org.hibernate.query.criteria.HibernateCriteriaBuilder` / `org.hibernate.query.sqm.NodeBuilder`.  Of course users could just implement them directly and add them to the Criteria tree, but having the builder methods would be nice

gavinking:
@sebersole OK, good point, I guess I was being a bit careful about adding new application-visible APIs without careful thought about what they should look like but if you like I can go and do some of that careful thinking.

Do you have any thoughts on which of these new things are higher/lower priority?

sebersole:
I was actually planning to do that already for the existing ones we've done
already.  More I just meant moving forward.  I need to keep it in mind also
when reviewing these.

gavinking:
OK.

PR #3365, opened 2020-04-22
6.0 envers test enablement

Naros:
Here's a quick pass to try and get Envers tests working.  

Run `org.hibernate.envers.test.integration.basic.Simple` test to observe the 
```
org.hibernate.sql.ast.SqlTreeCreationException: Could not locate TableGroup - NavigablePath[org.hibernate.envers.test.entities.IntTestEntity_AUD(e__).originalId.REV]
javax.persistence.PersistenceException: org.hibernate.sql.ast.SqlTreeCreationException: Could not locate TableGroup - NavigablePath[org.hibernate.envers.test.entities.IntTestEntity_AUD(e__).originalId.REV]
```

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/internal/AttributeFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/internal/AttributeFactory.java
@@ -220,8 +220,7 @@ public class AttributeFactory {
 				final Component component = (Component) typeContext.getHibernateValue();
 				final EmbeddableTypeImpl<Y> embeddableType;
 
-				if ( component.getComponentClass() != null
-						|| component.getComponentClassName() != null ) {
+				if ( component.getComponentClassName() != null ) {
 					// we should have a non-dynamic embeddable
 
 					final Class embeddableClass;
@@ -261,8 +260,16 @@ public class AttributeFactory {
 					return embeddableType;
 				}
 				else {
+
+					final ManagedTypeRepresentationStrategy representationStrategy = context.getTypeConfiguration()
+							.getMetadataBuildingContext()
+							.getBuildingOptions()
+							.getManagedTypeRepresentationResolver()
+							.resolveStrategy( component, context.getRuntimeModelCreationContext() );
+
 					embeddableType = new EmbeddableTypeImpl(
 							component.getRoleName(),
+							representationStrategy,
 							context.getJpaMetamodel()
 					);
 				}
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/internal/AttributeFactory.java
```diff
@@ -220,8 +220,7 @@ public Property getPropertyMapping() {
 				final Component component = (Component) typeContext.getHibernateValue();
 				final EmbeddableTypeImpl<Y> embeddableType;
 
-				if ( component.getComponentClass() != null
-						|| component.getComponentClassName() != null ) {
+				if ( component.getComponentClassName() != null ) {
```
Naros, line 223: One thing I noticed is that calling `Component#getComponentClassName()` will fail for map-mode mappings because no component-class-name has been set.  I could have reversed checks and used `&&` instead, I guess its a matter more of preference as the end-result is the same.

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/JpaMetamodelImpl.java
```diff
@@ -515,7 +515,7 @@ public void processJpa(
 			PersistentClass persistentClass,
 			MetadataContext context,
 			TypeConfiguration typeConfiguration) {
-		final Class javaType = persistentClass.getMappedClass();
+		final Class javaType = persistentClass.getMappedClass() == null ? Map.class : persistentClass.getMappedClass();
```
Naros, line 518: It's been a while but I seem to recall that `PersistentClass#getMappedClass` would return `Map.class` rather than `null` if the implementation was map-mode.  There is likely a better way to resolve this, so I leave that up to you guys on what's the best approach here.

sebersole:
I've gotten pretty far with this locally.  I created some tests for HBM based mappings in various scenarios including dynamic model with a composite-id including key-many-to-one.  Oddly it works for me from those tests, but the envers test you pointed me to (`Simple`) fails still.  I'll look more next week

Naros:
@sebersole if you have some HBM tests that pass, feel free to point me to those and I'll be happy to take a look too.

sebersole:
See `org.hibernate.orm.test.bootstrap.binding.hbm`

beikov:
Going to close this as we fully re-enabled the envers testsuite already.

PR #3366, opened 2020-04-22
wip/6.0 implement @Where and @WhereJoinTable

NathanQingyangXu:
https://trello.com/c/lF9R13J3/115-where-and-wherejointable

So `@Where` and `@WhereJoinTable` belongs to the same `filter` umbrella, together with `@Filter` and `@FilterJoinTable`. We simply generalize existing implementation to accommodate it easily.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
@@ -975,7 +975,7 @@ public abstract class CollectionBinder {
 		FilterJoinTable simpleFilterJoinTable = property.getAnnotation( FilterJoinTable.class );
 		if ( simpleFilterJoinTable != null ) {
 			if ( hasAssociationTable ) {
-				collection.addManyToManyFilter(simpleFilterJoinTable.name(), simpleFilterJoinTable.condition(),
+				collection.addFilter(simpleFilterJoinTable.name(), simpleFilterJoinTable.condition(),
 						simpleFilterJoinTable.deduceAliasInjectionPoints(),
 						toAliasTableMap(simpleFilterJoinTable.aliases()), toAliasEntityMap(simpleFilterJoinTable.aliases()));
 					}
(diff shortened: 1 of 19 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
```diff
@@ -990,7 +990,7 @@ private void bindFilters(boolean hasAssociationTable) {
 		if ( filterJoinTables != null ) {
 			for (FilterJoinTable filter : filterJoinTables.value()) {
 				if ( hasAssociationTable ) {
-					collection.addManyToManyFilter(filter.name(), filter.condition(),
+					collection.addFilter(filter.name(), filter.condition(),
 							filter.deduceAliasInjectionPoints(),
 							toAliasTableMap(filter.aliases()), toAliasEntityMap(filter.aliases()));
 				}
```
NathanQingyangXu, line 996: The above code change is to revert back erroneous changes in some previous PR.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -380,22 +381,20 @@ private void applyKeyRestriction(
...
+		assert joinable instanceof AbstractCollectionPersister;
+		final String tableExpression = joinable.getTableName();
+		final String tableAlias = tableGroup.resolveTableReference( tableExpression ).getIdentificationVariable();
+		final Predicate filterPredicate = FilterHelper.createFilterPredicate(
+				loadQueryInfluencers,
+				joinable,
+				tableAlias
+		);
+		if ( filterPredicate != null ) {
+			querySpec.applyPredicate( filterPredicate );
```
NathanQingyangXu, line 397: The above change is simply to get rid of `if ( loadQueryInfluencers.hasEnabledFilters() ) {` for it will add `@Where` and `@WhereJoinTable` filter content as well. This is the gist of this PR.

PR #3367, opened 2020-04-22
HHH-13969 Fix handling of large varbinary in SybaseASE15Dialect

Dan-Cao:
https://hibernate.atlassian.net/browse/HHH-13969

Updates the handling of the VARBINARY type so when a using column size larger than what Sybase accepts, the `image` type is used instead.

This change mirrors the behaviour of other Transact-SQL dialects such as [SQLServerDialect](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/SQLServerDialect.java#L42).

Document on `varbinary` - size only goes up to 16K:
https://wiki.ispirer.com/sqlways/sybase/data-types/varbinary

Document on the `image ` type:
https://wiki.ispirer.com/sqlways/sybase/data-types/image

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/SybaseASE15Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/SybaseASE15Dialect.java
@@ -30,6 +30,8 @@ public class SybaseASE15Dialect extends SybaseDialect {
 	public SybaseASE15Dialect() {
 		super();
 
+		registerColumnType( Types.VARBINARY, "image" );
+		registerColumnType( Types.VARBINARY, 8000, "varbinary($l)" );
 		registerColumnType( Types.LONGVARBINARY, "image" );
 		registerColumnType( Types.LONGVARCHAR, "text" );
 		registerColumnType( Types.BIGINT, "bigint" );


hibernate-core/src/main/java/org/hibernate/dialect/SybaseASE15Dialect.java
```diff
@@ -30,6 +30,8 @@
 	public SybaseASE15Dialect() {
 		super();
 
+		registerColumnType( Types.VARBINARY, "image" );
+		registerColumnType( Types.VARBINARY, 8000, "varbinary($l)" );
```
NathanQingyangXu, line 34: `8000` is the max length for SQLServer. Should we use `32767` instead here?

gavinking: That's what I decided it should be for Sybase Anywhere. (But apparently for ASE it depends on the version.)

gavinking: Actually it also depends on the page size which can be as small as 2k.

Noupoi: Yep, it's based on the database page size which is configurable.

On reflection, we should look up and use the actual page size instead of a fixed value of `8000`.

I couldn't find a quick way to do this after looking through the other dialects.

Can you give any pointers to how this would be done? I would be happy to give that a go.

NathanQingyangXu:
Sorry, I have zero experience on SyBase, but I guess there must be some way.

gavinking:
This is now handled differently, and the max is set to `16_384` in `SybaseASEDialect`.

PR #3371, opened 2020-04-23
HHH-13975 : geolatte-geom support for SpatialPredicates

daniel-shuy:
This enhances `SpatialPredicates` (added in version `5.4.13`, see #3159 for the original PR) to support `geolatte-geom` (in addition to `JTS`).

To clearly distinguish the classes, I've renamed `SpatialPredicates` to `JTSSpatialPredicates`, and added a new class `GeolatteSpatialPredicates` for `geolatte-geom`. ~~This unfortunately, is a minor breaking change~~.

Please review by commit, else the file renaming will mess with the Diffs.

### DIFF
--- a/hibernate-spatial/src/main/java/org/hibernate/spatial/predicate/GeolatteFilterPredicate.java
+++ b/hibernate-spatial/src/main/java/org/hibernate/spatial/predicate/GeolatteFilterPredicate.java
@@ -1,0 +1,104 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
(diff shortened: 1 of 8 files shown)


NathanQingyangXu:
@maesenka Would you mind take a look at this PR? You approved the last PR from the same author previously.

daniel-shuy:
I've added back `SpatialPredicates`, and made it extend `JTSSpatialPredicates` and added a deprecation notice instead so that this is no longer a breaking change

daniel-shuy:
Thanks again @NathanQingyangXu and @maesenka!

PR #3372, opened 2020-04-23
HHH-13976 - Add BEFORE_TRANSACTION_COMPLETION release mode

barreiro:


Sanne:
This originally failed the build but I can't see the problem now, did you fix it @barreiro ?

barreiro:
The test was missing credentials when creating the database, so it would fail when running in the test suite. I fixed that.

Sanne:
merged! thanks Luis!

PR #3377, opened 2020-04-24
HHH-13980 NullPointerException in AbstractEntityGraphVisitationStrategy.startingCollectionIndex

reda-alaoui:
https://hibernate.atlassian.net/browse/HHH-13980


sebersole:
Thanks.  And its "fine" you don't include a test, but realize that tests are not just about proving this PR fixes the immediate concern.  They also are insurance against them happening again.

I'm not going to write a test either.

I've taken a look and the change it simple, so I'll apply it.  But I'll wait until tomorrow to see if you change your mind about a test.

reda-alaoui:
Hi @sebersole ,

I know the test is to prevent regression. I would love to create a test, but I could not find the exact combination of elements leading to the NPE :/

sebersole:
Considering it happens in `#startingCollectionIndex` I would assume it is trying to process a Map probably with an embeddable or entity key.

sebersole:
This was a simple change, so as I said before I am ok applying this without a test.  From now on please include a test or we won't really be able to apply changes.

Thanks

PR #3379, opened 2020-04-26
wip/6.0 Implement @Formula

NathanQingyangXu:
https://trello.com/c/zUReFP2N/77-formula-support

Majority of the changed files are due to a new `isColumnFormula` method parameter to `ColumnConsumer`.

The main logic change is in the single `org.hibernate.sql.ast.tree.expression.ColumnReference` class which accepts a new boolean `isColumnFormula` constructor parameter to denote whether the existing `columnExpression` corresponds to a formula SQL fragment. Other than that, everything is straightforward.

The four `@Formula` related testing cases in `org.hibernate.test.annotations.formula` have been enabled, together with a `FormulaBasicsTest` copied from user guide.

Existing unimplemented method in `MappingModelCreationHelper` prohibits the implementation of `@JoinFormula` and for that reason, a Trello card was created to solve that separately: https://trello.com/c/ssoKQgnC/128-joinformula-and-joincolumnorformula-support


NathanQingyangXu:
forgot to implement HQL/Criteria. close for the time being.

NathanQingyangXu:
more testing cases need to be enabled

hibernate-envers/src/main/java/org/hibernate/envers/boot/EnversBootLogger.java
```diff
@@ -45,7 +45,7 @@ static Logger subLogger(String subName) {
 	@LogMessage( level = INFO )
 	@Message(
 			value = "Envers-generated HBM mapping...%n%s",
-			id = 90000001
+			id = 90005601
```
NathanQingyangXu, line 48: This code change is to bypass a CI compilation error that the id doesn't fall within the id range. Seems an overlooking by Steve's recent merging.

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/AttributeContainer.java
```diff
@@ -26,7 +26,7 @@
 
 		/**
 		 * Callback used when we have a singular id attribute of some form - either a simple id
-		 * or an aggregated composite id ({@link javax.persistence.EmbeddedId
+		 * or an aggregated composite id ({@link javax.persistence.EmbeddedId})
```
NathanQingyangXu, line 29: fixed another recent code changes made by Steve to bypass check style error

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/expression/ColumnReference.java
```diff
@@ -25,32 +25,43 @@
...
+		}
+		else {
+			this.columnExpression = columnExpression;
+		}
+		this.isColumnExpressionFormula = isColumnExpressionFormula;
+		this.referenceExpression = this.qualifier == null || isColumnExpressionFormula
+				? this.columnExpression
+				: this.qualifier + "." + this.columnExpression;
 		this.jdbcMapping = jdbcMapping;
 	}
```
NathanQingyangXu, line 56: This class change should be the facade of the code changes in this PR and starting with it is a good idea to grasp this seemingly big PR.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/expression/ColumnReference.java
```diff
@@ -108,7 +123,7 @@ public boolean equals(Object o) {
 
 	@Override
 	public int hashCode() {
-		return referenceExpression.hashCode();
+		return Objects.hashCode( referenceExpression );
```
beikov, line 126: Revert this. `referenceExpression` is non-null AFAICS

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/expression/ColumnReference.java
```diff
@@ -25,32 +25,43 @@
  * Models a reference to a Column in a SQL AST
  *
  * @author Steve Ebersole
+ * @author Nathan Xu
  */
 public class ColumnReference implements Expression, Assignable {
 	private final String qualifier;
 	private final String columnExpression;
+	private final boolean isColumnExpressionFormula;
```
beikov, line 33: AFAICS you are not using this so IMO you can remove this. It's enough to check in the constructor but there is no need for the state later

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -826,7 +826,8 @@ else if ( batchSize > 1 ) {
 				colAliases[k] = thing.getAlias( dialect, prop.getValue().getTable() );
 				if ( thing.isFormula() ) {
 					foundFormula = true;
-					( (Formula) thing ).setFormula( substituteBrackets( ( (Formula) thing ).getFormula() ) );
+					// ( (Formula) thing ).setFormula( substituteBrackets( ( (Formula) thing ).getFormula() ) );
+					// TOOD: uncomment the above statement when this#substituteBrackets(String) is implemented
```
beikov, line 830: Do you have a card for this so that we don't forget about it?

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/expression/ColumnReference.java
NathanQingyangXu: I changed this to align well with the `equals()` implementation, which has used `Objects.equals()` already.
I both reverted this and changed the `equals()` implementation to avoid unnecessary nullness checking.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/expression/ColumnReference.java
NathanQingyangXu: Yeah, currently it is not used publicly, but I do think it is a basic feature of this class and we are supposed to expose it for later use.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
NathanQingyangXu: Nope. In current v6 development, it would be ideal to fix issues related as much as possible. I left a `//TODO` comment above so I think it might be enough. There are too many `NotYetImplementedFor6Exception()` statements here and there in current v6 branch and creating a specific card for each of them might be too fine-grained, IMHO.
However, I did fix the typo (//TOOD -> //TODO).

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ColumnConsumer.java
```diff
@@ -15,5 +15,6 @@
 	void accept(
 			String containingTableExpression,
 			String columnExpression,
+			boolean isColumnExpressionFormula,
```
sebersole, line 18: Outside the scope of this PR, but as a larger question here we also need to start thinking about how `@ColumnTransformer` is supported here.  May be better to just pass along the "read" and "write" fragment portions?

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/BasicValuedSingularAttributeMapping.java
```diff
@@ -73,13 +78,21 @@ public BasicValuedSingularAttributeMapping(
...
 		this.mappedColumnExpression = mappedColumnExpression;
+		this.isMappedColumnExpressionFormula = isMappedColumnExpressionFormula;
 		this.valueConverter = valueConverter;
 		this.jdbcMapping = jdbcMapping;
 
 		if ( valueConverter == null ) {
 			domainTypeDescriptor = jdbcMapping.getJavaTypeDescriptor();
 		}
 		else {
+			if ( isMappedColumnExpressionFormula ) {
```
sebersole, line 89: Why disallow this?  In theory it should be fine.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ordering/ast/SequencePart.java
```diff
@@ -16,6 +16,7 @@
 public interface SequencePart {
 	SequencePart resolvePathPart(
 			String name,
+			boolean isFormula,
```
sebersole, line 19: The added `isFormula` is not needed.  It is always `false`.

NathanQingyangXu: Deleted it. Thanks for the comment.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/BasicValuedSingularAttributeMapping.java
NathanQingyangXu: I forgot what was in my mind when I coded this a long time ago, :). I deleted this as you suggested. Thanks.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ColumnConsumer.java
NathanQingyangXu: Good point. I am gonna create a Trello ticket for `@ColumnTransformer` soon. Yeah, adding "read" and "write" fragment seems accommodate every possible scenario (normal column, formula and columnTransformer). But from design perspective, maybe it is an idea to create a higher level class to include the three cases (maybe via class hierarchy). That would simplify code understanding and ease code maintaining.

NathanQingyangXu:
> This looks great Nathan. Thanks!
> 
> Wanted to discuss adding `isFormula` to the `#resolvePathPart` method, but otherwise +1.
> 
> Next step I think is to add support for `@ColumnTransformer`. I think all that is really needed is the read and the write fragment, along with the `JdbcMapping` in that signature:
> 
> ```
> public interface ColumnConsumer {
> 	void accept(
> 			String containingTableExpression,
> 			String name,
> 			String readExpression,
> 			String writeExpression,
> 			JdbcMapping jdbcMapping);
> ```
> 
> `name` added for informational mainly

That is definitively a good idea for it accommodates all possibilities, but it is tightly coupled with `@ColumnTransformer` and brings about some code understanding and maintaining issue in the long run. As first attempt I would consider creating a higher level class (maybe named `Column` and map the various cases into class hierarchy) to encapsulate the column complexity outside consumer. If so, we don't need the `ColumnConsumer` in the first place for we can replace it with `Consumer<Column>` directly.
I've created a Trello card for this: https://trello.com/c/aduHCQE5/139-columntransformer-support. Definitively I would explore your approach as the last resort for sure.

PR #3381, opened 2020-04-28
HHH-12997 Throwing custom exception when using @JoinFormula along with @OneToMany

Clopma:
Added a custom exception when failing to parse column as formula.

This annotation combination is not supported yet, even if [there are plans to add support to it](https://hibernate.atlassian.net/browse/HHH-9897?focusedCommentId=75063&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-75063) and some colaborators agree this is a [convinient feature to add support to](https://github.com/hibernate/hibernate-orm/pull/2547#issuecomment-438153730).

Steve Ebersole added a test to keep track of an error that ocurred previously when trying this, but now [this exception](https://hibernate.atlassian.net/browse/HHH-12997) is thrown. Anyway, if It is not going to be done in a short term, in my opinion there should be a custom exception being thrown.

### Previous exception:

`java.lang.ClassCastException: org.hibernate.mapping.Formula cannot be cast to org.hibernate.mapping.Column`

### New exception:

`org.hibernate.AnnotationException: Formula is not supported along with this type of relationship
`

Also added a test with `@FaliureExpected` in order to extend the track of the use case.


gavinking:
Thanks, this was a good idea; superseded by #9320

PR #3383, opened 2020-04-30
HHH-13991 - Make the TimstampType return the actual mapped Type (e.g., java.util.Date or java.sql.Timestamp) instead of always returning java.sql.Timestamp

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-13991


vladmihalcea:
@dreab8 @sebersole If you have time, can you take a look at this fix proposal?

Thanks.

gavinking:
Very out of date, and `java.sql` types are deprecated.

PR #3384, opened 2020-04-30
HHH-13996: Update documentation for HANA Cloud

breglerj:


hibernate-core/src/main/java/org/hibernate/boot/registry/selector/internal/DefaultDialectSelector.java
```diff
@@ -136,6 +137,9 @@
 		if ( name.equals( "H2" ) ) {
 			return H2Dialect.class;
 		}
+		if ( name.equals( "HANACloudColumnStore" ) ) {
```
Sanne, line 140: nice catch! thanks.

Sanne:
merged! thanks

PR #3385, opened 2020-05-01
HHH-13997 : Add methods to SessionImpl and StatefulPersistenceContext; make Loader method protected

gbadner:
https://hibernate.atlassian.net/browse/HHH-13997

* Change visibility for Loader#registerNonExists
* Methods for creating PersistenceContext and ActionQueue
* Add StatefulPersistenceContext#initializeNonLazyCollections(Consumer<PersistentCollection> initializeAction)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
@@ -221,8 +221,8 @@ public class SessionImpl
 	public SessionImpl(SessionFactoryImpl factory, SessionCreationOptions options) {
 		super( factory, options );
 
-		this.actionQueue = new ActionQueue( this );
-		this.persistenceContext = new StatefulPersistenceContext( this );
+		this.persistenceContext = createPersistenceContext();
+		this.actionQueue = createActionQueue();
 
 		this.autoClear = options.shouldAutoClear();
 		this.autoClose = options.shouldAutoClose();
@@ -267,6 +267,14 @@ public class SessionImpl
 		if ( log.isTraceEnabled() ) {
 			log.tracef( "Opened Session [%s] at timestamp: %s", getSessionIdentifier(), getTimestamp() );
 		}
+	}
+
+	protected StatefulPersistenceContext createPersistenceContext() {
+		return new StatefulPersistenceContext( this );
+	}
+
+	protected ActionQueue createActionQueue() {
+		return new ActionQueue( this );
 	}
 
 	private LockOptions getLockOptionsForRead() {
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
```diff
@@ -269,6 +269,14 @@ public SessionImpl(SessionFactoryImpl factory, SessionCreationOptions options) {
 		}
 	}
 
+	protected StatefulPersistenceContext createPersistenceContext() {
```
Sanne, line 272: Ah this is tricky; the new protected methods you've added are capturing "this", but since they are invoked from the constructor the reference to "this" isn't fully initialized yet.  This might seem to work but I find it a bit scary, especially as I expect you want to override the protected method from a subclass - this would lead to subclasss methods being run while the parent instance isn't initialized.

Could you explore some alternatives?

gbadner: @DavideD, FYI...

DavideD: I'm not a big fun of this solution either. In this case it works because all the constructor involved don't do anything complicated. I will check if I can find a different solution.

DavideD: I think I have an idea

Sanne: thanks :)   If there's no better way we can keep it of course.

DavideD: What about this? https://github.com/DavideD/hibernate-orm/commit/e4294f8e9b7bfefe5494e1cfc8606c69f2665a27

@gbadner Could you check that it works for the Rx PR, please? I've tried it on a different PR and it should be OK but better be sure. Thanks

DavideD:
I've created a new PR: https://github.com/hibernate/hibernate-orm/pull/3407

hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
gavinking: Ah, just noticed this discussion.

@DavideD the new approach in #3407 is just as bad. Well, actually, it's worse, really, since we're passing `this` to an arbitrary function, not just calling a method of the subclass.

If we *really* want to fix this properly the Good fix is to make an abstract superclass with `getPersistenceContext()` and `getActionQueue()` methods.

But I'm not sure that's worth it in this case.

PR #3386, opened 2020-05-03
wip/6.0 Enable legacy filter testing cases

NathanQingyangXu:
The following bugs of existing filter implementation is fixed:
1. filter is ignored for 'update' and 'delete' (majority of files changes in this PR is because of this);
2. many-to-many erroneous implementation;
3. table aliases mismatch. A new Joinable#filterFragment() was added to accept `TableGroup` instead of table alias to better utilize `TableGroup` to sync up table aliases. Another big benefit is if filter accesses table not in `from clause` (e.g. `select count(u) from User u`), the subclass join could be created and inserted into `from clause` automatically, just simulating v5 behaviour.

The original `org.hibernate.test.filter` package is migrated to `org.hibernate.orm.test.filter`. Some failing testing cases are attributed to current v6 imperfection, especially the `DynamicFilterTest` class.

For too many v6 imperfections(bugs or not-yet-implemented features), the `org.hibernate.test.subclassfilter` package is not enabled. For the same reason, a good portion of enabled testing cases are annotated with `@FailureExpected`, but working cases did a good job to cover various edge cases.


NathanQingyangXu:
tons of stuff undone, :(

sebersole:
Looks great as always Nathan.  Thanks!

PR #3387, opened 2020-05-04
HHH-14000  Allow a custom Type that was registered via TypeContributions to be resolved by its class, not just its name

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-14000

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/spi/BasicTypeRegistration.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/spi/BasicTypeRegistration.java
@@ -6,6 +6,10 @@
  */
 package org.hibernate.boot.spi;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 import org.hibernate.type.BasicType;
 import org.hibernate.type.CompositeCustomType;
 import org.hibernate.type.CustomType;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/boot/spi/BasicTypeRegistration.java
```diff
@@ -25,7 +29,23 @@ public BasicTypeRegistration(BasicType basicType) {
 
 	public BasicTypeRegistration(BasicType basicType, String[] registrationKeys) {
 		this.basicType = basicType;
-		this.registrationKeys = registrationKeys;
+		Class returnedClass = basicType.getReturnedClass();
+
+		if ( returnedClass != null ) {
```
jwgmeligmeyling, line 34: Is and should this handling be dependent of `@TypeDef(defaultForType = Clasz.class)` ? And specifically do not register if its not set as a default for any type? To prevent possible collisions between types between custom basic types?

vladmihalcea: I think it should work for TypeContributions too. Do you see any possible issue?

hibernate-core/src/main/java/org/hibernate/boot/spi/BasicTypeRegistration.java
```diff
@@ -25,7 +29,23 @@ public BasicTypeRegistration(BasicType basicType) {
...
+			String returnedClassName = returnedClass.getName();
+
+			List<String> registrationKeysList = registrationKeys == null || registrationKeys.length == 0 ?
+					new ArrayList<>() :
+					new ArrayList<>( Arrays.asList( registrationKeys ) );
+
+			if ( !registrationKeysList.contains( returnedClassName ) ) {
+				registrationKeysList.add( returnedClassName );
+			}
+			this.registrationKeys = registrationKeysList.toArray( new String[] {} );
```
NathanQingyangXu, line 44: There might be some minor performance optimization opportunity here when class name can be found in the array. For instance
```
boolean found = // maybe by some array helper method
if ( !found) {
    // go about clumsy array appending, maybe by some array helper method as well
}
```
That way we can some unnecessary transforming between array and list when the class name has been found within the array.

beikov:
Closing. See the Jira issue for details.

PR #3388, opened 2020-05-06
wip/6.0 Implement 'NULLS (FIRST | LAST)' in HQL

NathanQingyangXu:
https://trello.com/c/B5jMaXvx/73-allow-nulls-precedence-in-hql

seems straightforward implementation


gavinking:
Good idea.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/ParameterTests.java
```diff
@@ -145,7 +145,7 @@ public void testEmbeddableUseInPredicates() {
 	public void testNullParamValues() {
 		inTransaction(
 				session -> {
-					session.createQuery( "from Person p where p.name.first = :p" ).setParameter( "p", null ).list();
```
beikov, line 148: This change and similar others are unnecessary. You should add `NULLS`, `FIRST` and `LAST` to the `identifier` rule in HqlParser.g4 to make these soft keywords.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
```diff
@@ -226,6 +228,22 @@ public void visitQuerySpec(QuerySpec querySpec) {
 
 	@Override
 	public void visitSortSpecification(SortSpecification sortSpecification) {
+		NullPrecedence nullPrecedence = sortSpecification.getNullPrecedence();
+		final boolean hasNullPrecedence = nullPrecedence != null && nullPrecedence != NullPrecedence.NONE;
+		if ( hasNullPrecedence && ! dialect.supportsNullPrecedence() ) {
+			appendSql( "case when (" );
```
beikov, line 234: This should be avoided when the null precedence matches the DBMS default.

NathanQingyangXu:
> Good idea.

thanks. I simply wrap up on top of Steve's code.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
NathanQingyangXu: I have introduced a new method to `Dialect`, i.e. `supportsNullPrecedence()`. The above code is only run when the DBMS doesn't support it.
Am I missing something?

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/ParameterTests.java
NathanQingyangXu: I've done that and the above testing case would fail for all of sudden 'FIRST' and 'LAST' becomes SQL keywords. The above testing case has conflicting 'first' and 'last' columns, thus the code changes.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
beikov: DBMS have a default null precedence in which case you don't need to render the nulls clause or the emulation. Here is an example for MySQL: https://gregrs-uk.github.io/2011-02-02/mysql-order-by-with-nulls-first-or-last/
Another example from the PostgreSQL documentation `By default, null values sort as if larger than any non-null value; that is, NULLS FIRST is the default for DESC order, and NULLS LAST otherwise.`: https://www.postgresql.org/docs/8.3/queries-order.html

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/ParameterTests.java
beikov: As discussed on Zulip, the `FIRST`, `LAST` and `NULLS` keywords should probably be registered as keywords in the dialects so that matching identifiers are properly quoted

NathanQingyangXu: Good point. I experimented by adding the three keywords to dialect and found no escaping for 'first' happened. Maybe a missing feature in v6 currently. For that reason, I would maintain the current approach to rename the conflicting columns.

NathanQingyangXu:
> Not sure if this can be done yet, but using the select item ordinal instead of the expression if possible would reduce the SQL size and is probably easier for the DBMS to grok. Would be cool if we could use that. Apart from that, I think that the null precedence emulation could also use the select alias if we know it, but not sure if that information is available somewhere.

I don't think v6 supports order by ordinal currently, if desirable.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
NathanQingyangXu: Another excellent PR comment. It might be a good optimization effort, but given the current maturity status of v6, I wanna keep it out of the scope of this PR and might consider it in the future.

NathanQingyangXu:
> I guess it's fine for a first step but I would love to see some of the optimizations I mentioned. Ordering by ordinal is supported by many DBMS by the way.

@sebersole @gavinking How do you think about the suggestion?

gavinking:
> Not sure if this can be done yet, but using the select item ordinal instead of the expression if possible would reduce the SQL size and is probably easier for the DBMS to grok. 

While I agree that this is sometimes nicer, IIRC (and note that I'm relying here on my memory from years and years ago&mdash;I have not worked with the Hibernate 5 query parser *recently*) it was not so easy to implement this given the way the query parser worked in Hibernate 5.

(But don't take that as an authoritative statement on the subject because surely many things have changed since I last looked at this.)

gavinking:
Oh oops, we're talking Hibernate 6 here. In this case I seem to recall I already made what I guess was an abortive attempt to make that change in new query parser. But I seem to recall I ran up against preexisting bugs in the new parser.

sebersole:
> Not sure if this can be done yet, but using the select item ordinal instead of the expression if possible would reduce the SQL size and is probably easier for the DBMS to grok. Would be cool if we could use that. Apart from that, I think that the null precedence emulation could also use the select alias if we know it, but not sure if that information is available somewhere.

In principal it is implemented - org.hibernate.sql.ast.tree.expression.SqlSelectionExpression.  Do not remember if there are specifically any tests against diff DBs

sebersole:
Applied manually.  Thanks Nathan

PR #3389, opened 2020-05-06
Improve interaction of 2LC with Enhanced Proxies

Sanne:
- https://hibernate.atlassian.net/browse/HHH-14004

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
@@ -100,7 +100,7 @@ public class DefaultLoadEventListener implements LoadEventListener {
 			if ( loadType.isNakedEntityReturned() ) {
 				//do not return a proxy!
 				//(this option indicates we are initializing a proxy)
-				event.setResult( load( event, persister, keyToLoad, loadType ) );
+				event.setResult( load( event, persister, keyToLoad, loadType, false ) );
 			}
 			else {
 				//return a proxy if appropriate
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
```diff
@@ -278,6 +282,20 @@ private Object proxyOrLoad(
 				return managed;
 			}
 
+			//Check for presence of the required data in 2LC, as in this case there is no compelling reason to
+			//instantiate a lazily initialized enhanced proxy: might as well return it initialized.
```
sebersole, line 286: The concern here is the overhead of the L2C interactions when we do not necessarily have to.  Even the look into the cache to find an existing entry can be prohibitive depending on the underlying cache.  

That is perhaps not compelling in your use case, but this is something specifically discussed before.

beikov: How about a new method `Object getForProxy(SharedSessionContractImplementor session, Object key, long txTimestamp) throws CacheException` in the cache SPI `EntityRegionAccessStrategy` to allow handling this case?
The cache implementor can then decide to return null if the access would be too expensive.

Sanne: @sebersole I agree but how compelling is this concern in this context? This extra 2LC cost is only applied for entities (and eventsources) which actually do have caching enabled, so isn't this already something that the user has been explicitly opting-in?

How would you do it otherwise, do you think the "on proxy initialization hit 2LC" approach would fare better? Or are you thinking of having some configuration?

@beikov If we are to explore that direction, it would need to be more complicted than "just returning null" as we need to know if the lookup was actually attempted - so to ensure that eventually when we're sure the data is needed we know if we should _really_ attempt the 2LC load, or if we rather shouldn't bother as this was already happened.

As you can see in my PR, I make sure that the 2LC lookup happens _at most once_.

NathanQingyangXu:
I think `L2C` might bee better than `2LC` and less confusing

Sanne:
> I think `L2C` might bee better than `2LC` and less confusing

I always called it "2nd level caching" :)

@sebersole please let me know how you'd like me to change this? Or do you agree that in this context the concern you rised is the lesser evil?

sebersole:
As I said, if it were me I'd prefer getting the proxy and the proxy resolving itself from the second-level cache if that really does not happen at the moment.

But, if y'all agree that this is a better approach its fine.

beikov:
I'd personally prefer this to be configurable and controllable by the cache implementation i.e. let the cache decide if the access is costly, but still allow the user to override this decision.

Sanne:
thanks @sebersole I'll give it a shot.

@beikov I guess you're right that this could need a bit more flexibility. But beyond it depending on the cache implementation and configuration, e.g. Infinispan could be used clustered (costly) and local (less so) but there will always be some contention involved, which implies the user should have the last word.

So I'd rather keep it simple in the first shot, we can iterate later on the improvements but I'm a bit reticent in spending too much time on improvement on 5.x

Sanne:
I'm closing this PR as I've merged the alternative solution, based on initializing the proxy from the cache like Steve suggested.

See #3397

PR #3390, opened 2020-05-07
wip/6.0 add 'reverse' property to @RequiresDialectFeature annotation

NathanQingyangXu:
We all encounter such common scenario. There has been a `DialectFeatureCheck` implementation names `Supportsxxxx` in `DialectFeatureChecks` class. Then we need to cover the dialects which fail to be validated by it. Previously we have to create another one (`DoesNotSupportsxxxx`) manually which is tedious and not elegant.

This PR aims to streamline such common scenario by simply specifying a new `reverse` property of `@RequiresDialectFeature` to true.

### DIFF
--- a/hibernate-testing/src/main/java/org/hibernate/testing/orm/junit/DialectFeatureChecks.java
+++ b/hibernate-testing/src/main/java/org/hibernate/testing/orm/junit/DialectFeatureChecks.java
@@ -142,12 +142,6 @@ abstract public class DialectFeatureChecks {
 		}
 	}
 
-	public static class DoesRepeatableReadNotCauseReadersToBlockWritersCheck implements DialectFeatureCheck {
-		public boolean apply(Dialect dialect) {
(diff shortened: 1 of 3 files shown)


hibernate-testing/src/main/java/org/hibernate/testing/orm/junit/DialectFeatureChecks.java
```diff
@@ -185,12 +179,6 @@ public boolean apply(Dialect dialect) {
 		}
 	}
 
-	public static class DoesNotSupportRowValueConstructorSyntax implements DialectFeatureCheck {
-		public boolean apply(Dialect dialect) {
-			return dialect.supportsRowValueConstructorSyntax() == false;
-		}
-	}
-
```
NathanQingyangXu, line 193: We don't need the above clumsy syntax sugar classes any more

PR #3391, opened 2020-05-09
HHH-14007 Add travis support for s390x

cnnaik:
Add travis support for s390x
Signed-off-by: cnaik <Chandranana.Naik@ibm.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ dist: trusty
 language: java
 
 jobs:
+  fast_finish: true
+  allow_failures:
+    - arch: s390x
   include:
     - stage: Oracle JDK 8
       jdk: oraclejdk8
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,33 @@ jobs:
   include:
     - stage: Oracle JDK 8
       jdk: oraclejdk8
+    - stage: OpenJDK 11
+      jdk: openjdk11
+      dist: bionic
+      arch: s390x
     - stage: AdoptOpenJDK 11.0.3
       install:
         - curl -L -o install-jdk.sh https://github.com/sormuras/bach/raw/master/install-jdk.sh
         - source ./install-jdk.sh --target ./openjdk11 --url https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.3%2B7/OpenJDK11U-jdk_x64_linux_hotspot_11.0.3_7.tar.gz
-
+    - os: linux
+      dist: bionic
+      arch: s390x
+      install:
+        - curl -L -o install-jdk.sh https://github.com/sormuras/bach/raw/master/install-jdk.sh 
+        - source ./install-jdk.sh --target ./openjdk11 --url https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.3%2B7/OpenJDK11U-jdk_s390x_linux_hotspot_11.0.3_7.tar.gz
 before_script:
   - java -version
-  - ./gradlew assemble
+  - if [ "$HOSTTYPE" == "s390x" ];then
+      ./gradlew --no-daemon assemble;
+    else
+      ./gradlew assemble;
+    fi
 script:
-  - ./gradlew check -Plog-test-progress=true
+  - if [ "$HOSTTYPE" == "s390x" ];then
+      ./gradlew --no-daemon check -Plog-test-progress=true;
+    else
+      ./gradlew check -Plog-test-progress=true;
+    fi
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/


sebersole:
I'm not a fan of adding to the number of Travis stages.  These Travis jobs are used to verify pushes and pull requests - each stage added adds significant time to the process.  I'll let others chime in, but -1 from me

beikov:
I don't know how much Travis is used within the Hibernate development, but I am generally in favor of testing all supported platforms/environments as often/early as possible, assuming that platform is considered supported.
How about defining it in the `allow_failures` section and enable `fast_finish`? That way it doesn't block PRs but we still run it to see if it succeeds.

Sanne:
I would like for us to be aware of regressions on such platforms, but I agree with @sebersole that such tests shouldn't slow us down.

Would it be a reasonable tradeoff to have them run as nightly / weekly jobs on master?

Clearly that won't spot issues introduced in a new PR until it was merged, but the likelyhood for that to happen seems very low as we don't do platform specific development.

sebersole:
> Would it be a reasonable tradeoff to have them run as nightly / weekly jobs on master?

So long as we can run jobs against s390 as part of the Jenkins CI environment I think that is the route we should go here.  Do you know if that is possible @Sanne ?

Sanne:
@sebersole yes, I think we have the right contacts now to ask for an s390 machine for CI. But wouldn't it be easier to use the travis step? I'm fine either way, just wondering.

sebersole:
Well the concern is still time to run these steps.  Unless we can find a way to run them in parallel it already takes significant time to run these Travis jobs.  Open to suggestions to make that better

Sanne:
I've never used Travis; @yrodiere can we make the Travis job run in parallel so that they don't add on the total wait time? Or just nightly/weekly, rather than triggered on commits and PRs?

If not (or if that's simpler..) we can ask an s390 machine to run a Jenkins node.

cnnaik:
> I've never used Travis; @yrodiere can we make the Travis job run in parallel so that they don't add on the total wait time? Or just nightly/weekly, rather than triggered on commits and PRs?
> 
> If not (or if that's simpler..) we can ask an s390 machine to run a Jenkins node.

@sebersole @Sanne @beikov 
We have now added `allow_failures `for s390x arch, and also added `fast_finish `in the jobs so travis build status will not be dependent on s390x build completion.

Also we can provide s390x machines for CI, please let us know regarding the same.

beikov:
@Sanne: Usually, open source projects get 5 concurrent jobs for free, so I don't think it will be a problem to include 2 more matrix entries. Each matrix entry is a separate job and they all run in parallel.

yrodiere:
> I've never used Travis; @yrodiere can we make the Travis job run in parallel so that they don't add on the total wait time?

I think that's what @cnnaik did.

> Or just nightly/weekly, rather than triggered on commits and PRs?

I guess that's possible, but maybe we should not spread our use of CI environments even more? Using Travis for PR and Jenkins for everything else is confusing enough as it is.

Personally I'd move everything to Jenkins (including PR builds) and just leave a very simple Travis build enabled on branches for contributors to use in their forks. But then I don't really have time to work on this, so I'm fine with this PR.

sebersole:
> Each matrix entry is a separate job and they all run in parallel.

Are you sure about that?  We already had 2 stages defined and they run serially.

beikov:
>Are you sure about that? We already had 2 stages defined and they run serially.

Pretty sure. The thing is, the concurrent job slots are shared throughout the organization, so maybe another PR or maybe even another project blocks the concurrent job slots.

beikov:
As you can see here: 
image

... multiple jobs run in parallel.

I guess the current pipeline doesn't run in parallel because we are missing a top-level matrix entry i.e. we need

```
matrix:
  fast_finish: true
  include:
    ...
```

rather than

```
include:
  ...
```

cnnaik:
@beikov Running jobs through stages is the requirement? If not and we want all jobs to run in parallel, we will have to change the travis file to remove stages as shown in the code snippet below. 
```
jobs:
  fast_finish: true
  allow_failures:
    - arch: s390x
  include:
    - name: Oracle JDK 8
      jdk: oraclejdk8
    - name: OpenJDK 11
      jdk: openjdk11
      dist: bionic
      arch: s390x
    - name: AdoptOpenJDK 11.0.3
      install:
        - curl -L -o install-jdk.sh https://github.com/sormuras/bach/raw/master/install-jdk.sh
        - source ./install-jdk.sh --target ./openjdk11 --url https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.3%2B7/OpenJDK11U-jdk_x64_linux_hotspot_11.0.3_7.tar.gz
    - name: AdoptOpenJDK-s390x
      os: linux
      dist: bionic
      arch: s390x
      install:
        - curl -L -o install-jdk.sh https://github.com/sormuras/bach/raw/master/install-jdk.sh 
        - source ./install-jdk.sh --target ./openjdk11 --url https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.3%2B7/OpenJDK11U-jdk_s390x_linux_hotspot_11.0.3_7.tar.gz
```

beikov:
>Running jobs through stages is the requirement?

I don't think so. I guess this happened by accident.

> If not and we want all jobs to run in parallel, we will have to change the travis file to remove stages as shown in the code snippet below.

Yeah, that's what I suggested as well. It seems `jobs` is the new keyword and `matrix` is deprecated, although they are referring to `matrix.includes` everywhere in the documentation.

beikov:
Ahh, it seems the definition of stages imposes the sequential execution order. If we want to just name the jobs, we should use the `name` keyword instead of `stage`: https://docs.travis-ci.com/user/customizing-the-build#naming-jobs-within-matrices

Also see https://docs.travis-ci.com/user/build-stages/#how-to-define-build-stages

cnnaik:
>  we will have to change the travis file to remove stages as shown in the code snippet below.
> 
> ```
> jobs:
>   fast_finish: true
>   allow_failures:
>     - arch: s390x
>   include:
>     - name: Oracle JDK 8
>       jdk: oraclejdk8
>     - name: OpenJDK 11
>       jdk: openjdk11
>       dist: bionic
>       arch: s390x
>     - name: AdoptOpenJDK 11.0.3
>       install:
>         - curl -L -o install-jdk.sh https://github.com/sormuras/bach/raw/master/install-jdk.sh
>         - source ./install-jdk.sh --target ./openjdk11 --url https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.3%2B7/OpenJDK11U-jdk_x64_linux_hotspot_11.0.3_7.tar.gz
>     - name: AdoptOpenJDK-s390x
>       os: linux
>       dist: bionic
>       arch: s390x
>       install:
>         - curl -L -o install-jdk.sh https://github.com/sormuras/bach/raw/master/install-jdk.sh 
>         - source ./install-jdk.sh --target ./openjdk11 --url https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.3%2B7/OpenJDK11U-jdk_s390x_linux_hotspot_11.0.3_7.tar.gz
> ```

If required, i can update the PR to remove stages and use name keyword as shown in the above code snippet. Please let me know if that is fine.

beikov:
For me this is fine and this change would IMO have to be done anyway, but I don't know if we officially consider the `s390x` architecture to be supported

sebersole:
Well, again, my only real concern is the serial/parallel execution.  As I said, atm, each "step" in these builds is executed serially.  If this really does make them parallel then my major concern with doing this is removed.

However, where does this stop?

I understand why s390x is important now.  And we will add it because, well, we will have to.  I think we need an understanding that this is not "opening the door" for exploding the number of these "steps"

beikov:
> Well, again, my only real concern is the serial/parallel execution. As I said, atm, each "step" in these builds is executed serially. If this really does make them parallel then my major concern with doing this is removed.

Yeah, replacing `stage` with `name` fixes the serial vs. parallel execution issue.

>However, where does this stop?

Right, that is my concern as well, especially because I am unsure if testing against different architectures really makes sense as the JVM should abstract these details. I'd personally rather like to see different DBMS and Java versions being tested on travis than different architectures or JDK vendors. I guess a nightly run on Jenkins would be ok as well or maybe we can do a scheduled execution on travis as well?

yrodiere:
> However, where does this stop?

That was exactly the point of using stages, IIRC. Only use compute power to test additional environments when we are certain it works in a "reference" environment. But if it's free, I guess we can spend compute power...

> Right, that is my concern as well, especially because I am unsure if testing against different architectures really makes sense as the JVM should abstract these details. I'd personally rather like to see different DBMS and Java versions being tested on travis than different architectures or JDK vendors. I guess a nightly run on Jenkins would be ok as well or maybe we can do a scheduled execution on travis as well?

I'll take this opportunity to copy/paste what I suggested above :) 

> Personally I'd move everything to Jenkins (including PR builds) and just leave a very simple Travis build enabled on branches for contributors to use in their forks. But then I don't really have time to work on this, so I'm fine with this PR.

sebersole:
> > Personally I'd move everything to Jenkins (including PR builds) and just leave a very simple Travis build enabled on branches for contributors to use in their forks. But then I don't really have time to work on this, so I'm fine with this PR.
> 
> At the very least it'd be nice to discuss that kind of things at the team level, so that we don't start using wildly different CI tools for every single project.

That's kind of what we are doing ;)

My concern with Jenkins is that it is a PITA to set up jobs IMO.  Even "cloning" is a pain.  It now offers something similar to a Travis-style config.  Maybe that makes it easier.

I am looking more at Travis today because I want to start enabling cross-database testing for 6.  I'll investigate alternatives to Jenkins UI as well.

yrodiere:
> That's kind of what we are doing ;)

I meant with people from other projects ;)

> My concern with Jenkins is that it is a PITA to set up jobs IMO.

With the UI, true.

> Even "cloning" is a pain.

There's an option when creating a job to copy from another job. But I agree the UI being messy afterwards makes this a pain anyway.

> It now offers something similar to a Travis-style config IIUC. Maybe that makes it easier.?

Depends what you want to do. To get the same thing as Travis with the declarative syntax, I'd say it's good enough. To go above and beyond... it can be tough, but at least you can. Let's say the initial setup can be very complex, but after that you can evolve it rather easily.

I configured just that in Hibernate Search, and now we have a nice folder with one build per branch or PR: https://ci.hibernate.org/view/Search/job/hibernate-search/
... and a nice multi-configuration build on main branches: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/master/816/pipeline

The configuration file: https://github.com/hibernate/hibernate-search/blob/master/Jenkinsfile
It's quite long and a bit complex at first glance, but please keep in mind it could be much simpler with simpler requirements. I tried to do things that may not be easy to do with a declarative configuration (though that was a few years ago, maybe nowadays it's easier).

sebersole:
Awesome!  Thanks Yoann, I'll take a look at these.

cnnaik:
Closing the pull request as Travis support is removed - https://github.com/hibernate/hibernate-orm/commit/b166e684e1e5a7c8062d63e938a2a2deb0e9729b

PR #3392, opened 2020-05-09
wip/6.0 Implement @RowId

NathanQingyangXu:
https://trello.com/c/jNNzMzl2/108-implement-rowid-support


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/RootEntityResultImpl.java
```diff
@@ -105,6 +110,19 @@ public RootEntityResultImpl(
...
+		}
+		else {
+			rowIdResult = rowIdMapping.createDomainResult(
+					navigablePath.append( rowIdMapping.getRowIdName() ),
+					entityTableGroup,
+					AbstractEntityPersister.ROWID_ALIAS,
+					creationState
+			);
+		}
+
```
dreab8, line 125: I think the same should be done for `org.hibernate.sql.results.graph.entity.internal.EntityResultImpl`

NathanQingyangXu: I refactored the code and move the logic into `AbstractEntityResultGraphNode` instead. Thanks for the review.

PR #3393, opened 2020-05-10
wip/6.0 Replace 'AtomicInteger' with 'MutableInteger' in single thread scenario

NathanQingyangXu:
There seems to be an established pattern to use `AtomicInteger` as a convenient class of `mutable integer container`, regardless of whether it is used in single thread scenario (e.g. within method as a temporary variable without being exposed externally) or not. That is a misuse or sloppiness for we have to pay the performance cost of concurrency safety unnecessarily.

Luckily, in some recent PR by Andrea, a more appropriate substitute has been committed: `MutableInteger`.
 
Below is the **jmh** benchmark result between `AtomicInteger` and `MutableInteger` when it comes to increasing by 1:
```
Benchmark                Mode  Cnt   Score     Error    Units
benchmarkAtomicInteger   avgt   5       13.143 ± 6.360  ns/op
benchmarkMutableInteger  avgt   5       0.399 ± 0.045   ns/op
```

It is a simple fact that `MutableInteger` is much much quicker!

This is the simple **jmh** benchmark methods:
```
    @Benchmark
    @BenchmarkMode(Mode.AverageTime)
    @OutputTimeUnit(TimeUnit.NANOSECONDS)
    public void benchmarkAtomicInteger() {
        AtomicInteger counter = new AtomicInteger();
        int val = counter.incrementAndGet();
    }

    @Benchmark
    @BenchmarkMode(Mode.AverageTime)
    @OutputTimeUnit(TimeUnit.NANOSECONDS)
    public void benchmarkMutableInteger() {
        MutableInteger counter = new MutableInteger();
        int val = counter.getAndIncrement();
    }
```

Another common performance optimization is we can use `new AtomicInteger()` in lieu of `new AtomicInteger( 0 )`. **Jmh** proves that the performance difference is significant:

```
Benchmark                                 Mode  Cnt   Score    Error  Units
benchmarkAtomicIntegerDefaultConstructor  avgt    5  15.199 ± 13.034  ns/op
benchmarkAtomicIntegerExplicitConstructor avgt    5  35.029 ± 88.696  ns/op
```

```
    @Benchmark
    @BenchmarkMode(Mode.AverageTime)
    @OutputTimeUnit(TimeUnit.NANOSECONDS)
    public void benchmarkAtomicIntegerDefaultConstructor() {
        AtomicInteger counter = new AtomicInteger();
        int val = counter.incrementAndGet();
    }

    @Benchmark
    @BenchmarkMode(Mode.AverageTime)
    @OutputTimeUnit(TimeUnit.NANOSECONDS)
    public void benchmarkAtomicIntegerExplicitConstructor() {
        AtomicInteger counter = new AtomicInteger( 0 );
        int val = counter.incrementAndGet();
    }
```

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/util/MutableLong.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/util/MutableLong.java
@@ -6,33 +6,33 @@
  */
 package org.hibernate.internal.util;
 
-public class MutableInteger {
-	private int value;
+public class MutableLong {
+	private long value;
 
-	public MutableInteger() {
+	public MutableLong() {
 	}
 
-	public MutableInteger(int value) {
+	public MutableLong(long value) {
 		this.value = value;
 	}
 
-	public MutableInteger deepCopy() {
-		return new MutableInteger( value );
+	public MutableLong deepCopy() {
+		return new MutableLong( value );
 	}
 
-	public int getAndIncrement() {
+	public long getAndIncrement() {
 		return value++;
 	}
 
-	public int incrementAndGet() {
+	public long incrementAndGet() {
 		return ++value;
 	}
 
-	public int get() {
+	public long get() {
 		return value;
 	}
 
-	public void set(int value) {
+	public void set(long value) {
 		this.value = value;
 	}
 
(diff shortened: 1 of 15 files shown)


hibernate-core/src/main/java/org/hibernate/internal/util/MutableLong.java
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	public void set(long value) {
+		this.value = value;
+	}
+
+	public void increase() {
+		++value;
+	}
+}
```
NathanQingyangXu, line 42: I created this util class for 'long' case. Currently it is not used but it might be useful in the future.

dreab8:
Thanks @NathanQingyangXu , applied upstream.

sebersole:
:+1: Looked great Nathan

PR #3394, opened 2020-05-11
HHH-14006 Extended Bytecode enhancement fails for inherited public fields

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14006

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/FieldAccessEnhancer.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/FieldAccessEnhancer.java
@@ -11,6 +11,9 @@ import static net.bytebuddy.matcher.ElementMatchers.named;
 
 import javax.persistence.Id;
 
+import net.bytebuddy.description.NamedElement;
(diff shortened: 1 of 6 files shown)


barreiro:
@yrodiere  what about `javassist` bytecode enhancer ... have you ran the tests with `hibernate.bytecode.provider=javassist`?

yrodiere:
@yrodiere I did not. Isn't it deprecated and disabled by default?

I can see if there is a way to fix the problem there too, but if we want to drop support for javassist soon, I personally would vote against even trying.

yrodiere:
@barreiro I just ran the test, and it passes without any change with Javassist.

Turns out we use `getField`, not `getDeclaredField`, in the Javassist enhancer:

```
							|| !enhancementContext.isPersistentField( targetCtClass.getField( fieldName ) )
```

gsmet:
Yeah, I would say let's not even try if we don't have anyone explicitly asking for it.

barreiro:
@yrodiere thanks for checking. I usually try keep both implementations up to date so that we can compare how both behave when there are issues.

Sanne:
thank @yrodiere ! Merged it

PR #3395, opened 2020-05-12
wip/6.0 Get rid of session builder generics warnings

NathanQingyangXu:
We made extensive usage of class hierarchy builder pattern (as explained in Effective Java v3, Item 2, e.g.), but there are many misuses and unnecessary generics warnings which could be eliminated by some standard generics programming tricks, e.g.:

- recursive generics specification
- `getThis()` refactoring instead of returning `this` directly

Some 30 generics warnings are eliminated in this PR and our codebase becomes a little bit more elegant and safer.

Two typical pattern misuse was explained in two self-comments in this PR.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/AbstractDelegatingSessionBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/AbstractDelegatingSessionBuilder.java
@@ -25,17 +25,12 @@ import org.hibernate.resource.jdbc.spi.StatementInspector;
  * @author Gunnar Morling
  * @author Guillaume Smet
  */
-public abstract class AbstractDelegatingSessionBuilder<T extends SessionBuilder> implements SessionBuilder<T> {
+public abstract class AbstractDelegatingSessionBuilder<T extends AbstractDelegatingSessionBuilder<T>> implements SessionBuilder<T> {
 
 	private final SessionBuilder delegate;
 
 	public AbstractDelegatingSessionBuilder(SessionBuilder delegate) {
 		this.delegate = delegate;
-	}
-
-	@SuppressWarnings("unchecked")
-	protected T getThis() {
-		return (T) this;
 	}
 
 	protected SessionBuilder delegate() {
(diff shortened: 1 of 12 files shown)


NathanQingyangXu:
just realized there might be more improvement, so close the PR for the time being.

hibernate-core/src/main/java/org/hibernate/engine/spi/AbstractDelegatingSessionBuilder.java
```diff
@@ -25,19 +25,14 @@
  * @author Gunnar Morling
  * @author Guillaume Smet
  */
-public abstract class AbstractDelegatingSessionBuilder<T extends SessionBuilder> implements SessionBuilder<T> {
+public abstract class AbstractDelegatingSessionBuilder<T extends AbstractDelegatingSessionBuilder<T>> implements SessionBuilder<T> {
```
NathanQingyangXu, line 29: The generics type for builder hierarchy serves only one purpose, i.e. a placeholder for its subclass type per se. `SessionBuilder` chosen as the `extends` target is incorrect for it lacks the safety net against the possibility that a generics type not directly under `AbstractDelegatingSessionBuilder` could be used erroneously.
Similar code changes throughout the PR.

hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
```diff
@@ -1902,9 +1901,9 @@ public NClob createNClob(Reader reader, long length) {
 		}
 	}
 
-	private static class SharedSessionBuilderImpl<T extends SharedSessionBuilder>
-			extends SessionFactoryImpl.SessionBuilderImpl<T>
-			implements SharedSessionBuilder<T>, SharedSessionCreationOptions {
+	private static class SharedSessionBuilderImpl
+			extends SessionFactoryImpl.SessionBuilderImpl<SharedSessionBuilderImpl>
+			implements SharedSessionBuilder<SharedSessionBuilderImpl>, SharedSessionCreationOptions {
 		private final SessionImpl session;
```
NathanQingyangXu, line 1907: The above class is the leaf classes in the hierarchy (it is private class in this class) and there is no purpose to provide a generic type here (which is purely meant for subclassing). We are supposed to enjoy the generics at last by replacing `T` with the terminal class name per se, as seen in the below code changes. Time to harvest!

PR #3398, opened 2020-05-13
proposed change to SchemaManagementTool API for Hibernate Reactive

gavinking:
Hibernate Reactive needs a way to plug its own Vert.x-based database driver into the schema export tool.

Here's my not-very-deeply-thought-out proposal.

Hoping for feedback from @sebersole and @Sanne.


gavinking:
WTH?!

PR #3400, opened 2020-05-15
HHH-14020 - Allow Hibernate Types to have access to ServiceRegistry during initialization

vladmihalcea:
https://hibernate.atlassian.net/browse/HHH-14020

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/TypeFactory.java
@@ -17,6 +17,7 @@ import org.hibernate.classic.Lifecycle;
 import org.hibernate.engine.config.spi.ConfigurationService;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.internal.util.ReflectHelper;
+import org.hibernate.service.ServiceRegistry;
 import org.hibernate.tuple.component.ComponentMetamodel;
 import org.hibernate.type.spi.TypeBootstrapContext;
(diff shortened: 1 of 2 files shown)


vladmihalcea:
@sebersole @dreab8 If you have time, can you take a look at this optimization. It's a very useful addition to the Hibernate Types project.

NathanQingyangXu:
I think a more elegant way to get access to other stuff than configuration map is to add something into `TypeBootstrapContext` as following:
```
public interface TypeBootstrapContext {
	Map<String, Object> getConfigurationSettings();
        ServiceRegistry getServiceRegistry();
}
```
This way we don't need to make our magic overcomplicated and it is much easier to extend (simply adding new stuff into `TypeBootstrapContext`.

vladmihalcea:
@NathanQingyangXu That would not work for my use case. Because the `TypeBootstrapContext` class is only available in newer HIbernate 5.4 versions, I'll get a class loading error on Hibernate 5.2, 5.3, or previous 5.4 versions if the Hibernate Types project adds such a constructor to its Types.

That's why I proposed this change. In fact, that should have been the first option as the `TypeBootstrapContext` didn't help.

NathanQingyangXu:
Thanks for letting me know. TBH, the team members are busy with v6
developing and might not be too motivated by such hack to work around
backward compatibility. IMO, even if this PR is merged, we might change it
in v6 for it is a little bit messy. Actually, even the reflection usage
could be dropped totally by abstract parent class overloadded constructor.

NathanQingyangXu (approved):

gavinking:
superseded by #9313

PR #3401, opened 2020-05-15
Required for Hibernate Reactive for queries support

DavideD:
I haven't created the JIRAs yet, but could you let me know if these changes are viable first? Thanks

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/query/spi/QueryPlanCache.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/query/spi/QueryPlanCache.java
@@ -18,6 +18,7 @@ import java.util.concurrent.TimeUnit;
 import org.hibernate.Filter;
 import org.hibernate.MappingException;
 import org.hibernate.QueryException;
+import org.hibernate.Session;
 import org.hibernate.cfg.Environment;
 import org.hibernate.engine.query.spi.sql.NativeSQLQuerySpecification;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
@@ -152,7 +153,7 @@ public class QueryPlanCache implements Serializable {
 			final long startTime = ( stats ) ? System.nanoTime() : 0L;
 
 			LOG.tracev( "Unable to locate HQL query plan in cache; generating ({0})", queryString );
-			value = new HQLQueryPlan( queryString, shallow, enabledFilters, factory );
+			value = createHQLQueryPlan( queryString, shallow, enabledFilters, factory );
 
 			if ( stats ) {
 				final long endTime = System.nanoTime();
@@ -170,6 +171,10 @@ public class QueryPlanCache implements Serializable {
 			}
 		}
 		return value;
+	}
+
+	protected HQLQueryPlan createHQLQueryPlan(String queryString, boolean shallow, Map<String, Filter> enabledFilters, SessionFactoryImplementor factory) {
+		return new HQLQueryPlan( queryString, shallow, enabledFilters, factory );
 	}
 
 	/**
--- a/hibernate-core/src/main/java/org/hibernate/engine/query/spi/QueryPlanCache.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/query/spi/QueryPlanCache.java
@@ -18,7 +18,6 @@ import java.util.concurrent.TimeUnit;
 import org.hibernate.Filter;
 import org.hibernate.MappingException;
 import org.hibernate.QueryException;
-import org.hibernate.Session;
 import org.hibernate.cfg.Environment;
 import org.hibernate.engine.query.spi.sql.NativeSQLQuerySpecification;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
@@ -43,6 +42,11 @@ import org.hibernate.stat.spi.StatisticsImplementor;
 public class QueryPlanCache implements Serializable {
 	private static final CoreMessageLogger LOG = CoreLogging.messageLogger( QueryPlanCache.class );
 
+	@FunctionalInterface
+	public interface QueryPlanCreator {
+		HQLQueryPlan createQueryPlan(String queryString, boolean shallow, Map<String, Filter> enabledFilters, SessionFactoryImplementor factory);
+	}
+
 	/**
 	 * The default strong reference count.
 	 */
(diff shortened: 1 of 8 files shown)


gavinking:
Forks, can we expedite review of this patch, please? This is a blocking issue for further development work on Hibernate Reactive.

Thanks!

hibernate-core/src/main/java/org/hibernate/internal/SessionFactoryImpl.java
```diff
@@ -260,7 +265,7 @@ public SessionFactoryImpl(
 		LOG.debugf( "Session factory constructed with filter configurations : %s", filters );
 		LOG.debugf( "Instantiating session factory with properties: %s", properties );
 
-		this.queryPlanCache = new QueryPlanCache( this );
+		this.queryPlanCache = queryPlanCacheFunction.apply( this );
```
gavinking, line 268: @DavideD I'm not a big fan of this. I think it might be better to add a `getQueryPlanCache()` accessor to `SessionFactoryImpl` and override it on the subclass.

DavideD:
What's wrong with this?

hibernate-core/src/main/java/org/hibernate/internal/SessionFactoryImpl.java
gavinking: (You're passing a reference to a partially-uninitialized object to an arbitrary function. The subclass constructor hasn't been called at this point.)

gavinking: And note: you're using a function ref here precisely because you want to circumvent the JVM's restriction which is designed to prevent you doing precisely what you're doing here :-)

DavideD:
I've added two commits for HQL update reactive queries

hibernate-core/src/main/java/org/hibernate/internal/SessionFactoryImpl.java
DavideD: You got me :) OK, I will change it

hibernate-core/src/main/java/org/hibernate/engine/query/spi/QueryPlanCache.java
```diff
@@ -152,7 +153,7 @@ public HQLQueryPlan getHQLQueryPlan(String queryString, boolean shallow, Map<Str
 			final long startTime = ( stats ) ? System.nanoTime() : 0L;
 
 			LOG.tracev( "Unable to locate HQL query plan in cache; generating ({0})", queryString );
-			value = new HQLQueryPlan( queryString, shallow, enabledFilters, factory );
+			value = createHQLQueryPlan( queryString, shallow, enabledFilters, factory );
```
sebersole, line 156: Are you guys really using inheritance everywhere?  You *really* want to subclass all of these?  Why not just pass in functionals that can create the plan?  Then your SessionFactory (I assume you already subclass SessionFactory, Session, etc) can simply pass in a different one.

I mean, feel free to keep subclassing everything, but personally I'd be trying to minimize what I need to subclass

gavinking: @sebersole the new subclass `RxHQLQueryPlan` actually adds new public operations, it's not just overriding operations of `HQLQueryPlan`.

On the other hand, perhaps you're right about `RxQueryPlanCache`: I now notice we could probably eliminate that class and replace it by a new argument to `QueryPlanCache`. @DavideD do you want me to try and make that change?

DavideD: I think I can take care of it. Let me give it a go.

gavinking: I've almost got it done.

gavinking: @DavideD can I push to your branch?

DavideD: You are talking about adding a function for the creation of the QueryPlanCache, isn't it? Can I add a new `getHQLQueryPlan` with this additional argument? Or do  need to keep the API of `QueryPlanCache` the same?

DavideD: OK, push it on the branch. No problem

gavinking: @DavideD do I have permission?

    Authentication failed for 'https://github.com/DavideD/hibernate-orm.git/'

gavinking: Anyway the commit is c0ca1560e8bf385e0c3b9431b118780a8e0953e0

gavinking: And the corresponding change to your patch to hibernate-reactive is https://github.com/gavinking/hibernate-rx/commit/2f3d48ff4514225219d2855b3c988e778e11e327

DavideD:
I've squashed the commit in the [hibernate-reactive PR](https://github.com/hibernate/hibernate-reactive/pull/119) and add the  ORM commit to this one.

hibernate-core/src/main/java/org/hibernate/engine/query/spi/QueryPlanCache.java
DavideD: I've updated both PRs. I've squashed the commits in the Hibernate Reactive one

DavideD:
Thanks a lot! I'm going to create issues for this PR if there are not additional changes required

Sanne:
Thanks @DavideD . BTW I hadn't seen this - please ping me on zulip if you need a review, or just push it upstream as soon as it's good in your opinion. But also ping me when ORM is ready for release :)

DavideD:
Thanks a lot @sebersole, I noticed a compilation error and added this https://github.com/hibernate/hibernate-orm/commit/88a712bab771698c93feaa62fa91a9f7a1be576e

I hope it's ok

gavinking:
Excellent, thanks!

sebersole:
On Mon, May 18, 2020 at 6:57 AM DavideD <notifications@github.com> wrote:

> Thanks a lot @sebersole <https://github.com/sebersole>, I noticed a
> compilation error and added this 88a712b
> <https://github.com/hibernate/hibernate-orm/commit/88a712bab771698c93feaa62fa91a9f7a1be576e>
>

The linked Travis job finished successfully :(

>

PR #3402, opened 2020-05-15
wip/6.0 Fix some logging performance issues

NathanQingyangXu:
This is a labour intensively PR and feel free to have it merged as late as possible lest it breaks many existing PRs.

The following issues are taken care of in this PR as much as possible:

- avoid string concatenation (replace it with format parameter) 
- add level checking if the format parameter involves nontrivial computing (e.g. method invoking)
- keep from using format logging method if the content doesn't involve format placeholder
- misc bugs correction (e.g. missing format parameter, etc.)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/archive/internal/ArchiveHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/archive/internal/ArchiveHelper.java
@@ -93,7 +93,7 @@ public class ArchiveHelper {
 					"Unable to determine JAR Url from " + url + ". Cause: " + e.getMessage()
 			);
 		}
-		log.trace( "JAR URL from URL Entry: " + url + " >> " + jarUrl );
+		log.tracef( "JAR URL from URL Entry: %s >> %s", url, jarUrl );
 		return jarUrl;
 	}
 
--- a/hibernate-core/src/main/java/org/hibernate/boot/archive/internal/StandardArchiveDescriptorFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/archive/internal/StandardArchiveDescriptorFactory.java
@@ -130,12 +130,14 @@ public class StandardArchiveDescriptorFactory implements ArchiveDescriptorFactor
 			}
 			catch (MalformedURLException e) {
 				// allow to pass through to return the original URL
-				log.debugf(
-						e,
-						"Unable to adjust relative <jar-file/> URL [%s] relative to root URL [%s]",
-						filePart,
-						rootUrlFile.getAbsolutePath()
-				);
+				if ( log.isDebugEnabled() ) {
+					log.debugf(
+							e,
+							"Unable to adjust relative <jar-file/> URL [%s] relative to root URL [%s]",
+							filePart,
+							rootUrlFile.getAbsolutePath()
+					);
+				}
 			}
 
 			return url;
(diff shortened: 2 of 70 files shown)


hibernate-core/src/main/java/org/hibernate/boot/archive/internal/ArchiveHelper.java
```diff
@@ -93,7 +93,7 @@ else if ( "zip".equals( protocol )
 					"Unable to determine JAR Url from " + url + ". Cause: " + e.getMessage()
 			);
 		}
-		log.trace( "JAR URL from URL Entry: " + url + " >> " + jarUrl );
```
NathanQingyangXu, line 96: The above string concatenation will incur performance cost invariably even though trace logging level is disabled

hibernate-core/src/main/java/org/hibernate/boot/archive/internal/StandardArchiveDescriptorFactory.java
```diff
@@ -130,12 +130,14 @@ public URL adjustJarFileEntryUrl(URL url, URL rootUrl) {
...
-						"Unable to adjust relative <jar-file/> URL [%s] relative to root URL [%s]",
-						filePart,
-						rootUrlFile.getAbsolutePath()
-				);
+				if ( log.isDebugEnabled() ) {
+					log.debugf(
+							e,
+							"Unable to adjust relative <jar-file/> URL [%s] relative to root URL [%s]",
+							filePart,
+							rootUrlFile.getAbsolutePath()
```
NathanQingyangXu, line 138: I added logging level checking whenever the method invoking is involved. We never know how much performance cost will be paid for future coding might provide an overridden heavy weight method implementation. Adding logging level checking is simple and future proof.

hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
```diff
@@ -4221,7 +4227,7 @@ public void addAttributeBinding(Property attributeBinding) {
 
 		@Override
 		public void process() {
-			log.debugf( "Binding natural-id UniqueKey for entity : " + entityBinding.getEntityName() );
```
NathanQingyangXu, line 4224: The above pattern has two drawbacks:

- unnecessary string concatenation cost even the produced string is simply discarded later due to level disabling
- the produced string might contain magic chars (i.e. `%s`) and will explode during runtime

hibernate-core/src/main/java/org/hibernate/resource/beans/container/internal/JpaCompliantLifecycleStrategy.java
```diff
@@ -125,7 +125,7 @@ public void initialize() {
 				throw e;
 			}
 			catch (Exception e) {
-				log.debugf( "Error resolving CDI bean [%s] - using fallback" );
+				log.debugf( "Error resolving CDI bean [%s] - using fallback", beanType.getName() );
```
NathanQingyangXu, line 128: It is still a common bug to miss format parameter. I tried the Checker Framework (https://checkerframework.org/) but it crashed when I used it on hibernate-orm, :(.

hibernate-core/src/main/java/org/hibernate/boot/archive/internal/StandardArchiveDescriptorFactory.java
sebersole: Yes, it *is* future proof.  For whatever reason, and I may be in the minority which is fine, I prefer to not do the level check in trivial cases.  I don't think it is as easy as saying "well anything that calls a method.  E.g. here calling `#getAbsolutePath()` does have an overhead, so this is fine.  The reason I prefer to avoid the check is ofc because the logging system does the same checks already.  I'm just talking in the general case as a reply to your comment.  Again, here it is fine

NathanQingyangXu: Yeah, this is not clear-cut decision and it is hard to decide which scenario is trivial. Yeah, method invoking might not be a good criterion.

PR #3407, opened 2020-05-18
HHH-13997  Add methods to SessionImpl and StatefulPersistenceContext; make Loader method protected 

DavideD:
Follows: https://github.com/hibernate/hibernate-orm/pull/3385

Backported to 5.4


gavinking:
@DavideD I preferred the previous version of this patch.

Again it works by passing a not-fully-initialized instance of this `this` to an arbitrary function from within the superclass constructor.

And the thing is, we're already extending this class, so why not just override two more of its methods, instead of doing the same thing (overriding) by passing around function refs. To me that seems a much more straightforward design in the case that we're already have to extend the class for other reasons.

DavideD:
I'm going to revert the last commit so that we can release. And then later we can improve it. Does that work for you @gavinking ?

gavinking:
Yup, perfect.

DavideD:
Merged with the additional commit to revert what we discussed

PR #3410, opened 2020-05-19
HHH-14032 Parse Locales with scripts properly in LocaleTypeDescriptor.fromString

kilink:
Locales with scripts (e.g. zh-Hant-HK) were not parsed properly by LocaleTypeDescriptor,
causing the script part to get lost.


hibernate-core/src/test/java/org/hibernate/test/type/descriptor/java/LocaleTypeDescriptorTest.java
```diff
@@ -25,18 +25,19 @@
...
-		assertEquals( toLocale( "", "", "" ), LocaleTypeDescriptor.INSTANCE.fromString( "" ) );
+		assertEquals( toLocale( "de", null, null, null ), LocaleTypeDescriptor.INSTANCE.fromString( "de" ) );
+		assertEquals( toLocale( "de", "DE", null, null ), LocaleTypeDescriptor.INSTANCE.fromString( "de_DE" ) );
+		assertEquals( toLocale( null, "DE", null, null ), LocaleTypeDescriptor.INSTANCE.fromString( "_DE" ) );
+		assertEquals( toLocale( null, null, "ch123", null ), LocaleTypeDescriptor.INSTANCE.fromString( "__ch123" ) );
+		assertEquals( toLocale( null, "DE", "ch123", null ), LocaleTypeDescriptor.INSTANCE.fromString( "_DE_ch123" ) );
+		assertEquals( toLocale( "de", null, "ch123", null ), LocaleTypeDescriptor.INSTANCE.fromString( "de__ch123" ) );
+		assertEquals( toLocale( "de", "DE", "ch123", null ), LocaleTypeDescriptor.INSTANCE.fromString( "de_DE_ch123" ) );
+		assertEquals( toLocale( "zh", "HK", null, "Hant"), LocaleTypeDescriptor.INSTANCE.fromString( "zh_HK_#Hant" ) );
+		assertEquals( toLocale( "", "", "", null ), LocaleTypeDescriptor.INSTANCE.fromString( "" ) );
```
NathanQingyangXu, line 36: I think the above snippet misuses `assertEquals()` (as in the legacy code). Its first parameter is the expected value, not the second one.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
```diff
@@ -51,43 +51,33 @@ public Locale fromString(String string) {
...
-					position = i + 1;
-				}
-				else {
-					// On the second separator we have to check whether there are more chars available for variant
-					if ( chars.length > i + 1 ) {
-						// There is a variant so add it to the constructor
-						return new Locale( string, new String( chars, position, i - position ), new String( chars,
-								i + 1, chars.length - i - 1 ) );
+		Locale.Builder builder = new Locale.Builder();
+		String[] parts = string.split("_");
```
NathanQingyangXu, line 55: From performance perspective, we should avoid using `String.split()` for it will create RE and might involve unnecessary performance cost. There should be better way to replace the above `split()` usage.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
```diff
@@ -51,43 +51,33 @@ public Locale fromString(String string) {
...
+					if (i < parts.length - 1 || !s.startsWith("#")) {
+						builder.setVariant(s);
+						break;
 					}
-					else {
-						// No variant given, we just have language and country
-						return new Locale( string, new String( chars, position, i - position ), "" );
+				case 3:
+					if (s.startsWith("#")) {
+						s = s.substring(1);
```
NathanQingyangXu, line 73: It would be ideal to add `break` for `case 3` for best practice. Imagine in the future a new case is needed and it might create a subtle bug that the coder forgets to insert `break` for `case 3`.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
NathanQingyangXu: Maybe you can use the `StringHelper#split` for better performance.

hibernate-core/src/test/java/org/hibernate/test/type/descriptor/java/LocaleTypeDescriptorTest.java
```diff
@@ -47,6 +48,9 @@ public Locale toLocale(String lang, String region, String variant) {
 		if ( StringHelper.isNotEmpty( variant ) ) {
 			builder.setVariant( variant );
 		}
+		if ( StringHelper.isNotEmpty(script) ) {
+			builder.setScript(script);
```
NathanQingyangXu, line 52: it would be great to add space chars around `script` to adhere to existing coding format.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
NathanQingyangXu: The gist of this PR seems to lie in the above `case 2` and `case 3`. There is subtlety here that if s starts with `#` in `case 2`, it is equivalent to `case 3`. It would be great to add some comment like `// intentionally fall through` before `case 3` to help future code maintainer to understand the subtlety here.

NathanQingyangXu: TBH, there must be still room for code readability improvement by clearer logic. The break case falling through is to difficult to understand.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
kilink: I had originally used [StringTokenizer](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/util/StringTokenizer.html), which should have better performance, but its Javadoc says the following:

> StringTokenizer is a legacy class that is retained for compatibility reasons although its use is discouraged in new code. It is recommended that anyone seeking this functionality use the split method of String or the java.util.regex package instead.

That said, even in the latest JDK the class is technically not deprecated, not if the above advice should be heeded.

Looking at the suggested ```StringHelper#split``` method, it actually does use StringTokenizer internally. Depending on how careful we want to be about avoiding extra allocations, we can use StringTokenizer directly and avoid the array allocation. I don't have a strong opinion.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
kilink: I agree that the above code is less than ideal. Adding a break to case 2 would mean duplicating the code from case 3, I don't know which one is the lesser of evils. We could perhaps just avoid the switch altogether and unroll it into 4 if statements.

It could perhaps be rewritten to mimic this pattern as seen in [LanguageTag](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/sun/util/locale/LanguageTag.java#L202), which would require using something like StringTokenizer. It is unfortunate that this code cannot simply use [toLanguageTag](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/util/Locale.html#toLanguageTag()) / [forLanguageTag](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/util/Locale.html#forLanguageTag(java.lang.String))...

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
NathanQingyangXu: Thanks for knowledge sharing. I have no idea that `StringTokenizer` is not recommended. I took a look of the internal implementation of `String#split()` and found it has a fastpath for simple delimiter, so I think your current approach is perfectly fine.
Screen Shot 2020-06-04 at 4 42 41 PM

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
NathanQingyangXu: IMHO, using StringTokenizer is perfectly fine in Hibernate core codebase (it has been used more than 50 after a quick search). Code readability is a much bigger concern. The current switch case falling through is a clever approach but might make future code maintaining much difficult.

hibernate-core/src/test/java/org/hibernate/test/type/descriptor/java/LocaleTypeDescriptorTest.java
```diff
@@ -25,18 +25,19 @@
...
-		assertEquals( toLocale( "de", "DE", "ch123" ), LocaleTypeDescriptor.INSTANCE.fromString( "de_DE_ch123" ) );
-		assertEquals( toLocale( "", "", "" ), LocaleTypeDescriptor.INSTANCE.fromString( "" ) );
+		assertEquals( toLocale( "de", null, null, null ), LocaleTypeDescriptor.INSTANCE.fromString( "de" ) );
+		assertEquals( toLocale( "de", "DE", null, null ), LocaleTypeDescriptor.INSTANCE.fromString( "de_DE" ) );
+		assertEquals( toLocale( null, "DE", null, null ), LocaleTypeDescriptor.INSTANCE.fromString( "_DE" ) );
+		assertEquals( toLocale( null, null, "ch123", null ), LocaleTypeDescriptor.INSTANCE.fromString( "__ch123" ) );
+		assertEquals( toLocale( null, "DE", "ch123", null ), LocaleTypeDescriptor.INSTANCE.fromString( "_DE_ch123" ) );
+		assertEquals( toLocale( "de", null, "ch123", null ), LocaleTypeDescriptor.INSTANCE.fromString( "de__ch123" ) );
+		assertEquals( toLocale( "de", "DE", "ch123", null ), LocaleTypeDescriptor.INSTANCE.fromString( "de_DE_ch123" ) );
+		assertEquals( toLocale( "zh", "HK", null, "Hant"), LocaleTypeDescriptor.INSTANCE.fromString( "zh_HK_#Hant" ) );
```
NathanQingyangXu, line 35: It seems the above testing case is the sole motivation of this PR. Could we avoid touching existing code logic by a simple patch as following:

- test whether the string ends with '#xxx'
- if so, extract the `xxx` as `script` and drop the suffix from the string;
- pass the string using original logic.

Seems a better alternative than the current tricky switch implementation. Also, it is easy to understand and verify, so is more ideal from code maintaining perspective.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
```diff
@@ -51,43 +51,33 @@ public Locale fromString(String string) {
...
+				case 0:
+					builder.setLanguage(s);
+					break;
+				case 1:
+					builder.setRegion(s);
+					break;
+				case 2:
+					if (i < parts.length - 1 || !s.startsWith("#")) {
+						builder.setVariant(s);
+						break;
```
NathanQingyangXu, line 69: It would be great to adhere to hibernate code format (see https://github.com/hibernate/hibernate-ide-codestyles). Also, existing code convention uses `final` keyword as much as possible.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
Sanne: The `split` method accepts a regex which gets compiled at runtime, it would be best to declare the regex as a static final constant in the class, and use that to split strings.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
```diff
@@ -51,43 +51,33 @@ public Locale fromString(String string) {
...
 					}
-					else {
-						// No variant given, we just have language and country
-						return new Locale( string, new String( chars, position, i - position ), "" );
+				case 3:
+					if (s.startsWith("#")) {
```
Sanne, line 72: When needing to compare with characters, I'd prefer to use the char API:
    string.charAt(0) == '#'

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/LocaleTypeDescriptor.java
kilink: @Sanne as mentioned in the above comment, the split method has a fast path when splitting on a single character that avoids compiling a regex.

hibernate-core/src/test/java/org/hibernate/test/type/descriptor/java/LocaleTypeDescriptorTest.java
kilink: The existing code needs to be rewritten regardless, as the only way to set script or extensions is by switching to LocaleBuilder.

Also, looking at the Javadoc / implementation for [Locale.toString()](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/util/Locale.java#L1360), we can see the pattern is different based on whether the Locale has a variant or not (e.g., with a variant, script and extension are separated from the rest with "_#" vs just the "#" when no variant is present). Therefore I think the approach of testing if the string ends with "#" would need a special case for "_" anyway. I'll explore the more generic approach with just using a StringTokenizer or the like, if the main concern is the switch statement here?

kilink:
Sorry for the delay with this. I've gone with a different approach here that does not use a loop or switch statement at all. One caveat is that I decided not to attempt to parse extensions at all, as that is its own little rabbit hole; this PR strictly adds support for parsing scripts. I am not sure how prevalent bcp47 extensions are in practice, but I am assuming that script is more widely used and important. Extensions could be tackled in a follow up, as the way the code is rewritten would allow it to be extended fairly trivially.

gavinking:
I don't like the approach taken here. I think what we should be doing here is offering a whole new way of storing `Locale`s based on the language tag. Unfortunately, it doesn't look like we can do that in a completely backward compatible way, as I commented on the issue. So we'll need some way for the user to opt in to it. That's unfortunate.

Let's discuss it further on the issue.

gavinking:
i.e. we should really just be using `toLanguageTag()` and `forLanguageTag()` instead of parsing locales in our own code.

beikov:
Thanks for the work, I took it a bit further. Superseded by https://github.com/hibernate/hibernate-orm/pull/11052

PR #3412, opened 2020-05-19
[WIP] HHH-14180 Update h2 database from 1.4.197 to 2.0.x

famod:
~~https://hibernate.atlassian.net/browse/HHH-14031~~
Rescoped as agreed here: https://github.com/hibernate/hibernate-orm/pull/3412#issuecomment-679156850
Ticket is now: https://hibernate.atlassian.net/browse/HHH-14180

⚠️ This is work in progess as it will need many adjustments and discussions. This PR will grow gradually.

cc @tkalmar

### DIFF
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -10,7 +10,7 @@
 ext {
 
     junitVersion = '4.12'
-    h2Version = '1.4.196'
+    h2Version = '1.4.197'
     bytemanVersion = '4.0.13' //Compatible with JDK16
     jnpVersion = '5.0.6.CR1'
 
--- a/hibernate-core/src/test/java/org/hibernate/test/temporal/TimePropertyTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/temporal/TimePropertyTest.java
@@ -10,7 +10,6 @@ import java.sql.Time;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
-import java.util.Date;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
 import javax.persistence.Table;
@@ -41,6 +40,9 @@ public class TimePropertyTest extends BaseCoreFunctionalTestCase {
 		// See javadoc for java.sql.Time: 'The date components should be set to the "zero epoch" value of January 1, 1970 and should not be accessed'
 		// Other dates can potentially lead to errors in JDBC drivers, in particular MySQL ConnectorJ 8.x.
 		calendar.set( 1970, Calendar.JANUARY, 1 );
+		// H2Dialect uses TIME (without fractional seconds precision) so H2 would round half up if milliseconds were set
+		// See also: http://h2database.com/html/datatypes.html#time_type
+		calendar.set( Calendar.MILLISECOND, 0 );
 		eOrig.tAsDate = new Time( calendar.getTimeInMillis() );
 
 		Session s = openSession();
--- a/hibernate-core/src/test/java/org/hibernate/test/type/AbstractJavaTimeTypeTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/type/AbstractJavaTimeTypeTest.java
@@ -20,16 +20,10 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.function.Consumer;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
 import org.hibernate.cfg.AvailableSettings;
 import org.hibernate.cfg.Configuration;
 import org.hibernate.dialect.Dialect;
 import org.hibernate.dialect.H2Dialect;
-import org.hibernate.dialect.MariaDB10Dialect;
-import org.hibernate.dialect.Oracle8iDialect;
-import org.hibernate.dialect.PostgreSQL81Dialect;
 import org.hibernate.type.descriptor.sql.SqlTypeDescriptor;
 
 import org.hibernate.testing.TestForIssue;
@@ -304,8 +298,9 @@ abstract class AbstractJavaTimeTypeTest<T, E> extends BaseCoreFunctionalTestCase
 		}
 
 		protected final boolean isNanosecondPrecisionSupported() {
-			// Most databases apparently don't support nanosecond precision correctly
-			return dialect instanceof H2Dialect;
+			// This used to return true for H2Dialect, but as of 1.4.197 h2 does not use ns precision by default anymore.
+			// Bringing back ns precision would require timestamp(9) in the dialect class.
+			return false;
 		}
 
 		protected final S add(ZoneId defaultJvmTimeZone, Object ... subClassParameters) {
--- a/hibernate-testing/hibernate-testing.gradle
+++ b/hibernate-testing/hibernate-testing.gradle
@@ -22,6 +22,7 @@ dependencies {
     compile ( libraries.jboss_jta ) {
         transitive=false;
     }
+    compile( libraries.h2 )
 }
 
 // resources inherently exclude sources
--- a/hibernate-testing/src/main/java/org/hibernate/testing/junit4/BaseCoreFunctionalTestCase.java
+++ b/hibernate-testing/src/main/java/org/hibernate/testing/junit4/BaseCoreFunctionalTestCase.java
@@ -17,6 +17,7 @@ import java.util.Properties;
 import java.util.function.Consumer;
 import javax.persistence.SharedCacheMode;
 
+import org.h2.util.DateTimeUtils;
 import org.hibernate.HibernateException;
 import org.hibernate.Interceptor;
 import org.hibernate.Session;
@@ -350,6 +351,11 @@ public abstract class BaseCoreFunctionalTestCase extends BaseUnitTestCase {
 
 	@After
 	public final void afterTest() throws Exception {
+		// see https://github.com/hibernate/hibernate-orm/pull/3412#issuecomment-678338398
+		if ( getDialect() instanceof H2Dialect ) {
+			DateTimeUtils.resetCalendar();
+		}
+
 		completeStrayTransaction();
 
 		if ( isCleanupTestDataRequired() ) {
@@ -425,6 +431,8 @@ public abstract class BaseCoreFunctionalTestCase extends BaseUnitTestCase {
 	}
 
 	public static class RollbackWork implements Work {
+
+		@Override
 		public void execute(Connection connection) throws SQLException {
 			connection.rollback();
 		}
--- a/hibernate-testing/src/main/java/org/hibernate/testing/junit4/BaseNonConfigCoreFunctionalTestCase.java
+++ b/hibernate-testing/src/main/java/org/hibernate/testing/junit4/BaseNonConfigCoreFunctionalTestCase.java
@@ -18,6 +18,7 @@ import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
+import org.h2.util.DateTimeUtils;
 import org.hibernate.HibernateException;
 import org.hibernate.Interceptor;
 import org.hibernate.Session;
@@ -450,6 +451,11 @@ public class BaseNonConfigCoreFunctionalTestCase extends BaseUnitTestCase {
 
 	@After
 	public final void afterTest() throws Exception {
+		// see https://github.com/hibernate/hibernate-orm/pull/3412#issuecomment-678338398
+		if ( getDialect() instanceof H2Dialect ) {
+			DateTimeUtils.resetCalendar();
+		}
+
 		completeStrayTransaction();
 
 		if ( isCleanupTestDataRequired() ) {
@@ -505,6 +511,8 @@ public class BaseNonConfigCoreFunctionalTestCase extends BaseUnitTestCase {
 	}
 
 	public class RollbackWork implements Work {
+
+		@Override
 		public void execute(Connection connection) throws SQLException {
 			connection.rollback();
 		}


famod:
Ok, Travis is started even though this is a draft PR, which is good for my requirements.

This first commit should yield `10105 tests completed, 487 failed, 1325 skipped` in `hibernate-core`.

cc @Sanne

famod:
Interesting, Travis is reporting way more failures:

> 10105 tests completed, 912 failed, 1324 skipped

@Sanne Any idea why that can be? Is there any way to get the test reports from Travis?

Sanne:
good question.. no I'm not too familiar with Travis myself, we started using it only recently as an addition to ci.hibernate.org .

I was able to find the raw logs:
 - https://api.travis-ci.org/v3/job/688754932/log.txt

It's not very practical though.. @dreab8 @yrodiere is there a way to have it produce/publish the better reports that gradle produces?

dreab8:
Unfortunately I'm not a Travis expert, but I saw it is possible to access to the job log https://travis-ci.org/github/hibernate/hibernate-orm/jobs/688754932, so probably a way is to run gradle with the INFO logging level on (`./gradlew test -i` ), the downside is the huge amount of output printed to the console.

yrodiere:
@Sanne I've never done it, but from what I remember Travis recommends pushing data to AWS S3 buckets. It seems there are many storage options available nowadays: https://docs.travis-ci.com/user/deployment/
I doubt there is a "built-in" storage solution in Travis itself.

yrodiere:
> Unfortunately I'm not a Travis expert, but I saw it is possible to access to the job log https://travis-ci.org/github/hibernate/hibernate-orm/jobs/688754932, so probably a way is to run gradle with the INFO logging level on (./gradlew test -i ), the downside is the huge amount of output printed to the console.

From what I remember we had to reduce the verbosity of logs in the past. If the logs exceeds a certain size, Travis will flat out refuse to keep it and you won't see anything.

famod:
Thanks for your feedback. Unfortunately the stdout is not sufficient to see what is going on.
You only see:
```
org.hibernate.test.type.OffsetDateTimeTest > nativeWriteThenRead[2,017-11-6T19:19:1.0[+10:00] [JVM TZ: UTC-08:00, JDBC TZ: null, remapping dialect: null]] FAILED

    java.lang.AssertionError at Assert.java:88
```
but not the output of the test itself.

Can anyone of you build this branch locally? That would help to see whether it is a problem on my machine or whether there is something strange on Travis.

dreab8:
started a build on my machine

dreab8:
with java 8
```
java version "1.8.0_161"
Java(TM) SE Runtime Environment (build 1.8.0_161-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.161-b12, mixed mode)
```
locally I had `10105 tests completed, 708 failed, 1324 skipped`

famod:
@dreab8 Thanks for checking!

So we now have three different results. WTH?

I honestly don't want to install any of these ancient JDK8 version... (which might not even change anything).

dreab8:
@famod FYI I rebased your branch on latest version of master and then I run the tests

yrodiere:
For what it's worth, those tests are about timezones, and thus may be sensitive to the timezone of your machine. They take care of reseting the JVM locale when the test starts, but if H2 has internal configuration related to the locale and intializes that configuration before the test starts (for example in a static field), then the tests will lead to different results in different machines.

I do not know how to make the test suite more "stable" without starting a new JVM for each test parameter, which would obviously take ages.

That being said, the tests used to pass, so clearly something changed in H2... maybe there's some new "optimization" related to timezones in H2 that you can disable?

famod:
The timezone problems are more or less expected (although I still have to find the exact reason for that change in 1.4.197) but I didn't expect three different results. Btw, my branch should pretty much up to date, but I will (re-)check that.

I've kicked off another run locally with `-Duser.timezone=UTC` in `gradle.properties` (but IIRC I tried that already on the weekend).

famod:
I rebased on `master` but the only change is my own commit regarding `-Duser.language=en` in `gradle.properties`.

@dreab8 @yrodiere @Sanne 
I'd like to try having the travis build run with a recent AdoptOpenJDK8 version. Any objections?

famod:
> I'd like to try having the travis build run with a recent AdoptOpenJDK8 version.

I just tried that and it doesn't make a difference. 😕

famod:
I'll try to ignore the test failure differences for now. I am still testing, but it seems it is flip-flopping between 486 (sometimes 487?) and > 900 locally, which is a bit worrying as far as reproducible builds are concerned.

famod:
The first "real" fix is in (fixing ~50 test failures) and it should be discussed: https://github.com/hibernate/hibernate-orm/pull/3412/commits/a02854daa12a8807ace0a55cf8536259e738d0c8

The problem here is that multiple changes in 1.4.197 to TIMESTAMP precision and rounding cause the tests to fail that insert times like 23:59:59.999999999 (which menas nanosecond precision) and expect to get the same back.
This doesn't work anymore, _unless_ `Types.TIMESTAMP` is mapped to `timestamp(9)`, which I added to `H2Dialect`.
The _alternative_ to that would to adjust the tests which would basically mean that `AbstractJavaTimeTypeTest.AbstractParametersBuilder.isNanosecondPrecisionSupported()` has to be removed and in turn _no_ tests with nanosecond precision would be conducted:
```java
protected final boolean isNanosecondPrecisionSupported() {
	// Most databases apparently don't support nanosecond precision correctly
	return dialect instanceof H2Dialect;
}
```

@Sanne  (and maybe @dreab8?): WDYT?

famod:
I finally found out why I have only about half the failures locallly: The system properties from `grade.properties` are _not_ passed to the test execution.
I added the following to `java-module.gradle` (and to `gradle.properties` for consistency):
```groovy
test {
       systemProperty 'user.language', 'en'
       systemProperty 'user.country', 'US'
       systemProperty 'user.timezone', 'UTC'
       systemProperty 'file.encoding', 'UTF-8'
}
```

Sanne:
hi @famod, for convenience I've extracted your last commit and am including it already, as:
 - https://hibernate.atlassian.net/browse/HHH-14144

famod:
@Sanne That's a good idea! I'll rebase when I find some time.

WDYT about https://github.com/hibernate/hibernate-orm/pull/3412#issuecomment-633735805?

Sanne:
> @Sanne That's a good idea! I'll rebase when I find some time.
> 
> WDYT about [#3412 (comment)](https://github.com/hibernate/hibernate-orm/pull/3412#issuecomment-633735805)?

sorry I missed that. I think it would be fine to change `isNanosecondPrecisionSupported()` to "return false" and expand the comment with your findings, since it reflects reality.
I'd prefer keeping the method, in hope we could figure out that some dialects in fact support it .. but not strong feelings, maybe we should just clean it up.

famod:
PR rebased. I left the nanosecond test cases for now. Might clean them up once the rest is sorted out.
There is still a larger timezone related problem for which I don't really know the reason, yet.

Sanne:
@fax4ever could you help with this one, and see if we can finally upgrade?

fax4ever:
Hi @Sanne sure! I'm running the tests...

fax4ever:
Hi @Sanne,

I did some tests, you can find them [here](https://github.com/fax4ever/hibernate-orm/commit/7eb7abe723ae3f338d3180bd8d186b4c3171c84d) if are interested in.

Before continuing, I need to know more about what is the expected behavior here and about that I have two questions, one for us (the Hibernate team) and for @famod.

The former is:
From my understanding our tests in `AbstractJavaTimeTypeTest` update the default java time zone at writing time, meaning that when we created the schema we used a different default java to the one used to update the table.
If it is the case, I don’t know if it makes sense to test this. In my opinion, it is something that shouldn’t be supported. 
I’m saying that because the DB engine could put the timezone in the Table definition, but you know that I know a bit more about NoSQL than SQL stuff.

The latter question is for @famod and it is:
I wrote [this test](https://github.com/fax4ever/hibernate-orm/blob/HHH-14031-tests/hibernate-core/src/test/java/org/hibernate/test/type/tmp/JvmTimeZoneJdbcConsistentTest.java) to update the default time zone consistently (use the same at schema definition / writing / querying time).
Actually this doesn’t work with the current Hibernate H2 version too.
Is it supposed to be working?
In the test I tried to use the H2 JDBC Driver directly, If I did something wrong with it, I would be happy to fix it.

gavinking:
If you don't mind, can I just ask why we're bothering about time zone stuff on Hibernate 5 when we know that we just don't have proper support for time zones on this branch? (I implanted it for Hibernate 6, and so it should be tested there, it seems to me.)

Sanne:
+1 to do this just in the 6.x branch. Unless I'm missing an important reason?

famod:
So the plan is to remove those tests from `master` (and 5.x) and add/movee them to the wip/6.0 branch?
But the h2 update shall still land in `master`?
Is that correct or am I missing something?

fax4ever:
+ 1 to @gavinking  and @Sanne . Tests are supposed to be scoped on the expected behavior and do not test something that is not supposed to be supported.  I would move these tests to the Hibernate ORM 6 branch if the team is agree with that.

yrodiere:
For what it's worth, I wrote these tests while working on bug reports. Multiple bug reports, over the course of several months.

While I agree the support we have in ORM 5 is not great, it's still used by some people, it should still continue to work (if we can call the current behavior "working"), and we should avoid regressions.

Regarding this:
> From my understanding our tests in AbstractJavaTimeTypeTest update the default java time zone at writing time, meaning that when we created the schema we used a different default java to the one used to update the table.
> If it is the case, I don’t know if it makes sense to test this. In my opinion, it is something that shouldn’t be supported.
> I’m saying that because the DB engine could put the timezone in the Table definition, but you know that I know a bit more about NoSQL than SQL stuff.

The default JVM timezone definitely has an impact on our date/time code and also on the JDBC driver's date/time code, unfortunately. That's because most `java.util` date/time APIs rely on the JVM's default timezone. That's admittedly a terrible design decision, but that's how it is, and we're stuck with these APIs until ORM 6.

It's not a good idea *at all* to ignore the JVM timezone and to just test with whatever timezone is used on the computer running the tests. Most of the bugs only occur with a specific timezone.

The purpose of this change to the JVM timezone in `AbstractJavaTimeTypeTest` is to check that our date/time support works regardless of the JVM timezone. I agree it may not be a good idea to update it at runtime. Maybe we should move JVM timezone changes to some `@Before`/`@After` methods? I don't know why I did it differently, to be honest.

famod:
### Yay, h2 1.4.197 is passing all tests! 🎉 🎆 🥳 
I'll squash everything into on commit and continue with 1.4.198+.

PS, the Windows test fix is here: #3512

famod:
### Failures in 1.4.198+

**1.4.198** breaks the follwing tests:
- `EmbeddedIdDatabaseGeneratedValueTest` also breaks with:
   ```
   javax.persistence.PersistenceException: org.hibernate.MappingException: No Dialect mapping for JDBC type: 2014
   ```
   but I already found a remedy for this: `SELECT CURRENT_TIMESTAMP` -> `SELECT LOCALTIMESTAMP`.
   This is because of https://github.com/h2database/h2database/pull/1181:
   >     2. Return type of `CURRENT_TIMESTAMP` is changed from `TIMESTAMP` to `TIMESTAMP WITH TIME ZONE` to be compatible with SQL:1999, Oracle, PostgreSQL and may be some others.
   The question at this point: Do we have to change `current_timestamp` in `H2Dialect` from `current_timestamp` to `localtimestamp`?
   [As of 1.4.200](http://h2database.com/html/functions.html?highlight=CURRENT_TIMESTAMP&search=CURRENT_TIMESTAMP#current_date) other functions also include the time zone and might fail as well but those changes might have been made _after_ 1.4.198.

- `LockTest. testContendedPessimisticLock`
- `LockTest. testFindWithPessimisticWriteLockTimeoutException`
- `LockTest. testNamedQueryResultListPessimisticWriteLockTimeoutException`
- `LockTest. testQueryResultListPessimisticWriteLockTimeoutException`
- `LockTest. testQuerySingleResultPessimisticWriteLockTimeoutException`
- `ManyToOneWithFormulaTest. testManyToOneFromNonPkToNonPk`
- `ManyToOneWithFormulaTest. testManyToOneToPkWithOnlyFormula`
- `WithClauseTest. testWithClauseAsSubqueryWithKeyAndOtherJoinReference`
- `MapElementFormulaTest. testManyToManyFormula`

**1.4.199** breaks the following tests:
- `EmbeddedIdDatabaseGeneratedValueTest` (see 1.4.198)
- `LockTest. [...]` (see 1.4.198)

Because 1.4.199 has _fewer_ failures, I skipped 1.4.198 and had a look at the `LockTest` failures:
```
18:48:29,884 ERROR DriverManagerConnectionProviderImpl:300 - Connection leak detected: there are 2 unclosed connections upon shutting down pool jdbc:h2:mem:db1;DB_CLOSE_DELAY=-1;LOCK_TIMEOUT=10000
18:48:29,885  WARN SqlExceptionHelper:137 - SQL Error: 90030, SQLState: 90030
18:48:29,885 ERROR SqlExceptionHelper:142 - File corrupted while reading record: null. Possible solution: use the recovery tool; SQL statement:
select lock0_.id as id1_0_0_, lock0_.name as name2_0_0_, lock0_.version as version3_0_0_ from Lock_ lock0_ where lock0_.id=? for update [90030-199]
18:48:29,885  INFO CustomRunner:123 - Test: testFindWithTimeoutHint
18:48:29,885  INFO LogHelper:31 - HHH000204: Processing PersistenceUnitInfo [name: LockTest]
18:48:29,886  INFO DefaultLoadEventListener:116 - HHH000327: Error performing load command
org.hibernate.exception.JDBCConnectionException: could not extract ResultSet
	at org.hibernate.exception.internal.SQLExceptionTypeDelegate.convert(SQLExceptionTypeDelegate.java:48)
	[...]
	at org.hibernate.jpa.test.lock.LockTest.testFindWithPessimisticWriteLockTimeoutException(LockTest.java:95)
	[...]
Caused by: org.h2.jdbc.JdbcSQLNonTransientConnectionException: File corrupted while reading record: null. Possible solution: use the recovery tool; SQL statement:
select lock0_.id as id1_0_0_, lock0_.name as name2_0_0_, lock0_.version as version3_0_0_ from Lock_ lock0_ where lock0_.id=? for update [90030-199]
	at org.h2.message.DbException.getJdbcSQLException(DbException.java:617)
	at org.h2.message.DbException.getJdbcSQLException(DbException.java:427)
	at org.h2.message.DbException.get(DbException.java:194)
	at org.h2.mvstore.db.MVTableEngine$Store.convertIllegalStateException(MVTableEngine.java:205)
	at org.h2.mvstore.db.MVTable.convertException(MVTable.java:725)
	at org.h2.mvstore.db.MVPrimaryIndex.lockRow(MVPrimaryIndex.java:235)
	[...]
Caused by: java.lang.IllegalStateException: Map null(10) is closed. null [1.4.199/4]
	at org.h2.mvstore.DataUtils.newIllegalStateException(DataUtils.java:883)
	at org.h2.mvstore.MVMap.beforeWrite(MVMap.java:890)
	at org.h2.mvstore.MVMap.operate(MVMap.java:1649)
	at org.h2.mvstore.tx.TransactionMap.set(TransactionMap.java:306)
	at org.h2.mvstore.tx.TransactionMap.lock(TransactionMap.java:268)
	at org.h2.mvstore.db.MVPrimaryIndex.lockRow(MVPrimaryIndex.java:233)
	... 56 more
```
This looks rather scary so I checked whether 1.4.200 fixes this and _it does_! So better skip both 1.4.198 _and_ 1.4.199.

**1.4.200** breaks 293 tests, mostly related to time stuff (again).
1.4.200 officially supports `TIME WITH TIME ZONE` (see https://github.com/h2database/h2database/releases/tag/version-1.4.200=, but Hibernate 5 does not?

fax4ever:
Hi @famod, maybe for now we should just upgrade to `1.4.197`.

Moreover, we could avoid adding the H2 test compile dependency (as I did) and using reflection to call `DateTimeUtils.resetCalendar()`. What do you think about that?

Thanks again for this contribution!

famod:
> maybe for now we should just upgrade to `1.4.197`.

Yeah, seems we've kind of hit a wall for now.

See also https://groups.google.com/d/msg/h2-database/aoGRFlSX2hU/1N5xCaUuBQAJ:

> 1. 1.4.200 was the last release in 1.4 series of releases and upcoming 2.0.202 has significant changes everywhere and shouldn't be considered as drop-in replacement, you may need some adjustments for it.
> 
> 2. We have some known unintended regressions that need to be fixed before the release of H2 2.0, so the release is delayed until they will be fixed. There is no exact timeline.
> 
> 3. 1.4.200 definitely has a bug, but I have no idea why you ever use `ResultSet.getTimestamp()` on a `TIME` column. Applications normally use ResultSet.getObject(column, LocalTime.class) (this is the best choice for H2) or legacy ResultSet.getTime(column) (has problems at least in one time zone due to own design flaws of java.sql.Time). Even more, conversion from TIME to TIMESTAMP was aligned with the SQL Standard in H2 1.4.200 (due to bug in JDBC layer in this version you need to use CAST directly in the SQL code to see how it works). This conversion in not compatible with old versions of H2 anyway.

So it is unclear when h2 2.0 will be released. The third point is interesting though. Maybe Hibernate could adapt to that?

If I find some time I will conduct some tests with 2.0.201-SNAPSHOT to see whether anything is fixed or fails additionally.

> Moreover, we could avoid adding the H2 test compile dependency (as I did) and using reflection to call `DateTimeUtils.resetCalendar()`. What do you think about that?

Hm, not sure since it is only a test dependency? @Sanne @yrodiere @gavinking WDYT?

yrodiere:
> The third point is interesting though. Maybe Hibernate could adapt to that?

From what I understand that's what Hibernate 6 will do. Not sure that's feasible in Hibernate 5, because of the backward compatibility requirements.

> Hm, not sure since it is only a test dependency? @Sanne @yrodiere @gavinking WDYT?

I'm personally not very comfortable with having H2 in the classpath when testing PostgreSQL for example. I'd rather have the test environment as close as possible to what users will actually use... But that's just my opinion.

gavinking:
> > The third point is interesting though. Maybe Hibernate could adapt to that?
> 
> 
> 
> From what I understand that's what Hibernate 6 will do. Not sure that's feasible in Hibernate 5, because of the backward compatibility requirements.

Nope. It only uses `setObject()` for `OffsetDateTime` and `ZonedDateTime`. Not for the `Local` types.

And I guess that makes sense: remember that H6 still needs to work properly on all sorts of legacy JDBC drivers.

gavinking:
@famod 

> I have no idea why you ever use `ResultSet.getTimestamp()` on a TIME column. Applications normally use `ResultSet.getObject(column, LocalTime.class)` 

huh? are you sure??

As far as I cam see all time types use `TimeTypeDescriptor` which calls `ResultSet.getTime()`, which is correct, as far as I can see.

famod:
@gavinking That was a quote from one of the h2 maintainers and I guess he did not check the Hibernate code and was making an assumption.

gavinking:
The it sounds like he's just trying to make excuses for `getTime()` being broken.

gavinking:
Again, I dispute the notion that Hibernate is doing it wrong be using `getTime()` for local times. That is the oldest operation in the JDBC API and should therefore be the most robust one (across many platforms and drivers).

famod:
He was actually questioning `getTimeStamp()` and not so much `getTime()`: https://groups.google.com/d/msg/h2-database/aoGRFlSX2hU/1N5xCaUuBQAJ

gavinking:
OK bit that leaves me confused about why we think it's calling `getTimestamp()`? I don't see how it could be, unless this is some very specific case (a `UserType` or whatever).

famod:
A this point I suggest that I'll:
- reduce the scope of HHH-14031 from 1.4.200+ to 1.4.197
- create a new PR for 1.4.197
- create a new ticket for 1.4.197 -> 2.0.201
- change this PR to mention the new ticket
- rebase this PR once the other, smaller scoped PR is merged

This way we have a clear separation and we do not "loose" this valuable discussion.

famod:
I've created the PR updating to just 1.4.197 for now (including reflective call of `org.h2.util.DateTimeUtils`): #3516

gavinking:
On the whip/6.0 branch we are using 1.4.199. Did you try that version?

famod:
Interesting! Yes I did try that version but as written further up it fails in LockTest. It actually corrupts the database file!

gavinking:
> Interesting! Yes I did try that version but as written further up it fails in LockTest. It actually corrupts the database file!

Sorry, my bad, I must be blind.

gavinking:
> The question at this point: Do we have to change `current_timestamp` in H2Dialect from `current_timestamp` to `localtimestamp`?

@famod FTR I think this would be just fine. We already do that in some dialects, at least in H6.

famod:
> On the whip/6.0 branch we are using 1.4.199. Did you try that version?

h2 was updated to .199 in that branch via https://github.com/hibernate/hibernate-orm/commit/eab6107ec2e7b3a0c06146a9ff51b9964f4b3169
Strangely, I cannot find `LockTest` in the build log for this commit and when I build locally, the test is not listed in the test report?!

gavinking:
> Strangely, I cannot find LockTest in the build log for this commit and when I build wip/6.0 locally, the test is not listed in the test report?!

Yeah, so the thing about wip/6.0 is it has a lot fewer tests than master.

gavinking:
@famod I also found https://github.com/h2database/h2database/issues/2851

famod:
@gavinking Have you also tried that with 2.0.201-SNAPSHOT (from h2 `master`)?
I can share the gradle config bits needed to use the locally built h2 SNAPSHOT into Hibernate.

gavinking:
I have not tried, and I don't really have any more time to spend on this right now.

gavinking:
> > The question at this point: Do we have to change `current_timestamp` in H2Dialect from `current_timestamp` to `localtimestamp`?
> 
> @famod FTR I think this would be just fine. We already do that in some dialects, at least in H6.

So I did try this out on wip/6.0 and it does indeed seem to be the right thing to do on 1.4.200.

All I did was add:

```java
	@Override
	public String currentTime() {
		return "localtime";
	}

	@Override
	public String currentTimestamp() {
		return "localtimestamp";
	}
```

to `H2Dialect`.

famod:
> I have not tried, and I don't really have any more time to spend on this right now.

Ok, thanks for checking anyway. In case you have that test committed somewhere I could use it to test the SNAPSHOT when I find some time.

NathanQingyangXu:
After we update to 2.0.x, maybe we can revert back the version specific logic committed in https://github.com/hibernate/hibernate-orm/pull/3500/commits/3eb97f98c8944b549b244f6ca9c0459c1e6aa284?

Screen Shot 2020-09-07 at 11 32 56 AM

beikov:
If we remove the version specific logic, queries as shown in the test cases of the commit will stop working when using H2 1.4.197 or older. I doubt we want that. AFAIK many H2 users don't upgrade because of issues with the MVStore in newer H2 versions.

Sanne:
I just rebased this and then figured it was already included in `main` ... I'll close it.

PR #3413, opened 2020-05-19
HHH-14022: Revert "HHH-13322"

123Haynes:
https://hibernate.atlassian.net/browse/HHH-14022

This reverts commit 188c05cc333dec0fe3edfefd8f27fe19b63c702a.
The original issue being addressed by this commit was fixed by limiting the sequences to the current user, but this broke other setups.

The oracle docs on the different sequence views:
https://docs.oracle.com/cd/B28359_01/server.111/b28320/statviews_2053.htm#REFRN20244
"ALL_SEQUENCES describes all sequences accessible to the current user."  

Imho people should use different users which can only see the sequences they need to see to prevent the issue.


hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
```diff
@@ -420,7 +420,7 @@ else if ( initialValue > 0 && incrementSize < 0 ) {
 		else {
 			return
 				String.format(
-						"%s start with %d increment by %d",
+						"%s start with %d increment by  %d",
```
NathanQingyangXu, line 423: is the above space char intended?

hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
```diff
@@ -490,7 +490,7 @@ public boolean forUpdateOfColumns() {
 
 	@Override
 	public String getQuerySequencesString() {
-		return "select * from user_sequences";
+		return "select * from all_sequences";
```
NathanQingyangXu, line 493: I am the original committer to change `all_sequences` to `user_sequences`. Guess I made a mistake.

hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
123Haynes: @NathanQingyangXu idk. You removed it on your PR. This is a simple revert of all changes of you commit.
I can remove the space char again if you want.

dreab8:
applied on master and 5.4.

Thanks @123Haynes

PR #3414, opened 2020-05-19
HHH-14033 - SQL script parsing problem with multi-line comments

sebersole:
- Better handling of multi-line comments
- Restructured some internal classes to consolidate packages
- Added "system"-style SchemaToolingLogging

https://hibernate.atlassian.net/browse/HHH-14033


sebersole:
Consider specifically whether y'all think it is appropriate to switch the exception type for the missing `;` error

dreab8:
Hi Steve,  the specific message was introduce to fix https://github.com/quarkusio/quarkus/issues/3817, but I do not think that changing the exception type is a problem.

hibernate-core/hibernate-core.gradle
```diff
@@ -203,6 +203,9 @@ xjc {
     }
 }
 
+generateGrammarSource {
+    arguments += "-traceParser"
```
Sanne, line 207: Is this going to affect the regular HQL parser as well?

Specifically I'm wondering about performance impact, if this is to help during development maybe it shoduld be commented out when done with debugging?

sebersole: The impact - if you look in the generated parsers, you will now see Antlr has added traceIn/traceOut calls at the start and end of each rule method.  I use that as hooks to debug "parse tree info".  This is extremely useful information for debugging parse problems; unless we can demonstrate that it is a performance problem I'd prefer to leave these as-is.

sebersole: Its a good point though in that the default impl Antlr injects is to use sysout.  For HQL and for SQL script parsing I've overridden these to use logging.  I'll add the same override for the entity-graph parser and really the impact overall should be minimal

sebersole:
Right - I figured.  Which is why I kept the message unchanged

sebersole:
Going to push an additional commit.  If y'all agree I'll just puch this to master

dreab8:
@sebersole it is fine to me

Sanne:
> Going to push an additional commit. If y'all agree I'll just push this to master

sure!

PR #3415, opened 2020-05-21
Oracle schema validation fails due to missing sequences

varunsufi:
Fetching only user_sequences instead of all_sequences is a breaking change for many applications.

Hibernate-core is updated with spring-boot-dependencies update from 2.2.6 to 2.2.7.

Usually, user creating DB resources (sequences) is admin even if flyway/liquibase is used then user creating the DB resources is different from the DB user running the application.

With Spring config spring.jpa.hibernate.ddl-auto=validate the startup of application fails as it cannot find the user_sequences.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
@@ -490,7 +490,7 @@ public class Oracle8iDialect extends Dialect {
 
 	@Override
 	public String getQuerySequencesString() {
-		return "select * from user_sequences";
+		return "select * from all_sequences";
 	}
 
 	public SequenceInformationExtractor getSequenceInformationExtractor() {


123Haynes:
@varunsufi The related ticket is: https://hibernate.atlassian.net/browse/HHH-14022

varunsufi:
@123Haynes I will mark https://hibernate.atlassian.net/browse/HHH-14039 as duplicate and close it.

NathanQingyangXu:
Thanks for your PR. I approved @123Haynes and propose closing this one for duplication.

dreab8:
@NathanQingyangXu this change is breaking `OracleExtractSequenceMetadataTest` that is related with https://hibernate.atlassian.net/browse/HHH-1332 ( but not sure the test represents the issue described in this Jira)

I wonder if
```
@Override
		return    " select sequence_name from all_sequences as"
				+ "  union"
				+ " select synonym_name"
				+ "   from all_synonyms us, all_sequences asq"
				+ "  where asq.sequence_name = us.table_name"
				+ "    and asq.sequence_owner = us.table_owner";
	}

```
can fix also HHH-1332

123Haynes:
@dreab8 https://github.com/hibernate/hibernate-orm/pull/3413 removes that test as well.
The fix for HHH-13322 is to use different Users that don't have permissions to see the conflicting sequences in other schemas imho. I don't think hibernate needs to deal with that at all.

dreab8:
My point is that that the revert leaves open the issue with HHH-1332 and that probably `OracleExtractSequenceMetadataTest` was not a correct representation of HHH-1332 and that probably the correct solution is the change to `public String getQuerySequencesString()` i proposed

123Haynes:
@dreab8 But doesn't that issue only appear when you use the same db user for different applications or a user with sysdba rights?
Why not simply use different users for the applications. Then all_sequences simply won't show the conflicting sequences and the issue does not happen at all.

dreab8:
I'm not sure the reporter affirms the application was working, an update to hibernate caused the failure, in the description at some point he wrote 
> because the sequence is a synonym to a sequence in a different schem

dreab8:
@123Haynes can you check if 
```
@Override
		return    " select sequence_name from all_sequences as"
				+ "  union"
				+ " select synonym_name"
				+ "   from all_synonyms us, all_sequences asq"
				+ "  where asq.sequence_name = us.table_name"
				+ "    and asq.sequence_owner = us.table_owner";
	}
```
is working for you?

Thanks

123Haynes:
@dreab8 I'll try, but it could take a few hours.

My Point is:

"ALL_SEQUENCES describes all sequences accessible to the current user."

So if the current user does not have the rights to view the sequences of the other schemas, which by default he has not unless he has sysdba rights, the problem simply will not appear at all.
Imho this is simply a case of using a user with too many rights.

According to https://docs.oracle.com/cd/B19306_01/server.102/b14237/statviews_2091.htm#REFRN20273
ALL_SYNONYMS works the same. So I'm not sure how your statement would improve the situation there.
You'd still get the same sequence name twice if you have 2 instances of the same db in different schemas.

varunsufi:
> Thanks for your PR. I approved @123Haynes and propose closing this one for duplication.

Should I close this PR ?

dreab8:
applied https://github.com/hibernate/hibernate-orm/pull/3413 instead

Thanks @varunsufi and @123Haynes

PR #3416, opened 2020-05-21
HHH-14036 Add timezone('TIMEZONE', TIMESTAMP) to the out of the box functions for PostgreSQL10Dialect

dreab8:
https://hibernate.atlassian.net/browse/HHH-14036

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL10Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL10Dialect.java
@@ -6,13 +6,19 @@
  */
 package org.hibernate.dialect;
 
+import org.hibernate.dialect.function.StandardSQLFunction;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL10Dialect.java
```diff
@@ -6,14 +6,20 @@
...
 
 /**
  * An SQL dialect for Postgres 10 and later.
  */
 public class PostgreSQL10Dialect extends PostgreSQL95Dialect {
 
+	public PostgreSQL10Dialect() {
+		registerFunction( "timezone", new StandardSQLFunction( "timezone", StandardBasicTypes.TIMESTAMP ) );
+	}
+
```
NathanQingyangXu, line 22: I think `timezone` function is included in postgreSQL from the very beginning. E.g. https://www.postgresql.org/docs/8/functions-datetime.html

PR #3417, opened 2020-05-24
HHH-14041 - H2Dialect: fix referential integrity constraint constraint name extraction

Gittenburg:
https://hibernate.atlassian.net/browse/HHH-14041

The problem is that h2 does not directly substitute the constraint name but wraps it in a description: https://github.com/h2database/h2database/blob/9e2e212cf806fa75cd42437a13a33cb3185f5fb1/h2/src/main/org/h2/constraint/ConstraintReferential.java#L312


Sanne:
Thanks @Gittenburg ! We normally require a test but this looks like trivial enough, I'll merge it.

PR #3419, opened 2020-05-24
expose a private field to Hibernate Reactive

gavinking:
needed to support Criteria queries


DavideD:
I think we need this on 5.4 as well

dreab8:
@DavideD done

DavideD:
Thanks a lot

gavinking:
Thanks @dreab8

PR #3420, opened 2020-05-27
HHH-14044 : Extract code from Loader into protected methods

gbadner:
https://hibernate.atlassian.net/browse/HHH-14044

This is for Hibernate Reactive:
https://github.com/hibernate/hibernate-reactive/issues/120.

The changes:

* extract code from `Loader#prepareQueryStatement` that binds a `PreparedStatement` into `protected` method named `#bindPreparedStatement`;
* change `private processResultSet( ResultSet, RowSelection, LimitHandler, boolean, SharedSessionContractImplementor)` to `protected preprocessResultSet(...)`, since it really does preprocessing. (`#processResultSet(ResultSet , QueryParameters, SharedSessionContractImplementor, boolean, ResultTransformer, int, List<AfterLoadAction>) actually does the processing).


gavinking:
Great! This looks excellent, @gbadner. I have a couple of comments and questions but I want to actually try it out first. Hold on.

dreab8:
Applied on master and 5.4
Thanks @gbadner

PR #3422, opened 2020-05-27
make three private methods protected for use in Hibernate Reactive

gavinking:
I can delete a little bit of copy/pasted code if you expose these methods for me.


dreab8:
Hi @gavinking , 
does an associated Jira already exist? if not can you please create one?

Thanks

dreab8:
applied to master and 5.4 branch.

Thanks @gavinking

PR #3425, opened 2020-06-02
HHH-14055 - Add UUID support for fields annotated with @RevisionNumber

bbanchev:
Add fallback check if the field type is of type java.util.UUID
Update corresponding MappingException message

### DIFF
--- a/hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/RevisionInfoConfiguration.java
+++ b/hibernate-envers/src/main/java/org/hibernate/envers/configuration/internal/RevisionInfoConfiguration.java
@@ -8,6 +8,7 @@ package org.hibernate.envers.configuration.internal;
 
 import java.util.Date;
 import java.util.Set;
+import java.util.UUID;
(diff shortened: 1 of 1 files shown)


sebersole:
`@RevisionNumber` explicitly says that "values of this property should form a strictly-increasing sequence of numbers".  A UUID obviously does not fit that description.  

@Naros you are much better versed in envers than I... what do you think?

bbanchev:
> `@RevisionNumber` explicitly says that "values of this property should form a strictly-increasing sequence of numbers". A UUID obviously does not fit that description.
> 
> @Naros you are much better versed in envers than I... what do you think?

Actually this is one of the main reasons to ask for this feature - need unpredictable id sequence for revisions. On the other hand fits better in distributed environment.

sebersole:
When a contract explicitly says "should do X" and something explicitly
breaks that contract... I like to have an expert explain the ramifications.

Whether using a UUID would work better in distributed environment, it is
irrelevant if the code-base truly expects sequentiality.  So let's just
wait to hear from Chris

PR #3428, opened 2020-06-05
HHH-10186 - Session.load enhanced entity with eager collection loadin…

VladoKuruc:
…g mapping failded with PropertyAccessException reproducing test.
The issue is already fixed in the meantime, so it can be closed. Maybe this unit test is important for regress


gavinking:
@VladoKuruc reported the issue fixed, and the test uses features of Hibernate that have been removed in the meantime.

PR #3430, opened 2020-06-11
wip/6.0 Enable entity graph parser test cases

NathanQingyangXu:
The first step to entity graph verification. The following bugs were fixed:

1. ANTLR rule errors
2. entity graph subtype verification error

### DIFF
--- a/hibernate-core/src/main/antlr/org/hibernate/grammars/graph/GraphLanguageParser.g4
+++ b/hibernate-core/src/main/antlr/org/hibernate/grammars/graph/GraphLanguageParser.g4
@@ -33,15 +33,15 @@ attributeList
     ;
 
 attributeNode
-	: attributePath (subGraph)?
+	: attributePath subGraph?
 	;
 
 attributePath
-	: NAME attributeQualifier?
+	: ATTR_NAME attributeQualifier?
 	;
 
 attributeQualifier
-	: DOT NAME
+	: DOT ATTR_NAME
 	;
 
 subGraph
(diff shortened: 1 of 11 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/DomainModelHelper.java
```diff
@@ -29,7 +29,7 @@
 		}
 
 		// first, try to find it by name directly..
-		ManagedDomainType<S> subManagedType = jpaMetamodel.entity( subTypeName );
+		ManagedDomainType<S> subManagedType = jpaMetamodel.resolveHqlEntityReference( subTypeName );
```
NathanQingyangXu, line 32: usually subType is not specified as hibernate entity name, but as HQL entity name

hibernate-core/src/main/antlr/org/hibernate/grammars/graph/GraphLanguageParser.g4
```diff
@@ -33,21 +33,21 @@ attributeList
...
+	: DOT ATTR_NAME
 	;
 
 subGraph
 	: LPAREN (subType COLON)? attributeList RPAREN
 	;
 
 subType
-	: NAME
+	: TYPE_NAME
```
NathanQingyangXu, line 52: so the following errors flushed out

- uppercase should be allowed in NAME_CONTINUATION;
- subType should starts with uppercase;

PR #3431, opened 2020-06-11
HHH-14059 Cannot use full qualified enum name which has two uppercase alphabet prefix

zhengyhn:
Hi,
    [Here](https://hibernate.atlassian.net/browse/HHH-14059) is the jira issue about this pull request. Please review whether it's ok.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/internal/util/ReflectHelperTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/internal/util/ReflectHelperTest.java
@@ -230,4 +230,15 @@ public class ReflectHelperTest {
 	public void test_setMethod_nestedInterfaces_on_superclasses() {
 		assertNotNull( ReflectHelper.findSetterMethod( E.class, "id", String.class ) );
 	}
+
+	@TestForIssue(jiraKey = "HHH-14059")
+	@Test
+	public void test_getConstantValue_UpperCaseEnum() {
+		when( sessionFactoryOptionsMock.isConventionalJavaConstants() ).thenReturn( true );
+
+		when( classLoaderServiceMock.classForName( "com.example.UStatus" ) ).thenReturn( (Class) Status.class );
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/internal/util/ReflectHelperTest.java
```diff
@@ -230,4 +230,15 @@ public void test_getMethod_nestedInterfaces_on_superclasses()
...
+	@TestForIssue(jiraKey = "HHH-14059")
+	@Test
+	public void test_getConstantValue_UpperCaseEnum() {
+		when( sessionFactoryOptionsMock.isConventionalJavaConstants() ).thenReturn( true );
+
+		when( classLoaderServiceMock.classForName( "com.example.UStatus" ) ).thenReturn( (Class) Status.class );
+		Object value = ReflectHelper.getConstantValue( "com.example.UStatus.OFF", sessionFactoryImplementorMock);
+		assertEquals( OFF, value );
+		verify(classLoaderServiceMock, times(1)).classForName( eq("com.example.UStatus") );
+	}
```
NathanQingyangXu, line 243: In Hibernate testing, we seldom rely on mocking but go about testing through real integration test.

zhengyhn: Hi @NathanQingyangXu ,  thank you for your comment. I found I can extend `BaseEntityManagerFunctionalTestCase` to get an `entityManagerFactory` so that there is no need for mocking. However, the existing tests of this `ReflectHelper` use mocking. I want to be consistent with the [existing](https://github.com/hibernate/hibernate-orm/blob/1fcfbaa1c070846ba075ba3e40cbeceba7bf8047/hibernate-core/src/test/java/org/hibernate/internal/util/ReflectHelperTest.java).

dreab8:
applied upstream.

Thanks @zhengyhn

PR #3432, opened 2020-06-12
wip/6.0 Remove potential duplicated class loading in JpaMetamodelImpl

NathanQingyangXu:
The motivation of this PR is to remove the `todo` comment in `JpaMetamodelImpl`:
Screen Shot 2020-06-11 at 11 45 58 PM

The gist of this PR is to store loaded class to avoid duplicated loading later.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/JpaMetamodelImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/JpaMetamodelImpl.java
@@ -58,13 +58,23 @@ import org.hibernate.type.spi.TypeConfiguration;
  */
 public class JpaMetamodelImpl implements JpaMetamodel {
 	private static final EntityManagerMessageLogger log = HEMLogging.messageLogger( JpaMetamodel.class );
-	private static final String INVALID_IMPORT = "";
+	private static final ImportInfo<?> INVALID_IMPORT = new ImportInfo<>( null, null );
+
+	private static class ImportInfo<T> {
+		final String importedName;
+		Class<T> loadedClass; // could be null for boot metamodel import; not final to allow for populating later
+
+		ImportInfo(String importedName, Class<T> loadedClass) {
+			this.importedName = importedName;
+			this.loadedClass = loadedClass;
+		}
+	}
 
 	private final TypeConfiguration typeConfiguration;
 
 	private final Map<String, EntityDomainType<?>> jpaEntityTypeMap = new ConcurrentHashMap<>();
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/JpaMetamodelImpl.java
```diff
@@ -272,37 +289,32 @@ public void visitEmbeddables(Consumer<EmbeddableDomainType<?>> action) {
...
+		final ImportInfo<?> importInfo = resolveImport( queryName );
+		return importInfo == null ? null : importInfo.importedName;
 	}
 
-	private String resolveImportedName(String name) {
-		String result = nameToImportNameMap.get( name );
-		if ( result == null ) {
+	private <T> ImportInfo<T> resolveImport(String name) {
+		//noinspection unchecked
+		final ImportInfo<T> result = (ImportInfo<T>) nameToImportMap.computeIfAbsent( name, unknownName -> {
```
NathanQingyangXu, line 298: The benefit of using `ConcurrentHashMap#computeIfAbsent()` is thread safety and atomicity.

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/JpaMetamodelImpl.java
```diff
@@ -487,8 +499,6 @@ public void processJpa(
 
 			context.wrapUp();
 
-			this.nameToImportNameMap.putAll( bootMetamodel.getImports() );
```
NathanQingyangXu, line 490: @sebersole it seems the above statement is redundant for it has been done at the beginning of the method. Is that correct?

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/JpaMetamodelImpl.java
```diff
@@ -92,23 +102,30 @@ public TypeConfiguration getTypeConfiguration() {
...
-		final Class<X> requestedClass = resolveRequestedClass( entityName );
-		if ( requestedClass != null ) {
-			return resolveEntityReference( requestedClass );
+		if ( loadedClass == null ) {
+			loadedClass = resolveRequestedClass( entityName );
+			// populate class cache for boot metamodel imports
+			if ( importInfo != null && loadedClass != null ) {
+				importInfo.loadedClass = loadedClass;
+			}
+		}
```
NathanQingyangXu, line 123: The above code snippet is only meant to populate class cache for boot metamodel imports. Not sure whether it is needed. If not, the above code snippet can be deleted.

PR #3433, opened 2020-06-13
wip/6.0 Fix thread safety issue in QueryInterpretationCacheStandardImpl

NathanQingyangXu:
As explained in https://dzone.com/articles/concurrenthashmap-isnt-always-enough, **ConcurrentHashMap**'s **computeIfAbsent**() method enjoys the benefit of atomicity (for the same key only one thread is running while others are blocked) so we can avoid many complicated race conditions.

It is a pity that the **BoundedConcurrentHashMap** class doesn't provide atomic overridden '**computeIfAbsent**()' method as compared to the **ConcurrentHashMap** from JDK. However, the code changes in this PR still enjoys the fact that internally '**putIfAbsent**()' (which is atomic) is used for unknown key so we can avoid populating cache multiple times (the earliest one wins).

This PR fixes the atomicity issue of 'computeIfAbsent()' for **BoundedConcurrentHashMap** as well.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/internal/QueryInterpretationCacheStandardImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/internal/QueryInterpretationCacheStandardImpl.java
@@ -77,14 +77,7 @@ public class QueryInterpretationCacheStandardImpl implements QueryInterpretation
 			Supplier<SelectQueryPlan> creator) {
 		log.tracef( "QueryPlan#getSelectQueryPlan(%s)", key );
 
-		final SelectQueryPlan cached = (SelectQueryPlan) queryPlanCache.get( key );
-		if ( cached != null ) {
-			return cached;
-		}
-
-		final SelectQueryPlan plan = creator.get();
-		queryPlanCache.put( key, plan );
-		return plan;
+		return (SelectQueryPlan) queryPlanCache.computeIfAbsent( key, k -> creator.get() );
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/internal/QueryInterpretationCacheStandardImpl.java
```diff
@@ -77,14 +77,7 @@ public SelectQueryPlan resolveSelectQueryPlan(
...
 
-		final SelectQueryPlan cached = (SelectQueryPlan) queryPlanCache.get( key );
-		if ( cached != null ) {
-			return cached;
-		}
-
-		final SelectQueryPlan plan = creator.get();
-		queryPlanCache.put( key, plan );
-		return plan;
+		return (SelectQueryPlan) queryPlanCache.computeIfAbsent( key, k -> creator.get() );
```
NathanQingyangXu, line 80: internally the following default implementation of 'computeIfAbsent' from ConcurrentMap will be run. It is a pity we cannot enjoy the atomicity in ConcurrentHashMap in JDK, but the usage of 'putIfAbsent()' is definitively desirable:
Screen Shot 2020-06-13 at 2 46 19 PM

NathanQingyangXu:
I would try to fix the 'computeIfAbsent()' atomicity issue in BoundedConcurrnetHashMap. For this reason, I close this PR for the time being

NathanQingyangXu:
I need more time to solve the testing failure, so I close this PR for the time being.

Sanne:
hi @NathanQingyangXu , are you familiar with the notion of benign data races?

Please don't fix this: while there is a possibility of running a query twice, this is very unlikely. But having the lock would introduce a very severe limitation to horizontal scalability - so the tradeoff isn't good.

NathanQingyangXu:
okay, I have closed the PR if that is not a valid concern.

NathanQingyangXu:
that being said, the fix won't add overhead for we have used concurrent
map. We simply replace two locks with one. the real downside is code
complexity and maintainability. Seems Guava cache shares much with our
BoundedConcurrentHashmap, and it has atomic support for 'compute' map
methods.

Sanne:
The "atomic" implementation of JDK's ConcurrentHashMap requries a blocking lock on the segment, which would severly limit scalability.

I have not checked the implementation from Guava.. maybe there's something smarter there, but please check for that with caution: it's better to occasionally miss the cache than to introduce blocking operations while accessing the cache.

NathanQingyangXu:
the cache we are using is BoundedConcurrentHashmap, it was copied from
hibernate-infinity and is based on some early version of ConcurrentHashmap.
We have paid the cost already.

Sanne:
> the cache we are using is BoundedConcurrentHashmap, it was copied from
hibernate-infinity and is based on some early version of ConcurrentHashmap.

It was based on the library that preceded Caffeine - forgot the name now.

> We have paid the cost already.

I'm not understanding, what cost are you referring to?

Thanks!

NathanQingyangXu:
forget about it. I misunderstood something. Thanks for the clarification.

PR #3434, opened 2020-06-14
HHH-14071 Fix binding for FetchProfile for OneToOne(mappedBy=...) associations

oliversaggau:


Sanne:
Hi @oliversaggau , apologies for the late feedback. It's an excellent patch, thanks a lot! I've merged it in master and backported to 5.4, it will be included in the next 5.4 maintenance release.

Thanks again!

PR #3435, opened 2020-06-16
Circularity detection, EmbeddedId and other fixes

dreab8:
This PR contains :

- changes in the way circularity is detected (AssociationKey + mappedBy),  introduces `EntitySelectFetchByUniqueKeyInitializer ` and fixes for some issues with EmbeddedId (circularity and PluralAttribute) [commit](https://github.com/hibernate/hibernate-orm/pull/3435/commits/34d201400b98d65fa5d72d233813d71d4fe91730)

-  a fix for PluralAttribute manyToMany FK target part determination [commit](https://github.com/hibernate/hibernate-orm/pull/3435/commits/4eda5ffb393d389e519374c7938a11652d669228)

- a fix for fetching a ManyToOne  attribute using a join table using inner join when a left outer should be used instead [commit](https://github.com/hibernate/hibernate-orm/pull/3435/commits/1d0e69178d479aad0c858577f5e1bdfb738f1857)

- a fix for joined subclass discriminator column determination [commit](https://github.com/hibernate/hibernate-orm/pull/3435/commits/1ec72e856fdb6e5ba538925117f678a646c35576)

and re-enables more tests.


hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -419,15 +426,21 @@ private void applyOrdering(TableGroup tableGroup, OrderByFragment orderByFragmen
 		log.tracef( "Starting visitation of FetchParent's Fetchables : %s", fetchParent.getNavigablePath() );
 
 		final List<Fetch> fetches = new ArrayList<>();
+		String fullPath = fetchParent.getNavigablePath().getFullPath();
+		if ( visitedNavigablePath.containsKey( fullPath ) ) {
+			return visitedNavigablePath.get( fullPath );
+		}
```
NathanQingyangXu, line 432: It seems the above code snippet performance could be improved as following:
```
final List<Fetch> fullPathFetches = visitedNavigablePath.get( fullPath );
if ( fullPathFetches != null ) {
    return fullPathFetches;
}
```
It would incur only retrieval once instead of twice. Given that the map value is collection which is never null, we could simplify accordingly. Your code snippet would be mandatory if null is a possible value, though.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -419,15 +426,21 @@ private void applyOrdering(TableGroup tableGroup, OrderByFragment orderByFragmen
...
 
 		final BiConsumer<Fetchable, Boolean> processor = createFetchableBiConsumer( fetchParent, querySpec, creationState, fetches );
```
NathanQingyangXu, line 434: moving the declaration of 'fetches' before the above statement would improve code readability a little bit. As a general rule, putting variable declaration as near its usage as possible decrease human brain memorizing pressure, I think.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -495,7 +508,7 @@ else if ( fetchDepth > maximumFetchDepth ) {
 			}
 
 			try {
-				if ( !( fetchable instanceof BasicValuedModelPart ) ) {
+				if ( !( fetchable instanceof BasicValuedModelPart ) && !(fetchable instanceof EmbeddedAttributeMapping ) ) {
```
NathanQingyangXu, line 511: I am a little bit shamed to point out adding another space char after `(` in the above new code would make the format more consistent. Super minor for sure.

hibernate-core/src/main/java/org/hibernate/mapping/Column.java
```diff
@@ -276,6 +276,17 @@ public String getSqlType(Dialect dialect, Mapping mapping) throws HibernateExcep
 		return sqlType;
 	}
 
+	private Type getType(Mapping mapping, Type type) {
+		while ( !( type instanceof JdbcMapping ) ) {
+			//ManyToOneType doesn't implement JdbcMapping
+			type = mapping.getIdentifierType( ( (EntityType) type ).getAssociatedEntityName() );
+			if ( type instanceof ComponentType ) {
+				type = ( (ComponentType) type ).getSubtypes()[getTypeIndex()];
```
NathanQingyangXu, line 284: I saw there is some code duplication above (283-284 vs 251-252). Maybe we can eliminate it to avoid changing both when future refactoring is needed.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/EmbeddableMappingType.java
```diff
@@ -345,7 +345,18 @@ public void visitJdbcTypes(
...
+						( (ToOneAttributeMapping) attributeMapping ).getKeyTargetMatchPart().visitJdbcTypes(
+								action,
+								clause,
+								typeConfiguration
+						);
+					}
+					else {
+						attributeMapping.visitJdbcTypes( action, clause, typeConfiguration );
+					}
+				}
```
NathanQingyangXu, line 359: we can save a pair of `{}` in the lambda, though not a big deal

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EmbeddedForeignKeyDescriptor.java
```diff
@@ -330,17 +324,18 @@ public void visitTargetColumns(ColumnConsumer consumer) {
...
-			if ( !targetColumnExpressions.contains( columnNames[i] ) ) {
-				return false;
+	public AssociationKey getAssociationKey() {
+		if ( associationKey == null ) {
+			String[] associationKeyColumns = new String[keyColumnExpressions.size()];
+			int i = 0;
+			for ( String keyColumnExpression : this.keyColumnExpressions ) {
+				associationKeyColumns[i] = keyColumnExpression;
+				i++;
 			}
```
NathanQingyangXu, line 334: I am pretty sure `System.arrayCopy` seems to be much more performant, though not a big deal.

dreab8:
Thanks @NathanQingyangXu , applied your suggestions

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
```diff
@@ -525,17 +526,36 @@ else if ( persistentClass.isDiscriminatorValueNotNull() ) {
...
+		int subclassSpanMinusOne = subclassSpan - 1;
+		subclassClosure[subclassSpanMinusOne] = getEntityName();
 		if ( persistentClass.isPolymorphic() ) {
 			subclassesByDiscriminatorValue.put( discriminatorValue, getEntityName() );
 
-			discriminatorValuesByTableName = new LinkedHashMap<>( subclassSpan + 1 );
-			subclassNameByTableName = CollectionHelper.mapOfSize( subclassSpan + 1);
+			int initialCapacity = subclassSpan + 1;
+			discriminatorValuesByTableName = new LinkedHashMap<>( initialCapacity );
+			discriminatorColumnNameByTableName = new LinkedHashMap<>( initialCapacity );
```
NathanQingyangXu, line 536: It seems `LinkedHashMap` shares the same problem with `HashMap` in that the `initialCapacity` is not the same as `expectedSize` except the loadFactor is specified as 1. Only `ConcurrentHashMap` is immune with this annoying problem. If `CollectionHelper` doesn't support util static factory method for it, we might well create one there.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -169,101 +176,145 @@ public Fetch resolveCircularFetch(
...
-		if ( referencedEntityReference == null ) {
-			throw new HibernateException(
-					"Could not locate entity-valued reference for circular path `" + fetchablePath + "`"
-			);
+		if ( modelPart instanceof EntityCollectionPart ) {
+			if ( parentOfParent.getFullPath().endsWith( EntityIdentifierMapping.ROLE_LOCAL_NAME ) ) {
+				parentOfParent = parentOfParent.getParent();
+			}
+			return ( (PluralAttributeMapping) creationState.resolveModelPart( parentOfParent ) ).getMappedBy();
 		}
```
NathanQingyangXu, line 294: seems combining the above two if together in one if leads to better readability, though not a big deal

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/AssociationKey.java
```diff
@@ -0,0 +1,51 @@
...
+		}
+
+		final AssociationKey that = (AssociationKey) o;
+		return table.equals( that.table ) && columns.equals( that.columns );
+	}
+
+	@Override
+	public int hashCode() {
+		return table.hashCode();
+	}
```
NathanQingyangXu, line 39: why not include `columns' into hashCode as well? It has been used in 'equals'. Keep from using it doesn't violate the consistency between `hashCode` and `equals`, but it is a little bit confusing.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/AssociationKey.java
```diff
@@ -0,0 +1,51 @@
...
+
+	private String str;
+
+	@Override
+	public String toString() {
+		if ( str == null ) {
+			str = "AssociationKey(table=" + table + ", columns={" + String.join( ",", columns ) + "})";
```
NathanQingyangXu, line 46: I would use `, columns=[...]` instead of `, columns={..}` for '{}' is usually used for map. Not a big deal for sure.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/SimpleForeignKeyDescriptor.java
```diff
@@ -285,32 +288,25 @@ public EntityMappingType findContainingEntityMapping() {
...
 	@Override
-	public boolean areTargetColumnNamesEqualsTo(String[] columnNames) {
-		if ( columnNames.length != 1 ) {
-			return false;
+	public AssociationKey getAssociationKey() {
+		if ( associationKey == null ) {
+			List<String> associationKeyColumns = new ArrayList<>( 1 );
+			associationKeyColumns.add( keyColumnExpression );
+
+			associationKey = new AssociationKey( keyColumnContainingTable, associationKeyColumns );
```
NathanQingyangXu, line 307: Do we allow for the possibility that `associationKeyColumns` is mutable later? Otherwise `Collections.singletonList()` would be more appropriate and less confusing.

PR #3436, opened 2020-06-17
HHH-14075: Changes to loaders and TwoPhaseLoad to allow "internal" loading to be…

gbadner:
… reused by hibernate-reactive

https://hibernate.atlassian.net/browse/HHH-14075

Includes the following changes.

* Break `TwoPhaseLoad` into 2 methods, `#initializeEntityEntryLoadedState` and `#initializeEntityFromEntityEntryLoadedState`. `#initializeEntityEntryLoadedState` takes an argument of type `EntityResolver`, which is responsible for resolving to-one entity properties.
* A method was extracted from `Loader#processResultSet`, `#getRowsFromResultSet`, which reads all the rows from the `ResultSet`.
* Some methods in `Loader` and `OuterJoinLoader` were changed to `protected` or `public`, as needed so that Reactive implementations could make them `public`.
Changes to `AbstractLoadPlanBasedEntityLoader`, `AbstractLoadQueryDetails`, `EntityLoadQueryDetails`, `BatchingLoadQueryDetailsFactory`, to allow a `ResultSetProcessorResolver` to be provided by an implementation of `AbstractLoadPlanBasedEntityLoader`
* Changes to `ResultSetProcessorImpl` to allow hibernate-reactive loaders to extend it so that it can process the `ResultSet` and resolve hydrated objects reactively.
* Added `EntityType#isEager(Boolean overridingEager)` in order to determine if an entity attribute being resolved by `TwoPhaseLoad#initializeEntity` will be eager.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
@@ -35,13 +35,13 @@ import org.hibernate.event.spi.PreLoadEventListener;
 import org.hibernate.graph.spi.AttributeNodeImplementor;
 import org.hibernate.graph.spi.GraphImplementor;
 import org.hibernate.internal.CoreMessageLogger;
-import org.hibernate.internal.FastSessionServices;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.pretty.MessageHelper;
 import org.hibernate.property.access.internal.PropertyAccessStrategyBackRefImpl;
 import org.hibernate.proxy.HibernateProxy;
 import org.hibernate.stat.internal.StatsHelper;
 import org.hibernate.stat.spi.StatisticsImplementor;
+import org.hibernate.type.EntityType;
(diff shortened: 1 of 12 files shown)


hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
```diff
@@ -229,18 +248,36 @@ else if ( value != PropertyAccessStrategyBackRefImpl.UNKNOWN ) {
...
 		}
+	}
+
+	public static void initializeEntityFromEntityEntryLoadedState(
+			final Object entity,
+			final EntityEntry entityEntry,
+			final boolean readOnly,
+			final SharedSessionContractImplementor session,
+			final PreLoadEvent preLoadEvent,
+			final Iterable<PreLoadEventListener> preLoadEventListeners) throws HibernateException {
```
NathanQingyangXu, line 273: If this method is exposed to the public, we might change the last parameter type to `Iterable<? extends PreloadEventListener>` to be more robust. Not 100% sure, though.

PR #3441, opened 2020-06-19
HHH-14077  CVE-2019-14900 SQL injection issue using JPA Criteria API

dreab8:
https://hibernate.atlassian.net/browse/HHH-14077

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/sql/internal/JdbcLiteralFormatterCharacterData.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/sql/internal/JdbcLiteralFormatterCharacterData.java
@@ -19,6 +19,8 @@ import org.hibernate.type.descriptor.sql.spi.BasicJdbcLiteralFormatter;
  * @author Steve Ebersole
  */
 public class JdbcLiteralFormatterCharacterData extends BasicJdbcLiteralFormatter {
+	public static final String NATIONALIZED_PREFIX = "n";
+
 	private final boolean isNationalized;
 
 	public JdbcLiteralFormatterCharacterData(JavaTypeDescriptor javaTypeDescriptor) {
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/type/descriptor/sql/internal/JdbcLiteralFormatterCharacterData.java
```diff
@@ -34,12 +37,15 @@ public JdbcLiteralFormatterCharacterData(JavaTypeDescriptor javaTypeDescriptor,
...
+
 		if ( isNationalized ) {
 			// is there a standardized form for n-string literals?  This is the SQL Server syntax for sure
-			return String.format( Locale.ROOT, "n'%s'", literalValue );
+			return String.format( Locale.ROOT, NATIONALIZED_FORMAT, inlineLiteral );
+
 		}
 		else {
-			return String.format( Locale.ROOT, "'%s'", literalValue );
+			return String.format( Locale.ROOT, FORMAT, inlineLiteral );
```
NathanQingyangXu, line 48: Any difference between
```
String.format( Locale.ROOT, "%s", inlineLiteral );
```
and
```
inlineLiteral
```

A little bit curious for I think this might be one of my knowledge gaps
```

Comment (dreab8) [reply hibernate-core/src/main/java/org/hibernate/type/descriptor/sql/internal/JdbcLiteralFormatterCharacterData.java:48]: good point, I think we can avoid `return String.format( Locale.ROOT, FORMAT, inlineLiteral );` and just return `inlineLiteral`

File: hibernate-core/src/main/java/org/hibernate/type/descriptor/sql/internal/JdbcLiteralFormatterCharacterData.java [hunk 2]
```diff
@@ -34,12 +36,14 @@ public JdbcLiteralFormatterCharacterData(JavaTypeDescriptor javaTypeDescriptor,
...
 
+		final String inlineLiteral = dialect.inlineLiteral( literalValue );
+
 		if ( isNationalized ) {
 			// is there a standardized form for n-string literals?  This is the SQL Server syntax for sure
-			return String.format( Locale.ROOT, "n'%s'", literalValue );
-		}
-		else {
-			return String.format( Locale.ROOT, "'%s'", literalValue );
+			return String.format( Locale.ROOT, NATIONALIZED_FORMAT, inlineLiteral );
```
NathanQingyangXu, line 43: We can simplify the above without the usage of String.format as well. String.format is pretty slow!

hibernate-core/src/main/java/org/hibernate/type/descriptor/sql/internal/JdbcLiteralFormatterCharacterData.java
```diff
@@ -19,6 +19,8 @@
  * @author Steve Ebersole
  */
 public class JdbcLiteralFormatterCharacterData extends BasicJdbcLiteralFormatter {
+	public static final String NATIONALIZED_FORMAT = "n%s";
```
NathanQingyangXu, line 22: I just wanna double check `n'%s'` was discarded as previously intentionally? Seems a little bit weird by appending `n` only.

PR #3442, opened 2020-06-19
HHH-14080 Informix union all support

VladoKuruc:
All *UnionSubclassTest tests pass locally


gavinking:
This looks fine to me though I can't test it because I don't have Informix. The change would need to be replicated to H6.

hibernate-core/src/main/java/org/hibernate/dialect/InformixDialect.java
```diff
@@ -312,4 +312,19 @@ public IdentityColumnSupport getIdentityColumnSupport() {
...
 		return bool ? "'t'" : "'f'";
 	}
+	@Override
+	public boolean supportsUnionAll() {
+		return true;
+	}
+
+	@Override
+	public String getSelectClauseNullString(int sqlType) {
+		String typeName = getTypeName( sqlType, 1, 1, 0 );
```
gavinking, line 322: Would it be better to use `getCastTypeName()` here?

VladoKuruc: getCastTypeName() has the same result after trim off. The original implementation is the same as in PostgreSQL81Dialect.

VladoKuruc:
I closed this pull request because I created a new one #8027 based on the current version in the main branch.

PR #3450, opened 2020-07-03
StaxReader removal

Sanne:
See https://hibernate.atlassian.net/browse/HHH-14096

I think recommend backporting this to 5.3 as well, but as usual that's for @gbadner to decide.


hibernate-core/src/test/java/org/hibernate/internal/util/xml/XMLHelper.java
```diff
@@ -16,10 +16,11 @@
...
  *
- * @deprecated Currently only used for integration with HCANN.  The rest of Hibernate uses StAX now
- * for XML processing.  See {@link org.hibernate.boot.jaxb.internal.stax}
+ * This was part of Hibernate ORM core, but moved into the testsuite helpers to not expose
+ * access to the dom4j types. It's also used by Hibernate Envers, so we will need two copies
+ * until Envers is able to remove its reliance on dom4j.
+ * The rest of Hibernate uses StAX now for XML processing.  See {@link org.hibernate.boot.jaxb.internal.stax}
  */
-@Deprecated
 public final class XMLHelper {
```
gbadner, line 24: @Sanne, I'm surprised to see `@Deprecated` removed.

Shouldn't this class be deleted?

Sanne: I moved it into the tests folder: it is removed from the project, but still used in some tests.

gbadner: Oh, ok; I didn't realize.

Sanne:
thanks @gbadner , I've merged this in both `master` and `5.4`. Up to you to decide if and when to backport to `5.3` too.

PR #3451, opened 2020-07-03
wip/6.0 Enrich CollectionHelper by adding LinkedHashMap and LinkedHashSet processing

NathanQingyangXu:
This is the follow up of https://github.com/hibernate/hibernate-orm/pull/3279. In recent code review, it seems we need to tackle the common confusion in case of `LinkedHashMap` and `LinkedHashSet` counterparts.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/StandardOrderedMapSemantics.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/StandardOrderedMapSemantics.java
@@ -12,6 +12,7 @@ import java.util.Map;
 
 import org.hibernate.collection.spi.PersistentCollection;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.internal.util.collections.CollectionHelper;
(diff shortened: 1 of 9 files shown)


NathanQingyangXu:
> I don't understand the comment "confusion in case of LinkedHashMap and LinkedHashSet counterparts". Do you mean creating them with size/loadFactor? Aside from that (maybe, its hard to tell from the structure of the change) I do not see any real functional change.
> 
> That's simply a documentation / historical concern. Adding these new methods to CollectionHelper seems like a good change, so going to approve and merge it

It is functional change. So `LinkedHashMap` and `LinkedHashSet` extends from `HashMap` and `HashSet` respectively, so the following misuse to create a `LinkedHashMap` of expected size of 10
```
Map<String, String> linkedMap = new LinkedHashMap<>( 10 );
```
is similarly wrong as in
```
Map<String, String> hashMap = new HashMap<>( 10 );
``` 
As you can see, there are quite some misuse as above in our codebase and that motivates me providing a util method to avoid this.

PR #3452, opened 2020-07-04
HHH-14097 Add test for issue

dreab8:
https://hibernate.atlassian.net/browse/HHH-14097

PR #3453, opened 2020-07-04
HHH-14097 Fix redundant SQLs issue for fetch entity graph

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14097

The root cause is we lack fetch entity graph validation. For the initial multiple hydrated objects, only the first one is applicable to root entity graph.

Imported andrea's testing code from https://github.com/hibernate/hibernate-orm/pull/3452. Tons of thanks to him!

Remember only the testing case is needed to be merged into v6 for entity graph implementation is incompatible between v5 and v6 and the code changes except the testing case are not needed in v6 (actually I've carefully deleted the v5 implementation in v6 previously).

### DIFF
--- a/documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc
+++ b/documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc
@@ -188,8 +188,15 @@ In both cases, this resolves to exactly one database query to get all that infor
 [[fetching-strategies-dynamic-fetching-entity-graph]]
 === Dynamic fetching via JPA entity graph
 
-JPA 2.1 introduced entity graphs so the application developer has more control over fetch plans.
+JPA 2.1 introduced ``entity graph`` so the application developer has more control over fetch plans. It has two modes to choose from:
 
+fetch graph:::
+In this case, all attributes specified in the entity graph will be treated as FetchType.EAGER, and all attributes not specified will *ALWAYS* be treated as FetchType.LAZY.
+
+load graph:::
+In this case, all attributes specified in the entity graph will be treated as FetchType.EAGER, but attributes not specified use their static mapping specification.
+
+Below is an `fetch graph` dynamic fetching example:
 [[fetching-strategies-dynamic-fetching-entity-graph-example]]
 .Fetch graph example
 ====
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
@@ -206,61 +206,90 @@ public final class TwoPhaseLoad {
 		String entityName = persister.getEntityName();
 		String[] propertyNames = persister.getPropertyNames();
 		final Type[] types = persister.getPropertyTypes();
-		
-		final GraphImplementor<?> fetchGraphContext = session.getFetchGraphLoadContext();
-		
-		for ( int i = 0; i < hydratedState.length; i++ ) {
-			final Object value = hydratedState[i];
-			if ( debugEnabled ) {
-				LOG.debugf(
-					"Processing attribute `%s` : value = %s",
-					propertyNames[i],
-					value == LazyPropertyInitializer.UNFETCHED_PROPERTY ? "<un-fetched>" : value == PropertyAccessStrategyBackRefImpl.UNKNOWN ? "<unknown>" : value
-				);
-			}
 
-			if ( value == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
+		GraphImplementor fetchGraphContext = session.getFetchGraphLoadContext();
+		if ( fetchGraphContext != null && !fetchGraphContext.appliesTo( entity.getClass() ) ) {
+			LOG.warnf( "Entity graph specified is not applicable to the entity [%s]. Ignored.", entity);
+			fetchGraphContext = null;
+			session.setFetchGraphLoadContext( null );
+		}
+
+		try {
+			for ( int i = 0; i < hydratedState.length; i++ ) {
+				final Object value = hydratedState[i];
 				if ( debugEnabled ) {
-					LOG.debugf( "Resolving <un-fetched> attribute : `%s`", propertyNames[i] );
+					LOG.debugf(
+							"Processing attribute `%s` : value = %s",
+							propertyNames[i],
+							value == LazyPropertyInitializer.UNFETCHED_PROPERTY ?
+									"<un-fetched>" :
+									value == PropertyAccessStrategyBackRefImpl.UNKNOWN ? "<unknown>" : value
+					);
 				}
 
-				// IMPLEMENTATION NOTE: This is a lazy property on a bytecode-enhanced entity.
-				// hydratedState[i] needs to remain LazyPropertyInitializer.UNFETCHED_PROPERTY so that
-				// setPropertyValues() below (ultimately AbstractEntityTuplizer#setPropertyValues) works properly
-				// No resolution is necessary, unless the lazy property is a collection.
-				if ( types[i].isCollectionType() ) {
-					// IMPLEMENTATION NOTE: this is a lazy collection property on a bytecode-enhanced entity.
-					// HHH-10989: We need to resolve the collection so that a CollectionReference is added to StatefulPersistentContext.
-					// As mentioned above, hydratedState[i] needs to remain LazyPropertyInitializer.UNFETCHED_PROPERTY
-					// so do not assign the resolved, uninitialized PersistentCollection back to hydratedState[i].
-					Boolean overridingEager = getOverridingEager( session, entityName, propertyNames[i], types[i], debugEnabled );
-					types[i].resolve( value, session, entity, overridingEager );
+				if ( value == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
+					if ( debugEnabled ) {
+						LOG.debugf( "Resolving <un-fetched> attribute : `%s`", propertyNames[i] );
+					}
+
+					// IMPLEMENTATION NOTE: This is a lazy property on a bytecode-enhanced entity.
+					// hydratedState[i] needs to remain LazyPropertyInitializer.UNFETCHED_PROPERTY so that
+					// setPropertyValues() below (ultimately AbstractEntityTuplizer#setPropertyValues) works properly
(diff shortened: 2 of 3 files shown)


documentation/src/test/java/org/hibernate/userguide/fetching/GraphFetchingTest.java
```diff
@@ -89,7 +85,7 @@ public void test() {
 				Employee.class,
 				userId,
 				Collections.singletonMap(
-					"javax.persistence.fetchgraph",
+					GraphSemantic.FETCH.getJpaHintName(),
```
Sanne, line 88: Since this is an example in documentation, I think it would be clearer to just use the string (as it was)?

If you really want to use a constant in documentation, we should also show the import statement, or use the fully qualified classname.

NathanQingyangXu: Thanks. Didn't realize that. Reverted back the changes to the class. Thanks.

hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
```diff
@@ -206,62 +206,91 @@ public static void initializeEntityEntryLoadedState(
...
 
-			if ( session.getFetchGraphLoadContext() != fetchGraphContext ) {
 				session.setFetchGraphLoadContext( fetchGraphContext );
 			}
 		}
+		finally {
+			// HHH-14097
+			// Fetch entity graph should be applied only once on top level (for root hydrated object)
+			// e.g., see org.hibernate.loader.Loader for details
+			session.setFetchGraphLoadContext( null );
```
Sanne, line 292: I'm not sure about this, are we entirely sure that the fetch graph is no longer needed at this point?
Wouldn't it be possible for the fetch grapth to need to span multiple load events?

IF this is correct, should line 285 be removed as it seems redundant?

NathanQingyangXu:
in that case, the context will be set to null first then set to the upper
level context in line #285. I added 'on top level' for the comment already.
yeah, it is confusing. implementing fetch graph is sort if hacky in v5.

hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
NathanQingyangXu: For non-toplevel traversal, Line #285 will take effect to backtrack to upper level after the above statement. The above statement will impact only top level (I made it explicitly by including "on top level" in the comments above). I admit the code change is difficult to understand for basically 'fetch graph' implementing is sort of based on hack (existing v5 infrastructure makes it difficult to implement elegantly), but v6 based on path retrieval makes the implementation much easier and more elegant. After this PR gets merged, I'll port it into v6 to see whether the issue still remains.

dreab8:
Hi @NathanQingyangXu  still not sure we really need `session.setFetchGraphLoadContext( null );` as discussed privately I tried to add the kind of test you think may need this call
```
@Test
	public void testQueryByIdWithLoadGraph3() {
		Statistics statistics = entityManagerFactory().unwrap( SessionFactory.class ).getStatistics();
		statistics.clear();
		doInJPA(
				this::entityManagerFactory, entityManager -> {
					EntityGraph<AEntity> entityGraph = entityManager.createEntityGraph( AEntity.class );
					entityGraph.addAttributeNodes( "cList" );
					entityGraph.addSubgraph( "cList" ).addAttributeNodes( "a" );

					TypedQuery<AEntity> query = entityManager.createQuery(
							"select a from AEntity as a where a.id = :aid ",
							AEntity.class
					);
					query.setHint( GraphSemantic.LOAD.getJpaHintName(), entityGraph );
					query.setParameter( "aid", 1 );

					AEntity cEntity = query.getSingleResult();

					assertTrue( Hibernate.isInitialized( cEntity.getcList() ) );
					assertTrue( Hibernate.isInitialized( cEntity.getcList().get( 0 ) ) );

					assertEquals( 1L, statistics.getPrepareStatementCount() );
				} );
	}
```
if the test is correct, it passes with just
```
if ( fetchGraphContext != null && !fetchGraphContext.appliesTo( entity.getClass() ) ) {
			LOG.warnf( "Entity graph specified is not applicable to the entity [%s]. Ignored.", entity);
			fetchGraphContext = null;
}
```

NathanQingyangXu:
> Hi @NathanQingyangXu still not sure we really need `session.setFetchGraphLoadContext( null );` as discussed privately I tried to add the kind of test you think may need this call
> 
> ```
> @Test
> 	public void testQueryByIdWithLoadGraph3() {
> 		Statistics statistics = entityManagerFactory().unwrap( SessionFactory.class ).getStatistics();
> 		statistics.clear();
> 		doInJPA(
> 				this::entityManagerFactory, entityManager -> {
> 					EntityGraph<AEntity> entityGraph = entityManager.createEntityGraph( AEntity.class );
> 					entityGraph.addAttributeNodes( "cList" );
> 					entityGraph.addSubgraph( "cList" ).addAttributeNodes( "a" );
> 
> 					TypedQuery<AEntity> query = entityManager.createQuery(
> 							"select a from AEntity as a where a.id = :aid ",
> 							AEntity.class
> 					);
> 					query.setHint( GraphSemantic.LOAD.getJpaHintName(), entityGraph );
> 					query.setParameter( "aid", 1 );
> 
> 					AEntity cEntity = query.getSingleResult();
> 
> 					assertTrue( Hibernate.isInitialized( cEntity.getcList() ) );
> 					assertTrue( Hibernate.isInitialized( cEntity.getcList().get( 0 ) ) );
> 
> 					assertEquals( 1L, statistics.getPrepareStatementCount() );
> 				} );
> 	}
> ```
> 
> if the test is correct, it passes with just
> 
> ```
> if ( fetchGraphContext != null && !fetchGraphContext.appliesTo( entity.getClass() ) ) {
> 			LOG.warnf( "Entity graph specified is not applicable to the entity [%s]. Ignored.", entity);
> 			fetchGraphContext = null;
> }
> ```

I added a new testing case to showcase why `session.setFetchGraphLoadContext( null )` is needed. It is hard to explain and even confusing, but it ensures that entity graph is disabled for scenario wherein it is not applicable.

dreab8:
Thanks @NathanQingyangXu for the additional test, may be I'm missing something but I tried it with just 
```
GraphImplementor fetchGraphContext = session.getFetchGraphLoadContext();
		if ( fetchGraphContext != null && !fetchGraphContext.appliesTo( entity.getClass() ) ) {
			LOG.warnf( "Entity graph specified is not applicable to the entity [%s]. Ignored.", entity);
			fetchGraphContext = null;

		}
```
and it passes.

NathanQingyangXu:
> Thanks @NathanQingyangXu for the additional test, may be I'm missing something but I tried it with just
> 
> ```
> GraphImplementor fetchGraphContext = session.getFetchGraphLoadContext();
> 		if ( fetchGraphContext != null && !fetchGraphContext.appliesTo( entity.getClass() ) ) {
> 			LOG.warnf( "Entity graph specified is not applicable to the entity [%s]. Ignored.", entity);
> 			fetchGraphContext = null;
> 
> 		}
> ```
> 
> and it passes.

Sorry for my premature pushing. Now I updated the testing case and I am confident it should be fine.

documentation/src/main/asciidoc/userguide/chapters/fetching/Fetching.adoc
```diff
@@ -188,8 +188,15 @@ In both cases, this resolves to exactly one database query to get all that infor
...
-JPA 2.1 introduced entity graphs so the application developer has more control over fetch plans.
+JPA 2.1 introduced ``entity graph`` so the application developer has more control over fetch plans. It has two modes to choose from:
 
+fetch graph:::
+In this case, all attributes specified in the entity graph will be treated as FetchType.EAGER, and all attributes not specified will *ALWAYS* be treated as FetchType.LAZY.
+
+load graph:::
+In this case, all attributes specified in the entity graph will be treated as FetchType.EAGER, but attributes not specified use their static mapping specification.
+
+Below is an `fetch graph` dynamic fetching example:
```
NathanQingyangXu, line 199: The above content was copied from v6's counterpart. Thought the basic explanation of the two types of entity graph modes is supposed to exist in v5 as well.
Given the content has existed in v6 already, we don't need to bother merging this part to v6 as well.

hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
```diff
@@ -206,62 +206,91 @@ public static void initializeEntityEntryLoadedState(
...
-					value == LazyPropertyInitializer.UNFETCHED_PROPERTY ? "<un-fetched>" : value == PropertyAccessStrategyBackRefImpl.UNKNOWN ? "<unknown>" : value
-				);
-			}
 
-			if ( value == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
+		GraphImplementor fetchGraphContext = session.getFetchGraphLoadContext();
+		if ( fetchGraphContext != null && !fetchGraphContext.appliesTo( entity.getClass() ) ) {
+			LOG.warnf( "Entity graph specified is not applicable to the entity [%s]. Ignored.", entity);
+			fetchGraphContext = null;
+			session.setFetchGraphLoadContext( null );
```
dreab8, line 214: just a minor but I think the `session.setFetchGraphLoadContext( null );` is redundant

NathanQingyangXu: I think it is necessary because without setting it, the downstream will still try to locate some property as per the context grabbed from `session.getFetchGraphLoadContext()`, not from the local variable of `fetchGraphContext`. See the `org.hibernate.engine.internal.TwoPhaseLoad#isEagerFetchGraph(SharedSessionContractImplementor session, String associationName, Type associationType)` for details


I am sorry the current fetch entity graph implementation is so difficult to understand (and maintain). Luckily its status is improved much more in v6.

hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
NathanQingyangXu: I doubled checked and found the backtracking is working as expected (without it our graph testing cases would fail). Could you take a new look at this PR? Thanks.

Sanne:
thanks! Merging

PR #3456, opened 2020-07-09
HHH-14093 - Optimize PersistenceContext for better contains performance

Leuteris:
I've refactored StatefulPersistenceContext#registerInsertedKey and StatefulPersistenceContext#wasInsertedDuringTransaction 
to use HashSet instead of ArrayList.

https://hibernate.atlassian.net/projects/HHH/issues/HHH-14093


Sanne:
Very interesting, thanks @Leuteris . Just curious, how did you spot this? Did you see any problem at runtime with this?

I'm curious as we've been running many benchmarks (while profiling) and I've never observed this.

Sanne:
ah, sorry for my question. I see that was explained in the related JIRA.

Many thanks, I've merged it in master and in 5.4 !

PR #3457, opened 2020-07-09
HHH 13214 DML batch delete re-firing SQL from previous calls

dreab8:
https://hibernate.atlassian.net/browse/HHH-13214

this is an alternative solution of https://github.com/hibernate/hibernate-orm/pull/3348

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/spi/id/inline/AbstractInlineIdsDeleteHandlerImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/spi/id/inline/AbstractInlineIdsDeleteHandlerImpl.java
@@ -18,6 +18,7 @@ import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.hql.internal.ast.HqlSqlWalker;
 import org.hibernate.hql.spi.id.MultiTableBulkIdStrategy;
+import org.hibernate.internal.util.collections.ArrayHelper;
 import org.hibernate.persister.collection.AbstractCollectionPersister;
 import org.hibernate.persister.entity.Queryable;
 import org.hibernate.sql.Delete;
@@ -33,7 +34,7 @@ public abstract class AbstractInlineIdsDeleteHandlerImpl
 		extends AbstractInlineIdsBulkIdHandler
 		implements MultiTableBulkIdStrategy.DeleteHandler {
 
-	private List<String> deletes;
+	private List<String> deletes = new ArrayList<>();
 
 	public AbstractInlineIdsDeleteHandlerImpl(
 			SessionFactoryImplementor factory,
@@ -44,7 +45,7 @@ public abstract class AbstractInlineIdsDeleteHandlerImpl
 	@Override
 	public String[] getSqlStatements() {
 		if ( deletes.isEmpty() ) {
-			return new String[0];
+			return ArrayHelper.EMPTY_STRING_ARRAY;
 		}
 		return deletes.toArray( new String[deletes.size()] );
 	}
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/hql/spi/id/inline/AbstractInlineIdsDeleteHandlerImpl.java
```diff
@@ -42,6 +43,9 @@ public AbstractInlineIdsDeleteHandlerImpl(
 
 	@Override
 	public String[] getSqlStatements() {
+		if ( deletes.isEmpty() ) {
+			return new String[0];
```
NathanQingyangXu, line 47: Do we allow the returned value to be modifiable? If not, returning `Collections.emptyList()` would be more efficient and less confusing. The same suggestion applies for the later empty collection returned. Minor compared to the issue solved in this PR.

hibernate-core/src/main/java/org/hibernate/resource/beans/container/internal/JpaCompliantLifecycleStrategy.java
```diff
@@ -125,7 +125,7 @@ public void initialize() {
 				throw e;
 			}
 			catch (Exception e) {
-				log.debugf( "Error resolving CDI bean [%s] - using fallback" );
+				log.debugf( "Error resolving CDI bean [%s] - using fallback", beanType );
```
NathanQingyangXu, line 128: It is really annoying for such String format parameter mismatch issue. We encountered every now and then. Sometimes it gets exposed easily, sometimes the issue is hidden simply because the statement is seldom run.

hibernate-core/src/main/java/org/hibernate/hql/spi/id/inline/AbstractInlineIdsDeleteHandlerImpl.java
```diff
@@ -42,6 +43,9 @@ public AbstractInlineIdsDeleteHandlerImpl(
 
 	@Override
 	public String[] getSqlStatements() {
+		if ( deletes.isEmpty() ) {
```
NathanQingyangXu, line 46: Is there possibility that `deletes` is null here? Then a NPE would be thrown. IMHO, lazy initialization might be overused in Hibernate core codebase.

dreab8: solved thanks @NathanQingyangXu

hibernate-core/src/main/java/org/hibernate/hql/spi/id/inline/AbstractInlineIdsDeleteHandlerImpl.java
dreab8: we method return type is an array

NathanQingyangXu: my fault. we might use `ArrayHelper.EMPTY_STRING_ARRAY`, though not a big deal.

dreab8: done and thanks a lot for the review :)

Sanne: If we're nitpickin for optimisations, I might add that it's even faster to remove the `if` check  ;)

Sanne:
Looks great. Thanks @dreab8 and @NathanQingyangXu !

hibernate-core/src/main/java/org/hibernate/hql/spi/id/inline/AbstractInlineIdsDeleteHandlerImpl.java
Sanne: I'm joking, but let me clarify why: it's because I would not expect the `deleted` list to ever be empty, so the check is redundant.

PR #3458, opened 2020-07-11
Update basic-enums-Enumerated-ordinal-persistence-example.sql

Vishavjeet6:
as stated in https://docs.jboss.org/hibernate/orm/5.2/userguide/html_single/Hibernate_User_Guide.html
(2.3.7 after example 18  )that
 0
For the LAND_LINE enum
1
For the MOBILE enum
hence it should be 1


NathanQingyangXu:
Seems a good catch! @dreab8 do we need to create a ticket first for v5 bugfix?

PR #3460, opened 2020-07-12
Wip 6.0 Inheritance 

dreab8:
The PR re-enables a bunch on inheritance tests fixing issues mostly related with polymorphic associations.


hibernate-core/src/main/java/org/hibernate/engine/internal/Cascade.java
```diff
@@ -97,6 +97,7 @@ public static void cascade(
 						hasUninitializedLazyProperties &&
 						!persister.getBytecodeEnhancementMetadata().isAttributeLoaded( parent, propertyName );
 
+				Type type = types[i];
```
NathanQingyangXu, line 100: seems we can use `final` here for consistency reason

hibernate-core/src/main/java/org/hibernate/metamodel/internal/AbstractPojoInstantiator.java
```diff
@@ -34,4 +34,10 @@ public boolean isAbstract() {
 	public boolean isInstance(Object object, SessionFactoryImplementor sessionFactory) {
 		return mappedPojoClass.isInstance( object );
 	}
+
+	@Override
+	public boolean isSameClass(Object object, SessionFactoryImplementor sessionFactory) {
+		return object.getClass() == mappedPojoClass;
```
NathanQingyangXu, line 40: Should we take consideration the possibility that `object` could be null? This is a public method so it should be function appropriately for any external input. Nullness processing is always an gray area in Hibernate's codebase, IMHO.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/SimpleForeignKeyDescriptor.java
```diff
@@ -184,7 +180,6 @@ public Predicate generateJoinPredicate(
 			);
 		}
 		else {
-			assert rhsTableExpression.equals( keyColumnContainingTable );
```
NathanQingyangXu, line 187: One of my current ongoing entity graph testing case failed because of the above statement. Glad it has been removed.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/from/UnionTableGroup.java
```diff
@@ -63,24 +65,37 @@ public LockMode getLockMode() {
...
 	}
 
 	@Override
 	public void setTableGroupJoins(Set<TableGroupJoin> joins) {
+		if ( tableGroupJoins == null ) {
+			tableGroupJoins = new HashSet<>( joins );
+		}
+		else {
+			tableGroupJoins.addAll( joins );
+		}
```
NathanQingyangXu, line 83: seems a little bit confusing for what is done is `addAllTableGroupJoins()`, not replacing the previous content.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
```diff
@@ -117,10 +121,15 @@ protected AbstractEntityInitializer(
...
-						public Initializer resolveInitializer(NavigablePath navigablePath, Supplier<Initializer> producer) {
+						public Initializer resolveInitializer(
+								NavigablePath navigablePath,
+								ModelPart fetchedModelPart,
+								Supplier<Initializer> producer) {
 							for ( int i = 0; i < identifierInitializers.size(); i++ ) {
 								final Initializer existing = identifierInitializers.get( i );
-								if ( existing.getNavigablePath().equals( navigablePath ) ) {
+								if ( existing.getNavigablePath().equals( navigablePath )
+										&& fetchedModelPart.getNavigableRole().equals(
```
NathanQingyangXu, line 131: Just noticed some format inconsistency. Sometimes we split BEFORE boolean operator, sometimes we split AFTER it.

hibernate-core/src/test/java/org/hibernate/test/annotations/inheritance/joined/JoinedSubclassAndSecondaryTable.java
```diff
@@ -6,100 +6,121 @@
...
+					rowCount = getTableRowCount( session, "POOL_ADDRESS" );
+					assertEquals(
+
+							1,
+							rowCount,
+							"Now we should have a row in the pool address table "
+
+					);
+					assertNotNull( sp2.getAddress() );
+					assertEquals( sp2.getAddress().getAddress(), "Park Avenue" );
```
NathanQingyangXu, line 82: seems `Park Avenue` is the expected value and should be put as the first parameter. Not a big deal for sure.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/from/UnionTableGroup.java
dreab8: Fixed, thanks Nathan

hibernate-core/src/test/java/org/hibernate/test/annotations/inheritance/joined/JoinedSubclassAndSecondaryTable.java
dreab8: Fixed, thanks Nathan

hibernate-core/src/main/java/org/hibernate/metamodel/internal/AbstractPojoInstantiator.java
dreab8: it is an internal class, plus returning `false` seems wrong to me so in any case if `object` is null an exception should be raised.

dreab8:
merged

PR #3464, opened 2020-07-15
HHH-14108 - PADDED and DYNAMIC Batch loading strategies not compatible with the Enhanced Proxy feature

dreab8:
https://hibernate.atlassian.net/browse/HHH-14108

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
@@ -75,7 +75,6 @@ import org.hibernate.internal.FetchingScrollableResultsImpl;
 import org.hibernate.internal.ScrollableResultsImpl;
 import org.hibernate.internal.util.StringHelper;
 import org.hibernate.internal.util.collections.CollectionHelper;
-import org.hibernate.loader.entity.CascadeEntityLoader;
 import org.hibernate.loader.spi.AfterLoadAction;
 import org.hibernate.persister.collection.CollectionPersister;
 import org.hibernate.persister.entity.EntityPersister;
@@ -116,14 +115,12 @@ public abstract class Loader {
 	private final SessionFactoryImplementor factory;
 	private volatile ColumnNameCache columnNameCache;
 
-	private final boolean referenceCachingEnabled;
 	private final boolean enhancementAsProxyEnabled;
 
 	private boolean isJdbc4 = true;
 
 	public Loader(SessionFactoryImplementor factory) {
 		this.factory = factory;
-		this.referenceCachingEnabled = factory.getSessionFactoryOptions().isDirectReferenceCacheEntriesEnabled();
 		this.enhancementAsProxyEnabled = factory.getSessionFactoryOptions().isEnhancementAsProxyEnabled();
 	}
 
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/loader/Loader.java
```diff
@@ -1666,50 +1663,46 @@ protected void instanceAlreadyLoaded(
...
-									rs,
-									i,
-									persister,
-									key.getIdentifier(),
-									session
-							),
-							object,
-							requestedLockMode
-					);
+		if (persister.getBytecodeEnhancementMetadata().isEnhancedForLazyLoading() && enhancementAsProxyEnabled) {
```
NathanQingyangXu, line 1666: It might be considerable approach to put `enhancementAsProxyEnabled` to left of `&&` for its performance cost is lower and we might end up with performance gain due to short-circuiting.

hibernate-core/src/main/java/org/hibernate/loader/Loader.java
```diff
@@ -1610,7 +1607,7 @@ private void checkVersion(
 			else {
 				//If the object is already loaded, return the loaded one
 				object = session.getEntityUsingInterceptor( key );
-				if ( object != null ) {
+				if (object != null) {
```
NathanQingyangXu, line 1610: It is a pattern to get rid of space char uses previously advised. Is that a new format pattern to advocate then?

dreab8: No it is not, I'm going to fix it, thanks

hibernate-core/src/main/java/org/hibernate/loader/Loader.java
dreab8: I'll fix it thanks

hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/batch/DynamicBatchingTest.java
```diff
@@ -0,0 +1,56 @@
...
+public class DynamicBatchingTest extends AbstractBatchingTest {
+
+	@Override
+	protected void configureStandardServiceRegistryBuilder(StandardServiceRegistryBuilder ssrb) {
+		super.configureStandardServiceRegistryBuilder( ssrb );
+		ssrb.applySetting( AvailableSettings.BATCH_FETCH_STYLE, BatchFetchStyle.DYNAMIC.toString() );
+	}
+
+	@Test
+	public void testLoadParent() {
```
Sanne, line 36: Looks like the body of the test is exactly the same in both subclasses. Could be moved to the parent class to make this more clear?

dreab8: done thanks

hibernate-core/src/main/java/org/hibernate/loader/Loader.java
```diff
@@ -1666,38 +1663,34 @@ protected void instanceAlreadyLoaded(
 			);
 		}
 
-		if ( persister.getBytecodeEnhancementMetadata().isEnhancedForLazyLoading() && enhancementAsProxyEnabled ) {
-			if ( "merge".equals( session.getLoadQueryInfluencers().getInternalFetchProfile() ) ) {
```
Sanne, line 1670: This looks a bit puzzling. Do you know why it was limited to "merge" fetch profiles? 

Now you're expanding the set of operations to all, but the test just wants the load. I see we only use this for "merge" and "refresh" BTW... I suppose there's not  something special about "refresh" that we need to take into account during this load opertion?

dreab8: I do not remember the reason why it was limited to the `merge` fetch profile, for the test case the fetch profile value is null but we need to hydrate the entity state and I think anything special is needed for "refresh" if I'm correct it should not reach this point in case of `refresh` for an enhanced proxy (see `DefaultRefreshEventListener#onRefresh` lines 71 to75 )

Sanne: great, thanks

Sanne:
merged in master and backported on 5.4. Thanks!

PR #3467, opened 2020-07-17
Wip/6.0 Dependant Value resolution

dreab8:


hibernate-core/src/main/java/org/hibernate/mapping/DependantValue.java
```diff
@@ -64,4 +68,19 @@ public boolean isSame(DependantValue other) {
...
+		return true;
+	}
+
+	@Override
+	public BasicValue.Resolution<?> resolve() {
+		if ( wrappedValue instanceof BasicValue ) {
+			return ( (BasicValue) wrappedValue ).resolve();
+		}
+		// not sure it is ever possible
+		throw new UnsupportedOperationException("Trying to resolve the wrapped value but it is non a BasicValue");
```
NathanQingyangXu, line 83: `it is non a` => `it is not a`?

PR #3468, opened 2020-07-17
HHH-14109 use 'in expression count limit' if calculated count is to g…

OBreidenbach:
…reather as 'in expression count limit'
[HHH-14109](https://hibernate.atlassian.net/projects/HHH/issues/HHH-14109)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingsImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingsImpl.java
@@ -538,8 +538,8 @@ public class QueryParameterBindingsImpl implements QueryParameterBindings {
 			if ( inClauseParameterPaddingEnabled ) {
 				int bindValuePaddingCount = MathHelper.ceilingPowerOfTwo( bindValueCount );
 
-				if (inExprLimit > 0 && bindValuePaddingCount > inExprLimit ) {
-					bindValuePaddingCount = Math.max(bindValueCount, inExprLimit);
+				if ( inExprLimit > 0 && bindValuePaddingCount > inExprLimit ) {
+					bindValuePaddingCount = inExprLimit;
 				}
 
 				if ( bindValueCount < bindValuePaddingCount ) {
--- a/hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingsImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingsImpl.java
@@ -538,7 +538,11 @@ public class QueryParameterBindingsImpl implements QueryParameterBindings {
 			if ( inClauseParameterPaddingEnabled ) {
 				int bindValuePaddingCount = MathHelper.ceilingPowerOfTwo( bindValueCount );
 
-				if ( bindValueCount < bindValuePaddingCount && (inExprLimit == 0 || bindValuePaddingCount < inExprLimit) ) {
+				if (inExprLimit > 0 && bindValuePaddingCount > inExprLimit ) {
+					bindValuePaddingCount = Math.max(bindValueCount, inExprLimit);
+				}
+
+				if ( bindValueCount < bindValuePaddingCount ) {
 					bindValueMaxCount = bindValuePaddingCount;
 				}
 			}
--- a/hibernate-core/src/test/java/org/hibernate/query/MaxInExpressionParameterPaddingTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/query/MaxInExpressionParameterPaddingTest.java
@@ -92,6 +92,7 @@ public class MaxInExpressionParameterPaddingTest extends BaseEntityManagerFuncti
 		assertTrue(sqlStatementInterceptor.getSqlQueries().get( 0 ).endsWith( expectedInClause.toString() ));
 	}
 
+	@TestForIssue( jiraKey = "HHH-14109" )
 	@Test
 	public void testInClauseParameterPaddingToLimit() {
 		sqlStatementInterceptor.clear();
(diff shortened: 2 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingsImpl.java
```diff
@@ -538,7 +538,11 @@ public String expandListValuedParameters(String queryString, SharedSessionContra
 			if ( inClauseParameterPaddingEnabled ) {
 				int bindValuePaddingCount = MathHelper.ceilingPowerOfTwo( bindValueCount );
 
-				if ( bindValueCount < bindValuePaddingCount && (inExprLimit == 0 || bindValuePaddingCount < inExprLimit) ) {
+				if (inExprLimit > 0 && bindValuePaddingCount > inExprLimit ) {
+					bindValuePaddingCount = Math.max(bindValueCount, inExprLimit);
```
NathanQingyangXu, line 542: Could we simply the above with simply
```
bindValuePaddingCount = inExprLimit;
```
?

OBreidenbach: I don't think so. If bindValueCount is greather than inExprLimit we would ignore all values above the inExprLimit. 
We would not do log.tooManyInExpressions(...) and we would not add the values  above the inExprLimit to the prepared statement. So the data base would not throw an exception and we had silently ignored some values of the collection.

Excample:
Oracle has a limit of 1000. If I have a collection of 1002 values i cannot ignore the last two ones.

NathanQingyangXu: If I didn't misunderstand, if `bindValuePaddingCount` is set to 1000 as opposed to 1002, the later statements wouldn't change the final result for you have other logic to ensure that. My point is concern separation and duplicated logic makes the code unnecessarily difficult to understand. IMHO, `bindValuePaddingCount ` would be great to only concern padding issue and the `Math.max()` logic is redundant and might hurt code readability a little bit.
Not a big deal maybe, but I do think making code easy to understand makes for a more maintainable codebase. Ignore my comment if it is not appropriate, though.

OBreidenbach: As I said, IMHO your suggestion would ignore values above the in expression limit. No warning would be logged and the database would not throw an exception, but also would not find values referenced by values above the in expression limit. So IMHO we do need Math.max (or some other code, doing the same but different) to produce a correct result.

NathanQingyangXu: Maybe I misunderstood. If db limit is 1000 and existing list is 1002, the `bindValueMaxCount` won't be changed and so the for loop later will do nothing, correct?

NathanQingyangXu: Maybe I am wrong, but the example you mentioned is not a new issue and should have been solved already (see for loop statement for details). IMHO, we only need to focus on the key issue in this ticket (taking dialect limit into consideration) and don't need to duplicate unrelated logic again, which makes the code really difficult to figure out.

OBreidenbach: To be honest, I do not understand what you want.

Your suggested simplification will lead to the failure of ignoring values above the limit boarder. So we can not do that. Do you want (me) to find  some other simplification? Should I provide a test to show the different of both solutions.
My solutions puts 1002 parameters into the prepared statement yours 1000 parameters. I think there is a different and I do not see any safety net for that.

OBreidenbach: What do you mean by 

> the example you mentioned is not a new issue and should have been solved already (see for loop statement for details). IMHO, we only need to focus on the key issue in this ticket

IMHO I am focusing on the key issue of this ticket. If we do no padding a list of values larger than the inExpressionLimit is just passed as is to the prepared statement. You say I should only pass the first 1000 values if I do the padding, that is changing the behavior of hibernate.

NathanQingyangXu:
well, I give up. You simply don't understand.

NathanQingyangXu:
maybe I misunderstood. sorry about that.

OBreidenbach:
So, how do we get to a goog end?

hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingsImpl.java
NathanQingyangXu: If I didn't misunderstand, we don't need to worry about the 2 values above 1000 (take your example) for it has been appended to the SQL already.

NathanQingyangXu: So this is my point. it seems don't need to ensure `bindValuePaddingCount` not to be less `bindValueCount` for the following statement:
```
if ( bindValueCount < bindValuePaddingCount ) {
	bindValueMaxCount = bindValuePaddingCount;
}
```
Let us suppose the `bindValuePaddingCount` ends up with 1000 and `bindValueCount` is 1002, the above statement would be skipped and no additional value is padded (existing 1002 won't be touched). It seems to make no difference between 1000 or 1002 for the end result would be the same.
That is the motivation of my simplification suggestion. I do think the current logic is convoluted and has room for improvement.

OBreidenbach: OK, got it. I did take your suggestion.

Sanne:
Many thanks for the patch @OBreidenbach , and for the review @NathanQingyangXu !
Merging in both master and 5.4

PR #3469, opened 2020-07-17
HHH-14111 Upgrade build dependencies to tests JDK15/JDK16

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14111

Once this gets merged, the JDK15 build will pass: https://ci.hibernate.org/job/hibernate-orm-personal-yoann-jdk15/4/

The JDK16 build won't pass, however: https://ci.hibernate.org/view/Personal%20jobs/job/hibernate-orm-personal-yoann-jdk16/4/
The problem seems related to gradle, which ignores the `targetCompatibility` option for `GroovyCompile` tasks with JDK16 for some reason. I couldn't find a solution nor any mention of this problem online, so I'd say let's wait for the Gradle team to at least try out JDK16 (which is quite recent).

@Sanne Would you agree to backport this to 5.4?


yrodiere:
All checks passed; merging. Will open another PR for the backport to 5.4 (and maybe 5.3).

yrodiere:
FYI I won't backport to 5.3, which is on an older version of gradle.

Sanne:
thanks @yrodiere ! Feel free to push the 5.4 branch directly, if you're confident about it and CI is happy.

PR #3472, opened 2020-07-20
Avoid logging parameters when logging statements

Sanne:
- https://hibernate.atlassian.net/browse/HHH-11082

I've been wondering if we should still give an option to log the parameters, but considering that this log message was just recently introduced: seems users have been happy enough without any log at all for many years. So I'm proposing to keep it simple.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/jdbc/Expectations.java
+++ b/hibernate-core/src/main/java/org/hibernate/jdbc/Expectations.java
@@ -45,17 +45,17 @@ public class Expectations {
 			}
 		}
 
-		public final void verifyOutcome(int rowCount, PreparedStatement statement, int batchPosition) {
+		public final void verifyOutcome(int rowCount, PreparedStatement statement, int batchPosition, String statementSQL) {
 			rowCount = determineRowCount( rowCount, statement );
 			if ( batchPosition < 0 ) {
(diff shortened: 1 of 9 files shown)


hibernate-core/src/main/java/org/hibernate/jdbc/Expectations.java
```diff
@@ -80,11 +80,11 @@ else if ( rowCount == -3 ) {
 			}
 		}
 
-		private void checkNonBatched(int rowCount, PreparedStatement statement) {
+		private void checkNonBatched(int rowCount, PreparedStatement statement, String statementSQL) {
```
dreab8, line 83: just a minor but `PreparedStatement statement` parameter can be removed, the same for the `checkBatched` method

dreab8:
I agree with keeping it simple :+1:

hibernate-core/src/main/java/org/hibernate/jdbc/Expectations.java
Sanne: Thanks, I'll do that.

Sanne:
I need to re-merge, apparently I was using the wrong JIRA id.

Sanne:
merged properly now, and backported to 5.4. Thanks!

PR #3476, opened 2020-07-22
Wip/6.0 dialect version and tests

dreab8:
This PR is mainly focused on adapting `@SkipForDialect` and `@RequiresDialect` to the new Dialects using a version attribute.


hibernate-core/src/main/java/org/hibernate/collection/spi/BagSemantics.java
```diff
@@ -0,0 +1,14 @@
...
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.collection.spi;
+
+/**
+ * @author Andrea Boriero
+ */
+public interface BagSemantics<B> extends CollectionSemantics<B> {
+}
+
```
NathanQingyangXu, line 14: seems the content from another PR is included

hibernate-core/src/main/java/org/hibernate/dialect/CUBRIDDialect.java
```diff
@@ -87,6 +88,11 @@ public CUBRIDDialect() {
 		registerKeyword( "SEARCH" );
 	}
 
+	@Override
+	public int getVersion() {
+		return 0;
+	}
+
```
NathanQingyangXu, line 95: I do think a version field is deserved to be included into Dialect base class. I even created a PR for this previously: https://github.com/hibernate/hibernate-orm/pull/3316.

hibernate-core/src/main/java/org/hibernate/collection/spi/BagSemantics.java
dreab8: Good catch @NathanQingyangXu, going to fix it. Thanks

PR #3477, opened 2020-07-22
HHH-14113 Entity Graph attribute resolution not based on selected entity

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14113

I added a warning logging to remind the user that entity graph is not applicable to the root entity, as per Benjamin M.'s suggestion.

Note: this PR should not be merged into v6 for entity graph implementation in v6 has changed.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/graphs/EntityGraphAttributeResolutionTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/graphs/EntityGraphAttributeResolutionTest.java
@@ -1,0 +1,254 @@
+package org.hibernate.jpa.test.graphs;
+
+import java.util.EnumSet;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import javax.persistence.CollectionTable;
+import javax.persistence.ElementCollection;
+import javax.persistence.Entity;
+import javax.persistence.EntityGraph;
+import javax.persistence.EnumType;
+import javax.persistence.Enumerated;
+import javax.persistence.FetchType;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToMany;
+import javax.persistence.NamedAttributeNode;
+import javax.persistence.NamedEntityGraph;
+import javax.persistence.Table;
+
+import org.hibernate.graph.GraphSemantic;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/jpa/test/graphs/EntityGraphAttributeResolutionTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+	@Test
+	public void fetchAssocWithNamedFetchGraph() {
+		EntityManager em = getOrCreateEntityManager();
+		List result = em.createQuery( "SELECT u.groups FROM User u WHERE u.id = ?1" )
+				.setParameter(1, u.getId() )
+				.setHint( GraphSemantic.FETCH.getJpaHintName(), em.getEntityGraph( Group.ENTITY_GRAPH ) )
+				.getResultList();
+
+		assertEquals( Arrays.asList( g1, g2 ), result );
```
NathanQingyangXu, line 79: I noticed there seems to be some additional issue here. So from the HQL it seems the `result` should be a list of `Set<Group>`, but the real result is a list of `Group`. Is that a bug?

sebersole:
No.  The result of a query is always a List (getResult*List*)

hibernate-core/src/test/java/org/hibernate/jpa/test/graphs/EntityGraphAttributeResolutionTest.java
beikov: You are mixing things up. When encountering a path expression that ends on a non-basic typed attribute, it is implicit joined. So the query would be equivalent to the following `SELECT g FROM User u JOIN u.groups g WHERE u.id = ?1`

NathanQingyangXu: thanks for clarification. my knowledge gap here.

NathanQingyangXu:
I think my question is about the data type of the list. Anyway, Beikov answered already.

cptwunderlich:
Thanks for the fix Nathan!
I found the discussion on Zulip regarding this topic interesting - it seems like I'm not the only one who assumed the Entity Graph would be applicable in this scenario. Since this is not (easily) possible (or even desirable), I think emitting a warning is the way to go. Quietly not applying the graph would lead to confusion and long debugging sessions. This way, the user at least knows that applying the Entity Graph is not possible.

PR #3479, opened 2020-07-23
HHH-14112 - Always join all tables from superclass entity persisters that have a WHERE condition

beikov:
https://hibernate.atlassian.net/browse/HHH-14112

PR #3481, opened 2020-07-23
HHH-14112 Invalid count SQL for JOINED inheritance and WHERE clause

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14112

This is the new PR after @beikov authorized my creating a new PR after his (https://github.com/hibernate/hibernate-orm/pull/3479). Testing cases are added cloned from the bug reporter's repo.

Thanks for @beikov 's bug fixing!

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
@@ -142,6 +142,14 @@ public class JoinProcessor implements SqlTokenTypes {
 			if ( role != null ) {
 				result.add( fromElement.getOrigin().getPropertyTableName(role.substring(role.lastIndexOf('.') + 1)) );
 			}
+			final EntityPersister entityPersister = fromElement.getEntityPersister();
+			if ( entityPersister instanceof AbstractEntityPersister ) {
+				AbstractEntityPersister aep = (AbstractEntityPersister) entityPersister;
+				while ( !aep.filterFragment( "", Collections.emptyMap() ).isEmpty() && aep.getMappedSuperclass() != null ) {
+					Collections.addAll( result, aep.getTableNames() );
+					aep = (AbstractEntityPersister) walker.getSessionFactoryHelper().findEntityPersisterByName( aep.getMappedSuperclass() );
+				}
+			}
 			AST withClauseAst = fromElement.getWithClauseAst();
 			if ( withClauseAst != null ) {
 				collectReferencedTables( new ASTIterator( withClauseAst ), result );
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/JoinProcessor.java
```diff
@@ -142,6 +142,14 @@ public static JoinType toHibernateJoinType(int astJoinType) {
...
 				result.add( fromElement.getOrigin().getPropertyTableName(role.substring(role.lastIndexOf('.') + 1)) );
 			}
+			final EntityPersister entityPersister = fromElement.getEntityPersister();
+			if ( entityPersister instanceof AbstractEntityPersister ) {
+				AbstractEntityPersister aep = (AbstractEntityPersister) entityPersister;
+				while ( !aep.filterFragment( "", Collections.emptyMap() ).isEmpty() && aep.getMappedSuperclass() != null ) {
+					Collections.addAll( result, aep.getTableNames() );
+					aep = (AbstractEntityPersister) walker.getSessionFactoryHelper().findEntityPersisterByName( aep.getMappedSuperclass() );
+				}
+			}
```
NathanQingyangXu, line 152: @beikov so I have a question for you. It seems the logic fix is to include joins whenever there is where clause, but is there such possibility that the where clause can be found in child entity per se so we don't need to bother joining? Could we tell whether the joining is really necessary?

beikov: We don't know the structure of the `@Where` condition, so we can't check if the columns that are used in it are part of a specific table. Apart from that, I am also not sure if we could theoretically skip an intermediate join when e.g. `A3 extends A2 extends A1`, I guess the join condition to for adding A1 is based on columns of A2, so even if A2 is not used, we could not drop that table join if A1 defines a `@Where` condition.

This is IMO the best we can do

NathanQingyangXu: Thanks for the explanation. Hopefully v6 can be smarter on this.

beikov: We'd need to parse the SQL the user provides and I'm not sure this can be done perfectly in general, although we seem to do it already somehow because the column is prefixed properly with the table alias, so maybe there might be a way to do it right now also.

NathanQingyangXu: Yeah, that was my concern. We should have finished semantic parsing of filter fragment already.

beikov:
Anything else you want to add or discuss @NathanQingyangXu ? Otherwise I'd merge this.

NathanQingyangXu:
none. go ahead.

PR #3483, opened 2020-07-27
Hhh 14119

OBreidenbach:
Test case and naive solution for HHH-14119. I quess there are better solutions for this issue.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
@@ -228,7 +228,7 @@ public class InPredicate<T>
 							int bindValuePaddingCount = MathHelper.ceilingPowerOfTwo(bindValueCount);
 
 							if ( inExprLimit > 0 && bindValuePaddingCount > inExprLimit ) {
-								bindValuePaddingCount = Math.max(bindValueCount, inExprLimit);
+								bindValuePaddingCount = inExprLimit;
 							}
 
 							if ( bindValueCount < bindValuePaddingCount ) {
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
@@ -14,7 +14,11 @@ import java.util.List;
 import javax.persistence.criteria.Expression;
 import javax.persistence.criteria.Subquery;
 
+import org.hibernate.dialect.Dialect;
+import org.hibernate.engine.jdbc.spi.JdbcServices;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.internal.util.MathHelper;
+import org.hibernate.query.criteria.LiteralHandlingMode;
 import org.hibernate.query.criteria.internal.CriteriaBuilderImpl;
 import org.hibernate.query.criteria.internal.ParameterRegistry;
 import org.hibernate.query.criteria.internal.Renderable;
@@ -197,14 +201,51 @@ public class InPredicate<T>
 				}
 			}
 			else {
+				Expression lastValue = null;
 				buffer.append( '(' );
 				String sep = "";
 				for ( Expression value : values) {
+					lastValue = value;
 					buffer.append( sep )
 							.append( ( (Renderable) value )
 									.render( renderingContext ) );
 					sep = ", ";
 				}
+
+
+				if (renderingContext.getCriteriaLiteralHandlingMode() == LiteralHandlingMode.BIND &&
+						lastValue instanceof LiteralExpression) {
+
+					int bindValueCount = values.size();
+					int bindValueMaxCount = bindValueCount;
+
+					final SessionFactoryImplementor sfi = criteriaBuilder().getEntityManagerFactory().unwrap( SessionFactoryImplementor.class );
+					final Dialect dialect = sfi.getServiceRegistry().getService(JdbcServices.class).getJdbcEnvironment().getDialect();
+					final int inExprLimit = dialect.getInExpressionCountLimit();
+
+					boolean inClauseParameterPaddingEnabled = sfi.getSessionFactoryOptions().inClauseParameterPaddingEnabled() &&
+									bindValueCount > 2;
+
+					if ( inClauseParameterPaddingEnabled ) {
+						int bindValuePaddingCount = MathHelper.ceilingPowerOfTwo( bindValueCount );
+
+						if (inExprLimit > 0 && bindValuePaddingCount > inExprLimit ) {
+							bindValuePaddingCount = Math.max(bindValueCount, inExprLimit);
+						}
+
+						if ( bindValueCount < bindValuePaddingCount ) {
+							bindValueMaxCount = bindValuePaddingCount;
+						}
+
+						for (int i = bindValueCount; i < bindValueMaxCount; i++) {
+							buffer.append( sep )
+									.append( ( (Renderable) lastValue )
+											.render( renderingContext ) );
+						}
+					}
+				}
+
+
 				buffer.append( ')' );
 			}
 		}
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
@@ -201,50 +201,48 @@ public class InPredicate<T>
 				}
 			}
 			else {
-				Expression lastValue = null;
 				buffer.append( '(' );
 				String sep = "";
 				for ( Expression value : values) {
-					lastValue = value;
 					buffer.append( sep )
 							.append( ( (Renderable) value )
 									.render( renderingContext ) );
 					sep = ", ";
 				}
 
+				if ( renderingContext.getCriteriaLiteralHandlingMode() == LiteralHandlingMode.BIND ) {
+					Expression lastValue = values.get( values.size() - 1 );
+					if ( lastValue instanceof LiteralExpression ) {
+						final int bindValueCount = values.size();
+						int bindValueMaxCount = bindValueCount;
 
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
```diff
@@ -197,14 +201,51 @@ public String render(boolean isNegated, RenderingContext renderingContext) {
...
+
+				if (renderingContext.getCriteriaLiteralHandlingMode() == LiteralHandlingMode.BIND &&
+						lastValue instanceof LiteralExpression) {
+
+					int bindValueCount = values.size();
+					int bindValueMaxCount = bindValueCount;
+
+					final SessionFactoryImplementor sfi = criteriaBuilder().getEntityManagerFactory().unwrap( SessionFactoryImplementor.class );
+					final Dialect dialect = sfi.getServiceRegistry().getService(JdbcServices.class).getJdbcEnvironment().getDialect();
+					final int inExprLimit = dialect.getInExpressionCountLimit();
```
NathanQingyangXu, line 224: Could we move the above statements within the following if block:
```
if ( inclauseParameterPaddingEnabled ) {
    .. ...
}
```
Seems `inExprLimit` is only needed within the block and we can avoid unnecessary performance cost if padding is not enabled.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
```diff
@@ -197,14 +201,51 @@ public String render(boolean isNegated, RenderingContext renderingContext) {
...
+
+					boolean inClauseParameterPaddingEnabled = sfi.getSessionFactoryOptions().inClauseParameterPaddingEnabled() &&
+									bindValueCount > 2;
+
+					if ( inClauseParameterPaddingEnabled ) {
+						int bindValuePaddingCount = MathHelper.ceilingPowerOfTwo( bindValueCount );
+
+						if (inExprLimit > 0 && bindValuePaddingCount > inExprLimit ) {
+							bindValuePaddingCount = Math.max(bindValueCount, inExprLimit);
+						}
```
NathanQingyangXu, line 234: I think we can simplify the above as
```
if ( inExprLimit > 0 && bindValuePaddingCount > inExprLimit ) {
    bindValuePaddingCount = inExprLimit;
}
```
Seems the later logic will ensure correctness when `bindValueCount > inExprLimit` (in that case `bindValueMaxCount won't be assigned as per the if block below).

IMHO, we don't need `bindValueMaxCount` and we can simply the logic as following:
```
if ( inClauseParameterPaddingEnabled ) {
    int bindValuePaddingCount = MathHelper.ceilingPowerOfTwo( bindValueCount );

    if ( inExprLimit > 0 && bindValuePaddingCount > inExprLimit ) {
	bindValuePaddingCount = inExprLimit;
    }
    
    for (int i = bindValueCount; i < bindValuePaddingCount; i++) {
        ... ...
    }
```

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
```diff
@@ -197,14 +201,51 @@ public String render(boolean isNegated, RenderingContext renderingContext) {
...
+					lastValue = value;
 					buffer.append( sep )
 							.append( ( (Renderable) value )
 									.render( renderingContext ) );
 					sep = ", ";
 				}
+
+
+				if (renderingContext.getCriteriaLiteralHandlingMode() == LiteralHandlingMode.BIND &&
+						lastValue instanceof LiteralExpression) {
```
NathanQingyangXu, line 217: It would be great to add space chars after `if (` and before `) {`.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
```diff
@@ -197,14 +201,51 @@ public String render(boolean isNegated, RenderingContext renderingContext) {
...
+
+					int bindValueCount = values.size();
+					int bindValueMaxCount = bindValueCount;
+
+					final SessionFactoryImplementor sfi = criteriaBuilder().getEntityManagerFactory().unwrap( SessionFactoryImplementor.class );
+					final Dialect dialect = sfi.getServiceRegistry().getService(JdbcServices.class).getJdbcEnvironment().getDialect();
+					final int inExprLimit = dialect.getInExpressionCountLimit();
+
+					boolean inClauseParameterPaddingEnabled = sfi.getSessionFactoryOptions().inClauseParameterPaddingEnabled() &&
+									bindValueCount > 2;
```
NathanQingyangXu, line 227: swapping the parameter order around `&&` might be an idea for `bindValueCount > 2` is relatively lightweight and we can avoid some unnecessary cpu cycles due to short circuiting.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
```diff
@@ -197,14 +201,51 @@ public String render(boolean isNegated, RenderingContext renderingContext) {
...
+						}
+
+						if ( bindValueCount < bindValuePaddingCount ) {
+							bindValueMaxCount = bindValuePaddingCount;
+						}
+
+						for (int i = bindValueCount; i < bindValueMaxCount; i++) {
+							buffer.append( sep )
+									.append( ( (Renderable) lastValue )
+											.render( renderingContext ) );
```
NathanQingyangXu, line 243: Do you have some formatting issue above for as you can see the method chaining are not aligned vertically correctly.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
```diff
@@ -197,14 +201,51 @@ public String render(boolean isNegated, RenderingContext renderingContext) {
 				}
 			}
 			else {
+				Expression lastValue = null;
 				buffer.append( '(' );
 				String sep = "";
 				for ( Expression value : values) {
+					lastValue = value;
```
NathanQingyangXu, line 208: I think we don't need to assign it multiple times (suppose our list has 100 items, we need to assign to `lastValue` 100 times). We can simply get its value and assign it only once outside of the loop.
```

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
NathanQingyangXu, line 234: I think we'd better leave the above logic in the other PR of yours (i.e. https://github.com/hibernate/hibernate-orm/pull/3468) to make concerns separated.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
OBreidenbach: I just copied lines 209-211. I thought the formatting is ok, as it was done by a member of the hibernate team.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
OBreidenbach: So what do you suggest as a solution? 
I thought as Hibernate 6.0 is already in alpha state, and it is changing the generation of the sql completly, it would be a good enough solution to dublicate some code. 
Alternativ the InPredicate should put the values as list into one parameter so the QueryParameterBindingsImpl can do its job. But I do not know how to do this exactly.
Maybe it is enough to include my test case. So we can be sure this bug is fixed in release 6.0.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
OBreidenbach: It is just a copy&paste from QueryParameterBindingsImpl. I thought it is a goog idee to alter the code as little possible.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/InPredicate.java
NathanQingyangXu: I've thought this PR is targeting v5, not v6. I don't know too much about the implementation detail, but given you have created another PR for the above, including it here seems distracting.

OBreidenbach: Two different issues, hence two different solutions.
#3468 is fixing [HHH-14109](https://hibernate.atlassian.net/projects/HHH/issues/HHH-14109). This works if the value list is set as parameter to the prepared statement. For Oracle that means padding values between 513 and 999 to 1000. 

This PR fixes using criteria query and using literalHandlingMode == BIND does not do any parameter padding. As shown by my extension to the testcase.
Yes this PR is targeting v5. But I thought code duplication is no big deal for v5. As far I understand v6 correct it will rewrite the hole sql generation code anyway. 

As I said in my comment to the PR. That is a somewhat simple solution. May be you just ake my testcase from the testcase to make sure the bug will be fixed in v6.

NathanQingyangXu: I am not concerned with v6 since this PR is targeting v5. My concern is we should make ticket and PR independent. If the duplicated content is included, there are the following issues I can think of from ticket management:

1. we might end up with duplicated code review, which is a waste of time and efforts
2. if the duplicated content needs improvement, we need to modify multiple PRs, which is error-prone
3. if the duplicated content is not mandatory for this ticket (it seems so from my understanding), including it mislead the reviewer to miss the main point in the ticket.

Maybe I misunderstood something, but concern separating is a good thing in ticket management, just like in our routine coding practice.

gavinking:
This wasn't a regression, so it's not something we're likely to fix in H5.6.

And all this stuff changed completely in H6, and the test passes.

I'm going to close this PR; @sebersole can correct me if this was the wrong action to take.

gavinking:
See #9315

PR #3484, opened 2020-07-29
HHH-14124 Entity graph (fetch graph) is incorrectly applied to query results beyond the first one

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14124

Essentially, the current implementation of sub-graph handling relies on side-effects in the session: when it loads an entity with an entity graph, it will replace the entity graph in the session with the subgraph, and hope that the next loaded entity needs this subgraph.

This will sort of work, though it's probably riddled with bugs (what if there are multiple subgraphs? what if entities are not processed in the expected order?). These bugs are not what HHH-14124 is about, however, so I won't try to address them.

The real problem that caused HHH-14124 is that this implementation of sub-graph handling doesn't take care of restoring the "root" entity graph after we're done processing one row in the result set. It may leave an incorrect graph, or even set the graph to null. As a result, in many cases only the very first entity in the result set will be loaded correctly. The next ones won't have the correct entity graph in the session, and thus won't be loaded correctly.

This PR tries to solve the problem with a workaround: by storing the entities to load in a structure that lets us know when we're moving from one row to the next, we're able to "restore" the original entity graph before we load the entities of each row. That way, the bugs of the sub-graph handling implementation won't affect the next row, and simple graphs without a sub-graph will work correctly (like they used to). Sub-graphs will still be buggy, but that's been the case for a long time, and as I wrote above, solving this is not the goal of this PR.

You'll notice there are two commits. 

The first one uses a naive approach of storing the entities to load in a `List<List<Object>>`, each sub-list containing all the objects to load for a given row. That's a very straightforward approach, but it requires to allocate one list per result row, which I think is something we want to avoid. Thus I added a second commit which stores entities to load in a `List<Object>`, but inserts a `null` element to signal that the next elements were extracted from a different row.

Either solution is fine to me, but I think the second one, while quite ugly, is less likely to lead to performance regressions (since it'll allocate memory very similarly to the current code). It might be a good-enough compromise until we rewrite it all in Hibernate ORM 6.0?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
@@ -441,8 +441,10 @@ public abstract class Loader {
 						loadedKeys,
 						returnProxies
 				);
-				// Signal that a new row starts. Used in initializeEntitiesAndCollections
-				nullSeparatedHydratedObjects.add( null );
+				if ( nullSeparatedHydratedObjects != null ) {
+					// Signal that a new row starts. Used in initializeEntitiesAndCollections
+					nullSeparatedHydratedObjects.add( null );
+				}
 				if ( !keyToRead.equals( loadedKeys[0] ) ) {
 					throw new AssertionFailure(
 							String.format(
@@ -1051,8 +1053,10 @@ public abstract class Loader {
 					forcedResultTransformer
 			);
 			results.add( result );
-			// Signal that a new row starts. Used in initializeEntitiesAndCollections
-			nullSeparatedHydratedObjects.add( null );
+			if ( nullSeparatedHydratedObjects != null ) {
+				// Signal that a new row starts. Used in initializeEntitiesAndCollections
+				nullSeparatedHydratedObjects.add( null );
+			}
 			if ( createSubselects ) {
 				subselectResultKeys.add( keys );
 				keys = new EntityKey[entitySpan]; //can't reuse in this case
--- a/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
@@ -406,7 +406,7 @@ public abstract class Loader {
 		}
 
 		initializeEntitiesAndCollections(
-				hydratedObjects == null ? null : Collections.singletonList( hydratedObjects ),
+				hydratedObjects,
 				resultSet,
 				session,
 				queryParameters.isReadOnly( session )
@@ -423,7 +423,7 @@ public abstract class Loader {
 			final EntityKey keyToRead) throws HibernateException {
 
 		final int entitySpan = getEntityPersisters().length;
-		final List<List<Object>> hydratedObjectsPerRow = entitySpan == 0 ?
+		final List<Object> nullSeparatedHydratedObjects = entitySpan == 0 ?
 				null : new ArrayList<>( entitySpan );
 
 		Object result = null;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/loader/Loader.java
```diff
@@ -68,6 +68,7 @@
 import org.hibernate.event.spi.PostLoadEvent;
```
NathanQingyangXu, line 68: It seems a good catch. I am the community developer who implemented 'fetch graph' in v5. Also I implemented entity graph fully in v6. Yeah, I agree. The current implementation is buggy and difficult to maintain. That is because v5 infrastructure and design makes it difficult to implement and that is why 'fetch graph' was not implemented for a long time. The status has been changed in v6 and I think we have implemented the whole entity graph elegantly there already.

I would take a look at the PR and thanks for the two options to choose from. However, I am community contributor and have very limited free cycles out of my daily job. Seems an endless fixing of entity graph in v5, :(.

yrodiere: Right, I also noticed that a better solution to handle fetch graphs would require very significant changes to ORM 5. That's why I didn't even try :)

NathanQingyangXu: I would prefer non-intrusive approach to fix the issue without much performance cost. The improvement in v6 is we have complete access to the navigation role path of the entity to be loaded. There might be such role info already in TwoPhaseLoad, though no time to investigate.

NathanQingyangXu: Would you mind my creating another PR if I can solve the issue without changing Loader or creating any new data structure (reusing existing info instead)? Not necessarily possible. Just in case.

hibernate-core/src/main/java/org/hibernate/loader/Loader.java
```diff
@@ -1215,8 +1242,13 @@ private void initializeEntitiesAndCollections(
...
 
-		if ( hydratedObjects != null ) {
-			for ( Object hydratedObject : hydratedObjects ) {
+		if ( nullSeparatedHydratedObjects != null ) {
+			for ( Object hydratedObject : nullSeparatedHydratedObjects ) {
+				if ( hydratedObject == null ) {
+					// This is a hack to signal that we're starting to process a new row
+					// Ignore
+					continue;
+				}
```
NathanQingyangXu, line 1251: Did you conduct some MPH benchmark comparison to justify the hack? It would be great to avoid such hack for the entity graph implementation logic has been so difficult to understand and maintain.

yrodiere: I did not. I'm just considering that keeping the same data structure (a List) and adding a few elements is less change than moving to a `List<List<?>>`. I'm not saying it's the same, mind you. Just that it's the closest I can get.

hibernate-core/src/main/java/org/hibernate/loader/Loader.java
yrodiere: I wouldn't mind as long as my tests still pass, but I'd rather you create another ticket and a follow-up PR if you want to make more changes to address more problems. You know, one step at a time.

I also think it might not be the best use of your time, since 6.0 is the future, but that's _your_ time :)

NathanQingyangXu:
Suppose I created a new PR aiming to only improve code readibility without
solving new bug, is that okay for v5? I do think if we add backtracing
feature we might solve the issue in a simple way.

hibernate-core/src/main/java/org/hibernate/loader/Loader.java
NathanQingyangXu: I am not against such hack as long as it turns out to be the hot spot. Also, the fetch graph impl in v5 has been based on hack already. Overall the PR looks good from a quick review. I might focus on it more later in my off hours to finalize the review. Thanks for the bug reporting and quick fixing.

hibernate-core/src/main/java/org/hibernate/loader/Loader.java
```diff
@@ -1184,22 +1193,40 @@ private void initializeEntitiesAndCollections(
...
-			LOG.tracev( "Total objects hydrated: {0}", hydratedObjectsSize );
+		if ( nullSeparatedHydratedObjects != null && !nullSeparatedHydratedObjects.isEmpty() ) {
+			if ( LOG.isTraceEnabled() ) {
+				int hydratedObjectsSize = 0;
+				for ( Object hydratedObject : nullSeparatedHydratedObjects ) {
+					if ( hydratedObject != null ) {
+						++hydratedObjectsSize;
+					}
+				}
+				LOG.tracev( "Total objects hydrated: {0}", hydratedObjectsSize );
```
NathanQingyangXu, line 1204: Should the first format string to be consistent with the one of `String#format()`? Then we should use `%d` as opposed to `{0}`?

hibernate-core/src/main/java/org/hibernate/loader/Loader.java
NathanQingyangXu: I tried to create my PR but failed. Especially there is no way to even know some hydrated object is of root entity or not (the relationship among them is totally lost in TwoPhaseLoad class). Seems no better way to solve the issue. Great job and thanks a ton.

dreab8:
Thanks @yrodiere

yrodiere:
Thanks for the reviews!

gavinking:
These changes have broken Hibernate Reactive rather badly.

See https://github.com/hibernate/hibernate-reactive/issues/309.

PR #3486, opened 2020-07-30
wip/6.0 Make FastSessionServices faster

NathanQingyangXu:
The gist of this PR is to present a feasibility to improve FastSessionServices via EnumMap as opposed to HashMap for `defaultSessionProperties`.

If the `defaultSessionProperties` is a hot spot, the improvement might be worthwhile.

As well-known, EnumMap is compact and cannot be beat in terms of performance. Seems the `defaultSessionProperties` field is a good candidate for it has very limited keys (unlike AvailableSettings which has tons of entries).


NathanQingyangXu:
need more time to finalize it.

PR #3489, opened 2020-07-31
HHH-14133 fix obvious example caching code error in user guide

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14133

Simple error of including empty char at the end of hint property key. Should be intolerable for example code in user guide.


dreab8:
Merged, Thanks @NathanQingyangXu

PR #3490, opened 2020-08-03
HHH-14129 Bidirectional relationship with @NotNull fails to save

dreab8:
https://hibernate.atlassian.net/browse/HHH-14129

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/annotations/manytoone/NotNullManyToOneTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/annotations/manytoone/NotNullManyToOneTest.java
@@ -1,0 +1,93 @@
+package org.hibernate.test.annotations.manytoone;
+
+import javax.persistence.CascadeType;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/test/annotations/manytoone/NotNullManyToOneTest.java
```diff
@@ -0,0 +1,93 @@
...
+		inTransaction(
+				session -> {
+					Parent parent = new Parent( new Child() );
+					session.save( parent );
+				}
+		);
+	}
+
+	@Test(expected = javax.validation.ConstraintViolationException.class)
+	public void testSaveWithNullChild() {
```
NathanQingyangXu, line 49: the method name is a little bit confusing for the real issue is null `parent`

dreab8: fixed it, Thanks

PR #3492, opened 2020-08-06
Wip/6.0 member of predicate impl

dreab8:


hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
```diff
@@ -923,6 +922,18 @@ public void visitConversion(Conversion conversion) {
...
+	@Override
+	public void visitMemberOfPredicate(MemberOfPredicate memberOfPredicate) {
+		memberOfPredicate.getLeftHandExpression().accept( this );
+		if ( memberOfPredicate.isNegated() ) {
+			appendSql( " not " );
+		}
+		appendSql( " in (" );
+
+		visitQuerySpec( memberOfPredicate.getQuerySpec() );
+		appendSql( ")" );
```
NathanQingyangXu, line 934: Is there possibility that `visaitQuerySpec` adds nothing to the SQL and an erroneous SQL like `in ()` is issued (which only works for h2 but fails for any other dbs)? That is my first contribution to Hibernate and the solution is simply to change it to `in ( null )`.

hibernate-core/src/test/java/org/hibernate/orm/test/annotations/collectionelement/QueryTest.java
```diff
@@ -21,14 +27,38 @@
...
-				session ->
-						session.createQuery( "from EntityWithAnElementCollection e where 'abc' member of e.someStrings" )
-								.list()
+				session -> {
+					List<EntityWithAnElementCollection> list = session.createQuery(
+							"from EntityWithAnElementCollection e where 'abc' member of e.someStrings" )
+							.list();
+					assertThat( list.size(), is( 1 ) );
+				}
 		);
```
NathanQingyangXu, line 62: I would suggest a testing case in which `e.someStrings` is empty to see whether exception is thrown (because of the erroneous `in ()` SQL which only works in h2).

sebersole: This particular one, no.  For a member-of predicate it would be illegal for the sub-query to be empty.

However, you have a great point about other cases for in-predicates in the source query.  I'm not sure `in(null)` works on all databases - seems we have had this discussion before a few years ago.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
sebersole: This particular one, no. For a member-of predicate it would be illegal for the sub-query to be empty.

However, you have a great point about other cases for in-predicates in the source query. I'm not sure in(null) works on all databases - seems we have had this discussion before a few years ago.

hibernate-core/src/test/java/org/hibernate/orm/test/annotations/collectionelement/QueryTest.java
sebersole: You mean that `someStrings` for a particular `e` has no elements (e.getSomeStrings().isEmpty() == true)?  That's still ok.  The sub-query we generate would simply return no rows.  But the in-predicate sub-query expression itself is not null.  Meaning, in SQL, this is perfectly valid whether the sub-query returns no rows or more than zero:

```
select e 
from entity_with_elem_coll e
where 'abc' in (
    select ...
)
where i
```

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
```diff
@@ -373,7 +373,7 @@ predicate
 	| expression (NOT)? LIKE expression (likeEscape)?		# LikePredicate
 	| expression comparisonOperator expression				# ComparisonPredicate
 	| EXISTS expression										# ExistsPredicate
-	| MEMBER OF path										# MemberOfPredicate
+	| expression MEMBER OF path								# MemberOfPredicate
```
sebersole, line 376: Good catch

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/CollectionLoaderBatchKey.java
```diff
@@ -176,17 +175,7 @@ private void batchLoad(
...
-										public JdbcMapping getBindType() {
-											return type;
-										}
-
-										@Override
-										public Object getBindValue() {
-											return value;
-										}
-									}
+									new JdbcParameterBindingImpl( type, value )
```
sebersole, line 178: +1

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/BasicValuedCollectionPart.java
```diff
@@ -238,9 +239,8 @@ public void visitJdbcTypes(
 		action.accept( getJdbcMapping() );
 	}
 
-	//
```
sebersole, line 241: I'm curious... why remove this?

NathanQingyangXu:
thanks for explanation. seems a false alarm.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/AssignableSqmPathInterpretation.java
```diff
@@ -1,31 +0,0 @@
-/*
```
sebersole, line 1: I guess this is the interface you mentioned to me on Zulip?  I'm surprised it is not used, but then again I forget exactly how I implemented this part of UPDATE HQL/Criteria handling.

sebersole:
You're welcome :)

And thank you for your continued help with many of these pull requests

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/BasicValuedCollectionPart.java
dreab8: the commented code? I really do not remember

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/AssignableSqmPathInterpretation.java
dreab8: yes this is the one I mentioned you

dreab8:
Applied upstream.
@NathanQingyangXu @sebersole thanks for the review

PR #3493, opened 2020-08-07
wip/6.0 Reuse ArrayHelper's empty array constants to avoid unnecessary array creation

NathanQingyangXu:
We have the following empty array constants in `org.hibernate.internal.util.collections.ArrayHelper`:
```
        public static final String[] EMPTY_STRING_ARRAY = {};
	public static final int[] EMPTY_INT_ARRAY = {};
	public static final boolean[] EMPTY_BOOLEAN_ARRAY = {};
	public static final Class[] EMPTY_CLASS_ARRAY = {};
	public static final Object[] EMPTY_OBJECT_ARRAY = {};
	public static final Type[] EMPTY_TYPE_ARRAY = {};
	public static final byte[] EMPTY_BYTE_ARRAY = {};
```

This PR aims to reuse these constant in lieu of empty array creation to both boost performance and decrease memory footprint. Given the array is empty, it seems the replacement is hassle-free (unlike `Collections.emptyList()` which is immutable and adding element triggers exception).

As per https://stackoverflow.com/questions/30132902/how-is-the-zero-length-array-represented-in-memory, 32 bytes will be allocated for an empty array.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/AbstractPluralAttributeSourceImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/AbstractPluralAttributeSourceImpl.java
@@ -34,6 +34,7 @@ import org.hibernate.boot.model.source.spi.PluralAttributeSource;
 import org.hibernate.boot.model.source.spi.TableSpecificationSource;
 import org.hibernate.boot.model.source.spi.ToolingHintContext;
 import org.hibernate.cfg.NotYetImplementedException;
+import org.hibernate.internal.util.collections.ArrayHelper;
 
 /**
  * @author Steve Ebersole
@@ -169,7 +170,7 @@ public abstract class AbstractPluralAttributeSourceImpl
 
 	private static String[] extractSynchronizedTableNames(PluralAttributeInfo pluralAttributeElement) {
 		if ( pluralAttributeElement.getSynchronize().isEmpty() ) {
-			return new String[0];
+			return ArrayHelper.EMPTY_STRING_ARRAY;
 		}
 
 		final String[] names = new String[pluralAttributeElement.getSynchronize().size()];
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/CodeTemplates.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/CodeTemplates.java
@@ -25,6 +25,7 @@ import org.hibernate.engine.spi.CompositeOwner;
 import org.hibernate.engine.spi.CompositeTracker;
 import org.hibernate.engine.spi.ExtendedSelfDirtinessTracker;
 import org.hibernate.engine.spi.PersistentAttributeInterceptor;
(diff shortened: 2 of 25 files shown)


hibernate-core/src/main/java/org/hibernate/action/internal/BulkOperationCleanupAction.java
```diff
@@ -91,7 +92,7 @@ public BulkOperationCleanupAction(SharedSessionContractImplementor session, Quer
 			}
 		}
 
-		this.affectedTableSpaces = spacesList.toArray( new String[ spacesList.size() ] );
+		this.affectedTableSpaces = spacesList.toArray( ArrayHelper.EMPTY_STRING_ARRAY );
```
NathanQingyangXu, line 95: In modern JVM, `spacesList.toArray( new String[0] )` is faster than `spacesList.toArray( new String[ spacesList.size() ] )`. E.g. https://stackoverflow.com/questions/174093/toarraynew-myclass0-or-toarraynew-myclassmylist-size

Similar changes are done elsewhere in this PR.

beikov:
Why did you close the PR? I wasn't aware of it, but I would consider merging parts of it if you would agree to change/revert a few things.

NathanQingyangXu:
I've thought this PR is too minor to be serious., :). I am gonna reopen it
to let you review it.

hibernate-core/src/main/java/org/hibernate/cfg/annotations/reflection/JPAOverriddenAnnotationReader.java
```diff
@@ -129,6 +129,7 @@
 import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.internal.util.StringHelper;
+import org.hibernate.internal.util.collections.ArrayHelper;
```
beikov, line 132: Revert please
```suggestion
```

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -45,6 +45,7 @@
 import org.hibernate.engine.internal.NonNullableTransientDependencies;
 import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
+import org.hibernate.internal.util.collections.ArrayHelper;
```
beikov, line 48: Revert please
```suggestion
```

beikov (changes requested):
Please remove the unnecessary imports, then I will merge it

hibernate-core/src/main/java/org/hibernate/cfg/annotations/reflection/JPAOverriddenAnnotationReader.java
NathanQingyangXu: done. thanks.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
NathanQingyangXu: done. thanks.

PR #3497, opened 2020-08-13
Embedded identifer mapping

dreab8:
implementing SingleAttributeIdentifierMapping

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityResultGraphNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityResultGraphNode.java
@@ -6,8 +6,6 @@
  */
 package org.hibernate.sql.results.graph.entity;
 
-import java.util.ArrayList;
-
 import org.hibernate.LockMode;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityResultGraphNode.java
```diff
@@ -131,33 +120,42 @@ public AbstractEntityResultGraphNode(
...
+							);
+						}
+						else {
+							attributeMapping.createDomainResult(
+									navigablePath,
+									entityTableGroup,
+									null,
+									creationState
+							);
+						}
```
NathanQingyangXu, line 147: It might be minor, but the above code involves some duplication and might be simplified as
```
attributeMapping -> {
    final AttributeMapping realMapping;
    if ( attributeMapping instanceof ToOneAttributeMapping ) {
         realMapping = ( (ToOneAttributeMapping) attributeMapping ).getForeignKeyDescriptor();
    }
    else {
         realMapping = attributeMapping;
    }
    realMapping.createDomainResult(
        navigablePath,
        entityTableGroup,
        null,
        creationState
    );
}
```

dreab8:
applied upstream.

PR #3498, opened 2020-08-13
HHH-3326 Fix temp table collation issue for SQLServer dialect

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-3326

An ancient ticket to solve!

Implementation was copied from the code by the bug reporter, i.e. David Good, whose fixing was based on the stackoverflow page: https://stackoverflow.com/questions/21332105/temp-table-collation-conflict-error-cannot-resolve-the-collation-conflict-be

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/dialect/functional/SQLServerDialectTempTableCollationTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/dialect/functional/SQLServerDialectTempTableCollationTest.java
@@ -1,0 +1,161 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.test.dialect.functional;
+
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Inheritance;
+import javax.persistence.InheritanceType;
+import javax.persistence.Table;
+
+import org.hibernate.boot.registry.BootstrapServiceRegistry;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/test/java/org/hibernate/test/dialect/functional/SQLServerDialectTempTableCollationTest.java
```diff
@@ -0,0 +1,159 @@
...
+			catch (SQLException e) {
+				log.debug( e.getMessage() );
+			}
+			try ( Connection connection = serviceRegistry.getService( JdbcServices.class ).getBootstrapJdbcConnectionAccess().obtainConnection();
+				 Statement statement = connection.createStatement() ) {
+				connection.setAutoCommit( true );
+				statement.executeUpdate( "ALTER DATABASE CURRENT COLLATE " + changedDBCollation );
+			}
+			catch (SQLException e) {
+				log.debug( e.getMessage() );
```
beikov, line 110: Instead of logging a debug message, I think it would be better to have a catch clause in the outer try-finally block that rethrows as RuntimeException. Or is there a valid use case for a SQLException being thrown?

hibernate-core/src/test/java/org/hibernate/test/dialect/functional/SQLServerDialectTempTableCollationTest.java
```diff
@@ -0,0 +1,159 @@
...
+			);
+			try (Connection connection = serviceRegistry.getService( JdbcServices.class )
+					.getBootstrapJdbcConnectionAccess()
+					.obtainConnection();
+				 Statement statement = connection.createStatement()) {
+				connection.setAutoCommit( true );
+				statement.executeUpdate( "ALTER DATABASE CURRENT COLLATE " + originalDBCollation );
+			}
+			catch (SQLException e) {
+				log.debug( e.getMessage() );
```
beikov, line 68: Rethrow as RuntimeException

NathanQingyangXu:
I copied the pattern from another similar testing case. Seems the pattern
tries to avoid throwing exception for some reason.

beikov:
Thanks for the fix Nathan!

PR #3502, opened 2020-08-18
HHH-14152 reserve end-of-line chars in MultiLineImportExtractor

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14152

So the original implementation of `MultipleLinesSqlCommandExtractor` striped off end-of-line chars. However, end-of-line chars could be a meaningful delimiter as other white space chars. For instance, the following common SQL script
```
INSERT INTO ISSUERS
VALUES (1, 1, 'BIPlatform', NULL, NULL);
```
would be erroneously transformed to
```
INSERT INTO ISSUERSVALUES (1, 1, 'BIPlatform', NULL, NULL);
```

This PR tries to reserve such end-of-line chars.
Another strongly related minor issue is the trailing end-of-line chars given that they are reserved rather than skipped. I changed the parser rule to allow trailing white spaces to make the paring more robust (and less confusing. User would get misled and confused when they get `statement must end in ;` exception in face of trailing whitespaces).

The only testing script has been modified to make it more challenging. Without the code changes in this PR, both of the following testing cases would fail:

- org.hibernate.test.fileimport.MultiLineImportFileTest
- org.hibernate.test.fileimport.MultiLineImportExtractorTest

### DIFF
--- a/hibernate-core/src/main/antlr/sql-script.g
+++ b/hibernate-core/src/main/antlr/sql-script.g
@@ -56,7 +56,7 @@ script
     ;
 
 statement
-	: { statementStarted(); } (statementPart)*  DELIMITER { statementEnded(); }
+	: { statementStarted(); } (statementPart)* DELIMITER (WS_CHAR)* { statementEnded(); }
 	;
 
 statementPart
@@ -71,8 +71,11 @@ quotedString
 	;
 
 nonSkippedChar
-	: c:CHAR {
-   		out( c );
+	: w:WS_CHAR {
+   		out( w );
+   	}
+   	| o:OTHER_CHAR {
+   		out( o );
    	}
 	;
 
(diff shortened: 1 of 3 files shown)


hibernate-core/src/test/resources/org/hibernate/test/fileimport/multi-line-statements.sql
```diff
@@ -23,12 +23,14 @@ INSERT INTO test_data VALUES (2, 'Multi-line comment line 1
...
+VALUES
+	(
+    	4 -- another record
+     	, NULL
+    );
 -- comment;
 -- comment;
+
+
+
```
NathanQingyangXu, line 36: The above three lines are actually mixture of all kinds of white spaces, including ' ', '\t' and `\n`.

hibernate-core/src/main/antlr/sql-script.g
```diff
@@ -56,7 +56,7 @@ script
     ;
 
 statement
-	: { statementStarted(); } (statementPart)*  DELIMITER { statementEnded(); }
+	: { statementStarted(); } (statementPart)* DELIMITER (WS_CHAR)* { statementEnded(); }
```
beikov, line 59: What is the reason for parsing trailing whitespace chars?

NathanQingyangXu:
the reason is after reserving end-of-line chars, the original parser rule
doesn't apply. Seems intuitive to enlarge trailing end-of-line char to
other white space chars.

hibernate-core/src/main/antlr/sql-script.g
NathanQingyangXu: It seems intuitive to me, though not 100% necessary for this PR. After reserving end-of-line chars, we have to remember to remove the end-of-line after last `;`, which seems too demanding to me.

PR #3503, opened 2020-08-19
Return null early in the event of the CollectionEntry being null.

Jezza:
Avoids NPE, as the collection entry could not exist.


Sanne:
backport of #3499

PR #3504, opened 2020-08-19
HHH-14154 Incorrect SQL generated when concat() and function() methods are used together

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14154

The root cause is some sloppiness of code logic in `org.hibernate.hql.internal.ast.SqlGenerator` for the scenario wherein `FunctionNode`'s `sqlFunction` is null (e.g., the `FORMATDATETIME` function in the testing case of this PR).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/SqlGenerator.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/SqlGenerator.java
@@ -16,15 +16,12 @@ import org.hibernate.QueryException;
 import org.hibernate.dialect.function.SQLFunction;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.hql.internal.antlr.SqlGeneratorBase;
-import org.hibernate.hql.internal.antlr.SqlTokenTypes;
-import org.hibernate.hql.internal.ast.tree.CollectionPathNode;
 import org.hibernate.hql.internal.ast.tree.CollectionSizeNode;
 import org.hibernate.hql.internal.ast.tree.FromElement;
 import org.hibernate.hql.internal.ast.tree.FunctionNode;
 import org.hibernate.hql.internal.ast.tree.Node;
 import org.hibernate.hql.internal.ast.tree.ParameterContainer;
 import org.hibernate.hql.internal.ast.tree.ParameterNode;
-import org.hibernate.hql.internal.ast.util.ASTPrinter;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/SqlGenerator.java
```diff
@@ -192,13 +188,16 @@ protected void beginFunctionTemplate(AST node, AST nameNode) {
 		// 		METHOD_NAME
 		FunctionNode functionNode = (FunctionNode) node;
 		SQLFunction sqlFunction = functionNode.getSQLFunction();
+
+		outputStack.addFirst( writer );
+
 		if ( sqlFunction == null ) {
 			// if SQLFunction is null we just write the function out as it appears in the hql statement
+			writer = new StandardFunctionArguments();
 			super.beginFunctionTemplate( node, nameNode );
```
NathanQingyangXu, line 197: Previously no new `writer` was created and both `FORMATDATETIME` and `||` functions share the same writer, thus the strange SQL was created.

hibernate-core/src/test/java/org/hibernate/query/hhh14154/HHH14154Test.java
```diff
@@ -0,0 +1,60 @@
...
+	public void testNoExceptionThrown() {
+		doInJPA( this::sessionFactory, em -> {
+			final CriteriaBuilder cb = em.getCriteriaBuilder();
+			final CriteriaQuery<Foo> cq = cb.createQuery( Foo.class );
+			final Root<Foo> foo = cq.from( Foo.class );
+			cq.select(foo)
+				.where(
+					cb.lessThanOrEqualTo(
+						cb.concat(
+							cb.function( "FORMATDATETIME", String.class, foo.get( "startTime" ), cb.literal( "HH:mm:ss" ) ),
```
beikov, line 38: I don't think every DBMS supports this function, so please add a `@RequiresDialect(H2Dialect.class)`

beikov (changes requested):
Looks good, just a small change necessary

hibernate-core/src/test/java/org/hibernate/query/hhh14154/HHH14154Test.java
NathanQingyangXu: done. thanks. also added the bug reporter as one of the authors.

beikov (approved):

PR #3506, opened 2020-08-20
HHH-14160 Use a runtime used class for testing Bean Validation presence

gsmet:
https://hibernate.atlassian.net/browse/HHH-14160

I don't know the exact process but this should also be backported to 5.4.

Thanks!

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/beanvalidation/BeanValidationIntegrator.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/beanvalidation/BeanValidationIntegrator.java
@@ -32,7 +32,7 @@ public class BeanValidationIntegrator implements Integrator {
 
 	public static final String APPLY_CONSTRAINTS = "hibernate.validator.apply_to_ddl";
 
-	public static final String BV_CHECK_CLASS = "javax.validation.Validation";
+	public static final String BV_CHECK_CLASS = "javax.validation.ConstraintViolation";
 
 	public static final String MODE_PROPERTY = "javax.persistence.validation.mode";
 


hibernate-core/src/main/java/org/hibernate/cfg/beanvalidation/BeanValidationIntegrator.java
```diff
@@ -32,7 +32,7 @@
 
 	public static final String APPLY_CONSTRAINTS = "hibernate.validator.apply_to_ddl";
 
-	public static final String BV_CHECK_CLASS = "javax.validation.Validation";
+	public static final String BV_CHECK_CLASS = "javax.validation.ConstraintViolation";
 
 	public static final String MODE_PROPERTY = "javax.persistence.validation.mode";
```
NathanQingyangXu, line 38: Saw the original `javax.validation.Validation` comes from the following gradle dependency:
```
validation:     "javax.validation:validation-api:${validationApiVersion}",
```
Do we need to update it or delete it?

Sanne: We could remove it, but there's no benefit from removing it as it is a transitive dependency: better keep the explicit version control.

Sanne (approved):

hibernate-core/src/main/java/org/hibernate/cfg/beanvalidation/BeanValidationIntegrator.java
gsmet: I don't understand your question? The new class comes from the exact same dependency, there's nothing changed here.

Sanne:
@gsmet is this _necessary_ for a Quarkus fix, or a minor optimisation?

Just wondering if I should track anything on https://github.com/quarkusio/quarkus/issues/11231

gsmet:
No, it's not strictly necessary, just better as this class used to be entirely removed from the image. I tweaked the reflection registration to include as less things as possible.

But it would be nice to have it in the next 5.4 whenever it's released.

hibernate-core/src/main/java/org/hibernate/cfg/beanvalidation/BeanValidationIntegrator.java
NathanQingyangXu: It is my fault, I didn't know the library contains `javax.validation.ConstraintViolation` as well.

gsmet:
@Sanne how does the backport to 5.4 work?

Sanne:
> @Sanne how does the backport to 5.4 work?

I cherry pick it locally, run the tests, push (assuming no problems) and then track it on JIRA. Sometimes I might run extra checks, like Quarkus and Hibernate Reactive integrations - but I prefer batching these.

PR #3509, opened 2020-08-21
HHH-10282 Short-hand enum syntax support for HQL

beikov:


hibernate-core/src/main/java/org/hibernate/boot/internal/InFlightMetadataCollectorImpl.java
```diff
@@ -1664,7 +1664,7 @@ private void processValueResolvers(MetadataBuildingContext buildingContext) {
 
 		while ( ! valueResolvers.isEmpty() ) {
 			final boolean anyRemoved = valueResolvers.removeIf(
-					resolver -> resolver.apply( buildingContext )
+					resolver -> resolver == null || resolver.apply( buildingContext )
```
sebersole, line 1667: I'm curious where `null` is being put into this queue?  That seems wrong

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/JpaMetamodelImpl.java
```diff
@@ -493,6 +500,34 @@ public void processJpa(
...
 			}
+			Stream.concat(
+					context.getEntityTypesByEntityName().values().stream(),
+					Stream.concat(
+							context.getMappedSuperclassTypeMap().values().stream(),
+							context.getEmbeddableTypeSet().stream()
+					)
+			).forEach( managedDomainType -> {
+				managedDomainType.visitAttributes( persistentAttribute -> {
+					if ( persistentAttribute.getJavaType().isEnum() ) {
```
sebersole, line 511: Seems like this will break in cases of de-typed domain models (aka "map mode").  Depends how we end up deciding to handle that in building the JPA metamodel.  Regardless, its easy enough to handle here simply by checking for `persistentAttribute.getJavaType() == null`

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
```diff
@@ -1283,28 +1283,89 @@ else if ( ctx.GREATER_EQUAL()!=null ) {
 
 	@Override
 	public SqmComparisonPredicate visitComparisonPredicate(HqlParser.ComparisonPredicateContext ctx) {
+		ComparisonOperator comparisonOperator = (ComparisonOperator) ctx.comparisonOperator().accept( this );
```
sebersole, line 1286: Few things in this block to keep in mind...

1. We prefer to define final variables as `final`.  You have quite a few here.
2. We prefer code blocks use braces, including switch branches (i.e., `case NOT_EQUAL: { ... }`)

hibernate-core/src/main/java/org/hibernate/query/sql/internal/NativeSelectQueryPlanImpl.java
```diff
@@ -140,8 +144,12 @@ public NativeSelectQueryPlanImpl(
 							type = StandardBasicTypes.OBJECT_TYPE;
 						}
 
-						final JdbcMapping jdbcMapping = ( (BasicValuedMapping) type ).getJdbcMapping();
-						final JdbcParameterImpl jdbcParameter = new JdbcParameterImpl( jdbcMapping );
+						BasicValuedMapping basicValuedMapping = (BasicValuedMapping) type;
+						final JdbcMapping jdbcMapping = basicValuedMapping.getJdbcMapping();
+						final JdbcParameterImpl jdbcParameter = new JdbcParameterImpl(
```
sebersole, line 149: Really, the mapping here is only need to get the converter (if one).  I wonder if we should instead simply pass in the converter?

sebersole (commented):
Left a few minor comments.  

Regarding the enum work in particular, I wonder if there is a way we can clean that up so that it is handled once as part of the normal visitation of those nodes instead of having to deal with it across the various predicates.  As it is, if we add new predicates we have to remember that we might need to account for this enum handling.  Not saying its doable nor that this needs to change.  Just saying it is something I'd like to consider.

Aside from that and the minor comments LGTM!

hibernate-core/src/main/java/org/hibernate/boot/internal/InFlightMetadataCollectorImpl.java
beikov: You are right. Since the PR is quite old already, I still had that in as a workaround for a previously issue, but it's working fine now even without this. Thanks!

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/JpaMetamodelImpl.java
beikov: Nice catch, thanks. I adapted it.

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
beikov: Thanks I'll keep this in mind.

hibernate-core/src/main/java/org/hibernate/query/sql/internal/NativeSelectQueryPlanImpl.java
beikov: I moved the `MappingModelExpressable` to the `QueryParameterBinding` now like we discussed to handle the conversion during binding of the parameter value

beikov:
Thanks for the review @sebersole!

> Regarding the enum work in particular, I wonder if there is a way we can clean that up so that it is handled once as part of the normal visitation of those nodes instead of having to deal with it across the various predicates. As it is, if we add new predicates we have to remember that we might need to account for this enum handling. Not saying its doable nor that this needs to change. Just saying it is something I'd like to consider.

I don't think this is possible because we need to resolve the type of the non-literal side to dermine if the parse tree could be interpreted as enum literal. We might be able to pull this off for qualified literals, but since multiple enums with the same simple names might exists, we'd run into issues there again. I know it's not beautiful, but such a context sensitiveness probably never is :(

sebersole:
Yes, the "problem" is that we are trying something here that is "out of place" in terms of design.  Its really the next phase where interpretation/guessing of types is supposed to be done.  Things like this done out of place never seem clean.

I wonder if we could design an SQM node that is "delayed" in part of it resolution.  Something for us to think about

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/JpaMetamodelImpl.java
```diff
@@ -493,6 +500,34 @@ public void processJpa(
 			for ( EmbeddableDomainType<?> embeddable : context.getEmbeddableTypeSet() ) {
 				this.jpaEmbeddableDescriptorMap.put( embeddable.getJavaType(), embeddable );
 			}
+			Stream.concat(
+					context.getEntityTypesByEntityName().values().stream(),
+					Stream.concat(
+							context.getMappedSuperclassTypeMap().values().stream(),
+							context.getEmbeddableTypeSet().stream()
+					)
+			).forEach( managedDomainType -> {
```
NathanQingyangXu, line 509: We might consider the following pattern:
```
Stream.of( 
					context.getEntityTypesByEntityName().values().stream(),
					context.getMappedSuperclassTypeMap().values().stream(),
					context.getEmbeddableTypeSet().stream()
			).flatMap( Function.identity() ).forEach( ... )
```
As per https://www.techempower.com/blog/2016/10/19/efficient-multiple-stream-concatenation-in-java/, the flatmap approach performs better when the stream sizes are not large.

beikov: According to the article the suggestion is to use `flatMap(Stream::concat)` so I'd rather stick to this pattern.

hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingImpl.java
```diff
@@ -227,6 +229,15 @@ public void setBindValues(
...
 
+	@Override
+	public MappingModelExpressable getType() {
+		return type;
+	}
+
+	@Override
+	public void setType(MappingModelExpressable type) {
+		this.type = type;
+	}
```
NathanQingyangXu, line 240: I think we could eliminate two unnecessary generics warning by making use of `<T>` above.

hibernate-core/src/test/java/org/hibernate/test/enums/OrdinalEnumTypeTest.java
```diff
@@ -109,6 +109,46 @@ public void testEnumAsBindParameterAndExtract() {
...
+	@Test
+	@TestForIssue(jiraKey = "HHH-10282")
+	public void hqlTestEnumQualifiedShortHandSyntaxInPredicate() {
+		doInHibernate( this::sessionFactory, session -> {
+			session.createQuery(
+					"select id from Person where originalHairColor in (HairColor.BLONDE, HairColor.BROWN)")
+					.getResultList();
+		} );
+	}
+
```
NathanQingyangXu, line 151: I think it would be ideal to include negation counterparts for each of the new testing cases above.

hibernate-core/src/main/java/org/hibernate/sql/exec/internal/JdbcParameterImpl.java
```diff
@@ -13,6 +13,7 @@
  * @author Steve Ebersole
  */
 public class JdbcParameterImpl extends AbstractJdbcParameter {
+
 	public JdbcParameterImpl(JdbcMapping jdbcMapping) {
```
NathanQingyangXu, line 17: I think it is one of @sebersole 's personal styles to leave out the blank line before the first element statement in a class. I prefer the above change but it is not a big deal.

Another style of Steve is to leave out the space delimiter between generic types as following:
```
Map<K,V> map; // I would prefer Map<K, V> map
```
But I don't think it is a big deal and they have become my style components as well, :).

gavinking:
Great! We should have done this a long time ago but much better late than never! Thanks!

PR #3510, opened 2020-08-23
HHH-14181 fix bug where HQL update/delete query resulted in same table being updated twice

gavinking:
The `constraintOrderedTableNames` and `constraintOrderedKeyColumnNames` arrays in `UnionSubclassEntityPersister` contain a duplicate entry for the table of the persisted class.

This resulted in dupe SQL `update` and `delete` statements being executed for HQL bulk queries.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/UnionSubclassEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/UnionSubclassEntityPersister.java
@@ -144,12 +144,12 @@ public class UnionSubclassEntityPersister extends AbstractEntityPersister {
 				persistentClass.getEntityName()
 		);
 		if ( persistentClass.isPolymorphic() ) {
-			Iterator iter = persistentClass.getSubclassIterator();
+			Iterator<Subclass> subclassIter = persistentClass.getSubclassIterator();
 			int k = 1;
-			while ( iter.hasNext() ) {
-				Subclass sc = (Subclass) iter.next();
-				subclassClosure[k++] = sc.getEntityName();
-				subclassByDiscriminatorValue.put( sc.getSubclassId(), sc.getEntityName() );
+			while ( subclassIter.hasNext() ) {
+				Subclass subclass = subclassIter.next();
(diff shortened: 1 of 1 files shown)


beikov (changes requested):
I prefer to remove code instead of just commenting it out. Are you still doing some testing or is this the final state of the PR?
Also, it would be great if you could create a JIRA issue for tracking purposes and a testcase so we can make sure we don't run into the same problem with 6.0.

gavinking:
Well I'm not sure how to write a test that assets that we don't execute the exact same update statement twice.

Is that something we *typically* have teats for?

beikov:
We do, take a look at `org.hibernate.jpa.test.graphs.EntityGraphWithFetchAnnotationTest` for an example.

gavinking:
> We do

Now, I might be wrong, but I checked and AFAICS there are precisely zero tests of this nature for HQL queries.

So no, I don't have time to write the very first one of this nature right now because I'm actually working on other completely different things and this was just something I happened to run across.

If *you* think you would like to add a test like this, then *you* can write it.

Otherwise I would appreciate it if you would just merge the PR. Thanks in advance.

gavinking:
https://hibernate.atlassian.net/browse/HHH-14181

gavinking:
@beikov I've also opened https://hibernate.atlassian.net/browse/HHH-14182 which addresses the fact that there are no `SQLStatementInterceptor`-based tests for any HQL `update`/`delete` queries. I agree that it would be good for someone to spend some time writing at least a few tests like that, but I'm certainly not going to do it right now.

For now, there are plenty of other tests which already exercise the code path touched by this commit. If the commit had broken anything, I think we would know.

Sanne:
@sebersole , @dreab8 : we talked about the need for testing (and possibly recording) of the SQL statements actually being send to the DB; I take the discussion above ^ as another example of such needs.  Is there already better infrastructure in 6 for such an approach, or should I open/track this need on JIRA?

Sanne:
I've merged this in master and 5.4

gavinking:
Thanks!!

PR #3512, opened 2020-08-23
HHH-14179 fix MultiLineImportExtractorTest on Windows

famod:
Fixes https://hibernate.atlassian.net/browse/HHH-14179

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/fileimport/MultiLineImportExtractorTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/fileimport/MultiLineImportExtractorTest.java
@@ -40,11 +40,13 @@ public class MultiLineImportExtractorTest {
 				assertThat( commands, notNullValue() );
 				assertThat( commands.length, is( 6 ) );
 
+				// for Windows compatibility, System.lineSeparator() has to be used instead of just "\n"
+
 				assertThat( commands[0], startsWith( "CREATE TABLE test_data" ) );
 
 				assertThat( commands[1], is( "INSERT INTO test_data VALUES (1, 'sample')" ) );
 
-				assertThat( commands[2], is( "DELETE\n  FROM test_data" ) );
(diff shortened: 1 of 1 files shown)


beikov:
Hmm, I think we should normalize the line endings to just `\n` in the Antlr SQL generator to avoid this issue. The problem is that Git uses the platform line endings by default but can be configured to use `\n` instead.

NathanQingyangXu:
> Hmm, I think we should normalize the line endings to just `\n` in the Antlr SQL generator to avoid this issue. The problem is that Git uses the platform line endings by default but can be configured to use `\n` instead.

IMHO, the default git config makes sense for otherwise windows user might have trouble to read the resource file per se. Also, I think we'd better keep the original sql script content as it is (without normalizing).

beikov:
I am personally using `\n` line endings in Git on Windows and never had a problem with any tool. I think every tool handles all line ending characters already, so normalizing the line endings here should be fine. Unless someone has a reason for not doing this, I'd strongly prefer normalization.

famod:
What about `.gitattributes`?
```
*.sql         text eol=lf
```

NathanQingyangXu:
IDE will take care of end-of-line, but people might open the text file by
other means. I would prefer not to touch the git global setting for other
unexpected ripple effect. I think the code change in this PR seems enough.

NathanQingyangXu:
maybe we can raise the git global end-of-line config on zulip or mail list?
but it seems a little bit overkill for this ticket, IMHO.

famod:
I certainly won't resist if this PR is merged as is. 😉

NathanQingyangXu:
Your pattern based git config approach is great, but I think maybe the current fix is enough? maybe add one line comment explaining the reason?

famod:
> maybe add one line comment explaining the reason?

In the test, describing why `System.lineSeparator()` is required? Sure, why not!

PR #3513, opened 2020-08-24
HHH-14178 Make new collection skip delete-orphan 

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14178

The root cause of the unexpected exception reported is our failure to detect whether some collection is newly instantiated. 

So below is the original code snippet:
```
final boolean deleteOrphans = style.hasOrphanDelete()
				&& action.deleteOrphans()
				&& elemType.isEntityType()
				// a newly instantiated collection can't have orphans
				&& child instanceof PersistentCollection;

		if ( deleteOrphans ) {
```
Notice the comment for the last condition to decide whether deleteOrphan should be done. However, `child` is always of `PersistentCollection` type (we have collection field substitution logic before reaching here), so it boils down to the fact that we have no safety net against running 'delete-orphan' for new collection field.

How to detect whether some `PersistentCollection` is newly instantiated?

I added a new default method to `PersistentCollection' as following and added its usage in the above code snippet:
```
	default boolean isNewlyInstantiated() {
		return getKey() == null && !isDirty();
	}
```

I am open to suggestion regarding the above code logic or even better way.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/Cascade.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/Cascade.java
@@ -545,8 +545,9 @@ public final class Cascade {
 		final boolean deleteOrphans = style.hasOrphanDelete()
 				&& action.deleteOrphans()
 				&& elemType.isEntityType()
+				&& child instanceof PersistentCollection
 				// a newly instantiated collection can't have orphans
-				&& child instanceof PersistentCollection;
+				&& ! ( (PersistentCollection) child ).isNewlyInstantiated();
 
 		if ( deleteOrphans ) {
 			final boolean traceEnabled = LOG.isTraceEnabled();
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
@@ -269,7 +269,7 @@ public abstract class AbstractSaveEventListener
 		boolean substitute = substituteValuesIfNecessary( entity, id, values, persister, source );
 
 		if ( persister.hasCollections() ) {
-			substitute = substitute || visitCollectionsBeforeSave( entity, id, values, types, source );
+			substitute = visitCollectionsBeforeSave( entity, id, values, types, source ) || substitute;
 		}
 
 		if ( substitute ) {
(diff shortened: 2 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/event/internal/AbstractSaveEventListener.java
```diff
@@ -269,7 +269,8 @@ protected Serializable performSaveOrReplicate(
 		boolean substitute = substituteValuesIfNecessary( entity, id, values, persister, source );
 
 		if ( persister.hasCollections() ) {
-			substitute = substitute || visitCollectionsBeforeSave( entity, id, values, types, source );
+			final boolean substituteBecauseOfCollections = visitCollectionsBeforeSave( entity, id, values, types, source );
+			substitute = substitute || substituteBecauseOfCollections;
```
NathanQingyangXu, line 273: `visitCollectionsBeforeSave()` is a method with side effect. The side effect would not happen if `substitute` has been true (e.g., versioned entity) due to boolean operator short-circuiting.

beikov:
Would you mind integrating the test case that was attached to the issue so we avoid the same problem for 6.0?

gavinking:
Replicated fix in HR: https://github.com/hibernate/hibernate-reactive/commit/569caf1444c63f311b200505fc62fce720ef08dc

NathanQingyangXu:
> Would you mind integrating the test case that was attached to the issue so we avoid the same problem for 6.0?

Testing cases added. Maintained the original logic as much as possible. However, the original testing cases are too verbose and some of them are not needed. Also, an important testing scenario was overlooked (versioned map + versioned mapping).

hibernate-core/src/main/java/org/hibernate/engine/internal/Cascade.java
```diff
@@ -565,6 +566,10 @@ private static void cascadeCollectionElements(
 		}
 	}
 
+	private static boolean isCollectionNewlyInstantiated(PersistentCollection pc) {
```
beikov, line 569: Remove

beikov (changes requested):
Apart from a small issue, looks good!

hibernate-core/src/main/java/org/hibernate/engine/internal/Cascade.java
NathanQingyangXu: done. thanks for spotting this.

beikov (approved):

PR #3514, opened 2020-08-24
fix an obvious mistake in LocalTimeJavaDescriptor (H6)

gavinking:
This code is obviously wrong, though perhaps it doesn't actually produce any real bug.


Sanne:
@dreab8 there's the same commit merged in master as HHH-14184  : I suppose you'll prefer to just close this one and take it during the usual rebase process?

dreab8:
@Sanne, I'm going to apply it. Thanks @gavinking

PR #3515, opened 2020-08-24
fix an obvious mistake in LocalTimeJavaDescriptor (H5)

gavinking:
This code is obviously wrong, though perhaps it doesn't actually produce any real bug.


Sanne:
Surprising! Thanks, will merge soon - just needs a JIRA

gavinking:
@Sanne the reason nothing blows up is that `unwrap()` is normally called with `java.sql.Time` as an argument.

We don't have a `LocalTimeTypeDescriptor` which writes `LocalTime`s to JDBC.

PR #3516, opened 2020-08-24
HHH-14031 Update h2 to 1.4.197

famod:
Resolves https://hibernate.atlassian.net/browse/HHH-14031

Notes:
- h2 1.4.197+ does not support ns precision by default anymore
- h2 DateTimeUtils.resetCalendar() is required for many tests to avoid failures due to h2 internal caching

This was separated from #3412 which originally aimed at updating to 1.4.200+ but that path seems blocked for now: https://github.com/hibernate/hibernate-orm/pull/3412#issuecomment-678813877 (and subsequent comments).

### DIFF
--- a/hibernate-testing/src/main/java/org/hibernate/testing/junit4/BaseNonConfigCoreFunctionalTestCase.java
+++ b/hibernate-testing/src/main/java/org/hibernate/testing/junit4/BaseNonConfigCoreFunctionalTestCase.java
@@ -40,6 +40,7 @@ import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SessionImplementor;
 import org.hibernate.internal.build.AllowPrintStacktrace;
 import org.hibernate.internal.build.AllowSysOut;
+import org.hibernate.internal.util.ReflectHelper;
 import org.hibernate.internal.util.StringHelper;
 import org.hibernate.jdbc.Work;
 import org.hibernate.mapping.Collection;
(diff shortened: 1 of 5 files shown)


famod:
@Sanne (or someone else) Should be good to merge, no?

hibernate-testing/src/main/java/org/hibernate/testing/junit4/BaseNonConfigCoreFunctionalTestCase.java
```diff
@@ -450,6 +451,11 @@ protected void prepareTest() throws Exception {
 
 	@After
 	public final void afterTest() throws Exception {
+		// see https://github.com/hibernate/hibernate-orm/pull/3412#issuecomment-678338398
+		if ( getDialect() instanceof H2Dialect ) {
+			ReflectHelper.getMethod( Class.forName( "org.h2.util.DateTimeUtils" ), "resetCalendar" ).invoke( null );
```
NathanQingyangXu, line 456: I might be wrong, but it seems we can invoke the public util method directly:
```
DateTimeUtils.resetCalendar();
```
No?

NathanQingyangXu, line 456: Could we centralize the h2 workaround into `BaseUnitTestCase` to avoid ugly duplication?

NathanQingyangXu: Also, we have other testing base classes (e.g. `BaseEntityManagerFunctionalTestCase`) but they are inherit `BaseUnitTestCase`.

famod: We could, but in #3412 some people voted for reflective invocation so that no dependency is needed here.

famod: IIRC (on the phone right now), I had a look already and that `BaseUnitTestCase` does not know the current dialect.

NathanQingyangXu (approved):

Sanne:
many thanks, having a look...

Sanne:
rebased and merged. Thanks all!

PR #3517, opened 2020-08-24
explicitly specify precision/scale for @Column in test

gavinking:
If you have a `@Column` annotation, then by default you will get `precision=0`, `scale=0`, resulting in a `numeric(0,0)`.

This caused a test failure on H2 1.4.200. Not sure why it wasn't causing problems earlier.


gavinking:
Note: the related issue is https://hibernate.atlassian.net/browse/HHH-14031, but this change should be merged independently of that.

Sanne:
https://hibernate.atlassian.net/browse/HHH-14192

Sanne:
merged

PR #3519, opened 2020-08-25
on H2 prior to 2.0 map NUMERIC to DECIMAL (H6)

gavinking:
This is helpful to the schema update tool.

This workaround was lost during my work on `Dialect`s.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
@@ -101,6 +101,12 @@ public class H2Dialect extends Dialect {
 			this.sequenceInformationExtractor = SequenceInformationExtractorNoOpImpl.INSTANCE;
 			this.querySequenceString = null;
 		}
+
+		if ( version < 200 ) {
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
```diff
@@ -101,6 +101,12 @@ public H2Dialect(int version, int buildId) {
 			this.sequenceInformationExtractor = SequenceInformationExtractorNoOpImpl.INSTANCE;
 			this.querySequenceString = null;
 		}
+
+		if ( version < 200 ) {
```
famod, line 105: Shouldn't this be `<=` instead? Current version of h2 is 1.4.200.

gavinking: I don't think so, wasn't this the thing we all agreed that they changed in the .200 release?

gavinking: By "changed" I mean fixed.

beikov:
Thanks!

PR #3405, opened 2020-05-17
HHH-14023: Adjust H2Dialect for 1.4.201: sequence handling, DECIMAL, NUMERIC and VARBINARY

famod:
See also:
- https://github.com/h2database/h2database/issues/2625
- https://hibernate.zulipchat.com/#narrow/stream/132096-hibernate-user/topic/Locale.20problem.20when.20building.20hibernate-orm

cc @Sanne

To test this, you will need a current h2 SNAPSHOT (build it yourself from `master`) _with adjusted `org.h2.engine.Constants.BUILD_ID`_ (which is still on 200 until the release of 201).
I was able to get my local SNAPSHOT in to Gradle via:
```diff
diff --git a/gradle/java-module.gradle b/gradle/java-module.gradle
index 1928f00d42..8d2ea3dbbc 100644
--- a/gradle/java-module.gradle
+++ b/gradle/java-module.gradle
@@ -5,6 +5,11 @@
  * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
  */

+repositories {
+       mavenLocal()
+       mavenCentral()
+}
+
 buildscript {
        repositories {
                mavenCentral()
@@ -81,7 +86,7 @@ dependencies {
        testRuntime( libraries.byteBuddy )

        //Databases
-       testRuntime( libraries.h2 )
+       testRuntime( libraries.h2 ) { changing = true }
        testRuntime( libraries.hsqldb )
        testRuntime( libraries.postgresql )
        testRuntime( libraries.mysql )
diff --git a/gradle/libraries.gradle b/gradle/libraries.gradle
index cc53eb9448..8481c51267 100644
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -10,7 +10,7 @@
 ext {

     junitVersion = '4.12'
-    h2Version = '1.4.196'
+    h2Version = '1.4.201-SNAPSHOT'
     bytemanVersion = '4.0.8' //Compatible with JDK14
     jnpVersion = '5.0.6.CR1'

@@ -177,6 +177,9 @@ ext {
 }

 configurations.all {
+    resolutionStrategy {
+      cacheChangingModulesFor 0, 'seconds'
+    }
     resolutionStrategy.eachDependency { DependencyResolveDetails details ->
         //Force the "byte buddy agent" version to match the Byte Buddy version we use, as Mockito might p
         if (details.requested.group + ":" + details.requested.name == 'net.bytebuddy:byte-buddy-agent') {
[...]

PR #3520, opened 2020-08-25
on H2 after 2.0 don't need column type workaround (H5)

gavinking:
Apparently H2 correctly reports column types via JDBC metadata since 2.0, so this workaround is unnecessary.

Also add a default mapping for DECIMAL in superclass, just for completeness.

cc @famod WDYT? Does this look good?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
@@ -122,7 +122,7 @@ public class H2Dialect extends Dialect {
 		registerColumnType( Types.BIT, "boolean" );
 		registerColumnType( Types.CHAR, "char($l)" );
 		registerColumnType( Types.DATE, "date" );
-		registerColumnType( Types.DECIMAL, buildId >= 201 ? "numeric($p,$s)" : "decimal($p,$s)" );
+		registerColumnType( Types.DECIMAL, "decimal($p,$s)" );
 		registerColumnType( Types.NUMERIC, buildId >= 201 ? "numeric($p,$s)" : "decimal($p,$s)" );
 		registerColumnType( Types.DOUBLE, "double" );
 		registerColumnType( Types.FLOAT, "float" );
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
```diff
@@ -122,7 +122,7 @@ public H2Dialect() {
 		registerColumnType( Types.BIT, "boolean" );
 		registerColumnType( Types.CHAR, "char($l)" );
 		registerColumnType( Types.DATE, "date" );
-		registerColumnType( Types.DECIMAL, buildId >= 201 ? "numeric($p,$s)" : "decimal($p,$s)" );
+		registerColumnType( Types.DECIMAL, "decimal($p,$s)" );
 		registerColumnType( Types.NUMERIC, buildId >= 201 ? "numeric($p,$s)" : "decimal($p,$s)" );
 		registerColumnType( Types.DOUBLE, "double" );
 		registerColumnType( Types.FLOAT, "float" );
```
NathanQingyangXu, line 128: btw, there is a related ticket for the above `float' issue: https://github.com/hibernate/hibernate-orm/pull/2855

gavinking: `AbstractSchemaValidator.validateColumnType()` is the culprit here.

gavinking: And `Table.validateColumns()`, I suppose, except that that method doesn't seem to be called anywhere.

hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
```diff
@@ -122,7 +122,7 @@ public H2Dialect() {
 		registerColumnType( Types.BIT, "boolean" );
 		registerColumnType( Types.CHAR, "char($l)" );
 		registerColumnType( Types.DATE, "date" );
-		registerColumnType( Types.DECIMAL, buildId >= 201 ? "numeric($p,$s)" : "decimal($p,$s)" );
+		registerColumnType( Types.DECIMAL, "decimal($p,$s)" );
```
famod, line 125: I'm confused. Does this mean a part of #3405 was not necessary at all?
Or do we have the same problem as here: https://github.com/hibernate/hibernate-orm/pull/3519#discussion_r483961535?

gavinking: Well that was the result of the long discussion we just had, wasn't it? I mean, honestly I found it pretty hard to understand what he was saying, but after much back and forth it seems to me we agreed that the workaround wasn't required at all from .200 on.

beikov:
Thanks!

PR #3521, opened 2020-08-25
on H2 1.4.200 and above use localtime/localtimestamp (H5)

gavinking:
As suggested by @famod we need to use `localtime` instead of `current_time` because of changes in H2.

In particular the JDBC driver now refuses to convert `TIME/TIMESTAMP WITH TIME ZONE` to plain `TIME/TIMESTAMP`.

Does this look right @famod?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
@@ -210,8 +210,12 @@ public class H2Dialect extends Dialect {
 		registerFunction( "curtime", new NoArgSQLFunction( "curtime", StandardBasicTypes.TIME ) );
 		registerFunction( "curtimestamp", new NoArgSQLFunction( "curtimestamp", StandardBasicTypes.TIME ) );
 		registerFunction( "current_date", new NoArgSQLFunction( "current_date", StandardBasicTypes.DATE ) );
-		registerFunction( "current_time", new NoArgSQLFunction( "current_time", StandardBasicTypes.TIME ) );
-		registerFunction( "current_timestamp", new NoArgSQLFunction( "current_timestamp", StandardBasicTypes.TIMESTAMP ) );
+		// H2 made a nasty breaking change that changed the type of
+		// - current_timestamp to timestamp with time zone
+		// - current_time to time with time zone
+		// and also refuses to implicitly convert the type
+		registerFunction( "current_time", new NoArgSQLFunction( buildId >= 200 ? "localtime" : "current_time", StandardBasicTypes.TIME ) );
+		registerFunction( "current_timestamp", new NoArgSQLFunction( buildId >= 200 ? "localtimestamp" : "current_timestamp", StandardBasicTypes.TIMESTAMP ) );
 		registerFunction( "datediff", new StandardSQLFunction( "datediff", StandardBasicTypes.INTEGER ) );
 		registerFunction( "dayname", new StandardSQLFunction( "dayname", StandardBasicTypes.STRING ) );
 		registerFunction( "dayofmonth", new StandardSQLFunction( "dayofmonth", StandardBasicTypes.INTEGER ) );


hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
```diff
@@ -210,8 +210,8 @@ public H2Dialect() {
 		registerFunction( "curtime", new NoArgSQLFunction( "curtime", StandardBasicTypes.TIME ) );
 		registerFunction( "curtimestamp", new NoArgSQLFunction( "curtimestamp", StandardBasicTypes.TIME ) );
 		registerFunction( "current_date", new NoArgSQLFunction( "current_date", StandardBasicTypes.DATE ) );
-		registerFunction( "current_time", new NoArgSQLFunction( "current_time", StandardBasicTypes.TIME ) );
-		registerFunction( "current_timestamp", new NoArgSQLFunction( "current_timestamp", StandardBasicTypes.TIMESTAMP ) );
+		registerFunction( "current_time", new NoArgSQLFunction( buildId >= 200 ? "localtime" : "current_time", StandardBasicTypes.TIME ) );
```
famod, line 213: Shouldn't this (and the next line) be >=198 instead?
https://github.com/hibernate/hibernate-orm/pull/3412#issuecomment-678813877

My sentence
> As of 1.4.200 other functions also include the time zone and might fail as well but those changes might have been made after 1.4.198.
might have been a bit misleading. "As of 1.4.200" means that the current online documentation (which matches 1.4.200) is saying that even more functions include the time zone but I haven't checked whether those functions were added also in 198 or in 199 or 200.

gavinking: Well I know for sure that the both forms work fine on 1.4.199 because I've tested them both in H6.

It's a bit harder to test them on H5 because there are other distracting test failures.

gavinking: I mean the real problem here is that he's insisting on not implicitly casting `TIME WITH TIME ZONE` to `TIME`, a cast which I'm pretty sure is implicit on most other databases.

gavinking:
Note, the related issue is https://hibernate.atlassian.net/browse/HHH-14031.

beikov:
@dreab8 and I were discussing this and think this might cause a problem for users that try to extract the timezone from the current timestamp. Not sure if we should really merge it to master and 5.4 as it would not be backwards compatible. Wdyt @Sanne?

gavinking:
So I don't understand what's going on here, have we just decided to leave H2 support at 1.4.197 and never upgrade to 1.4.200?

Because the current situation, as I understand it, from what @famod has found, is that the a HQL query with `select current_timestamp` is broken on H2 1.4.200 (which was released back in 2019!) and that's why we're still stuck on 1.4.197.

And now this PR has been sitting here for 5 months.

It doesn't seem to me like ¯\\_(ツ)_/¯ is a terribly great resolution to the issue.

gavinking:
> @dreab8 and I were discussing this and think this might cause a problem for users that try to extract the timezone from the current timestamp. 

I don't really understand this objection. The change to `current_timestamp` that changed it to return `timestamp with time zone` was made in 1.4.198. This was the change that broke Hibernate. That's why we are currently stuck on H2 1.4.197.

<https://github.com/hibernate/hibernate-orm/pull/3412#issuecomment-678813877>

Older versions of H2 didn't include a timezone in `current_timestamp`, so how could users possibly be relying on that behavior?

The only scenario I can see is that they're using an unsupported later release of H2 with a version of Hibernate on which the `current_timestamp` function is broken, and they just haven't run into the brokenness yet.

So in order to not break people who're using an unsupported version of H2, you're proposing to never support that version of H2 _or any other future versions_.

That doesn't seem reasonable to me.

beikov:
>Older (i.e. supported) versions of H2 didn't include a timezone in current_timestamp, so how could users possibly be relying on that behavior?

I wasn't aware of that. In that case, I think this is fine. Could you please rebase the PR?

gavinking:
I rebased and added a comment.

PR #3522, opened 2020-08-25
HHH-11877 Wrap CompoundPredicate's expression list

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-11877

The issue can be summarized as following:

Suppose we have an entity class `Foo` that contains a boolean field `bar`, the following criteria building will produce invalid SQL:
```
cq.select( foo ).where( cb.and( cb.and(), foo.get( Foo_.bar ) ) );
```
whereas the following is working fine:
```
cq.select( foo ).where( foo.get( Foo_.bar ) );
```

The root cause is internally in `CriteriaBuilderImpl#wrap()` method we fixed boolean type Path to ensure creating SQL snippet like `Foo_.bar = 1` rather than `Foo_.bar` (which will trigger invalid SQL). But we didn't go deeper than root level and this PR fixed the loophole to fix non-root level boolean type Path as well.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/CompoundPredicate.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/CompoundPredicate.java
@@ -80,7 +80,10 @@ public class CompoundPredicate
 
 	private void applyExpressions(List<Expression<Boolean>> expressions) {
 		this.expressions.clear();
-		this.expressions.addAll( expressions );
+		final CriteriaBuilderImpl criteriaBuilder = criteriaBuilder();
+		for ( Expression<Boolean> expression : expressions ) {
+			this.expressions.add( criteriaBuilder.wrap( expression ) );
+		}
 	}
 
 	@Override
--- a/hibernate-core/src/test/java/org/hibernate/query/criteria/internal/hhh11877/Foo.java
+++ b/hibernate-core/src/test/java/org/hibernate/query/criteria/internal/hhh11877/Foo.java
@@ -1,0 +1,31 @@
+package org.hibernate.query.criteria.internal.hhh11877;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+
+@Entity
+public class Foo {
+
+	private long id;
+	private boolean bar;
+
+	@Column(nullable = false)
+	@GeneratedValue(strategy = GenerationType.IDENTITY)
+	@Id
+	public long getId() {
+		return this.id;
+	}
+	public void setId(final long id) {
+		this.id = id;
+	}
+
+	public boolean isBar() {
+		return this.bar;
+	}
+	public void setBar(final boolean bar) {
+		this.bar = bar;
+	}
+}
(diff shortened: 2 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/query/criteria/internal/CriteriaBuilderImpl.java
```diff
@@ -255,6 +255,12 @@ public Order desc(Expression<?> x) {
 
 	public Predicate wrap(Expression<Boolean> expression) {
 		if ( Predicate.class.isInstance( expression ) ) {
+			if ( CompoundPredicate.class.isInstance( expression ) ) {
```
beikov, line 258: IMO we should instead call wrap in CompoundPredicate when adding elements.

beikov (changes requested):

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/CriteriaBuilderImpl.java
NathanQingyangXu: You mean adding a new `wrap` method to `CompoundPredicate`? I did consider that but the code logic needs invoking the `CriteriaBuilderImpl#wrap()` recursively which makes the code difficult to read and understand. The current approach centralizes everything related to `wrapping` in this single method.

Another alternative is to add a new overloading method in the same class as following:
```
public Predicate wrap(Predicate predicate) {
        return predicate;
}

public Predicate wrap(CompoundPredicate compoundPredicate) {
        final List<Expression<Boolean>> expressions = compoundPredicate.getExpressions();
        for ( int i = 0; i < expressions.size(); i++ ) {
		expressions.set( i, wrap( expressions.get( i ) ) );
	}
        return compoundPredicate;
}
```

This approach seems to succeed in avoiding `Class#isInstance()` usage and to be a little bit more elegant, but it has downside in that the wrap method user might misuse the API. There is a good example of such misuse I encountered: https://github.com/jboss-logging/jboss-logging/pull/36.

NathanQingyangXu: I refactored as you suggested. There is both pros and cons, but if the new public method is useful in other scenarios than `CriteriaBuilderImpl`, it might be worthwhile.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/CompoundPredicate.java
```diff
@@ -125,6 +125,13 @@ public Predicate not() {
 		return new NegatedPredicateWrapper( this );
 	}
 
+	public void wrap() {
+		final CriteriaBuilderImpl criteriaBuilder = criteriaBuilder();
+		for ( int i = 0, size = expressions.size(); i < size; i++ ) {
+			expressions.set( i, criteriaBuilder.wrap( expressions.get( i ) ) );
```
beikov, line 131: Maybe it's better to do the wrapping in `applyExpressions` as that is the only code where expressions are added to the list.

beikov (changes requested):

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/CompoundPredicate.java
NathanQingyangXu: done. thanks.

beikov (approved):

PR #3523, opened 2020-08-25
HHH-14187 - Lazy loading fails if an entity appears twice in the object graph when enhancement as proxy is enabled

dreab8:
https://hibernate.atlassian.net/browse/HHH-14187


hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementAsProxyLazinessInterceptor.java
```diff
@@ -39,6 +39,8 @@
 
 	private boolean initialized;
 
+	private boolean inizializing;
```
Sanne, line 42: rather than 2 booleans `initialized` and `inizializing`, should we have a single three-state field, like an enum?

That would better restrict the combinations.

dreab8: yeap, definitively  better. Done. Thanks @Sanne

Sanne (approved):
great, thanks!

Sanne:
merged and backported

PR #3491, opened 2020-08-03
HHH-14137 Empty String check performance improvements

isadounikau:
https://hibernate.atlassian.net/browse/HHH-14137

PR #3524, opened 2020-08-26
HHH-14186 - @ColumnTransformer support

sebersole:


hibernate-core/src/main/java/org/hibernate/cfg/Ejb3Column.java
```diff
@@ -240,15 +240,17 @@ protected void initMappingColumn(
 			this.mappingColumn.setSqlType( sqlType );
 			this.mappingColumn.setUnique( unique );
 
-			if(writeExpression != null && !writeExpression.matches("[^?]*\\?[^?]*")) {
+			if ( writeExpression != null && !writeExpression.matches("[^?]*\\?[^?]*" ) ) {
```
NathanQingyangXu, line 243: There is a missing space char right after `.matches(`. The expression seems nontrivial. Will the above statement incur repeated performance cost to compile the regular expression?

hibernate-core/src/main/java/org/hibernate/mapping/Column.java
```diff
@@ -425,6 +435,11 @@ public String getCustomRead() {
 		return customRead;
 	}
 
+	public void setResolvedCustomRead(String customRead) {
+		assert customRead == null || ! customRead.trim().equals( "" );
```
NathanQingyangXu, line 439: As https://github.com/hibernate/hibernate-orm/pull/3491 suggested, we can replace
```
!customRead.trim().equals( "" );
```
with
```
customRead.trim().length() > 0;
```
As reported in https://medium.com/javarevisited/micro-optimizations-in-java-string-equals-22be19fd8416, the performance boot is nontrivial. @Sanne even created some util method for this.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/BasicValuedCollectionPart.java
```diff
@@ -57,21 +59,32 @@
...
 		this.valueConverter = valueConverter;
 		this.tableExpression = tableExpression;
 		this.columnExpression = columnExpression;
+
+		this.customColumnReadExpr = customColumnReadExpr == null
+				? columnExpression
+				: customColumnReadExpr;
+		this.customColumnWriteExpr = customColumnWriteExpr == null
+				? DEFAULT_COLUMN_WRITE_EXPRESSION
+				: customColumnWriteExpr;
```
NathanQingyangXu, line 87: We might well use the util method of `org.apache.commons.lang3.ObjectUtils#defaultIfNull()` to improve readability. The class seems present in the classpath.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationProcess.java
```diff
@@ -15,37 +15,44 @@
 import org.hibernate.metamodel.mapping.NonTransientException;
 import org.hibernate.metamodel.spi.RuntimeModelCreationContext;
 import org.hibernate.persister.entity.EntityPersister;
+import org.hibernate.query.sqm.function.SqmFunctionRegistry;
 
 /**
  * @author Steve Ebersole
  */
 public class MappingModelCreationProcess {
+	private final String EOL = System.getProperty( "line.separator" );
```
NathanQingyangXu, line 24: from Javadoc, `System.lineSeparator` always returns the same value as above.

hibernate-core/src/main/java/org/hibernate/mapping/Column.java
Sanne: Not super important, but it's worth pointing out that if we had JDK11, you could use:
 - https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#isBlank()

Of course I do agree that it's not worth dropping 8 compatibility just for this, but it's a good example of the small many things, which will eventually add up into a more efficient library.

I also think it's a good example because it's simply not worth it to have two(more) copies of this class if we wanted to use multi-release jars: I suppose we all agree that it would be insane to maintain two copies of the same sources just for such a little optimisation.

gavinking: Ahem. Folks, it should actually use `StringHelper.isEmptyOrWhitespace()`.

When we use util methods for stuff like this it makes it easy to optimize them later.

hibernate-core/src/main/java/org/hibernate/cfg/Ejb3Column.java
sebersole: However, I went a different approach and simply iterate the char values of the String counting the number of `?` chars by adding `StringHelper#count` which is similar to `StringHelper#countUnquoted`

hibernate-core/src/main/java/org/hibernate/mapping/Column.java
sebersole: As for performance... its an assert... ;)

sebersole: Well, actually @gavinking - the check is not whether the text is empty or a whitespace... the check is that the text is either null or non-empty ("" should fail the check here)

sebersole: Nm, you probably just mean the second condition

gavinking: ah, haha right, sorry :)

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/BasicValuedCollectionPart.java
sebersole: It is not part of the dependencies for hibernate-core.  The only usage of this package in the entire project is in the documentation project.  So we should not use that.

Actually though, looking at this again, this should not be doing this anyway.  It should be using the passed values strictly.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationProcess.java
sebersole: Pushed upstream with suggested changes

PR #3526, opened 2020-08-26
HHH-14152 fix the non deterministic grammar warning caused by changes to sql-script.g 

dreab8:
https://hibernate.atlassian.net/browse/HHH-14152

the warning seems caused by changes introduced by https://github.com/hibernate/hibernate-orm/pull/3502

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/fileimport/MultiLineImportExtractorTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/fileimport/MultiLineImportExtractorTest.java
@@ -33,9 +33,9 @@ public class MultiLineImportExtractorTest {
 	public void testExtraction() throws IOException {
 		final ClassLoader classLoader = ClassLoader.getSystemClassLoader();
 
-		try ( final InputStream stream = classLoader.getResourceAsStream( IMPORT_FILE ) ) {
+		try (final InputStream stream = classLoader.getResourceAsStream( IMPORT_FILE )) {
 			assertThat( stream, notNullValue() );
-			try ( final InputStreamReader reader = new InputStreamReader( stream ) ) {
+			try (final InputStreamReader reader = new InputStreamReader( stream )) {
 				final String[] commands = extractor.extractCommands( reader );
 				assertThat( commands, notNullValue() );
 				assertThat( commands.length, is( 6 ) );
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/test/fileimport/MultiLineImportExtractorTest.java
```diff
@@ -46,15 +46,15 @@ public void testExtraction() throws IOException {
 
 				assertThat( commands[1], is( "INSERT INTO test_data VALUES (1, 'sample')" ) );
 
-				assertThat( commands[2], is( "DELETE" + System.lineSeparator() + "  FROM test_data" ) );
+				assertThat( commands[2], is( "DELETE   FROM test_data" ) );
```
Sanne, line 49: so now we convert a newline into a standard whitespace? Was the intention not to also emit multi-line SQL, possibly for better logging?

dreab8: the issue is when we have something like
 ```
DELETE
FROM TEST DATA
```
where there is no space after `DELETE` and no space before `FROM`. by the way the `System.lineSeparator()` was introduced by the previous PR.

dreab8: probably another solution is to keep the grammar as it was before the  before https://github.com/hibernate/hibernate-orm/pull/3502 change and in some way detect and add the space on case it it needed in the `SqlScriptParser.out()` method.

gavinking: > so now we convert a newline into a standard whitespace? Was the intention not to also emit multi-line SQL, possibly for better logging?

I wonder what happens if hibernate.format_sql is enables when you use a multiline script? Do you get rubbish logging with random-looking newlines?

gavinking: (Before this latest change, I mean.)

dreab8: I don't know, I should try

Sanne: I might be wrong, but I don't think we apply SQL formatting on the import scripts?

dreab8:
improved by @sebersole and merged on master.

Sanne:
backported. Thanks all!

PR #3528, opened 2020-08-26
HHH-13058 Fix issue left join root cannot be replaced by correlated parent in subquery

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-13058

The current correlated subquery ignores the first left join and ends up with 'inner join' instead invariably. So suppose for the below entities:
```
@Entity(name = "Task")
public class Task {
    @Id
    Long id;

    @ManyToOne
    Patient patient;
}

@Entity(name = "Patient")
public class Patient {
    @Id
    Long id;

    @ManyToOne
    Site site;
}

@Entity(name = "Site")
public class Site {
    @Id
    Long id;
}
```
Note that `@ManyToOne`'s `optional` is `true` by default.

the following Criteria would skip the `Task` without `Patient`:
```
final Subquery<Task> subquery = outerQuery.subquery( Task.class );
final Root<Task> subtask = subquery.correlate( outerTask );
final From<Task, Patient> patient = subtask.join( Task_.patient, JoinType.LEFT );
final From<Patient, Site> site = patient.join( Patient_.site, JoinType.LEFT );
outerQuery.where(
    builder.exists(
        subquery.select( subtask )
            .where(
                builder.or(
                    patient.isNull(),
                    site.in( validSites )
                )
        )
    )
);
```
because internally the jpasql generated is:
```
select generatedAlias0 from Task as generatedAlias0 
where exists (
    select generatedAlias0 
    from generatedAlias0.patient as generatedAlias1 
        left join generatedAlias1.site as generatedAlias2 
    where ( generatedAlias1 is null ) or ( generatedAlias2 in (:param0, :param1) 
))
```
Note that `generatedAlias1 is null` can never be true for the left join from Task to Patient has been skipped by correlate parent's alias substitution; but given that Task's Patient is optional we should reserve the `left join` from `Task` to `Patient`. 
The fix in this PR is to create the following jpasql instead:
```
select generatedAlias0 from Task as generatedAlias0 
where exists (
    select generatedAlias0 
    from Task as generatedAlias1 
        left join generatedAlias1.patient as generatedAlias1 
        left join generatedAlias1.site as generatedAlias2 
    where 
        generatedAlias1 = generatedAlias0 and (( generatedAlias1 is null ) or ( generatedAlias2 in (:param0, :param1))
))
```
The basic idea is to generate new alias for correlated root and add `newAlias = oldAlias` to the `WHERE` jpasql clause, thus reserving the `LEFT JOIN` semantic of the correlated root in subquery.
However, the old pattern still remains for performance reason and the above changes only apply when `LEFT JOIN` is involved.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
@@ -318,18 +318,33 @@ public class QueryStructure<T> implements Serializable {
 						final FromImplementor correlationParent = correlationRoot.getCorrelationParent();
 						correlationParent.prepareAlias( renderingContext );
 						final String correlationRootAlias = correlationParent.getAlias();
-						for ( Join<?, ?> correlationJoin : correlationRoot.getJoins() ) {
-							final JoinImplementor correlationJoinImpl = (JoinImplementor) correlationJoin;
-							// IMPL NOTE: reuse the sep from above!
+						if ( correlationRoot.canBeReplacedByCorrelatedParentInSubQuery() ) {
+							for ( Join<?, ?> correlationJoin : correlationRoot.getJoins() ) {
+								final JoinImplementor correlationJoinImpl = (JoinImplementor) correlationJoin;
+								// IMPL NOTE: reuse the sep from above!
+								jpaqlQuery.append( sep );
+								correlationJoinImpl.prepareAlias( renderingContext );
+								jpaqlQuery.append( correlationRootAlias )
+										.append( '.' )
+										.append( correlationJoinImpl.getAttribute().getName() )
+										.append( " as " )
+										.append( correlationJoinImpl.getAlias() );
+								sep = ", ";
+								renderJoins( jpaqlQuery, renderingContext, correlationJoinImpl.getJoins() );
+							}
+						}
+						else {
+							correlationRoot.prepareAlias( renderingContext );
 							jpaqlQuery.append( sep );
-							correlationJoinImpl.prepareAlias( renderingContext );
-							jpaqlQuery.append( correlationRootAlias )
-									.append( '.' )
-									.append( correlationJoinImpl.getAttribute().getName() )
-									.append( " as " )
-									.append( correlationJoinImpl.getAlias() );
 							sep = ", ";
-							renderJoins( jpaqlQuery, renderingContext, correlationJoinImpl.getJoins() );
+							jpaqlQuery.append( correlationRoot.renderTableExpression( renderingContext ) );
+							renderJoins( jpaqlQuery, renderingContext, correlationRoot.getJoins() );
+							if ( correlationRoot instanceof Root ) {
+								Set<TreatedRoot> treats = ( (RootImpl) correlationRoot ).getTreats();
+								for ( TreatedRoot treat : treats ) {
+									renderJoins( jpaqlQuery, renderingContext, treat.getJoins() );
+								}
+							}
 						}
 					}
 				}
@@ -341,18 +356,46 @@ public class QueryStructure<T> implements Serializable {
 	}
 
 	protected void renderWhereClause(StringBuilder jpaqlQuery, RenderingContext renderingContext) {
-		if ( getRestriction() == null ) {
+		final String correlationRestrictionWhereFragment = getCorrelationRestrictionsWhereFragment();
+		if ( getRestriction() == null && correlationRestrictionWhereFragment.isEmpty() ) {
 			return;
 		}
 
 		renderingContext.getClauseStack().push( Clause.WHERE );
 		try {
-			jpaqlQuery.append( " where " )
-					.append( ( (Renderable) getRestriction() ).render( renderingContext ) );
+			jpaqlQuery.append( " where " );
+			jpaqlQuery.append( correlationRestrictionWhereFragment );
+			if ( getRestriction() != null ) {
+				if ( !correlationRestrictionWhereFragment.isEmpty() ) {
+					jpaqlQuery.append( " and ( " );
+				}
+				jpaqlQuery.append( ( (Renderable) getRestriction() ).render( renderingContext ) );
+				if ( !correlationRestrictionWhereFragment.isEmpty() ) {
+					jpaqlQuery.append( " )" );
+				}
+			}
 		}
 		finally {
 			renderingContext.getClauseStack().pop();
 		}
+	}
+
+	private String getCorrelationRestrictionsWhereFragment() {
+		if ( !isSubQuery || correlationRoots == null ) {
+			return "";
+		}
+		StringBuilder buffer = new StringBuilder();
+		String sep = "";
+		for ( FromImplementor<?, ?> correlationRoot : correlationRoots ) {
+			if ( !correlationRoot.canBeReplacedByCorrelatedParentInSubQuery() ) {
+				buffer.append( sep );
+				sep = " and ";
+				buffer.append( correlationRoot.getAlias() )
+						.append( "=" )
+						.append( correlationRoot.getCorrelationParent().getAlias() );
+			}
+		}
+		return buffer.toString();
 	}
 
 	protected void renderGroupByClause(StringBuilder jpaqlQuery, RenderingContext renderingContext) {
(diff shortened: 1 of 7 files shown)


NathanQingyangXu:
will open when it is finalized

NathanQingyangXu:
Note for v6 porting. The code changes might not be applicable to v6 and nontrivial v6 specific tweaking might be needed, but the testing case is invaluable.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/path/AbstractFromImpl.java
```diff
@@ -81,12 +81,7 @@ protected boolean canBeDereferenced() {
 	@Override
 	public void prepareAlias(RenderingContext renderingContext) {
 		if ( getAlias() == null ) {
```
beikov, line 83: I'm not sure this will generate the correct HQL. Maybe this should be the following
```suggestion
		if ( super.getAlias() == null ) {
```

and you should revert the other changes in this file

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
```diff
@@ -286,52 +287,17 @@ private FromImplementor locateImplicitSelection() {
...
-						renderFetches( jpaqlQuery, renderingContext, treat.getFetches() );
-					}
-				}
-				renderFetches( jpaqlQuery, renderingContext, root.getFetches() );
-			}
+			doRenderFrom( getRoots(), jpaqlQuery, renderingContext );
 
 			if ( isSubQuery ) {
 				if ( correlationRoots != null ) {
-					for ( FromImplementor<?, ?> correlationRoot : correlationRoots ) {
```
beikov, line 317: IMO this should be reverted and we should specially handle the left join case instead as this kind of querying might result in worse performance for some cases. A subquery like `select ... from alias.elementCollection` will only use the join table in the SQL. AFAIU, the changes you are doing here will always use the owner entity table in the from clause and then join the join table.

beikov (changes requested):
Requires some further work as the implementation produces SQL queries that aren't as efficient as before.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/path/AbstractFromImpl.java
NathanQingyangXu: I think it should be fine. I removed the `getAlias()` overloading method in this class (to ensure new alias is generated rather than reusing the correlated parent's alias), so both `getAlias()` and 'setAlias()` would be from the parent class.

But as you suggested, we need to revert back partially to the previous implementation for performance reason (only create new alias when the correlated root is of LEFT JOIN), then I need to reconsider the issue.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
NathanQingyangXu: yeah, that is the gist of the original implementation plan. I agree we should consider the LEFT JOIN correlated root as a special case to boost performance, though the downside is the code might be harder to read and maintain.

Will do as you suggested. Again, this PR might involve lots of back and forth.

NathanQingyangXu: After investigation, I found HHH-14197 has different root cause so I exclude that ticket outside this PR and will create its own PR in near future.

Updated as your suggestion. Thanks.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
```diff
@@ -341,20 +358,48 @@ private void renderFromClause(StringBuilder jpaqlQuery, RenderingContext renderi
...
 
 		renderingContext.getClauseStack().push( Clause.WHERE );
 		try {
-			jpaqlQuery.append( " where " )
-					.append( ( (Renderable) getRestriction() ).render( renderingContext ) );
+			jpaqlQuery.append( " where " );
+			jpaqlQuery.append( correlationRestrictionWhereFragment );
+			if ( getRestriction() != null ) {
+				if ( !correlationRestrictionWhereFragment.isEmpty() ) {
+					jpaqlQuery.append( " and (" );
```
beikov, line 372:
```suggestion
					jpaqlQuery.append( " and " );
```

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
```diff
@@ -341,20 +358,48 @@ private void renderFromClause(StringBuilder jpaqlQuery, RenderingContext renderi
...
+				}
+				jpaqlQuery.append( ( (Renderable) getRestriction() ).render( renderingContext ) );
+				if ( !correlationRestrictionWhereFragment.isEmpty() ) {
+					jpaqlQuery.append( ")" );
+				}
```
beikov, line 377:
```suggestion
```

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
```diff
@@ -318,18 +318,35 @@ private void renderFromClause(StringBuilder jpaqlQuery, RenderingContext renderi
...
-							renderJoins( jpaqlQuery, renderingContext, correlationJoinImpl.getJoins() );
+							jpaqlQuery.append( correlationRoot.renderTableExpression( renderingContext ) );
+							renderJoins( jpaqlQuery, renderingContext, correlationRoot.getJoins() );
+							if ( correlationRoot instanceof Root ) {
+								Set<TreatedRoot> treats = ( (RootImpl) correlationRoot ).getTreats();
+								for ( TreatedRoot treat : treats ) {
+									renderJoins( jpaqlQuery, renderingContext, treat.getJoins() );
+									renderFetches( jpaqlQuery, renderingContext, treat.getFetches() );
+								}
+								renderFetches( jpaqlQuery, renderingContext, correlationRoot.getFetches() );
```
beikov, line 348:
```suggestion
```

Subquery can't have fetches

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
```diff
@@ -318,18 +318,35 @@ private void renderFromClause(StringBuilder jpaqlQuery, RenderingContext renderi
...
-									.append( correlationJoinImpl.getAlias() );
 							sep = ", ";
-							renderJoins( jpaqlQuery, renderingContext, correlationJoinImpl.getJoins() );
+							jpaqlQuery.append( correlationRoot.renderTableExpression( renderingContext ) );
+							renderJoins( jpaqlQuery, renderingContext, correlationRoot.getJoins() );
+							if ( correlationRoot instanceof Root ) {
+								Set<TreatedRoot> treats = ( (RootImpl) correlationRoot ).getTreats();
+								for ( TreatedRoot treat : treats ) {
+									renderJoins( jpaqlQuery, renderingContext, treat.getJoins() );
+									renderFetches( jpaqlQuery, renderingContext, treat.getFetches() );
```
beikov, line 346:
```suggestion
```

Subquery can't have fetches

beikov (changes requested):
Looking good!

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
NathanQingyangXu: My concern is without the parenthesis, there might be issue if the original SQL is 'or' compound predicate. For instance, the original SQL might be:
```
where alias0.name = 'haha' or alias0.field1 is null
```
then without parenthesis, the following erroneous SQL would be created:
```
where alias1 = alias0 and alias0.name = 'haha' or alias0.name is null
```
No such concern exists?

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
NathanQingyangXu: deleted. thanks.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
NathanQingyangXu: thanks. deleted.

hibernate-core/src/main/java/org/hibernate/query/criteria/internal/QueryStructure.java
NathanQingyangXu: updated as you suggest

beikov (approved):

PR #3530, opened 2020-08-27
generate sensibly-lengthed columns for UUIDs

gavinking:
Previously, Hibernate generated column definitions of length 255 for `BINARY`/`CHAR` columns that will have `UUID`s stored in them. It should be:

- length 16 for `BINARY`
- length 36 for `CHAR`


gavinking:
Fixes https://hibernate.atlassian.net/browse/HHH-14195.

PR #3531, opened 2020-08-28
HHH-14199 fix gradle 'setDataBase' task failure

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14199


dreab8:
Thanks @NathanQingyangXu applied it upstream and also  to 5.4

PR #3533, opened 2020-08-31
HHH-14148 fix issue mapping-based order fragment might lead to incorrect SQL

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14148

The issue is invalid SQL like `... order by ... nulls last nulls last` would be generated for the combination of `@OrderBy on collection field` + `entity graph dictating the collection field to be fetched eagerly`.

So it turns out that the root cause is the following ANTLR2 snippet in sql-gen.g:

```
orderExprs { String ordExp = null; String ordDir = null; String ordNul = null; }
	// TODO: remove goofy space before the comma when we don't have to regression test anymore.
	// Dialect is provided a hook to render each ORDER BY element, so the expression is being captured instead of
	// printing to the SQL output directly. See Dialect#renderOrderByElement(String, String, String, NullPrecedence).
	: { captureExpressionStart(); } ( expr ) { captureExpressionFinish(); ordExp = resetCapture(); }
	    (dir:orderDirection { ordDir = #dir.getText(); })? (ordNul=nullOrdering)?
	        { out( renderOrderByElement( ordExp, ordDir, ordNul ) ); }
	    ( {out(", "); } orderExprs )?
	;
```

The above case only applies for explicit HQL order by processing. There is another implicit mapping-based order by SQL fragment to go through this matching rule (e.g. the order by fragment added by @OrderBy). In that case, the SQL order by fragment has been complete (including the potential null precedence like `nulls last`). Going through the above processing would add another null precedence fragment again, thus ending up with incorrect SQL.

The solution in this PR is simply to add another rule for such complete SQL order fragment case to have it printed intact.

### DIFF
--- a/hibernate-core/src/main/antlr/sql-gen.g
+++ b/hibernate-core/src/main/antlr/sql-gen.g
@@ -146,7 +146,7 @@ selectStatement
 		from
 		( #(WHERE { out(" where "); } whereExpr ) )?
 		( #(GROUP { out(" group by "); } groupExprs ( #(HAVING { out(" having "); } booleanExpr[false]) )? ) )?
-		( #(ORDER { out(" order by "); } orderExprs ) )?
+		( #(ORDER { out(" order by "); } orderBySqlFragmentOrExprs ) )?
 	)
 	;
 
@@ -189,6 +189,11 @@ whereClause
 whereClauseExpr
 	: (SQL_TOKEN) => conditionList
 	| booleanExpr[ false ]
+	;
+
+orderBySqlFragmentOrExprs
+	: sqlToken // for the purpose of mapping-defined orderBy SQL fragment
+	| orderExprs
 	;
 
 orderExprs { String ordExp = null; String ordDir = null; String ordNul = null; }
@@ -277,7 +282,7 @@ distinctOrAll
 	;
 
 countExpr
-	// Syntacitic predicate resolves star all by itself, avoiding a conflict with STAR in expr.
+	// Syntactic predicate resolves star all by itself, avoiding a conflict with STAR in expr.
 	: ROW_STAR { out("*"); }
 	| simpleExpr
 	;
--- a/hibernate-core/src/test/java/org/hibernate/test/annotations/onetomany/OrderByTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/annotations/onetomany/OrderByTest.java
@@ -24,15 +24,17 @@ import org.hibernate.Criteria;
 import org.hibernate.Hibernate;
 import org.hibernate.NullPrecedence;
 import org.hibernate.Session;
+import org.hibernate.cfg.AvailableSettings;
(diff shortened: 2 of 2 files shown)


beikov:
I didn't look too deep, but wouldn't the asc/desc part also be duplicated? Maybe we can find out the token type in `org.hibernate.hql.internal.ast.SqlGenerator#renderOrderByElement` and decide based on that whether to invoke the renderOrderByElement of the dialect.

NathanQingyangXu:
> I didn't look too deep, but wouldn't the asc/desc part also be duplicated? Maybe we can find out the token type in `org.hibernate.hql.internal.ast.SqlGenerator#renderOrderByElement` and decide based on that whether to invoke the renderOrderByElement of the dialect.

The asc/desc part is not duplicated. The root cause must has something to do with `entity graph` integration, which is not organic and has some edge case hidden.

> I didn't look too deep, but wouldn't the asc/desc part also be duplicated? Maybe we can find out the token type in `org.hibernate.hql.internal.ast.SqlGenerator#renderOrderByElement` and decide based on that whether to invoke the renderOrderByElement of the dialect.

The 'asc/desc' part is not duplicated. The duplicated 'nulls' only happens for `entity graph`, not for normal HQL. Guess `entity graph` is a later add-on and it was not organically integrated. Seems there is no way to detect the token type in `org.hibernate.hql.internal.ast.SqlGenerator#renderOrderByElement` for the `expression` parameter of the method is of string type (we can parse it but seems overkill). For some reason, only `entity graph` case contains `nulls` in `expression`.

Will investigate whether there is more elegant way to fix this.

NathanQingyangXu:
Need more time to come up with a much better solution.

NathanQingyangXu:
So it turns out that the root cause is the following ANTLR2 snippet in `sql-gen.g`:
```
orderExprs { String ordExp = null; String ordDir = null; String ordNul = null; }
	// TODO: remove goofy space before the comma when we don't have to regression test anymore.
	// Dialect is provided a hook to render each ORDER BY element, so the expression is being captured instead of
	// printing to the SQL output directly. See Dialect#renderOrderByElement(String, String, String, NullPrecedence).
	: { captureExpressionStart(); } ( expr ) { captureExpressionFinish(); ordExp = resetCapture(); }
	    (dir:orderDirection { ordDir = #dir.getText(); })? (ordNul=nullOrdering)?
	        { out( renderOrderByElement( ordExp, ordDir, ordNul ) ); }
	    ( {out(", "); } orderExprs )?
	;
```
The above case only applies for explicit HQL order by processing. There is another implicit mapping-based order by SQL fragment to go through this matching rule. In that case, the SQL order by fragment has been complete(including the potential nullness precedence like `nulls last`). Going through the above processing would add another nullness precedence fragment again, thus ending up with incorrect SQL.

The solution in this PR is simply to add another rule for such complete SQL order fragment case to have it printed intact.

PR #3536, opened 2020-09-02
introduce Dialect.equivalentTypes

gavinking:
Tools like `SchemaValidator` have problems when the database reports a different JDBC type code for a column to the one Hibernate uses for DDL generation. This has resulted in nasty workarounds in the dialect subclasses like mapping `Types.NUMERIC` to `decimal` or whatever.

The problem arises because many databases have fewer internal types than the set that JDBC defines. This change lets a `Dialect` treat certain JDBC type codes as equivalent.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -260,6 +260,37 @@ public abstract class Dialect implements ConversionContext {
 	}
 
 	/**
+	 * Do the given JDBC type codes, as defined in {@link Types} represent
+	 * essentially the same type in this dialect of SQL? The default
+	 * implementation treats {@link Types#NUMERIC NUMERIC} and
+	 * {@link Types#DECIMAL DECIMAL} as the same type, and
+	 * {@link Types#FLOAT FLOAT}, {@link Types#REAL REAL}, and
+	 * {@link Types#DOUBLE DOUBLE} as essentially the same type, since the
+	 * ANSI SQL specification fails to meaningfully distinguish them.
+	 *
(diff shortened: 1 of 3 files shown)


beikov (approved):

gavinking:
https://hibernate.atlassian.net/browse/HHH-14204

dreab8:
Thanks @gavinking for the head up, I will do it.

dreab8:
Thanks @gavinking I will do it.

Sanne:
Looks nice, but it's not clear to me which implications this brings; would you recommend we backport this to 5.4?

gavinking:
> Looks nice, but it's not clear to me which implications this brings; would you recommend we backport this to 5.4?

Well, I'm not sure. I don't see why not, since it doesn't really change the behavior of anything except the schema validator and it makes it more forgiving. So I don't see how it could be considered and *incompatible* change per se.

Sanne:
good point, I'll backport it. thanks!

Sanne:
merged

Sanne:
Just a note: there's also  `JdbcTypeFamilyInformation` which encodes related information. Maybe in 6 we could reorganize things a bit.

quaff:
Should `CLOB` be equivalent to `VARCHAR`?

gavinking:
> Should `CLOB` be equivalent to `VARCHAR`?

Unclear. It's almost like the sort of difference I would want the validador to give me a warning about but perhaps not an error?

And as for stuff like `CLOB` vs `LONGVARCHAR` and `BLOB` vs `LONGVARBINARY`, that's going to depend on the `Dialect`, which is why I made this as an operation of the `Dialect`.

quaff:
I agree warning better than error.

PR #3537, opened 2020-09-03
HHH-14205 Replace StringBuffer with StringBuilder

quaff:
avoid unnecessary synchronization


NathanQingyangXu:
I created a JIRA ticket for you: https://hibernate.atlassian.net/browse/HHH-14205. Without such associated ticket, it would be pretty hard to have such PR merged. Please update your title and git commit message accordingly. Thanks.

NathanQingyangXu (approved):
good catch

PR #3539, opened 2020-09-03
HHH-14201 Fix EntityJoinFromElement JOIN order issue

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14201

Screen Shot 2020-09-07 at 12 07 41 PM

Sometimes we generated SQL with wrong JOIN order. The root cause is we always reserve all the `EntityJoinFromElement`s without inserting them into AST tree,  until the last moment when we simply append them to the root level. However, it seems `EntityJoinFromElement` could show up anywhere and our special treatment of `EntityJoinFromElement` ends up with wrong SQL (some dbs could optimize by reverting back join order; whereas others simply accept it as it is and end up with error).

The solution is to insert all `EntityJoinFromElement`s back into the AST tree to ensure their original position or depth-first traversal order.

### DIFF
--- a/hibernate-core/src/main/antlr/sql-gen.g
+++ b/hibernate-core/src/main/antlr/sql-gen.g
@@ -309,12 +309,13 @@ fromTable
 	// Write the table node (from fragment) and all the join fragments associated with it.
 	: #( a:FROM_FRAGMENT  { out(a); } (tableJoin [ a ])* { fromFragmentSeparator(a); } )
 	| #( b:JOIN_FRAGMENT  { out(b); } (tableJoin [ b ])* { fromFragmentSeparator(b); } )
-	| #( e:ENTITY_JOIN    { out(e); } (tableJoin [ e ])* { fromFragmentSeparator(e); } )
+	| #( c:ENTITY_JOIN    { out(c); } (tableJoin [ c ])* { fromFragmentSeparator(c); } )
 	;
 
 tableJoin [ AST parent ]
-	: #( c:JOIN_FRAGMENT { out(" "); out(c); } (tableJoin [ c ] )* )
-	| #( d:FROM_FRAGMENT { nestedFromFragment(d,parent); } (tableJoin [ d ] )* )
+	: #( d:JOIN_FRAGMENT { out(" "); out(d); } (tableJoin [ d ] )* )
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/antlr/sql-gen.g
```diff
@@ -309,12 +309,13 @@ fromTable
...
+	| #( c:ENTITY_JOIN    { out(c); } (tableJoin [ c ])* { fromFragmentSeparator(c); } )
 	;
 
 tableJoin [ AST parent ]
-	: #( c:JOIN_FRAGMENT { out(" "); out(c); } (tableJoin [ c ] )* )
-	| #( d:FROM_FRAGMENT { nestedFromFragment(d,parent); } (tableJoin [ d ] )* )
+	: #( d:JOIN_FRAGMENT { out(" "); out(d); } (tableJoin [ d ] )* )
+	| #( e:FROM_FRAGMENT { nestedFromFragment(e,parent); } (tableJoin [ e ] )* )
+	| #( f:ENTITY_JOIN   { out(" "); out(f); } (tableJoin [ f ] )* )
 	;
```
NathanQingyangXu, line 319: The above changes include:
* add `ENTITY_JOIN` to `tableJoin` so it can show up anywhere and SQL generated won't ignore it;
* change the variable name to better reflect that `ENTITY_JOIN` is not special (original naming implies that it is only at the top level).

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromClause.java
```diff
@@ -61,8 +61,6 @@
 	 */
 	private List impliedElements = new LinkedList();
 
-	private List<EntityJoinFromElement> entityJoinFromElements;
```
NathanQingyangXu, line 64: Now the above field has been redundant for it is included in `fromElements` which reserves join order via `LinkedHashSet`.

NathanQingyangXu:
I closed it for I need more time to analyze and fix it in much better way.

NathanQingyangXu:
I closed the wrong PR. This one is ready for review. Reopened.

beikov:
Great work! Thanks Nathan 👍

PR #3540, opened 2020-09-03
HHH-14197 Fix correlated subquery empty 'from' SQL issue

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14197

The root cause is we didn't test non-root correlated parent and some fix to FromElementFactory is needed to avoid setting text to emptiness aggressively.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementFactory.java
@@ -226,7 +226,7 @@ public class FromElementFactory implements SqlTokenTypes {
 		}
 
 		if ( explicitSubqueryFromElement ) {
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromElementFactory.java
```diff
@@ -226,7 +226,7 @@ else if ( elementType.isComponentType() ) {
 		}
 
 		if ( explicitSubqueryFromElement ) {
-			// Treat explict from elements in sub-queries properly.
+			// Treat expliict from elements in sub-queries properly.
```
andrei-ivanov, line 229: still wrong 😀

NathanQingyangXu: thanks. corrected.

PR #3543, opened 2020-09-07
clean up Dialect-specific lock syntax code (H6)

gavinking:
This code was very convoluted and difficult to follow and had multiple bugs.

Also it did not support the `WAIT n` syntax on MariaDB.

Note that I worked from the documentation here, and I have *not* tested this code on MariaDB!


gavinking:
@dreab8 Even if I can't guarantee that this code isn't quite perfect and has zero bugs, I doubt it has more bugs than the current code, which must have grown organically over lots of tiny commits and must never have been properly reviewed as a whole. If it passes the CI build, I recommend integrating it.

dreab8:
Thanks @gavinking applied upstream!

gavinking:
Cheers￼

PR #3545, opened 2020-09-08
HHH-14210 locking and DB2 (H6)

gavinking:
Contains the following changes:

- support for `skip locked data` on DB2
- support for `use and keep share locks` on DB2
- removal of support for `for update` in `IngresDialect` and RDMS, since those databases only support `for update` in cursors, AFAICS
- fixed documentation of `Dialect.forUpdateOfColumns()`, which was completely wrong about its semantics

Fixes HHH-14210

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.dialect;
 
+import org.hibernate.LockOptions;
 import org.hibernate.NullPrecedence;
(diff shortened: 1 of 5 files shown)


gavinking:
Useful link, FTR: <https://www.ibm.com/support/knowledgecenter/SSEPEK_12.0.0/sqlref/src/tpc/db2z_sql_isolationclause.html>

gavinking:
Actually `for update` on DB2 seems to be for use with cursors only.

gavinking:
@dreab8 WDYT?

dreab8:
It looks good to me.
Applied upstream, Thanks @gavinking,

gavinking:
Great, thanks.

PR #3546, opened 2020-09-10
HHH-14212 Fix Fetch Graph ultimately

NathanQingyangXu:

- revert back all the previous PRs related to Fetch Graph implementation and bug fixing (HHH-8776, HHH-14097, HHH-14212);


yrodiere:
From what I can see, the problem lies in the fact that we reset the entity graph context between after each row, but a single entity may be hydrated from multiple rows, and that's the case here.

Now how will we determine whether a row is part of the same entity than the previous one... that remains to be seen. We don't have much context to work with in this part of the code, unfortunately. Still looking into this.

NathanQingyangXu:
close this PR in favour of another PR to reflect the new status

PR #3547, opened 2020-09-13
HHH 14212 ultimate Fetch Graph fix

NathanQingyangXu:

- revert back all the previous PRs related to Fetch Graph implementation and bug fixing (HHH-8776, HHH-14097, HHH-14124);

I intentionally organized the commit list to streamline code review. All the commits except the last one are reverting back previous PR commits. You can go to the last commit directly to understand the new implementation.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
@@ -415,11 +415,7 @@ public final class TwoPhaseLoad {
 	}
 
 	/**
-	 * Check if eager of the association is overridden (i.e. skipping metamodel strategy), including (order sensitive):
-	 * <ol>
-	 *     <li>fetch graph</li>
-	 *     <li>fetch profile</li>
-	 * </ol>
+	 * Check if eager of the association is overridden by anything.
 	 *
 	 * @param session session
 	 * @param entityName entity name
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
```diff
@@ -521,38 +502,9 @@ private static Boolean isEagerFetchProfile(SharedSessionContractImplementor sess
...
-			}
-			else {
-				// implicit 'fetch graph' applies, so fetch lazily
-				return false;
-			}
-		}
 
-		return null;
+	private static Boolean isEagerFetchGraph(SharedSessionContractImplementor session) {
+		return session.isEnforcingFetchGraph() ? false : null;
```
NathanQingyangXu, line 507: The above one-liner is the gist of the new implementation. We simply return false. Even if we return true for a Fetch Graph node, it won't make any difference for it will be loaded anyway (the corresponding JOIN has been present in the SQL generated). What really matters is the decision of entity not covered by entity graph.

hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
```diff
@@ -465,25 +446,14 @@ private static Boolean getOverridingEager(
...
-							associationName,
-							overridingEager
-					);
-				}
-
-				return overridingEager;
+			// we can return false invariably for if the entity has been covered by entity graph,
+			// its associated JOIN has been present in the SQL generated and hence it would be loaded anyway
+			if ( session.isEnforcingFetchGraph() ) {
+				return false;
```
NathanQingyangXu, line 452: The above one-liner is the gist of the new implementation.

NathanQingyangXu:
I have to close this one again in favour of another one

PR #3548, opened 2020-09-13
HHH-14212 Ultimate Fetch Graph fix

NathanQingyangXu:

- revert back all the previous PRs related to Fetch Graph implementation and bug fixing (HHH-8776, HHH-14097, HHH-14124);

I intentionally organized the commit list to streamline code review. All the commits except the last one are reverting back previous PR commits. You can go to the last commit directly to understand the new implementation: https://github.com/hibernate/hibernate-orm/pull/3548/commits/46856068ebd952d181e6314011b782b7b0d80e82

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
@@ -32,8 +32,6 @@ import org.hibernate.event.spi.PostLoadEvent;
 import org.hibernate.event.spi.PostLoadEventListener;
 import org.hibernate.event.spi.PreLoadEvent;
 import org.hibernate.event.spi.PreLoadEventListener;
-import org.hibernate.graph.spi.AttributeNodeImplementor;
-import org.hibernate.graph.spi.GraphImplementor;
 import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.pretty.MessageHelper;
@@ -207,8 +205,6 @@ public final class TwoPhaseLoad {
 		String[] propertyNames = persister.getPropertyNames();
 		final Type[] types = persister.getPropertyTypes();
 		
-		final GraphImplementor<?> fetchGraphContext = session.getFetchGraphLoadContext();
-		
 		for ( int i = 0; i < hydratedState.length; i++ ) {
 			final Object value = hydratedState[i];
 			if ( debugEnabled ) {
@@ -256,10 +252,6 @@ public final class TwoPhaseLoad {
 				if ( debugEnabled ) {
 					LOG.debugf( "Skipping <unknown> attribute : `%s`", propertyNames[i] );
 				}
-			}
-
-			if ( session.getFetchGraphLoadContext() != fetchGraphContext ) {
-				session.setFetchGraphLoadContext( fetchGraphContext );
 			}
 		}
 	}
@@ -414,11 +406,7 @@ public final class TwoPhaseLoad {
 	}
 
 	/**
-	 * Check if eager of the association is overridden (i.e. skipping metamodel strategy), including (order sensitive):
-	 * <ol>
-	 *     <li>fetch graph</li>
-	 *     <li>fetch profile</li>
-	 * </ol>
+	 * Check if eager of the association is overridden by anything.
 	 *
 	 * @param session session
 	 * @param entityName entity name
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/engine/internal/TwoPhaseLoad.java
```diff
@@ -465,25 +428,14 @@ private static Boolean getOverridingEager(
...
-							associationName,
-							overridingEager
-					);
-				}
-
-				return overridingEager;
+			// we can return false invariably for if the entity has been covered by entity graph,
+			// its associated JOIN has been present in the SQL generated and hence it would be loaded anyway
+			if ( session.isEnforcingFetchGraph() ) {
+				return false;
```
NathanQingyangXu, line 434: The above one-liner is the gist of the ultimate fix. Its comment is self-exlanatory.

yrodiere (approved):
Ok, I understand better. Thanks for the clarification. So essentially this leverages the pre-existing "load graph" implementation, adding just one flag in the session to let us know whether we're in "fetch graph" mode or "load graph" mode, and according to that flag we'll decide whether attributes that are not mentioned in the graph are forced to lazy or left to their default value.

Looks good to me!

beikov (approved):
Looking good!

beikov:
Do we want/need anyone else to review this? If not, I'd say we merge it! 😃

NathanQingyangXu:
> Ok, I understand better. Thanks for the clarification. So essentially this leverages the pre-existing "load graph" implementation, adding just one flag in the session to let us know whether we're in "fetch graph" mode or "load graph" mode, and according to that flag we'll decide whether attributes that are not mentioned in the graph are forced to lazy or left to their default value.
> 
> Looks good to me!

The session tweaking is not new and it was applied for my previous fix direction (in HHH-8776). What is really inspiring is the simple logic to return 'false' always regardless of whether the entity should be eagerly loaded or not (we have added JOINs for those entity graph nodes already so they will be loaded sooner or later; what really matters is the NEW nodes). The inspiration occurred to me when I was accompanying my little kid in playground. It only took 15 mins to verify my idea. I am really excited by its beauty and simplicity.

gavinking:
@NathanQingyangXu so I suppose this one needs to be replicated in Hibernate Reactive?

gavinking:
Perhaps it even breaks HR?

senthilaru:
When it will be available to the public? I mean the release version.

NathanQingyangXu:
> When it will be available to the public? I mean the release version.

guess in next release. @beikov could we have it merged?

beikov:
What are the concerns here @gavinking ? Do you think that merging this PR will cause problems for HR?

gavinking:
No, I'm just saying that we will (I think) need to make changes to HR once this is merged.

beikov:
I see, thanks for the fast response. I'll merge this then :)

NathanQingyangXu:
> No, I'm just saying that we will (I think) need to make changes to HR once this is merged.

Feel free to ask for my help on that, though I have zero experience on HR right now.

gavinking:
@NathanQingyangXu why does this patch remove type arguments and go back there to using *raw types*?

gavinking:
> Feel free to ask for my help on that, though I have zero experience on HR right now.

@NathanQingyangXu Well for a start we need to just test HR against the latest ORM snapshot.

NathanQingyangXu:
> @NathanQingyangXu why does this patch remove type arguments and go back there to using _raw types_?

That change is done in yrodiere's PR, which has nothing to do with entity graph. I reverted his PR totally so this change is reverted as well. This PR involves many reverting and I need to be very cautious so I focus on the core issue instead.

gavinking:
> I reverted his PR totally so this change is reverted as well. 

Right, OK, gotcha.

gavinking:
HR issue is here https://github.com/hibernate/hibernate-reactive/issues/384

yrodiere:
@NathanQingyangXu @beikov Shouldn't this be backported to the 5.4 branch, since the bug was reported on 5.4?

beikov:
@yrodiere I proposed this to @Sanne and I think he is in the process of testing Quarkus with these changes.

NathanQingyangXu:
@Sanne many people are asking whether we can back port it into 5.4 for it is a regression issue. Seems failure to back port it might hurt our credit for entity graph is a enormous popular feature of JPA.

Sanne:
thanks @NathanQingyangXu , yes I'll backport this. Sorry I've been a bit swamped, hope to "batch" a series of backports so to test them all together in downstream projects.

chrisknoll:
Has this been back ported?  We're seeing it on v5.4.21, so we're having to downgrade to v5.4.18, but there are other vulnerabilities that were addressed post v5.4.18 so we'd really like to use the latest release int he 5.4.x release.

yrodiere:
See https://hibernate.atlassian.net/browse/HHH-14212 . It's been backported in 5.4.22.

PR #3550, opened 2020-09-14
HHH-14201 fixup

beikov:


NathanQingyangXu (approved):

hibernate-core/src/main/antlr/sql-gen.g
```diff
@@ -191,18 +191,15 @@ whereClauseExpr
...
 	// TODO: remove goofy space before the comma when we don't have to regression test anymore.
 	// Dialect is provided a hook to render each ORDER BY element, so the expression is being captured instead of
 	// printing to the SQL output directly. See Dialect#renderOrderByElement(String, String, String, NullPrecedence).
-	: { captureExpressionStart(); } ( expr ) { captureExpressionFinish(); ordExp = resetCapture(); }
+	: { captureExpressionStart(); } ( e:expr ) { captureExpressionFinish(); ordExp = resetCapture(); }
 	    (dir:orderDirection { ordDir = #dir.getText(); })? (ordNul=nullOrdering)?
-	        { out( renderOrderByElement( ordExp, ordDir, ordNul ) ); }
+	    // SQL Tokens without a direction and null ordering can be passed through as-is.
+	    // These tokens could be mapping defined order by fragments which are already rendered via the dialect hook
+	        { out( #e.getType() == SQL_TOKEN && ordDir == null && ordNul == null ? ordExp : renderOrderByElement( ordExp, ordDir, ordNul ) ); }
```
NathanQingyangXu, line 202: I might be wrong. It seems the `SQL_TOKEN` usually implies null `ordDir` and `ordNul`. Also, you might need to update the comment as well.

beikov: You are right! Thanks for the hint

hibernate-core/src/main/antlr/sql-gen.g
```diff
@@ -191,18 +191,15 @@ whereClauseExpr
...
 
-orderBySqlFragmentOrExprs
-	: sqlToken // for the purpose of mapping-defined orderBy SQL fragment
-	| orderExprs
-	;
-
 orderExprs { String ordExp = null; String ordDir = null; String ordNul = null; }
 	// TODO: remove goofy space before the comma when we don't have to regression test anymore.
 	// Dialect is provided a hook to render each ORDER BY element, so the expression is being captured instead of
 	// printing to the SQL output directly. See Dialect#renderOrderByElement(String, String, String, NullPrecedence).
```
NathanQingyangXu, line 197: so the above comment only applies in HQL. Now we allow for another possibility that "printing to the SQL output directly". We need to add more explanation to the comment above, for it is not accurate any more.

PR #3552, opened 2020-09-15
HHH-12012 UnsupportedOperationException thrown for Stateless session query

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-12012


NathanQingyangXu:
close this PR unless there is some feedback from the bug reporter.

Shrexpecial:
A few days ago, I ran into the same issue while migrating our system from hibernate-core 5.1 to 5.4. However, correct me if I'm wrong, but this does not provide a resolution to the issue and upon execution of QueryImpl.scroll() (or QueryImpl.list() for that matter), an UnsupportedOperationException is thrown. There does not seem to be a way of explicitly setting neither cacheStoreMode nor cacheRetrieveMode to null, as setCacheMode calls in turn the respective CacheHelper methods.

PR #3553, opened 2020-09-15
Add syntax highlighting to the logged SQL

gavinking:
Using ANSI escape codes, we can make the logged SQL a little more readable.

For https://hibernate.atlassian.net/browse/HHH-14217

### DIFF
--- a/documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
+++ b/documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
@@ -621,6 +621,9 @@ Write all SQL statements to the console. This is an alternative to setting the l
 `*hibernate.format_sql*` (e.g. `true` or `false` (default value))::
 Pretty-print the SQL in the log and console.
 
+`*hibernate.highlight_sql*` (e.g. `true` (default value) or `false`)::
+Colorize the SQL in the console using ANSI escape codes.
+
 `*hibernate.use_sql_comments*` (e.g. `true` or `false` (default value))::
 If true, Hibernate generates comments inside the SQL, for easier debugging.
 
--- a/documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
+++ b/documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
@@ -621,7 +621,7 @@ Write all SQL statements to the console. This is an alternative to setting the l
 `*hibernate.format_sql*` (e.g. `true` or `false` (default value))::
 Pretty-print the SQL in the log and console.
 
-`*hibernate.highlight_sql*` (e.g. `true` (default value) or `false`)::
+`*hibernate.highlight_sql*` (e.g. `true` or `false` (default value))::
 Colorize the SQL in the console using ANSI escape codes.
 
 `*hibernate.use_sql_comments*` (e.g. `true` or `false` (default value))::
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/HighlightingFormatter.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/HighlightingFormatter.java
@@ -1,0 +1,105 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.engine.jdbc.internal;
+
+import org.hibernate.engine.jdbc.env.spi.AnsiSqlKeywords;
+import org.hibernate.internal.util.StringHelper;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.StringTokenizer;
+
+/**
+ * Performs basic syntax highlighting for SQL using ANSI escape codes.
+ *
+ * @author Gavin King
+ */
+public class HighlightingFormatter implements Formatter {
+
+	private static final Set<String> KEYWORDS = new HashSet<>( AnsiSqlKeywords.INSTANCE.sql2003() );
+	static {
+		// additional keywords not reserved by ANSI SQL 2003
+		KEYWORDS.addAll( Arrays.asList( "KEY", "SEQUENCE", "CASCADE", "INCREMENT" ) );
+	}
+
+	public static final Formatter INSTANCE =
+			new HighlightingFormatter(
+					"34", // blue
+					"36", // cyan
+					"32"
+			);
+
+	private static String escape(String code) {
+		return "\u001b[" + code + "m";
+	};
+
+	private final String keywordEscape;
+	private final String stringEscape;
+	private final String quotedEscape;
+	private final String normalEscape;
+
+	/**
+	 * @param keywordCode the ANSI escape code to use for highlighting SQL keywords
+	 * @param stringCode the ANSI escape code to use for highlighting SQL strings
+	 */
+	public HighlightingFormatter(String keywordCode, String stringCode, String quotedCode) {
+		keywordEscape =escape(keywordCode);
+		stringEscape = escape(stringCode);
+		quotedEscape = escape(quotedCode);
+		normalEscape = escape("0");
+	}
+
+	@Override
+	public String format(String sql) {
+		String symbolsAndWs = "=><!+-*/()',.|&`\"?" + StringHelper.WHITESPACE;
+		StringBuilder result = new StringBuilder();
+		boolean inString = false;
+		boolean inQuoted = false;
+		for (StringTokenizer tokenizer = new StringTokenizer( sql, symbolsAndWs, true );
+				tokenizer.hasMoreTokens(); ) {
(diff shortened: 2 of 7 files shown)


gavinking:
Do we need some way to turn this off?

NathanQingyangXu:
I really like the idea and this is the first PR I even cloned locally to inspect with curiosity. Currently it seems the formatting interferes with testing and most of the testing case failed (at the `drop table if exists ...` statement which has been changed to invalid SQL ending up with exception).
But this is a really cool feature!

hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/HighlightingFormatter.java
```diff
@@ -0,0 +1,30 @@
...
+ * Performs basic syntax highlighting for SQL using ANSI escape codes.
+ *
+ * @author Gavin King
+ */
+public class HighlightingFormatter implements Formatter {
+
+    private static final Pattern keywords =
+            Pattern.compile("\\b(create|drop|alter|add|table|sequence|primary|foreign|unique|key|constraint|cascade|references|not|null|if|exists|select|as|delete|insert|update|set|from|in|into|values|left|right|inner|outer|join|on|where|having|order|group|by|desc|asc)\\b");
+    private static final Pattern strings =
+            Pattern.compile("'[^']*'");
```
NathanQingyangXu, line 15: Guess we don't need to make everything perfect, but RE based matching might be broken for the following SQL:
```
update book b set b.title = 'Bob Dylan\'s dream' where b.id = 1;
```

gavinking: Yeah look, I'm not sure if it's really worth trying to make it completely perfect, but in principle I could ultimately get rid of the regexes and write a proper tokenizer for SQL. That's not hard.

gavinking: Oh and also the correct way to escape a ' in a SQL string is '', isn't it?

gavinking:
So we have tests which look at the output of the logged SQL?

Because the formatters shouldn't be affecting the SQL sent to the database should they?

gavinking:
> Because the formatters shouldn't be affecting the SQL sent to the database should they?

Oh for DDL the formatted SQL is executed?!

Is that ... *supposed* to work like that?

hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/HighlightingFormatter.java
NathanQingyangXu: Yeah. It is double single quotes.

NathanQingyangXu:
> > Because the formatters shouldn't be affecting the SQL sent to the database should they?
> 
> Oh for DDL the formatted SQL is executed?!
> 
> Is that ... _supposed_ to work like that?

Saw you have fixed the testing case.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/HighlightingFormatter.java
NathanQingyangXu: Yeah, if it is really worth trying to make it perfect, we might well get formatting info from SQL generation phase, which might be more robust than RE. Anyway, if something imperfect is released, we might need to fix some bug reported by our users sooner or later.

gavinking: I've changed it to use a handwritten tokenizer.

NathanQingyangXu:
It is definitively pretty cool and helpful for developer inspecting SQL in IDE, but I am not sure formatting is needed in all scenarios (like in high performance server side), so I think adding option to allow user to choose whether to format doesn't hurt. Maybe we can even go as far as disabling it by default, though personally I would enable this cool feature without any hesitation for sure.

NathanQingyangXu (approved):

gavinking:
>  think adding option to allow user to choose whether to format doesn't hurt. 

Well I just checked and it seems that editors and even `less` and `more` don't know what to do with ANSI escape codes, and since there doesn't seem to be any way to automatically detect that we're logging to a file, I guess we will need a way to turn it off.

gavinking:
> I guess we will need a way to turn it off

Alternatively I could limit it to be a feature of `hibernate.show_sql`.

gavinking:
> > I guess we will need a way to turn it off
> 
> Alternatively I could limit it to be a feature of `hibernate.show_sql`.

I've added `hibernate.highlight_sql` to enable this. But I'm thinking that probably it should always be enabled for output in the console from `hibernate.show_sql`.

gavinking:
Upon reflection, I've decided that this only really makes sense for the console output with `hibernate.show_sql`, and doesn't really make sense for the log. So for `hibernate.show_sql` I've turned it on by default, but let you explicitly disable it in case it's a problem.

quaff:
It should ignore formatting if no console.
see `noConsoleNoAnsi=true` https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/pattern/HighlightConverter.html

gavinking:
> It should ignore formatting if no console.

OK, good idea. I've made that change.

gavinking:
> > It should ignore formatting if no console.
> 
> OK, good idea. I've made that change.

Actually, no, scratch that, that actually *doesn't* seem to be the right thing to do, since the console is actually `null` when running tests inside IntelliJ.

quaff:
> @quaff
> 
> > > It should ignore formatting if no console.
> > 
> > 
> > OK, good idea. I've made that change.
> 
> Actually, no, scratch that, that actually _doesn't_ seem to be the right thing to do, since the console is actually `null` when running tests inside IntelliJ.

It will be mess in eclipse <strong>or</strong> windows cmd, maybe there is no better way.

gavinking:
As I understand it, windows terminal now supports ANSI escapes.

I didn't know about the Eclipse console, does it really not?!

gavinking:
> I didn't know about the Eclipse console, does it really not?!

Ugh. Apparently you need a plug-in :(

hibernate-core/src/main/java/org/hibernate/engine/jdbc/spi/SqlStatementLogger.java
```diff
@@ -120,14 +133,16 @@ public void logStatement(String statement) {
...
+				&& ( logToStdout || LOG.isDebugEnabled() ) ) {
+			statement = formatter.format( statement );
 		}
 		LOG.debug( statement );
 		if ( logToStdout ) {
-			System.out.println( "Hibernate: " + statement );
+			if ( highlight ) {
+				statement = FormatStyle.HIGHLIGHT.getFormatter().format( statement );
+			}
+			System.out.println( "\u001b[35m[Hibernate]\u001b[0m " + statement );
```
quaff, line 145: If not highlight you should not print Ansi escapes.

quaff (commented):
> As I understand it, windows terminal now supports ANSI escapes.
> 
> I didn't know about the Eclipse console, does it really not?!

Windows `cmd` doesn't supports ANSI escapes, I really like this feature, but I think it should be disabled by default.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/spi/SqlStatementLogger.java
gavinking: Fixed.

gavinking:
> Windows `cmd` doesn't supports ANSI escapes

Fine, so it's now disabled by default.

Sanne:
Very nice!  About the default, I believe maybe Quarkus already has a general way to know if the output supports it?  I hope we could tie into that so to switch it on by default in Quarkus when it hints this won't break things?

@maxandersen , @gsmet ?

maxandersen:
There isn't really a reliable way to know. 

Currently we disable quarkus ansi logging if running on windows no matter if there is a console or not. This is done very early on before any config set/configured if I recall. 

I agree we should try and see if those flags can be matched / aligned.

Sanne:
thanks @maxandersen . I suppose that if Quarkus is emitting ansi codes, it's fine if Hibernate does so as well; to be continued on the Quarkus tracker ;)

documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
```diff
@@ -621,6 +621,9 @@ Write all SQL statements to the console. This is an alternative to setting the l
 `*hibernate.format_sql*` (e.g. `true` or `false` (default value))::
 Pretty-print the SQL in the log and console.
 
+`*hibernate.highlight_sql*` (e.g. `true` (default value) or `false`)::
```
yrodiere, line 624:
```suggestion
`*hibernate.highlight_sql*` (e.g. `true` or `false` (default value))::
```

Sanne: thanks @yrodiere !

Sanne:
Merged and backported. Thanks!

gavinking:
@Sanne thanks.

PR #3555, opened 2020-09-17
Stop generating fragments of uppercase SQL

gavinking:
Hibernate generates lowercase SQL, because it isn't 1980 anymore, and hasn't been for a while.

(I already fixed this in H6, this fix is for H5.)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/InterbaseDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/InterbaseDialect.java
@@ -77,7 +77,7 @@ public class InterbaseDialect extends Dialect {
 
 	@Override
 	public String getSequenceNextValString(String sequenceName) {
-		return "select " + getSelectSequenceNextValString( sequenceName ) + " from RDB$DATABASE";
+		return "select " + getSelectSequenceNextValString( sequenceName ) + " from rdb$database";
 	}
 
 	@Override
@@ -92,12 +92,12 @@ public class InterbaseDialect extends Dialect {
 
 	@Override
 	public String getDropSequenceString(String sequenceName) {
-		return "delete from RDB$GENERATORS where RDB$GENERATOR_NAME = '" + sequenceName.toUpperCase(Locale.ROOT) + "'";
+		return "delete from rdb$generators where rdb$generator_name = '" + sequenceName.toUpperCase(Locale.ROOT) + "'";
 	}
 
 	@Override
 	public String getQuerySequencesString() {
-		return "select RDB$GENERATOR_NAME from RDB$GENERATORS";
+		return "select rdb$generator_name from rdb$generators";
 	}
 
 	@Override
@@ -150,7 +150,7 @@ public class InterbaseDialect extends Dialect {
 		// TODO : not sure which (either?) is correct, could not find docs on how to do this.
 		// did find various blogs and forums mentioning that select CURRENT_TIMESTAMP
 		// does not work...
-		return "{?= call CURRENT_TIMESTAMP }";
+		return "{?= call current_timestamp }";
 //		return "select CURRENT_TIMESTAMP from RDB$DATABASE";
 	}
 
--- a/hibernate-core/src/main/java/org/hibernate/dialect/MariaDB103Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/MariaDB103Dialect.java
@@ -57,7 +57,7 @@ public class MariaDB103Dialect extends MariaDB102Dialect {
 
 	@Override
 	public String getQuerySequencesString() {
-		return "select table_name from information_schema.TABLES where table_schema = database() and table_type = 'SEQUENCE'";
+		return "select table_name from information_schema.tables where table_schema = database() and table_type = 'SEQUENCE'";
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Oracle8iDialect.java
@@ -745,7 +745,7 @@ public class Oracle8iDialect extends Dialect {
 
 	@Override
 	public String getCurrentSchemaCommand() {
-		return "SELECT SYS_CONTEXT('USERENV', 'CURRENT_SCHEMA') FROM DUAL";
+		return "select sys_context('USERENV', 'CURRENT_SCHEMA') from dual";
 	}
 
 	@Override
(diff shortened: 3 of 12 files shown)


Sanne:
I've merged this in master but not backporting it yet.

Sanne:
- https://hibernate.atlassian.net/browse/HHH-14240

katzyn:
@gavinking
Hibernate really should use upper-case names of all tables and columns in the `INFORMATION_SCHEMA` for H2. `select * from INFORMATION_SCHEMA.sequences` doesn't work in some versions of H2 (in 1.4.200, for example) when `DATABASE_TO_UPPER=FALSE` setting is in use.

Sanne:
Many thanks @katzyn , I'll revert this.  We can then think about options without any rush.

Sanne:
Reverted, sorry Gavin! This will need a bit more consideration.

I'm quite fond of using H2 with `DATABASE_TO_UPPER=FALSE`, I remember needing to use it in specific benchmark tests, so I believe we should generate SQL that (by default) matches its internal expectations.

gavinking:
@Sanne 


> I'm quite fond of using H2 with `DATABASE_TO_UPPER=FALSE`, I remember needing to use it in specific benchmark tests, so I believe we should generate SQL that (by default) matches its internal expectations.

Well I think I fixed that, since there's a test for it (that still passes after this change).

It was only the schema name that needed to be uppercase, so I left that uppercase.

gavinking:
(See `H2DialectDataBaseToUpperTest`.)

gavinking:
> `select * from INFORMATION_SCHEMA.sequences` doesn't work in some versions of H2 (in 1.4.200, for example)

OK, fine, so apparently there was a change in 1.4.200, which I didn't catch because we test on H2 1.4.197.

I've now made the tiny change so that the patch also works on H2 1.4.200 with `DATABASE_TO_UPPER=FALSE`.

Would you please now re-apply this.

(By the way, SQL is defined to be case insensitive for unquoted identifiers, and so I would call this setting a misfeature of H2, but whatever.)

gavinking:
Oh, and for the record, your preferred syntax syntax `select * from INFORMATION_SCHEMA.SEQUENCES` is broken if you set `DATABASE_TO_UPPER=TRUE` in H2.

So I don't know why we're even trying to accommodate these silly, completely non-standard and non-default behaviors of H2!

katzyn:
> Oh, and for the record, your preferred syntax syntax `select * from INFORMATION_SCHEMA.SEQUENCES` is broken if you set `DATABASE_TO_LOWER=TRUE` in H2.

No, it isn't broken. Unquoted upper-case names of information schema tables and columns work in all versions of H2 with all combinations of settings. Unquoted lower-case and mixed-case names normally work, but in H2 1.4.198, 1.4.199, and 1.4.200 they don't work if and only if `DATABASE_TO_UPPER=FALSE` is specified **and** neither `DATABASE_TO_LOWER=TRUE` nor `CASE_INSENSITIVE_IDENTIFIERS=TRUE` are specified.

> completely contrary-to-standard

In the SQL Standard everything is upper-case and lower-case and mixed-case identifiers are normalized to upper case. But some DBMS, such as PostgreSQL and MySQL, convert everything to lower case, and there are many other deviations across different DBMS. Many people use H2 as replacement of other DBMS in tests and for similar purposes, that's why all these weird settings exist in H2. I don't like them too, but they are used (and sometimes misused) very often.

Sanne:
This change has likely more impact than what we can think of during a simple PR review, so I want to be cautious.

As mentioned I have grounds to suspect an impact on performance and other subtle compatibility issues with various tools, neither of which are covered by SQL standards. But also I won't have time to spend a week to perform an in depth analysis of the implications, so either someone else volunteers or I will keep having my doubts about this.

Happy to change this going forward in Hibernate ORM 6 as it's a new major release, but I see no good reason to risk such a change in 5.

gavinking:
> Unquoted upper-case names of information schema tables and columns work in all versions of H2 with all combinations of settings. 

OK, I thought I had tested that and found that it *didn'*t work on 1.4.198, but perhaps I'm mistaken.

> In the SQL Standard everything is upper-case and lower-case and mixed-case identifiers are normalized to upper case. But some DBMS, such as PostgreSQL and MySQL, convert unquoted identifiers to lower case, and there are many other deviations across different DBMS.

And yet the net effect is that every SQL database acts as a case-insensitive language from the point of view of the user, and offers a well-defined way (quoted identifiers) to achieve case-sensitivity if necessary.

And *that* is the behavior that Hibernate does and should support.

Hibernate doesn't need to support weird and pointless toggles that make SQL act in a non-standard way, especially toggles that exhibit inconsistent and differently-broken behavior across different versions of the database.

(If you have something that behaves differently between version 1.4.197 and 1.4.198 then we already know it is bad and wrong and shouldn't be used.)

Now, *for Hibernate 5*, where we do have a very strong requirement for backward compatibility, then sure, I agree that it's a good idea to maintain the current behavior, which is why I quickly fixed this as soon as it was pointed out, and why the current patch rolls back that tiny change.

But going forward (H6) I would say that this is the sort of thing we don't need to support, and it's *MUCH, MUCH* more important for Hibernate to not generate ugly unprofessional-looking SQL with a mix of lowercase and uppercase identifiers.

gavinking:
> This change has likely more impact than what we can think of during a simple PR review, so I want to be cautious.

I strongly disagree. These are changes I already made in H6 after very extensive testing.

(Indeed, I'm pretty sure that I'm *the only* person who has invested so much time testing across so many different SQL dialects in recent times, as evidenced by the sheer number of wrong things I picked up on and fixed in the process.)

And OK, fine, so I missed something that only occurs:

- on certain versions of H2,
- when you run H2 with a very specific setting.

So my bad _I guess_ but I really don't see what the problem is. If we break something on some weird version of some database then I'm sure users can let us know about that and we can fix it.

I think it's of *far more serious concern* that Hibernate 5 currently generates a random and entirely unprofessional-looking mix of uppercase and lowercase keywords and identifiers on some platforms, than that this patch carries a slight risk of breaking some completely obscure corner cases.

Indeed, if I were a user of one of these platforms, and saw my ORM generating SQL with lowercase `select` and uppercase `SKIP`, I think I would assume that I was working with a rubbish product and should look for a different solution.

gavinking:
> These are changes I already made in H6 after very extensive testing.

In fairness, I wasn't able to test on Informix, nor on old versions of SQL server, but I would like to say I'm 99.999% certain that neither of these databases ever exhibited case-sensitive *keywords*.

gavinking:
@Sannw would you please reapply the change.

Sanne:
Merged Gavin's reviewed version as https://hibernate.atlassian.net/browse/HHH-14240

Will be included in 5.5.0.Final

gavinking:
Thanks @Sanne

PR #3558, opened 2020-09-18
Add option to always include collection fields in the default fetch group

gavinking:
There is no good reason to lazily-instantiate a collection wrapper, since that operation never requires access to the database.

See discussion here:

   https://github.com/hibernate/hibernate-reactive/issues/374

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
@@ -15,6 +15,7 @@ import org.hibernate.bytecode.BytecodeLogger;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.internal.SessionFactoryRegistry;
+import org.hibernate.mapping.Collection;
 import org.hibernate.mapping.OneToOne;
 import org.hibernate.mapping.Property;
 import org.hibernate.mapping.ToOne;
@@ -30,7 +31,8 @@ public class EnhancementHelper {
 	public static boolean includeInBaseFetchGroup(
 			Property bootMapping,
 			boolean isEnhanced,
-			boolean allowEnhancementAsProxy) {
+			boolean allowEnhancementAsProxy,
+			boolean collectionsInDefaultFetchGroupEnabled) {
 		final Value value = bootMapping.getValue();
 
 		if ( ! isEnhanced ) {
@@ -63,7 +65,8 @@ public class EnhancementHelper {
 			return true;
 		}
 
-		return ! bootMapping.isLazy();
+		return collectionsInDefaultFetchGroupEnabled && ( value instanceof Collection )
+				|| ! bootMapping.isLazy();
 	}
 
 	public static <T> T performWork(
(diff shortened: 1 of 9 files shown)


gavinking:
So I note that there are *actual tests* for this feature, but I still think it's wrong and *extremely* dangerous.

You see I had code that was working, and I turned on the bytecode enhancer, and it still *seemed* to be working, and I didn't notice for literally *months* that where I had written `author.books` I wasn't getting the persistent collection of books, but instead just the empty `ArrayList` I instantiated in the constructor.

Because, you see, this turning on the enhancer changed the semantics of code that I had not marked `@Basic(fetch = LAZY)`, where I would be expecting to see some field-level lazy fetching.

Now, OK, sure, I would have been fine if my code had been using getters (`getBooks()`), but as it happens I wasn't.

gavinking:
So look, since this is apparently intended behavior, what I've done for now is to add an option so that it can at least be disabled in Hibernate Reactive, where it's a real problem (since lazy fetching is never transparent in HR).

But I personally would turn that option on by default except in Quarkus.

Thoughts, @Sanne @sebersole?

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
```diff
@@ -63,7 +65,8 @@ public static boolean includeInBaseFetchGroup(
 			return true;
 		}
 
-		return ! bootMapping.isLazy();
+		return collectionsInDefaultFetchGroupEnabled && ( value instanceof Collection )
+				|| ! bootMapping.isLazy();
```
NathanQingyangXu, line 69: This is minor and feel free to ignore this comment. IMO we could enhance code readability by swapping the operand order as following:
```
return ! bootMapping.isLazy || ( value instanceof Collection ) && collectionsInDefaultFetchGroupEnabled;
```
because short-circuiting avoids unnecessary checking and implies the following preconditions:
* the new logic only applies when `bootMapping.isLazy()` returns true;
* the `collectionsInDefaultFetchGroupEnabled` checking only applies when value is of `Collection` type;

There is no performance implication, however, for `instanceof` has been super fast in modern JVM. E.g., https://stackoverflow.com/questions/103564/the-performance-impact-of-using-instanceof-in-java

Sanne:
ok, I can test and merge this later today.

I need to understand you better on this:

> But I personally would turn that option on by default except in Quarkus.

You mean for Hibernate Reactive only?

We can control configurations which are specifically needed for Reactive (and for Quarkus/ORM blocking) but it's generally useful to reduce the configuration flags that people can set, so to reduce the testing permutations we need to look at. Also because we can't test all permutations, which translates into a quality and maintenance risk.

I'm of course fine to flip a default if you think it's _the right thing_: we don't have to be uber-conservative with these branches.

gavinking:
> You mean for Hibernate Reactive only?

Well, no, I mean for plain Hibernate ORM as well ... I don't see why collections need to be twice-lazy.

Now, in Quarkus, where we can enhance the client code, I guess I don't really care, and I suppose it's a very minor performance gain to not have to instantiate a `PersistentCollection` when you load an entity.

But if we're not enhancing client code, it seems to me that the current default is a bad default.

Sanne:
Merged as https://hibernate.atlassian.net/browse/HHH-14238

Now that I see it better I agree that it's a bit puzzling. I'll go with care so not switching the default just yet but I agree we should do it soon.

gavinking:
Thanks @Sanne.

PR #3560, opened 2020-09-21
HHH-14222 id with space padding column definition in fails to load fully 

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14222

The issue seems pretty tricky for it involves id inconsistency (id we assigned in code and its db representation). Usually it is not a good idea to make it happen but for completeness we still need to fix it.

Ongoing. Currently only broken testing case is included.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/id/IdColumnWithSpacePaddingDefinitionTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/id/IdColumnWithSpacePaddingDefinitionTest.java
@@ -1,0 +1,94 @@
+package org.hibernate.test.id;
+
+import java.util.Collections;
+import java.util.List;
+import javax.persistence.CascadeType;
+import javax.persistence.Column;
+import javax.persistence.Entity;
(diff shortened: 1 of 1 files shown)


mefernandez:
Hi,

I currently work around this problem by preemptively right-padding ids with spaces up to the length of the column.

```java
String id = StringUtils.rightPad("a",10," ");
A a = entityManager.find( A.class, id );
```

NathanQingyangXu:
I think that is a good workaround. Fixing it in Hibernate is pretty tricky
(and even risky) and seems very few people are experiencing this issue.

mefernandez:
It's fine. 
I acknowledge not many people face a Sybase database with fixed-width primary/foreign keys.

In all, taking a deep dive to properly submit this issue and its test case helped me find this workaround.

Thanks for taking the time to look into it Nathan.

NathanQingyangXu:
As bug reporter has found some workaround and I have no free cycles to spare from my own job, I am gonna close this PR. Anyone who has interest to solve it can still make use of the testing case in this PR.

PR #3561, opened 2020-09-23
wip/6.0 Fix the typo of `extracter` in codebase

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14601

A simple refactoring to correct the typo of `extracter` (should be `extractor`) which I found during working on my previous PR.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/exception/internal/SQLStateConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/exception/internal/SQLStateConverter.java
@@ -18,12 +18,12 @@ import org.hibernate.exception.spi.ViolatedConstraintNameExtractor;
  */
 @Deprecated
 public class SQLStateConverter extends StandardSQLExceptionConverter implements SQLExceptionConverter {
-	public SQLStateConverter(final ViolatedConstraintNameExtractor extracter) {
+	public SQLStateConverter(final ViolatedConstraintNameExtractor extractor) {
 		super();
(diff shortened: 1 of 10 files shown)


sebersole:
This is not specifically about you or this PR.  I've seen a few now
changing `er` to `or` or `or` to `er`.  Both are valid in English.

gavinking:
Steve I don't think that's quite correct. It might be the case for some words (perhaps due to differences between American and British English) but I don't believe it's generally true, and according to Google, it's not true for "extractor".

gavinking:
Then indispensable Language Log: https://languagelog.ldc.upenn.edu/nll/?p=22378

sebersole:
http://www.differencebetween.net/language/difference-between-er-and-or/#:~:text=2.,end%20with%20%E2%80%9C%2Date.%E2%80%9D

sebersole:
And before you go there... yes this one is ok.

I'm just speaking about all of them in general here in one place.  I forget
the others I have seen.  Some of those were for sure irrelevant changes
meaning either version worked.  I didn't say anything before because well
it does not really matter.  If someone wants to spend time doing this,
great.  Just pointing out that some I have seen are correct either way and
going through the effort of automatically picking one or the other might be
a waste of people's time

NathanQingyangXu:
Yeah, I have googled before creating this PR. I also found either 'updateable' and 'updatable' is correct and I think I've correct the former to latter in my code comment typo fixing PR.

sebersole:
Any chance you can fix these compile errors?

PR #3562, opened 2020-09-25
HHH-14230 Fix generics to avoid ClassCastException

quaff:
https://hibernate.atlassian.net/browse/HHH-14230

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/mapping/Table.java
+++ b/hibernate-core/src/main/java/org/hibernate/mapping/Table.java
@@ -39,7 +39,7 @@ import org.jboss.logging.Logger;
  *
  * @author Gavin King
  */
-@SuppressWarnings("unchecked")
+@SuppressWarnings("deprecation")
 public class Table implements RelationalModel, Serializable, Exportable {
 	private static final Logger log = Logger.getLogger( Table.class );
 
@@ -50,14 +50,14 @@ public class Table implements RelationalModel, Serializable, Exportable {
 	/**
 	 * contains all columns, including the primary key
 	 */
-	private Map columns = new LinkedHashMap();
+	private Map<String, Column> columns = new LinkedHashMap<>();
 	private KeyValue idValue;
 	private PrimaryKey primaryKey;
-	private Map<ForeignKeyKey, ForeignKey> foreignKeys = new LinkedHashMap<ForeignKeyKey, ForeignKey>();
-	private Map<String, Index> indexes = new LinkedHashMap<String, Index>();
-	private Map<String,UniqueKey> uniqueKeys = new LinkedHashMap<String,UniqueKey>();
+	private Map<ForeignKeyKey, ForeignKey> foreignKeys = new LinkedHashMap<>();
+	private Map<String, Index> indexes = new LinkedHashMap<>();
+	private Map<String,UniqueKey> uniqueKeys = new LinkedHashMap<>();
 	private int uniqueInteger;
-	private List<String> checkConstraints = new ArrayList<String>();
+	private List<String> checkConstraints = new ArrayList<>();
 	private String rowId;
 	private String subselect;
 	private boolean isAbstract;
@@ -246,7 +246,7 @@ public class Table implements RelationalModel, Serializable, Exportable {
 	}
 
 	public Column getColumn(int n) {
-		Iterator iter = columns.values().iterator();
+		Iterator<Column> iter = columns.values().iterator();
 		for ( int i = 0; i < n - 1; i++ ) {
 			iter.next();
 		}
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/mapping/Table.java
```diff
@@ -50,7 +50,7 @@
 	/**
 	 * contains all columns, including the primary key
 	 */
-	private Map columns = new LinkedHashMap();
+	private Map<String, Column> columns = new LinkedHashMap<String, Column>();
```
NathanQingyangXu, line 53: We don't need to duplicate generics spec on the right due to diamond operator. Other than that, this PR seems good. A good example we need to pay more attention to generics usage, partly because of legacy reason, partly because of coding sloppiness.

NathanQingyangXu (changes requested):
I will approve the PR if you delete unnecessary generics spec on right.

hibernate-core/src/main/java/org/hibernate/mapping/Table.java
quaff: I'd prefer diamond operator, but it breaks consistency with legacy code, It’s necessary to cleanup entire codebase in another commit.

dreab8:
Hi @quaff thanks for the PR, can you please also provide a test for the issue?

quaff:
> Hi @quaff thanks for the PR, can you please also provide a test for the issue?

Coming soon.

hibernate-core/src/main/java/org/hibernate/mapping/Table.java
```diff
@@ -420,10 +420,9 @@ public void validateColumns(Dialect dialect, Mapping mapping, TableMetadata tabl
 				throw new HibernateException( "Missing column: " + col.getName() + " in " + Table.qualify( tableInfo.getCatalog(), tableInfo.getSchema(), tableInfo.getName()));
 			}
 			else {
-				final boolean typesMatch =
-						dialect.equivalentTypes( columnInfo.getTypeCode(), col.getSqlTypeCode( mapping ) )
-						|| col.getSqlType( dialect, mapping ).toLowerCase(Locale.ROOT)
-								.startsWith( columnInfo.getTypeName().toLowerCase(Locale.ROOT) );
+				final boolean typesMatch = col.getSqlType( dialect, mapping ).toLowerCase(Locale.ROOT)
+						.startsWith( columnInfo.getTypeName().toLowerCase(Locale.ROOT) )
+						|| columnInfo.getTypeCode() == col.getSqlTypeCode( mapping );
```
dreab8, line 425: Why this change?

hibernate-core/src/main/java/org/hibernate/mapping/Table.java
dreab8: I agree with @NathanQingyangXu , better to use the diamond operator.

dreab8:
> Coming soon.

Thanks

hibernate-core/src/main/java/org/hibernate/mapping/Table.java
quaff: done.

hibernate-core/src/main/java/org/hibernate/mapping/Table.java
quaff: Fixed changes by accidental formatting.

NathanQingyangXu (approved):

NathanQingyangXu:
I've approved this PR. I just have one open question here: is it preferable to name testing class by ticket number or descriptive  natural language? I saw coexistence of both in our codebase. Personally I always opt for the latter, though sometimes I found I am facing trouble to come up with terse name restrictive to the ticket's scope.

quaff:
> I've approved this PR. I just have one open question here: is it preferable to name testing class by ticket number or descriptive natural language? I saw coexistence of both in our codebase. Personally I always opt for the latter, though sometimes I found I am facing trouble to come up with terse name restrictive to the ticket's scope.

Naming is difficult, especially for those people don't speak english, you should understand.

NathanQingyangXu:
> > I've approved this PR. I just have one open question here: is it preferable to name testing class by ticket number or descriptive natural language? I saw coexistence of both in our codebase. Personally I always opt for the latter, though sometimes I found I am facing trouble to come up with terse name restrictive to the ticket's scope.
> 
> Naming is difficult, especially for those people don't speak english, you should understand.

One of the benefits of natural language test class name is for searching by keyword. It is invaluable if someone wanna search to see whether some similar bugs have been fixed.

Sanne:
Thanks! merged

NathanQingyangXu:
@quaff , I have started an ambitious PR to improve the generics usage in our codebase in v6: https://github.com/hibernate/hibernate-orm/pull/3587. It might take a couple of weeks or even months to have it done.

PR #3563, opened 2020-09-26
HHH-14231 Fix ClassCastException in ScrollableResultsImpl for primitive array value

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14231

straightforward implementation. Array of primitive cannot be cast to array of object.


Sanne:
Thanks! merged and backported to 5.4

PR #3564, opened 2020-09-26
HHH-14223 - nullpointer exception on @JoinFormula with implicit join

bukajsytlos:
Can't add label but JIRA issue Test Case ;)

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/persister/entity/JoinFormulaImplicitJoinTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/persister/entity/JoinFormulaImplicitJoinTest.java
@@ -1,0 +1,167 @@
+package org.hibernate.persister.entity;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
+
+import java.util.List;
+import java.util.Map;
+
+import javax.persistence.CascadeType;
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.OneToMany;
+
+import org.hibernate.annotations.JoinColumnOrFormula;
+import org.hibernate.annotations.JoinColumnsOrFormulas;
+import org.hibernate.annotations.JoinFormula;
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.jpa.test.BaseEntityManagerFunctionalTestCase;
+import org.hibernate.testing.TestForIssue;
+import org.junit.Before;
+import org.junit.Test;
+
+@TestForIssue(jiraKey = "HHH-14223")
+public class JoinFormulaImplicitJoinTest extends BaseEntityManagerFunctionalTestCase {
+    @Override
+    protected Class<?>[] getAnnotatedClasses() {
+        return new Class<?>[] {
+            Person.class, PersonVersion.class
+        };
+    }
+
+    @Override
+    protected void addConfigOptions(Map options) {
+        options.put(
+            AvailableSettings.GLOBALLY_QUOTED_IDENTIFIERS,
+            Boolean.TRUE
+        );
+    }
+
+    @Before
+    public void setUp() {
+        doInJPA( this::entityManagerFactory, entityManager -> {
+            final Person person = new Person();
+            entityManager.persist(person);
+
+            for (int i = 0; i < 3; i++) {
+                final PersonVersion personVersion = new PersonVersion();
+                personVersion.setName("Name" + i);
+                personVersion.setVersion(i);
+                personVersion.setPerson(person);
+                entityManager.persist(personVersion);
+            }
+        });
+    }
+
+    protected int entityCount() {
+        return 5;
+    }
+
+    @Test
+    public void testImplicitJoin() {
+        doInJPA( this::entityManagerFactory, entityManager -> {
+            entityManager.createQuery(
+                "SELECT person\n" +
+                    "FROM Person AS person\n" +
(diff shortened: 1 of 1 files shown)


NathanQingyangXu:
do you plan to fix it further in this PR, btw?

bukajsytlos:
propably not, I am kind of lost in hibernate abstractions

hibernate-core/src/test/java/org/hibernate/persister/entity/JoinFormulaImplicitJoinTest.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Test
+    public void testImplicitJoin() {
+        doInJPA( this::entityManagerFactory, entityManager -> {
+            entityManager.createQuery(
+                "SELECT person\n" +
+                    "FROM Person AS person\n" +
+                    "    LEFT JOIN FETCH person.latestPersonVersion\n" +
+                    "order by person.latestPersonVersion.id desc\n"
```
NathanQingyangXu, line 70: as a workaround, you might use the following HQL to bypass the bug:
```
SELECT person
FROM Person AS person 
     LEFT JOIN FETCH person.latestPersonVersion AS latestVersion 
ORDER BY latestVersion.id desc 
```

hibernate-core/src/test/java/org/hibernate/persister/entity/JoinFormulaImplicitJoinTest.java
```diff
@@ -0,0 +1,167 @@
...
+    public static class Person {
+
+        @Id
+        @GeneratedValue
+        private Long id;
+
+        @OneToMany(mappedBy = "person", cascade = CascadeType.ALL, orphanRemoval = true)
+        private List<PersonVersion> personVersions;
+
+        @ManyToOne
```
NathanQingyangXu, line 85: When I changed the above to `@OneToOne`, another NPE was triggered. Seems another bug we can file.

hibernate-core/src/test/java/org/hibernate/persister/entity/JoinFormulaImplicitJoinTest.java
beikov: Just by the way, the path expression `person.latestPersonVersion.id` will create another join, so I doubt you want to use such a query anyway. The query Nathan suggested is what you should normally write.

bukajsytlos: well this is the query framework we use will issue. should not hibernate be smart enough to not execute unnecessary join?

beikov: This is not about being smart, an implicit join is just something different than an explicit join. Also, would you want the SQL to contain the subquery in the order by clause or rather the id column of the person table? The query you wrote would lead to a subquery in the order by clause. The query Nathan proposed would lead to the person table column being used in the order by clause.

NathanQingyangXu:
Btw, I tested in Hibernate v5.3 and found it seems fine. So maybe a
regression issue?

hibernate-core/src/test/java/org/hibernate/persister/entity/JoinFormulaImplicitJoinTest.java
bukajsytlos: Ah sorry I understood/read it wrong. Btw the error is present even without join fetch part

NathanQingyangXu: Regardless of whether your `ORDER BY` is misused, definitively Hibernate should be smarter than throwing NPE. E.g., it should throw more meaningful exception when it found `ORDER BY` is misused, ideally during the HQL creation statement executing.

bukajsytlos: hm, why misused? it was working in older versions. I don't think throwing different error will solve this issue

bukajsytlos: this will fail too
```
SELECT person
FROM Person AS person 
ORDER BY person.latestPersonVersion.id desc
```

NathanQingyangXu: Still investigating why it used to be working in v5.3. From current implementation in v5.4, it seems `ORDER BY` cannot go deeper than first level so internally it treats `latestPersonVersion.id` as the property name and naturally cannot find it, ending up with NPE.

NathanQingyangXu:
Finally https://github.com/hibernate/hibernate-orm/pull/2394 is nailed down to be the culprit. Will investigate it further.

hibernate-core/src/test/java/org/hibernate/persister/entity/JoinFormulaImplicitJoinTest.java
jwgmeligmeyling: `person.latestPersonVersion.id` can actually be dereferenced - not from the primary table obviously, but the join table is included by default in the entity subclass table span. Not dereferencing it against actually leads to a third join to be rendered in this scenario.

The issue is caused because of improper handling of ID attributes in `AbstractPropertyMapping`. I'll provide a patch for that.

jwgmeligmeyling:
I've implemented a fix in #3570

NathanQingyangXu:
@bukajsytlos you can close this PR for the issue has been fixed in another PR which has included your testing case. Thanks for helping us to fix this regression issue.

bukajsytlos:
Thanks for your effort and fix

PR #3565, opened 2020-09-27
HHH-14228 Introduce session.amend()

quaff:
save an an unnecessary update if possible, like git commit --amend.


quaff:
Supersede by https://github.com/hibernate/hibernate-orm/pull/3566 and https://github.com/hibernate/hibernate-orm/pull/3567

PR #3567, opened 2020-09-27
HHH-14228 Reuse insert if entity is unflushed

quaff:
Compare to https://github.com/hibernate/hibernate-orm/pull/3566 ,  no need call `session.update()`

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
@@ -228,6 +228,10 @@ public class ActionQueue {
 		isTransactionCoordinatorShared = false;
 	}
 
+	public ExecutableList<AbstractEntityInsertAction> getInsertions() {
+		return insertions;
+	}
+
 	public void clear() {
 		EXECUTABLE_LISTS_MAP.forEach( (k,listProvider) -> {
 			ExecutableList<?> l = listProvider.get( this );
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -228,6 +228,10 @@ public ActionQueue(SessionImplementor session) {
 		isTransactionCoordinatorShared = false;
 	}
 
+	public ExecutableList<AbstractEntityInsertAction> getInsertions() {
+		return insertions;
+	}
+
```
NathanQingyangXu, line 234: Backward compatibility is a concern in Hibernate codebase evolution and we would be cautious to change its public API (like adding a new public method) for other Hibernate projects or jboss product might be broken. So approach without breaking backward compatibility has a pros accordingly.

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultFlushEntityEventListener.java
```diff
@@ -470,6 +473,21 @@ protected final boolean isUpdateNecessary(FlushEntityEvent event) throws Hiberna
...
+						if ( insertAction.getInstance() == event.getEntity() ) {
+							Object[] oldState = insertAction.getState();
+							Object[] newState = event.getPropertyValues();
+							for ( int i = 0; i < dirtyProperties.length; i++ ) {
+								oldState[dirtyProperties[i]] = newState[dirtyProperties[i]];
+							}
+							return false;
+						}
+					}
+				}
```
NathanQingyangXu, line 490: would be great if `final` is used as aggressively as in other locations in Hibernate codebase.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
quaff: Sure it's not a good idea, but I need it here, I think it's safe enough since adding new getter doesn't break backward compatibility normally.

quaff:
Supersede by #3566

PR #3568, opened 2020-09-27
HHH-14235 Polishing OneToOneUnidirectionalTest

quaff:
child should be persisted before parent, an update sql can be saved after this commit

insert into Phone (details_id, `number`, id) values (null, '123-456-7890', 1)
insert into PhoneDetails (provider, technology, id) values ('T-Mobile', 'GSM', '2')
update Phone set details_id=2, `number`='123-456-7890' where id=1
==>
insert into PhoneDetails (provider, technology, id) values ('T-Mobile', 'GSM', '1')
insert into Phone (details_id, `number`, id) values (1, '123-456-7890', 2)

### DIFF
--- a/documentation/src/test/java/org/hibernate/userguide/associations/OneToOneUnidirectionalTest.java
+++ b/documentation/src/test/java/org/hibernate/userguide/associations/OneToOneUnidirectionalTest.java
@@ -39,8 +39,8 @@ public class OneToOneUnidirectionalTest extends BaseEntityManagerFunctionalTestC
 			PhoneDetails details = new PhoneDetails( "T-Mobile", "GSM" );
 
 			phone.setDetails( details );
-			entityManager.persist( phone );
 			entityManager.persist( details );
+			entityManager.persist( phone );
 		} );
 	}
 


quaff:
Such usage is legal but not recommended, I found this because it blocks https://github.com/hibernate/hibernate-orm/pull/3566 and https://github.com/hibernate/hibernate-orm/pull/3567, caused by foreign key constraint violation.

@vladmihalcea Would you have a look at this?

vladmihalcea:
@quaff I no longer work for Red Hat, so you are better off asking the Hibernate devs to check it out. Cheers.

NathanQingyangXu:
Usually we require associated JIRA ticket to be created (and include its `HHH-xxxx` as git commit prefix) for such PR targeting Hibernate v5, unless your PR is targeting v6 (`wip/6.0`).

NathanQingyangXu (approved):

NathanQingyangXu:
Please include ticket number as the git commit message prefix. That way we can correlate git commit and its associated ticket more easily.

beikov:
@quaff you mentioned that your changes in #3566 and #3567 break this test(valid) and now you want to change it so that it doesn't fail anymore? I didn't look into the test yet, but this should not be necessary. Maybe your fixes aren't that perfect yet?

beikov:
Or is the foreign key created due to #3569 not being fixed?

quaff:
> @quaff you mentioned that your changes in #3566 and #3567 break this test(valid) and now you want to change it so that it doesn't fail anymore? I didn't look into the test yet, but this should not be necessary. Maybe your fixes aren't that perfect yet?
My point is fix the misuse in this test.
Indeed #3566 or #3567 will breaks such usage, maybe I need introduce a new flag and disabled by default, It's still work in progress. There is one important question, which one is better? I would hear advice from hibernate team.

quaff:
> Or is the foreign key created due to #3569 not being fixed?

#3569 is a perfect fix from my perspective, It's not related to this.

quaff:
Keep this anti pattern, prevent breaking compatibility, misuse better than broken.

quaff:
@vladmihalcea I suggest you detect such usage in Hypersistence Optimizer.

vladmihalcea:
@quaff I don't understand thr issue. What do you mean by "child should be persisted before parent"? A parent entity should be persisted first.

quaff:
> @quaff I don't understand thr issue. What do you mean by "child should be persisted before parent"? A parent entity should be persisted first.

For example entity `Foo` refer to entity `Bar`, then `Bar` should be persisted before `Foo`, otherwise `foo.bar` will be null while inserting then updated with `bar` later.

vladmihalcea:
@quaff Sorry, but I've never understood what those `Foo` or `Bar` terminologies stand for since they don't have any resemblance to the real world.

For instance, in all the examples I have on my blog, I use `Post` for the parent entity and `PostComment` for the one-to-many child. This is because it's easier for people to reason about the domain model.

In this case, because we are talking about a one-to-one association, the parent `Post` entity has a `PostDetails` child entity, which, again, makes sense since anyone will be able to visualize the relationship.

So, considering this model, it makes sense to persist the `Post` first and then the `PostDetails`. Those can happen in different transactions too. 

Now, is this issue related to the `cascade` attribute of the `@OneToOne` annotation?

quaff:
`Post` should be persisted before `PostDetails` to save one unnecessary update statement, It's not related to `cascade`.

vladmihalcea:
And isn't the `Post` persisted by Hibernate prior to the `PostDetails`? This was working fine in the past. If it doesn't work like that, then it's a bug that will have to be fixed. 

Hibernate should insert the parent first and the child afterward, no matter the order the user has called persist on the entities.

quaff:
> And isn't the `Post` persisted by Hibernate prior to the `PostDetails`? This was working fine in the past. If it doesn't work like that, then it's a bug that will have to be fixed.
> 
> Hibernate should insert the parent first and the child afterward, no matter the order the user has called persist on the entities.

I change the order of persist call in this PR, It indeed eliminate one update statement. I'm not sure it violate JPA specification.

vladmihalcea:
I don't think the JPA specification imposes any restriction related to how the SQL statements should be executed. This is an implementation detail that's left for the JPA provider to decide.

quaff:
> I don't think the JPA specification imposes any restriction related to how the SQL statements should be executed. This is an implementation detail that's left for the JPA provider to decide.

Then it's an improvement not a bug since it works functionally.

quaff:
> Hibernate should insert the parent first and the child afterward, no matter the order the user has called persist on the entities.

Do this rely on `hibernate.order_inserts` ? @vladmihalcea

vladmihalcea:
The `hibernate.order_inserts` is a different mechanism. The outcome should be the same. The parent should be inserted first, then the child and no UPDATE should be generated for the one-to-one association.

The problem with that initial `Phone` and `PhoneDetails` mapping is that it's not even a real one-to-one table relationship, but a one-to-many database table relationship with a UNIQUE constraint on the FK. The true one-to-one relationship uses a single column that's both the PK and the PK. So, only `@MapsId` can generate the property one-to-one relationship. The default one is both less efficient and not even compliant to the DB schema design considerations.

quaff:
Superseded by https://hibernate.atlassian.net/browse/HHH-14254

PR #3571, opened 2020-09-28
HHH-14220 Skip test when bytecode provider is Javassist

dreab8:
This PR skip the execution of the test when the bytecode provides is Javassist, the fix for https://hibernate.atlassian.net/browse/HHH-14220 was implemented only for ByteBuddy.

Not sure this is the best way to skip the test, any suggestion is more than welcome.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/inlinedirtychecking/DirtyCheckPrivateUnMappedCollectionTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/inlinedirtychecking/DirtyCheckPrivateUnMappedCollectionTest.java
@@ -21,6 +21,7 @@ import javax.persistence.MappedSuperclass;
 import org.hibernate.boot.MetadataSources;
 import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
 import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.cfg.Environment;
 
 import org.hibernate.testing.bytecode.enhancement.BytecodeEnhancerRunner;
 import org.hibernate.testing.bytecode.enhancement.CustomEnhancementContext;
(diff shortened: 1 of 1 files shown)


NathanQingyangXu:
> This PR skip the execution of the test when the bytecode provides is Javassist, the fix for https://hibernate.atlassian.net/browse/HHH-14220 was implemented only for ByteBuddy.
> 
> Not sure this is the best way to skip the test, any suggestion is more than welcome.

IHMO I think the best solution is to come up with some thing similar to `@SkipForDialect`. The current approach is not authentic skipping. Off course, it all depends whether creating such abstract is worthwhile or not.

hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/inlinedirtychecking/DirtyCheckPrivateUnMappedCollectionTest.java
```diff
@@ -45,11 +48,21 @@ protected void configureStandardServiceRegistryBuilder(StandardServiceRegistryBu
 
 	@Override
 	protected void applyMetadataSources(MetadataSources sources) {
-		sources.addAnnotatedClass( Measurement.class );
+		String byteCodeProvider = Environment.getProperties().getProperty( AvailableSettings.BYTECODE_PROVIDER );
+		if ( byteCodeProvider != null && !Environment.BYTECODE_PROVIDER_NAME_BYTEBUDDY.equals( byteCodeProvider ) ) {
```
NathanQingyangXu, line 52: testing against `javaassist` might be more accurate and maintainable. The above logic is based on the precondition that the only non-bytebuddy provider is `javaassist`, which might not be correct all the time.

Sanne:
Hi @NathanQingyangXu , those are good suggestions and I would agree with them if it was about maintaining a feature for the long term, but consider that we don't plan to maintain the javassist enhancer for longer - and don't allow to plug a custom bytecode provider either, as it's too advanced and nobody has ever asked to try :)

I'll take @dreab8 's quick fix because I trust it's short lived.

PR #3572, opened 2020-09-28
HHH-14233 fix issue @ManyToOne(fetch = FetchType.LAZY) + @JoinColumn end up with eager loading

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14233

Assume the following entity:

```
	@Entity(name = "Car")
	static class Car implements Serializable {
		@Id @GeneratedValue
		Long id;

		@JoinColumn(name = "siren", referencedColumnName = "siren")
		Company company;
	}
```

The `company` will be eager loaded, though `FetchType.LAZY` has been specified. However, if `@JoinColumn` is removed, lazy loading would take effect.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/lazyload/JoinColumnLazyLoadingTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/lazyload/JoinColumnLazyLoadingTest.java
@@ -1,0 +1,92 @@
+package org.hibernate.test.lazyload;
+
+import java.io.Serializable;
+import java.util.HashSet;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


NathanQingyangXu:
After some primitive investigating, it seems the root cause is obvious. Currently Hibernate's proxy is based on `id` or `primary key` and it won't work for `unique key` as `@JoinColumn` requires.
From what I know, even v6 has no plan to implement `unique key` for proxy, so I guess this issue cannot be easily fixed.
I closed this PR for the time being and left a comment in the ticket. Anyone who is willing to solve it can still work on top of the testing case in this PR.

PR #3574, opened 2020-09-29
HHH-14234 Fix issue denormalized table inherits unique constraints from parent table

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14234

In denormalized table (child table with `TABLE_PER_CLASS` inheritance strategy is the only case in Hibernate), it shares many columns with its parent table, but it should stop duplicating any further. Among others, no need to duplicate unique constraint from parent table to child table or denormalized table.


sebersole (approved):
Looks good

PR #3577, opened 2020-09-30
HHH-14227 ActionQueue's insert ordering ultimate fix

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14227

Aimed to solve ActionQueue's insert order issue once for all via Topological Sorting. The current impl has the following serious drawbacks:
* data structure representation error: it is a dependency graph (one node could have multiple dependencies or multiple parent nodes) not a dependency tree; no wonder sometimes we fixed a bug ending up with creating a new one;
* sorting algorithm is based on terrible heuristics: it even uses time complexity literally as iteration halting condition;
* difficult to maintain simply because it is terribly difficult to understand; seems the current impl is purely based on ad-hoc testing cases;
* no way to locate circular dependency so cannot report meaningful debug info on such circular dependency;
* unnecessarily inefficient;

Another big problem is even though we have many testing cases, but surprisingly a good portion of existing testing cases lack SQL order verification. Needless to say, no 'foreign key integrity violation' exception is thrown doesn't necessarily mean the SQL ordering is correct, especially in face of code refactoring.

- [x] import the testing case form bug reporter
- [x] add SQL ordering verification logic to existing testing cases
- [ ] rewrite the insert ordering algorithm

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/insertordering/BaseInsertOrderingTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/insertordering/BaseInsertOrderingTest.java
@@ -1,0 +1,72 @@
+package org.hibernate.test.insertordering;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.util.Map;
+
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.cfg.Environment;
+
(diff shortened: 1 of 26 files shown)


NathanQingyangXu:
closed for it seems more complicated than I orginally expected.

PR #3578, opened 2020-10-01
HHH-14244 Force dirty mark on write to uninitialized lazy property

pludov:
The generated field writer method is modified. 

I added a check to detect an uninitialized lazy field and force marking it dirty.
But that check is only usefull when the new value is the default for the type (null/0/false...). For other values, the existing comparison is enough to mark the dirty. So, the new code is only triggered when received the default value, to avoid doing the check  too frequently

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/InlineDirtyCheckingHandler.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/InlineDirtyCheckingHandler.java
@@ -15,6 +15,8 @@ import javax.persistence.Id;
 
 import org.hibernate.bytecode.enhance.internal.bytebuddy.EnhancerImpl.AnnotatedFieldDescription;
 import org.hibernate.bytecode.enhance.spi.EnhancerConstants;
+import org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor;
+import org.hibernate.engine.spi.PersistentAttributeInterceptor;
 
 import net.bytebuddy.ClassFileVersion;
 import net.bytebuddy.asm.Advice;
@@ -37,10 +39,13 @@ final class InlineDirtyCheckingHandler implements Implementation, ByteCodeAppend
 
 	private final FieldDescription.InDefinedShape persistentField;
 
-	private InlineDirtyCheckingHandler(Implementation delegate, TypeDescription managedCtClass, FieldDescription.InDefinedShape persistentField) {
+	private final boolean checkLazyUninitialized;
+
+	private InlineDirtyCheckingHandler(Implementation delegate, TypeDescription managedCtClass, FieldDescription.InDefinedShape persistentField, boolean checkLazyUninitialized) {
 		this.delegate = delegate;
 		this.managedCtClass = managedCtClass;
 		this.persistentField = persistentField;
+		this.checkLazyUninitialized = checkLazyUninitialized;
 	}
 
 	static Implementation wrap(
@@ -58,7 +63,7 @@ final class InlineDirtyCheckingHandler implements Implementation, ByteCodeAppend
 					&& !( persistentField.getType().asErasure().isAssignableTo( Collection.class )
 					&& enhancementContext.isMappedCollection( persistentField ) ) ) {
 				implementation = new InlineDirtyCheckingHandler( implementation, managedCtClass,
-						persistentField.asDefined() );
+						persistentField.asDefined(), enhancementContext.isLazyLoadable(persistentField) );
 			}
 
 			if ( enhancementContext.isCompositeClass( persistentField.getType().asErasure() )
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/InlineDirtyCheckingHandler.java
```diff
@@ -58,7 +63,7 @@ else if ( !persistentField.hasAnnotation( Id.class )
 					&& !( persistentField.getType().asErasure().isAssignableTo( Collection.class )
 					&& enhancementContext.isMappedCollection( persistentField ) ) ) {
 				implementation = new InlineDirtyCheckingHandler( implementation, managedCtClass,
-						persistentField.asDefined() );
+						persistentField.asDefined(), enhancementContext.isLazyLoadable(persistentField) );
```
NathanQingyangXu, line 66: It would be great to comply to Hibernate coding format aggressively in non-testing code, including:
* use `final` keyword for temporary variable within method as much as possible;
* use space delimiter after `(` and before `)'.

hibernate-core/src/test/java/org/hibernate/jpa/test/enhancement/TestDirtyCheckOnLazy.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.Arrays;
+import java.util.Map;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@TestForIssue( jiraKey = "HHH-14244" )
+@RunWith( BytecodeEnhancerRunner.class )
+public class TestDirtyCheckOnLazy extends BaseEntityManagerFunctionalTestCase {
```
NathanQingyangXu, line 36: existing testing class naming convention is `xxxxTest`. Would be great to maintain consistent naming convention for maintenance purpose.

NathanQingyangXu:
Usually PR is targeting `master` branch rather than some specific version branch, so when it is merged to master, core  team member can determine whether to back port to 5.4 or not, in addition to merging to current 5.5 branch.

hibernate-core/src/test/java/org/hibernate/jpa/test/enhancement/TestDirtyCheckOnLazy.java
```diff
@@ -0,0 +1,174 @@
...
+            }
+
+            @Override
+            public void afterTransactionCompletion() {
+                assertTrue( Hibernate.isPropertyInitialized( entity, "lazyData" ) );
+            }
+        } );
+
+        checkLazyField( entity, testArray );
+    }
```
NathanQingyangXu, line 112: seems the above two methods were copied from `TestLazyPropertyOnPreUpdate`. Are they relevant to the nullness issue? We might delete them if not.

hibernate-core/src/test/java/org/hibernate/jpa/test/enhancement/TestDirtyCheckOnLazy.java
NathanQingyangXu: nvm, saw there is another testing class with similar naming pattern in the same package. Ignore my comment above.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/InlineDirtyCheckingHandler.java
```diff
@@ -96,6 +101,93 @@ public Size apply(
 			MethodVisitor methodVisitor,
 			Context implementationContext,
 			MethodDescription instrumentedMethod) {
+
+		// Skip comparaison when lazy not loaded
```
NathanQingyangXu, line 105: would be great to correct the above obvious typo: comparaison -> comparison

NathanQingyangXu:
@Sanne @dreab8 It seems there is some good stuff here. Could you take a look when you have time?

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/InlineDirtyCheckingHandler.java
```diff
@@ -96,6 +101,93 @@ public Size apply(
...
+		Label uninitializedLazy = new Label();
+
+		if (checkLazyUninitialized) {
+			// No dirty check is required for non initialized lazy fields.
+			Label skipUnitializedCheckWithPersistentAttributeInterceptor = new Label();
+			Label skipUnitializedCheck = new Label();
```
NathanQingyangXu, line 111: typo? `skipUnitializedXXX` => `skipUninitializedXXX`?

pludov:
> Usually PR is targeting `master` branch rather than some specific version branch, so when it is merged to master, core team member can determine whether to back port to 5.4 or not, in addition to merging to current 5.5 branch.

I was primarly interested on 5.4, so... :-) 
However, the branch applies verbatim on master and gradle build is ok as well.

NathanQingyangXu:
> > Usually PR is targeting `master` branch rather than some specific version branch, so when it is merged to master, core team member can determine whether to back port to 5.4 or not, in addition to merging to current 5.5 branch.
> 
> I was primarly interested on 5.4, so... :-)
> However, the branch applies verbatim on master and gradle build is ok as well.

well, targeting `master` branch is more future-proof, :). Also, it would motive code reviewers much more as a generic patch, as opposed to a branch patch only.

Sanne:
Many thanks @pludov , this seems a rather important fix, but I'm rusty on ASM. I need to set aside some proper review time next week.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/InlineDirtyCheckingHandler.java
```diff
@@ -96,6 +101,93 @@ public Size apply(
...
+
+			// When field is not initialized, comparison is useless
+			// if ( interceptor.isAttributeLoaded( fieldName ) ) goto skipUninitializedCheck
+			methodVisitor.visitJumpInsn( Opcodes.IFNE, skipUninitializedCheck );
+
+			// Sure that the lazy is not initialized. No need for comparison, directly set dirty
+			methodVisitor.visitJumpInsn( Opcodes.GOTO, uninitializedLazy );
+
+			methodVisitor.visitLabel( skipUninitializedCheckWithPersistentAttributeInterceptor );
+			if ( implementationContext.getClassFileVersion().isAtLeast( ClassFileVersion.JAVA_V6 ) ) {
```
Sanne, line 181: Could you comment about the class version requirement?

I'm also wondering if we should test this on pre-6 java class versions, but that's going to be complicated with our CI setup as the main codebase requires Java 8.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/InlineDirtyCheckingHandler.java
```diff
@@ -96,6 +101,93 @@ public Size apply(
 			MethodVisitor methodVisitor,
 			Context implementationContext,
 			MethodDescription instrumentedMethod) {
+
```
Sanne, line 104: just wondering, but was there a specific reason to have preferred using an ASM visitor over using a ByteBuddy template? Could you add a comment about the decision, for sake of future maintainers?

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/InlineDirtyCheckingHandler.java
pludov: Stackmap bytecodes have been introduced in JAVA6 : https://stackoverflow.com/questions/25109942/what-is-a-stack-map-frame.
I just followed already existing pattern here (there was already that kind of test near the end).

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/InlineDirtyCheckingHandler.java
pludov: No particular reason. I've just been following the already existing pattern in that class.

dreab8:
Hi @pludov ,

thanks for your PR and sorry for taking so long to review it. It seems this issue has been solved ([PR](https://github.com/hibernate/hibernate-orm/pull/9386)) so I'm going to close this.

PR #3579, opened 2020-10-02
Fix automatic releease script, changelog.txt contains wrong Jira rele…

dreab8:
…ase url

### DIFF
--- a/release/release.gradle
+++ b/release/release.gradle
@@ -299,8 +299,6 @@ task changeLogFile( dependsOn: [releaseChecks] ) {
     }
 }
 
-
-
 task addVersionCommit( dependsOn: [changeLogFile] ) {
     group = "Release"
     description = "Adds a commit for the released version and push the changes to github"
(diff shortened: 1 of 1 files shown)


release/release.gradle
```diff
@@ -433,6 +430,17 @@ class ChangeLogFile {
...
+    private static getVersionId(jsonReleaseNotes, String restReleaseVersion) {
+        def fixVersions = jsonReleaseNotes.issues.get( 0 ).fields.fixVersions
+
+        for ( def fixVersion : fixVersions ) {
+            if ( fixVersion.name.equals( restReleaseVersion ) ) {
+                return fixVersion.id
+            }
+        }
+        throw new GradleException( "Unable to determine the version id of the current release." )
+    }
```
NathanQingyangXu, line 443: There is more idiomatic Groovy way to achieve the logic as
```
return Optional.ofNullable( fixVersions.find { it.name.equals( restReleaseVersion ) } ).orElseThrow{new GradleException( 'Unable to determine the version id of the current release.' ) }
```
though not a big deal and feel free to ignore my comment.

PR #3580, opened 2020-10-05
HHH-14227 fix some ActionQueue insert ordering regression issues

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14227

It turns out to be a simple regression issue. 

We have complete insert ordering testing cases, but a good portion of them lacks SQL ordering verification (no referential integrity exception thrown is not necessarily identical to insert ordering correctness). In light of this big defect in existing testing code, I tried to add as many as possible SQL batch ordering verification.

The two new testing cases demonstrating regression failure are:
* `org.hibernate.test.insertordering.InsertOrderingSelfReferenceTest`
* `org.hibernate.test.insertordering.InsertOrderingReferenceDifferentSubclassTest`

The share the same pattern that null foreign keys are not ignored during dependency checking.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
@@ -45,6 +45,7 @@ import org.hibernate.cache.CacheException;
 import org.hibernate.engine.internal.NonNullableTransientDependencies;
 import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
+import org.hibernate.internal.util.collections.CollectionHelper;
 import org.hibernate.metadata.ClassMetadata;
 import org.hibernate.proxy.HibernateProxy;
 import org.hibernate.proxy.LazyInitializer;
@@ -105,7 +106,7 @@ public class ActionQueue {
 	 */
 	private static final LinkedHashMap<Class<? extends Executable>,ListProvider> EXECUTABLE_LISTS_MAP;
 	static {
-		EXECUTABLE_LISTS_MAP = new LinkedHashMap<Class<? extends Executable>,ListProvider>( 8 );
+		EXECUTABLE_LISTS_MAP = new LinkedHashMap<>( CollectionHelper.determineProperSizing( 8 ) );
 
 		EXECUTABLE_LISTS_MAP.put(
 				OrphanRemovalAction.class,
@@ -1101,7 +1102,7 @@ public class ActionQueue {
 			}
 
 			/**
-			 * Check if the this {@link BatchIdentifier} has a parent or grand parent
+			 * Check if this {@link BatchIdentifier} has a parent or grand parent
 			 * matching the given {@link BatchIdentifier} reference.
 			 *
 			 * @param batchIdentifier {@link BatchIdentifier} reference
@@ -1268,7 +1269,9 @@ public class ActionQueue {
 				for ( int i = 0; i < propertyValues.length; i++ ) {
 					Object value = propertyValues[i];
 					Type type = propertyTypes[i];
-					addParentChildEntityNameByPropertyAndValue( action, batchIdentifier, type, value );
+					if ( value != null ) {
+						addParentChildEntityNameByPropertyAndValue( action, batchIdentifier, type, value );
+					}
 				}
 
 				if ( identifierType.isComponentType() ) {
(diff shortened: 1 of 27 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -1129,9 +1127,6 @@ private boolean hasParent(BatchIdentifier batchIdentifier, List<BatchIdentifier>
 			}
 		}
 
-		// the mapping of entity names to their latest batch numbers.
-		private List<BatchIdentifier> latestBatches;
```
NathanQingyangXu, line 1133: the above should be temporary variable within method, as opposed to a private field

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -105,7 +106,7 @@
 	 */
 	private static final LinkedHashMap<Class<? extends Executable>,ListProvider> EXECUTABLE_LISTS_MAP;
 	static {
-		EXECUTABLE_LISTS_MAP = new LinkedHashMap<Class<? extends Executable>,ListProvider>( 8 );
+		EXECUTABLE_LISTS_MAP = new LinkedHashMap<>( CollectionHelper.determineProperSizing( 8 ) );
```
NathanQingyangXu, line 109: The above usage is a common anti-pattern which almost always ends up with rehashing.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -1268,7 +1269,9 @@ private void addParentChildEntityNames(AbstractEntityInsertAction action, BatchI
 				for ( int i = 0; i < propertyValues.length; i++ ) {
 					Object value = propertyValues[i];
 					Type type = propertyTypes[i];
-					addParentChildEntityNameByPropertyAndValue( action, batchIdentifier, type, value );
+					if ( value != null ) {
+						addParentChildEntityNameByPropertyAndValue( action, batchIdentifier, type, value );
+					}
```
NathanQingyangXu, line 1274: The above simple code change fixes the regression issues. When the property value is null, we don't need to consider its referential integrity.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -1300,11 +1303,9 @@ private void addParentChildEntityNameByPropertyAndValue(AbstractEntityInsertActi
...
 						batchIdentifier.getParentEntityNames().add( entityName );
 					}
-					if ( value != null ) {
-						String valueClass = value.getClass().getName();
-						if ( !valueClass.equals( entityName ) ) {
-							batchIdentifier.getParentEntityNames().add( valueClass );
-						}
+					String valueClass = value.getClass().getName();
+					if ( !valueClass.equals( entityName ) ) {
+						batchIdentifier.getParentEntityNames().add( valueClass );
```
NathanQingyangXu, line 1308: Now we know `value` cannot be null

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
Sanne: Are you sure? when hinted at the initial-capacity, this specific constructor already does a similar calculation.

If you're right that there are improvements to be done here, I'd still prefer to see them in a separate issue (and commit). I hope you don't mind, the reason is the fix seems very important but it also has become very large.

Sanne:
Awesome to see many tests added, I really appreciate that!

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
Sanne: Some tests seem to disagree :)
See `FumTest`

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
Sanne: Ok I see now the fix is actually very small, sorry I got scared by seeing "27 files changed" ...

Sanne:
I've reworked the changesets a bit. Could you check if you like https://github.com/hibernate/hibernate-orm/pull/3581 ?
Closing this one.

PR #3583, opened 2020-10-05
HHH 14241 Support ImplicitNamingStrategyComponentPathImpl with IdClass

jrenaat:


dreab8 (approved):
It looks good to me :+1:

sebersole (changes requested):
Generally I am not a fan of processing Strings that are the result of other String processing.

E.g. here you allow a recursive call to continue building a String(Builder/Buffer).  You then return that buffer in full and perform a sub-string on it.  I think a better solution would be to handle this within the `#process` method instead

dreab8:
@sebersole in this particular case is it not better in terms ofr performance to try the replacement just once in the `transformAttributePath` method, instead of multiple times inside the `process` method?

sebersole:
@dreab8 Fair point.  But unless I am mistaken, these only ever appear at the start of a path.  So in `#process` you would already be limited to the `attributePath.getParent() == null` case, which we already check for anyway:

```
	public static void process(AttributePath attributePath, StringBuilder sb) {
		if ( attributePath.getParent() != null ) {
			process( attributePath.getParent(), sb );
			if ( StringHelper.isNotEmpty( attributePath.getParent().getProperty() ) ) {
				sb.append( '_' );
			}
		}
		else {
			// I added this else - this where we would check attributePath.getProperty() for "_identifierMapper"
		}

		String property = attributePath.getProperty();
		property = property.replace( "<", "" );
		property = property.replace( ">", "" );

		sb.append( property );
	}
```

In fact, I'd change that like:

```
	public static boolean process(AttributePath attributePath, StringBuilder sb) {
		String property = attributePath.getProperty();

		if ( attributePath.getParent() != null ) {
			final parentAdded = process( attributePath.getParent(), sb );
			if ( parentAdded ) ) {
				sb.append( '_' );
			}
		}
		else if ( "_identifierMapper".equals( property ) {
			// skip it, do not pass go
			return false;
		}

		property = property.replace( "<", "" );
		property = property.replace( ">", "" );

		sb.append( property );

		return true;
	}
```

I _think_ that will be more efficient, but cannot say for sure I guess

sebersole:
Either way, that's just a matter of timing.   Good job finding the problem and solution @jrenaat

sebersole (approved):
I'm going to remove the request for changes.  As Andrea points out, the performance difference is probably not worth requiring a change

jrenaat:
> Good job finding the problem and solution @jrenaat

Ok, some irony here @sebersole ... ;DD
But I can certainly appreciate that, thanks for the help

dreab8:
@sebersole I think your idea of checking  `"_identifierMapper".equals( property )` inside the `process` method is the best in term of performance. I just noticed that `attributePath.getParent()` can be not null but with a `property` with an empty string value, so probably the change should be

```
public static void process(AttributePath attributePath, StringBuilder sb) {
		String property = attributePath.getProperty();

		final AttributePath parent = attributePath.getParent();
		if ( parent != null && !parent.getProperty().equals( "" ) ) {
			process( parent, sb );
			if ( StringHelper.isNotEmpty( parent.getProperty() ) ) {
				sb.append( '_' );
			}
		}
		else if ( PropertyPath.IDENTIFIER_MAPPER_PROPERTY.equals( property ) ) {
			// skip it, do not pass go
			sb.append( "id" );
			return;
		}
		property = property.replace( "<", "" );
		property = property.replace( ">", "" );

		sb.append( property );
	}
```
@jrenaat really good job :+1:

sebersole:
@dreab8 You can remove the following block now too:

```
if ( StringHelper.isNotEmpty( parent.getProperty() ) ) {
	sb.append( '_' );
}
```

```
public static void process(AttributePath attributePath, StringBuilder sb) {
		String property = attributePath.getProperty();

		final AttributePath parent = attributePath.getParent();
		if ( parent != null && StringHelper.isNotEmpty( parent.getProperty() ) ) {
			process( parent, sb );
			sb.append( '_' );
		}
		...
}
```

hibernate-core/src/main/java/org/hibernate/boot/model/naming/ImplicitNamingStrategyComponentPathImpl.java
```diff
@@ -30,14 +31,17 @@ protected String transformAttributePath(AttributePath attributePath) {
...
-		if ( attributePath.getParent() != null ) {
-			process( attributePath.getParent(), sb );
-			if ( StringHelper.isNotEmpty( attributePath.getParent().getProperty() ) ) {
-				sb.append( '_' );
-			}
-		}
-
 		String property = attributePath.getProperty();
+		final AttributePath parent = attributePath.getParent();
+		if ( parent != null && !parent.getProperty().equals( "" ) ) {
```
NathanQingyangXu, line 36: I remember we have discussed the bad performance of `xxx.equals( "" )` and @Sanne has created a util method to avoid it. We can simply use `StringHelper.isNotEmpty( parent.getProperty )` here after parent's nullness checking.

NathanQingyangXu: see https://github.com/hibernate/hibernate-orm/pull/3491

hibernate-core/src/test/java/org/hibernate/id/idclass/IdClassNamingStrategyTest.java
```diff
@@ -0,0 +1,48 @@
...
+		 * create table MyEntity (
+         *   id_idA bigint not null,
+         *   id_idB bigint not null,
+         *   _identifierMapper_idA bigint not null, <-- ??
+         *   _identifierMapper_idB bigint not null, <-- ??
+         *   notes varchar(255),
+         *   primary key (id_idA, id_idB)
+         * )
+		 */
+		configuration.setImplicitNamingStrategy( new ImplicitNamingStrategyComponentPathImpl() );
```
NathanQingyangXu, line 36: It would be great to avoid TAB and space mixture for the above comments to avoid inconsistent presenting. The above is a good example.

hibernate-core/src/test/java/org/hibernate/id/idclass/IdClassNamingStrategyTest.java
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	@Test
+	public void test() {
```
NathanQingyangXu, line 40: We'd better add `@TestForIssue` annotation above

hibernate-core/src/test/java/org/hibernate/id/idclass/MyEntity.java
```diff
@@ -0,0 +1,41 @@
...
+	@Id
+	private Long idA;
+
+	@Id
+	private Long idB;
+
+	private String notes;
+
+	public MyEntityId getId() {
+		return new MyEntityId( idB, idA );
```
NathanQingyangXu, line 26: I am wondering why the above is not `new MyEntityId( idA, idB )`, which seems counter-intuitive and confusing.

hibernate-core/src/test/java/org/hibernate/id/idclass/MyEntityId.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.id.idclass;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+public class MyEntityId implements Serializable {
```
NathanQingyangXu, line 12: It would be good practise to use package default visibility (remove `public` above) for these classes are only used by the testing case within the same package.

hibernate-core/src/main/java/org/hibernate/boot/model/naming/ImplicitNamingStrategyComponentPathImpl.java
```diff
@@ -30,14 +31,17 @@ protected String transformAttributePath(AttributePath attributePath) {
...
+			process( parent, sb );
+			sb.append( '_' );
+		}
+		else if ( PropertyPath.IDENTIFIER_MAPPER_PROPERTY.equals( property ) ) {
+			// skip it, do not pass go
+			sb.append( "id" );
+			return;
+		}
 		property = property.replace( "<", "" );
 		property = property.replace( ">", "" );
```
NathanQingyangXu, line 46: There must be a more efficient way to replace the above two statements with one without much performance cost, though I am not able to find one in our existing classes (StringUtils from apache commons seems forbidden).

NathanQingyangXu:
@jrenaat you enjoy the luxury to get @sebersole 's attention. My others would envy you a lot (including me), :).

hibernate-core/src/main/java/org/hibernate/boot/model/naming/ImplicitNamingStrategyComponentPathImpl.java
sebersole: Unless I am mistaken, the '<' and '>' can only occur as the first and last in the prop-name.  Direct access to the underlying char array is usually quite fast - 

```
if ( property.charAt( 0 ) == '<' && property.charAt( property.length() -1 ) == '>' ) {
    property = property.substring( ... );
}
```

sebersole:
> @jrenaat you enjoy the luxury to get @sebersole 's attention. My others would envy you a lot (including me), :).

Sorry Nathan, I have been taking a little break from 6.0 dev for the past 2 weeks or so to work on some Gradle and Quarkus stuff.  I'll get back to your commits :)

hibernate-core/src/main/java/org/hibernate/boot/model/naming/ImplicitNamingStrategyComponentPathImpl.java
sebersole: Yes, that was burried in my comment at https://github.com/hibernate/hibernate-orm/pull/3583#issuecomment-705059462 but I completely agree it should use `#isNotEmpty`

hibernate-core/src/test/java/org/hibernate/id/idclass/MyEntityId.java
jrenaat: I prefer leaving this as-is

jrenaat:
> @jrenaat you enjoy the luxury to get @sebersole 's attention. My others would envy you a lot (including me), :).

I'm blackmailing him ;)
Anyway, thanks for your comments, I've taken most of them into account

PR #3584, opened 2020-10-05
HHH-14227 fix testing failure on Oracle

NathanQingyangXu:
Oracle stores null value when Lob field contains empty content, which happens to be the case for the original testing case.


Sanne:
thanks!

PR #3581, opened 2020-10-05
HHH-14227 fix some ActionQueue insert ordering regression issues

Sanne:
Replaces #3580 , just reorganized the changest a bit. All credit to @NathanQingyangXu 

 - https://hibernate.atlassian.net/browse/HHH-14227

PR #3585, opened 2020-10-06
HHH-14248 minor improvements to ActionQueue

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14248

some minor improvements to ActionQueue.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
@@ -1129,9 +1129,6 @@ public class ActionQueue {
 			}
 		}
 
-		// the mapping of entity names to their latest batch numbers.
-		private List<BatchIdentifier> latestBatches;
-
 		// the map of batch numbers to EntityInsertAction lists
 		private Map<BatchIdentifier, List<AbstractEntityInsertAction>> actionBatches;
 
@@ -1143,8 +1140,10 @@ public class ActionQueue {
 		 */
 		public void sort(List<AbstractEntityInsertAction> insertions) {
 			// optimize the hash size to eliminate a rehash.
-			this.latestBatches = new ArrayList<>( );
 			this.actionBatches = new HashMap<>();
+
+			// the mapping of entity names to their latest batch numbers.
+			final List<BatchIdentifier> latestBatches = new ArrayList<>();
 
 			for ( AbstractEntityInsertAction action : insertions ) {
 				BatchIdentifier batchIdentifier = new BatchIdentifier(
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
@@ -1155,8 +1155,6 @@ public class ActionQueue {
 								.getRootEntityName()
 				);
 
-				// the entity associated with the current action.
-				Object currentEntity = action.getInstance();
 				int index = latestBatches.indexOf( batchIdentifier );
 
 				if ( index != -1 )  {
@@ -1266,8 +1264,8 @@ public class ActionQueue {
 
 				for ( int i = 0; i < propertyValues.length; i++ ) {
 					Object value = propertyValues[i];
-					Type type = propertyTypes[i];
 					if ( value != null ) {
+						Type type = propertyTypes[i];
 						addParentChildEntityNameByPropertyAndValue( action, batchIdentifier, type, value );
 					}
 				}


hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -105,7 +105,7 @@
 	 */
 	private static final LinkedHashMap<Class<? extends Executable>,ListProvider> EXECUTABLE_LISTS_MAP;
 	static {
-		EXECUTABLE_LISTS_MAP = new LinkedHashMap<Class<? extends Executable>,ListProvider>( 8 );
+		EXECUTABLE_LISTS_MAP = new LinkedHashMap<Class<? extends Executable>,ListProvider>( CollectionHelper.determineProperSizing( 8 ) );
```
NathanQingyangXu, line 108: Given that `LinkedHashMap` extends from `HashMap`, the following common usage:
```
new LinkedHashMap<>( 8 );
```
almost always end up with resizing for load factor is 0.75 by default, not 1. So in this specific scenario, the map will be resized when the 7th class is inserted into the map, which defeats the purpose of the constructor parameter.

A common misuse, just similar to the case for `HashMap`. Only `ConcurrentHashMap`'s constructor take load factor into account.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -634,7 +632,7 @@ private static boolean areTablesToBeUpdated(UnresolvedEntityInsertActions action
 	}
 
 	private static String[] convertTimestampSpaces(Set spaces) {
-		return (String[]) spaces.toArray( new String[ spaces.size() ] );
+		return (String[]) spaces.toArray( new String[0] );
```
NathanQingyangXu, line 635: As well known, modern JVM has better performance using zero length string as the parameter of `toArray()`.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -1129,9 +1127,6 @@ private boolean hasParent(BatchIdentifier batchIdentifier, List<BatchIdentifier>
 			}
 		}
 
-		// the mapping of entity names to their latest batch numbers.
-		private List<BatchIdentifier> latestBatches;
-
```
NathanQingyangXu, line 1134: The above field should be put into method as temporary variable rather than a field

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -1290,7 +1285,7 @@ private void addParentChildEntityNameByPropertyAndValue(AbstractEntityInsertActi
 				final String entityName = entityType.getName();
 				final String rootEntityName = action.getSession().getFactory().getMetamodel().entityPersister( entityName ).getRootEntityName();
 
-				if ( entityType.isOneToOne() && OneToOneType.class.cast( entityType ).getForeignKeyDirection() == ForeignKeyDirection.TO_PARENT ) {
+				if ( entityType.isOneToOne() && entityType.getForeignKeyDirection() == ForeignKeyDirection.TO_PARENT ) {
```
NathanQingyangXu, line 1288: We don't need to down cast to invoke the method for Java's polymorphism is builtin feature. Maybe the casting enjoys better performance?

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -1302,11 +1297,9 @@ private void addParentChildEntityNameByPropertyAndValue(AbstractEntityInsertActi
...
 						batchIdentifier.getParentEntityNames().add( entityName );
 					}
-					if ( value != null ) {
-						String valueClass = value.getClass().getName();
-						if ( !valueClass.equals( entityName ) ) {
-							batchIdentifier.getParentEntityNames().add( valueClass );
-						}
+					String valueClass = value.getClass().getName();
+					if ( !valueClass.equals( entityName ) ) {
+						batchIdentifier.getParentEntityNames().add( valueClass );
```
NathanQingyangXu, line 1302: The above is leftover of https://github.com/hibernate/hibernate-orm/pull/3581. We don't need nullness checking here for the only upstream method ensures that the value is never null.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
Sanne: hi @NathanQingyangXu , I know about map resizes - but I still don't see the point.

First off, this is a static constant which is initialized only once per classloader so it's not really worth your time, nor mine, to try avoid a single collection resize.

Also this is the maintenance branch, on which each change needs to then be ported to 6 and it slows down the 6 development - which we can't have :) Let's please do only important fixes or changes we actually need: such minor improvements are not healthy for the project as they steal vital focus from other things.

But even if I don't think we should apply this, I'm happy to discuss this case in more detail as it's interesting. I believe 8 is a reasonable estimate, it's not possible to have this exact. It's a map, so keys are going to be hashed and therefore there are likely collisions, leading to some elements to share a bucket: that's why the constructor applies a load factor.

You can debug this and verify that most of the times with 8 there is either 1 or 2 resizings that need to happen: one is always going to be needed as the buckets are lazily allocated.
It's "most of the times" and not "always" as the key is a Class, and its hashcode will depend on its identity, which might be different if you run this different times. Working hard on making this resize only once in all cases would come at a memory penalty for the cases in which a single resize was good enough, so it's not a silver bullet either.

IMO if you want to really optimise this code path that's great, but I think there's potential here to not use a LinkedHashMap at all, benefit from immutability, and the change would belong in 6. At very least since this is a constant you could replace the 8 with a slightly larger value; e.g. using "10" would make a resize need more unlikely - but still not guaranteed either.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
Sanne: As I mentioned in the previous PR, that's wrong and there's failing tests if you make this change.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
Sanne: I totally agree with this one, could you extract it into a separate patch?

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
Sanne: you're right but the difference is extremely tiny, not worth doing in 5.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
NathanQingyangXu: After second-thought, I reverted back this code change for we have solve it in v6 in systematic way. No need to bring about merging conflict furthermore.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
NathanQingyangXu: yeah, I realized for I found the broken TumTest case. Reverted back.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
NathanQingyangXu: I reverted it back, together with many other trivial changes.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
NathanQingyangXu: What did you mean? Creating a separate git commit in this PR?

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -1341,12 +1337,11 @@ else if ( type.isComponentType() && value != null ) {
...
-			List<AbstractEntityInsertAction> actions = actionBatches.get( batchIdentifier );
+			List<AbstractEntityInsertAction> actions = actionBatches.computeIfAbsent(
+					batchIdentifier,
+					k -> new LinkedList<>()
+			);
 
-			if ( actions == null ) {
-				actions = new LinkedList<>();
-				actionBatches.put( batchIdentifier, actions );
-			}
```
NathanQingyangXu, line 1349: @Sanne , do you think the above code change is necessary? I can revert it back without problem.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -1156,8 +1154,6 @@ public void sort(List<AbstractEntityInsertAction> insertions) {
 								.getRootEntityName()
 				);
 
-				// the entity associated with the current action.
-				Object currentEntity = action.getInstance();
```
NathanQingyangXu, line 1160: The above variable is never used.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
Sanne: yes, we have occasionally needed an explicit cast introduced by the performance team. I don't remember if this is the case, but it would be safer to not clean these up blindly: an excplicit cast has the ability to simplify a megamorphic call and we have a couple of extreme cases.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
Sanne: yes, either a separate git commit or a separate PR, so to be able to backport this fix only (and only this), in case of need.

No need to have a separate PR if most other changes are undone.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
NathanQingyangXu: reverted it back.

hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
NathanQingyangXu: Now after many reverting, seems not much code change is left. Maybe we can keep it as it is now?

NathanQingyangXu: done. Now a separate git commit is created for this change exlusively.

gavinking:
@NathanQingyangXu please let me know when you're done with this so we can replicate the changes to HR. Or just send us a pull.

NathanQingyangXu:
I have done already. Awaiting approval. Super minor changes left and feel
free to ignore it.

Sanne:
Thanks Nathan, nice cleanup. I've merged it.

gavinking:
We missed replicating this in HR.

gavinking:
> We missed replicating this in HR.

Done now.

PR #3594, opened 2020-10-13
Add test cases for HHH-14256

tobyhammett:
These tests currently fail, but demonstrate HHH-14256.

https://hibernate.atlassian.net/browse/HHH-14256


NathanQingyangXu:
@yrodiere feel free to take a look at this ticket for the reporter claimed it is a regression issue introduced by https://hibernate.atlassian.net/browse/HHH-13266. Seems complicated for it seems your PR was reverted back later.

yrodiere:
@NathanQingyangXu Thanks, but I don't really have the time right now. I'll have a look in a few weeks if nobody else does. If my Anyway, if my PR was reverted, I don't see how it could be causing a regression.

NathanQingyangXu:
> @NathanQingyangXu Thanks, but I don't really have the time right now. I'll have a look in a few weeks if nobody else does. If my Anyway, if my PR was reverted, I don't see how it could be causing a regression.

I got confused as well, thus pinged you. Anyway, feel free to ignore and thanks for your response.

gavinking:
We should prioritize fixing this bug, even though it's really a bug in the JDK. From the description in the issue, I think the fix should be straightforward,

jrenaat:
Superseded by https://github.com/hibernate/hibernate-orm/pull/9595

PR #3595, opened 2020-10-15
HHH-14257 An Entity A with a map collection having as index an Embeddable with a an association to the Entity A fails with a NPE

dreab8:
https://hibernate.atlassian.net/browse/HHH-14257

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/plan/build/internal/AbstractLoadPlanBuildingAssociationVisitationStrategy.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/plan/build/internal/AbstractLoadPlanBuildingAssociationVisitationStrategy.java
@@ -12,8 +12,6 @@ import java.util.Map;
 
 import org.hibernate.HibernateException;
 import org.hibernate.engine.FetchStrategy;
-import org.hibernate.engine.FetchStyle;
-import org.hibernate.engine.FetchTiming;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.internal.util.StringHelper;
 import org.hibernate.loader.PropertyPath;
@@ -44,6 +42,7 @@ import org.hibernate.persister.walking.spi.AnyMappingDefinition;
 import org.hibernate.persister.walking.spi.AssociationAttributeDefinition;
 import org.hibernate.persister.walking.spi.AssociationKey;
 import org.hibernate.persister.walking.spi.AttributeDefinition;
+import org.hibernate.persister.walking.spi.AttributeSource;
 import org.hibernate.persister.walking.spi.CollectionDefinition;
 import org.hibernate.persister.walking.spi.CollectionElementDefinition;
 import org.hibernate.persister.walking.spi.CollectionIndexDefinition;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/test/collection/map/EmbeddableIndexTest.java
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	@Test
+	public void testIt() {
+		inSession(
+				session -> {
+					TheOne one = session.get( TheOne.class, "1" );
+					TheMapKey theMapKey = one.getTheManys().keySet().iterator().next();
+					assertThat( theMapKey, is( notNullValue() ) );
+					assertThat( theMapKey.getTheOne(), is( one ) );
```
NathanQingyangXu, line 63: It might be more accurate to use `sameInstance()` instead of `is()` for the following reasons:
* `is()` is based on `equalsTo()` but it seems there is no `equalsTo()` and `hashcode()` overriding in the class;
* should we allow `theMapKey.getTheOne()` returns another instance other than `one`? Seems not, then `sameInstance()` might be more accurate to express your intention.

hibernate-core/src/main/java/org/hibernate/loader/plan/build/internal/AbstractLoadPlanBuildingAssociationVisitationStrategy.java
```diff
@@ -690,6 +692,14 @@ public void foundCircularAssociation(AssociationAttributeDefinition attributeDef
 		}
 	}
 
+	private Joinable locateJonable(AttributeSource source) {
```
Sanne, line 695: there's a typo in the method name

Sanne (changes requested):

hibernate-core/src/test/java/org/hibernate/test/collection/map/EmbeddableIndexTest.java
dreab8: Thanks @NathanQingyangXu fixed it

hibernate-core/src/main/java/org/hibernate/loader/plan/build/internal/AbstractLoadPlanBuildingAssociationVisitationStrategy.java
dreab8: Thanks @Sanne fixed it

hibernate-core/src/main/java/org/hibernate/loader/plan/build/internal/AbstractLoadPlanBuildingAssociationVisitationStrategy.java
```diff
@@ -690,6 +690,14 @@ public void foundCircularAssociation(AssociationAttributeDefinition attributeDef
 		}
 	}
 
+	private Joinable locateJoinable(AttributeSource source) {
+		if ( source instanceof CompositionDefinition ) {
+			return locateJoinable( ( (CompositionDefinition) source ).getSource() );
+		}
+
+		return (Joinable) ( (EntityDefinition) source ).getEntityPersister();
```
Sanne, line 698: I don't know much about this code section, but it seems a bit suspicious that the "else" branch here is taking an entirely different approach than the current code?

It seems that the bugfix should only affect the case in which is it a `CompositionDefinition`, and it makes me a bit uncomfortable to change the other branch as well without understanding more about the reasons.

dreab8: @Sanne I have worked on a different solution https://github.com/hibernate/hibernate-orm/pull/3635

Sanne:
Closing in favour of #3635

PR #3600, opened 2020-10-19
HHH-14234 Fix CI building error on HANA dialect

NathanQingyangXu:
HANA dialect has special requirement regarding table name in terms of case sensibility. Renaming table names to UPPERCASE should solve the issue (another option is to put quotation marks around table names). See https://stackoverflow.com/questions/29974507/sap-hana-invalid-table-name

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/inheritance/DenormalizedTablePhysicalIncludedTableConstraintTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/inheritance/DenormalizedTablePhysicalIncludedTableConstraintTest.java
@@ -12,11 +12,7 @@ import javax.persistence.Table;
 import javax.persistence.UniqueConstraint;
 
 import org.hibernate.cfg.AvailableSettings;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/test/java/org/hibernate/test/inheritance/DenormalizedTablePhysicalIncludedTableConstraintTest.java
```diff
@@ -26,10 +22,6 @@
  * @author Nathan Xu
  */
 @TestForIssue( jiraKey = "HHH-14234" )
-@SkipForDialects( {
-		@SkipForDialect( value = MySQLDialect.class, comment = "skip it for it support constraint name uniqueness in table, not db" ),
-		@SkipForDialect( value = MariaDBDialect.class, comment = "skip it for it support constraint name uniqueness in table, not db" )
-} )
```
NathanQingyangXu, line 32: Seems the above two dialects are not the only ones which don't treat uniqueness constraint name unique within db. Anyway, including them doesn't hurt.

PR #3602, opened 2020-10-19
HHH-14255 Allow to define an association between an entity and a MappedSuperclass that extends a JOINED or UNION Inheritance

dreab8:
https://hibernate.atlassian.net/browse/HHH-14255

Not sure if there is a better way for managing the `ProxyFactory` for a `MappedSuperclass` https://github.com/hibernate/hibernate-orm/commit/d948e6d7fa7f8fd6f6e6320464ff9cdc1c13a16f
`

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
@@ -123,6 +123,7 @@ import org.hibernate.persister.walking.spi.AttributeDefinition;
 import org.hibernate.persister.walking.spi.EntityIdentifierDefinition;
 import org.hibernate.pretty.MessageHelper;
 import org.hibernate.property.access.internal.PropertyAccessStrategyBackRefImpl;
+import org.hibernate.proxy.ProxyFactory;
 import org.hibernate.sql.Alias;
 import org.hibernate.sql.Delete;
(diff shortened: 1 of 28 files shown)


hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -5011,10 +5016,34 @@ protected final OptimisticLockStyle optimisticLockStyle() {
...
+		mappedSuperclassSubclassProxyFactory.put( mappedSuperclassName, proxyFactory );
+	}
+
+	@Override
+	public boolean isMappedSuperclassSubclass(String className) {
+		if ( mappedSuperclassSubclassProxyFactory == null ) {
+			return false;
+		}
+		return mappedSuperclassSubclassProxyFactory.containsKey( className );
+	}
```
NathanQingyangXu, line 5045: usually IDE would warn that the above code snippet can be simplified as
```
return mappedSuperclassSubclassProxyFactory != null && mappedSuperclassSubclassProxyFactory.containsKey( className );
```
Personally I agree with the IDE's suggestion. Ignore my comment if you disagree, though.

gavinking:
This will impact HR. Please let us know when it merges.

dreab8:
I'm going to close this PR since the Jira has been rejected

PR #3604, opened 2020-10-20
show more detailed errors when getting connection metadata

DanielYWoo:
We have some customization with the connection and datasource, when exceptions happen we cannot see where the problem is. E.g, when getConnection is called and a NPE is thrown, all we get is a very generic NPE error message. The change is to show more details about the error, easier to troubleshoot.

BTW, I see two methods could have typo, added two lines of comments too.


Sanne:
Good idea, thanks! https://hibernate.atlassian.net/browse/HHH-14268

Sanne:
Thanks @DanielYWoo , I've merged it and backported to 5.4

PR #3607, opened 2020-10-22
HHH-14276 test case for strange AnnotationException: MapsId in IdClass context

kawoolutions:
Minimal test case for HHH-14276


beikov:
Thanks for the test. Superseded by https://github.com/hibernate/hibernate-orm/pull/3641

PR #3612, opened 2020-10-25
HHH-14280 - audit query does not work when @IdClass composite identifier contains association to a not audited entity

ivanrapenok:
The first commit contains test case
The second one contains possible solution


Naros:
Superseded by PR #4490.

PR #3614, opened 2020-10-26
HHH-14279 Test and fix for join ordering issue when entity key joins are involved

beikov:
https://hibernate.atlassian.net/browse/HHH-14279

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromClause.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/FromClause.java
@@ -94,6 +94,9 @@ public class FromClause extends HqlSqlWalkerNode implements HqlSqlTokenTypes, Di
 	void moveFromElementToEnd(FromElement element) {
 		fromElements.remove( element );
 		fromElements.add( element );
+		// We must move destinations which must come after the from element as well
+		fromElements.removeAll( element.getDestinations() );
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/generated/org/hibernate/resource/beans/internal/BeansMessageLogger_$logger.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+    @Override
+    public final void stoppingBeanContainer(final BeanContainer beanContainer) {
+        log.logf(FQCN, INFO, null, stoppingBeanContainer$str(), beanContainer);
+    }
+    private static final String stoppingBeanContainer = "HHH10005004: Stopping BeanContainer : %s";
+    protected String stoppingBeanContainer$str() {
+        return stoppingBeanContainer;
+    }
+}
```
NathanQingyangXu, line 62: Seems the above classes can be auto generayed and are not supposed to be committed?

NathanQingyangXu (approved):
Looks good. Not sure whether we should commit the generated logger classes, though.

hibernate-core/src/main/generated/org/hibernate/resource/beans/internal/BeansMessageLogger_$logger.java
beikov: Right, I removed them again from the commit

PR #3617, opened 2020-10-30
Revert to HCANN 5.1.0.final compatibility

Sanne:
This is a *draft* and at best we'll apply it only to 5.4

Created because we suspect enforcing the requirement of updating to HCANN 5.1.1 is not suitable for a micro release, especially as WildFly users would need to also update their HCANN modules.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
@@ -310,7 +310,7 @@ public final class AnnotationBinder {
 		if ( packaze == null ) {
 			return;
 		}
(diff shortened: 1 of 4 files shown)


yrodiere (approved):
LGTM. Using reflection is not very satisfying, that's for sure, but it was like that before, we're just expanding it to a few other methods. I think it's good compromise if we want the next micro release to be backward-compatible.

I also tested these changes in Hibernate Search 5.11: everything works fine after this change, even in an older version of WildFly where HCANN wasn't updated.

Sanne:
thanks - assigned a JIRA and merged.

Sanne:
( https://hibernate.atlassian.net/browse/HHH-14298 )

PR #3618, opened 2020-11-02
HHH-14306 TcclLookupPrecedence parameter in JPA properties is ignored

PashaTurok:
Full description is [ here](https://hibernate.atlassian.net/browse/HHH-14306)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/ClassLoaderServiceImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/ClassLoaderServiceImpl.java
@@ -114,10 +114,9 @@ public class ClassLoaderServiceImpl implements ClassLoaderService {
 		addIfSet( providedClassLoaders, AvailableSettings.HIBERNATE_CLASSLOADER, configValues );
 		addIfSet( providedClassLoaders, AvailableSettings.ENVIRONMENT_CLASSLOADER, configValues );
 
-		TcclLookupPrecedence providedPrecedence = TcclLookupPrecedence.AFTER;
-		final String tcclLookupPrecedence = (String) configValues.get( AvailableSettings.TC_CLASSLOADER );
-		if ( tcclLookupPrecedence != null ) {
-			providedPrecedence =  TcclLookupPrecedence.valueOf( tcclLookupPrecedence.toUpperCase( Locale.ROOT ) );
+		TcclLookupPrecedence providedPrecedence = TcclLookupPrecedence.from( configValues );
+		if ( providedPrecedence == null ) {
+			providedPrecedence = TcclLookupPrecedence.AFTER;
 		}
 		return new ClassLoaderServiceImpl( providedClassLoaders,providedPrecedence );
 	}
--- a/hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/ClassLoaderServiceImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/ClassLoaderServiceImpl.java
@@ -114,11 +114,11 @@ public class ClassLoaderServiceImpl implements ClassLoaderService {
 		addIfSet( providedClassLoaders, AvailableSettings.HIBERNATE_CLASSLOADER, configValues );
 		addIfSet( providedClassLoaders, AvailableSettings.ENVIRONMENT_CLASSLOADER, configValues );
 
-                TcclLookupPrecedence providedPrecedence = TcclLookupPrecedence.AFTER;
-                final String tcclLookupPrecedence = (String) configValues.get( AvailableSettings.TC_CLASSLOADER );
-                if ( tcclLookupPrecedence != null ) {
-                        providedPrecedence =  TcclLookupPrecedence.valueOf( tcclLookupPrecedence.toUpperCase( Locale.ROOT ) );
-                }
+		TcclLookupPrecedence providedPrecedence = TcclLookupPrecedence.AFTER;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/ClassLoaderServiceImpl.java
```diff
@@ -113,7 +114,12 @@ public static ClassLoaderServiceImpl fromConfigSettings(Map configValues) {
 		addIfSet( providedClassLoaders, AvailableSettings.HIBERNATE_CLASSLOADER, configValues );
 		addIfSet( providedClassLoaders, AvailableSettings.ENVIRONMENT_CLASSLOADER, configValues );
 
-		return new ClassLoaderServiceImpl( providedClassLoaders,TcclLookupPrecedence.AFTER );
+		TcclLookupPrecedence providedPrecedence = TcclLookupPrecedence.AFTER;
```
sebersole, line 117:
```
final TcclLookupPrecedence tcclPrecedence = TcclLookupPrecedence.fromSetting( configValues );
return new ClassLoaderServiceImpl( providedClassLoaders,providedPrecedence );
```

```
public enum TcclLookupPrecedence {
    ...

    public static TcclLookupPrecedence from(Map<?,?> settings) {
        final String explicitSetting = (String) settings.get( AvailableSettings.TC_CLASSLOADER );
        if ( explicitSetting == null ) {
            return AFTER;
        }

        if ( NEVER.name().equalsIgnoreCase( explicitSetting ) ) {
            return NEVER;
        }
		
        if ( BEFORE.name().equalsIgnoreCase( explicitSetting ) ) {
            return BEFORE;
        }

        return AFTER;
    }
}
```

sebersole (changes requested):
Thanks for the contribution!  2 things:

- This should have a test
- It would be better to define a static "resolution" method on `TcclLookupPrecedence`.  Not critical here, but will perform better.  Gave an example of what I mean

hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/ClassLoaderServiceImpl.java
PashaTurok: I am glad to help!

1.  About static resolution. I copied code from this line [here](https://github.com/hibernate/hibernate-orm/blob/67fd24ecff9f9a9a2e848805c04590b4b4ee2b3e/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java#L447). If I add `TcclLookupPrecedence.from(Map)` method, should I also change and that line 447?

2. About test - unfortunately I know about hibernate too little and I just can't make a test. I need someone's help.

sebersole: > 1. About static resolution. I copied code from this line [here](https://github.com/hibernate/hibernate-orm/blob/67fd24ecff9f9a9a2e848805c04590b4b4ee2b3e/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java#L447). If I add `TcclLookupPrecedence.from(Map)` method, should I also change and that line 447?

Awesome, thanks!  Looks good.  With regard to line 447, in my opinion that should change also.

> 2. About test - unfortunately I know about hibernate too little and I just can't make a test. I need someone's help.

We have templates you can use to create tests.  They already do most of the heavy lifting for you.  See https://github.com/hibernate/hibernate-test-case-templates

We won't apply this without a test.  But I'll try to find some time to create one.  Would be awesome if you could write the test though.

PashaTurok: I would be very thankful if you provide a test. If I make a pull request again I will study it.

PashaTurok: @sebersole What about line 445 - 447, maybe in TcclLookupPrecedence#from we should check if map is Properties and if true then to use getProperty method? [Here](https://stackoverflow.com/a/11104420/5057736) it is said it is said that get() not always returns the same as getProperty(). It depends on how the properties were created. What do you think?

PashaTurok: @sebersole Any ideas?

PashaTurok:
No feedback. Closed.

karge-itestra:
@sebersole I would appreciate if this PR was reconsidered, the issue is still there 3 years later, and no one bothered to test this when the bug was created in 2016.

karge-itestra:
@PavelTurkish I would not generalize it, I had great encounters with some of the maintainers here, and while @sebersole is very strict about some things, I would not see this as bad attitude - the only thing I wish for is that issues are really followed through and not ignored after they become a little stale.

karge-itestra:
> Do you think I want something different?

Not at all, and I can totally understand your frustration. Let´s hope for the best :)

sebersole:
Hm, to be honest I thought we dropped `TcclLookupPrecedence`; but apprently not.  We don't often play in that part of the code.

I'll take a look.

sebersole:
>  When a few years I realized what the attitude

Look, I can understand your frustration.  But take that "attitude" and multiply times a fair amount of the user base.  On top of that, look at the sheer number of reports we get over 25 years.  So yeah, when things go stale we tend to lose sight of them.

PR #3620, opened 2020-11-04
renamed `NamedQueryRepository` to `NamedObjectRepository`

sebersole:
As per subject


sebersole:
Decided on a PR just to gauge the impact of changing this name in consumers

yrodiere (approved):
LGTM, the new name makes sense and this interface was incubating after all.

Hibernate Search does not use this interface.

If someone is worried about backward compatibility, you can just leave the `NamedQueryRepository` interface around, and have `NamedObjectRepository` extend it. As long as users don't implement `NamedQueryRepository` (and I don't think they are supposed to?), upgrades will go smoothly.

Sanne (approved):
Seems ok to me.

sebersole:
Thanks guys!

> If someone is worried about backward compatibility, you can just leave the `NamedQueryRepository` interface around, and have `NamedObjectRepository` extend it. As long as users don't implement `NamedQueryRepository` (and I don't think they are supposed to?), upgrades will go smoothly.

Yep I have used this frequently in the past.  As you said though, this is an incubating contract - in fact this whole network of objects starting from QueryEngine is new.  So if it won't negatively affect Search or Quarkus then I think its not worth it and I should just do the change as proposed in the PR

PR #3625, opened 2020-11-09
HHH-14316 HHH-14317 Avoid accessing potentially null variables

gsmet:
No rocket science here, when something is wrong in the setup (typically we cannot connect to the database), these variables can be null and we shouldn't end up with a NPE.

Noted while debugging a bug report in Quarkus.

https://hibernate.atlassian.net/browse/HHH-14316
https://hibernate.atlassian.net/browse/HHH-14317

Should be backported to 5.4.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
@@ -168,11 +168,17 @@ public class DriverManagerConnectionProviderImpl
 
 	@Override
 	public Connection getConnection() throws SQLException {
+		if ( state == null ) {
+			throw new IllegalStateException( "Cannot get a connection as the driver manager is not properly initialized" );
+		}
 		return state.getConnection();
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/engine/jdbc/connections/internal/DriverManagerConnectionProviderImpl.java
```diff
@@ -168,11 +168,17 @@ private static Driver loadDriverIfPossible(String driverClassName, ServiceRegist
 
 	@Override
 	public Connection getConnection() throws SQLException {
+		if ( state == null ) {
```
gsmet, line 171: I wasn't sure about this one. The Agroal implementation returns a `null` connection but that really doesn't sound like a good idea.

I'll let you judge of that, I can adjust (or even drop these 2 checks if you think they are useless).

Sanne: This seems sensible, thanks.

Sanne:
Thanks, merged and backported.

PR #3626, opened 2020-11-09
HHH-14315 Upgrade to Gradle 6.7 and use toolchains for per-JDK builds

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14315

After this PR, we will need a few change in our CI jobs, and then we will be able to test JDK16 (spoiler alert: it works).

This introduces similar changes to https://github.com/hibernate/hibernate-reactive/pull/429.
By default, nothing changes, except that we now use Gradle 6.7.

When specific options are enabled:

* We run Gradle with JDK8
* We compile the project with JDK 8
* We run the tests with a separate JVM version 11, 15 or 16.

Different configurations are also possible, e.g. run Gradle with JDK11, compile with JDK8 and run the tests with JDK15. But the one above is the most interesting and most likely the one we'll use.

@Sanne I think that, considering you need specific options to enable toolchains, we can safely backport to 5.4?


yrodiere:
Example build with JDK16: https://ci.hibernate.org/view/Personal%20jobs/job/hibernate-orm-personal-yoann-jdk16/12/console

gradle/java-module.gradle
```diff
@@ -112,17 +112,72 @@ dependencies {
 
 tasks.withType( JavaCompile ) {
 	options.encoding = 'UTF-8'
-	sourceCompatibility = project.baselineJavaVersion
```
Sanne, line 115: just wondering, was it necessary to remove these?

settings.gradle
```diff
@@ -13,35 +13,79 @@ if ( !JavaVersion.current().java8Compatible ) {
...
-        logger.warn( "Gradle does not support this JDK, because it is too recent; build is likely to fail." +
-                " To avoid failures, you should specify an older Java version in the 'testedJavaVersion' parameter." +
-                " Just append the following to your gradle command:" +
-                " '-PtestedJavaVersion=<major version of your newer JDK, e.g. 14>'" )
-        // Use a hack to retrieve the major as a string.
-        // This only works for Java 9+ (we're at least on Java 18 here).
-        gradle.ext.testedJavaVersion = System.getProperty( 'java.specification.version' )
+gradle.ext.baselineJavaVersion = JavaLanguageVersion.of( 8 )
+
+// Gradle does bytecode transformation on tests.
```
Sanne, line 18: hu?? Have a pointer about what kinds of things it needs doing?

Sanne (approved):

Sanne:
Looks great! I'll merge it in master, but defer backporting it a bit. This might allow making some more changes in case there's need for CI setup?

@sebersole you might be interested? Much more flexibility and control of JDKs being used for build/release/testing.

settings.gradle
yrodiere: Not really, but you can try setting the max supported version to 16 and running the build on 16: you'll see errors related to "incompatible minor.major" pop up, originating from Gradle.

Maybe Gradle just inspects the bytecode, I don't know. But it definitely is sensitive to the bytecode version.

Sanne: ok, thanks.

gradle/java-module.gradle
yrodiere: I didn't remove these, I just moved them further below. We can't use `sourceCompatibility` and such when toolchains are enabled, so I needed an `if` block.

PR #3628, opened 2020-11-09
HHH-14313 - natural-id support (6.0)

sebersole:
Support for both simple and compound natural-ids. Includes new features:

- multi-load
- easier compound loading
- event-less listeners


hibernate-core/src/main/java/org/hibernate/loader/ast/internal/CompoundNaturalIdLoader.java
```diff
@@ -0,0 +1,103 @@
...
+			);
+		}
+
+		// make sure we've exhausted all JDBC parameters
+		assert ! jdbcParamItr.hasNext();
+	}
+
+	@Override
+	public Object resolveIdToNaturalId(Object id, SharedSessionContractImplementor session) {
+		return null;
```
beikov, line 96: Shouldn't this throw a NotYetImplementedIn6Exception?

beikov (commented):
LGTM. Just a few small comments.

hibernate-core/src/main/java/org/hibernate/loader/ast/LoaderLogging.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+package org.hibernate.loader.ast;
+
+import org.jboss.logging.Logger;
+
+/**
+ * @author Steve Ebersole
+ */
+public interface LoaderLogging {
+	String LOGGER_NAME = "";
```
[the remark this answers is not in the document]
sebersole: lol, yep

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/CompoundNaturalIdLoader.java
sebersole: No, I should just implement it ;)

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/MultiNaturalIdLoaderImpl.java
```diff
@@ -0,0 +1,223 @@
...
+					entityDescriptor.getNaturalIdMapping().getJdbcTypeCount( sessionFactory.getTypeConfiguration() ),
+					numberOfIdsLeft
+			);
+		}
+
+		int idPosition = 0;
+		while ( numberOfIdsLeft > 0 ) {
+			final int batchSize = Math.min( numberOfIdsLeft, maxBatchSize );
+
+			final Object[] naturalIdsInBatch = new Object[ batchSize ];
```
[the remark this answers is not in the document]
sebersole: The statement is dependent upon the number of keys to load.  So yeah we could, but that would require quite a few more changes here.  But probably worth it.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/MultiNaturalIdLoaderImpl.java
```diff
@@ -0,0 +1,223 @@
...
+				entityDescriptor.getNaturalIdMapping(),
+				null,
+				naturalIdsInBatch.length,
+				session.getLoadQueryInfluencers(),
+				lockOptions,
+				jdbcParameters::add,
+				sessionFactory
+		);
+
+		final JdbcSelect jdbcSelect = sqlAstTranslatorFactory.buildSelectTranslator( sessionFactory ).translate( sqlSelect );
```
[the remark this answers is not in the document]
sebersole: See previous

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/MultiNaturalIdLoaderImpl.java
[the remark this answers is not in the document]
beikov: Can't we reuse the same statement by binding `null` for the last few ids to fill up the batch? Since we use `id IN (?, ?)` which is the same semantically as `id = ? or id = ?` this should work well.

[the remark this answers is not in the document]
sebersole: It really makes no difference whether we use IN or OR... the "problem" is the same.  I do not like the "padded" solution.  Was actually thinking more what I do with dynamic batching.

Working on this change...

beikov (approved):

hibernate-core/src/main/java/org/hibernate/NaturalIdMultiLoadAccess.java
```diff
@@ -0,0 +1,130 @@
...
+	 *
+	 * @return {@code this}, for method chaining
+	 */
+	NaturalIdMultiLoadAccess<T> with(CacheMode cacheMode);
+
+	/**
+	 * Define a load graph to be used when retrieving the entity
+	 */
+	default NaturalIdMultiLoadAccess<T> with(RootGraph<T> graph) {
+		return with( graph, GraphSemantic.LOAD );
```
NathanQingyangXu, line 43: Any reason to choose `LOAD` as the default entity graph type? Seems `FETCH` is the default type as per JPA spec.

hibernate-core/src/main/java/org/hibernate/internal/MultiIdentifierLoadAccessImpl.java
```diff
@@ -0,0 +1,170 @@
...
+				// change it back
+				session.setCacheMode( sessionCacheMode );
+			}
+		}
+	}
+
+	@Override
+	@SuppressWarnings( "unchecked" )
+	public <K> List<T> multiLoad(List<K> ids) {
+		return perform( () -> entityPersister.multiLoad( ids.toArray( new Object[ 0 ] ), session, this ) );
```
NathanQingyangXu, line 168: You might well reuse the `ArrayHelper.EMPTY_OBJECT_ARRAY` (which was introduced recently by https://github.com/hibernate/hibernate-orm/pull/3493) to improve performance a little bit.

hibernate-core/src/main/java/org/hibernate/internal/util/collections/CollectionHelper.java
```diff
@@ -361,4 +362,27 @@ public static Properties asProperties(Map<?,?> map) {
...
+	@SuppressWarnings( "unchecked" )
+	public static <K,S> Map<K, S> asMap(Object[] elements) {
+		assert elements != null;
+		assert elements.length % 2 == 0;
+
+		final HashMap<K,S> map = new HashMap<>();
+		collectMapEntries( map::put, elements );
+		for ( int i = 0; i < elements.length; i += 2 ) {
+			map.put( (K) elements[ i ], (S) elements[ i+1 ] );
+		}
```
NathanQingyangXu, line 384: Do we put the elements twice? Once is in `collectMapEntries( map::put, elements )` and then we insert them again afterwards?

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/AbstractNaturalIdLoader.java
```diff
@@ -141,18 +144,33 @@ public Callback getCallback() {
...
+		postLoadListener.completedLoadByNaturalId( entityDescriptor, result, session );
+
+		return result;
 	}
 
+	protected abstract Object resolveNaturalIdBindValue(Object naturalIdToLoad, SharedSessionContractImplementor session);
+
+	protected abstract void applyNaturalIdAsJdbcParameters(
+			Object naturalIdToLoad,
+			List<JdbcParameter> jdbcParameters,
```
NathanQingyangXu, line 157: In theory `List<? extends JdbcParameter>` is more flexible, unless we are sure JdbcParameter could never have child class.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/CompoundNaturalIdLoader.java
```diff
@@ -0,0 +1,98 @@
...
+			EntityMappingType entityDescriptor,
+			MappingModelCreationProcess creationProcess) {
+		super( naturalIdMapping, preLoadListener, postLoadListener, entityDescriptor, creationProcess );
+	}
+
+	@Override
+	protected Object resolveNaturalIdBindValue(Object naturalIdValue, SharedSessionContractImplementor session) {
+		// the "real" form as an array, although we also accept here a Map and reduce it to
+		// the appropriately ordered array
+		if ( naturalIdValue instanceof Object[] ) {
```
NathanQingyangXu, line 43: If `naturalIdValue` is of primitive type array, the above statement would return false. If that were not the original intention, we can use the following statement instead:
```
if ( naturalIdValue.getClass().isArray() ) {
... ...
}
```

hibernate-core/src/main/java/org/hibernate/internal/util/collections/CollectionHelper.java
```diff
@@ -361,4 +362,27 @@ public static Properties asProperties(Map<?,?> map) {
...
+			mapEntryConsumer.accept( (K) mappings[i], (V) mappings[i+1] );
+		}
+	}
+
+	@SuppressWarnings( "unchecked" )
+	public static <K,S> Map<K, S> asMap(Object[] elements) {
+		assert elements != null;
+		assert elements.length % 2 == 0;
+
+		final HashMap<K,S> map = new HashMap<>();
```
NathanQingyangXu, line 380: Seems we could reuse existing util method in the same class to avoid Map rehashing:
```
final HashMap<K, S> map = CollectionHelper.mapOfSize( elements.length / 2 );
```

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/CompoundNaturalIdLoader.java
```diff
@@ -0,0 +1,98 @@
...
+			List<JdbcParameter> jdbcParameters,
+			JdbcParameterBindings jdbcParamBindings,
+			SharedSessionContractImplementor session) {
+		assert naturalIdToLoad instanceof Object[];
+		final Object[] naturalIdValueArray = (Object[]) naturalIdToLoad;
+
+		final Iterator<JdbcParameter> jdbcParamItr = jdbcParameters.iterator();
+
+		for ( int i = 0; i < naturalIdMapping().getNaturalIdAttributes().size(); i++ ) {
+			final SingularAttributeMapping attrMapping = naturalIdMapping().getNaturalIdAttributes().get( i );
```
NathanQingyangXu, line 74: Is that true that we don't need to avoid repeated invoking "naturalIdMapping().getNaturalIdAttributes()" simply because it is too light-weight? Why not ensure we only invoke it only once as:
```
for ( SingularAttributeMapping attrMapping : naturalIdMapping().getNaturalIdAttributes()) {

```

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -224,32 +251,47 @@ private LoaderSelectBuilder(
...
 		}
-		else {
-			for ( ModelPart restrictedPart : restrictedParts ) {
-				if ( restrictedPart instanceof ForeignKeyDescriptor ) {
-					forceIdentifierSelection = true;
-				}
+
+		if ( restrictedParts.size() == 1 ) {
+			final ModelPart restrictedPart = restrictedParts.get( 0 );
+			if ( Objects.equals( restrictedPart.getPartName(), NaturalIdMapping.PART_NAME ) ) {
```
NathanQingyangXu, line 269: Given `NaturalIdMapping.PART_NAME` is not null, it seems we can replace the above with
```
if ( NaturalIdMapping.PART_NAME.equals( restrictedPart.getPartName() ) {
```
to save some unnecessary null checking in `Objects.equals()`.

hibernate-core/src/main/java/org/hibernate/internal/MultiIdentifierLoadAccessImpl.java
NathanQingyangXu: It seems we often need to invoke the #multiLoad() with Java generic collection rather than an array, maybe we can add another happy path to avoid such ugly "list -> aray" low level processing. For one thing, reference array is a deprecated usage in JDK in terms of both performance and code readability. It would be great to see the array usage as little as possible (only for primitive value array) rather than adhering to the outdated pattern.

hibernate-core/src/main/java/org/hibernate/internal/util/collections/CollectionHelper.java
```diff
@@ -361,4 +362,27 @@ public static Properties asProperties(Map<?,?> map) {
...
+		// even numbered
+		assert mappings.length %2 == 0;
+
+		for ( int i = 0; i < mappings.length; i += 2 ) {
+			mapEntryConsumer.accept( (K) mappings[i], (V) mappings[i+1] );
+		}
+	}
+
+	@SuppressWarnings( "unchecked" )
+	public static <K,S> Map<K, S> asMap(Object[] elements) {
```
NathanQingyangXu, line 376: It would be great to provide method accepting Collections in CollectionHelper's method (in favour of array counterpart).

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -207,6 +209,31 @@ public static SelectStatement createSubSelectFetchSelect(
 	private int fetchDepth;
 	private Map<OrderByFragment, TableGroup> orderByFragments;
 
+	private LoaderSelectBuilder(
+			SqlAstCreationContext creationContext,
+			Loadable loadable,
+			List<? extends ModelPart> partsToSelect,
+			List<ModelPart> restrictedParts,
```
NathanQingyangXu, line 216: Why there is inconsistency above for the two list (one uses bounded generic type whereas the other doesn't)? That would confuse the future code reader always.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/MultiNaturalIdLoaderStandard.java
```diff
@@ -0,0 +1,108 @@
...
+import org.hibernate.metamodel.mapping.EntityMappingType;
+import org.hibernate.metamodel.mapping.internal.MappingModelCreationProcess;
+import org.hibernate.sql.ast.tree.expression.JdbcParameter;
+import org.hibernate.sql.ast.tree.select.SelectStatement;
+import org.hibernate.sql.results.LoadingLogger;
+
+/**
+ * Standard MultiNaturalIdLoader implementation
+ */
+public class MultiNaturalIdLoaderStandard<E> implements MultiNaturalIdLoader<E> {
```
NathanQingyangXu, line 29: I would suggest append "Impl" to the class name. Some class with name ending with "Standard" implies some interface like JPA spec classes.

NathanQingyangXu: Now I got it. You wanna avoid duplicating with an existing class in the same package.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/SimpleNaturalIdLoader.java
```diff
@@ -0,0 +1,262 @@
...
+						return QueryParameterBindings.NO_PARAM_BINDINGS;
+					}
+
+					@Override
+					public Callback getCallback() {
+						return afterLoadAction -> {
+						};
+					}
+				},
+				row -> row,
```
NathanQingyangXu, line 154: We might well use `Function.identity()` above. Super minor.

hibernate-core/src/main/java/org/hibernate/loader/entity/CacheEntityLoaderHelper.java
```diff
@@ -62,6 +62,46 @@
...
+			if ( options.isAllowNulls() ) {
+				final EntityPersister persister = session.getFactory()
+						.getRuntimeMetamodels()
+						.getMappingMetamodel()
+						.getEntityDescriptor( keyToLoad.getEntityName() );
+				if ( ! persister.isInstance( old ) ) {
+					LOG.debugf(
+							"Load request found matching entity in context, but the matched entity was of an inconsistent return type.  " +
+									"Setting status as `%s`",
+							EntityStatus.INCONSISTENT_RTN_CLASS_MARKER
```
NathanQingyangXu, line 93: We might well add log level checking here for the log message involves string interpolation, so we are gonna pay the cost regardless of whether debug level is enabled or not.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/CompoundNaturalIdMapping.java
```diff
@@ -0,0 +1,216 @@
...
+				declaringType,
+				creationProcess
+		);
+		multiLoader = new MultiNaturalIdLoaderStandard<>( declaringType, creationProcess );
+	}
+
+	@Override
+	@SuppressWarnings( "rawtypes" )
+	public Object normalizeValue(Object incoming, SharedSessionContractImplementor session) {
+		if ( incoming instanceof Object[] ) {
```
NathanQingyangXu, line 78: Is that true that natural ids cannot be primitive values?

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/CompoundNaturalIdMapping.java
```diff
@@ -0,0 +1,216 @@
...
+
+	public CompoundNaturalIdMapping(
+			EntityMappingType declaringType,
+			List<SingularAttributeMapping> attributes,
+			String cacheRegionName,
+			MappingModelCreationProcess creationProcess) {
+		super( declaringType, cacheRegionName );
+		this.attributes = attributes;
+
+		final List<JdbcMapping> jdbcMappings = new ArrayList<>();
```
NathanQingyangXu, line 58: Could we make an educated guess about the list size based on attributes? Seems most of the time it is predicatable.

hibernate-core/src/test/java/org/hibernate/orm/test/metamodel/mapping/naturalid/CompoundNaturalIdTests.java
```diff
@@ -0,0 +1,266 @@
...
+		assert values.length % 2 == 0;
+
+		final HashMap<String,String> valuesMap = new HashMap<>();
+		for ( int i = 0; i < values.length; i += 2 ) {
+			valuesMap.put( values[i], values[i+1] );
+		}
+
+		return valuesMap;
+	}
+
```
NathanQingyangXu, line 59: could we replace the above with the usage of the new CollectionHelper.asMap()?

hibernate-core/src/main/java/org/hibernate/NaturalIdMultiLoadAccess.java
```diff
@@ -0,0 +1,130 @@
...
+	}
```
NathanQingyangXu, line 44: Screenshot from 2021-01-23 10-11-42
It seems the default entity graph corresponds to 'FETCH' semantics. It is also the case in Spring Data JPA.

hibernate-core/src/main/java/org/hibernate/NaturalIdMultiLoadAccess.java
```diff
@@ -0,0 +1,130 @@
...
+	 * Perform a load of multiple entities by natural-id.
+	 *
+	 * See {@link #enableOrderedReturn} and {@link #enableReturnOfDeletedEntities}
+	 * for options which effect the size and "shape" of the return list.
+	 *
+	 * @param ids The natural-id values to load
+	 *
+	 * @return The managed entities.
+	 */
+	List<T> multiLoad(Object... ids);
```
NathanQingyangXu, line 108: So the above method implies that primitive array (like `int[]` or `long[]`) cannot be passed in directly. Luckily, we have a util method (`toObjectArray(Object array)`) created by me in ArrayHelper for the transformation.
It might be worthwhile considering providing overloaded methods accepting primitive type array, if for completeness sake.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/CompoundNaturalIdLoader.java
NathanQingyangXu: But I guess this PR assumed that primitive nutural id array has gone through the transformation to Object[] already.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/CompoundNaturalIdLoader.java
NathanQingyangXu: At least it hurts code readability, IMHO.

hibernate-core/src/main/java/org/hibernate/internal/util/collections/CollectionHelper.java
```diff
@@ -361,4 +362,27 @@ public static Properties asProperties(Map<?,?> map) {
 		}
 	}
 
+	@SuppressWarnings( "unchecked" )
+	public static <K,V> void collectMapEntries(BiConsumer<K, V> mapEntryConsumer, Object[] mappings) {
+		// even numbered
+		assert mappings.length %2 == 0;
```
NathanQingyangXu, line 368: super minor. would be great to leave a space char after `%`

hibernate-core/src/main/java/org/hibernate/internal/MultiIdentifierLoadAccessImpl.java
NathanQingyangXu: Out of pure curiosity, I did some JMH experiments as following:

1. a big List of String has been created in @Setup method;
2. I tested the following two dimensions:  a) whether to transform to Array first; b) whether to user iterator or index forloop

Below is the interesting results:

List + iterator: 
Benchmark              Mode  Cnt  Score   Error  Units
Benchmark.iterateList  avgt    5  0.075 ± 0.009  ms/op

List + forloop:
Benchmark              Mode  Cnt  Score   Error  Units
Benchmark.iterateList  avgt    5  0.068 ± 0.003  ms/op

Array + iterator:
Benchmark              Mode  Cnt  Score   Error  Units
Benchmark.iterateList  avgt    5  0.074 ± 0.002  ms/op

Array + forloop:
Benchmark              Mode  Cnt  Score   Error  Units
Benchmark.iterateList  avgt    5  0.074 ± 0.003  ms/op

The JDK is openjdk8. I draw some primitive conclusion as following:

- there is not much performance difference to iterate array or index forloop;
- index forloop is 10% faster than iterator in case of List;
- there is not much performance difference between List or Array as long as List usage is restricted to iterator

So suppose `list` is a List of String, the fastest way to iterate its elements is as following (10% faster than all the other three possibilities):
```
for (int i = 0; i < list.size(); i++) {
    String elem = list.get(i);
    // do something with elem
}
```

I also compared the following three ways to transform a List into an Array (in JDK8):

- list.toArray(new String[0])
- list.toArray(new String[list.size()])
- list.stream().toArray(String[]::new)

The first approach is significantly faster than the other two; whereas there is not much difference between the other two.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/CompoundNaturalIdLoader.java
sebersole: I believe we discussed this one before... the general case... enhanced-for-loops are not efficient.  `#forEach` *might be* efficient (depends on a few things).

hibernate-core/src/main/java/org/hibernate/NaturalIdMultiLoadAccess.java
sebersole: Not sure what you mean.  The JPA spec says nothing about loading by natural-id, so how can either be "per spec"?

NathanQingyangXu:
I meant default entity graph semantic is "FETCH", so maybe it is the case
here as well. Definitively natural id loading has no specific spec, but
maybe generic convention allies here as well?

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/CompoundNaturalIdLoader.java
NathanQingyangXu: Sorry for the confusion. My point here is the duplication avoidance. Obviously, Java compiler won't be smart enough to change:
```
for ( int i = 0; i < naturalIdMapping().getNaturalIdAttributes().size(); i++ ) {
			final SingularAttributeMapping attrMapping = naturalIdMapping().getNaturalIdAttributes().get( i );

```
into more efficient
```
var naturalIdAttrs = naturalIdMapping().getNaturalIdAttributes();
for ( int i = 0; i < naturalIdAttrs.size(); i++ ) {
			final SingularAttributeMapping attrMapping = naturalIdAttrs.get( i );

```
for the following reasons:

1. the loop body might change the natural id attributes collection;
2. naturalIdMapping().getNaturalIdAttributes() might have side effect, for instance, the common lazy initialization pattern.
The avoidance of duplication leads to better code readability also, I believe.

sebersole:
Was integrated upstream

PR #3635, opened 2020-11-11
HHH-14257  An Entity A with a map collection having as index an Embeddable with a an association to the Entity A fails with a NPE

dreab8:
https://hibernate.atlassian.net/browse/HHH-14257


Sanne:
this is great, thanks! Looks much better than the previous soluion, I'm glad you tried again.

PR #3636, opened 2020-11-11
HHH-14322 Fix HBM many-to-one property ref support

beikov:
https://hibernate.atlassian.net/browse/HHH-14322


Sanne:
Nice ! merged and backporting to 5.4

PR #3641, opened 2020-11-12
HHH-14276 Test and fix for issues during composite identifier column reference lookup due to quoting

beikov:
https://hibernate.atlassian.net/browse/HHH-14276


Sanne:
Many thanks, I merged it - but added an additional commit to fix some style and line ending issues.

Backporting next...

kawoolutions:
Before you go, could you please check https://hibernate.atlassian.net/browse/HHH-14343 before backporting?

PR #3643, opened 2020-11-12
HHH-14325 - Add Query hint for specifying "query spaces" for native q…

sebersole:
…ueries

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/jpa/QueryHints.java
+++ b/hibernate-core/src/main/java/org/hibernate/jpa/QueryHints.java
@@ -19,6 +19,7 @@ import static org.hibernate.annotations.QueryHints.FETCH_SIZE;
 import static org.hibernate.annotations.QueryHints.FLUSH_MODE;
 import static org.hibernate.annotations.QueryHints.FOLLOW_ON_LOCKING;
 import static org.hibernate.annotations.QueryHints.NATIVE_LOCKMODE;
+import static org.hibernate.annotations.QueryHints.NATIVE_SPACES;
 import static org.hibernate.annotations.QueryHints.PASS_DISTINCT_THROUGH;
 import static org.hibernate.annotations.QueryHints.READ_ONLY;
 import static org.hibernate.annotations.QueryHints.TIMEOUT_HIBERNATE;
@@ -26,8 +27,6 @@ import static org.hibernate.annotations.QueryHints.TIMEOUT_JPA;
 
 /**
  * Defines the supported JPA query hints
- *
- * @author Steve Ebersole
  */
 public class QueryHints {
 	/**
@@ -91,30 +90,27 @@ public class QueryHints {
 	 * 
 	 * Note: Currently, attributes that are not specified are treated as FetchType.LAZY or FetchType.EAGER depending
 	 * on the attribute's definition in metadata, rather than forcing FetchType.LAZY.
-	 *
-	 * @deprecated (since 5.4) Use {@link GraphSemantic#FETCH}'s {@link GraphSemantic#getJpaHintName()} instead
 	 */
-	@Deprecated
 	public static final String HINT_FETCHGRAPH = GraphSemantic.FETCH.getJpaHintName();
 	
 	/**
 	 * Hint providing a "loadgraph" EntityGraph.  Attributes explicitly specified as AttributeNodes are treated as
 	 * FetchType.EAGER (via join fetch or subsequent select).  Attributes that are not specified are treated as
 	 * FetchType.LAZY or FetchType.EAGER depending on the attribute's definition in metadata
-	 *
-	 * @deprecated (since 5.4) Use {@link GraphSemantic#LOAD}'s {@link GraphSemantic#getJpaHintName()} instead
 	 */
-	@Deprecated
 	public static final String HINT_LOADGRAPH = GraphSemantic.LOAD.getJpaHintName();
 
 	public static final String HINT_FOLLOW_ON_LOCKING = FOLLOW_ON_LOCKING;
 
 	public static final String HINT_PASS_DISTINCT_THROUGH = PASS_DISTINCT_THROUGH;
 
+	public static final String HINT_NATIVE_SPACES = NATIVE_SPACES;
+
+
 	private static final Set<String> HINTS = buildHintsSet();
 
 	private static Set<String> buildHintsSet() {
-		HashSet<String> hints = new HashSet<String>();
+		HashSet<String> hints = new HashSet<>();
 		hints.add( HINT_TIMEOUT );
 		hints.add( SPEC_HINT_TIMEOUT );
 		hints.add( HINT_COMMENT );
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/query/internal/NativeQueryImpl.java
```diff
@@ -559,6 +567,33 @@ protected boolean isNativeQuery() {
...
 	}
 
+	@Override
+	protected boolean applyQuerySpaces(Object value) {
+		if ( value == null ) {
+			return false;
+		}
+
+		if ( value instanceof List ) {
+			this.querySpaces.addAll( (List) value );
```
Sanne, line 577: I suspect `querySpaces` might not be initialized here. All other writing methods check for it, and initialize it with `new ArrayList<>();`.

Secondarily, but why is it using a list and not a Set? Is a List really preferreable here?

hibernate-core/src/main/java/org/hibernate/jpa/QueryHints.java
```diff
@@ -91,30 +90,27 @@
 	 * 
 	 * Note: Currently, attributes that are not specified are treated as FetchType.LAZY or FetchType.EAGER depending
 	 * on the attribute's definition in metadata, rather than forcing FetchType.LAZY.
-	 *
-	 * @deprecated (since 5.4) Use {@link GraphSemantic#FETCH}'s {@link GraphSemantic#getJpaHintName()} instead
```
Sanne, line 95: Are the "deprecated" removals intentional?

Sanne (changes requested):

gavinking:
@sebersole what is the use case for this? stored procedures and views? or something else?

gavinking:
(I'm wondering because perhaps HR needs a way to specify it since we don't support generic query hints.)

beikov:
Currently, when executing a native query, we flushe all pending changes. With query spaces, we can defer flushes for other query spaces.

Sanne:
@gavinking we had a request for this in the context of native SQL queries. The users were making changes in unrelated (unmapped) tables, and still ORM would invalidate the query cache.

sebersole:
> @sebersole what is the use case for this? stored procedures and views? or something else?

"JPA portability".  It performs the same as the `SynchronizeableQuery` but without casting to Hibernate API.

hibernate-core/src/main/java/org/hibernate/jpa/QueryHints.java
sebersole: Yes.

hibernate-core/src/main/java/org/hibernate/query/internal/NativeQueryImpl.java
sebersole: I chose List because it is more conceptually similar to array.  We can certainly accept both.

Checking for null was just an oversight

sebersole:
> @gavinking we had a request for this in the context of native SQL updates. The users were making changes in unrelated (unmapped) tables, and still ORM would invalidate the query cache.

That's only a (very small ) part of it.  They had plenty of other queries against the mapped domain as well.  The concern was casting rather than "pure" JPA usage.

hibernate-core/src/main/java/org/hibernate/query/internal/NativeQueryImpl.java
Sanne: > Checking for null was just an oversight

WDYM, don't you think it is necessary here?

sebersole: Or do you mean for the field?  For the field, yes, that's actually not at all new and its done that way because it more closely matches array, which is what the code that uses this actually expects ultimately

sebersole: Oversight means it was overlooked.  I just missed it

Sanne: np :) I'm just wondering if you're planning to fix that.

No need to change the List/Set - that question was an unrelated curiosity, I was just wondering

gavinking:
Wait, let me back up a little and see if I understand exactly what we already supported here:

- When you have a native query that returns entities, the entity mapped tables are automatically added to the query spaces, but
- not any other table that the native query touches.

And there was no way to add a query space via `@NamedNativeQuery` (this half-surprised me).

Is that right?

gavinking:
> "JPA portability". It performs the same as the `SynchronizeableQuery` but without casting to Hibernate API.

Ah, OK, this was the first time I noticed that interface! So you can do it there. (But not via an annotation, correct?)

Then we should definitely expose `addSynchronizedQuerySpace()` in HR. This was an oversight!

sebersole:
> > "JPA portability". It performs the same as the `SynchronizeableQuery` but without casting to Hibernate API.
> 
> Ah, OK, this was the first time I noticed that interface! So you can do it there. (But not via an annotation, correct?)
> 
> Then we should definitely expose `addSynchronizedQuerySpace()` in HR. This was an oversight!

The interface is "new" to a degree - I added it when adding callable support to unify those contracts.  But the methods are quite old

sebersole:
> Wait, let me back up a little and see if I understand exactly what we already supported here:
> 
> * When you have a native query that returns entities, the entity mapped tables are automatically added to the query spaces, but
> * not any other table that the native query touches.

If you supply the resultset-mapping with entity returns, then yes we use that information.  I'd have to double check if we do that when supplying a return-class that happens to be an entity.

When you have a native-query of any kind we do not automatically add any query spaces - we'd have to do a real parse of the SQL to know that.  So a "fresh" NativeQuery will have no query spaces.  

This dove-tails with a JPA "requirement" that NativeQuery perform an auto-flush.  I put it on quotes because they added a TCK test that required this behavior (we used to do no flush for NativeQuery execution).

 
> And there was no way to add a query space via `@NamedNativeQuery` (this half-surprised me).

This surprised me as well now that you mention it and I looked.  Now of course you can via the hint

sebersole:
> > And there was no way to add a query space via `@NamedNativeQuery` (this half-surprised me).
> 
> This surprised me as well now that you mention it and I looked. Now of course you can via the hint

I also went back and added:

```
public @interface NamedNativeQuery {

	/**
	 * The query spaces to apply for the query.
	 * 
	 * @see org.hibernate.SynchronizeableQuery
	 */
	String[] querySpaces();
}
```

gavinking:
> When you have a "raw" native-query we do not automatically add any query spaces - we'd have to do a real parse of the SQL to know that. So a "fresh" NativeQuery will have no query spaces.

Yes, what I thought, and it's quite reasonable, of course.

> they added a TCK test that required this behavior

*Really?* Interesting.

> This surprised me as well now that you mention it and I looked. Now of course you can via the hint

Yeah, but even so it might be worth adding a specific annotation member to do it more comfortably.

dreab8:
apart checking `this.querySpaces` is not null, it looks good to me :+1:

sebersole:
Pushed manually

PR #3644, opened 2020-11-12
HHH-14318 Work on @JoinFormula and @JoinColumnOrFormula support

beikov:
https://hibernate.atlassian.net/browse/HHH-14318

Here a list of changes:

* Support for join column formulas and unified handling of column/formula/read/write expressions
* Introduce abstraction for mappings called `SelectionMapping` and `SelectionMappings` (for embeddables)
* Change `ColumnConsumer` to `SelectionConsumer` that accepts an index and `SelectionMapping`
* Naming changes in Bindable => embrace the forEachXXX naming scheme that returns the amount of elements
   * `visitColumns` => `forEachSelection`
   * `visitJdbcTypes` => `forEachJdbcType`
   * `visitJdbcValues` => `forEachJdbcValue`
   * `visitDisassembledJdbcValues` => `forEachDisassembledJdbcValue`
* For indexed consumers, introduce method variant forEachXXX that accepts an offset which is added to index for consumer


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EmbeddedForeignKeyDescriptor.java
```diff
@@ -181,11 +186,11 @@ public DomainResult createDomainResult(
 							),
 							s ->
 									new ColumnReference(
-											identificationVariable,
```
sebersole, line 184: I'm tempted to say that `ColumnReference` should accept the `ColumnMapping` rather than the individual pieces.  I understand you probably chose that because it matches the existing contracts.  

But this is point I brought up in regards to model parts accepting a `ColumnConsumer`, but that really a `ColumnMappingConsumer` makes sense as well.  There is never a time when a model part wont have the  `ColumnMapping` reference(s) already, so concerns about instantiations isn't really a hinderance there.

I'm not sure there is ever a time when we would not want a `ColumnMapping` on the `ColumnReference` for some reason ("PK tables", e.g.), but if so we could e.g. add a `ModelColumnReference` also that takes the ColumnMapping.

I just think we added this `ColumnMapping` contract because it cleaned up a lot of code.  It would be a shame to not use it in other places where we could be using it.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EmbeddedForeignKeyDescriptor.java
```diff
@@ -255,46 +260,49 @@ public Predicate generateJoinPredicate(
...
-							false,
-							null,
-							null,
-							jdbcMapping,
-							creationContext.getSessionFactory()
-					)
-			);
-			predicate.add( comparisonPredicate );
-		}
+		lhsMappings.forEachColumn(
```
sebersole, line 278: The more I see it the more I really like your naming pattern here better - `#forEachColumn` as opposed to `#visitColumns`.  It better matches with the Java naming.  

I'm actually thinking of a general refactoring to make that change.  What do y'all think?

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EmbeddedForeignKeyDescriptor.java
```diff
@@ -115,35 +116,39 @@ public DomainResult createCollectionFetchDomainResult(
...
-			for ( int i = 0; i < keyColumnExpressions.size(); i++ ) {
-				final JdbcMapping jdbcMapping = jdbcMappings.get( i );
-				final String columnExpression = targetColumnExpressions.get( i );
-				final SqlSelection sqlSelection = sqlExpressionResolver.resolveSqlSelection(
-						sqlExpressionResolver.resolveSqlExpression(
-								SqlExpressionResolver.createColumnReferenceKey(
-										tableReference,
-										columnExpression
+			List<SqlSelection> sqlSelections = new ArrayList<>( targetColumnMappings.getColumnSpan() );
+			targetColumnMappings.forEachColumn(
```
sebersole, line 120: Another case where this code would be a lot cleaner passing along the `ColumnMapping`

sebersole (commented):
I love how `ColumnMapping` cleans up the code!  Great job!  

In fact my biggest question across many of the in-line comments is whether we should use it in even more places.

There are also some code style comments.  The recurring one about multi-line statements and how to chop them down.  I'm making a PR on top of yours changing these to use the style we try to use in ORM.  Being consistent helps us quickly look at code and not have to get caught up on extraneous things like formatting.

hibernate-core/src/main/java/org/hibernate/boot/model/TypeDefinition.java
```diff
@@ -50,7 +51,7 @@
  */
 @SuppressWarnings({"rawtypes", "unchecked"})
 public class TypeDefinition implements Serializable {
-	public static final MutableInteger NAME_COUNTER = new MutableInteger();
+	public static final AtomicInteger NAME_COUNTER = new AtomicInteger();
```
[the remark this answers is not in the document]
beikov: Since this is static and thus accessible by multiple threads, this must use a thread safe data structure like AtomicInteger.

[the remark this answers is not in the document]
sebersole: Ah, missed the static.. carry on.

PR #3645, opened 2020-11-13
HHH-14329 Test and fix for DirtinessTracker usage for enhanced entities doesn't respect mutable types

beikov:
https://hibernate.atlassian.net/browse/HHH-14329

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
@@ -12,6 +12,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.BitSet;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -67,7 +68,6 @@ import org.hibernate.engine.jdbc.spi.JdbcCoordinator;
 import org.hibernate.engine.jdbc.spi.JdbcServices;
 import org.hibernate.engine.spi.CachedNaturalIdValueSource;
 import org.hibernate.engine.spi.CascadeStyle;
-import org.hibernate.engine.spi.CascadingAction;
 import org.hibernate.engine.spi.CascadingActions;
 import org.hibernate.engine.spi.CollectionKey;
 import org.hibernate.engine.spi.EntityEntry;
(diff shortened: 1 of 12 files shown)


gavinking:
This change needs to be replicated in HR.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -2261,6 +2259,44 @@ public boolean isDefinedOnSubclass(int i) {
 		return Arrays.copyOf( fields, counter );
 	}
 
+	@Override
+	public int[] resolveDirtyAttributeIndexes(String[] attributeNames) {
+		final int[] mutablePropertiesIndexes = entityMetamodel.getMutablePropertiesIndexes();
```
Sanne, line 2264: This looks a bit scary, but after reading through it it seems all it does is a union Set of the indexes of mutable properties + the other named attributes.
Did I read that correctly? Maybe we could add an implementation comment?

hibernate-core/src/main/java/org/hibernate/tuple/entity/EntityMetamodel.java
```diff
@@ -97,7 +97,7 @@
 
 	private final Map<String, Integer> propertyIndexes = new HashMap<>();
 	private final boolean hasCollections;
-	private final boolean hasMutableProperties;
+	private final int[] mutablePropertiesIndexes;
```
Sanne, line 100: You think we could use BitSet(s) to track such things?  Probably not worth our time here, but I'd like to take that in consideration in ORM 6.

beikov: Yeah, a `BitSet` is a good candidate for something like that.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
beikov: I will adapt the implementation a bit as it is not 100% correct yet. Working on it atm.

Sanne:
excellent, merging and backporting. Thanks @beikov !

gavinking:
So, @Sanne, @beikov I have a couple of doubts about this, relating to https://github.com/hibernate/hibernate-reactive/issues/453.

1. If I'm understanding correctly, there's a new call to `this.collection.size()` that seems to sometimes cause actual fetching the collection. That doesn't feel right to me. Do we really need to fetch an unfetched collection to check it's field-level dirtiness? Should that invocation of `size()` be protected by a call to `isInitialized()` or something?
2. Also, why should this be done at all for the `mappedBy` side of an association? A `mappedBy` collection is never "dirty".

Granted I don't understand what's going on here especially well, but it doesn't seem right that this change should have started causing collections to get fetched that weren't being fetched before.

Sanne:
Hi @gavinking (I'm back from a few days off..), good catch - I didn't notice it, and agree we shouldn't initialize a collection in this case.

@beikov I assume this wasn't intentional, or are we missing something? Sorry for letting you play wack-a-mole BTW, it seems this area is very lacking of integration tests.

beikov:
Yeah this can be considered a bug. What I don't understand is, why this wasn't a problem before as I didn't touch this generated code at all. I'll look into it.

beikov:
I created https://hibernate.atlassian.net/browse/HHH-14348 for this

gavinking:
Cheers, thanks guys!

PR #3646, opened 2020-11-13
HHH-14325 - Add Query hint for specifying "query spaces" for native queries (5.3)

sebersole:
Cherry-pick of HHH-14325 work back to 5.3


sebersole:
Pushed manually

PR #3647, opened 2020-11-17
HHH-13945: catch exception and fallback to call mapToMapFromEntity directly

showuon:
Fix: https://hibernate.atlassian.net/browse/HHH-13945

In HHH-13760, we tried to fix ClassCastException when Envers inserts audit rows that use lazy many-to-one mappings, but we didn't consider that there might be cases that the proxy object cannot initialized successfully. In our project, it's because the deleted object is an inherited object, and while hibernate tried to initialize the same object with parent class name, it will be the state that (1) not initialized (2) cannot load from session cache and 2nd cache (3) not existed in Data store, so the objectNotFoundException will be thrown.

In the bug discussion thread, there are other cases caused by the getImplementation() method. Add a fallback mechanism, so that we can try to eagerly initialize the proxy object, and if failed, call the mapToMapFromEntity directly as before.


showuon:
@Naros @NathanQingyangXu , please help review. Thanks.

hibernate-envers/src/main/java/org/hibernate/envers/internal/entities/mapper/relation/ToOneIdMapper.java
```diff
@@ -58,8 +60,16 @@ public boolean mapToMapFromEntity(
...
+
+		// fix HHH-13760 - try to aggressively un-proxy this entity to help get the correct type of data later
+		// in mapToMapFromEntity. But it might fail while getImplementation() if object is deleted or other reasons.
+		// We catch the exception and fallback to call mapToMapFromEntity directly with the HibernateProxy entity
 		if ( lazyMapping && entity instanceof HibernateProxy ) {
-			entity = ( (HibernateProxy) entity ).getHibernateLazyInitializer().getImplementation();
+			try {
+				entity = ((HibernateProxy) entity).getHibernateLazyInitializer().getImplementation();
+			} catch ( PersistenceException e ) {
+				// Ignore the exception and fallback to call mapToMapFromEntity directly
```
beikov, line 71: IMO the exception should at least be logged with a DEBUG level. Since this throws a `PersistenceException` I assume that this will cause a transaction rollback due to the semantics of JPA, so this might not be the best way to fix this. Can you verify if the TX is marked as rollback only here?

beikov (changes requested):
Other than that, it looks good to me.

hibernate-envers/src/main/java/org/hibernate/envers/internal/entities/mapper/relation/ToOneIdMapper.java
beikov: Ok, I just checked the code and the TX is not marked as rollback-only in this case, so this is fine.

If you introduce the logging, I will merge this.

showuon:
@beikov  , thanks for the comments. I've added debug log during exception thrown in this commit: https://github.com/hibernate/hibernate-orm/pull/3647/commits/c0c3324ea602eee59adf06fdf53dd0b1d76f471d. Thank you.

showuon:
All tests passed. Thanks.

showuon:
@beikov , could this PR merged into 5.4 release? Our project have been blocked by this issue for a long time. Thanks.

beikov:
I pushed it to 5.4 as well.

PR #3650, opened 2020-11-17
Add package for Micrometer MeterBinder support

ebullient:
After discussion with @Sanne , create a project that brings support for Micrometer MeterBinders closer to hibernate source.
This first pass brings the 1.6.1 MeterBinder implementations over to work against hibernate 5.4.

Note this is a draft, one of many things to be sorted is how best to test this so we know it is doing what we want it to.


sebersole (approved):
Looks great to me.  In fact, I'm going to apply this for certain to our 6.0 work.

Thanks Erin!

Sanne:
Looks nice, thanks Erin! As soon as you have some tests we'll merge it.

ebullient:
Ok. Tests appended. PLEASE TELL ME IF I DID THIS WRONG. ;) 
I am not testing for output, only for expected construction. I did stumble on one thing I think I have to fix in the existing micrometer MeterBinder in the meanwhile. The changes around metrics for second level caching are not as clear as they could be.. so please look at what I did and ask questions. I have no confidence that what I'm doing matches what was intended (I only know that it at least behaves).

sebersole:
Let's not say "wrong" ;)

Ideally new tests would use annotations rather than the legacy hbm.xml format.  I'll work on that part and create a pull request for you.  Funny enough I literally just converted these this entity on our 6.0 branch so it'll be super easy :D

It would also be nice to not force push to the PR branch.  Instead just push the individual commits and we can combine them before we merge it upstream.  It helps us to see what changed between - its easier to not re-review parts we've already reviewed.

I'll look at the L2 cache stats stuff.  Anything in particular?

sebersole:
Also, this really needs a Jira... it is not a trivial change.

I went ahead and created one: https://hibernate.atlassian.net/browse/HHH-14337

Commits for these issues should being with the Jira key (HHH-14337).  This allows Jira and other tooling to "link" the commits with the Jira issue

sebersole:
I cannot find your repository as a PR target.  Not sure why.  Anyway here are the changes related to the tests: https://github.com/sebersole/hibernate-core/tree/topic/HHH-14337.

Regarding tests, I think we do want some functional tests as well.  E.g. when you save an entity check the number of Sessions opened/closed, number of SQL statements executed, etc.  It vets that your impl is collecting things properly

sebersole:
Sorry for  spamming you so early in the morning :D

If you care about such things, you can create a Jira account and I can assign that issue to you to get "credit" for doing the work

ebullient:
> It would also be nice to not force push to the PR branch. Instead just push the individual commits and we can combine them before we merge it upstream. It helps us to see what changed between - its easier to not re-review parts we've already reviewed.

Habit from other projects. ;) Easy enough to not do here.

> I'll look at the L2 cache stats stuff. Anything in particular?

In 1.6, Micrometer made the change to attempt to collect metrics per region.. the problem is that the Statistics API did not surface a way to get the new/unqualified cache region names, so you have the old(er) method to list names (which includes legacy/qualified names). Then you pass those names to the new methods... but the new methods want unqualified names (so I had to strip the qualifier myself, which as far as I understand it is just the prefix.. but, TBH, I have no idea!). Part of moving this into the hibernate repo is to allow the binder to do _whatever is the right thing_. =)

ebullient:
https://github.com/ebullient/hibernate-orm/tree/5.4 ?!  It isn't private.. :( 
Maybe draft PR status?

No worries about early morning spam; I am glad you're looking!

> Regarding tests, I think we do want some functional tests as well. E.g. when you save an entity check the number of Sessions opened/closed, number of SQL statements executed, etc. It vets that your impl is collecting things properly

I was a little unsure about reproducing functional tests, which is why I didn't. I guessed two... (especially since the second level cache metrics via `getDomainDataRegionStatistics` was a little wonky).

sebersole:
> It isn't private.. :(

Did you fork the repo?  Or just clone it locally directly?  I'm trying to think of things that could lead to this situation, but really not sure.  Not forking is my last idea lol

> In 1.6, Micrometer made the change to attempt to collect metrics per region.. the problem is that the Statistics API did not surface a way to get the new/unqualified cache region names...

Yep, the cache region naming stuff is a mess.  Historical unfortunately.  I plan on revisiting this in 6 and dropping any "legacy concerns".  So it should get better.  And anyway, the work is in the Hibernate repo now so we can adjust that as needed moving forward.

ebullient:
It's a proper fork (the link works? https://github.com/ebullient/hibernate-orm/)

sebersole:
"Forking" is a specific GitHub process.  From the original repo you want to work on, there is an option to "fork it".  There is also an option to simply clone it (which simply pulls the code locally to your machine).  Anyway, its not critical.  You can integrate my commit into your repo using cherry-pick.  Or manually copy the changes.

I almost forked your D&D repo :smiley:

sebersole:
To be clear... we can "see" your repo.  That's not the problem.  But when you create a PR there is a drop-down to select the "base branch" which for my commit would have been the branch on your repo.  However your repo is not in that drop-down even though there are literally hundreds of them. :shrug:

ebullient:
GitHub tells me "I already forked that repo" =)  So I don't feel too crazy. I'll cherry pick your patch, and hope that GH finds its brains sooner or later.

sebersole:
> GitHub tells me "I already forked that repo" =) So I don't feel too crazy. I'll cherry pick your patch, and hope that GH finds its brains sooner or later.

Ok, then I have no idea ;)

Sanne:
Erin's fork does say "forked from hibernate/hibernate-orm" so that's probably not it.

ebullient:
Added some tests for values coming through. I will be honest that the accumulation around the second level cache doesn't make sense to me either...

sebersole:
What do you mean?  You don't think cache put/hit/miss counts are useful?

ebullient:
> What do you mean? You don't think cache put/hit/miss counts are useful?

how it gets to the value of 2, e.g.
```
		Assert.assertEquals( 1, registry.get("hibernate.cache.natural.id.puts").functionCounter().count(), 0);
		Assert.assertEquals(2, registry.get("hibernate.second.level.cache.puts").tags("region", REGION).functionCounter().count(), 0);
```
I just don't know what I'm looking at, so I have no idea if that is expected or not.

sebersole:
Some stats are aggregations of others.  Think of it like a roll-up, though not the fruit kind :)

Here you have an entity which (1) is cached and (2) has a natural-id.  So there will be 2 regions associated with that entity : one for the "data cache" and another for caching natural-id -> pk resolutions.  When this entity is saved, there will be 2 puts for all regions - one per region I just mentioned.

To be honest I'm surprised there is not 3... we'd also put an entry into the "update timestamps cache".

So I'm assuming that your `hibernate.second.level.cache.puts` equates to `Statistics#getSecondLevelCachePutCount`?  That would be the aggregation value I mentioned

ebullient:
Where the values are coming from is all in the PR.. :-P 
Can you go through hibernate-micrometer/src/main/java/org/hibernate/stat/HibernateMetrics.java with a hyper-ridiculous-fine-tooth-comb to make sure the metrics being gathered make sense?

sebersole:
Not sure how fine of a comb it will be, but I can definitely look.

I'm just speaking overall, especially in terms of how the existing impl works

sebersole:
So it is a little hard for me to follow the code you want me to look at closely (HibernateMetrics) because most of it is Micrometer specific and I have no knowledge of Micrometer.  

At high-level, I was surprised looking closer that you don't actually re-implement the Statistics(Implementor) contract.  Was that a conscious decision?  I expected HibernateMetrics to act as the Statistics(Implementor) and consume the events itself.

Beyond that, as I said, its a bit difficult to follow since it is mostly Micrometer calls.

One thing we could discuss is to not expose the "aggregate" counts as part of this integration?

ebullient:
> So it is a little hard for me to follow the code you want me to look at closely (HibernateMetrics) because most of it is Micrometer specific and I have no knowledge of Micrometer.
> 
> At high-level, I was surprised looking closer that you don't actually re-implement the Statistics(Implementor) contract. Was that a conscious decision? I expected HibernateMetrics to act as the Statistics(Implementor) and consume the events itself.

SOOO. For this first pass.. I'm just bringing over what Micrometer already did, which was just tracking the Hibernate Statistics  functions that already exist. I was making the brash assumption that you had no idea they did this, which is also why I wanted you to look.

> Beyond that, as I said, its a bit difficult to follow since it is mostly Micrometer calls.

The micrometer calls to set up function counters do what they say w/o much that is interesting. 
What I am more interested in (and have talked to @Sanne about to some extent) is whether or not the information being gathered is actually useful, and in what context.

> One thing we could discuss is to not expose the "aggregate" counts as part of this integration? It seems to be these aggregates that cause you trouble

And yes, that's the question. I don't understand enough to know if the aggregates being collected are useful (outside of sanity checking a configuration). It is outside my experience, so I want to ask. ;)

Once this binder is over in this project, it would be completely ok to re-implement the inside to satisfy the Statistics(Implementor) contract. I think that was a harder case to make when the binders existed in another repo entirely.

sebersole:
As for "warping" versus implementing StatisticsImplementor... for 5.x I'd say we leave it as you have it.  And look at moving to the implementation approach for 6

sebersole:
So for sure, things that make sense:

* Session open/close count
* connection counts
* transaction counts
* prepared statement counts

For entity, collection and natural-id counts I think we should *either* aggregation counts or per-entity, ...  I'm not sure which makes the most sense tbh.  I think it depends how do most people consume this kind of info, which you'd know better than me.  I am talking specifically the Micrometer metrics.  Are "nested" metrics common and/or useful in this context?

Similar for queries/query-cache

ebullient:
I honestly don't know that anyone knows. "It exposes the same statistics as would be exposed when calling {@link Statistics#logSummary()}." .. I am not sure any more thought went into it than that..

Sanne:
I think we have two goals here:
 - let ORM 5.4.x users migrate easily to the Micrometer extension which lives here
 - take better ownership of these metrics and improve them, as a subsequent step (branch ORM6?)

I would suggest to perform these two steps separately, so that people can easily swap the existing configuration for this one.
It might not be perfect, and some metrics might be questionable, but that would help for the transition.

As second step, it's easier to explain to users that metrics and statistics have been revisited when they are ready to upgrade to 6 - honestly I don't expect them to read much of a migration guide before that time.

sebersole:
I don't understand your distinction @Sanne 

WDYM "migrate easily to the Micrometer extension"?  Migrate from what exactly?  This is a branch new module.. nothing to migrate from.

If you mean migrate from Statistics, its still a completely different thing.  They can still use Statistics as they do atm.  This PR simply adds another method "on top of that.

So even for 5.x, "[taking] better ownership of these metrics" is still entirely reasonable imo.

I do think for 6 we ought to consider changing this module to provide an alternative Statistics impl to act as both the sink and the metric publisher.  But not for 5.  The reason being that in 6 we can redesign the actual Statistics(Implementor) contract based on this "better ownership"

Sanne:
> WDYM "migrate easily to the Micrometer extension"? Migrate from what exactly? This is a branch new module.. nothing to migrate from.

Sorry I didn't explain that!

This module is new within the Hibernate repository, but the Micrometer project was already distributing the same thing. This is about transferring the existing functionality, which potentially already has many users.

And I agree about the Statistics - I also expected them to plug a custom Statistics implementation, but that's another thing I'd do as subsequent step.

Sanne:
And also: Quarkus is already exposing these same statistics, using the existing Micrometer binder code (having the odd metrics).

sebersole:
Are Micrometer metrics "consumed" via API?  I thought it is similar to JMX stats...

ebullient:
Micrometer metrics go to registries that go to any number of places, but I would say the most typical (right now) is prometheus

sebersole:
Ok, so its a push model as opposed to pull.  Gotchya

sebersole:
@Sanne Do you mean that Quarkus simply applies this binder?  If it does not pull the metrics and does not really even push them... I'm not seeing how simplifying the metrics would be an issue.  

For consumers.. sure I get that, assuming the explicitly consume the exact metric(s) we decide to remove

jkschneider:
> Ok, so its a push model as opposed to pull. Gotchya

@sebersole Micrometer is unopinionated about this, it's up to the registry implementation. Many are push. Prometheus is pull.

hibernate-micrometer/src/main/java/org/hibernate/stat/HibernateMetrics.java
```diff
@@ -0,0 +1,369 @@
...
+				Statistics::getNaturalIdCachePutCount
+		);
+
+		counter(registry,
+				"hibernate.query.natural.id.executions",
+				"The number of naturalId queries executed against the database",
+				Statistics::getNaturalIdQueryExecutionCount
+		);
+
+		TimeGauge.builder(
```
jkschneider, line 268: I wonder if we should wrap this in a `TimeWindowMax` so this value decays more usefully. Could also publish both this "all-time" max and a decaying one separately.

hibernate-micrometer/src/test/java/org/hibernate/test/stat/AccountId.java
```diff
@@ -0,0 +1,43 @@
...
+    private final int id;
+
+    protected AccountId() {
+        this.id = 0;
+    }
+    
+    public AccountId(int id) {
+        this.id = id;
+    }
+    public int intValue() {
```
jkschneider, line 22: style nit on lack of blank line.

hibernate-micrometer/src/test/java/org/hibernate/test/stat/MicrometerStatisticsTest.java
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	private SimpleMeterRegistry registry = new SimpleMeterRegistry();
+	private HibernateMetrics hibernateMetrics;
+
+	@Override
+	protected void configure(Configuration configuration) {
+		super.configure( configuration );
+
+		configuration.setProperty( Environment.USE_SECOND_LEVEL_CACHE, "false" );
```
jkschneider, line 44: style nit on inconsistent use of space before/after first and last argument here and elsewhere in pr.

hibernate-micrometer/src/main/java/org/hibernate/stat/HibernateMetrics.java
ebullient: @sebersole ? This is one of those questions that comes back to which metrics are useful. ;)

I'm tempted to leave this one as it is/was in 5.4 (5.x?)

sebersole: Well I think this comes back to the generalized approach we want here...  The general consensus seemed to be:

1) keep exposed metrics as they are today in your existing integration for 5.x
2) redesign/redefine the metrics that actually make sense to expose for 6.x

I'm not sure that second part will ever come to fruition - people become loathe to change things after they are exposed.  Even more so here I would think for a "soft API".  But y'all have more experience with changing these "APIs" from release to release, so I'll defer to what you think.

ebullient: In a way, it's an optionally used thing.. people can always copy the existing one if they need to keep existing behavior... which is why I can see an argument for updating the behavior here if that makes more sense.

OTOH, this is also an older stream, so I could justify doing more along the lines of "recommended metrics" in version 6 (as there are a lot of other changes in the area, too).

What isn't apparent from the 5.x API is which metrics are there for historical reasons, which are there for sanity checking / debugging, and which are actually useful for gaining some notion of health or performance you'd want to monitor periodically over time, etc.

So I can go either way here: pick up TimeWindowMax so the max decays usefully (or publish both to preserve existing values), or leave everything as-is/was, and we can make better recommendations in the 6.x stream.

sebersole: > In a way, it's an optionally used thing, people can always copy the existing one if they need to keep existing behavior. Which is why I can see either side as valid. ;)

Unless I misunderstand, simply dropping in the old one would not work as I'd plan on making at least some of the corresponding changes to the Statistics API in 6.  Many of those do not make much sense, and the cache region name stuff at the very least need to be simplified


> What isn't apparent from the 5.x API is which metrics are there for historical reasons, which are there for sanity checking / debugging, and which are actually useful for gaining some notion of health or performance you'd want to monitor periodically over time, etc.

I was not involved in the original metrics impl nor consumed it, so I'm not the one to answer that.


> So whichever you want: if you think including better decaying behavior in 5.x is a cool thing to do, I can add it. (both existing max + decaying max?) If you would rather focus improvement on gathered metrics for 6.x, I'm fine w/ that, too.

I think the decaying max is more useful.  And I just can't imagine that the distinction is going to negatively impact anyone.  Famous last words...

ebullient: I'll add it to the 5.x one then. 6.x has API changes.. and I think we can/should redo that one to contain useful metrics, full stop. ;) If that means dropping things, it means dropping things (but 5 -> 6 should allow that)

sebersole:
@ebullient If you have a Trello account, I'd love to get any thoughts you have wrt 6.0 - https://trello.com/c/SHe6BhCb/178-revisit-statistics

ebullient:
> https://trello.com/c/SHe6BhCb/178-revisit-statistics

I do (redhat email)

Sanne:
I've squashes the commits and merged them - pushing in both `5.4` and `master`. Thanks @ebullient !

ebullient:
Thanks, Sanne! =)

PR #3652, opened 2020-11-19
HHH-13414 Fix issues with zero based query parameter binding for old style native queries

beikov:
https://hibernate.atlassian.net/browse/HHH-13414


gavinking:
This will slightly impact HR.

beikov:
Steve mentioned in the JIRA issue that we won't do this.

PR #3653, opened 2020-11-19
HHH-14335 Remember the resolved type in SimpleValue#getType instead of resolving multiple times

beikov:
https://hibernate.atlassian.net/browse/HHH-14335


Sanne:
I think the `type` field should be made `volatile` (at least) ? But maybe I'm misunderstanding the context of when this getType() method is invoked? I'm not familiar with the use case so I might be wrong.

Sanne:
Ok that's not necessary. Discussed on zulip.

PR #3656, opened 2020-11-22
Test case for HHH-14343

kawoolutions:
This is a follow-up issue for HHH-14276.


beikov:
Thanks for the test. Superseded by https://github.com/hibernate/hibernate-orm/pull/3658

PR #3662, opened 2020-11-24
Work on column alias resolving

beikov:


hibernate-core/src/main/java/org/hibernate/query/internal/FetchBuilderJpa.java
```diff
@@ -6,31 +6,46 @@
...
+	private final List<String> columnNames;
 
-	public FetchBuilderJpa(NavigablePath navigablePath, String attributePath) {
+	public FetchBuilderJpa(
+			NavigablePath navigablePath,
+			String attributePath,
+			List<String> columnNames) {
 		this.navigablePath = navigablePath;
 		this.attributePath = attributePath;
+		this.columnNames = columnNames;
```
sebersole, line 48: Are these passed `columns` ever used here?

hibernate-core/src/main/java/org/hibernate/query/internal/FetchMementoJpa.java
```diff
@@ -19,10 +20,12 @@
 
 	private final NavigablePath fetchPath;
 	private final String attributePath;
+	private final List<String> columnNames;
 
-	public FetchMementoJpa(NavigablePath fetchPath, String attributePath) {
+	public FetchMementoJpa(NavigablePath fetchPath, String attributePath, List<String> columnNames) {
```
sebersole, line 25: Here again, the only use of the passed `columns` is to pass them to `FetchBuilderJpa` which seems to never use them.

sebersole (approved):
Added a few comments, but other than that lgtm

hibernate-core/src/main/java/org/hibernate/query/internal/FetchBuilderJpa.java
beikov: Should I delete `FetchBuilderJpa` and `FetchMementoJpa`? They were only used by this code section https://github.com/hibernate/hibernate-orm/pull/3662/files#diff-ef7d256eac99c70fe4b496c2b04708b3dc6f71a99d333ac1aeb18c344ef16484L398 which I changed

PR #3663, opened 2020-11-25
[Feature OP#91771] New PostgreSQL Dialect mapping CLOB/BLOB to VARCHAR/BYTEA in favor of using OID

mvalchev:


mvalchev:
mistake

PR #3664, opened 2020-11-25
HHH-14348 Special handling in bytecode enhancement for lazy PersistentCollection fields

beikov:
https://hibernate.atlassian.net/browse/HHH-14348

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/LazyAttributesMetadata.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/LazyAttributesMetadata.java
@@ -34,7 +34,8 @@ public class LazyAttributesMetadata implements Serializable {
 	public static LazyAttributesMetadata from(
 			PersistentClass mappedEntity,
 			boolean isEnhanced,
-			boolean allowEnhancementAsProxy) {
+			boolean allowEnhancementAsProxy,
+			boolean collectionsInDefaultFetchGroupEnabled) {
 		final Map<String, LazyAttributeDescriptor> lazyAttributeDescriptorMap = new LinkedHashMap<>();
 		final Map<String, Set<String>> fetchGroupToAttributesMap = new HashMap<>();
 
@@ -48,7 +49,7 @@ public class LazyAttributesMetadata implements Serializable {
 					property,
 					isEnhanced,
 					allowEnhancementAsProxy,
-					false
+					collectionsInDefaultFetchGroupEnabled
 			);
 			if ( lazy ) {
 				final LazyAttributeDescriptor lazyAttributeDescriptor = LazyAttributeDescriptor.from( property, i, x++ );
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/LazyAttributesMetadata.java
```diff
@@ -49,7 +48,7 @@ public static LazyAttributesMetadata from(
 					property,
 					isEnhanced,
 					allowEnhancementAsProxy,
-					collectionsInDefaultFetchGroupEnabled
+					false
```
gavinking, line 51: Just curious: what was the reason for this change?

beikov: We have to consider attributes that are lazy from a persistence perspective, as lazy in bytecode enhancement. The flag would have caused the bytecode enhancement to see collections as part of the base fetch group which more or less means initialized. This is why the size code was triggered.

The whole bytecode enhancement logic is a bit of a mess in my opinion and I plan to improve this significantly in 6.0.

gavinking: @beikov well if that's the case then that sounds like we should be rolling back my patch which added `collectionsInDefaultFetchGroupEnabled` in the first place! Apparently it only worked by coincidence because there was another bug.

The semantics of `collectionsInDefaultFetchGroupEnabled` was _supposed_ to be:

1. that the field would be initialized, but
2. that the collection wrapper would remain uninitialized and unfetched.

It was definitely _not_ supposed to mean that the collection would be eagerly fetched.

But please define what you mean by "more or less" here.

beikov: By not considering this flag here, it gets your desired semantics.

This part of the code determines if an attribute is considered `lazy`, which bytecode enhancement uses for determining whether it needs to initialize the object on access. Any attributes that are not listed as lazy are considered to be initialized. So when we access the collection then through calling `size()` this causes lazy initialization.

gavinking:
OK, I confirm that the problem is fixed with even w/o the changes to `collectionsInDefaultFetchGroupEnabled`.

@Sanne do you think we could expedite a release of `hibernate-core` with this fix in it? So we can get the HR build back to normal?

gavinking (approved):

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/LazyAttributesMetadata.java
gavinking: FTR: @beikov and I chatted on the phone, and the decided that the changes related to `collectionsInDefaultFetchGroupEnabled` were not necessary.

Sanne:
> OK, I confirm that the problem is fixed with even w/o the changes to `collectionsInDefaultFetchGroupEnabled`.
> 
> @Sanne do you think we could expedite a release of `hibernate-core` with this fix in it? So we can get the HR build back to normal?

Sure, there's a couple more fixes we need also but I think we have PRs for all of them now. Monday I'll merge the PRs you need, backport some more and tag it.

gavinking:
OK great.

Sanne:
Looks great, many thanks @beikov !

gavinking:
yup, thanks @beikov

PR #3665, opened 2020-11-26
HHH-14350 - Set lock wait timeout in seconds instead of miliseconds

NicklasWallgren:


Sanne:
Many thanks, I've merged this in master and backported to 5.4

PR #3667, opened 2020-11-26
Various fixes for 6.0

beikov:
A quick overview of what I did:
* Implement support for group by and having clause
* Implement tuple emulation handling
* Follow SQL standard regarding collations i.e. collations can be applied to any scalar expression
* Fix timezone issues with timestamp literal parsing
* Fix a few Derby issues
* Pass through jdbc timezone for rendering timestamp literals with offsets/zones
* Implement cache key based on copied binding values for query result caching
* Fix a few generics warnings
* Allow configuring if parameters or literals should be created for values passed to JPA Criteria API. Don't interfere with the rendering process of literals to SQL anymore
* Caching for temporal literal parser formats


hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
```diff
@@ -404,23 +396,23 @@ likeEscape
...
+	| entityVersionReference collationSpecification?	# EntityVersionExpression
+	| entityNaturalIdReference collationSpecification?	# EntityNaturalIdExpression
+	| path collationSpecification?						# PathExpression
+	| function collationSpecification?					# FunctionExpression
+	| signOperator expression							# UnaryExpression
+	| expression datetimeField  						# ToDurationExpression
+	| expression BY datetimeField						# FromDurationExpression
+	| expression multiplicativeOperator expression		# MultiplicationExpression
+	| expression additiveOperator expression			# AdditionExpression
+	| expression DOUBLE_PIPE expression					# ConcatenationExpression
```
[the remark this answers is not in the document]
gavinking: @beikov 

What I mean is, I would much rather see the error 

    signed expression may not have a collatiotion

Than

    unexpected token: identifier

hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
```diff
@@ -572,6 +572,8 @@ protected SqlTypeDescriptor getSqlTypeDescriptorOverride(int sqlCode) {
 			// Therefore here we overwrite the sql type descriptors to
 			// use the non-N variants which are supported.
 			switch ( sqlCode ) {
+				case Types.BOOLEAN:
+					return SmallIntTypeDescriptor.INSTANCE;
```
gavinking, line 576: Why precisely was this needed?

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -87,11 +91,14 @@ public DerbyDialect(int version) {
 		registerColumnType( Types.BIT, 1, "boolean" ); //no bit
 		registerColumnType( Types.BIT, "smallint" ); //no bit
 		registerColumnType( Types.TINYINT, "smallint" ); //no tinyint
+		registerColumnType( Types.CHAR, "char(1)" );
```
gavinking, line 94: This should not be necessary, and is probably wrong.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -87,11 +91,14 @@ public DerbyDialect(int version) {
...
 
 		//HHH-12827: map them both to the same type to
 		//           avoid problems with schema update
 //		registerColumnType( Types.DECIMAL, "decimal($p,$s)" );
 		registerColumnType( Types.NUMERIC, "decimal($p,$s)" );
+		registerColumnType( Types.FLOAT, "float" );
+		registerColumnType( Types.DOUBLE, "double" );
```
gavinking, line 101: No, we're not using this mapping anymore in Hibernate 6, that was what we used to using in Hibernate 5 and it was wrong and I changed it across all the `Dialect`s.

I'm pretty sure you shouldn't need to be changing type mappings for Derby because I tested the shit out of all this on Derby when I cleaned up the dialects.

gavinking: See 

- https://db.apache.org/derby/docs/10.8/ref/rrefsqlj27281.html
- https://db.apache.org/derby/docs/10.8/ref/rrefsqljdoubleprecision.html

The default type mappings in `Dialect` should be perfect for Derby.

If you map `Types.FLOAT` to `float`, then Java `float`s and `doubles will both be mapped to the precision-less SQL type `float`.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -440,9 +453,16 @@ public boolean supportsTupleDistinctCounts() {
 	}
 
 	protected SqlTypeDescriptor getSqlTypeDescriptorOverride(int sqlCode) {
-		return sqlCode == Types.NUMERIC
-				? DecimalTypeDescriptor.INSTANCE
-				: super.getSqlTypeDescriptorOverride(sqlCode);
+		switch ( sqlCode ) {
+			case Types.BOOLEAN:
+				return SmallIntTypeDescriptor.INSTANCE;
```
gavinking, line 458: Same as above: why was this necessary?

gavinking (changes requested):

gavinking:
> Implement support for group by and having clause

Fantastic!

> Implement tuple emulation handling

What is this, precisely?

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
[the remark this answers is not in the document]
beikov: Gotcha, I'll improve that.

hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
beikov: DB2 has no booleans. I adapted this from https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java#L502 as I thought this was missed.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
beikov: Adapted from https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java#L95 which was the super type of the derby dialect

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
beikov: Ok, I just tried to copy over parts from 5 that seemed to be missing, but if you did this on purpose, I'll remove the double and float registrations.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
beikov: Same rationale as for DB2 from above: https://github.com/hibernate/hibernate-orm/pull/3667#discussion_r531191675

Derby has no booleans

beikov:
@gavinking 
>>* Implement tuple emulation handling

>What is this, precisely?

Emulate expressions like `(a, b) OP (1, 2)` for DBs that don't support that or only have partial support.

* `(a, b) = (1, 2)` emulated as `a = 1 and b = 2`
* `(a, b) > (1, 2)` emulated as `a > 1 or a = 1 and b = 2`
* `(a, b) > any (select x, y ...)` emulated as `exists (select 1 ... where (a, b) > (x, y) and ...)`
* `(a, b) > all (select x, y ...)` emulated as `(a, b) > (select x, y ... order by x desc, y desc limit 1)`
* `(a, b) > all (select x, y ...)` emulated as `not exists (select 1 ... where (a, b) <= (x, y) and ...)`

gavinking:
Excellent, that's what I hoped it would be.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
gavinking: Yeah but if you look in `Dialect`, we now have `registerColumnType( Types.CHAR, "char($l)" );`, and you're overriding that here with `"char(1)"` which wrongly treats the SQL `char()` type as a single character instead of a string.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
gavinking: No no, please don't do that. Nothing is missing, it's all defined in the superclass now. If you just copy stuff over from Hibernate 5, it will be wrong.

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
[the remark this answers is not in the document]
gavinking: thanks

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
gavinking: Are you certain? https://db.apache.org/derby/docs/10.8/ref/rrefsqljBoolean.html

hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
gavinking: > DB2 has no booleans.

I'm pretty sure that can't be right, since I know for sure I extensively tested the new `DB2Dialect`, and I'm using `boolean` as the mapping for both `BIT` and `BOOLEAN`.

gavinking: Apparently the answer is that `BOOLEAN` was introduced in DB2 11. So it looks like we need to make that type mapping conditional on the database version.

gavinking:
@beikov I created https://github.com/hibernate/hibernate-orm/pull/3668/commits/a5370b37976047d733dbd963db181a9f4f052ec3

But it's probably easier to just cherry-pick that commit onto your branch and close my P.R.

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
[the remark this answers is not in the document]
beikov: Should be fixed now

beikov:
Thanks for the comments @gavinking. I fixed the parts you requested.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -252,6 +262,12 @@ public String castPattern(CastType from, CastType to) {
 					return "case ?1 when false then 0 when true then 1 end";
 				}
 				break;
+			case STRING:
+				// See https://issues.apache.org/jira/browse/DERBY-2072
+				if ( from.getKind() == CastTypeKind.NUMERIC ) {
+					return "cast(cast(?1 as char(38)) as ?2)";
```
gavinking, line 268: 38?

Is there something magical about this number?

gavinking: Oh, right, I guess it's the maximum number of digits in a 32 bit integer....

gavinking: Rmm, nope, it's not even the max number of digits in a long.

gavinking: Curious to know where 38 came from.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -1257,7 +1258,8 @@ public String getSelectGUIDString() {
 	 * @return the SQL equivalent to Oracle's {@code from dual}.
 	 */
 	public String getFromDual() {
-		return "";
+		// The standard SQL solution to get a dual table is to use the VALUES clause
+		return "from (values (0)) as dual";
```
gavinking, line 1262: IIRC, Postgres doesn't need this, so if you're going to add this here, please also override it on `PostgresDialect` to return `""`.

Right now I don't recall if there were other databases which also don't require that.

gavinking (changes requested):

gavinking:
> Thanks for the comments @gavinking. I fixed the parts you requested.

Thank you!

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
beikov: This was the suggested workaround in the comment https://issues.apache.org/jira/browse/DERBY-2072?focusedCommentId=17154473&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17154473

I don't know what the maximum decimal precision is, so we might just as well use 254(char max capacity) to be safe, but 32 and 64 bit values should definitely fit into that data type. I'll just use the maximum capacity then and document this.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
beikov: If a query uses other clauses like `where`, `group by` etc. it needs a from clause. I can add another method `supportsSelectQueryWithoutFromClause` to handle these cases if you like.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
gavinking: > I don't know what the maximum decimal precision is

I think it's 31, but I might be wrong.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
gavinking: > If a query uses other clauses like `where`, `group by` etc. it needs a from clause. 

Ah, OK, right. But does it make sense to have these clauses if there is no `from` clause? I don't see how they could possibly be useful.

> I can add another method `supportsSelectQueryWithoutFromClause` to handle these cases if you like.

I would prefer that because `from (values (0)) as dual` is really pretty ugly and noisy.

beikov: For the record, I needed this in the emulation for tuple comparison like `(a, b) in ((1, 2), (2, 3))` for DBMS that only support tuples for IN predicates with subqueries like e.g. Oracle. I'll add an additional method to handle omitting the from clause altogether though.

gavinking: OK, that makes sense. 👍

beikov:
@sebersole: @dreab8 and I discussed that `SessionLazyDelegatorBaseImpl` is a temporary workaround to keep this PR managable. We'd rather use `WrapperOptions` in `org.hibernate.type.descriptor.sql.JdbcLiteralFormatter#toJdbcLiteral` to avoid the need for this class, but that then results in a change in every type to adapt `org.hibernate.type.descriptor.sql.spi.BasicJdbcLiteralFormatter#unwrap` which I wanted to avoid in this PR as that would add a lot of changes. I'll do this change in a follow up PR

beikov:
Forget that last comment. Andrea noticed that this can be done more easily and I went ahead and did it.

PR #3668, opened 2020-11-26
DB2 did not allow 'boolean' as a column type prior to 11

gavinking:
We should not use the DB2 `boolean` type as a column type on old versions.

cc @beikov


gavinking:
Note, I have *no way at all* of testing this change 🤷‍♂️

beikov:
This was already merged through some other PR, so this PR is obsolete now. Thanks anyway!

PR #3670, opened 2020-11-28
Unused parameter on createCriteriaUpdate(Delete)

Sam-Kruglov:
https://hibernate.atlassian.net/browse/HHH-13691

In order to create a manipulation query I have to pass the entity class twice because the first time it's not used:
```java
CriteriaDelete<T> criteriaDelete = entityManager.getCriteriaBuilder().createCriteriaDelete(User.class);
criteriaDelete.from(User.class);
```
This change makes it so that the first time I pass it, the class actually gets registered.


Sam-Kruglov:
> Unable to decrypt Maven password using PlexusCipher
org.sonatype.plexus.components.cipher.PlexusCipherException: java.lang.ArrayIndexOutOfBoundsException

(c) CI

beikov:
Let's continue the discussion on the JIRA issue. This is not as simple as you think.

gavinking:
The referenced issue was closed.

PR #3676, opened 2020-12-11
HHH-14365 PreparedStatementCount too high with criteria api and parameterExpression

Sputnik-git:
Add test case to  replicate the issue
#HHH-14365


gavinking:
AFAICT this is perfectly normal. I don't think that statistic reflects anything about prepared statement caching.

PR #3677, opened 2020-12-14
HHH-14370 + HHH-14371 + HHH-14372 - Add required --add-opens options for Gradle, Weld, Javassist

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14370
https://hibernate.atlassian.net/browse/HHH-14371
https://hibernate.atlassian.net/browse/HHH-14372

Workaround for https://github.com/gradle/gradle/issues/15538 , and a few more `--add-opens` to make tests work when illegal access is disabled. Strangely, we weren't using `--illegal-access=deny`? Anyway, JDK16-ea28 makes that the default.

This should fix compilation and tests on JDK16-ea28.
See https://ci.hibernate.org/view/Personal%20jobs/job/hibernate-orm-personal-yoann-jdk16/20/

Please do not backport to 5.4 yet: we need to backport the upgrade to toochains first.


yrodiere:
I added a few commits to also fix tests.

Sanne (approved):
Thanks!
I assume we should backport it?

yrodiere:
>  I assume we should backport it?

Yes, but only after we backport the toolchain support.

PR #3679, opened 2020-12-17
New Test annotation @NotImplementedYet

jrenaat:


hibernate-testing/src/main/java/org/hibernate/testing/orm/junit/NotImplementedYetExtension.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.TestExecutionExceptionHandler;
+
+import org.jboss.logging.Logger;
+
+public class NotImplementedYetExtension
+		implements ExecutionCondition, AfterEachCallback, TestExecutionExceptionHandler {
+
+	private static final Logger log = Logger.getLogger( NotImplementedYetExtension.class );
+	private static final String EXPECTED_FAILURE_STORE_KEY = "EXPECTED_FAILURE";
```
sebersole, line 15: This should be unique.  `STORE_KEY = "NOT_YET_IMPLEMENTED"` e.g.

hibernate-testing/src/main/java/org/hibernate/testing/orm/junit/NotImplementedYetExtension.java
```diff
@@ -0,0 +1,61 @@
...
+	@Override
+	public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+		log.debugf( "#evaluateExecutionCondition(%s)", context.getDisplayName() );
+
+		if ( !context.getElement().isPresent() ) {
+			throw new RuntimeException( "Unable to determine how to handle given ExtensionContext : " + context.getDisplayName() );
+		}
+
+		// Test this in case some other annotation were extended with NotImplementedYetExtension
+		if ( !TestingUtil.hasEffectiveAnnotation( context, NotImplementedYet.class ) ) {
```
sebersole, line 48: Why are you disabling all tests that are not marked with this new annotation?

sebersole (changes requested):
I made a few in-line comments.  But overall I wanted to point out that this extension should only effect methods/classes that are marked with this annotations.  And it should only "excuse" exceptions that are of type `NotYetImplementedFor6Exception`, at the moment it excuses any type of exception.

PR #3682, opened 2020-12-20
HHH-2558: Fix batching inserts for multi-table entities

muzmov:
The original fix of https://hibernate.atlassian.net/browse/HHH-2558 caused https://hibernate.atlassian.net/browse/HHH-12470 and was reverted by adding back jdbcCoordinator.executeBatch() in StatementPreparerImpl.prepareStatement(String sql, final boolean isCallable)

The problem with deletes in HHH-12470 was caused by batch deletes not being flushed before non batch delete. That's why we still need executeBatch() in prepareStatement(String sql, final boolean isCallable) that is used by non batch delete

The idea behind this fix is that we use prepareStatement(String sql) (which doesn't trigger executeBatch()) for AbstractBatchImpl.buildBatchStatement with non callable sql


gavinking:
The referenced issue was closed, apparently fixed. And the code has changed very significantly.

PR #3671, opened 2020-12-02
HHH-13944 HQL/JPQL size() does not work (anymore) with nested expression

dreab8:
https://hibernate.atlassian.net/browse/HHH-13944

PR #3684, opened 2020-12-23
HHH-13944 : HQL/JPQL size() does not work (anymore) with nested expression (added tests)

gbadner:
I've rebased https://github.com/hibernate/hibernate-orm/pull/3671, and added some tests that do join fetches. There are a couple of things that I am unsure about. Details are in comments below.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/hql/size/SizeTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/hql/size/SizeTest.java
@@ -92,6 +92,17 @@ public class SizeTest extends BaseNonConfigCoreFunctionalTestCase {
 	}
 
 	@Test
+	public void testSizeAddFetch() {
+		inTransaction( session -> {
+			List<Teacher> teachers = session.createQuery(
+					"select distinct teacher from Teacher teacher left join fetch teacher.skills join teacher.skills skills where size(skills) > 0",
+					Teacher.class
+			).getResultList();
+			assertEquals( 2, teachers.size() );
+		} );
+	}
+
+	@Test
 	public void testSizeWithoutNestedPath() {
 		inTransaction( session -> {
 			List<Teacher> teachers = session.createQuery(
@@ -103,6 +114,31 @@ public class SizeTest extends BaseNonConfigCoreFunctionalTestCase {
 	}
 
 	@Test
+	public void testSizeWithoutNestedPathAddFetch() {
+		inTransaction( session -> {
+			List<Teacher> teachers = session.createQuery(
+					"select teacher from Teacher teacher left join fetch teacher.skills where size(teacher.skills) > 0",
+					Teacher.class
+			).getResultList();
+			assertEquals( countNumberOfJoins( sqlStatementInterceptor.getSqlQueries().get( 0 ) ), 2 );
+			assertEquals( 2, teachers.size() );
+		} );
+	}
+
+	@Test
+	public void testSizeWithoutNestedPathAddFetchDistinct() {
+		inTransaction( session -> {
+			List<Teacher> teachers = session.createQuery(
+					"select distinct teacher from Teacher teacher left join fetch teacher.skills where size(teacher.skills) > 0",
+					Teacher.class
+			).getResultList();
+			assertEquals( countNumberOfJoins( sqlStatementInterceptor.getSqlQueries().get( 0 ) ), 2 );
+			assertEquals( 2, teachers.size() );
+		} );
+	}
+
+
+	@Test
 	public void testSizeWithNestedPathAndImplicitJoin() {
 		doInJPA( this::sessionFactory, em -> {
 			List<Student> students = em.createQuery(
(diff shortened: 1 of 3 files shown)


hibernate-core/src/test/java/org/hibernate/test/hql/size/SizeTest.java
```diff
@@ -91,6 +91,17 @@ public void testSize() {
 		} );
 	}
```
gbadner, line 93: The following works properly because it uses `distinct`.

hibernate-core/src/test/java/org/hibernate/test/hql/size/SizeTest.java
```diff
@@ -91,6 +91,17 @@ public void testSize() {
...
 
+	@Test
+	public void testSizeAddFetch() {
+		inTransaction( session -> {
+			List<Teacher> teachers = session.createQuery(
+					"select distinct teacher from Teacher teacher left join fetch teacher.skills join teacher.skills skills where size(skills) > 0",
+					Teacher.class
+			).getResultList();
```
gbadner, line 100: The following assertion passes because the query uses `distinct`.

hibernate-core/src/test/java/org/hibernate/test/hql/size/SizeTest.java
```diff
@@ -91,6 +91,17 @@ public void testSize() {
...
+			assertEquals( 2, teachers.size() );
```
gbadner, line 101: The above assertion passes because the query uses distinct.

hibernate-core/src/test/java/org/hibernate/test/hql/size/SizeTest.java
```diff
@@ -102,6 +113,31 @@ public void testSizeWithoutNestedPath() {
...
 
+	@Test
+	public void testSizeWithoutNestedPathAddFetch() {
+		inTransaction( session -> {
+			List<Teacher> teachers = session.createQuery(
+					"select teacher from Teacher teacher left join fetch teacher.skills where size(teacher.skills) > 0",
+					Teacher.class
+			).getResultList();
+			assertEquals( countNumberOfJoins( sqlStatementInterceptor.getSqlQueries().get( 0 ) ), 2 );
+			assertEquals( 2, teachers.size() );
```
gbadner, line 124: The above assertion fails because the `Teacher` with 2 `Skill` objects in its collection ends up getting duplicated.

I believe this is a bug. I don't think it should be necessary to use `distinct` when adding a `join fetch`, but I am not 100% sure about that.

hibernate-core/src/test/java/org/hibernate/test/hql/size/SizeTest.java
```diff
@@ -102,6 +113,31 @@ public void testSizeWithoutNestedPath() {
...
+
+	@Test
+	public void testSizeWithoutNestedPathAddFetchDistinct() {
+		inTransaction( session -> {
+			List<Teacher> teachers = session.createQuery(
+					"select distinct teacher from Teacher teacher left join fetch teacher.skills where size(teacher.skills) > 0",
+					Teacher.class
+			).getResultList();
+			assertEquals( countNumberOfJoins( sqlStatementInterceptor.getSqlQueries().get( 0 ) ), 2 );
+			assertEquals( 2, teachers.size() );
```
gbadner, line 136: The above assertion passes because `distinct` is used in the query. 

I don't believe that using `distinct` should be necessary though. See https://github.com/hibernate/hibernate-orm/pull/3684/commits/9c1f805e53d15c377cb0ded21ba1f859658bc412#r547585526 .

hibernate-core/src/test/java/org/hibernate/test/hql/size/SizeTest.java
```diff
@@ -114,6 +150,30 @@ public void testSizeWithNestedPathAndImplicitJoin() {
...
 
+	@Test
+	public void testSizeWithNestedPathAndImplicitJoinAddFetch() {
+		doInJPA( this::sessionFactory, em -> {
+			List<Student> students = em.createQuery(
+					"select student from Student student left join fetch student.teacher t left join fetch t.skills where size(student.teacher.skills) > 0",
+					Student.class
+			).getResultList();
+			assertEquals( countNumberOfJoins( sqlStatementInterceptor.getSqlQueries().get( 0 ) ), 3 );
+			assertEquals( 1L, students.size() );
```
gbadner, line 161: The above assertion fails for the same reason as `#testSizeWithoutNestedPathAddFetch`.

hibernate-core/src/test/java/org/hibernate/test/hql/size/SizeTest.java
```diff
@@ -114,6 +150,30 @@ public void testSizeWithNestedPathAndImplicitJoin() {
...
+
+	@Test
+	public void testSizeWithNestedPathAndImplicitJoinAddFetchDistinct() {
+		doInJPA( this::sessionFactory, em -> {
+			List<Student> students = em.createQuery(
+					"select distinct student from Student student left join fetch student.teacher t left join fetch t.skills where size(student.teacher.skills) > 0",
+					Student.class
+			).getResultList();
+			assertEquals( countNumberOfJoins( sqlStatementInterceptor.getSqlQueries().get( 0 ) ), 3 );
+			assertEquals( 1L, students.size() );
```
gbadner, line 173: The above assertion works for the same reason as `#testSizeWithoutNestedPathAddFetchDistinct`.

hibernate-core/src/test/java/org/hibernate/test/hql/size/SizeTest.java
beikov: I agree with you, this seems like a bug.

gbadner:
I'm closing this PR because I found that the JPA spec says that `distinct` should be added to remove duplicates when a query has `join`. I was thinking that `join fetch` would be treated differently in this respect from a simple `join`. Apparently, the same requirement applies to `join fetch`.

gavinking:
> I'm closing this PR because I found that the JPA spec says that `distinct` should be added to remove duplicates when a query has `join`. I was thinking that `join fetch` would be treated differently in this respect from a simple `join`. Apparently, the same requirement applies to `join fetch`.

That does gel with my hazy recollection of this. I believe Hibernate always behaved that way, and the behavior was inherited by JPA.

hibernate-core/src/test/java/org/hibernate/test/hql/size/SizeTest.java
gbadner: @beikov, my guess for requiring ```distinct``` is that a ```join fetch``` is a kind of ```join```, and not something special with different rules. That's just my guess though.

PR #3685, opened 2020-12-23
Add support for FETCH clause, CTEs and set operations

beikov:
Here a quick overview of the changes I did:

* SqmTranslatorFactory now receives SqmStatement during creation instead during translate call
* SqlAstTranslatorFactory now receives Statement during creation instead during translate call
* SqlAstTranslator receives `JdbcParameterBindings` and `QueryOptions` for `translate` which it can use for SQL rendering if needed
  => For some DBMS it is necessary to render parameters as literals in certain clauses
  => Note that the generated `JdbcOperation` has the capability to report if a different `JdbcParameterBindings` + `QueryOptions` are compatible
  => To avoid applying `QueryOptions` multiple times, the SQL related parts are "omitted" when passed to the executor through `SqlOmittingQueryOptions.omitSqlQueryOptions`
* Removed SimpleSqmUpdateTranslator, SimpleSqmDeleteTranslator, SqmInsertTranslator and SqmSelectTranslator and use the generic variant `SqmTranslator<? extends Statement>`
  => This allows to have a single translator class per dialect whereas before, we required 4
* Removed SqlAstUpdateTranslator, SqlAstDeleteTranslator, SqlAstInsertTranslator and SqlAstSelectTranslator and use the generic variant `SqlAstTranslator<? extends JdbcOperation>`
  => This allows to have a single translator class per dialect whereas before, we required 4
* Rework CTE support in HQL and SQL AST model to reflect what the SQL standard does
* Rework CTE SqmMultiTableMutationStrategy to use modifiable CTEs as it previously was equivalent to using inline VALUES clause
* QueryPart and SqmQueryPart as abstraction over set producing query expressions
* Add support for set operations (UNION, INTERSECT, EXCEPT) in HQL and SQL AST model
* Move most SQM AST translation logic to BaseSqmToSqlAstConverter for simpler extension for Dialect
* Move most SQL AST translation logic to AbstractSqlAstTranslator for simpler extension for Dialect
* Avoid rendering subclass or secondary table joins in general and especially for multi-table DML when possible
* Avoid rendering null precedence, and thus also avoid emulation, if it matches the null ordering of the database
* Implement all limit handler logic in `SqlAstTranslator` along with emulations for `percent` and `with ties` fetching if possible
* Integrate `LimitHandler`, lock-, hint- and comment rendering to SQL as preparation for native query support
* Use `Limit` in `LimitHandler` instead of now deprecated `RowSelection`
* Fix `PathRegistry` and `FromClauseIndex` related alias issues when using correlations in subqueries
* Implementing correlation path support for subqueries in SQM and JPA Criteria

In an upcoming PR I still want to:

* JPA Criteria support for set operations + more set operation tests
* Emulate INTERSECT and EXCEPT with EXISTS and NOT EXISTS
* Emulate UNION/INTERSECT/EXCEPT ALL with window functions


gavinking:
Wow, this sounds great!

Can you please give us a few examples of the kinds of queries that are supported here?

beikov:
For set operations, the prime example would be `FROM Person WHERE ... UNION ALL FROM Person WHERE ...` or `SELECT name FROM person UNION ALL SELECT name FROM some_other_relation`.

Deletes with the new CTE strategy implementation would generate queries like this:

```
with cte_ids as(
  select id from person where ...
), delete_1 as (
  delete from collection_table where person_id in (select id from cte_ids) returning person_id
), delete_2 as (
  delete from person where id in (select id from cte_ids) returning id
)
select count(*) from cte_ids
```

instead of multiple queries, which reduces the roundtrip time. I have planned to add another strategy called "anonymous block" which can be used for Oracle and other DBMS to combine multiple DML statements into a single statement.

The fetch clause part is essentially based on what PostgreSQL supports: https://www.postgresql.org/docs/current/sql-fetch.html

PR #3686, opened 2021-01-06
HHH-14392: Test case for CASE WHEN ... THEN ... END failing in @org.hibernate.annotations.OrderBy

kawoolutions:
Adding test case for https://hibernate.atlassian.net/browse/HHH-14392

NOTE THAT THE FULL TEST CASE CODE ISN'T REQUIRED. -> the code of methods `setUp` and `testOrderByCaseWhen` could be commented completely.

It is only required after fixing to test the bottommost, most complex `@OrderBy` example commented in the code above the `teams` relationship to show that the CASE WHEN ordering is working.


kawoolutions:
What needs to be done here?

beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/3895

PR #3687, opened 2021-01-07
Fix most test issues on GH actions

beikov:
Fixes:

* https://hibernate.atlassian.net/browse/HHH-14432
* https://hibernate.atlassian.net/browse/HHH-14433
* https://hibernate.atlassian.net/browse/HHH-14434
* https://hibernate.atlassian.net/browse/HHH-14435


docker_db.sh
```diff
@@ -22,10 +22,10 @@ postgresql_9_5() {
 
 db2() {
     docker rm -f db2 || true
-    docker run --name db2 --privileged -e DB2INSTANCE=orm_test -e DB2INST1_PASSWORD=orm_test -e DBNAME=orm_test -e LICENSE=accept -p 50000:50000 -d ibmcom/db2:11.5.0.0a
+    docker run --name db2 --privileged -e DB2INSTANCE=orm_test -e DB2INST1_PASSWORD=orm_test -e DBNAME=orm_test -e LICENSE=accept -e AUTOCONFIG=false -e ARCHIVE_LOGS=false -e TO_CREATE_SAMPLEDB=false -e REPODB=false -p 50000:50000 -d ibmcom/db2:11.5.5.0
     # Give the container some time to start
     OUTPUT=
-    while [[ $OUTPUT != *"Setup has completed"* ]]; do
+    while [[ $OUTPUT != *"INSTANCE"* ]]; do
         echo "Waiting for DB2 to start..."
```
NathanQingyangXu, line 29: Rather than ad-hoc logging statement pattern, we might well use robust way to detect whether container has been ready:
```
until [ "`docker inspect -f {{.State.Health.Status}} db2`" == "healthy" ]; do sleep 10; done
```

docker_db.sh
```diff
@@ -35,7 +35,23 @@ db2() {
...
+        sleep 5
+    done
+    if [ "$n" -ge 5 ]; then
+      echo "SQL Server failed to start and configure after 25 seconds"
+    else
+      echo "SQL Server successfully started"
+    fi
 }
 
 oracle() {
```
NathanQingyangXu, line 57: It might be considerable to make it a rule to always provide image tag. Without explicit tag, the latest version would always be fetched, which might not have been tested if a new version is released.

docker_db.sh
beikov: The DB2 image doesn't have a built-in health check

docker_db.sh
beikov: True, but it wasn't updated in 2 years and also probably won't be updated.

docker_db.sh
```diff
@@ -12,7 +12,7 @@ mysql_8_0() {
 
 mariadb() {
     docker rm -f mariadb || true
-    docker run --name mariadb -e MYSQL_USER=hibernate_orm_test -e MYSQL_PASSWORD=hibernate_orm_test -e MYSQL_DATABASE=hibernate_orm_test -p3306:3306 -d mariadb:10.5.8 --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci
+    docker run --name mariadb -e MYSQL_USER=hibernate_orm_test -e MYSQL_PASSWORD=hibernate_orm_test -e MYSQL_DATABASE=hibernate_orm_test -e MYSQL_ROOT_PASSWORD=hibernate_orm_test -p3306:3306 -d mariadb:10.5.8 --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci
```
NathanQingyangXu, line 15: We might well avoid the error message for 'rm' statement by simply including "--rm=true" into the "docker run" statement. "docker rm -f xxx || true" seems a little over engineering.

ci/build.sh
```diff
@@ -4,15 +4,15 @@ goal=
...
+  goal="-Pdb=oracle_ci"
 elif [ "$RDBMS" == "db2" ]; then
-  goal="-Pdb=db2 -Dhibernate.connection.url=jdbc:db2://localhost:50000/orm_test -Dhibernate.connection.username=orm_test -Dhibernate.connection.password=orm_test"
+  goal="-Pdb=db2_ci"
 elif [ "$RDBMS" == "mssql" ]; then
-  goal="-Pdb=mssql -Dhibernate.connection.url=jdbc:sqlserver://localhost:1433;databaseName= -Dhibernate.connection.username=sa -Dhibernate.connection.password=hibernate_orm_test"
+  goal="-Pdb=mssql_ci"
 fi
 
 exec ./gradlew check ${goal} -Plog-test-progress=true --stacktrace
```
NathanQingyangXu, line 18: if the command or arguments involve string interpolation, it would be much less error-prone to use 'eval' instead of 'exec'. I learned painful lesson on this when I need to pass JVM option with value containing quotes (e.g. **-Dhibernate.connection.username="sys as dba"**).
Seems 'exec' is mainly used for hard coded command and arguments. When dynamic string value is involved, eval is more appropriate (you can combine two together if you really wanna the benefit of exec).

hibernate-testing/src/main/java/org/hibernate/testing/cleaner/AbstractMySQLDatabaseCleaner.java
```diff
@@ -0,0 +1,186 @@
...
+			}
+			catch (SQLException e) {
+				throw new RuntimeException( e );
+			}
+		} );
+	}
+
+	@Override
+	public void clearData(Connection connection, String schemaName) {
+		clearData0( connection, schemaName, s -> {
```
sebersole, line 125: Stylistically I find this difficult to read.  I know everyone will have different takes on this.  But understanding the arguments here was not instantaneous - it took me a few times looking between the call and the signature to figure it out.

sebersole: IMO, this is best:
```
clearData0(
    connection,
    schemaName,
    (tableName) -> {...}
);
```

PR #3688, opened 2021-01-07
Upgrade the DB2 container image boot script

Sanne:
hi @beikov , when updating the DB2 container it no longer required "privileged", and this makes it compatible with podman so more people could use the scripts :)

Also added a couple variables which help a little bit with its horrible boot times.

Finally, changed the message it's waiting for, as "Setup has completed" is a little too early still.

Does this need a JIRA?  Feel free to take the commit and include it in other related works..

### DIFF
--- a/docker_db.sh
+++ b/docker_db.sh
@@ -22,10 +22,10 @@ postgresql_9_5() {
 
 db2() {
     docker rm -f db2 || true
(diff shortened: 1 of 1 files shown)


beikov:
Thanks, I applied this to my existing PR that tries to fix a few test issues on master: https://github.com/hibernate/hibernate-orm/pull/3687

beikov:
By the way, I had to add the `--privileged` flag again as it wasn't working without that.

PR #3691, opened 2021-01-13
HHH-14406 fix Architecture svg image in user guide

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14406

The PR will replace the previous svg image:

Screen Shot 2021-01-13 at 6 02 05 PM

with
Screen Shot 2021-01-13 at 6 02 27 PM

The issue is the ‘session’ suite diagram is confusing for it implies it is special or different from the other two suites. That confuses user guide reader a lot.


NathanQingyangXu:
@Sanne , could you take a look at this PR? Not sure whether it is necessary or not, though I did spend lots of time tweaking the svg file. Who is the last svg editor? Maybe he knows better why 'session' suite were dealt with differently than the other two ones?

NathanQingyangXu:
@vladmihalcea I saw your were the original svg file author. Could you take a look at this PR? Still wondering whether my understanding is correct or not. Do we have any reason to treat 'session' suite differently?

vladmihalcea:
The SVG was taken with IntelliJ IDEA 5 years ago when the `EntityManagerImpl` was having a reference to the `Session`. But, that's not the case anymore since the `Session` implements `EntityManager`.

NathanQingyangXu:
> The SVG was taken with IntelliJ IDEA 5 years ago when the `EntityManagerImpl` was having a reference to the `Session`. But, that's not the case anymore since the `Session` implements `EntityManager`.

Thanks. Now the SVG was not be previewed in latest IDEA and it would make its maintenance more and more difficult (nobody dares to touch it without preview effect). I would modify the SVG file to make it simple and maintainable (currently there are tons of dead hyperlink URLs).

NathanQingyangXu:
Finally I found the reason why SVG preview failed in Intellij. As long as it contains image element with data URI, preview will fail. However, it seems image with data URI is so common (every web browser can open the svg without problem). Not sure why Intellij plugin makes such decision (maybe for security reason).

NathanQingyangXu:
Now the svg preview works in latest Intellij IDE. This is a good doc improvement. Who can take a look at it?

dreab8:
Thanks @NathanQingyangXu

PR #3692, opened 2021-01-13
HHH-10550: Replaced int cast/usage for long implementation

luiscla27:
Fixes 
https://hibernate.atlassian.net/browse/HHH-10550

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/BlobProxy.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/BlobProxy.java
@@ -201,7 +201,7 @@ public final class BlobProxy implements Blob, BlobImplementer {
 
 		@Override
 		public long getLength() {
-			return (int) length;
+			return length;
 		}
 
 		@Override
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/BlobProxy.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/BlobProxy.java
@@ -105,6 +105,18 @@ public final class BlobProxy implements Blob, BlobImplementer {
 	}
 
 	@Override
+	@Deprecated
+	public byte[] getBytes(final long start, final int length) throws SQLException {
+		if ( start < 1 ) {
+			throw new SQLException( "Start position 1-based; must be 1 or more." );
+		}
+		if ( length < 0 ) {
+			throw new SQLException( "Length must be great-than-or-equal to zero." );
+		}
+		return DataHelper.extractBytes( getBinaryStream() );
+	}
+
+	@Override
 	public byte[] getBytes(final long start, final long length) throws SQLException {
 		if ( start < 1 ) {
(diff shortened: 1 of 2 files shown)


NathanQingyangXu:
Thsnks for the bug reporting and the PR. Seems a good catch and fixing. The only thing that is missing is a testing case which shows:

1. without the code fixes it would be broken;
2. with the code fixes it would succeed.

Usually there is slim chance for the PR to be accepted without such compelling testing case (unless the PR is to fix some documentation). Please see existing testing cases to get the clue how to come up with one. It is not difficult and coding such testing case is both fun and mandatory.

luiscla27:
Thank you for taking a look in this PR :)

I'll take a look on the existing tests, it might be specially hard because the test-case would need to hold in memory a binary which exceeds the value of the `Integer.MAX_VALUE` constant (2,147,483,647 bytes). Any suggestion??

In my real world use-case issue that's actually what is happening...

NathanQingyangXu:
> Thank you for taking a look in this PR :)
> 
> I'll check existing tests, it might be specially hard to create one for this issue, specially because the test-case would need to hold in memory a binary which exceeds the value of the `Integer.MAX_VALUE` constant (2,147,483,647 bytes). Any suggestion??
> 
> In my real world use-case issue that's actually what is happening...

Could you mock something to simulate that then?

hibernate-core/src/main/java/org/hibernate/engine/jdbc/BlobProxy.java
```diff
@@ -201,7 +207,7 @@ public byte[] getBytes() {
 
 		@Override
 		public long getLength() {
-			return (int) length;
+			return length;
```
gavinking, line 210: This was already fixed a while ago.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/BinaryStreamImpl.java
```diff
@@ -18,7 +18,7 @@
  * @author Steve Ebersole
  */
 public final class BinaryStreamImpl extends ByteArrayInputStream implements BinaryStream {
-	private final int length;
+	private final long length;
```
gavinking, line 21: A java array can't have a `long` number of elements, so this change is unnecessary.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/BlobProxy.java
```diff
@@ -112,6 +112,9 @@ public byte[] getBytes(final long start, final int length) throws SQLException {
 		if ( length < 0 ) {
 			throw new SQLException( "Length must be great-than-or-equal to zero." );
 		}
+		if ( length == length()) {
+			return DataHelper.extractBytes( getBinaryStream() );
+		}
```
gavinking, line 117: This is maybe not a terrible idea, but it's also not quite right because `start` should also be checked.

gavinking:
Closing this because the bug was already fixed.

PR #3693, opened 2021-01-14
HHH-14326 + HHH-14404 Connection closing fixes

yrodiere:
* [HHH-14404](https://hibernate.atlassian.net/browse/HHH-14404): SessionBuilder.connectionHandlingMode is ignored
* [HHH-14326](https://hibernate.atlassian.net/browse/HHH-14326): StoredProcedureQuery instances should be closed by Hibernate ORM

Necessary for https://github.com/quarkusio/quarkus/issues/13273 and https://github.com/quarkusio/quarkus/issues/7242 in particular.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/LogicalConnectionManagedImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/LogicalConnectionManagedImpl.java
@@ -205,15 +205,7 @@ public class LogicalConnectionManagedImpl extends AbstractLogicalConnectionImple
 		// when releasing resources, we'll abort the batch statement,
 		// which will trigger "logicalConnection.afterStatement()",
 		// which in some configurations will release the connection.
-
-		//Some managed containers might trigger this release concurrently:
-		//this is not how they should do things, still we try to detect it to trigger a more clear error.
-		boolean concurrentUsageDetected = ( this.physicalConnection == null );
 		this.physicalConnection = null;
-		if ( concurrentUsageDetected ) {
-			throw new HibernateException( "Detected concurrent management of connection resources." +
-					" This might indicate a multi-threaded use of Hibernate in combination with managed resources, which is not supported." );
-		}
 		try {
 			try {
 				getResourceRegistry().releaseResources();
--- a/hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/LogicalConnectionManagedImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/LogicalConnectionManagedImpl.java
@@ -194,16 +194,28 @@ public class LogicalConnectionManagedImpl extends AbstractLogicalConnectionImple
 	}
 
 	private void releaseConnection() {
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/LogicalConnectionManagedImpl.java
```diff
@@ -194,16 +194,28 @@ public void manualReconnect(Connection suppliedConnection) {
...
+		// when releasing resources, we'll abort the batch statement,
+		// which will trigger "logicalConnection.afterStatement()",
+		// which in some configurations will release the connection.
+
+		//Some managed containers might trigger this release concurrently:
+		//this is not how they should do things, still we try to detect it to trigger a more clear error.
+		boolean concurrentUsageDetected = ( this.physicalConnection == null );
+		this.physicalConnection = null;
+		if ( concurrentUsageDetected ) {
+			throw new HibernateException( "Detected concurrent management of connection resources." +
```
Sanne, line 214: I guess it's now impossible to ever get into this state?

I'd be OK with just removing the exception; in the original version before the "attempt to detect" we actually had an NPE triggered - that doesn't seem a likely outcome today so feel free to simplify this a bit.

yrodiere: I suppose we could end up in this state in some cases since the read and write operations on `this.physicalConnection` are not atomic, but I'm not sure. In any case, I can imagine ending up in that state, but yes that sounds unlikely. Even concurrent execution of the code below sounds quite unlikely, though possible.

We could be absolutely sure concurrent execution never happens by using an `AtomicReference` for `this.physicalConnection`, but that feels like overkill, considering people are not supposed to release connections concurrently in the first place?

Anyway, I added a commit to remove this check.

Sanne: thanks :)

And yes consider that this field isn't even `volatile`, compiled code is most likely reading it just once - or even if it did twice, it's not having any memory barrier. So even the very little window which seems present is unlikely to be a real possibility.

hibernate-core/src/test/java/org/hibernate/test/connections/BeforeCompletionReleaseTest.java
```diff
@@ -56,12 +71,40 @@ protected Map getConfig() {
...
+        } );
+
+        Connection connectionSpy = connectionSpies[0];
+
+        // Must close the resources, then the connection, then commit
+        InOrder inOrder = inOrder( statementMock, connectionSpy, transactionSpy );
+        inOrder.verify( statementMock ).close();
+        inOrder.verify( connectionSpy ).close();
+        inOrder.verify( transactionSpy ).commit( any(), anyBoolean() );
+    }
```
Sanne, line 138: :100:

Sanne:
great, merged it in master, checking backports now..

yrodiere:
Thanks!

Sanne:
all merged

PR #3694, opened 2021-01-15
HHH-14407 NPE in AbstractMultiTableBulkIdStrategyImpl for column 'hib_sess_id' of PersistentTableBulkIdStrategy

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14407

The NPE was introduced by https://github.com/hibernate/hibernate-orm/pull/3498. Specifically, the following new line introduced in that PR didn't take into consideration that some Column has null value (like the '**hib_sess_id**' column in **PersistentTableBulkIdStrategy**) so it would trigger NPE:

Screen Shot 2021-01-14 at 11 37 11 PM

The fix is simple as the following two steps:

1. set the sql code type of the '**hib_sess_id**' to **Types.VARCHAR**;
2. in the above highlighted line, the **Column#getSqlTypeCode(Mapping mapping)** method is only invoked when **Column#getSqlTypeCode()** returns null; otherwise simply return the internal sqlTypeCode value directly, which is in line with the Javadoc of the method:

Screen Shot 2021-01-14 at 11 42 06 PM


beikov:
Thanks for the fix!

Sanne:
great, thanks @NathanQingyangXu !

PR #3695, opened 2021-01-15
Various fixes for 6.0

beikov:
Changes:
* Handle funky Derby casting requirements
* Change binder to ignore `precision < 1` to fix float type on DBs that support specifying a precision for float
* Extend `CastTarget` to support casting from/to `YesNoBoolean`, `TrueFalseBooleanType`, `NumericBooleanType` and `BooleanType` to/from String and Integer types in a consistent manner.
* Create emulation for `LEAST`/`GREATEST`
* Fix overlay emulation issues for Derby due to strict substring handling
* Properly implement `WEEK` and `DAY_OF_WEEK` extraction for Derby
* Fix sequence support for Derby
* Support translating same `SqmParameter` instance multiple times => Needed for e.g. Derby to emulate the `OVERLAY` function
* Add some keywords for SQL Template to support more SQL expressions for `@Formula`
* Introduce `DialectContext` for caching `Dialect` to use for dialect precondition checks, which is independent of `SessionFactory` used for tests. This fixes a circularity problem in dialect check extensions
* Fix a few tests and fix issues encountered while testing MariaDB
* Consider size 255 for UUID, Character and Boolean as "default JPA size"
* Fix support for updating associations to `null` through DML queries


gavinking:
Could you please give more details and explanation for these changes. It's really difficult to see the motivation behind some of them, even though some I can guess at.

For example, I want to see a proper argument made for changing the casting logic to something stringly-typed. I'm not sure that's a good change at all.

Also, with respect to the Derby-related changes, I would like to know precisely which tests were failing because I know for sure I tested extensively on Derby.

Thanks.

gavinking:
I mean, we're supposed to be trying to make _better_ use of static typing. The reappearance of string comparisons in the code looks like a step backwards.

gavinking:
> * Support translating same SqmParameter instance multiple times => Needed for e.g. Derby to emulate the OVERLAY function

This is great. Has been needed for a long time!

beikov:
Hey Gavin,

> Could you please give more details and explanation for these changes. It's really difficult to see the motivation behind some of them, even though some I can guess at.

I tried to summarize the motivation in the description. The main motivation is to fix tests that are failing right now. If you have specific doubts or questions I'm happy to answer them. Some of the tests didn't fail previously because there was no data in the tables which would trigger the failures. I fixed these tests. The queries were mostly syntactically ok, but when certain functions were applied to data, they produced errors. The overlay function for example rendered something along the lines of `..||'overlayed string'||substr('abc', 4)` but that failed because `4` is out of bounds for `substr` and Derby doesn't like that (other DBs produce an empty string). Due to that, I had to introduce a strict `substr` mode where the `substr` expression is wrapped in a `case when` to avoid the error.

> For example, I want to see a proper argument made for changing the casting logic to something stringly-typed. I'm not sure that's a good change at all.
> I mean, we're supposed to be trying to make better use of static typing. The reappearance of string comparisons in the code looks like a step backwards.

Currently, in `CastType#from`, the enum for a java class is determined based on `==` or `equals`: https://github.com/hibernate/hibernate-orm/pull/3695/files#diff-dc83ddc1ad1003ce54d39a6aefc357abe497a23f1aabb6957a4913e806e9117cL55

The if-else cascade has linear time complexity so that's one point I wanted to improve. Using the class name in a string switch reduces the complexity to constant time.

Right now, the code determines a `CastType` based on the linear check and then just passes `CastType` to the `castPattern` method, which is not enough information, as Derby has some very specific requirements(it needs to do some intermediate casts for certain SQL types. Also see the links I added to the comments for more information). The decision to use the class names in a string switch is based on the fact that this is a constant time lookup. We could of course use some kind of indirection through a hash-map if you like, but I don't think it gets much better. I'd also like to avoid using an enum as that wouldn't be extendible for users.

> Also, with respect to the Derby-related changes, I would like to know precisely which tests were failing because I know for sure I tested extensively on Derby.

You can take a look at the test runs on GH actions for derby on wip/6.0 to see the failures: https://github.com/hibernate/hibernate-orm/suites/1830358936/artifacts/35333168

Compare that to the GH actions run for this PR and you will see that I fixed all function and type-size related failures for Derby.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -245,7 +249,14 @@ public String extractPattern(TemporalUnit unit) {
 			case DAY_OF_YEAR:
 				return "({fn timestampdiff(sql_tsi_day, date(char(year(?2),4)||'-01-01'),?2)}+1)";
 			case DAY_OF_WEEK:
-				return "(mod({fn timestampdiff(sql_tsi_day, {d '2000-01-01'}, ?2)},7)+1)";
+				// Use the approach as outlined here: https://stackoverflow.com/questions/36357013/day-of-week-from-seconds-since-epoch
+				return "(mod(mod({fn timestampdiff(sql_tsi_day, {d '1970-01-01'}, ?2)}+4,7)+7,7)+1)";
```
beikov, line 253: A test in `FunctionTests` failed because it produced the week number `21` instead of `22` I think.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -245,7 +249,14 @@ public String extractPattern(TemporalUnit unit) {
...
+			case WEEK:
+				// Use the approach as outlined here: https://www.sqlservercentral.com/articles/a-simple-formula-to-calculate-the-iso-week-number
+				// In SQL Server terms this is (DATEPART(dy,DATEADD(dd,DATEDIFF(dd,'17530101',@SomeDate)/7*7,'17530104'))+6)/7
+				return "(({fn timestampdiff(sql_tsi_day, date(char(year(?2),4)||'-01-01'),{fn timestampadd(sql_tsi_day, {fn timestampdiff(sql_tsi_day, {d '1753-01-01'}, ?2)}/7*7, {d '1753-01-04'})})}+7)/7)";
```
beikov, line 257: Since we had a test that needed the `WEEK` extraction, I looked into this topic and found the article I mentioned in the comment. The implementation/emulation works great, so I thought why not translate this for the Derby dialect.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -245,7 +249,14 @@ public String extractPattern(TemporalUnit unit) {
...
+			case QUARTER:
+				return "((month(?2)+2)/3)";
```
beikov, line 259: Not sure if we had a test for that, but I wanted to add it for completeness sake.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
beikov: Good I looked into this again. Looks like I can now enable this test which failed before: https://github.com/hibernate/hibernate-orm/blob/2f1c7f60f7faea7d9947761cfc3b4ee97090bb05/hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java#L824

gavinking:
> The if-else cascade has linear time complexity so that's one point I wanted to improve. Using the class name in a string switch reduces the complexity to constant time.

To begin with, I really doubt that we need to be concerning ourselves with micro-optimizing the performance of `CastType.from()`. I doubt it's ever called in any sort of hot path, and it should be super-fast, being at worst 13 `==` tests.

The constant integer 13 does not have linear time complexity. It's _O(1)_.

And it's not clear to me that a string switch is necessarily faster, since in principle it can involve computing a hash code.

> which is not enough information, as Derby has some very specific requirements(it needs to do some intermediate casts for certain SQL types. Also see the links I added to the comments for more information).

I don't think it's a good idea to make our code worse and less typesafe just because Derby is rubbish. Derby is not at all a priority: if we can make it work, great, but if it requires doing horrible things it's just as legitimate to leave it unsupported on Derby.

Anyway, I would request that you roll back all the changes that have introduced long lists of string switches in the `Dialects`s. If performance is really of concern here, then we can discuss that, and figure out if there's a better solution. But I doubt that there is a real performance problem here.

gavinking:
By the way, if you really think `Class.equals()` is slow (I don't think it is, but I can't be bothered testing) then you could much more easily change `CastType.from()` to do the string-based `switch` and leave all the `Dialect` alone.

beikov:
> The constant integer 13 does not have linear time complexity. It's O(1).
> By the way, if you really think Class.equals() is slow (I don't think it is, but I can't be bothered testing) then you could much more easily change CastType.from() to do the string-based switch and leave all the Dialect alone.

Right, but we both know that the N by which the complexity rises is the amount of types, and if we add more types it becomes more expensive. It's probably still not worth bothering especially since this is only part of the cast function.

> And it's not clear to me that a string switch is necessarily faster, since in principle it can involve computing a hash code.

The hash code in a string is usually cached.

Anyway, performance isn't my main concern here, but the fact that it's not flexible enough (need to access the Java Class or JDBC type code), so the change to passing `JdbcMapping` instead of `CastType` is IMO a good idea to get more flexibility. I hope you agree with that part.

> I don't think it's a good idea to make our code worse and less typesafe just because Derby is rubbish. Derby is not at all a priority: if we can make it work on Derby, great, but if it requires doing horrible things it's just as legitimate to leave it unsupported on Derby.

I can of introduce `CastType.from()` calls into the dialect code to transform the Java Class to an enum, to then do a switch on enum, but the more general problem with the enum IMO is that it isn't extendible. Imagine someone introduces a funky Java type for e.g. `String` and suddenly casts don't work anymore because we don't handle this new type. How could a user extend this to add support for that if all he get's passed in the dialect is `CastType.UNKNOWN` because we currently use the Java Class type for determining the `CastType`?

gavinking:
Well when I thought about this (and it was a while ago, so I certainly don't recall all my reasoning at this point) it seemed pretty clear to me that this was something that simply wasn't likely to be extensible, and unless I'm very much mistaken, your changes don't in fact make it extensible. I just can't cast an arbitrary database column to an arbitrary Java class, and I don't think you've fixed that. If that were an actual requirement, it seems to me that we would need to do a _whole lot_ more work.

gavinking:
Actually, if you see my comment in `CastType`, I think the original idea was to use the database column type in `CastType.from()`, which at the time I did this work was impossible, but now looks possible.

So if I'm remembering all this correctly, really the `switch` in `CastType.from()` should probably have been rewritten to use the `SqlTypeDescriptor.getSqlType()`.

And note: the list of SQL types is bounded. It's just not extensible.

gavinking:
Actually now I'm struggling to see why we're not just using JDBC type codes from `Types` instead of the  `CastType`  enum. I wonder if it's just because when I wrote this code I didn't have a way to reliably obtain the `JdbcMapping`.

It seems very clear that `CastType`a represent SQL types, not Java types, not Hibernate types.

My guess is that I added that class because I just didn't have a way to get hold of the JDBC type code.

So if we're going to do anything, it should be to change all those `switch`es to use the JDBC type codes, it seems to me.

beikov:
The JDBC type codes are just that, numeric codes which refer to common JDBC/SQL types, but there are more SQL types than that. I think a UserType can return arbitrary type codes, so this is where the extensibility comes from as well.

Generally, I agree that most switches could/should probably be changed to work on JDBC type codes instead, but I didn't want to do such a invasive change here yet, especially because we would have to treat all kinds of string types (char, varchar, ..) like `String`. Note though, that the Java type still plays a role here e.g. in the cast from Boolean to String we might have to use a `case when` if the base type is a numeric type.

The use of the class name makes the new implementation "compatible" to the old one without much change. I'm open to changing this to use the JDBC type codes though if you like that more :)

gavinking:
Yeah, digging deeper, I think it would probably be safe to replace the use of the `CastTarget` enum values with type codes from `java.sql.Types`.

But it's still not going to be at all extensible. To see why, look at `TypeConfiguration.resolveCastTargetType()`.

gavinking:
>  I think a UserType can return arbitrary type codes, so this is where the extensibility comes from as well.

But you can't cast to a `UserType`. What would the syntax for that even look like??

beikov:
This is not specifically about casting **to** a UserType, but about casting a value **of** a UserType **to** some other type.

I'll take a deeper look tomorrow though. Thanks for the feedback. I'll adapt the code to use the JDBC type codes.

gavinking:
> Generally, I agree that most switches could/should probably be changed to work on JDBC type codes instead

I'm *pretty sure* that this is what I had originally wanted to do.

> but I didn't want to do such a invasive change here yet, 

Well the change you made is just as invasive and much uglier.

> especially because we would have to treat all kinds of string types (char, varchar, ..) like String. 

I don't understand. That's what we already do. And I don't see why we would not want to keep doing it.

> Note though, that the Java type still plays a role here e.g. in the cast from Boolean to String we might have to use a case when if the base type is a numeric type.

What I'm saying is that you would get two JDBC type codes in `castPattern()`. I mean, it would be:

    public String castPattern(int from, int to)

I don't *love* `int` here, but we do generally pass type codes around in other places.

gavinking:
> This is not specifically about casting to a UserType, but about casting a value of a UserType to some other type.

Well that's perfectly fine. That's the `from` parameter.

beikov:
> I don't understand. That's what we already do. And I don't see why we would not want to keep doing it.

The point was just, that the PR would have gotten bigger/uglier if I added all the case labels for the different "String" JDBC type codes, but in the long run, IMO this is the right thing to do.

> `public String castPattern(int from, int to)`

This way, we can't support `cast(booleanAttribute as String)` on DBs that use a JDBC type code other than `Types.BOOLEAN`. To support this, we need to know the source Java Type, which is boolean, as well. Thus we need `public String castPattern(JdbcMapping from, JdbcMapping to)`. Do you agree?

> Well that's perfectly fine. That's the from parameter.

Right, and with the JDBC type code being available in `castPattern`, a user can customize the casting.

gavinking:
Just to summarize before I go to sleep, my objection to changing these `switch`es to use Java class names is:

1. they model casts between database types, not Java types; you can't cast between Java types in SQL (`CastType` models a database type, at least approximately), and
2. while it's perfectly OK to have `switch`es with string literal cases (especially if you're using the `switch` to translate the string literals to something statically-typed) it's preferable to localize it in *one place*, not scatter the same stringly-typed names all over.

gavinking:
> This way, we can't support cast(booleanAttribute as String) on DBs that use a JDBC type code other than `Types.BOOLEAN`. 

Unless I'm mistaken, Hibernate always uses `Types.BOOLEAN`, even when the database column type is `BIT`. That's why we have code like:

    registerColumnType( Types.BOOLEAN, "bit" );

(Taken from `MySQLDialect`.)

>  Thus we need `public String castPattern(JdbcMapping from, JdbcMapping to)`. Do you agree?

I don't see why. We shouldn't be looking at the `JavaTypeDescriptor` to do a cast in SQL. And the SQLTypeDescriptor really doesn't have anything beyond just the JDBC type code.

gavinking:
> Unless I'm mistaken, Hibernate always uses `Types.BOOLEAN`, even when the database column type is `BIT`.

Ah well, perhaps there's some inconsistency there. Perhaps on Oracle and Sybase you *do* get `Types.BIT`.

So perhaps in `OracleDialect.castPattern()` it also needs to treat `Types.BIT` like a boolean. Anyhow, that seems like a pretty minor issue.

beikov:
But BIT could also be mapped to short, integer or long ;)

Let's have a chat tomorrow if you want. I'm pretty sure we need the whole JdbcMapping though. The JDBC type code is just the one returned by the SqlTypeDescriptor which is concerned only about reading a value from a JDBC result set.

gavinking:
> But BIT could also be mapped to short, integer or long ;)

Right, sure, but that's not a problem at all, AFAICS.

The SQL type code used internally in Hibernate `Dialect`s does not directly represent the database column type. That's what I tried to explain above. The `Dialect` maps the SQL type `Boolean` to the column type `BIT` for the purpose of DDL generation because that's what's in the `TypeNames`, but it goes on using `BooleanTypeDescriptor`/`Types.BOOLEAN`.

The _only_ case I was worried about was casting `Boolean` to `String` on a database which does something like this in `OracleDialect`:

	protected SqlTypeDescriptor getSqlTypeDescriptorOverride(int sqlCode) {
		return sqlCode == Types.BOOLEAN
				? BitTypeDescriptor.INSTANCE
				: super.getSqlTypeDescriptorOverride( sqlCode );
	}

But I just checked and this override actually doesn't seem to affect anything we care about. 

(TBH I have never been able to figure out precisely what  `getSqlTypeDescriptorOverride()` actually *does* affect, and I strongly suspect it's something we should just remove or completely rework, because it doesn't seem to be implemented at all consistently across dialects. Indeed, I strongly suspect that people keep adding useless entries in `getSqlTypeDescriptorOverride()` because they think it does something it actually doesn't do.)

Anyway, the point is I think the problem you're imagining just can't occur because Hibernate doesn't use the SQL type codes in the way you imagine it does.

gavinking:
> The point was just, that the PR would have gotten bigger/uglier if I added all the case labels for the different "String" JDBC type codes, but in the long run, IMO this is the right thing to do.

Well look, if what you're worried about is having `CharTypeDescriptor`, `NVarcharTypeDescriptor`, `NCharTypeDescriptor`, and `LongVarcharTypeDescriptor` all work correctly, then that's a rather strong argument for retaining my `CastType` enumeration and leaving things essentially alone.

Because for whatever other flaws it has, what `CastType` certainly *does* do is it abstracts over all these uninteresting variations with the single member `STRING`. And hell, perhaps *that's* another reason I introduced it. (I simply don't recall.)

Now, what you *could* consider doing is actually giving each `SqlTypeDescripter` a `CastType` (i.e. add `getCastType()` to SqlTypeDescripter`.) and remove `CastType.from()`.

That feels like probably the best option to me now.

gavinking:
P.S. I should also mention that the current implementation has a bug where casting `TrueFalseType` or `YesNoType` to `String` doesn't work (and your proposed implementation doesn't fix that, because it still only looks at the Java type).

But this bug would I guess be fixed by migrating to either JDBC type codes, or to using the `CastType` belonging to a `SqlTypeDescriptor`.

beikov:
If you look into e.g. https://github.com/hibernate/hibernate-orm/pull/3695/files#diff-b569b438f401c092432d54437122c7f260fc7bc5f3d57939411132fcdcf5e63eR708 you will see that code currently simply assumes that booleans are represented as numerics/strings for certain DB versions. It's probably a good idea to use the JDBC type code for this decision as well though. I'll add tests for the various variants as well as implement it base on the JDBC type code and we can discuss that then.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -3441,54 +3437,73 @@ public String formatBinaryLiteral(byte[] bytes) {
...
+				Integer precision,
+				Integer scale,
+				Long length) {
 			final Size size = new Size();
 			int jdbcTypeCode = sqlType.getJdbcTypeCode();
 
 			switch (jdbcTypeCode) {
 				case Types.BIT:
+					// BIT has an implicit length of 1
+					// If a database like e.g. MySQL supports values greater than 1, it can override this
```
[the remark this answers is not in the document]
beikov: Good catch. I guess I was confused that some DBs only support the value 1, but I'll adapt that.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -3441,54 +3437,73 @@ public String formatBinaryLiteral(byte[] bytes) {
...
 				case Types.LONGVARCHAR:
 				case Types.LONGNVARCHAR:
 				case Types.LONGVARBINARY:
 					size.setLength( javaType.getLongSqlLength() );
 					break;
 				case Types.FLOAT:
+					if ( precision != null && precision < 1 ) {
+						// Use the default precision
+						precision = null;
+					}
```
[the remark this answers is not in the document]
beikov: When the JPA `@Column` annotation is used, which has a default precision of 0, we receive a precision of 0 here, but e.g. a `float(0)` is illegal.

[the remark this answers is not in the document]
gavinking: OK but this seems like the wrong layer to handle interpretation of `@Column`. Surely it should be the job of the `Binder` to interpret 0 as `null` when reading the annotation.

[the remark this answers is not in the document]
beikov: For precision, you might be right that we should interpret 0 as null to get the default precision, but for `scale` it's more complicated as you can read in the discussion: https://hibernate.zulipchat.com/login/#narrow/stream/132094-hibernate-orm-dev/topic/Minimum.20sizes.20for.20types

My main argument for interpreting the precision here is to have a single place to interpret all size related values.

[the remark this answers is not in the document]
gavinking: Honestly `scale` I'm simply not sure about. I don't think there's a clear intent in the JPA spec, and I don't recall us ever thinking terribly hard about this. I think we just accepted that there were no nulls in annotations so defaulting to `0` was the best we could possibly do.

*Probably* `scale=0` should be taken literally.

But that's not the case with precision, since `precision=0` simply makes no sense.

Unfortunate, but anyway with the `getDefaultSqlPrecision()` and `getDefaultSqlScale()` stuff I implemented, explicitly specifying precision and scale less-often needed.

hibernate-core/src/main/java/org/hibernate/dialect/AbstractTransactSQLDialect.java
```diff
@@ -43,7 +43,7 @@ public AbstractTransactSQLDialect() {
 		registerColumnType( Types.BOOLEAN, "bit" );
 
 		//SQL Server/Sybase 'bit' type is always exactly one bit
-		registerColumnType( Types.BIT, "smallint" );
+		registerColumnType( Types.BIT, "bit" );
```
gavinking, line 46: This change is incorrect, which is why there is a comment there. The `bit` type in these databases is a literal bit and can't be null. That's why it's set to `smallint`.

gavinking: http://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.infocenter.dc36271.1550/html/blocks/X34468.htm

Also they can't have indexes which is a problem if you want to put 'em in a PK.

beikov: Ok, I just looked into the SQL Server documentation and there it states null is possible: https://docs.microsoft.com/en-US/sql/t-sql/data-types/bit-transact-sql?view=sql-server-ver15

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -270,39 +288,35 @@ public String translateExtractField(TemporalUnit unit) {
 	 * type, but it doesn't know how to cast to it. Worse,
 	 * Derby makes us use the {@code double()} function to
 	 * cast things to its floating point types.
+	 * @param from
+	 * @param to
```
gavinking, line 292: if you're going to add these, you need to add descriptive text.

hibernate-core/src/main/java/org/hibernate/dialect/AbstractTransactSQLDialect.java
beikov: Does it make sense to use `bit` then for boolean if it doesn't support null on Sybase?

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
beikov: Thanks, I'll remove that.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -556,6 +555,64 @@ public String extractPattern(TemporalUnit unit) {
...
+					case INTEGER:
+					case INTEGER_BOOLEAN:
+					case LONG:
+						return "(?1<>0)";
+					case YN_BOOLEAN:
+						return "(?1<>'N')";
+					case TF_BOOLEAN:
+						return "(?1<>'F')";
+				}
+		}
```
gavinking, line 615: The only reservation I have here is that I guess there are some databases which will already natively do the right thing in some of these casts without the need for verbose `case`/`when` constructs.

gavinking: But it's _very_ nice that we can now cast between all the various different ways to represent booleans. That's actually rather useful.

beikov: I thought about that as well, but if this turns out to be an issue, it could be optimized later. I wanted to keep it as simple as possible :)

gavinking: Well, actually we can only cast _from_ then to regular `Boolean`, because there's actually no way to specify things like `YesNoBooleanType` in the `cast()` function. So this is actually a bit less useful than it looks, unless we add a syntax for doing that.

So perhaps we should let you write something like `cast(foo as Boolean(T,F))` or something like that.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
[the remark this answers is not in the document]
gavinking: By the way, @beikov, have you checked that none of these changes break my work on defaulting the right precision/scale/length for a type?

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
```diff
@@ -216,27 +215,16 @@ public SqlAstTranslatorFactory getSqlAstTranslatorFactory() {
 	/**
 	 * Firebird 2.5 doesn't have a real {@link java.sql.Types#BOOLEAN}
 	 * type, so...
+	 * @param from
+	 * @param to
```
gavinking, line 219: Also missing the text description.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
```diff
@@ -216,27 +215,16 @@ public SqlAstTranslatorFactory getSqlAstTranslatorFactory() {
...
-			}
-			if ( to==STRING && from==BOOLEAN ) {
-				return "trim(decode(?1,0,'false','true'))";
-			}
-		}
-		else {
-			if ( from==BOOLEAN
-					&& (to==LONG || to==INTEGER)) {
-				return "decode(?1,false,0,true,1)";
-			}
```
gavinking, line 239: It's at least a little bit sad that we're now using `case`/`when` for this on databases which have the much less verbose `decode()`.

hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
```diff
@@ -117,6 +117,8 @@ public H2Dialect(int version) {
 			this.sequenceInformationExtractor = SequenceInformationExtractorNoOpImpl.INSTANCE;
 			this.querySequenceString = null;
 		}
+		// H2 doesn't allow a length for BIT
+		registerColumnType( Types.BIT, "bit" );
```
gavinking, line 121: I don't think this change is a good one. H2 doesn't allow a length in `bit` because it's a synonym for `boolean`. So if you're trying to use a `bit(10)` on H2, then that should give you an *error*.

Making this change means you would get the error later: not when the schema is exported, but when you try to write to the column. I think it's better to get the error earlier than later, so I would not make this change.

hibernate-core/src/main/java/org/hibernate/dialect/HSQLDialect.java
```diff
@@ -223,15 +219,6 @@ public SqlAstTranslatorFactory getSqlAstTranslatorFactory() {
 		};
 	}
 
-	@Override
-	public String castPattern(CastType from, CastType to) {
-		if ( from== BOOLEAN
-				&& (to== INTEGER || to== LONG)) {
-			return "casewhen(?1,1,0)";
```
gavinking, line 230: Again, it's kinda sad to lose this nice terse SQL syntax, and replace it with a verbose `case`/`when`.

hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
```diff
@@ -334,31 +360,6 @@ public String extractPattern(TemporalUnit unit) {
...
-					case STRING:
-//						return "if(?1 rlike '^(t|f|true|false)$', ?1 like 't%', null)";
-						return "if(lower(?1) in('t','f','true','false'), ?1 like 't%', null)";
-					case LONG:
-					case INTEGER:
-						return "(?1<>0)";
-				}
-			case STRING:
-				if (from == BOOLEAN) {
-					return "if(?1,'true','false')";
```
gavinking, line 355: same comment here: this `if()` function is much nicer than a `case`\`when`

hibernate-core/src/main/java/org/hibernate/dialect/AbstractTransactSQLDialect.java
gavinking: Wait, I'm confused. How do we have:

    registerColumnType( Types.BOOLEAN, "bit" );

I thought that was wrong.

gavinking: I right, I see. Sorry, my explanation for that line is wrong. What's wrong with mapping `Types.BIT` to Transact SQL `bit` is that on Transact SQL the `bit` type is always exactly one bit. 

But `Dialect` already has a registration for that case:

    registerColumnType( Types.BIT, 1, "bit" );

This line:

    registerColumnType( Types.BIT, "smallint" );

is just overriding the mapping for length>1. Honestly I'm not sure if it's even worth having it there and perhaps we should simply delete that line. But whatever, it certainly doesn't make sense to map length>1 to `bit` in Transact SQL.

(It's `SybaseASEDialect` that handles the issue with mapping `Types.BOOLEAN` to Transact SQL `bit`, since it seems I misremembered and that is an issue only on ASE, not on SQL Server or Sybase Anywhere.)

hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
beikov: Passing null would produce the else-result, so this wouldn't be compatible with other implementations. Thus it needs case when.

hibernate-core/src/main/java/org/hibernate/dialect/HSQLDialect.java
beikov: Passing null would produce the else-result, so this wouldn't be compatible with other implementations. Thus it needs case when.

hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
beikov: I'll remove that.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
beikov: I'll readd that as good as I can

hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
gavinking: Note, this is kinda similar to the situation with the Transact SQL dialect above.

hibernate-core/src/main/java/org/hibernate/dialect/AbstractTransactSQLDialect.java
beikov: Two lines above. That was already there, but I guess this might have been a mistake then.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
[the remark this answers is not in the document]
beikov: Break "how" and which work are you referring to? If there are tests for that, this is covered.

I'll change the binder code to handle precision 0 the same as `null` then.

hibernate-core/src/main/java/org/hibernate/dialect/AbstractTransactSQLDialect.java
gavinking: So there's actually 5 different `Dialect`s which map `Types.BIT` to `smallint` for length>1. This is something we inherited from H5 it seems, and it's inconsistent (H2 doesn't do it, for example), and honestly it's not clear to me that it even makes all that much sense.

So I dunno, delete it or leave it be, I don't much care. The main danger is that it's confusing to people maintaining the `Dialect`s.

hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
gavinking: To be consistent with other dialects we could map `bit` with length>1 to `smallint` if we decide that makes sense.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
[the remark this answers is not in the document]
gavinking: Well I mean all the work I did to make sure that we generate DDL with a good precision/scale by default for each Hibernate type. (In Hibernate 5 the precision/scale often defaulted to rubbish.)

I don't think there's tests, because I didn't have a good way to test the types in generated DDL.

hibernate-core/src/main/java/org/hibernate/dialect/AbstractTransactSQLDialect.java
```diff
@@ -40,11 +40,6 @@
 	public AbstractTransactSQLDialect() {
 		super();
 
-		registerColumnType( Types.BOOLEAN, "bit" );
```
gavinking, line 43: Removing this line breaks `SQLServerDialect` and `SybaseAnywhereDialect`, I believe.

Really I wish you wouldn't mess too much with these type mappings because I put a *lot* of work into fixing them, and the truth is I just don't remember all the details of every mapping off the top of my head.

If you want to change type mappings, please do it in *small* pull requests, please don't roll changes like that into a big "omnibus" PR.

hibernate-core/src/main/java/org/hibernate/dialect/HSQLDialect.java
gavinking: Actually HSQL has a `decode()` function, so we could use that, like in Oracle.

hibernate-core/src/main/java/org/hibernate/dialect/SybaseDialect.java
```diff
@@ -49,6 +49,10 @@ public SybaseDialect(int version) {
 		this.version = version;
 		//Sybase ASE didn't introduce bigint until version 15.0
 		registerColumnType( Types.BIGINT, "numeric(19,0)" );
+
+		//Sybase 'bit' type is always exactly one bit i.e. doesn't support null
+		registerColumnType( Types.BIT, "smallint" );
+		registerColumnType( Types.BOOLEAN, "smallint" );
```
gavinking, line 55: I *think*, but I'm not 100% sure, that we can use `bit` for `BOOLEAN` on Sybase Anywhere. It looks like it's only ASE where we can't use it.

hibernate-core/src/main/java/org/hibernate/dialect/AbstractTransactSQLDialect.java
gavinking: Oh, I see you also made changes to the subclasses. But it's still a change to behavior on Sybase Anywhere that I don't think we want. Whatever, such a change doesn't belong in *this* pull request.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
```diff
@@ -219,24 +222,87 @@ public SqlAstTranslatorFactory getSqlAstTranslatorFactory() {
...
+					case INTEGER_BOOLEAN:
+						return "decode(?1,null,null,0,'false','true')";
+					case TF_BOOLEAN:
+						return "decode(?1,'T','true','F','false',null)";
+					case YN_BOOLEAN:
+						return "decode(?1,'Y','true','N','false',null)";
+					case BOOLEAN:
+						return "decode(?1,true,'true',false,'false',null)";
+				}
+				break;
```
gavinking, line 305: I suppose this is just a copy/paste job of what's in `OracleDialect`. If so, it should be moved to a static method somewhere, to eliminate the duplication, and also to reuse it in HSQLDB.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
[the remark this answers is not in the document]
beikov: Since I reverted the changes regarding the minimum size check, this is equivalent to the previous behavior now.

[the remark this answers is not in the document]
gavinking: OK, 👍 thanks for confirming

hibernate-core/src/main/java/org/hibernate/dialect/SybaseDialect.java
beikov: You are right, SQL Anywhere supports marking the BIT type as `NULL`. I'll fix that.

hibernate-core/src/main/java/org/hibernate/dialect/HSQLDialect.java
beikov: You are right, thanks. I'll add that.

hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
beikov: I'd like to remove all those re-mappings since the BitTypeDescriptor always expects true/false i.e. it even uses `getBoolean`/`setBoolean` on the JDBC driver. Passing through the length if the DB supports it is ok I guess, but otherwise I'd just omit it. WDYT?

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
```diff
@@ -79,9 +79,9 @@ queryExpression
 	;
 
 queryGroup
-	: querySpec                                            										# QuerySpecQueryGroup
-	| LEFT_PAREN queryGroup RIGHT_PAREN															# NestedQueryGroup
-	| queryGroup queryOrder? (setOperator (querySpec | LEFT_PAREN queryGroup RIGHT_PAREN))+		# SetQueryGroup
+	: querySpec                                            												# QuerySpecQueryGroup
+	| LEFT_PAREN queryGroup queryOrder? RIGHT_PAREN														# NestedQueryGroup
+	| queryGroup queryOrder? (setOperator (querySpec | LEFT_PAREN queryGroup queryOrder? RIGHT_PAREN))+	# SetQueryGroup
```
gavinking, line 84: This rule looks like it could do with a refactoring, to remove the duplication of `queryGroup queryOrder?` and `LEFT_PAREN queryGroup queryOrder? RIGHT_PAREN`.

Split out a separate rule.

gavinking (changes requested):

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
beikov: Fixed

hibernate-core/src/main/java/org/hibernate/dialect/AbstractTransactSQLDialect.java
beikov: Ok I removed these changes form this PR and will create a separate one or discuss on your PR

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
```diff
@@ -78,10 +78,14 @@ queryExpression
...
+	: LEFT_PAREN queryGroup queryOrder? RIGHT_PAREN
+	;
+
 queryGroup
-	: querySpec                                            										# QuerySpecQueryGroup
-	| LEFT_PAREN queryGroup RIGHT_PAREN															# NestedQueryGroup
-	| queryGroup queryOrder? (setOperator (querySpec | LEFT_PAREN queryGroup RIGHT_PAREN))+		# SetQueryGroup
+	: querySpec                                            						# QuerySpecQueryGroup
+	| nestedQueryExpression														# NestedQueryGroup
+	| queryGroup queryOrder? (setOperator (querySpec | nestedQueryExpression))+	# SetQueryGroup
```
gavinking, line 88: Ok that's better. so now that I understand what's going on here, I wonder if why it's ambiguous without the parents around the one with the order by.

Got an example of a query which would be ambiguous?

beikov: The last order by clause should always refer to the whole query group i.e. `select 1 union all select 1 order by 1` should result in `QueryGroup({QuerySpec, QuerySpec}, orderBy = {1})` and if you want to order only a query spec, you need parenthesis.

gavinking: Well is that what the ANSI SQL spec says?

Because that seems like a rubbish grammar to me.

It would be better to make you write 

    (select 1 union all select 1) order by 1

or even 

    select * from (select 1 union all select 1) order by 1

fo that case.

Because all the other clauses bind to the most recent `select`. It's not normal that there should be a special use for `order by`.

hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
```diff
@@ -1103,16 +1146,16 @@ public static Replacer datetimeFormat(String format, boolean useFm) {
...
 
 				//minute
 				.replace("mm", "MI")
-				.replace("m", fm + "MI")
+				.replace("m", fm + "MI" + fm)
 
 				//second
 				.replace("ss", "SS")
-				.replace("s", fm + "SS")
+				.replace("s", fm + "SS" + fm)
```
gavinking, line 1158: I don't understand the reason for this change. According to the Oracle documentation, the leading `fm` suppresses trailing blanks as well as leading zeros. Why did you add this?

Also, note that this function is called by other dialects; we should move it somewhere else.

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
beikov: I don't have a SQL standard other than the one from 92 which was published publicly. I used the Oracle and PostgreSQL, syntax which also seems supported for TSQL:

* https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/The-UNION-ALL-INTERSECT-MINUS-Operators.html#GUID-B64FE747-586E-4513-945F-80CB197125EE
* https://www.postgresql.org/docs/current/queries-order.html
* https://github.com/antlr/grammars-v4/blob/master/sql/tsql/TSqlParser.g4#L3002

I agree that wrapping the query expression is "cleaner" and I want to support that eventually. But right now, we don't allow query expressions yet in the FROM clause

hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
beikov: According to the Oracle documentation this is a "trigger" to switch the behavior and it can occur multiple times. PostgreSQL seems to reset this trigger when encountering a "literal", just as I would have expected from Oracle, but apparently Oracle keeps this mode, so it needs to be reset, otherwise the rest of the pattern will use the modifier as well.

hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
```diff
@@ -163,6 +165,25 @@ else if( "myisam".equals( storageEngine.toLowerCase() ) ) {
...
+					Integer precision,
+					Integer scale,
+					Long length) {
+				final int jdbcTypeCode = sqlType.getSqlType();
+				switch ( jdbcTypeCode ) {
+					case Types.BIT:
+						// MySQL allows BIT with a length up to 64
+						if ( length != null ) {
+							return Size.length( Math.min( Math.max( length, 1 ), 64 ) );
+						}
```
gavinking, line 182: Do you think this is really necessary? It's not something we do with other types, nor in other dialects.

What's the reason for doing it specifically for this one case?

hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
gavinking: So you're saying you include it to *undo* the effect of `fm`?

But in fact we never wish to turn it off, it seems to me. If it could be one setting at the start of the pattern that would be fine by me. Right?

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
gavinking: Right, but we need to anticipate future evolution of the grammar, and not break people's queries.

why not just support the syntax:

    from (select 1 union all select 1) order by 1

as a special case for now (just especially) for this thing.

hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
beikov: Since MySQL uses the bit type for booleans, and the JPA default when using the `@Column` annotation is 255, this might be relevant, although I think I handled this already in the `Dialect`, so it might be unnecessary to keep this around, although I think if we want to support the bit type, it might make sense to still keep this around.

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
beikov: What is the specific part that you don't like about this change? The parenthesis are useful for other things as well like operation ordering. A query like `... intersect ... union all ...` is not the same as `... intersect (... union all ...)` because set operations are executed from left to right.

gavinking: `order by` should, by nature, have a higher precedence than `union` and `intersect`, no?

I mean, every other clause of a `select` query does, right?

It's an issue of grammar design. When I originally saw that rule, I just new that something was screwy, it smells.

gavinking: We never want to say "oh this construct has such-and-such relative precedence _except_ in this special context". Or "oh in this special case you need parens when you would not expect it".

(That's the Perl way.)

hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
beikov: I'm saying that if you just use `fm` once in Oracle, every following pattern will trim leading zeros even if your desired format should retain them. To avoid this, you have to "switch" again by using the fm modifier again.

beikov: Also see https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements004.htm#SQLRF00216

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
beikov: I'm pretty sure the SQL standard says that order by applies to the query group/block as all databases implement it this way. The reasoning for this is readability in multiple documentations. Here is a reference to the syntax in Oracle: https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/About-Queries-and-Subqueries.html#GUID-DB7521FE-9329-415E-B583-EA4467E990A7

In a query like `select 1 union all select 2 order by 1 desc` the order by clause does not refer to the subquery `select 2` but instead refers to the whole query group/block.

gavinking: Perhaps that's why I hate locals so much, come to think of it: they're a missed opportunity to make a new function.

hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
gavinking: > every following pattern will trim leading zeros even if your desired format should retain them

but in what circumstance do we ever want to retain them. IIRC our `format()` function always has the semantics of no padding, doesn't it?

(now where did I document that??)

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
gavinking: Well even if that's right, it's an objectively bad syntax, and I don't think we should copy it.

gavinking: I mean, consider:

    select foo, sum(bar) from Foo1 union select foo, sum(bar) from Foo2 group by foo order by foo

do you *really* think it's a good idea that `group by` refers to the second `select` but `order by` refers to the whole union?

That's just objectively bad.

*Either* you require people to always write

    (select foo, sum(bar) from Foo1) union (select foo, sum(bar) from Foo2 group by foo) order by foo

i.e. *always* put parens around subqueries in unions, 

*or* you require them to write

    from (select foo, sum(bar) from Foo1 union select foo, sum(bar) from Foo2 group by foo) order by foo

when they want to order the whole union.

beikov: I have only access to SQL:2003 BNF, but ordering and limiting in subqueries was only introduced in SQL:2008 so I can't give you a definitive reference to the SQL Standard for this. Like I wrote before, all databases implement it this way (ordering applies to query group) and I bet the reason for that is that the SQL Standard mandates this.

I understand your point and I know it might not be 100% clear to a reader that doesn't know how this works. I also agree that we should (in addition) support and advocate the `select * from (...) order by ...` syntax, but before doing that we need to discuss support for query expressions in the from clause. Regardless of all of that, I'm pretty sure the SQL standard specifies that the ordering applies to the whole query group which is why I would like to match this behavior in HQL. As a user, I would find it confusing if HQL went a different route than SQL on this regard. To avoid this confusion, we'd have to prohibit ordering for set operation query groups in general and force the user to wrap the query group in `select * from (...)` or `from (...)` for the sake of adding ordering.

I'm already seeing a future user complaining and asking why Hibernate has to deviate from what the SQL Standard specifies in this regard. I'd prefer to be as close to the SQL Standard as possible and this behavior, although not immediately clear to everyone, is clearly specified.

So where do we go from here? I think that the argument "match what the SQL Standard specifies" should be enough to resolve this discussion, but if feel strongly about this, maybe this should be a team discussion/decision?

hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
beikov: AFAICT our `format()` uses the same format like `java.time.format.DateTimeFormatter` and we translate that to the DBMS specific format. To get the semantics right, it is necessary to "reset"/"switch" the modifier by using `fm` again.

gavinking: Sorry, you're right and I'm an idiot. The main reason we're doing this is because we _sometimes_ want to eliminate leading zeros. The purpose isn't the stupid trailing spaces in Oracle. I forgot.

Still not certain we want to add the trailing `fm`s for other dialects apart from Oracle though. I think we should add another boolean toggle to the method signature, so this is something we only do on Oracle.

gavinking: FTR, the only documentation I wrote for this feature (so far) is the Javadoc in `Dialect.datetimeFormat()`.

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
gavinking: >  I can't give you a definitive reference to the SQL Standard for this

Anyway it's beside the point, HQL isn't SQL, and we don't need to slavishly copy SQL's mistakes.

> before doing that we need to discuss support for query expressions in the from clause

Really, I don't see why. I think the correct syntax is from `from`-clause subqueries is pretty obvious, and I don't think there's going to be any debate about it at all. We know what a `from`-clause subquery is going to look like, and we have for decades&mdash;we just didn't implement it yet, because it's not so terribly useful in the context of ORM.

> As a user, I would find it confusing if HQL went a different route than SQL

> As a user, I would find it confusing if HQL went a different route than SQL on this regard.

As a user I find it confusing when languages have confusing semantics.

We should definitely not do objectively bad things just because other people did objectively bad things before us.

Ask yourself: would you be better off today if Java had reproduced more confusing things from the C/C++ languages? Was it good, for example, that Java copied the fall-through semantics of C's `switch`?

> I'm already seeing a future user complaining and asking why Hibernate has to deviate from what the SQL Standard specifies in this regard. 

And I can see _many_ more users, Java developers, who don't know SQL inside-out, being as surprised as I was that `order by` has a different precedence to `group by`.

> So where do we go from here? 

If you really think there's need for further discussion of this issue (I don't, because I know I'm right about this), then the right thing to do is to just drop this feature from this pull request, and get the other stuff that we agree on integrated.

(That's the problem with huge multi-feature pull requests.)

gavinking: > (That's the problem with huge multi-feature pull requests.)

It was actually just *luck* that I caught this before it got pushed into the language. You didn't even mention it in the list of changes!

hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
beikov: Ok, I'll adapt it for Oracle then. It's just that it's not "wrong" to toggle the `fm` mode off by using it again. All DBMS do the right thing. It seems that at least DB2 and PostgreSQL reset the `fm` mode automatically when encountering a literal, but not sure if all DBMS work this way. At the same time, these DBMS also handle a trailing `fm` correctly in the sense that it also toggles the `fm` mode.

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
beikov: > Anyway it's beside the point, HQL isn't SQL, and we don't need to slavishly copy SQL's mistakes.

I don't know if it is a mistake or a deliberate decision, but the SQL standard probably also won't tell us that. In general I  agree that we shouldn't copy mistakes, but I guess this particular case is more complex as it involves not only "taste" but also the question if we want to align with what is known (SQL).

> Really, I don't see why. I think the correct syntax for from from-clause subqueries is pretty obvious, and I don't think there's going to be any debate about it at all. We know what a from-clause subquery is going to look like, and we have for decades—we just didn't implement it yet, because it's not so terribly useful in the context of ORM.

I've seen plenty of requests for it from various users, so I think subqueries in the from clause are very useful. The problem is not necessarily the HQL syntax, but more about how we represent the "type" in the query model.

> Ask yourself: would you be better off today if Java had reproduced more confusing things from the C/C++ languages? Was it good, for example, that Java copied the fall-through semantics of C's switch?

Like I wrote before, we should probably have a discussion about this and decide within the team as there are more aspects to this IMO.

> If you really think there's need for further discussion of this issue (I don't, because I know I'm right about this), then the right thing to do is to just drop this feature from this pull request, and get the other stuff that we agree on integrated.
> (That's the problem with huge multi-feature pull requests.)

I know this PR is too big, and we discussed this on Zulip already. It's not like I wanted this to be this way. I was just trying to fix testsuite issues and one thing lead to another. Future PRs will be more focused and have a corresponding issue.

> It was actually just luck that I caught this before it got pushed into the language. You didn't even mention it in the list of changes!

The feature (ordering of query groups) was there already before this PR. The changes to the grammar I did were to add support for ordering nested query groups as I apparently didn't add that originally.

hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
gavinking: Whatever you decide to do is fine by me. My original objection to this was wrong.

gavinking (approved):
Looks good now!

Thanks so much!!

PR #3696, opened 2021-01-15
HHH-14384 Fix QueryProducer.createNativeQuery documentation

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14384

Minor doc PR, but I think the complaint is valid.


hibernate-core/src/main/java/org/hibernate/query/QueryProducer.java
```diff
@@ -50,7 +50,7 @@
 	 * Create a typed {@link Query} instance for the given HQL/JPQL query string.
 	 *
 	 * @param queryString The HQL/JPQL query
-	 *
+	 * @param resultClass The type of the query result
```
NathanQingyangXu, line 53: The above is to fix a missing Javadoc element along the way, not directly related to the ticket.

beikov:
Thanks!

PR #3697, opened 2021-01-17
fix Github Action on Oracle matrix

NathanQingyangXu:


NathanQingyangXu:
the current Oracle Github CI on master seems working as expected (36 broken testing cases after 3 hours of running). Closed this PR to focus on fixing the 36 broking Oracle testing cases instead.

PR #3698, opened 2021-01-21
Various fixes

dreab8:
- Fix EntitiyInitialization when exists a loaded a proxy
- Fix NPE for InferredBasicVaueResolution
- Fix issue issue with loading by Unique key
- Fix FkDescriptor for Embeddable 
- other minor fixes
- Some code refactoring


hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -133,6 +133,45 @@ public static SelectStatement createSelect(
...
+			int numberOfKeysToLoad,
+			LoadQueryInfluencers loadQueryInfluencers,
+			LockOptions lockOptions,
+			Consumer<JdbcParameter> jdbcParameterConsumer,
+			SessionFactoryImplementor sessionFactory) {
+		final LoaderSelectBuilder process = new LoaderSelectBuilder(
+				sessionFactory,
+				loadable,
+				partsToSelect,
+				Arrays.asList( restrictedPart ),
```
NathanQingyangXu, line 163: Using `Collections.singleton()` would be more natural and readable. JMH benchmark should show its performance advantage as well (though not tried yet).

NathanQingyangXu: The article (https://dzone.com/articles/singleton-list-showdown-collectionssingletonlist-v) shows the `Arrays.asList()` is twice as slow than `Collections.singletonList()`.

hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
```diff
@@ -987,6 +987,9 @@ public Object immediateLoad(String entityName, Object id) throws HibernateExcept
 			event.setResult( null );
 			loadEvent = event;
 		}
+		if ( result instanceof HibernateProxy ) {
+			return ( (HibernateProxy) result ).getHibernateLazyInitializer().getImplementation();
```
beikov, line 991: Doesn't this break the identity guarantee, that entities with the same PK have the same identity, required by JPA? IMO we should trigger lazy loading here, but still return the proxy.

hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
```diff
@@ -743,7 +743,9 @@ public Object narrowProxy(Object proxy, EntityPersister persister, EntityKey key
 
 			if ( object != null ) {
 				final LazyInitializer li = ( (HibernateProxy) proxy ).getHibernateLazyInitializer();
-				li.setImplementation( object );
+				if ( li.isUninitialized() ) {
```
beikov, line 746: Should it even be possible that this is already initialized? Maybe this should be an assert and we should fix the callers such that the state is set on the existing entity object rather than creating a new one and ignore it here?

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/EmbeddableMappingType.java
```diff
@@ -528,7 +528,21 @@ public int forEachJdbcValue(
 		for ( int i = 0; i < attributes.size(); i++ ) {
 			final AttributeMapping attributeMapping = attributes.get( i );
 			final Object o = attributeMapping.getPropertyAccess().getGetter().get( value );
-			span += attributeMapping.forEachJdbcValue( o, clause, span + offset, consumer, session );
+			if ( attributeMapping instanceof ToOneAttributeMapping ) {
```
beikov, line 531: I recently had a SO exception due to something similar. What do you think about changing the implementation of `ToOneAttributeMapping` to traverse only the foreign key by default? I don't know how join fetching is currently implemented so I don't know how such a change would affect that, but I have a feeling that such missing instanceof checks might be hidden problems that we will yet discover.

hibernate-core/src/test/java/org/hibernate/orm/test/bidi/AuctionTest.java
```diff
@@ -0,0 +1,147 @@
...
+ */
+@DomainModel(
+		xmlMappings = "org/hibernate/orm/test/bidi/Auction.hbm.xml"
+)
+@SessionFactory
+public class AuctionTest {
+
+	@Test
+	@SuppressWarnings({ "unchecked" })
+	@SkipForDialect(dialectClass = PostgreSQL81Dialect.class, matchSubTypes = true, reason = "doesn't like boolean=1")
```
beikov, line 39:
```suggestion
	@SkipForDialect(dialectClass = PostgreSQLDialect.class, matchSubTypes = true, reason = "doesn't like boolean=1")
```

beikov (changes requested):
Looks good so far, but I'm a bit unsure about the identity guarantee that JPA mandates

hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
dreab8: it is due to the change in the Initializers logic, now in case of existing proxy they always return the proxy initialised, in 5.x  the entity was loaded and then based on the case a corresponding proxy was searched and initialised.

dreab8: I think the result is the same but not 100% sure

hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
dreab8: ok based on the changes to Initializers the proxy should be always initialised, I add an assertion here

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
dreab8: Thanks @NathanQingyangXu , i'm going to use your suggestion

beikov (approved):

PR #3699, opened 2021-01-21
HHH-14413 fix issue that EntityUpdateAction increments version despite veto on update

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14413

Straightforward implementation and testing.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/event/PreUpdateEventListenerVetoTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/event/PreUpdateEventListenerVetoTest.java
@@ -1,0 +1,84 @@
+package org.hibernate.event;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Version;
+
+import org.hibernate.event.service.spi.EventListenerRegistry;
+import org.hibernate.event.spi.EventType;
+
+import org.hibernate.testing.TestForIssue;
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInHibernate;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Nathan Xu
+ * @author Tassilo Karge
+ */
+@TestForIssue( jiraKey = "HHH-14413" )
+public class PreUpdateEventListenerVetoTest extends BaseCoreFunctionalTestCase {
+
+	private static final Long EXAMPLE_ID_VALUE = 1L;
+
+	@Override
+	protected Class<?>[] getAnnotatedClasses() {
+		return new Class<?>[] { ExampleEntity.class };
+	}
+
+	@Override
+	protected void afterSessionFactoryBuilt() {
+		super.afterSessionFactoryBuilt();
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/event/PreUpdateEventListenerVetoTest.java
```diff
@@ -0,0 +1,76 @@
...
+	@Test
+	public void testVersionNotTouchedWhenPreUpdateEventVetoed() {
+		doInHibernate( this::sessionFactory, session -> {
+			ExampleEntity entity = session.byId( ExampleEntity.class ).load( EXAMPLE_ID_VALUE );
+			entity.name = "new_name";
+			session.update( entity );
+
+			// without fixing HHH-14413, the following statement would fail due to incremented version
+			// even though updating has been vetoed
+			session.delete( entity );
```
karge-itestra, line 60: Probably we should test the version directly as well, because the delete statement´s inner working could be changed. I propose additionally to the current test:
```
ExampleEntity entity = session.byId( ExampleEntity.class ).load( EXAMPLE_ID_VALUE );
Long initialVersion = entity.version;
//...
session.update( entity );
assertEquals(initialVersion, entity.version, "The entity version must not change when update is vetoed");
//...
```

NathanQingyangXu: Good idea. It also align well with the testing method name. 
The issue is `entity.version` doesn't change. What is changed is the internal `EntityPersister`. I am afraid we have to go about low level testing which is tightly coupled with implementation detail.
The current approach is the only means I could think of without testing internal implementation detail. Let me know your thoughts.

NathanQingyangXu: Finally I figured it out. We need to invoke `session.flush()` before assertion. Then it works perfectly! Thanks again for your good bug reporting and code review.

karge-itestra: Not sure if I misunderstand these lines in `postUpdate(...)`:
```java
		if ( getPersister().isVersioned() ) {
			this.version = nextVersion;
			getPersister().setPropertyValue( entity, getPersister().getVersionProperty(), nextVersion );
		}
```
My understanding is that this does not change the persister, but the entity. Please correct me if I am wrong. So the value of the version property should actually be changed.

karge-itestra: Alright, during me writing this you already found the solution :D Thank you!

NathanQingyangXu: btw, I also included your name as the second author of the testing case, :).

karge-itestra: That´s great to hear, thanks for doing the heavy lifting of implementation, I would not have been able to make the test case that concise.

gavinking:
This change should be replicated in HR, in `ReactiveEntityUpdateAction.reactiveExecute()`.

beikov (approved):
LGTM

karge-itestra:
Can we merge this meanwhile?

PR #3701, opened 2021-01-23
remove BitTypeDescriptor

gavinking:
Let's see what breaks if we just remove this thing.

cc @beikov


gavinking:
OK, well, apparently id did break something on Oracle :-(

Is there some way I can get access to the test report to see exactly what went wrong?

gavinking:
@beikov I don't have a way of running this on Ora right now. Do you?

Or alternatively, can someone explain me how I can run the Hibernate test suite on Docker Desktop?

beikov:
We have a script `docker_db.sh` which you can use to start a docker container e.g. use `docker_db.sh oracle` to start Oracle and then run the tests with `./gradlew check -Pdb=oracle_ci -DdbHost=127.0.0.1` with a host wherever Oracle is then running on.

Andrea and I in general agree with the change.

gavinking:
Thanks, I will try.

hibernate-core/src/main/java/org/hibernate/dialect/function/CastFunction.java
```diff
@@ -27,32 +28,52 @@
  */
 public class CastFunction extends AbstractSqmSelfRenderingFunctionDescriptor {
 
-	private Dialect dialect;
+	private final Dialect dialect;
+	private final CastType booleanCastType;
```
gavinking, line 32: I would just save the `int preferredSqlTypeCodeForBoolean` instead of the `CastType` here. That way you need one less method.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/BooleanTypeDescriptor.java
```diff
@@ -159,7 +158,7 @@ public int getDefaultSqlScale() {
 
 	@Override
 	public String getCheckCondition(String columnName, SqlTypeDescriptor sqlTypeDescriptor, Dialect dialect) {
-		return sqlTypeDescriptor instanceof BitTypeDescriptor && !dialect.supportsBitType()
+		return dialect.toBooleanValueString(true).equals("1") && !dialect.supportsBitType()
 				? columnName + " in (0,1)"
 				: null;
```
gavinking, line 163: I'm thinking we should now move this logic to `Dialect`. Add a new `getBooleanCheckConstraint()` method or something like that.

gavinking (commented):
It looks good. I've left some minor comments.

However, in addition, I believe we can now go ahead and remove all the column type mappings for `Types.BIT` from the `Dialect`s, can't we?

beikov:
> However, in addition, I believe we can now go ahead and remove all the column type mappings for `Types.BIT` from the Dialects, can't we?

We could, but if a user type uses the type code, I think we would generate `blob`/`varbinary` as column type for DDL. Not sure how important that is, but I wanted to mention it.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/BooleanTypeDescriptor.java
beikov: I had to change the implementation to respect the `SqlTypeDescriptor` again because otherwise types like e.g. `YesNoType` or `TrueFalseType` wouldn't work anymore.

hibernate-core/src/main/java/org/hibernate/dialect/function/CastFunction.java
beikov: Not sure I understand. Do you want me to move the switch block of `getBooleanCastType` into the if of `getCastType`?

gavinking: Well you have one method which translates an `int` into a `CastType`, which you save in this field.

And then another method which use this `CastType` to do something else.

I would just save away the `int`, not the `CastType` and have just the second method.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/BooleanTypeDescriptor.java
gavinking: I'm not objecting to this change. I'm just saying that instead of having this fragment of code

```
dialect.toBooleanValueString(true).equals("1") && !dialect.supportsBitType()
				? columnName + " in (0,1)"
				: null;
```

in `BooleanTypeDescriptor`, which is not supposed to know about rendering SQL, we move that fragment of code to a new method of `Dialect`, and dispatch from `BooleanTypeDescriptor. getCheckCondition()` back to `Dialect.getBooleanCheckConstraint()`.

It's a cleaner separation of concerns, and potentially more flexible.

gavinking:
> We could, but if a user type uses the type code, I think we would generate `blob`/`varbinary` as column type for DDL. Not sure how important that is, but I wanted to mention it.

Yeah, I know, and originally I was thinking along those lines, but then when I realized that this type was removed from the SQL spec, I think I changed my mind.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/BooleanTypeDescriptor.java
beikov: Since the check constraint depends on the `true`/`false` encodings and `org.hibernate.type.descriptor.java.JavaTypeDescriptor#getCheckCondition` is also implemented by `EnumJavaTypeDescriptor` I'd rather leave this as it is for now. I agree that `org.hibernate.type.descriptor.java.JavaTypeDescriptor#getCheckCondition` is not necessarily the "right place" for generating a check constraint. Maybe moving this to `org.hibernate.type.BasicType` is more appropriate? Anyway, I think this should be done in a separate PR unless you insist on doing this here already.

hibernate-core/src/main/java/org/hibernate/dialect/function/CastFunction.java
beikov: Is that a request or a comment? I like it this way, but if you insist I'll change it.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/BooleanTypeDescriptor.java
gavinking: No that's fine. I don't think you've understood what I'm saying, so I'll just do it myself later.

hibernate-core/src/main/java/org/hibernate/dialect/function/CastFunction.java
gavinking: It's no big deal.

beikov:
>Yeah, I know, and originally I was thinking along those lines, but then when I realized that this type was removed from the SQL spec, I think I changed my mind.

I just quickly googled for `sql bit deprecated` but couldn't find any reference that the SQL type was deprecated. It's not totally clear to me what SQL type the JDBC `BIT` maps to and the JDBC specification just mentions to use `Boolean` as Java type for the JDBC type `BIT`, so I assume the JDBC type `BIT` is just the predecessor of `BOOLEAN`. I'll remove the registrations in a dedicated commit so we can revert it easily if we decide that we want to interpret `BIT` like the SQL `BIT` i.e. as `bit string`.

gavinking (commented):
OK, great, merge it and then I will do the cleanup I suggested for the check constraint stuff.

PR #3702, opened 2021-01-24
Updating refs to current IDEA and Eclipse integration docs

jottinger:


Sanne:
many thanks @jottinger !

PR #3704, opened 2021-01-25
Implicit casting support for numerics and switch to argument type as return for floor/ceiling

beikov:
`floor` and `ceiling` should not return `int` or `long` as that could lead to precision loss when extracting that value. Imagine if one of the two functions used `int`, then a query like this `select ceiling(2147483647.1)` (note that 2147483647 is the maximum integer value) would lead to an overflow error or precision loss as the correct return value is 2147483648.

On top of that, this PR enables "implicit casting" between numerics for the purpose of extraction from a JDBC result set. It should be possible to extract an `int` out of e.g. `double` if known to be compatible with `int` like it would be the case for certain `floor`/`ceiling` invocations, like in `org.hibernate.dialect.function.ExtractFunction#extractWeek`.


gavinking:
This looks good to me.

sebersole (approved):
Looks great Christian!

PR #3711, opened 2021-02-02
HHH-14424 When enhanced as a proxy with dirty checking is enabled, on flush uninitialized entities containing collections are updated setting all fields to null 

dreab8:
https://hibernate.atlassian.net/browse/HHH-14424


hibernate-core/src/main/java/org/hibernate/engine/internal/AbstractEntityEntry.java
```diff
@@ -347,22 +346,20 @@ public boolean requiresDirtyCheck(Object entity) {
...
-			boolean uninitializedProxy = false;
+			boolean mayBeDirty = true;
 			if ( entity instanceof PersistentAttributeInterceptable ) {
 				final PersistentAttributeInterceptable interceptable = (PersistentAttributeInterceptable) entity;
 				final PersistentAttributeInterceptor interceptor = interceptable.$$_hibernate_getInterceptor();
 				if ( interceptor instanceof EnhancementAsProxyLazinessInterceptor ) {
-					EnhancementAsProxyLazinessInterceptor enhancementAsProxyLazinessInterceptor = (EnhancementAsProxyLazinessInterceptor) interceptor;
 					// When a proxy has dirty attributes, we have to treat it like a normal entity to flush changes
-					uninitializedProxy = !enhancementAsProxyLazinessInterceptor.isInitialized() && !( (SelfDirtinessTracker) entity ).$$_hibernate_hasDirtyAttributes();
+					mayBeDirty = ((EnhancementAsProxyLazinessInterceptor) interceptor).mayBeDirty();
```
beikov, line 355: I think we might consider more entities to be dirty due to this change, possibly leading to other performance problems. What is the reason for not considering the `SelfDirtinessTracker#$$_hibernate_hasDirtyAttributes()` method if the `enhancementAsProxyLazinessInterceptor.isInitialized()` is `true`?

beikov (changes requested):
Looking good in general, but I have a concern about dropping the dirty attribute check.

hibernate-core/src/main/java/org/hibernate/engine/internal/AbstractEntityEntry.java
dreab8: the issue is that `( (SelfDirtinessTracker) entity ).$$_hibernate_hasDirtyAttributes()` with an entity having a collection initialized ad the field level like the following
```
class User {
  public Set<Role> roles = new HashSet<>();
}
```
returns always true.
By the way I think I have a little improvement for this part.

beikov: But when is this relevant? If the entity is initialized and a hash set is assigned, then we can't know if it is dirty and thus have to consider it "maybe dirty". If the entity is not initialized, this is irrelevant and "maybe dirty" should be false.

dreab8: the problem is that in the specific case the entity is not initialized but we return that it may be dirty

dreab8:
Applied upstream.

hibernate-core/src/main/java/org/hibernate/engine/internal/AbstractEntityEntry.java
Sanne: How is that it "is may be dirty" if it's not initialized ?

dreab8: if the entity is annotated with `@DYnamicUpdate` when we set a field the `EnhancementAsProxyLazinessInterceptor` does not initialize the entity but it keep track of the attribute name in  the `writtenFieldNames` set.

PR #3713, opened 2021-02-03
HHH-14408 Fix wrong property type for integrator provided user types applied 5.4

beikov:
Backport of https://github.com/hibernate/hibernate-orm/pull/3712 for 5.4

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryBuilderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryBuilderImpl.java
@@ -462,7 +462,6 @@ public class SessionFactoryBuilderImpl implements SessionFactoryBuilderImplement
 
 	@Override
 	public SessionFactory build() {
-		metadata.validate();
 		final StandardServiceRegistry serviceRegistry = metadata.getMetadataBuildingOptions().getServiceRegistry();
 		BytecodeProvider bytecodeProvider = serviceRegistry.getService( BytecodeProvider.class );
 		addSessionFactoryObservers( new SessionFactoryObserverForBytecodeEnhancer( bytecodeProvider ) );
--- a/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/service/internal/EventListenerRegistryImpl.java
@@ -150,7 +150,7 @@ public class EventListenerRegistryImpl implements EventListenerRegistry, Stoppab
 					propertyIterator.hasNext(); ) {
 				Property property = (Property) propertyIterator.next();
 
-				if ( property.getType().isComponentType() ) {
+				if ( property.isComposite() ) {
 					callbackBuilder.buildCallbacksForEmbeddable(
 							property,
 							mappedClass,
--- a/hibernate-core/src/main/java/org/hibernate/internal/SessionFactoryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/SessionFactoryImpl.java
@@ -295,6 +295,7 @@ public class SessionFactoryImpl implements SessionFactoryImplementor {
 				);
 				identifierGenerators.put( model.getEntityName(), generator );
 			} );
+			metadata.validate();
 
 			LOG.debug( "Instantiated session factory" );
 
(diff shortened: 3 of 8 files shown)


Sanne (changes requested):
So I was testing this with Quarkus, as moving the `metadata.validate();` looked as something that could possibly break it, and unfortunately my hunch was right. This would break the native-image compilation.

The problem is that `validate()` actually performs a lot of operations, and we need to ensure this happens in the augmentation phase of Quarkus (which means: run only once during the native-image compilation itself, and not run again when the produced binary is started).

Why are you moving the `validate()` method?

Sanne:
I'm experimenting with a patch on the Quarkus side which might make this less brittle ( and make your patch work)

beikov:
The problem is that the validation requires property types which might not be available yet because integrators can contribute them. I deferred the validation until after the integrators ran so that this is not an issue.

Sanne:
Right, so I think the problem is the `Integrator` contract: integrators are passed a SessionFactory, but in its very nature the `SessionFactory` should be immutable - so it's too late.

Like the javadoc on SessionFactoryImpl states:

> This class must appear immutable to clients, even if it does all kinds of caching and pooling under the covers.

IMO it's not legal to expect `Integrator` (s) to be allowed to mutate its state other than enlist listeners, but here it seems it's used also to register user types?

Not sure how to crack this w/o changing the SPIs.  @sebersole WDYT, is this something we could improve on in 6?

I'll try workaround this in Quarkus but it's a big challenge to not being able to rely on the metadata and metamodel being "fully complete" yet when we're constructing the `SessionFactory`. IMO if there's a need for intermediate steps, we should represent them with an intermediate object to hold the intermediate state.

Sanne:
> IMO it's not legal to expect Integrator (s) to be allowed to mutate its state other than enlist listeners,

To elaborate more on that; I wonder if even the "enlist liteners" aspect is a bit questionable: I mentioned it to be ok only because it's perhaps less problematic and I tend to be pragmatic about such extension needs.

But thinking in terms of better design, pehaps we should just expose the Metadata and ServiceRegisty; this would allow modifying EventListenerRegistry, etc.. and all other components in a previous phase.

I suspect that the SessionFactory is handed over because a reference to it might be useful? We should look at the use cases and see if we can defer providing the reference.

beikov:
User types are registered through the `org.hibernate.boot.spi.MetadataImplementor`. The SPI even says that

> The session factory being created

i.e. it suggests that it is in the process of creation. IMO it's fine how it is, but we could certainly try to move integrators out. The only issue then is that we have to move certain other parts out as well and pass a different context object to the integrators so that they can do changes. Maybe we can try to split the validation into two steps instead to do whatever is causing problems for Quarkus earlier?

Sanne:
yes the SPI is being clear, but "session factory being created" is a bit vague - I guess intentionally :)

I'd prefer to have the integration-with-integrators moved out, but we can't do that today as the `Integrator` SPI expects to be passed a reference  to the SessionFactory instance. Which doesn't exist before we invoke its constructor ..

Regarding validating in two steps.. maybe that's a compromise, but I'm not sure how far that makes sense a I would *prefer* to being able to run all validations (including the second step) as part of the build. But of course while that's ideal, if we hit a roadblock we'll only validate what's possible.

Anyhow, I managed to workaround this change in Quarkus now, and it's not looking too bad so we could merge this. Just be aware that (in Quarkus only) this validation phase will still run earlier. We also don't support such integrators yet, so we can face that problem when we get there.

beikov:
Workaround how? What do you mean by we don't support _such integrators yet_?

The proper way for contributing types in Hibernate is IMO to use a `org.hibernate.boot.spi.MetadataContributor` and do it there as that is called while the metadata is being built. We could tell users to just switch to this, but I don't know if there are other purposes for integrators that could interfere with the metadata.

beikov:
IMO the problem is that we pass the `Metadata` which allows changes. Maybe for 6.0 we can drop that parameter?

Sanne:
> Workaround how?

Not very polished yet but this works:
 - https://github.com/quarkusio/quarkus/commit/9dbb5c5566d454e14c39617bf0835eeb867c26c7

As dicussed, but I'll repeat it here as note to share with others, the drawback of my Quarkus workaround is that the validation() invocation happens before the last-minute `Integrator` (s) are applied. This gets mitigated by the fact that I don't really know of `Integrator` implementations which are supposed to work, we could improve that bommit by scanning for them and logging a warning about how to do the better integration (as discussed below).

> What do you mean by we don't support such integrators yet?

Just that in Quarkus most implementations of `Integrator` are unlikely to work OOB; they might work with some luck but I'd prefer adding explicit extensions for the most popular ones.

And once we have extensions, we could make them bind over the `org.hibernate.boot.spi.MetadataContributor` like you suggest: I agree that's the better way forward for this use case.

I'll merge this next; hopefully we can improve on this soonish but that's not something orthogonal to this PR. Thanks!

PR #3716, opened 2021-02-04
HHH-13658 : make NO_PROXY unnecessary

sebersole:
- Better handle `FetchModeType#LAZY` for to-one associations based on whether bytecode-enhancement-as-proxy is enabled.  Minimize the cases a user is likely to need to use `@LazyToOne`
- See also EAP7-1402
- Currently has 5 failing tests, all related to assertions regarding handling of polymorphic to-one associations which is behavior these changes explicitly affect.  So most likely these are "expected"

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
@@ -25,11 +25,18 @@ import org.hibernate.mapping.Value;
  * @author Steve Ebersole
  */
 public class EnhancementHelper {
+
+	@FunctionalInterface
+	public interface InheritanceChecker {
+		boolean hasSubclasses(String entityName);
+	}
+
 	/**
 	 * Should the given property be included in the owner's base fetch group?
 	 */
 	public static boolean includeInBaseFetchGroup(
 			Property bootMapping,
+			InheritanceChecker inheritanceChecker,
 			boolean isEnhanced,
 			boolean allowEnhancementAsProxy,
 			boolean collectionsInDefaultFetchGroupEnabled) {
@@ -39,7 +46,7 @@ public class EnhancementHelper {
 			if ( value instanceof ToOne ) {
 				if ( ( (ToOne) value ).isUnwrapProxy() ) {
 					BytecodeLogger.LOGGER.debugf(
-							"To-one property `%s#%s` was mapped with LAZY + NO_PROXY but the class was not enhanced",
+							"`%s#%s` was mapped with LAZY + NO_PROXY but the class was not enhanced",
 							bootMapping.getPersistentClass().getEntityName(),
 							bootMapping.getName()
 					);
@@ -51,15 +58,32 @@ public class EnhancementHelper {
 		if ( value instanceof ToOne ) {
 			final ToOne toOne = (ToOne) value;
 			if ( toOne.isLazy() ) {
-				if ( toOne.isUnwrapProxy() ) {
-					if ( toOne instanceof OneToOne ) {
-						return false;
-					}
-					// include it in the base fetch group so long as the config allows
-					// using the FK to create an "enhancement proxy"
-					return allowEnhancementAsProxy;
+				// include it in the base fetch group so long as the config allows
+				// using the FK to create an "enhancement proxy"
+
+				// however a few special cases to handle...
+
+				if ( ! toOne.isReferenceToPrimaryKey() ) {
+					// we do not have a reference to the associated primary-key
+					return false;
 				}
 
+				if ( inheritanceChecker.hasSubclasses( toOne.getReferencedEntityName() ) ) {
+					// the associated type has subclasses - we cannot use the enhanced proxy and will generate a HibernateProxy
+					if ( ! toOne.isUnwrapProxyImplicit() ) {
+						// NO_PROXY was explicitly requested
+						BytecodeLogger.LOGGER.debugf(
+								"`%s#%s` was mapped with LAZY with explicit NO_PROXY but the associated entity (`%s`) has subclasses",
+								bootMapping.getPersistentClass().getEntityName(),
+								bootMapping.getName(),
+								toOne.getReferencedEntityName()
+						);
+					}
+					// however, select the fk to create the proxy
+					return true;
+				}
+
+				return allowEnhancementAsProxy;
 			}
 
 			return true;
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
@@ -83,7 +83,13 @@ public class EnhancementHelper {
 					return true;
 				}
 
-				return allowEnhancementAsProxy;
+				if ( toOne.isUnwrapProxy() && ! toOne.isUnwrapProxyImplicit() ) {
+					return allowEnhancementAsProxy;
+				}
+
+//				if ( toOne.isUnwrapProxy() || toOne.isUnwrapProxyImplicit() ) {
+//					return allowEnhancementAsProxy;
+//				}
 			}
 
 			return true;
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/LazyAttributesMetadata.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/LazyAttributesMetadata.java
@@ -17,8 +17,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.hibernate.boot.spi.MetadataImplementor;
 import org.hibernate.mapping.PersistentClass;
 import org.hibernate.mapping.Property;
+import org.hibernate.persister.spi.PersisterCreationContext;
 
 /**
  * Information about all of the bytecode lazy attributes for an entity
@@ -35,7 +37,8 @@ public class LazyAttributesMetadata implements Serializable {
 			PersistentClass mappedEntity,
 			boolean isEnhanced,
 			boolean allowEnhancementAsProxy,
-			boolean collectionsInDefaultFetchGroupEnabled) {
+			boolean collectionsInDefaultFetchGroupEnabled,
+			PersisterCreationContext creationContext) {
 		final Map<String, LazyAttributeDescriptor> lazyAttributeDescriptorMap = new LinkedHashMap<>();
 		final Map<String, Set<String>> fetchGroupToAttributesMap = new HashMap<>();
 
@@ -47,6 +50,12 @@ public class LazyAttributesMetadata implements Serializable {
 			final Property property = (Property) itr.next();
 			final boolean lazy = ! EnhancementHelper.includeInBaseFetchGroup(
 					property,
+					(entityName) -> {
+						final MetadataImplementor metadata = creationContext.getMetadata();
+						final PersistentClass entityBinding = metadata.getEntityBinding( entityName );
+						assert entityBinding != null;
+						return entityBinding.hasSubclasses();
+					},
 					isEnhanced,
 					allowEnhancementAsProxy,
 					collectionsInDefaultFetchGroupEnabled
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
@@ -3163,7 +3163,9 @@ public final class AnnotationBinder {
 		}
 		else {
 			toOne.setLazy( fetchType == FetchType.LAZY );
-			toOne.setUnwrapProxy( false );
+
+			toOne.setUnwrapProxy( true );
+			toOne.setUnwrapProxyImplicit( true );
 		}
 		if ( fetch != null ) {
 			if ( fetch.value() == org.hibernate.annotations.FetchMode.JOIN ) {
--- a/hibernate-core/src/main/java/org/hibernate/tuple/PropertyFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/tuple/PropertyFactory.java
@@ -10,6 +10,7 @@ import java.lang.reflect.Constructor;
 
 import org.hibernate.EntityMode;
 import org.hibernate.HibernateException;
+import org.hibernate.boot.spi.MetadataImplementor;
 import org.hibernate.boot.spi.SessionFactoryOptions;
 import org.hibernate.bytecode.enhance.spi.interceptor.EnhancementHelper;
 import org.hibernate.engine.internal.UnsavedValueFactory;
@@ -22,6 +23,7 @@ import org.hibernate.mapping.KeyValue;
 import org.hibernate.mapping.PersistentClass;
 import org.hibernate.mapping.Property;
 import org.hibernate.persister.entity.EntityPersister;
+import org.hibernate.persister.spi.PersisterCreationContext;
 import org.hibernate.property.access.spi.Getter;
 import org.hibernate.property.access.spi.PropertyAccess;
 import org.hibernate.property.access.spi.PropertyAccessStrategy;
@@ -148,6 +150,7 @@ public final class PropertyFactory {
 	 * @param property The mapped property.
 	 * @param lazyAvailable Is property lazy loading currently available.
 	 *
+	 * @param creationContext
 	 * @return The appropriate NonIdentifierProperty definition.
 	 */
 	public static NonIdentifierAttribute buildEntityBasedAttribute(
@@ -155,7 +158,8 @@ public final class PropertyFactory {
 			SessionFactoryImplementor sessionFactory,
 			int attributeNumber,
 			Property property,
-			boolean lazyAvailable) {
+			boolean lazyAvailable,
+			PersisterCreationContext creationContext) {
 		final Type type = property.getValue().getType();
 
 		final NonIdentifierAttributeNature nature = decode( type );
@@ -173,6 +177,12 @@ public final class PropertyFactory {
 		SessionFactoryOptions sessionFactoryOptions = sessionFactory.getSessionFactoryOptions();
 		final boolean lazy = ! EnhancementHelper.includeInBaseFetchGroup(
 				property,
+				(entityName) -> {
+					final MetadataImplementor metadata = creationContext.getMetadata();
+					final PersistentClass entityBinding = metadata.getEntityBinding( entityName );
+					assert entityBinding != null;
+					return entityBinding.hasSubclasses();
+				},
 				lazyAvailable,
 				sessionFactoryOptions.isEnhancementAsProxyEnabled(),
 				sessionFactoryOptions.isCollectionsInDefaultFetchGroupEnabled()
--- a/hibernate-core/src/main/java/org/hibernate/tuple/entity/EntityMetamodel.java
+++ b/hibernate-core/src/main/java/org/hibernate/tuple/entity/EntityMetamodel.java
@@ -20,6 +20,7 @@ import java.util.Set;
 import org.hibernate.EntityMode;
 import org.hibernate.HibernateException;
 import org.hibernate.MappingException;
+import org.hibernate.boot.spi.MetadataImplementor;
 import org.hibernate.boot.spi.SessionFactoryOptions;
 import org.hibernate.bytecode.enhance.spi.interceptor.EnhancementHelper;
 import org.hibernate.bytecode.spi.BytecodeEnhancementMetadata;
@@ -36,6 +37,7 @@ import org.hibernate.mapping.Component;
 import org.hibernate.mapping.PersistentClass;
 import org.hibernate.mapping.Property;
 import org.hibernate.persister.entity.EntityPersister;
+import org.hibernate.persister.spi.PersisterCreationContext;
 import org.hibernate.tuple.GenerationTiming;
 import org.hibernate.tuple.IdentifierProperty;
 import org.hibernate.tuple.InDatabaseValueGenerationStrategy;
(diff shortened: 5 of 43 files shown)


gbadner:
@sebersole, it looks like you are making this the default behavior for bytecode-proxies enabled in master branch. I think it's the right thing to do in master. I just want to confirm that's your intention.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
```diff
@@ -25,11 +25,18 @@
...
+
 	/**
 	 * Should the given property be included in the owner's base fetch group?
 	 */
 	public static boolean includeInBaseFetchGroup(
 			Property bootMapping,
+			InheritanceChecker inheritanceChecker,
 			boolean isEnhanced,
 			boolean allowEnhancementAsProxy,
 			boolean collectionsInDefaultFetchGroupEnabled) {
```
gbadner, line 42: I'm assuming that changing this SPI is OK in master.

Note to self: when this gets backported, the old version of the method should be deprecated.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
```diff
@@ -51,15 +58,38 @@ public static boolean includeInBaseFetchGroup(
...
+						BytecodeLogger.LOGGER.debugf(
+								"`%s#%s` was mapped with LAZY with explicit NO_PROXY but the associated entity (`%s`) has subclasses",
+								bootMapping.getPersistentClass().getEntityName(),
+								bootMapping.getName(),
+								toOne.getReferencedEntityName()
+						);
 					}
-					// include it in the base fetch group so long as the config allows
-					// using the FK to create an "enhancement proxy"
+					// however, select the fk to create the proxy
```
gbadner, line 82: In `DefaultLoadEventListener#proxyOrLoad` there is a check for the existence of a `ProxyFactory` [here](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java#L283). 

Do we need to check for a `ProxyFactory` here and factor that into the decision for the return value?

IIUC, this will apply regardless of how `allowEnhancementAsProxy` is set. I think that's fine for master.

When this gets backported, I don't think this should return `true` when `allowEnhancementAsProxy == false`. WDYT?

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
```diff
@@ -51,15 +58,38 @@ public static boolean includeInBaseFetchGroup(
...
+					return true;
+				}
+
+				if ( toOne.isUnwrapProxy() && ! toOne.isUnwrapProxyImplicit() ) {
```
gbadner, line 86: I'm unsure about this `if` statement. I'm going to do more checking on this Monday.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/LazyAttributesMetadata.java
```diff
@@ -35,7 +37,8 @@ public static LazyAttributesMetadata from(
 			PersistentClass mappedEntity,
 			boolean isEnhanced,
 			boolean allowEnhancementAsProxy,
-			boolean collectionsInDefaultFetchGroupEnabled) {
+			boolean collectionsInDefaultFetchGroupEnabled,
+			PersisterCreationContext creationContext) {
```
gbadner, line 41: I'm assuming that changing this SPI is OK in master.

Note to self: when this gets backported, the old version of the method should be deprecated.

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
```diff
@@ -3163,7 +3163,9 @@ else if ( oneToOne != null ) {
 		}
 		else {
 			toOne.setLazy( fetchType == FetchType.LAZY );
-			toOne.setUnwrapProxy( false );
+
```
gbadner, line 3166: Should the following be set to `true` when ` fetchType != FetchType.LAZY`

Will this change behavior for non-lazy attributes that end up getting loaded as a proxy from the `PersistenceContext`?

I'd like to look more deeply at this on Monday.

hibernate-core/src/main/java/org/hibernate/tuple/PropertyFactory.java
```diff
@@ -148,14 +150,16 @@ public static VersionProperty buildVersionProperty(
...
 	 * @return The appropriate NonIdentifierProperty definition.
 	 */
 	public static NonIdentifierAttribute buildEntityBasedAttribute(
 			EntityPersister persister,
 			SessionFactoryImplementor sessionFactory,
 			int attributeNumber,
 			Property property,
-			boolean lazyAvailable) {
+			boolean lazyAvailable,
+			PersisterCreationContext creationContext) {
```
gbadner, line 162: I'm assuming that changing this SPI is OK in master.

Note to self: when this gets backported, the old version of the method should be deprecated.

hibernate-core/src/main/java/org/hibernate/tuple/entity/EntityMetamodel.java
```diff
@@ -131,8 +133,8 @@
 	public EntityMetamodel(
 			PersistentClass persistentClass,
 			EntityPersister persister,
-			SessionFactoryImplementor sessionFactory) {
-		this.sessionFactory = sessionFactory;
+			PersisterCreationContext creationContext) {
```
gbadner, line 136: I'm assuming that changing this SPI is OK in master.

Note to self: when this gets backported, the old version of the method should be deprecated.

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
pblanchardie: I think yes. I tried, and it fixed the bug ;)

beikov (commented):
Since all uses of `includeInBaseFetchGroup` negate the result to determine if the property is lazy, it's a bit hard to follow the logic for me. If this needs changes anyway, could we maybe rename this to `isPropertyLazy` and remove the negations?

Also, do I understand this correct, that lazy in this context means that the entity field/property does not receive a proxy? I conclude that from the fact that `to-one` associations to non-primary keys are always treated as "lazy" for which we can never have proxies, because proxies can only use the primary key. If this understanding is correct, I think the term _lazy_ is still a bit confusing when considering that associations can be lazy as well even if the fields/properties receive a proxy. How about using an enum then and naming the method `getPropertyInitializationMode` with options `NULL`, `PROXY` and `VALUE`. This would be a great help for my understanding of what is going on and as far as I understood (correct me if I am wrong), this is essentially what we are trying to model here.

sebersole:
> @sebersole, it looks like you are making this the default behavior for bytecode-proxies enabled in master branch. I think it's the right thing to do in master. I just want to confirm that's your intention.

That is the way it is defined now in this work.  We can talk about whether that makes sense for 5.3 (anecdotally, I think it does).
The main problem with not (and this is the main difficulty with this feature at all in 5.x) is that its made in multiple places that sometimes do not have all the needed info.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
sebersole: I really cannot imagine user call this.  But sure, whatever you think wrt 5.3

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
sebersole: That's a good question.  The thought process here was that the check for ProxyFactory was really just needed for entities with subclasses.  That is maybe not exactly correct yet.  This is obviously the main point of change for this improvement and its tricky to get just right - it comes back to what I mentioned earlier about timing and access to info.  This block is exactly why this is still a draft :)

Keep in mind also that this simply defines what columns should be selected as part of the "base fetch group".  I'm not sure I agree with your statement about `allowEnhancementAsProxy == false`.  Still need to play with it some more

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
sebersole: This check is actually the exact improvement.  Previously, if `@LazyToOne` was not specified (implicit) then unwrap-proxy would be set to true.  However, if `@LazyToOne` is implicit we want to treat that differently.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/LazyAttributesMetadata.java
sebersole: Again, I cannot see this being a method users will call.  This is specifically creating these objects

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
sebersole: I think that is probably true

hibernate-core/src/main/java/org/hibernate/tuple/PropertyFactory.java
sebersole: Same comment as previous

hibernate-core/src/main/java/org/hibernate/tuple/entity/EntityMetamodel.java
sebersole: Again, same as previous.  For this though *maybe* other Hibernate projects use it, but I doubt it

sebersole:
> Since all uses of `includeInBaseFetchGroup` negate the result to determine if the property is lazy, it's a bit hard to follow the logic for me. If this needs changes anyway, could we maybe rename this to `isPropertyLazy` and remove the negations?

I prefer it to remain this way.  Your phrasing is just as confusing imo.  Lazy in what sense, which is what this comes down to.  The problem has always been that we have 2 distinct concepts that we call "lazy" so that phrasing gets confusing.


> Also, do I understand this correct, that lazy in this context means that the entity field/property does not receive a proxy?

Not really, it means whether that property is considered part of the entity's base fetch group.  When we first initialize as entity, what state should we select from the database?  That's what this method determines.  That is not the same as whether there is a proxy or not.

I also thought of a different return value here.  I thought of an enum also (great minds...).  The problem is that of compatibility.  I think we should absolutely go that route with 6.0.  But for 5.x, to a large degree, it comes down to how much change in these signatures we want to introduce.

sebersole:
> Not really, it means whether that property is considered part of the entity's base fetch group. When we first initialize as entity, what state should we select from the database?

To clarify a bit.. there are 3 conditions that need to be considered:

1. No enhancement
2. Enhanced but enhanced proxies disallowed
3. Enhanced and enhanced proxies allowed

The outcome here is generally different for all 3 cases.

gbadner:
@sebersole, thanks for the feedback. I'll dig into this more today.

gbadner:
I'm seeing 20 failures locally:

org.hibernate.jpa.test.callbacks.PrivateConstructorEnhancerTest > testGetReferenceEntity FAILED
    org.hibernate.testing.junit4.CallbackException at TestClassMetadata.java:208
        Caused by: javax.persistence.PersistenceException at PrivateConstructorEnhancerTest.java:46
            Caused by: org.hibernate.PersistentObjectException at PrivateConstructorEnhancerTest.java:46

org.hibernate.jpa.test.callbacks.PrivateConstructorEnhancerTest > testListEntity FAILED
    org.hibernate.testing.junit4.CallbackException at TestClassMetadata.java:208
        Caused by: javax.persistence.PersistenceException at PrivateConstructorEnhancerTest.java:46
            Caused by: org.hibernate.PersistentObjectException at PrivateConstructorEnhancerTest.java:46

org.hibernate.jpa.test.callbacks.PrivateConstructorEnhancerTest > testListJoinFetchEntity FAILED
    org.hibernate.testing.junit4.CallbackException at TestClassMetadata.java:208
        Caused by: javax.persistence.PersistenceException at PrivateConstructorEnhancerTest.java:46
            Caused by: org.hibernate.PersistentObjectException at PrivateConstructorEnhancerTest.java:46

org.hibernate.jpa.test.callbacks.PrivateConstructorEnhancerTest > testLoadProxyAssociation FAILED
    org.hibernate.testing.junit4.CallbackException at TestClassMetadata.java:208
        Caused by: javax.persistence.PersistenceException at PrivateConstructorEnhancerTest.java:46
            Caused by: org.hibernate.PersistentObjectException at PrivateConstructorEnhancerTest.java:46

org.hibernate.orm.test.mapping.lazytoone.InstrumentedJoinedLazyToOneTest > testEnhancedButProxyNotAllowed FAILED
    java.lang.AssertionError at InstrumentedJoinedLazyToOneTest.java:96

org.hibernate.orm.test.mapping.lazytoone.InstrumentedLazyToOneTest > testEnhancedButProxyNotAllowed FAILED
    java.lang.AssertionError at InstrumentedLazyToOneTest.java:88

org.hibernate.test.bytecode.enhancement.lazy.proxy.LazyGroupWithInheritanceTest > loadEntityWithAssociationToAbstract FAILED
    java.lang.AssertionError at LazyGroupWithInheritanceTest.java:57

org.hibernate.test.bytecode.enhancement.lazy.proxy.LazyGroupWithInheritanceTest > queryEntityWithAssociationToAbstract FAILED
    java.lang.AssertionError at LazyGroupWithInheritanceTest.java:88

org.hibernate.test.bytecode.enhancement.lazy.LazyLoadingIntegrationTest > test FAILED
    java.lang.NullPointerException at LazyLoadingIntegrationTest.java:82

org.hibernate.test.bytecode.enhancement.lazy.LazyLoadingTest > test FAILED
    java.lang.AssertionError at LazyLoadingTest.java:87

org.hibernate.test.bytecode.enhancement.lazy.LazyCollectionDetachTest > testDetachProxy FAILED
    javax.persistence.PersistenceException at LazyCollectionDetachTest.java:85
        Caused by: org.hibernate.HibernateException at LazyCollectionDetachTest.java:85
            Caused by: java.lang.NoSuchMethodException at LazyCollectionDetachTest.java:85

org.hibernate.test.bytecode.enhancement.lazy.QueryScrollingWithInheritanceTest > testScrollableWithSession FAILED
    java.lang.AssertionError at QueryScrollingWithInheritanceTest.java:160

org.hibernate.test.bytecode.enhancement.lazy.QueryScrollingWithInheritanceTest > testScrollableWithStatelessSession FAILED
    java.lang.AssertionError at QueryScrollingWithInheritanceTest.java:106

org.hibernate.test.bytecode.enhancement.lazy.NaturalIdInUninitializedAssociationTest > testImmutableNaturalId FAILED
    java.lang.AssertionError at NaturalIdInUninitializedAssociationTest.java:57

org.hibernate.test.bytecode.enhancement.lazy.NaturalIdInUninitializedAssociationTest > testMutableNaturalId FAILED
    java.lang.AssertionError at NaturalIdInUninitializedAssociationTest.java:74

org.hibernate.test.bytecode.enhancement.lazy.LazyCollectionLoadingTest > testNoTransaction FAILED
    org.hibernate.HibernateException at LazyCollectionLoadingTest.java:111
        Caused by: java.lang.NoSuchMethodException at LazyCollectionLoadingTest.java:111

org.hibernate.test.bytecode.enhancement.lazy.LazyCollectionLoadingTest > testTransaction FAILED
    org.hibernate.HibernateException at LazyCollectionLoadingTest.java:90
        Caused by: java.lang.NoSuchMethodException at LazyCollectionLoadingTest.java:90

org.hibernate.test.bytecode.enhancement.lazy.group.LazyGroupTest > testAccess FAILED
    java.lang.AssertionError at LazyGroupTest.java:104

org.hibernate.test.bytecode.enhancement.lazy.group.LazyGroupTest > testUpdate FAILED
    java.lang.AssertionError at LazyGroupTest.java:145

org.hibernate.test.bytecode.enhancement.ondemandload.OnDemandLoadTest > testClosedSession FAILED
    java.lang.AssertionError at OnDemandLoadTest.java:99

10304 tests completed, 20 failed, 1330 skipped

sebersole:
> I'm seeing 20 failures locally:

You could have just seen those in the CI runs ;)

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
gbadner: OK, I see that the existence of a {{ProxyFactory}} isn't known at the time this is called.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
```diff
@@ -39,7 +46,7 @@ public static boolean includeInBaseFetchGroup(
 			if ( value instanceof ToOne ) {
 				if ( ( (ToOne) value ).isUnwrapProxy() ) {
```
gbadner, line 47: Due to the change below to `AnnotationBinder#defineFetchingStrategy`, if `@LazyToOne` is not mapped, then the following will be set:
```
toOne.setUnwrapProxy( true );
toOne.setUnwrapProxyImplicit( true );
```
and the message will end up getting DEBUG logged.

I think it should only get logged if NO_PROXY is explicitly mapped:

```
final ToOne toOne = (ToOne) value;
if ( toOne.isUnwrapProxy() && !toOne.isUnwrapProxyImplicit() ) {
```

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
```diff
@@ -51,15 +58,38 @@ public static boolean includeInBaseFetchGroup(
...
+				if ( inheritanceChecker.hasSubclasses( toOne.getReferencedEntityName() ) ) {
+					// the associated type has subclasses - we cannot use the enhanced proxy and will generate a HibernateProxy
+					if ( ! toOne.isUnwrapProxyImplicit() ) {
+						// NO_PROXY was explicitly requested
+						BytecodeLogger.LOGGER.debugf(
+								"`%s#%s` was mapped with LAZY with explicit NO_PROXY but the associated entity (`%s`) has subclasses",
+								bootMapping.getPersistentClass().getEntityName(),
+								bootMapping.getName(),
+								toOne.getReferencedEntityName()
+						);
```
gbadner, line 80: I know it's a little late to ask this, but should this be a warning, or at least something a little stronger than just DEBUG logged, since `@LazyToOne(NO_PROXY)` is not compatible with an association with an entity class that has subclasses?

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
gbadner: OK, I understand this case (explicitly has NO_PROXY) now, so this looks OK.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
```diff
@@ -51,15 +58,38 @@ public static boolean includeInBaseFetchGroup(
...
+				if ( toOne.isUnwrapProxy() && ! toOne.isUnwrapProxyImplicit() ) {
 					return allowEnhancementAsProxy;
 				}
 
+//				if ( toOne.isUnwrapProxy() || toOne.isUnwrapProxyImplicit() ) {
+//					return allowEnhancementAsProxy;
+//				}
 			}
 
 			return true;
```
gbadner, line 95: A lazy to-one association that does not explicitly have `@LazyToOne(NO_PROXY)`, and does not fall into the other special cases will end up returning `true`.

That makes sense for `allowEnhancementAsProxy == true`.

Is this correct if `allowEnhancementAsProxy == false`? Does that mean that the associated entity will be loaded as an uninitialized `HibernateProxy`?

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
gbadner: This is fine.

gbadner:
@sebersole, I checked all the places where `ToOne#setUnwrapProxy` gets called, to make sure that the corresponding `ToOne#isUnwrapProxyImplicit` gets set properly.

I see that the calls from `ModelBinder` don't set `ToOne#isUnwrapProxyImplicit`. Since `ToOne#isUnwrapProxyImplicit` defaults to `false`, that DEBUG message in `EnhancementHelper#includeInBaseFetchGroup` will log the message for HBM mappings.

`OneToOneSecondPass#doSecondPass` also calls  `ToOne#setUnwrapProxy` without calling `ToOne#isUnwrapProxyImplicit` [here](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/cfg/OneToOneSecondPass.java#L199). I think this one is a bug that needs to be fixed.

hibernate-core/src/main/java/org/hibernate/cfg/AnnotationBinder.java
sebersole: What I found is that knowing when the LazyToOne was implicit is just as important.  To that end, this is now as follows in my newest work, which I'll be pushing as a new PR soon -

```
toOne.setUnwrapProxy( fetchType != FetchType.LAZY );
toOne.setUnwrapProxyImplicit( true );
```

this new "implicitness" is checked later.  Ideally unwrap-proxy would have been a Boolean value allowing null rather than a boolean forcing a true/false here.  But I felt adding a new method was less disruptive since that mapping model is considered an API

sebersole:
Closing in preference of https://github.com/hibernate/hibernate-orm/pull/3739

PR #3717, opened 2021-02-05
HHH-14438 - ValueHandlers are broken

andreybpanfilov:
https://hibernate.atlassian.net/browse/HHH-14438


gavinking:
There are no `ValueHandler`s in H6 and H7, and at this point I doubt we're going to fix this in H5.6 at this point, since it wasn't a regression. Furthermore, this pull request does not come with a test.

So I'm going to close this, even though the suggested fix looks like it was probably correct.

@sebersole can correct me if this was the wrong action to take here.

PR #3719, opened 2021-02-08
HHH-14443: add hashcode to ObjectTypeCacheEntry

Francoisvandelft:
The hashcode of ObjectTypeCacheEntry is used in the key of the query cache. Without the hashcode implementation, the lookup on a query with an anytype will fail.


Sanne:
The Oracle CI job is in trouble, sorry about that. I'll ignore that job and see if someone could review soon.

Sanne:
Thanks, I've merged this and backported to branch 5.4.

PR #3720, opened 2021-02-08
HHH-14444 Remove synchronization from PooledLoThreadLocalOptimizer

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14444

@Sanne as discussed in https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/We.20have.20a.20blocker .
Not sure we want to backport this to 5.4, maybe let's see if it really is the root of the problem in the performance tests first?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/id/enhanced/PooledLoThreadLocalOptimizer.java
+++ b/hibernate-core/src/main/java/org/hibernate/id/enhanced/PooledLoThreadLocalOptimizer.java
@@ -31,15 +31,6 @@ public class PooledLoThreadLocalOptimizer extends AbstractOptimizer {
 			PooledLoOptimizer.class.getName()
 	);
 
-	private static class GenerationState {
-		// last value read from db source
-		private IntegralDataTypeHolder lastSourceValue;
-		// the current generator value
-		private IntegralDataTypeHolder value;
-		// the value at which we'll hit the db again
-		private IntegralDataTypeHolder upperLimitValue;
-	}
-
 	private final ThreadLocal<GenerationState> singleTenantState = ThreadLocal.withInitial( GenerationState::new );
 	private final ThreadLocal<Map<String, GenerationState>> multiTenantStates = ThreadLocal.withInitial( HashMap::new );
 
@@ -59,22 +50,8 @@ public class PooledLoThreadLocalOptimizer extends AbstractOptimizer {
 
 	@Override
 	public Serializable generate(AccessCallback callback) {
-		final GenerationState generationState = locateGenerationState( callback.getTenantIdentifier() );
-		return generate( generationState, callback );
-	}
-
-	private Serializable generate(GenerationState generationState, AccessCallback callback) {
-		if ( generationState.value == null
-				|| !generationState.value.lt( generationState.upperLimitValue ) ) {
-			generationState.lastSourceValue = callback.getNextValue();
-			generationState.upperLimitValue = generationState.lastSourceValue.copy().add( incrementSize );
-			generationState.value = generationState.lastSourceValue.copy();
-			// handle cases where initial-value is less that one (hsqldb for instance).
-			while ( generationState.value.lt( 1 ) ) {
-				generationState.value.increment();
-			}
-		}
-		return generationState.value.makeValueThenIncrement();
+		return locateGenerationState( callback.getTenantIdentifier() )
+				.generate( callback, incrementSize );
 	}
 
 	private GenerationState locateGenerationState(String tenantIdentifier) {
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/id/enhanced/PooledLoThreadLocalOptimizer.java
```diff
@@ -58,26 +58,28 @@ public PooledLoThreadLocalOptimizer(Class returnClass, int incrementSize) {
 	public Serializable generate(AccessCallback callback) {
 		if ( callback.getTenantIdentifier() == null ) {
 			final GenerationState local = localAssignedIds.get();
-			if ( local.value != null && local.value.lt( local.upperLimitValue ) ) {
-				return local.value.makeValueThenIncrement();
-			}
```
Sanne, line 63: Can you help me understand this commit? The comment suggests you want to avoid a `synchronization`, but this block was already unsynched?

hibernate-core/src/main/java/org/hibernate/id/enhanced/PooledLoThreadLocalOptimizer.java
```diff
@@ -127,4 +122,10 @@ public IntegralDataTypeHolder getLastSourceValue() {
 	public boolean applyIncrementSizeToSourceValues() {
 		return true;
 	}
+
+	private static class GenerationStates {
```
Sanne, line 103: I'm thinking that this implementation is not a very good choice for people who are using multi-tenancy, when it's combined with "many tenants" and many threads, as the state gets rather large, with significant ID fragmentation on the DB level.

So while it's of course a choice we want to allow as the definition of such trade-offs is subjective, I would expect this Optimizer being used primarily by people who are not using multitenancy; but this wrapper object penalises them as it leads to more memory waste by repeating this structure for each thread.

Could be simple to use two different ThreadLocals, one holding `GenerationState` for single-tenant usage, and one holding the map for multi-tenant?

yrodiere:
```
I'm thinking that this implementation is not a very good choice for people who are using multi-tenancy, when it's combined with "many tenants" and many threads, as the state gets rather large, with significant ID fragmentation on the DB level.
```

It's a decent choice if you only have a few tenants, and a bad one if you have thousands of them.
Regardless, this is how this optimizer is supposed to work, as described in the documentation. People who don't want this state to be held in a thread-local can simply use the `pooled-lo` optimizer instead of the `pooled-lotl` optimizer.

```
I would expect this Optimizer being used primarily by people who are not using multitenancy; but this wrapper object penalises them as it leads to more memory waste by repeating this structure for each thread.
```

The map is `null` initially, and stays `null` as long as you don't use multi-tenancy. So people relying on single-tenancy are only penalized by the extra `GenerationStates` instance, which doesn't seem so bad?

Anyway, I'll do that if you feel strongly about this.

hibernate-core/src/main/java/org/hibernate/id/enhanced/PooledLoThreadLocalOptimizer.java
yrodiere: If you look closely, when the condition was `false`, we would pass through this block and reach the synchronized` block, even for single-tenancy. This is what I'm changing.

Basically we used to synchronize whenever we needed to call the database to renew the pool. Which is completely unnecessary. More importantly, it *may* be a source of deadlock in some situations. Simplified example:

1. Thread 1 (T1) already has a connection (from a pool of size 1)
2. Thread 2 (T2) doesn't have a connection.
3. Thread 2 requests a new ID, enters the synchronization block and waits for a connection from T1.
4. Thread 1 requests a new ID, tries to enter the synchronization block and waits for T2 to leave it.
5. T1 waits for T2, T2 waits for T1: deadlock... until T2 times out waiting for a connection.

Admittedly that's an extreme scenario. Still, I think it's a valid concern that could happen even with a pool of a normal size.

To be more explicit, here is the code before:

```
if ( single-tenant ) {
  if ( we have more values in the pool ) {
    return the next value from the pool
  }
  else {
    synchronized (this) {
      get the next low value from the sequence
      return the next value from the pool
    }
  }
}
else { // multi-tenant
  synchronized (this) {
    get the pool for this tenant
    if ( we have more values in the pool ) {
      return the next value from the pool
    }
    else {
        get the next low value from the sequence
        return the next value from the pool
    }
  }
}
```

And here is the code after this change

```
if ( single-tenant ) {
  if ( we have more values in the pool ) {
    return the next value from the pool
  }
  else { // NO SYNCHRONIZATION HERE
    get the next low value from the sequence
    return the next value from the pool
  }
}
else { // multi-tenant
  synchronized (this) {
    get the pool for this tenant
    if ( we have more values in the pool ) {
      return the next value from the pool
    }
    else {
        get the next low value from the sequence
        return the next value from the pool
    }
  }
}
```

hibernate-core/src/main/java/org/hibernate/id/enhanced/PooledLoThreadLocalOptimizer.java
yrodiere: Done.

hibernate-core/src/main/java/org/hibernate/id/enhanced/PooledLoThreadLocalOptimizer.java
Sanne: ah, many thanks - I missed the case when reading yesterday. It's great you spotted this!

hibernate-core/src/main/java/org/hibernate/id/enhanced/PooledLoThreadLocalOptimizer.java
Sanne: thanks, and sorry for the nitpicking.

FWIW I wasn't feeling strongly about it, just a small suggestion but yes I think it's slightly better - even for null values, the wrapping object still needs to be allocated (and null values still need space for the pointer, even if it's null).

Sanne (approved):
Great catch! thanks

Sanne:
merged and backported to 5.4

PR #3722, opened 2021-02-09
HHH-14447 Add missing EventGroupListeners into FastSessionService and…

Sanne:
https://hibernate.atlassian.net/browse/HHH-14447

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/action/internal/EntityDeleteAction.java
+++ b/hibernate-core/src/main/java/org/hibernate/action/internal/EntityDeleteAction.java
@@ -184,20 +184,10 @@ public class EntityDeleteAction extends EntityAction {
 		final EventListenerGroup<PostDeleteEventListener> eventListeners = getFastSessionServices()
 				.eventListenerGroup_POST_COMMIT_DELETE;
 		if (success) {
-			eventListeners.fireLazyEventOnEachListener( this::newPostDeleteEvent, EntityDeleteAction::postCommitDeleteOnSuccess );
+			eventListeners.fireLazyEventOnEachListener( this::newPostDeleteEvent, PostDeleteEventListener::onPostDelete );
 		}
 		else {
 			eventListeners.fireLazyEventOnEachListener( this::newPostDeleteEvent, EntityDeleteAction::postCommitDeleteOnUnsuccessful );
-		}
-	}
-
-	private static void postCommitDeleteOnSuccess(PostDeleteEventListener listener, PostDeleteEvent event) {
-		if ( listener instanceof PostCommitDeleteEventListener ) {
-			listener.onPostDelete( event );
-		}
-		else {
-			//default to the legacy implementation that always fires the event
-			listener.onPostDelete( event );
 		}
 	}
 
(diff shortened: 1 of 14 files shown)


hibernate-core/src/main/java/org/hibernate/action/internal/EntityInsertAction.java
```diff
@@ -263,13 +250,12 @@ protected boolean cacheAfterInsert(EntityDataAccess cache, Object ck) {
 
 	@Override
 	protected boolean hasPostCommitEventListeners() {
-		final EventListenerGroup<PostInsertEventListener> group = listenerGroup( EventType.POST_COMMIT_INSERT );
+		final EventListenerGroup<PostInsertEventListener> group = getFastSessionServices().eventListenerGroup_POST_COMMIT_INSERT;
 		for ( PostInsertEventListener listener : group.listeners() ) {
```
dreab8, line 254: I think `getFastSessionServices().eventListenerGroup_POST_COMMIT_INSERT` can be `null`

dreab8: sorry looking at the code and it seems not

hibernate-core/src/main/java/org/hibernate/action/internal/EntityDeleteAction.java
```diff
@@ -166,47 +165,49 @@ protected boolean preDelete() {
...
+			eventListeners.fireLazyEventOnEachListener( this::newPostDeleteEvent, EntityDeleteAction::postCommitDeleteOnSuccess );
+		}
+		else {
+			eventListeners.fireLazyEventOnEachListener( this::newPostDeleteEvent, EntityDeleteAction::postCommitDeleteOnUnsuccessful );
+		}
+	}
+
+	private static void postCommitDeleteOnSuccess(PostDeleteEventListener listener, PostDeleteEvent event) {
+		if ( listener instanceof PostCommitDeleteEventListener ) {
+			listener.onPostDelete( event );
```
beikov, line 196: Missing cast?

beikov (changes requested):
Looking good overall, there is just one small issue I think

hibernate-core/src/main/java/org/hibernate/action/internal/EntityDeleteAction.java
Sanne: hum yes that is looking odd. I suspect I can just remove the `if` - need to look better.

dreab8: I think it should be something like
```
if ( listener instanceof PostCommitDeleteEventListener ) {
if ( success ) {

					listener.onPostDelete( event );
				}
				else {
					((PostCommitDeleteEventListener) listener).onPostDeleteCommitFailed( event );
				}
}
```

Sanne: I'll do the same I did for `PostInsertEventListener`

Sanne: pushed an additional commit

beikov (approved):

dreab8:
Thanks @Sanne and @beikov , merged on master and 5.4

PR #3723, opened 2021-02-09
Fixes for testsuite issues

beikov:
This should fix issues that our testsuite has on PostGIS, Oracle and HANA.

* Recreate PostgreSQL extensions after schema clearing
* Avoid dropping system generated sequences in Oracle cleaner
* Always drop id tables in tests
* Fix global temp table test to use actual dialect strategy
* Add `on commit delete rows` option for hana dialects to avoid https://answers.sap.com/questions/8824628/error-while-trying-to-drop-global-temporary-table.html


dreab8:
Thanks @beikov, merged upstream!

PR #3724, opened 2021-02-09
Migrate tests from org.hibernate.jpa.test to org.hibernate.orm.test.jpa (3)

jrenaat:


jrenaat:
Branched off at wrong starting point ... closing

PR #3727, opened 2021-02-11
HHH-14446 Integrate Spatial Dialect integration tests in the CI

maesenka:


docker_db.sh
```diff
@@ -20,6 +20,11 @@ postgresql_9_5() {
     docker run --name postgres -e POSTGRES_USER=hibernate_orm_test -e POSTGRES_PASSWORD=hibernate_orm_test -e POSTGRES_DB=hibernate_orm_test -p5432:5432 -d postgres:9.5
 }
 
+postgis_9_6(){
+  docker rm -f postgis || true
+  docker run --name postgis -e POSTGRES_USER=hibernate_orm_test -e POSTGRES_PASSWORD=hibernate_orm_test -e POSTGRES_DB=hibernate_orm_test -p5432:5432 -d mdillon/postgis:9.6
```
beikov, line 25: It seems this image here is "more" official: https://github.com/postgis/docker-postgis

Any reason why you chose the one from `mdillon`?

beikov (commented):
Other than the question about the PostgreSQL image, this looks good!

docker_db.sh
maesenka: You're right, those are the official ones. I just used the docker image I'm most familiar with. But after testing, I find that the official image works just as wel. I'll change this

beikov:
Thanks!

PR #3729, opened 2021-02-12
HHH-14456 Add DialectResolutionInfo#unwrap to be able to detect Cockroach

beikov:
https://hibernate.atlassian.net/browse/HHH-14456


gavinking:
Looks fine to me.

PR #3733, opened 2021-02-15
changes to allow non-primary unique key associations in HR

gavinking:
see https://github.com/hibernate/hibernate-reactive/issues/565

This is a draft, which I need to think more about. But here so at least @Sanne and @DavideD can see what I'm thinking about doing.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/entity/AbstractEntityLoader.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/entity/AbstractEntityLoader.java
@@ -61,6 +61,11 @@ public abstract class AbstractEntityLoader
 		return load( session, id, optionalObject, id, lockOptions, readOnly );
 	}
 
+	@Override
+	public Object load(Object id, SharedSessionContractImplementor session, LockOptions lockOptions) {
+		return load( session, id, null, null, lockOptions, null );
+	}
+
 	protected Object load(
 			SharedSessionContractImplementor session,
 			Object id,
--- a/hibernate-core/src/main/java/org/hibernate/loader/entity/EntityLoader.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/entity/EntityLoader.java
@@ -11,7 +11,6 @@ import org.hibernate.LockOptions;
 import org.hibernate.MappingException;
 import org.hibernate.engine.spi.LoadQueryInfluencers;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
-import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.persister.entity.OuterJoinLoadable;
 import org.hibernate.type.Type;
 
@@ -138,14 +137,6 @@ public class EntityLoader extends AbstractEntityLoader {
 					lockOptions.getTimeOut(),
 					getSQLString() );
 		}
-	}
-
-	public Object loadByUniqueKey(SharedSessionContractImplementor session, Object key) {
-		return loadByUniqueKey( session, key, null );
-	}
-
-	public Object loadByUniqueKey(SharedSessionContractImplementor session, Object key, Boolean readOnly) {
-		return load( session, key, null, null, LockOptions.NONE, readOnly );
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/loader/entity/UniqueEntityLoader.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/entity/UniqueEntityLoader.java
@@ -23,15 +23,16 @@ public interface UniqueEntityLoader {
 	 * Load an entity instance. If <tt>optionalObject</tt> is supplied,
 	 * load the entity state into the given (uninitialized) object.
 	 *
+	 * @throws HibernateException indicates problem performing the load.
+	 *
 	 * @deprecated use {@link #load(java.io.Serializable, Object, SharedSessionContractImplementor, LockOptions)} instead.
 	 */
 	@SuppressWarnings( {"JavaDoc"})
 	@Deprecated
-	Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session) throws HibernateException;
-
-	default Object load(Serializable id, Object optionalObject, SharedSessionContractImplementor session, Boolean readOnly) throws HibernateException {
-		return load( id, optionalObject, session );
-	}
+	Object load(
+			Serializable id,
+			Object optionalObject,
+			SharedSessionContractImplementor session);
 
 	/**
 	 * Load an entity instance by id.  If <tt>optionalObject</tt> is supplied (non-<tt>null</tt>,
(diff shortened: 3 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/loader/entity/AbstractEntityLoader.java
```diff
@@ -61,6 +61,11 @@ public Object load(Serializable id, Object optionalObject, SharedSessionContract
 		return load( session, id, optionalObject, id, lockOptions, readOnly );
 	}
 
+	@Override
+	public Object load(Object id, SharedSessionContractImplementor session, LockOptions lockOptions) {
+		return load( session, id, null, null, lockOptions, null );
+	}
+
```
gavinking, line 68: This is a replacement for `loadByUniqueKey()`.

hibernate-core/src/main/java/org/hibernate/loader/entity/EntityLoader.java
```diff
@@ -140,14 +139,6 @@ public EntityLoader(
...
 	}
 
-	public Object loadByUniqueKey(SharedSessionContractImplementor session, Object key) {
-		return loadByUniqueKey( session, key, null );
-	}
-
-	public Object loadByUniqueKey(SharedSessionContractImplementor session, Object key, Boolean readOnly) {
-		return load( session, key, null, null, LockOptions.NONE, readOnly );
-	}
-
```
gavinking, line 150: These really weren't doing a lot of work, and HR can't override them.

hibernate-core/src/main/java/org/hibernate/loader/entity/UniqueEntityLoader.java
```diff
@@ -60,4 +69,11 @@ default Object load(
 			Boolean readOnly) {
 		return load( id, optionalObject, session, lockOptions );
 	}
+
+	default Object load(
+			Object id,
+			SharedSessionContractImplementor session,
+			LockOptions lockOptions) {
+		throw new UnsupportedOperationException();
+	}
```
gavinking, line 78: This is a replacement for `loadByUniqueKey()`. Unimplemented by default. But in future we could extend unique key loading to other `UniqueEntityLoader`s.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -255,7 +255,7 @@
 
 	private final EntityLoaderLazyCollection loaders = new EntityLoaderLazyCollection();
 
-	private volatile Map<String,EntityLoader> uniqueKeyLoaders;
+	private volatile Map<String,UniqueEntityLoader> uniqueKeyLoaders;
```
gavinking, line 258: This change is necessary to allow HR to provide its own reactive loaders.

DavideD:
Shouldn't this be a GitHub draft?

gavinking:
Yes I meant to do that but I forgot.

gavinking:
So here's the thing: the overloaded `load()` operations of `UniqueEntityLoader` are really a bit of a mess at this point. So, should I take this as an opportunity to clean it up, or do I just add that one new `default` method and leave everything else alone? (And clean it up for H6.)

WDYT, @Sanne, @sebersole?

Sanne:
> So here's the thing: the overloaded load() operations of UniqueEntityLoader are really a bit of a mess at this point. So, should I take this as an opportunity to clean it up, or do I just add that one new default method and leave everything else alone? (And clean it up for H6.)

I guess it depends what you plan?  We can of course trust you and have no problem with small refactorings, but I'd still prefer us being highly conservative with 5.x in case you have semantic changes in mind.

N.B. I'll be mostly offline this week as I'm moving home - but I'm sure you all know what you're doing

sebersole:
This is already refactored in 6, so no worry about compatibility - its already something we planned on being not compatible.

However, these are public APIs and probably better to leave alone for 5.x unless there is a clear reason.

gavinking:
OK, fine. So then are the changes I've proposed here OK?

I added a `default` method to `UniqueEntityLoader` and removed a couple of methods from `EntityLoader`. Does anyone object to those changes?

hibernate-core/src/main/java/org/hibernate/loader/entity/EntityLoader.java
Sanne: I'll change this to deprecate them, suggesting to use `load` instead. Ok?

gavinking: Totally fine by me :)

Sanne:
I'm a bit puzzled to see that @gavinking 's changes are conflicting with patches I wrote ~5 months ago. @gavinking may I suggest to pull changes sometimes :)

gavinking:
Uh..... wha?

gavinking:
According to GH this branch is only 16 commits behind master. (I sent it two weeks ago.)

Sanne:
tracking as https://hibernate.atlassian.net/browse/HHH-14474

Sanne:
> According to GH this branch is only 16 commits behind master. (I sent it two weeks ago.)

I got confused as I had made some changes recently as well - still there were conflicts, looks like with other improvements of yours that have been merged in the meantime. FWIW this will look a little different in branch 5.4.

Sanne:
This was merged now.

gavinking:
Thanks so much @Sanne, appreciated!!

Sanne:
thanks @gavinking :)  Could you check your HR branch using a fresh build from ORM branch `5.4` ? 

I had to make some little changes here - should be equivalent to your original patch, but I'd rather verify this before we tag.

gavinking:
On it.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -2469,17 +2469,17 @@ private void internalInitSubclassPropertyAliasesMap(String path, Iterator proper
...
 			String propertyName,
 			Object uniqueKey,
-			SharedSessionContractImplementor session) throws HibernateException {
-		return getAppropriateUniqueKeyLoader( propertyName, session ).loadByUniqueKey( session, uniqueKey );
+			SharedSessionContractImplementor session) {
+		return getAppropriateUniqueKeyLoader( propertyName, session ).load( uniqueKey, session, LockOptions.NONE );
 	}
 
-	private EntityLoader getAppropriateUniqueKeyLoader(String propertyName, SharedSessionContractImplementor session) {
+	protected UniqueEntityLoader getAppropriateUniqueKeyLoader(String propertyName, SharedSessionContractImplementor session) {
```
gavinking, line 2476: @Sanne I think we might have lost this change:

`getAppropriateUniqueKeyLoader()` needs to be made visible.

Sanne: Ok, I wondered about that. Isn't access to override `createUniqueKeyLoader` good enough for HR ?

Sanne: Ok, I wondered about that. Isn't access to override `createUniqueKeyLoader` good enough for HR ?

gavinking: No i need it.

Sanne: Ok, done. Thanks!

PR #3734, opened 2021-02-15
Fix issue with ForeignKey creation

dreab8:
…g#forEachJdbcValue method

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
@@ -59,7 +59,6 @@ import org.hibernate.metamodel.mapping.CompositeIdentifierMapping;
 import org.hibernate.metamodel.mapping.EmbeddableMappingType;
 import org.hibernate.metamodel.mapping.EmbeddableValuedModelPart;
 import org.hibernate.metamodel.mapping.EntityIdentifierMapping;
-import org.hibernate.metamodel.mapping.EntityMappingType;
 import org.hibernate.metamodel.mapping.ForeignKeyDescriptor;
 import org.hibernate.metamodel.mapping.JdbcMapping;
 import org.hibernate.metamodel.mapping.ManagedMappingType;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
```diff
@@ -1048,6 +1052,30 @@ else if ( fkTarget instanceof EmbeddableValuedModelPart ) {
 		}
 	}
 
+	private static void setReferencedEmbeddableForeignKeyDescriptor(
```
beikov, line 1055: IMO this method should be removed i.e. revert to using `buildEmbeddableForeignKeyDescriptor`. Either this method does the same as `buildEmbeddableForeignKeyDescriptor`, except for calling `setForeignKeyDescriptor`, or `buildEmbeddableForeignKeyDescriptor` is using the wrong selection mapping for the target columns and it should be fixed. Or am I missing something?

beikov (changes requested):

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
dreab8: it means to make tobuildEmbeddableForeignKeyDescriptor more complex, decide about the target of referenced columns and selections.

dreab8: in this case we can just inline it

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -653,6 +653,12 @@ public int forEachDisassembledJdbcValue(Object value, Clause clause, int offset,
 
 	@Override
 	public int forEachJdbcValue(Object value, Clause clause, int offset, JdbcValuesConsumer consumer, SharedSessionContractImplementor session) {
-		return foreignKeyDescriptor.forEachJdbcValue( value, clause, offset, consumer, session );
+		return foreignKeyDescriptor.forEachJdbcValue(
+				entityMappingType.getEntityPersister().getIdentifier( value, session ),
```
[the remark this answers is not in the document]
dreab8: it does not work with `getPropertyAccess().getGetter().get(value)`

beikov:
Thanks, but as discussed, I'm closing this in favor of https://github.com/hibernate/hibernate-orm/pull/3735

PR #3735, opened 2021-02-15
Fix issues with FK creation

beikov:
Here a quick overview of the changes:

* Fix issues with inverse foreign key descriptor creation
* Use proper FK property for `ToOneAttributeMapping` instead of always using the target entity PK
* Introduce `EntityAssociationMapping#createDelayedDomainResult` for forcing proxy fetching
* Register `BulkOperationCleanupAction` after bulk operations to properly clear caches
* Rewrite `DatabaseSnapshotExecutor` to produce the state that structurally matches the entity state
* Fix natural id extraction issues

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/action/internal/BulkOperationCleanupAction.java
+++ b/hibernate-core/src/main/java/org/hibernate/action/internal/BulkOperationCleanupAction.java
@@ -7,9 +7,11 @@
 package org.hibernate.action.internal;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
+import java.util.List;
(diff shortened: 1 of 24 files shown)


dreab8 (approved):

sebersole (approved):
Really liked the changes to `BulkOperationCleanupAction` based on the SqmStatement - very clean

sebersole (changes requested):
Actually this leads to tests failures that are repeatable.  Could you take a look at why?

sebersole:
FWIW I applied my latest work for natural-id on top of this and the changes here resolve the database snapshot issues I was having

beikov:
You are right @sebersole I fixed that now. Apparently Derby does not like plain `null` as select item because it can't infer a type. I now implemented casts of plain parameters and `null` when appearing as select item.

PR #3739, opened 2021-02-17
HHH-13658 : make LazyToOne unnecessary

sebersole:
https://hibernate.atlassian.net/browse/HHH-13658


hibernate-core/src/test/java/org/hibernate/orm/test/mapping/lazytoone/Flight.java
```diff
@@ -25,7 +24,6 @@
 public class Flight {
 	@Id
 	private Integer id;
-	@Column(name = "flight_number")
```
beikov, line 28: Please revert this change. That causes an issue on Oracle as `number` is a keyword.

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/lazytoone/mappedby/InverseToOneExplicitOptionTests.java
```diff
@@ -0,0 +1,239 @@
...
+
+					// this should trigger loading the entity's base state
+					supplementalInfo.getSomething();
+					assertThat( sqlStatementInterceptor.getSqlQueries().size(), is( 1 ) );
+					assertThat(
+							supplementalInfoEnhancementMetadata.extractLazyInterceptor( supplementalInfo ),
+							instanceOf( LazyAttributeLoadingInterceptor.class )
+					);
+
+					// should not trigger a load and the `customer` reference should be an uninitialized enhanced proxy
```
beikov, line 121: Seems the comment does not reflect what is happening as you are asserting that a query is fired.

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/lazytoone/onetoone/JoinFetchedOneToOneAllowProxyTests.java
```diff
@@ -0,0 +1,197 @@
...
+
+					// access the id - should do nothing with db
+					supplementalInfo.getId();
+					assertThat( sqlStatementInterceptor.getSqlQueries().size(), is( 0 ) );
+
+					// this should trigger loading the entity's base state which should include join fetching Customer
+					supplementalInfo.getSomething();
+					assertThat( sqlStatementInterceptor.getSqlQueries().size(), is( 1 ) );
+
+					// should not trigger a load and the `customer` reference should be an uninitialized enhanced proxy
```
beikov, line 85: This comment is a bit confusing as the customer was already initialized through the access of the entity's base state. The combination `@Fetch( JOIN )` and `@ManyToOne( fetch = LAZY )` seems odd to me. Would `session.byId( SupplementalInfo.class ).load( 1 )` also cause a join? If so, IMO this is "wrong".

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/lazytoone/polymorphic/JoinFetchedPolymorphicToOneTests.java
```diff
@@ -0,0 +1,236 @@
...
+
+					System.out.println( "Order # " + order.getId() );
+					assertThat( sqlStatementInterceptor.getQueryCount(), is( 0 ) );
+
+					System.out.println( "  - amount : " + order.getAmount() );
+					// triggers load of base fetch state
+					assertThat( sqlStatementInterceptor.getQueryCount(), is( 1 ) );
+
+					final Customer customer = order.getCustomer();
+					// customer is part of base fetch state
```
beikov, line 61: Again, a confusing combination of `@ManyToOne( fetch = LAZY )` and `@Fetch( JOIN )`. Shouldn't the access to the customer trigger the loading of the customer?

hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/LazyGroupWithInheritanceTest.java
```diff
@@ -38,11 +42,25 @@
...
+
+		inTransaction(
+				(session) -> {
+					final Order loaded = session.byId( Order.class ).load( 1 );
+					assert Hibernate.isPropertyInitialized( loaded, "customer" );
+					assertThat( stats.getPrepareStatementCount(), is( 1L ) );
+					assertThat( loaded, instanceOf( PersistentAttributeInterceptable.class ) );
+					final PersistentAttributeInterceptor interceptor = ((PersistentAttributeInterceptable) loaded).$$_hibernate_getInterceptor();
+					assertThat( interceptor, instanceOf( BytecodeLazyAttributeInterceptor.class ) );
+					final BytecodeLazyAttributeInterceptor interceptor1 = (BytecodeLazyAttributeInterceptor) interceptor;
```
beikov, line 57: The method does not have a `@Test` annotation and the interceptor is unused. Did you forget to push some changes or was this added by accident?

hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/LazyGroupWithInheritanceTest.java
```diff
@@ -61,8 +79,8 @@ public void queryEntityWithAssociationToAbstract() {
 					//			The only viable solution I see would be to join to the "other side" and read the
 					//			version/discriminator[1].  But of course that means doing the join which is generally
 					//			what the application is trying to avoid in the first place
-					//expectedQueryCount.set( 1 );
-					expectedQueryCount.set( 4 );
+					expectedQueryCount.set( 1 );
```
beikov, line 82: Is the effect of this PR, that `Order#customer` is eagerly fetched now or "bytecode initialized".

beikov (changes requested):
Overall looks good, but there are a few things that are not clear to me.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
```diff
@@ -47,27 +53,89 @@ public static boolean includeInBaseFetchGroup(
...
+				// is no column for the association on the owner table - do not add
+				// the association to the base group (which would force an immediate
+				// select from the association table effectively making this
+				// association non-lazy)
+				return false;
+			}
+
+			final boolean unwrapExplicitlyRequested = toOne.isUnwrapProxy() && !toOne.isUnwrapProxyImplicit();
+
+			if ( inheritanceChecker.hasSubclasses( toOne.getReferencedEntityName() ) ) {
```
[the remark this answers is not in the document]
sebersole: IMO, no - it belongs here

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
```diff
@@ -47,27 +53,89 @@ public static boolean includeInBaseFetchGroup(
...
+				// the associated type has subclasses - we cannot use the enhanced proxy and will generate a HibernateProxy
+				if ( unwrapExplicitlyRequested ) {
+					// NO_PROXY was explicitly requested
```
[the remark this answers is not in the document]
sebersole: I think this is a general confusion.  The return here is not whether we get a proxy or not (nor what kind).  We are simply deciding whether to select the column, here the FK.  

We want to select the FK, so we return true

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/lazytoone/Flight.java
sebersole: +1 - not sure why I changed that

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/lazytoone/mappedby/InverseToOneExplicitOptionTests.java
sebersole: Copy-paste error.  You can change these fwiw ;)

sebersole: You can see by the "overall expectation" comment ~ line 82 what the expectations are

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/lazytoone/onetoone/JoinFetchedOneToOneAllowProxyTests.java
sebersole: There is a discrepancy here between what should happen (imo) and what does happen.  What does happen (this is due to nothing I changed) is that this mapping chooses JOIN (eager) and ignores the LAZY.

This comes back to something I tried explaining to you and Andrea the other day.  What I think should happen here (although this would have to wait for 6) is that we'd simply not select the FK initially and later do a join when we access that association.

I think it is "odd" because of our implementation, not because the combo does not make sense

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/lazytoone/polymorphic/JoinFetchedPolymorphicToOneTests.java
sebersole: Same conversation as above

hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/LazyGroupWithInheritanceTest.java
sebersole: It was a gap in the original test.  I was only running it in IJ apparently (which does not care if the method is annotated with `@Test`)

sebersole: For the unused variable... meh

hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/proxy/LazyGroupWithInheritanceTest.java
sebersole: It is a HibernateProxy

sebersole:
I'm making a few changes here based on the feedback (although its really all just test comments) and will push upstream

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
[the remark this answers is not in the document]
beikov: But doesn't that currently result in either a proxy initialized or eager loading of the target entity?

I thought that, by deferring the FK loading, we could avoid this. Or is what I described (proxy initialized or eager loading) not what is happening?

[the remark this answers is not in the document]
sebersole: if it has subclasses we create a HibernateProxy.  To create a HibernateProxy we need to know the FK

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/lazytoone/onetoone/JoinFetchedOneToOneAllowProxyTests.java
beikov: >What does happen (this is due to nothing I changed) is that this mapping chooses JOIN (eager) and ignores the LAZY.

That's what I suspected. Just as follow-up, are we logging a warning for this (LAZY is ignored due to `@Fetch( JOIN )` already, or do you think this is not necessary?

>This comes back to something I tried explaining to you and Andrea the other day. What I think should happen here (although this would have to wait for 6) is that we'd simply not select the FK initially and later do a join when we access that association.

I'm a bit unsure what the effect of `@Fetch( JOIN )` should be for `fetch = LAZY` associations. In 5.x we apparently mark the association as `EAGER`. Do I understand that correctly, that you would like to honor `fetch = LAZY` in 6.0 by deferring the fetching of the FK? We can continue the discussion elsewhere if you like.

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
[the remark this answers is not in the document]
beikov: Ok, I thought that we could honor the explicitly requested `NO_PROXY` by not selecting the FK, but I'm not 100% sure what the effect of doing that would be.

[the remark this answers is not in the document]
sebersole: We could, but it would mean selecting the FK separately which is something we try to avoid when we can.  In fact, it would mean 2 selects to initialize the association - 1 to select the FK and then another to select from the association table.  (see why joins would be beneficial?)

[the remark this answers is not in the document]
beikov: Ok so in 6.0 we would return false when we support doing that in 1 query?

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/lazytoone/onetoone/JoinFetchedOneToOneAllowProxyTests.java
sebersole: > > What does happen (this is due to nothing I changed) is that this mapping chooses JOIN (eager) and ignores the LAZY.
> 
> That's what I suspected. Just as follow-up, are we logging a warning for this (LAZY is ignored due to `@Fetch( JOIN )` already, or do you think this is not necessary?

We do not as far as I know.  Not sure it is super important.  In general there are many combinations of annotations where one is ignored because of others that we do not log.  

Also, here, there is the whole "this will (likely) change later" aspect.


> I'm a bit unsure what the effect of `@Fetch( JOIN )` should be for `fetch = LAZY` associations. In 5.x we apparently mark the association as `EAGER`. Do I understand that correctly, that you would like to honor `fetch = LAZY` in 6.0 by deferring the fetching of the FK? We can continue the discussion elsewhere if you like.

Yeah, I don't think this is the place to discuss that.

But in general, its more than delaying fetching the FK.  Yes we would not select the FK but when we go to initialize it we would do the join.  E.g.

```
@Entity
class Address {
    ... 
    @ManyToOne( ..., fetch=LAZY )
    @Fetch( JOIN )
    PostalCode postalCode;
}

@Entity
class PostalCode {
    ...
}

Session session = ...;

// Loads the Address "base state", which does not include `postalCode`
//     select a.id, ... from address a where a.id = 1
Address addr = session.get( Address.class, 1 );

// Access Address#postalCode which triggers initializing PostalCode
//     select a.postal_code_id, p.id, p... from address a join postal_code p on ... where a.id = 1
```

This gets into a more in-depth discussion regarding how we can specify when FKs get selected.  To me, the first level is `@LazyGroup`.  This example mapping almost makes me think we should have an "implicit" group for any many-to-one with `FetchMode#JOIN`.

But again, imo these are details outside the scope of this PR

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/spi/interceptor/EnhancementHelper.java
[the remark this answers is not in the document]
sebersole: That is one option, yes.  It comes back to the join discussion.  This becomes a very involved discussion as I think there are a lot of conditions that we simply do not account for today such as an explicit `@LazyGroup`, FetchMode#JOIN, etc

sebersole:
The recent oracle job failure is specific to this GH Actions / Docker set up.  I set up a Jenkins job and it passes fine - https://ci.hibernate.org/view/Personal%20jobs/job/steve-hibernate-orm-HHH-13658/1/console

I did not configure the job to keep resources (test reports, etc) but you can see from the console that the tests are run and pass

sebersole:
Squashing locally in prep for push upstream

beikov:
There is an issue with LockTest on Oracle XE but I haven't yet figured out the exact problem. I'm also not sure how to skip the test when running against Oracle XE.

sebersole:
I set up a job in Jenkins against Oracle and it works fine.  So yeah, maybe
a difference in XE.

But this is not the first time I've seen these jobs fail for test failures
which we cannot reproduce

PR #3740, opened 2021-02-17
HHH-12301 : Add Support for "bitand" Function to MySQL5 & HSQL Dialects

gbadner:
https://hibernate.atlassian.net/browse/HHH-12301

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/MySQL5Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/MySQL5Dialect.java
@@ -9,10 +9,12 @@ package org.hibernate.dialect;
 import java.sql.SQLException;
 import java.sql.Types;
 
+import org.hibernate.dialect.function.VarArgsSQLFunction;
 import org.hibernate.dialect.hint.IndexQueryHintHandler;
 import org.hibernate.exception.spi.TemplatedViolatedConstraintNameExtracter;
 import org.hibernate.exception.spi.ViolatedConstraintNameExtracter;
 import org.hibernate.internal.util.JdbcExceptionHelper;
+import org.hibernate.type.StandardBasicTypes;
 
 /**
  * An SQL dialect for MySQL 5.x specific features.
@@ -20,6 +22,12 @@ import org.hibernate.internal.util.JdbcExceptionHelper;
  * @author Steve Ebersole
  */
 public class MySQL5Dialect extends MySQLDialect {
+
+	public MySQL5Dialect() {
+		super();
+		// bitwise functions
+		registerFunction( "bitand", new VarArgsSQLFunction( StandardBasicTypes.LONG, "(", "&", ")" ) );
+	}
 	@Override
 	protected void registerVarcharTypes() {
 		registerColumnType( Types.VARCHAR, "longtext" );
--- a/hibernate-core/src/main/java/org/hibernate/dialect/HSQLDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/HSQLDialect.java
@@ -257,6 +257,9 @@ public class HSQLDialect extends Dialect {
 		// function templates
 		registerFunction( "concat", new VarArgsSQLFunction( StandardBasicTypes.STRING, "(", "||", ")" ) );
 
+		// bitwise functions
+		registerFunction( "bitand", new StandardSQLFunction( "bitand" ) );
+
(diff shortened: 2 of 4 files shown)


gbadner:
@dreab8 or @beikov, could one of you take a quick look at this?

beikov (approved):
Looks good to me

hibernate-core/src/main/java/org/hibernate/dialect/MySQL5Dialect.java
```diff
@@ -9,17 +9,25 @@
...
  *
  * @author Steve Ebersole
  */
 public class MySQL5Dialect extends MySQLDialect {
+
+	public MySQL5Dialect() {
+		super();
+		// bitwise functions
+		registerFunction( "bitand", new VarArgsSQLFunction( StandardBasicTypes.LONG, "(", "&", ")" ) );
+	}
```
dreab8, line 30: what about adding also `registerFunction( "bit_and", new StandardSQLFunction( "bit_and" ) );` ?
`

dreab8 (approved):

gavinking:
Wait, `bitand()` but no `bitor()`, `bitnot()`, `bitxor()`?

If this is worth doing, at least do all of them!

gavinking:
(P.S. if you look on the H6 branch, you can quickly see exactly which of these functions are supported on which databases.)

hibernate-core/src/main/java/org/hibernate/dialect/MySQL5Dialect.java
gavinking: Does MySQL even *have* a `bitand()`? According to the MySQL docs it's called `bit_and()`.

gavinking:
> (P.S. if you look on the H6 branch, you can quickly see exactly which of these functions are supported on which databases.)

I say this, because there are lots more dialects which support some variation of `bitand()` and `bit_and()`.

This patch looks _very_ incomplete to me, and I'm not even sure why we're wasting time with piecemeal fixes like this on the H5 branch, when all this is all already *properly* fixed in H6, and we should be directing efforts to trying to get H6 finished!

This is just a waste of time, IMO.

sebersole:
> This is just a waste of time, IMO.

We'd all love to focus solely on 6.0 - we == the dev team.  But the simple fact is that we need to continue to maintain 5.x for reasons we all know.

gavinking:
Well surely there's *some* maintenance required on H5, but that doesn't mean we need to be expending effort on very-marginal things like this. H6 is much more important.

gbadner:
@gavinking, @sebersole, I decided to do this PR because someone submitted a patch and test case for this particular function. 

In general, I am not comfortable defining functions in dialects without checking to see if they are actually supported in the relevant version. In this particular case, I found {{bitand}} documented in the MySQL 5.0 user guide, but not in earlier versions. For that reason, I put {{bitand}} in {{MySQL5Dialect}}.

I see now that it is defined in `MySQLDialect` in wip/6.0:
https://github.com/hibernate/hibernate-orm/blob/wip/6.0/hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java#L266 .

Has this 6.0 been test on MySQL 4.x? Maybe it doesn't matter.

In any case, I thought this would be a quick, easy fix. It is turning out to be neither, since now I'm not even sure of the proper dialect . 

I'm going to close this PR, close HHH-12301, and say that it is fixed in 6.0.

gavinking:
> someone submitted a patch and test case for this particular function.

Yeah, sure, I just think that sometimes it takes just as much time to review and fix up a user patch as it would take to write the code from scratch. I especially dislike these small patches that add one thing to one dialect without aiming for uniformity of support across all the main dialects.

> Has this fix in 6.0 been tested on MySQL 4.x? 

In `wip/6.0` I abstract over all the functions, so you write `bitand()` in HQL, but it gets translated to `bit_and()`, which if I'm not mistaken works on older versions of MySQL, as well as on MySQL 8.

gavinking:
Oh wait, the whole situation is even more complicated on MySQL. `bit_and()` and `bit_or()` are actually _aggregate_ functions on MySQL.

The correct translation of `bitand` is actually `(?1&?2)` according to the code in H6.

gbadner:
@gavinking , yes, you're correct about `bit_and` in MySQL. MySQL does not actually support a function named `bitand` at all. Starting in MySQL 5, the `&` operator is documented.

gbadner:
@gavinking, I think that 6.0 is a better place to get these details worked out.

PR #3743, opened 2021-02-19
HHH-14462 Impala dialect support for Hadoop

geonyeongkim:
https://hibernate.atlassian.net/browse/HHH-14462


beikov:
We are discussing the way forward for new Dialects in https://github.com/hibernate/hibernate-orm/discussions/3783

beikov:
As per https://hibernate.atlassian.net/browse/HHH-14693 and https://github.com/hibernate/hibernate-orm/discussions/3783 we decided not to include any dialects in 5.5 anymore for now and focus on 6.0.

IMO it would be fine to add it to 5.5 though if the Dialect for 6 qualifies for hibernate-core according to the requirements document: https://github.com/hibernate/hibernate-orm/blob/wip/6.0/dialects.adoc

PR #3744, opened 2021-02-20
PersistentCollection generics

gavinking:
I've made a good start on adding proper use of generics to the `PersistentCollection` hierarchy.

I was doing this work on the 6.0 branch, but I just noticed that it actually applies cleanly to master as well.


gavinking:
I will next take a look at extending this work to the `CollectionSemantics` hierarchy.

beikov:
Nice work, thanks!

PR #3745, opened 2021-02-22
Use java.lang.reflect.Type instead of java.lang.Class for JavaTypeDescriptor#getJavaType to support parameterized types

beikov:
Change `JavaTypeDescriptor#getJavaType` to return `java.lang.reflect.Type` to support parameterized type access. I also un-deprecated `JavaTypeDescriptor#getJavaTypeClass` which now returns the raw type of `java.lang.reflect.Type`.

The main use case is to support using Java types like e.g. `List<String>` for `varchar[]` and allow to understand the component type in functions.


sebersole:
Looks good to me, but a lot of test failures

beikov:
wip/6.0 does not yet pass all tests, but my other two PRs are trying to fix that so that we can immediately see if a PR breaks something.

* https://github.com/hibernate/hibernate-orm/pull/3735
* https://github.com/hibernate/hibernate-orm/pull/3662

PR #3746, opened 2021-02-22
improve use of generics in CollectionSemantics hierarchy

gavinking:


sebersole:
Looks good.  However the test failures seem to be real.

gavinking:
> Looks good. However the test failures seem to be real.

Ah, OK, I didn't realize. Strange that adding generics could break *behavior*.

gavinking:
> Looks good. However the test failures seem to be real.

@sebersole are you *sure* about that? I've just run the tests on five databases locally, and all fine, no failures.

gavinking:
I've rebased and repushed. Let's see if that helps.

beikov:
The tests that are failing are also failing on wip/6.0 already, so the change does not introduce test failures.

beikov:
Thanks!

gavinking:
> Thanks!

Thank *you*.

sebersole:
I'm sure the tests failed on GH Actions, yes.

But I've seen false failures on the "other" database jobs on a few of my
PRs as well.  I think those jobs may just be unreliable

PR #3749, opened 2021-02-23
HHH-14463 Upgrade to ByteBuddy 1.10.21: OOB support for JDK17

Sanne:
Meant for both master and `5.4`

 - https://hibernate.atlassian.net/browse/HHH-14463


yrodiere:
Merged, thanks.
Also, ported to master.

PR #3750, opened 2021-02-23
more efficient loading by @NaturalId

gavinking:
More efficient loading for entities with a single @NaturalId property only.

Uses a unique key EntityLoader instead of two SQL selects.

This is just a proof of concept for @sebersole to take a look at and give his feedback on.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultResolveNaturalIdEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultResolveNaturalIdEventListener.java
@@ -7,7 +7,6 @@
 package org.hibernate.event.internal;
 
 import java.io.Serializable;
-import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import org.hibernate.HibernateException;
@@ -128,12 +127,10 @@ public class DefaultResolveNaturalIdEventListener
 		final Serializable pk;
 		EntityPersister persister = event.getEntityPersister();
 		LockOptions lockOptions = event.getLockOptions();
-		if ( persister instanceof UniqueKeyLoadable
-				&& naturalIdValues.length==1 ) {
+		if ( persister instanceof UniqueKeyLoadable) {
 			UniqueKeyLoadable rootPersister = (UniqueKeyLoadable)
 					persister.getFactory().getMetamodel().entityPersister( persister.getRootEntityName() );
-			Map.Entry<String, Object> e = event.getNaturalIdValues().entrySet().iterator().next();
-			Object entity = rootPersister.loadByUniqueKey( e.getKey(), e.getValue(), lockOptions, session );
+			Object entity = rootPersister.loadByNaturalId( naturalIdValues, lockOptions, session );
 			if ( entity == null ) {
 				pk = null;
 			}
@@ -141,7 +138,7 @@ public class DefaultResolveNaturalIdEventListener
 				if ( !persister.isInstance(entity) ) {
 					throw new WrongClassException(
 							"loaded object was of wrong class " + entity.getClass(),
-							e.getKey(),
+							naturalIdValues,
 							persister.getEntityName()
 					);
 				}
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultResolveNaturalIdEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultResolveNaturalIdEventListener.java
@@ -7,11 +7,12 @@
 package org.hibernate.event.internal;
 
 import java.io.Serializable;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import org.hibernate.HibernateException;
-import org.hibernate.cache.spi.access.NaturalIdDataAccess;
-import org.hibernate.engine.spi.PersistenceContext;
+import org.hibernate.LockOptions;
+import org.hibernate.WrongClassException;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.event.spi.EventSource;
 import org.hibernate.event.spi.ResolveNaturalIdEvent;
@@ -19,6 +20,7 @@ import org.hibernate.event.spi.ResolveNaturalIdEventListener;
 import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.persister.entity.EntityPersister;
+import org.hibernate.persister.entity.UniqueKeyLoadable;
 import org.hibernate.pretty.MessageHelper;
 import org.hibernate.stat.spi.StatisticsImplementor;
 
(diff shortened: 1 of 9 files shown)


hibernate-core/src/main/java/org/hibernate/event/internal/DefaultResolveNaturalIdEventListener.java
```diff
@@ -120,29 +123,48 @@ protected Serializable loadFromDatasource(final ResolveNaturalIdEvent event) {
...
+							"loaded object was of wrong class " + entity.getClass(),
+							e.getKey(),
+							persister.getEntityName()
+					);
+				}
+				pk = persister.getIdentifier( entity, session );
+			}
+		}
+		else {
+			pk = persister.loadEntityIdByNaturalId( naturalIdValues, lockOptions, session );
```
gavinking, line 152: original implementation for more than one @NaturalId

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultResolveNaturalIdEventListener.java
```diff
@@ -120,29 +123,48 @@ protected Serializable loadFromDatasource(final ResolveNaturalIdEvent event) {
...
+
+		final Serializable pk;
+		EntityPersister persister = event.getEntityPersister();
+		LockOptions lockOptions = event.getLockOptions();
+		if ( persister instanceof UniqueKeyLoadable
+				&& naturalIdValues.length==1 ) {
+			UniqueKeyLoadable rootPersister = (UniqueKeyLoadable)
+					persister.getFactory().getMetamodel().entityPersister( persister.getRootEntityName() );
+			Map.Entry<String, Object> e = event.getNaturalIdValues().entrySet().iterator().next();
+			Object entity = rootPersister.loadByUniqueKey( e.getKey(), e.getValue(), lockOptions, session );
```
gavinking, line 136: optimized loading for case of just one @NaturalId

hibernate-core/src/test/java/org/hibernate/test/naturalid/inheritance/cache/InheritedNaturalIdNoCacheTest.java
```diff
@@ -57,10 +58,21 @@ public void testLoadExtendedByNormal() {
...
+			}
+			catch (WrongClassException expected) {
+				// expected outcome
+			}
+			catch (Exception other) {
+				throw new AssertionError(
+						"Unexpected exception type : " + other.getClass().getName(),
+						other
+				);
+			}
```
gavinking, line 75: This test was inconsistent with `InheritedNaturalIdCacheTest`, so I had to change one or the other of them.

We should at least define a consistent behavior for this!

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -2486,6 +2486,23 @@ public Object loadByUniqueKey(
 		return getAppropriateUniqueKeyLoader( propertyName, session ).loadByUniqueKey( session, uniqueKey );
 	}
 
+	public Object loadByUniqueKey(
```
sebersole, line 2489: This is only used from natural-id loading and at the moment limits the usefulness to single attribute natural-ids.

IMO this should be changed to something like

```
public Object loadByNaturalId(Object key, ...) {
    ...
}
```

Then this can be used for both single- and multi-attribute natural-ids and clean up the event listener

hibernate-core/src/main/java/org/hibernate/persister/entity/UniqueKeyLoadable.java
```diff
@@ -16,7 +17,11 @@
 	 * Load an instance of the persistent class, by a unique key other
 	 * than the primary key.
 	 */
-	Object loadByUniqueKey(String propertyName, Object uniqueKey, SharedSessionContractImplementor session);
+	Object loadByUniqueKey(String propertyName, Object uniqueKey,
```
sebersole, line 20: Format thing... if you are going to "chop down" arguments, can you please chop them all down?  

E.g.

```
	Object loadByUniqueKey(
		String propertyName,
		Object uniqueKey,
		SharedSessionContractImplementor session);

	Object loadByUniqueKey(
		String propertyName,
		Object uniqueKey, 
		LockOptions lockOptions,
		SharedSessionContractImplementor session);
```

hibernate-core/src/test/java/org/hibernate/test/naturalid/inheritance/cache/InheritedNaturalIdNoCacheTest.java
sebersole: Yep, we had this same conversation in regards to 6.

The inconsistency came down whether the id-resolution was cached or not.  Personally I prefer the exception

sebersole (changes requested):
A few suggestions

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
gavinking: Yup, if we're going to actually do this, I plan to generalize it to multiple `@NaturalId`s. However, there is a small difference between the two cases that I need to account for: `createUniqueKeyLoader()` has these lines:

```java
		if ( uniqueKeyType.isEntityType() ) {
			String className = ( (EntityType) uniqueKeyType ).getAssociatedEntityName();
			uniqueKeyType = getFactory().getMetamodel().entityPersister( className ).getIdentifierType();
		}
```

They're wrong for the case of `@NaturalId` loading.

Not quite sure what to do about this difference other than having two different methods.

hibernate-core/src/main/java/org/hibernate/persister/entity/UniqueKeyLoadable.java
gavinking: Fixed.

hibernate-core/src/test/java/org/hibernate/test/naturalid/inheritance/cache/InheritedNaturalIdNoCacheTest.java
gavinking: Fine, then my coin-toss came down on the right side.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
sebersole: You don't necessarily have to "go through" `createUniqueKeyLoader()` specifically.  In fact, not sure I see a benefit to doing that.  

You may run into similar problems with EntityLoader itself though.

gavinking:
OK, I've pushed a new implementation which correctly handles multiple `@NaturalId`s. We could actually remove the old implementation now.

You might not like `NaturalIdType` much&mdash;I don't like it myself&mdash;but it's nice that it lets use use `EntityLoader`s for loading by `@NaturalId` without much extra impact.

gavinking:
Alright, now I'm actually proposing this change. 

- It's a good performance optimization.
- It will make it really easy for my to implement `@NaturalId` support in HR.

gavinking:
WTH is this about:

```
Configure project :hibernate-agroal
41
Unable to decrypt Maven password using PlexusCipher
42
org.sonatype.plexus.components.cipher.PlexusCipherException: java.lang.ArrayIndexOutOfBoundsException
43
	at org.sonatype.plexus.components.cipher.PBECipher.decrypt64(PBECipher.java:193)
44
	at org.sonatype.plexus.components.cipher.DefaultPlexusCipher.decrypt(DefaultPlexusCipher.java:72)
45
	at org.sonatype.plexus.components.cipher.DefaultPlexusCipher.decryptDecorated(DefaultPlexusCipher.java:86)
46
	at 
```

(The tests pass locally, FTR.)

sebersole:
As you mentioned, not a fan of NaturalIdType but this is a temporary thing specific to 5.x.  And given previous Type-driven metamodel approach, probably not any real better solution.  So going to integrate this

gavinking:
Thanks Steve!

PR #3751, opened 2021-02-23
Support special parameter rendering through functions

beikov:
Changes:

* `SelfRenderingExpression`, `FunctionRenderingSupport` now receive a `SqlAstTranslator` which has the capability to render nodes with a parameter inlining mode
* Fix Derby rendering modes for `length`
* Use standard pipe operator for concatentation on Derby
* Implement `lpad` and `rpad` emultations for Derby to properly handle rendering mode
* Use pipe operator for concatenation on Spanner
* Don't infer type of parameter expression as function argument through function context
  => Previously `intField = func( ? )` would infer that the parameter has type int


hibernate-core/src/main/java/org/hibernate/dialect/function/TimestampdiffFunction.java
```diff
@@ -58,7 +59,7 @@ public void render(
 				TypeConfiguration.getSqlTemporalType( to.getExpressionType() )
 		);
 
-		new PatternRenderer( pattern ).render( sqlAppender, arguments, walker );
+		new PatternRenderer( pattern, SqlAstNodeRenderingMode.DEFAULT ).render( sqlAppender, arguments, walker );
```
gavinking, line 62: Since you almost always seem to pass `SqlAstNodeRenderingMode.DEFAULT` to `new PatternRenderer()`, wouldn't it make more sense to overload the constructor?

hibernate-core/src/main/java/org/hibernate/sql/ast/SqlAstNodeRenderingMode.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.sql.ast;
+
+/**
+ * @author Christian Beikov
+ */
+public enum SqlAstNodeRenderingMode {
```
gavinking, line 12: Would appreciate a bit more Javadoc here, explaining why we need it, what functionality it relates to, etc.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -90,6 +101,84 @@ public void visitAssignment(Assignment assignment) {
...
+						case Types.TINYINT:
+						case Types.INTEGER:
+						case Types.BIGINT:
+						case Types.DOUBLE:
+						case Types.REAL:
+						case Types.FLOAT:
+						case Types.NUMERIC:
+						case Types.DECIMAL:
+							sqlAstNode.accept( this );
+							appendSql( "+0" );
```
gavinking, line 128: you don't need to add any parens to the generated SQL?

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
```diff
@@ -2632,7 +2633,11 @@ private void visitLiteral(Literal literal) {
 
 				final LiteralAsParameter<Object> jdbcParameter = new LiteralAsParameter<>( literal );
 				if ( clauseStack.getCurrent() == Clause.SELECT && dialect.requiresCastingOfParametersInSelectClause() ) {
-					castFunction().render( this, Collections.singletonList( jdbcParameter ), this );
+					castFunction().render(
+							this,
+							Collections.singletonList( jdbcParameter ),
+							(SqlAstTranslator<?>) this
```
gavinking, line 2639: Ugh, that's ugly. I feel like it should not be necessary. What's the error?

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -90,6 +101,84 @@ public void visitAssignment(Assignment assignment) {
...
+			case NO_PLAIN_PARAMETER:
+				if ( sqlAstNode instanceof SqmParameterInterpretation ) {
+					sqlAstNode = ( (SqmParameterInterpretation) sqlAstNode ).getResolvedExpression();
+				}
+				if ( sqlAstNode instanceof JdbcParameter ) {
+					final JdbcMapping jdbcMapping = ( (JdbcParameter) sqlAstNode ).getExpressionType().getJdbcMappings()
+							.get( 0 );
+					// We try to avoid inlining parameters if possible which can be done by wrapping the parameter
+					// in an expression that is semantically unnecessary e.g. numeric + 0 or concat with an empty string
+					switch ( jdbcMapping.getSqlTypeDescriptor().getJdbcTypeCode() ) {
```
gavinking, line 116: Do we _really_ not have some utility function for determining if a SQL type code is numeric anywhere?

gavinking: (Or a string.)

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
beikov: The `render` method now requires a `SqlAstTranslator` rather than `SqlAstWalker`. I can try to move down this logic into `SqlAstTranslator` to avoid the cast. The `SqlAstTranslator` is also needed in `visitSelfRenderingPredicate` and `visitSelfRenderingExpression` for calling `renderToSql`. I could move these methods down into `SqlAstTranslator` as well.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
beikov: We don't, but I can introduce it if you want.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
beikov: You are right, it's safer to add parenthesis. I'll add that.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
gavinking: I'm _always_ ready  to work hard to remove a typecast.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
gavinking: I would. Makes the method shorter and more readable.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
beikov: Hmm, I could do that, but that requires a few changes again so now I am questioning if we actually need both `SqlAstWalker` and `SqlAstTranslator` i.e. why not just have one contract and thus also merge `AbstractSqlAstWalker` and `AbstractSqlAstTranslator`?

hibernate-core/src/main/java/org/hibernate/query/sqm/produce/function/internal/PatternRenderer.java
```diff
@@ -94,14 +101,16 @@ else if ( c  == '?' ) {
 			chunkList.add( chunk.toString() );
 		}
 
-		varargParam = vararg;
-		maxParamIndex = max;
+		this.varargParam = vararg;
+		this.maxParamIndex = max;
```
gavinking, line 105: why the useless `this`s?

hibernate-core/src/main/java/org/hibernate/type/descriptor/sql/SqlTypeDescriptor.java
```diff
@@ -95,6 +95,48 @@ default int getJdbcTypeCode() {
...
+			case Types.NCHAR:
+			case Types.VARCHAR:
+			case Types.NVARCHAR:
+			case Types.LONGVARCHAR:
+			case Types.LONGNVARCHAR:
+				return true;
+		}
+		return false;
+	}
+
```
gavinking, line 139: In one of my PRs I have added functions like these to `Dialect`, so we need to unify them in some `Helper` class I guess.

gavinking (commented):
OK, great, in principle it LGTM.

hibernate-core/src/main/java/org/hibernate/query/sqm/produce/function/internal/PatternRenderer.java
beikov: For consistency sake as I added a local variable for `paramIndexes` to only set the final variable `this.paramIndexes` once the array is fully initialized.

hibernate-core/src/main/java/org/hibernate/type/descriptor/sql/SqlTypeDescriptor.java
beikov: I suppose you mean the changes from https://github.com/hibernate/hibernate-orm/pull/3758 ? Yeah we need to unify that somehow. I think we can use these methods I introduced in this PR though, if we move the `getCheckCondition` method to `BasicType` like I suggested.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -456,167 +688,3155 @@ public void bindParameterValue(
...
+
+			parameterBinders.add( jdbcParameter.getParameterBinder() );
+			jdbcParameters.addParameter( jdbcParameter );
+		}
+	}
+
+	@Override
+	public void render(SqlAstNode sqlAstNode, SqlAstNodeRenderingMode renderingMode) {
+		switch ( renderingMode ) {
+			case NO_PLAIN_PARAMETER:
```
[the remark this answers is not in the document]
sebersole: Out of curiosity, why not just cast?  

If you you really think the concat / +0 approach is better, then maybe just as a better fallback to calling `#renderExpressionAsLiteral`.  

```
if ( sqlAstNode instanceof JdbcParameter ) {
    final JdbcMapping jdbcMapping = ...;

    if ( jdbcMapping.getSqlTypeDescriptor().isNumber() ) {
        ...
    }
    else if ( jdbcMapping.getSqlTypeDescriptor().isString() ) {
        ...
    }
    else {
        try {
            final String targetType = dialect.getRawTypeName( 
                    jdbcMapping.getSqlTypeDescriptor().getJdbcTypeCode() 
            );
            appendSql( "cast(? as " );
            appendSql( targetType );
            appendSql( ")" );
        }
        catch( Exception e ) {
            // the Dialect did not know a proper database type for the code...
            renderExpressionAsLiteral();
        }
}
```

If the intention really is to avoid inlining the parameters as much as possible, this helps with that

sebersole (approved):
Commented on some minor things we can discuss moving forward.  

This looks great(!) and I don't want to hold up getting it integrated upstream.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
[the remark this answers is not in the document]
beikov: You are right, I adapted that. Thanks!

PR #3747, opened 2021-02-22
expose more AbstractEntityPersister operations to HR

gavinking:
Hibernate Reactive needs access to these two methods.

PR #3752, opened 2021-02-23
Make AbstractEntityPersister#isValueGenerationRequired public

DavideD:
So that Hibernate Reactive can use it.

A simple one I think, it avoids some duplicated code in HR
Is this fine? I'm going to create an issue if there are no problems.


gavinking:
@DavideD I already did this for you :-)

gavinking:
See https://github.com/hibernate/hibernate-orm/pull/3747

DavideD:
Thanks :-)

PR #3753, opened 2021-02-24
HHH-14466 :  StackOverflowError loading an entity with eager one-to-m…

gbadner:
…any if bidirectional and many-to-one side is the ID

https://hibernate.atlassian.net/browse/HHH-14466

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/OneToOneSecondPass.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/OneToOneSecondPass.java
@@ -237,18 +237,6 @@ public class OneToOneSecondPass implements SecondPass {
 				boolean referenceToPrimaryKey  = referencesDerivedId || mappedBy == null;
 				value.setReferenceToPrimaryKey( referenceToPrimaryKey );
 
-				// If the other side is an entity with an ID that is derived from
-				// this side's owner entity, and both sides of the association are eager,
-				// then this side must be set to FetchMode.SELECT; otherwise,
-				// there will be an infinite loop attempting to load the derived ID on
-				// the opposite side.
-				if ( referencesDerivedId &&
-						!value.isLazy() &&
-						value.getFetchMode() == FetchMode.JOIN &&
-						!otherSideProperty.isLazy() ) {
-					value.setFetchMode( FetchMode.SELECT );
-				}
-
 				String propertyRef = value.getReferencedPropertyName();
 				if ( propertyRef != null ) {
 					buildingContext.getMetadataCollector().addUniquePropertyReference(
--- a/hibernate-core/src/main/java/org/hibernate/loader/plan/build/internal/FetchStyleLoadPlanBuildingAssociationVisitationStrategy.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/plan/build/internal/FetchStyleLoadPlanBuildingAssociationVisitationStrategy.java
@@ -16,13 +16,20 @@ import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.internal.CoreLogging;
 import org.hibernate.loader.plan.spi.CollectionReturn;
 import org.hibernate.loader.plan.spi.EntityReturn;
+import org.hibernate.loader.plan.spi.FetchSource;
 import org.hibernate.loader.plan.spi.LoadPlan;
 import org.hibernate.loader.plan.spi.Return;
+import org.hibernate.persister.collection.CollectionPersister;
+import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.persister.walking.spi.AssociationAttributeDefinition;
 import org.hibernate.persister.walking.spi.EncapsulatedEntityIdentifierDefinition;
 import org.hibernate.persister.walking.spi.EntityIdentifierDefinition;
 import org.hibernate.persister.walking.spi.NonEncapsulatedEntityIdentifierDefinition;
 import org.hibernate.persister.walking.spi.WalkingException;
+import org.hibernate.type.CompositeType;
+import org.hibernate.type.EmbeddedComponentType;
+import org.hibernate.type.EntityType;
+import org.hibernate.type.Type;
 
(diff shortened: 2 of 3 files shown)


beikov (approved):

gbadner:
@beikov, I just noticed that you reviewed and approved this PR after I pushed a second commit.

I realized that this issue ([HHH-14466](https://hibernate.atlassian.net/browse/HHH-14466)) is somewhat similar to [HHH-14390](https://hibernate.atlassian.net/browse/HHH-14390). 

HHH-14466 deals with a bidirectional one-to-many/many-to-one.

HHH-14390 deals with a bidirectional one-to-one association.

I think it is more appropriate to adjust the `FetchStrategy`for HHH-14390 in `FetchStyleLoadPlanBuildingAssociationVisitationStrategy#adjustJoinFetchIfNeeded` instead of setting `FetchMode.SELECT` for the association in `OneToOneSecondPass#doSecondPass` as I originally fixed it. It seems to me that the result of `OneToOneSecondPass#doSecondPass` should reflect what was actually mapped.

The second commit (https://github.com/hibernate/hibernate-orm/pull/3753/commits/12acde9c9b559ec1b742e344a444f3be29f0f88a) makes this change for HHH-14390. I'm using the same jira issue since it hasn't been released yet.

@sebersole, @beikov, what do you think about fixing these in `FetchStyleLoadPlanBuildingAssociationVisitationStrategy#adjustJoinFetchIfNeeded`?

hibernate-core/src/main/java/org/hibernate/cfg/OneToOneSecondPass.java
```diff
@@ -237,18 +237,6 @@ else if ( otherSideProperty.getValue() instanceof ManyToOne ) {
 				boolean referenceToPrimaryKey  = referencesDerivedId || mappedBy == null;
 				value.setReferenceToPrimaryKey( referenceToPrimaryKey );
 
-				// If the other side is an entity with an ID that is derived from
-				// this side's owner entity, and both sides of the association are eager,
-				// then this side must be set to FetchMode.SELECT; otherwise,
-				// there will be an infinite loop attempting to load the derived ID on
-				// the opposite side.
-				if ( referencesDerivedId &&
```
beikov, line 245: I guess this is now handled by `adjustJoinFetchIfNeeded`?

beikov (approved):
Looks good to me

hibernate-core/src/main/java/org/hibernate/cfg/OneToOneSecondPass.java
gbadner: @beikov, yes, that's correct.

hibernate-core/src/main/java/org/hibernate/loader/plan/build/internal/FetchStyleLoadPlanBuildingAssociationVisitationStrategy.java
```diff
@@ -190,11 +200,107 @@ protected FetchStrategy adjustJoinFetchIfNeeded(
...
+						collectionPersister.isInverse() ) {
+					// This is an eager "mappedBy" (inverse) side of a bidirectional
+					// one-to-many/many-to-one association
+					final EntityType elementType = (EntityType) collectionPersister.getElementType();
+					final Type elementIdType = ( (EntityPersister) elementType.getAssociatedJoinable( sessionFactory() ) ).getIdentifierType();
+					if ( elementIdType.isComponentType() && ( (CompositeType) elementIdType ).isEmbedded() ) {
+						final EmbeddedComponentType elementIdTypeEmbedded = (EmbeddedComponentType) elementIdType;
+						if ( elementIdTypeEmbedded.getSubtypes().length == 1 &&
+								elementIdTypeEmbedded.getPropertyNames()[ 0 ].equals( collectionPersister.getMappedByProperty() ) ) {
+							// The associated entity's ID is the other (many-to-one) side of the association.
```
sebersole, line 245: Could we simplify this comment a bit?  Something like

```
// The owning side of the inverse collection is defined by the associated entity's id.
// 
// Because of how Loaders process ids when processing the ResultSet, this condition would 
// lead to an infinite loop.  Adjust the fetch to use a SELECT fetch instead of JOIN to avoid
```

?

hibernate-core/src/main/java/org/hibernate/loader/plan/build/internal/FetchStyleLoadPlanBuildingAssociationVisitationStrategy.java
```diff
@@ -190,11 +200,107 @@ protected FetchStrategy adjustJoinFetchIfNeeded(
...
+							final EmbeddedComponentType associatedNonEncapsulatedIdentifierType =
+									(EmbeddedComponentType) associatedIdentifierType;
+							if ( associatedNonEncapsulatedIdentifierType.getSubtypes().length == 1 &&
+									EntityType.class.isInstance( associatedNonEncapsulatedIdentifierType.getSubtypes()[0] ) ) {
+								final EntityType otherSideEntityType =
+										( (EntityType) associatedNonEncapsulatedIdentifierType.getSubtypes()[0] );
+								if ( otherSideEntityType.isLogicalOneToOne() &&
+										otherSideEntityType.isReferenceToPrimaryKey() &&
+										otherSideEntityType.getAssociatedEntityName().equals( currentEntityPersister.getEntityName() ) ) {
+									// The associated entity's ID is the other side of the association.
```
sebersole, line 290: Similar thought about code comment here...

```
// The owning side of the inverse to-one is defined by the associated entity's id.
// 
// Because of how Loaders process ids when processing the ResultSet, this condition would 
// lead to an infinite loop.  Adjust the fetch to use a SELECT fetch instead of JOIN to avoid
```

sebersole (commented):
Minor comment about simplifying a few code comments

Otherwise, LGTM

gbadner:
@sebersole, thanks for reviewing. I've made your suggested changes and pushed to master.

PR #3755, opened 2021-02-24
HHH-14467 AnnotationException: A Foreign key refering <SomeEntity> from <SomeOtherEntity> has the wrong number of column. should be 0

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14467

See also https://github.com/quarkusio/quarkus/issues/12228

To be backported to 5.4, and maybe 5.3 if it makes sense (@gbadner, any opinion ?).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/ToOneFkSecondPass.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/ToOneFkSecondPass.java
@@ -66,19 +66,32 @@ public class ToOneFkSecondPass extends FkSecondPass {
 			//try explicit identifier property
 			return path.startsWith( property.getName() + "." );
 		}
-		else {
-			//try the embedded property
-			//embedded property starts their path with 'id.' See PropertyPreloadedData( ) use when idClass != null in AnnotationSourceProcessor
-			if ( path.startsWith( "id." ) ) {
-				KeyValue valueIdentifier = persistentClass.getIdentifier();
-				String localPath = path.substring( 3 );
-				if ( valueIdentifier instanceof Component ) {
-					Iterator it = ( (Component) valueIdentifier ).getPropertyIterator();
-					while ( it.hasNext() ) {
-						Property idProperty = (Property) it.next();
-						if ( localPath.startsWith( idProperty.getName() ) ) return true;
+		//try the embedded property
+		//embedded property starts their path with 'id.' See PropertyPreloadedData( ) use when idClass != null in AnnotationSourceProcessor
+		else if ( path.startsWith( "id." ) ) {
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/ToOneFkSecondPass.java
```diff
@@ -66,19 +66,32 @@ else if ( property != null) {
...
-				KeyValue valueIdentifier = persistentClass.getIdentifier();
-				String localPath = path.substring( 3 );
-				if ( valueIdentifier instanceof Component ) {
-					Iterator it = ( (Component) valueIdentifier ).getPropertyIterator();
-					while ( it.hasNext() ) {
-						Property idProperty = (Property) it.next();
-						if ( localPath.startsWith( idProperty.getName() ) ) return true;
+		//try the embedded property
+		//embedded property starts their path with 'id.' See PropertyPreloadedData( ) use when idClass != null in AnnotationSourceProcessor
+		else if ( path.startsWith( "id." ) ) {
```
gbadner, line 71: It is possible to have an attribute that is not mapped as `@Id` that is named "`id`". If that happens, I don't think this will work.

gbadner: Oh, I see this code is just moved from above, so maybe it's OK?

hibernate-core/src/main/java/org/hibernate/cfg/ToOneFkSecondPass.java
```diff
@@ -66,19 +66,32 @@ else if ( property != null) {
...
+					Property idProperty = (Property) it.next();
+					if ( localPath.startsWith( idProperty.getName() ) ) {
+						return true;
+					}
+				}
+			}
+		}
+		// Try the case where a @ManyToOne is also an ID property
+		// E.g. @ManyToOne @Id SomeEntity other;
+		else if ( !path.contains( "." ) ) {
```
gbadner, line 86: In the debugger, I'm seeing @OneToOne @Id from `EntityWithOneToOneDerivedId#idsource` falling into this area.

gbadner:
@yrodiere, I think it can be simplified. Please take a look at https://github.com/hibernate/hibernate-orm/pull/3760.

yrodiere:
Superseded by #3760

hibernate-core/src/main/java/org/hibernate/cfg/ToOneFkSecondPass.java
yrodiere: The comment is wrong, I also meant to cover `@OneToOne` here.

PR #3756, opened 2021-02-24
HHH-14439 QueryException: Unrecognized parameter label when executing the same query with subselects twice with different list parameters

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14439

I will backport to 5.4 once this gets merged.


sebersole (approved):
LGTM

yrodiere:
Merged, thanks!

PR #3757, opened 2021-02-24
Dialect callback to resolve SqlTypeDescriptor from JDBC type code, precision and scale

beikov:


hibernate-core/src/main/java/org/hibernate/dialect/TeradataDialect.java
```diff
@@ -120,6 +122,24 @@ public TeradataDialect(int version) {
...
 
+	@Override
+	public SqlTypeDescriptor resolveSqlTypeDescriptor(
+			int jdbcTypeCode,
+			int precision,
+			int scale,
+			SqlTypeDescriptorRegistry sqlTypeDescriptorRegistry) {
+		switch ( jdbcTypeCode ) {
+			case Types.BIT:
+				return sqlTypeDescriptorRegistry.getDescriptor( Types.BOOLEAN );
```
gavinking, line 133: Can't we get rid of stuff like this now?

hibernate-core/src/main/java/org/hibernate/dialect/TimesTenDialect.java
```diff
@@ -95,6 +97,18 @@ public int getVersion() {
...
 	}
 
+	@Override
+	public SqlTypeDescriptor resolveSqlTypeDescriptor(
+			int jdbcTypeCode,
+			int precision,
+			int scale,
+			SqlTypeDescriptorRegistry sqlTypeDescriptorRegistry) {
+		if ( jdbcTypeCode == Types.BIT ) {
+			return sqlTypeDescriptorRegistry.getDescriptor( Types.BOOLEAN );
```
gavinking, line 107: This too.

hibernate-core/src/main/java/org/hibernate/dialect/TeradataDialect.java
beikov: I don't think we can because this is for supporting native queries.

gavinking: Oh it's for reflecting on the metadata of the native query JDBC result set and guessing the Hibernate types?

OK, that makes sense.

beikov: Correct

sebersole (approved):
LGTM

PR #3758, opened 2021-02-24
move the code for generating check constraints to Dialect

gavinking:
cc @beikov this is what I meant.

I think that's cleaner, and potentially more flexible, even if perhaps not entirely necessary.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -264,6 +264,86 @@ public abstract class Dialect implements ConversionContext {
 				: size;
 	}
 
+	/**
+	 * Does the given JDBC type code represent some sort of
+	 * numeric type?
+	 * @param sqlType a JDBC type code from {@link Types}
+	 */
+	private static boolean isNumericType(int sqlType) {
+		switch (sqlType) {
+			case Types.BIT:
+			case Types.SMALLINT:
+			case Types.TINYINT:
+			case Types.INTEGER:
+			case Types.BIGINT:
+			case Types.DOUBLE:
+			case Types.REAL:
+			case Types.FLOAT:
+			case Types.NUMERIC:
+			case Types.DECIMAL:
+				return true;
+			default:
+				return false;
+		}
+	}
+
+	/**
+	 * Does the given JDBC type code represent some sort of
+	 * string type?
+	 * @param sqlType a JDBC type code from {@link Types}
+	 */
+	private static boolean isCharacterType(int sqlType) {
+		switch (sqlType) {
+			case Types.CHAR:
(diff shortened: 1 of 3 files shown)


beikov:
Not sure what is wrong there, but there are many test failures.

gavinking:
OK, lemme see what's going on.

gavinking:
> OK, lemme see what's going on.

Yeah, OK, no need to panic. What's going on is I'm a retard, as usual.

gavinking:
@beikov I believe it's fixed, despite the remaining failure on Derby.

beikov:
Looking good now! No new failures when comparing to wip/6.0 so this is fine. I think though, that the `getCheckCondition` method should be part of `org.hibernate.type.BasicType` to allow generating check conditions for other types as well. I understand your argument about SQL generation shouldn't be part of the `JavaTypeDescriptor` so what do you think about moving this to `org.hibernate.type.BasicType`?

gavinking:
> Looking good now! No new failures when comparing to wip/6.0 so this is fine. I think though, that the `getCheckCondition` method should be part of `org.hibernate.type.BasicType` to allow generating check conditions for other types as well. I understand your argument about SQL generation shouldn't be part of the `JavaTypeDescriptor` so what do you think about moving this to `org.hibernate.type.BasicType`?

I *think* I thought it should go on `JavaTypeDescriptor` so that other types like user types could take advantage of this, but it's possible I got that wrong.

WDYT, @sebersole?

sebersole:
> > the `getCheckCondition` method should be part of `org.hibernate.type.BasicType` to allow generating check conditions for other types as well. I understand your argument about SQL generation shouldn't be part of the `JavaTypeDescriptor` so what do you think about moving this to `org.hibernate.type.BasicType`?
> 
> I _think_ I thought it should go on `JavaTypeDescriptor` so that other types like user types could take advantage of this, but it's possible I got that wrong.
> 
> WDYT, @sebersole?

I'd have to pull this locally and see exactly what is calling this functionality but my initial thought was adding this as part of `BasicValuedModelPart`.  It would have access to all of the needed information.  And unless I miss something, we'd only ever issue constraints for basic types so that ModelPart would be proper.

What that does under the covers is the question.  I'd be inclined to make the method on `BasicValuedModelPart` default since it seems like the determination here would be "static".  I like it on Dialect if this varies by Dialect; but IIUC that's not what happens ATM and I cannot think if a case where it would.

Personally I'd have that delegate to JTD

gavinking:
I'm going to merge this. After rethinking it I decided that it really is cleaner and more flexible to move the part of this that renders SQL to the `Dialect` hierarchy.

I had a look at `ModelPart` and couldn't quite figure out what Steve is thinking about that. @sebersole if you want to move the contract from the descriptor to `ModelPart`, it's best you do it.

PR #3759, opened 2021-02-24
natural-id caching (on top of #3735)

sebersole:
natural-id caching 

- on top of Christian's PR #3735 which fixes a problem with pulling entity snapshots from the database which effects natural-id handling and caused test failures here


beikov (approved):
Your changes look good to me, but needs to be rebased as I merge my PR for the FK fix already with a fix for the failures that you saw.

PR #3760, opened 2021-02-25
HHH-14467 :  AnnotationException: A Foreign key refering <SomeEntity> from <SomeOtherEntity> has the wrong number of column. should be 0 #3755

gbadner:
https://hibernate.atlassian.net/browse/HHH-14467

An alternative to https://github.com/hibernate/hibernate-orm/pull/3755

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/ToOneFkSecondPass.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/ToOneFkSecondPass.java
@@ -79,7 +79,7 @@ public class ToOneFkSecondPass extends FkSecondPass {
 				Iterator it = ( (Component) valueIdentifier ).getPropertyIterator();
 				while ( it.hasNext() ) {
 					Property idProperty = (Property) it.next();
-					if ( localPath.startsWith( idProperty.getName() ) ) {
+					if ( localPath.equals( idProperty.getName() ) || localPath.startsWith( idProperty.getName() + "." ) ) {
 						return true;
 					}
 				}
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/ToOneFkSecondPass.java
```diff
@@ -67,29 +67,19 @@ else if ( property != null) {
...
-		// Try the case where a @ManyToOne is also an ID property
-		// E.g. @ManyToOne @Id SomeEntity other;
-		else if ( !path.contains( "." ) ) {
-			KeyValue valueIdentifier = persistentClass.getIdentifier();
-			if ( valueIdentifier instanceof Component ) {
 				Iterator it = ( (Component) valueIdentifier ).getPropertyIterator();
 				while ( it.hasNext() ) {
 					Property idProperty = (Property) it.next();
-					if ( path.equals( idProperty.getName() ) ) {
+					if ( localPath.startsWith( idProperty.getName() ) ) {
```
yrodiere, line 82: My main concern with this `startsWith` business, and the reason I used `equals`, is that a path can start with the name of an ID property without being related to it.

For example, IIUC, the following model would trigger a false positive for the association `EmployeeRecord.employeeStatus`: it's not in the ID, but it would be detected as such because it starts with `employee`, which is an ID property.

```
@Entity
public class EmployeeRecord {

    @Id
    @ManyToOne
    Company company;

    @Id
    @ManyToOne
    Employee employee;

    @ManyToOne
    EmployeeStatus employeeStatus;

}
```

Maybe we should go with something like this instead?

```suggestion
					if ( localPath.equals( idProperty.getName() ) || localPath.startsWith( idProperty.getName() + "." ) ) { 
```

yrodiere (commented):
The code is cleaner, but I'm not sure it's more correct... see my comment.

gbadner:
@yrodiere, thanks for the suggestion! I just added your suggested commit. I'm going to do some testing...

gbadner:
@yrodiere , this looks fine to me. I think it's OK to push to master and 5.4. I'm not going to backport this to 5.3 for now.

yrodiere:
Thanks. I clarified the commit messages and merged. I will open another PR to launch the tests on 5.4.

PR #1013, opened 2015-07-19
HHH-9952 Formula in join columns lead to several errors

racon-am:
pull-request for https://hibernate.atlassian.net/browse/HHH-9952

PR #3761, opened 2021-02-25
HHH-9952 Formula in join columns lead to several errors

dreab8:
https://hibernate.atlassian.net/browse/HHH-9952

This is a rebased version of https://github.com/hibernate/hibernate-orm/pull/1013.

I have also applied some changes to the original PR :
- I changed the `AbstractCollectionPersister#columnNames` impl
- I made a little change to the `extractMutationTexts` and moved it to the `AbstractSelectExpression` (not sure if this is the right place)

I'm not entirely sure about the impact of the change to `AbstractEntityPersister#toColumns(String propertyName)`

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/AbstractSelectExpression.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/AbstractSelectExpression.java
@@ -6,9 +6,15 @@
  */
 package org.hibernate.hql.internal.ast.tree;
 
+import java.util.Arrays;
+
+import org.hibernate.HibernateException;
+import org.hibernate.hql.internal.antlr.HqlSqlTokenTypes;
+import org.hibernate.internal.util.StringHelper;
 import org.hibernate.type.Type;
 
 import antlr.SemanticException;
+import antlr.collections.AST;
 
 /**
  * Partial implementation of SelectExpression for all the nodes that aren't constructors.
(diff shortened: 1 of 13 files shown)


hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -2821,7 +2822,23 @@ else if ( isAllOrDirtyOptLocking() && oldFields != null ) {
 					String[] propertyColumnWriters = getPropertyColumnWriters( i );
 					boolean[] propertyNullness = types[i].toColumnNullness( oldFields[i], getFactory() );
 					for ( int k = 0; k < propertyNullness.length; k++ ) {
-						if ( propertyNullness[k] ) {
+						if ( propertyColumnNames[k] == null && propertyColumnFormulaTemplates[k] != null ) {
+							String formula = StringHelper.replace(
+									propertyColumnFormulaTemplates[k][i],
```
beikov, line 2827: Shouldn't this be the other way around? `i` is the property index and `k` seems to be the column span index for the current property.

```suggestion
									propertyColumnFormulaTemplates[i][k],
```

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -2821,7 +2822,23 @@ else if ( isAllOrDirtyOptLocking() && oldFields != null ) {
 					String[] propertyColumnWriters = getPropertyColumnWriters( i );
 					boolean[] propertyNullness = types[i].toColumnNullness( oldFields[i], getFactory() );
 					for ( int k = 0; k < propertyNullness.length; k++ ) {
-						if ( propertyNullness[k] ) {
+						if ( propertyColumnNames[k] == null && propertyColumnFormulaTemplates[k] != null ) {
```
beikov, line 2825: Same issue as below, the first index to use for `propertyColumnFormulaTemplates` should be the property index.

```suggestion
						if ( propertyColumnNames[k] == null && propertyColumnFormulaTemplates[i] != null ) {
```

beikov (changes requested):
The changes look ok to me mostly, but there is a test failure. The test `org.hibernate.test.subselect.CompositeIdTypeBindingTest` fails on databases that support row value constructors, like e.g. PostgreSQL.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
dreab8: You are right, thanks @beikov I'm going to fix it

dreab8:
Hi @beikov I'm working on that

dreab8:
the issue should be fixed

beikov (approved):

gavinking:
This looks like it might impact HR, need to look more closely.

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/AbstractSelectExpression.java
```diff
@@ -55,4 +61,40 @@ public void setScalarColumn(int i) throws SemanticException {
 	public int getScalarColumnIndex() {
 		return scalarColumnIndex;
 	}
+
+	protected static String[] extractMutationTexts(Node operand, int count) {
```
NathanQingyangXu, line 65: Do we have reason to use the combination of 'protected' + 'static'? I even googled for this: https://www.google.com/search?q=protected+static+java+method&rlz=1C5CHFA_enCA943CA943&oq=protected+static+java+method&aqs=chrome..69i57j0i22i30l3.4244j1j7&sourceid=chrome&ie=UTF-8

dreab8:
I created a test using annotation instead of the deprecated hbm mapping showing the issue has been resolved in recent Hibernate versions.

https://github.com/hibernate/hibernate-orm/pull/9450

PR #3762, opened 2021-02-25
clean up the org.hibernate.annotations

gavinking:
For Hibernate 6:

- remove deprecated `@Entity` annotation
- remove deprecated  `AccessType` annotation
- remove deprecated `@Sort` annotation
- remove deprecated and unused members of `QueryHints`
- remove deprecated `FlushModeType.NEVER`
- remove deprecated `CascadeType.EVICT`
- deprecate `@Tuplizer` since `EntityTuplizer` is deprecated
- get rid of useless members of `@Tuplizer` annotation

I'm also considering whether we should remove the deprecated `@Index` and `@ForeignKey` annotations, and we need to discuss this.


sebersole (commented):
In general, things that have alternatives I think it is fine to remove.  E.g. we broke down all of the info on `@Entity` into separate annotations.  So there is already a way to specify all of the things on `@Entity`.  I think that is the pattern (aside, ofc, from things like Tuplizer that we are simply dropping).

Some of these however do not fit that and we should discuss...

- `@AccessType` - it is true users should prefer the JPA `@Access` annotation.  However we support more access types than JPA defines.  Either we say its not valid for users to specify those "extra" access types via annotations or leave this
- I vote to remove `@Tuplizer` as well.

gavinking:
> * `@AccessType` - it is true users should prefer the JPA `@Access` annotation.  However we support more access types than JPA defines.  Either we say its not valid for users to specify those

Well, OK, but then if that's the case `@Access` shouldn't be deprecated, or we should be providing something to replace it.

sebersole:
> > * `@AccessType` - it is true users should prefer the JPA `@Access` annotation.  However we support more access types than JPA defines.  Either we say its not valid for users to specify those
> 
> Well, OK, but then if that's the case `@Access` shouldn't be deprecated, or we should be providing something to replace it.

Everything before the but is bs as they say...

Ok true, but the deprecation message says it as well.

gavinking:
@sebersole OK now I'm just confused. The deprecation message says to use `@AttributeAccessor`, which appears on its surface to be a superset of `@AccessType`.

Or are there things you can do with `@AccessType` that you can't do with `@AttributeAccessor`?

PR #3764, opened 2021-02-25
remove extra-lazy collection fetching

gavinking:
This is a proposal to remove support for `@LazyCollection` and in particular `@LazyCollection(EXTRA)`.

Important note: as a side-effect we lost most of the support for "queued" collection operations, something that someone (@gbadner?) clearly put some work into.

Am I right in understanding that queued puts were only for extra-lazy collections, or did I screw up?


gavinking:
A thought to potentially rescue some of this functionality: instead of having collections being defined as extra-lazy, we could allow certain operations to be performed via static methods of the `Hibernate` class, e.g:

    Hibernate.getSize(my.getItems())

Or:

    Hibernate.contains(my.getItems(), item)

Or

    Hibernate.remove(my.getItems(), item)

Or:

    Hibernate.put(my.getMap(), key, value)

This would be much more useful than the previous functionality, though of course I would not be able to delete quite as much code.

Thoughts?

gavinking:
So apparently there is some evidence that this is a feature some people actually use.

If that's the case, I don't think it's critical to remove it, and I'm withdrawing this proposal for now.

PR #3766, opened 2021-03-01
HHH-14471 Fix concurrency issue due to builder sharing in DynamicBatchingEntityLoader

beikov:
https://hibernate.atlassian.net/browse/HHH-14471


Sanne (commented):
Good catch! Looks urgent, but I'll wait for the user to confirm.

BTW I'm a little worried about the allocation cost of the `EntityLoader` now happening on each operation. Perhaps this could be polished further by splitting the object into an immutable-and-reusable component, and having `idsToLoad.lenght` as only variable input?

I'd be 100% ok for that to be a 6-exclusive improvement - or to defer it as I don't mean to side-track your priorities.

beikov:
I did this to fix a problem a user reported for 5.4. The EntityLoader was created on-demand previously as well. The fix is just that the builder for the entity loader is now, not shared anymore.

jensborrmann:
We verified that the fix works in our use case.

Thanks for the quick reaction!

jensborrmann:
Could you please confirm that the changes will be applied to 5.4 as well?

Sanne:
@jensborrmann yes we'll backport it.Thanks!

Sanne (approved):

PR #3770, opened 2021-03-01
type ids as Object instead of Serializable

gavinking:
Here's a draft.

I have done this without impacting user-visible APIs like `Interceptor` and `Tuplizer`, but we need to decide what to do about those.


beikov (approved):
Looks good to me. I think it's also a good idea to change this in `Interceptor` and `Tuplizer`.

gavinking:
> Looks good to me. I think it's also a good idea to change this in `Interceptor` and `Tuplizer`.

Well yeah but that's a breaking change to user code, so we need to discuss it first.

gavinking:
OK, this now also incorporates the change to the `EntityTuplizer` methods.

gavinking:
OK, I've now fixed up the `Interceptor` and `Lifecycle` interfaces. This is now good to go, it seems to me.

sebersole (approved):

PR #3774, opened 2021-03-02
HHH-14473 Resolve managed class name with class loader as well

JohnNiang:
Related to https://hibernate.atlassian.net/browse/HHH-14473

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/process/internal/ScanningCoordinator.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/process/internal/ScanningCoordinator.java
@@ -30,9 +30,9 @@ import org.hibernate.boot.internal.ClassLoaderAccessImpl;
 import org.hibernate.boot.jaxb.Origin;
 import org.hibernate.boot.jaxb.SourceType;
 import org.hibernate.boot.registry.classloading.spi.ClassLoaderService;
+import org.hibernate.boot.registry.classloading.spi.ClassLoadingException;
 import org.hibernate.boot.spi.BootstrapContext;
 import org.hibernate.boot.spi.ClassLoaderAccess;
-import org.hibernate.boot.spi.MetadataBuildingOptions;
 import org.hibernate.boot.spi.XmlMappingBinderAccess;
 import org.hibernate.cfg.AttributeConverterDefinition;
 import org.hibernate.service.ServiceRegistry;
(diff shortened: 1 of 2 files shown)


Sanne:
Interesting, thanks @JohnNiang .But I don't completely understand the need; could you elaborate on which environment this was a problem, and how I could try reproduce?

JohnNiang:
Thank you for your reply.

In my case, attribute converters are generated by ByteBuddy as follow:

```java
    return new ByteBuddy()
        .with(new NamingStrategy.AbstractBase() {
          @Override
          protected String name(TypeDescription superClass) {
            return superClass.getName()
                + "$" + clazz.getSimpleName() + "$"
                + AUTO_GENERATION_SUFFIX;
          }
        })
        .subclass(
            parameterizedType(AbstractEnumConverter.class, clazz, Integer.class)
                .build())
        .annotateType(ofType(Converter.class).define("autoApply", true).build())
        .make()
        .load(classLoader, ClassLoadingStrategy.Default.INJECTION)
        .getLoaded();
```

If I put the auto-generated class names of attribute converter into managed class names, `org.hibernate.boot.registry.classloading.spi.ClassLoaderService#locateResource` won't find them correctly, because auto-generated classes aren't exist in the class path.

The moment, I have to delegate the `ClassLoaderService` to complete those things as follow:

```java
  @Override
  public URL locateResource(String name) {
    var url = delegate.locateResource(name);
    if (log.isTraceEnabled()) {
      log.trace("Located resource: {}, url: {}", name, url);
    }
    if (StringUtils.endsWithIgnoreCase(name, ".class")
        && !StringUtils.endsWithIgnoreCase(name, "/package-info.class")) {
      // if the name ends with .class, then trying to load by class
      var className = name.replace(".class", "").replace('/', '.');
      if (AttributeConverterAutoGenerator.isGeneratedByByteBuddy(className)) {
        // handle for class generated by byte buddy.
        try {
          ClassUtils.resolveClassName(className, this.classLoader);
          url = new URL("file:///auto-generated/" + name);
        } catch (IllegalArgumentException | IllegalStateException ignored) {
          // ignore this exception
        } catch (MalformedURLException e) {
          log.error("Failed to create url of name " + name, e);
        }
      }
    }
    return url;
  }
```

JohnNiang:
@Sanne I'll provide a test case as soon as possible.

Sanne:
Thanks for the explanation! Makes sense. And yes, a test would be great.

Sanne:
Many thanks @JohnNiang , I've reabased and merged this.

PR #3775, opened 2021-03-02
HHH-11076 Log a warning if uninitialized collection unsets session when filters are enabled

beikov:
This is the implementation that was discussed during the PR meet-up. Note that I also provided an alternative implementation that snapshots the load query influencers on detach: https://github.com/hibernate/hibernate-orm/pull/3776

https://hibernate.atlassian.net/browse/HHH-11076


sebersole (approved):
LGTM

PR #3776, opened 2021-03-02
HHH-11076 Snapshot load query influencers on detach and apply in temporary session

beikov:
This is a possible solution for the problem as described in the issue. The alternative, which is what has been discussed in the PR meeting, is to log a warning instead: https://github.com/hibernate/hibernate-orm/pull/3775

https://hibernate.atlassian.net/browse/HHH-11076

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/AbstractPersistentCollection.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/AbstractPersistentCollection.java
@@ -25,6 +25,7 @@ import org.hibernate.collection.spi.PersistentCollection;
 import org.hibernate.engine.internal.ForeignKeys;
 import org.hibernate.engine.spi.CollectionEntry;
 import org.hibernate.engine.spi.EntityEntry;
+import org.hibernate.engine.spi.LoadQueryInfluencersSnapshot;
 import org.hibernate.engine.spi.PersistenceContext;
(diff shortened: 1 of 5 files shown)


sebersole:
Simply copying my replies from Zulip:

> As I mentioned when we discussed this, its not just when that setting is enabled.
>
> if the argument is that logically the filters in effect when the PersistentCollection was created, then that needs to be consistent and covers cases like possible later merges
>
> we talked about "well the merge Session should have the same filters...". that may be true, but also its a logical inconsistency in how these collections are handled
>
> in other words, you'd have to do this snapshoting whenever the collection is detached. and actually, even that is not enough to be logically consistent... really you'd have to take the snapshot when the PersistentCollection is created!
>
> i'll simply re-iterate that I think this is a terrible idea

Steve Ebersole: (not the logging, I think that is great)

beikov:
As discussed with @sebersole on Zulip, we are not going to implement this in this way. If we were to work on this at all, we'd have to capture the load influencers for every collection at initialization time already and honor that for all operations which is an overkill. Closing this in favor of https://github.com/hibernate/hibernate-orm/pull/3775

sebersole (changes requested):
Christian and I discussed this on Zulip.  *If* this is something we really want to do (and I personally question this) I think this is a "half solution" and we should handle all cases

PR #3781, opened 2021-03-02
make @AttributeAccessor annotation typesafe

gavinking:
See discussion here:

https://github.com/hibernate/hibernate-orm/discussions/3780#discussioncomment-423468

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/annotations/PropertyBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/annotations/PropertyBinder.java
@@ -41,6 +41,7 @@ import org.hibernate.mapping.RootClass;
 import org.hibernate.mapping.SimpleValue;
 import org.hibernate.mapping.ToOne;
 import org.hibernate.mapping.Value;
+import org.hibernate.property.access.spi.PropertyAccessStrategy;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/annotations/PropertyBinder.java
```diff
@@ -276,7 +277,17 @@ public Property makeProperty() {
 
 			if ( property.isAnnotationPresent( AttributeAccessor.class ) ) {
 				final AttributeAccessor accessor = property.getAnnotation( AttributeAccessor.class );
-				prop.setPropertyAccessorName( accessor.value() );
+				String value = accessor.value();
+				Class<?> type = accessor.strategy();
+				if ( !value.equals("") ) {
```
NathanQingyangXu, line 282: This might be minor, but it is a good practice to avoid "equals()" to decide whether a string is empty or not. For instance, https://howtodoinjava.com/java/string/always-use-length-instead-of-equals-to-check-empty-string-in-java/. Definitively we can save some CPU cycles. If I were not mistaken, we have some util method (created by @Sanne ?) in v5, though not sure whether it exists in v6.

PR #3792, opened 2021-03-05
HHH-14478 : Allow DialectResolvers to be discovered by ServiceLoader

sebersole:
HHH-14478 : Allow DialectResolvers to be discovered by ServiceLoader

- Note : the changes to `org.hibernate.dialect.resolver.DialectFactoryTest` were needed to make them pass.  However, the failures were not caused by changes done in this work.  They mostly likely "broke" when the initial Dialect clean up work was done.  This test was part of the "filtered out" tests in terms of the Gradle build, so they would not show up on CI runs e.g. after those earlier changes

- Note 2 : at least locally, the test `org.hibernate.test.annotations.derivedidentities.bidirectional.OneToOneWithDerivedIdentityTest`  is failing the build even though it is outside of the filtered include package.  No idea why that test gets picked up; nor why it just started now..

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/dialect/internal/DialectResolverSet.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/dialect/internal/DialectResolverSet.java
@@ -8,6 +8,7 @@ package org.hibernate.engine.jdbc.dialect.internal;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 
 import org.hibernate.dialect.Dialect;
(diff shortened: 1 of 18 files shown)


hibernate-core/src/main/javadoc/package.html
```diff
@@ -33,7 +33,7 @@ <h3>Extensions</h3>
     <li>{@link org.hibernate.Transaction} / {@link org.hibernate.transaction.TransactionFactory}</li>
     <li>{@link org.hibernate.context.CurrentSessionContext}</li>
     <li>{@link org.hibernate.dialect.Dialect}</li>
-    <li>{@link org.hibernate.dialect.resolver.DialectResolver}</li>
+    <li>{@link org.hibernate.orm.test.dialect.resolver.DialectResolver}</li>
```
beikov, line 36: Looks like this happened by accident.

```suggestion
    <li>{@link org.hibernate.dialect.resolver.DialectResolver}</li>
```

hibernate-core/src/main/java/org/hibernate/engine/jdbc/dialect/internal/DialectResolverSet.java
```diff
@@ -60,23 +61,19 @@ public Dialect resolveDialect(DialectResolutionInfo info) {
...
-	 * priority than any other existing resolvers.
-	 *
-	 * @param resolver The resolver to add.
-	 */
-	public void addResolverAtFirst(DialectResolver resolver) {
-		resolvers.add( 0, resolver );
+	public void addResolverAtFirst(DialectResolver... resolvers) {
+		// in reverse so that the first item in the incoming list is added
+		// to the first position in `this.resolutions`
+		for ( int i = resolvers.length - 1; i >= 0; i-- ) {
```
beikov, line 71: Minor, but could be done by using `this.resolvers.addAll( 0, Arrays.asList( resolvers ) );`

beikov (commented):
Looks good!

beikov:
> Note 2

I just fixed that. After a rebase this should be fine.

sebersole:
> I just fixed that. After a rebase this should be fine.

curiosity sake... what was the issue?

hibernate-core/src/main/javadoc/package.html
sebersole: Strange.  Thanks for the fix

sebersole:
Squashed, rebased and applied upstream

gavinking:
Excellent!

beikov:
@sebersole actually `org.hibernate.orm.test.sql.exec.onetoone.OneToOneWithDerivedIdentityTest` failed due to the `NotImplementedYetExtension` running it's checks only after test execution, but the test already failed in the "before-lifecycle" because the EntityManagerFactory/SessionFactory was built eagerly. I switched to lazy building which now causes the failure to happen during the test run and the `NotImplementedYetExtension` to work properly.

PR #3793, opened 2021-03-06
HHH-14482 Do not discard prior implicit join by key

Naros:
https://hibernate.atlassian.net/browse/HHH-14482


Naros:
While this PR solves the concern with the specific sub-query in the associated jira, there are other sub-query use cases where this solution doesn't account for and therefore the following exception continues to be raised:
```
Caused by: org.hibernate.sql.ast.SqlTreeCreationException: Could not locate TableGroup - StringListEntity_strings_AUD(ee__).originalId.REV
	at org.hibernate.sql.ast.spi.FromClauseAccess.getTableGroup(FromClauseAccess.java:35)
```
I'll continue to refine this PR.

dreab8:
Thanks @Naros,

 I'm closing this one in favour of https://github.com/hibernate/hibernate-orm/pull/3807

PR #3794, opened 2021-03-07
Don't add superfluous "and" when there is no discriminator filter

Naros:
When `@WhereJoinTable` gets applied to a collection, a superfluous `and` is added to the rendered SQL unnecessarily when no discriminator filter is rendered as part of the predicate condition.


dreab8:
Thanks @Naros

PR #3795, opened 2021-03-07
Resolve JavaTypeDescriptor for StandardRowReader

Naros:
Assuming a simple query as follows:
```
session.createNativeQuery( "SELECT group_id FROM GroupMember" ).addScalar( "group_id", IntegerType.INSTANCE )
```
This led to a NPE in the `StandardRowReader#getResultJavaType` method because the underlying `ResultAssembler` was passed a null JavaTypeDescriptor.  This change guarantees that if no explicit JavaTypeDescriptor is present but a type has been resolved, we resolve the JavaTypeDescriptor from the resolved type.


dreab8:
Thank @Naros

PR #3796, opened 2021-03-07
Fix NPE with DelayedCollectionInitializer

Naros:
A NPE is thrown due to `DelayedCollectionInitializer` relying on the `AbstractCollectionInitializer` behavior for `resolveKey`.  The overridden method does not register the resolution listener since the instance hasn't yet been resolved but only resolves the key.


dreab8:
Thanks @Naros

PR #3797, opened 2021-03-08
HHH-14484 Improve support for Firebird 4

mrotteveel:
Also includes:
- Some general Firebird improvements
- Tweak tests to run (or be skipped) against Firebird
- Fix to BooleanDecoder
- Add statistical/linear regression function definitions in CommonFunctionFactory
- Added the relevant functions to other dialects where I could verify they apply
- Fix for unions with parenthesized selects
- Add casts for parameters in functions

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
@@ -186,6 +186,7 @@ public class DB2Dialect extends Dialect {
 		CommonFunctionFactory.median( queryEngine );
 		CommonFunctionFactory.stddev( queryEngine );
 		CommonFunctionFactory.stddevPopSamp( queryEngine );
+		CommonFunctionFactory.regrLinearRegressionAggregates( queryEngine );
 		CommonFunctionFactory.variance( queryEngine );
 		CommonFunctionFactory.stdevVarianceSamp( queryEngine );
 		CommonFunctionFactory.addYearsMonthsDaysHoursMinutesSeconds( queryEngine );
--- a/hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
@@ -8,6 +8,7 @@ package org.hibernate.dialect;
 
 import org.hibernate.HibernateException;
 import org.hibernate.NotYetImplementedFor6Exception;
+import org.hibernate.NullOrdering;
 import org.hibernate.cfg.Environment;
 import org.hibernate.dialect.function.CommonFunctionFactory;
 import org.hibernate.dialect.identity.FirebirdIdentityColumnSupport;
@@ -34,6 +35,7 @@ import org.hibernate.metamodel.spi.RuntimeModelCreationContext;
 import org.hibernate.query.CastType;
 import org.hibernate.query.TemporalUnit;
 import org.hibernate.query.spi.QueryEngine;
+import org.hibernate.query.sqm.function.SqmFunctionRegistry;
 import org.hibernate.query.sqm.mutation.internal.idtable.AfterUseAction;
 import org.hibernate.query.sqm.mutation.internal.idtable.GlobalTemporaryTableStrategy;
 import org.hibernate.query.sqm.mutation.internal.idtable.IdTable;
@@ -55,6 +57,7 @@ import java.sql.DatabaseMetaData;
 import java.sql.SQLException;
 import java.sql.Types;
 import java.time.temporal.TemporalAccessor;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.TimeZone;
@@ -77,7 +80,7 @@ public class FirebirdDialect extends Dialect {
 	private final int version;
 
 	public FirebirdDialect() {
-		this(250);
+		this( 250 );
 	}
 
 	public FirebirdDialect(DialectResolutionInfo info) {
@@ -87,7 +90,7 @@ public class FirebirdDialect extends Dialect {
 	// KNOWN LIMITATIONS:
 
 	// * no support for format()
-	// * extremely low maximum decimal precision (18)
+	// * (Firebird 3 and earlier) extremely low maximum decimal precision (18)
 	//   making BigInteger/BigDecimal support useless
 	// * can't select a parameter unless wrapped in a
 	//   cast (not even when wrapped in a function call)
@@ -114,18 +117,31 @@ public class FirebirdDialect extends Dialect {
 		//no precision for 'timestamp' type
 		registerColumnType( Types.TIMESTAMP, "timestamp" );
 		if ( getVersion() < 400 ) {
+			// No time zone support, map to without time zone types
 			registerColumnType( Types.TIMESTAMP_WITH_TIMEZONE, "timestamp" );
+			registerColumnType( Types.TIME_WITH_TIMEZONE, "time" );
 		}
 		else {
 			registerColumnType( Types.TIMESTAMP_WITH_TIMEZONE, "timestamp with time zone" );
 		}
+		// Single byte character sets can be 32_765 characters, but assume use of UTF8
 		registerColumnType( Types.VARCHAR, 8_191, "varchar($l)" );
 		registerColumnType( Types.VARCHAR, "blob sub_type text" );
 
-		registerColumnType( Types.BINARY, 32_767, "char($l) character set octets" );
+		if ( getVersion() < 400 ) {
+			registerColumnType( Types.BINARY, 32_767, "char($l) character set octets" );
+		}
+		else {
+			registerColumnType( Types.BINARY, 32_767, "binary($l)" );
+		}
 		registerColumnType( Types.BINARY, "blob sub_type binary" );
 
-		registerColumnType( Types.VARBINARY, 32_765, "varchar($l) character set octets" );
+		if ( getVersion() < 400 ) {
+			registerColumnType( Types.VARBINARY, 32_765, "varchar($l) character set octets" );
+		}
+		else {
+			registerColumnType( Types.VARBINARY, 32_765, "varbinary($l)" );
+		}
 		registerColumnType( Types.VARBINARY, "blob sub_type binary" );
 
 		registerColumnType( Types.BLOB, "blob sub_type binary" );
@@ -166,14 +182,29 @@ public class FirebirdDialect extends Dialect {
 
 	@Override
 	public String getTypeName(int code, Size size) throws HibernateException {
-		//precision of a Firebird 'float(p)' represents
-		//decimal digits instead of binary digits
-		return super.getTypeName( code, binaryToDecimalPrecision( code, size ) );
+		if ( getVersion() < 400 ) {
+			//precision of a Firebird 3 and earlier 'float(p)' represents
+			//decimal digits instead of binary digits
+			return super.getTypeName( code, binaryToDecimalPrecision( code, size ) );
+		}
+		else {
+			// Firebird 4 and higher supports standard 'float(p)' (with precision in binary digits)
+			return super.getTypeName( code, size );
+		}
 	}
 
 	@Override
 	public int getFloatPrecision() {
-		return 21; // -> 7 decimal digits
+		return getVersion() < 400
+				? 21 // -> 7 decimal digits (actually 24, but needed for Dialect#binaryToDecimalPrecision(int,size))
+				: 24;
+	}
+
+	@Override
+	public int getDefaultTimestampPrecision() {
+		// Formally, Firebird has a (fixed) precision of 4 (100 microseconds),
+		// but things like CURRENT_TIMESTAMP produce values with a maximum of 3, so report that
+		return 3;
 	}
 
 	@Override
@@ -182,6 +213,17 @@ public class FirebirdDialect extends Dialect {
 
 		CommonFunctionFactory.concat_pipeOperator( queryEngine );
 		CommonFunctionFactory.cot( queryEngine );
+		CommonFunctionFactory.cosh( queryEngine );
+		CommonFunctionFactory.sinh( queryEngine );
+		CommonFunctionFactory.tanh( queryEngine );
+		if ( getVersion() >= 300 ) {
+			CommonFunctionFactory.moreHyperbolic( queryEngine );
+			CommonFunctionFactory.stddevPopSamp( queryEngine );
+			CommonFunctionFactory.varPopSamp( queryEngine );
+			CommonFunctionFactory.covarPopSamp( queryEngine );
+			CommonFunctionFactory.corr( queryEngine );
+			CommonFunctionFactory.regrLinearRegressionAggregates( queryEngine );
+		}
 		CommonFunctionFactory.log( queryEngine );
 		CommonFunctionFactory.log10( queryEngine );
 		CommonFunctionFactory.pi( queryEngine );
@@ -198,18 +240,70 @@ public class FirebirdDialect extends Dialect {
 		CommonFunctionFactory.reverse( queryEngine );
 		CommonFunctionFactory.bitandorxornot_binAndOrXorNot( queryEngine );
 		CommonFunctionFactory.leastGreatest_minMaxValue( queryEngine );
-		CommonFunctionFactory.stddevPopSamp( queryEngine );
-		CommonFunctionFactory.varPopSamp( queryEngine );
-		//TODO: lots more statistical functions
 
-		//TODO: gen_uid() and friends, gen_id()
-
-		queryEngine.getSqmFunctionRegistry().registerBinaryTernaryPattern(
+		SqmFunctionRegistry functionRegistry = queryEngine.getSqmFunctionRegistry();
+		functionRegistry.registerBinaryTernaryPattern(
 				"locate",
 				StandardBasicTypes.INTEGER,
 				"position(?1 in ?2)",
-				"(position(?1 in substring(?2 from ?3)) + (?3) - 1)"
-		).setArgumentListSignature("(pattern, string[, start])");
+				"position(?1, ?2, ?3)"
+		).setArgumentListSignature( "(pattern, string[, start])" );
+		functionRegistry.namedDescriptorBuilder( "ascii_val" )
+				.setExactArgumentCount( 1 )
+				.setInvariantType( StandardBasicTypes.SHORT )
+				.register();
+		functionRegistry.registerAlternateKey( "ascii", "ascii_val" );
+		functionRegistry.namedDescriptorBuilder( "ascii_char" )
+				.setExactArgumentCount( 1 )
+				.setInvariantType( StandardBasicTypes.CHARACTER )
+				.register();
+		functionRegistry.registerAlternateKey( "chr", "ascii_char" );
+		functionRegistry.registerAlternateKey( "char", "ascii_char" );
+		functionRegistry.registerPattern(
+				"radians",
+				"((?1)*pi()/180e0)",
+				StandardBasicTypes.DOUBLE
+		);
+		functionRegistry.registerPattern(
+				"degrees",
+				"((?1)*180e0/pi())",
+				StandardBasicTypes.DOUBLE
+		);
+
+		if ( getVersion() >= 400 ) {
+			Arrays.asList( "md5", "sha1", "sha256", "sha512" )
+					.forEach( hash -> functionRegistry.registerPattern(
+							hash,
+							"crypt_hash(?1 using " + hash + ")",
+							StandardBasicTypes.BINARY
+					) );
+			functionRegistry.registerAlternateKey( "sha", "sha1" );
+			functionRegistry.registerPattern(
+					"crc32",
+					"hash(?1 using crc32)",
+					StandardBasicTypes.INTEGER
+			);
+		}
+	}
+
+	@Override
+	public String currentLocalTime() {
+		if ( supportsTimezoneTypes() ) {
+			return "localtime";
+		}
+		else {
+			return super.currentLocalTime();
+		}
+	}
+
+	@Override
+	public String currentLocalTimestamp() {
+		if ( supportsTimezoneTypes() ) {
+			return "localtimestamp";
+		}
+		else {
+			return super.currentLocalTimestamp();
+		}
 	}
 
 	@Override
@@ -231,10 +325,6 @@ public class FirebirdDialect extends Dialect {
 	public String castPattern(CastType from, CastType to) {
 		String result;
 		switch ( to ) {
-			case FLOAT:
-				return "cast(double(?1) as real)";
-			case DOUBLE:
-				return "double(?1)";
 			case INTEGER:
 			case LONG:
 				result = BooleanDecoder.toInteger( from );
@@ -269,7 +359,8 @@ public class FirebirdDialect extends Dialect {
 			case STRING:
 				result = BooleanDecoder.toString( from );
 				if ( result != null ) {
-					return result;
+					// trim converts to varchar to prevent padding with spaces
+					return "trim(" + result + ")";
 				}
 				break;
 		}
@@ -278,6 +369,9 @@ public class FirebirdDialect extends Dialect {
 
 	@Override
 	public long getFractionalSecondPrecisionInNanos() {
+		// Formally, Firebird can store values with 100 microsecond precision (100_000 nanoseconds).
+		// However, some functions (e.g. CURRENT_TIMESTAMP) will only return values with millisecond precision
+		// So, we report millisecond precision
 		return 1_000_000; //milliseconds
 	}
 
@@ -292,15 +386,17 @@ public class FirebirdDialect extends Dialect {
 		switch ( unit ) {
 			case DAY_OF_WEEK:
 			case DAY_OF_YEAR:
-				return "(" + super.extractPattern(unit) + "+1)";
+				return "(" + super.extractPattern( unit ) + "+1)";
+			case QUARTER:
+				return "((extract(month from ?2)+2)/3)";
 			default:
-				return super.extractPattern(unit);
+				return super.extractPattern( unit );
 		}
 	}
 
 	@Override
 	public String timestampaddPattern(TemporalUnit unit, TemporalType temporalType) {
-		switch (unit) {
+		switch ( unit ) {
 			case NATIVE:
 				return "dateadd((?2) millisecond to ?3)";
 			case NANOSECOND:
@@ -308,7 +404,7 @@ public class FirebirdDialect extends Dialect {
 			case WEEK:
 				return "dateadd((?2)*7 day to ?3)";
 			case QUARTER:
-				return "dateadd((?2)*4 month to ?3)";
+				return "dateadd((?2)*3 month to ?3)";
 			default:
 				return "dateadd(?2 ?1 to ?3)";
 		}
@@ -316,7 +412,7 @@ public class FirebirdDialect extends Dialect {
 
 	@Override
 	public String timestampdiffPattern(TemporalUnit unit, TemporalType fromTemporalType, TemporalType toTemporalType) {
-		switch (unit) {
+		switch ( unit ) {
 			case NATIVE:
 				return "datediff(millisecond from ?2 to ?3)";
 			case NANOSECOND:
@@ -324,16 +420,20 @@ public class FirebirdDialect extends Dialect {
 			case WEEK:
 				return "datediff(day from ?2 to ?3)/7";
 			case QUARTER:
-				return "datediff(month from ?2 to ?3)/4";
+				return "datediff(month from ?2 to ?3)/3";
 			default:
 				return "datediff(?1 from ?2 to ?3)";
 		}
 	}
 
 	@Override
+	public boolean supportsTemporalLiteralOffset() {
+		return getVersion() >= 400;
+	}
+
+	@Override
 	public int getDefaultDecimalPrecision() {
-		//the extremely low maximum
-		return 18;
+		return getVersion() < 400 ? 18 : 38;
 	}
 
 	@Override
@@ -348,7 +448,7 @@ public class FirebirdDialect extends Dialect {
 
 	@Override
 	public int getMaxAliasLength() {
-		return 20;
+		return getVersion() < 400 ? 20 : 52;
 	}
 
 	public IdentifierHelper buildIdentifierHelper(
@@ -356,6 +456,24 @@ public class FirebirdDialect extends Dialect {
 			DatabaseMetaData dbMetaData) throws SQLException {
 		// Any use of keywords as identifiers will result in token unknown error, so enable auto quote always
 		builder.setAutoQuoteKeywords( true );
+
+		// Additional reserved words
+		// The Hibernate list of SQL:2003 reserved words doesn't contain all SQL:2003 reserved words,
+		// and Firebird is finicky when it comes to reserved words
+		int version = getVersion();
+		if ( version >= 300 ) {
+			builder.applyReservedWords(
+					"AVG", "BOOLEAN", "CHARACTER_LENGTH", "CHAR_LENGTH", "CORR", "COUNT",
+					"COVAR_POP", "COVAR_SAMP", "EXTRACT", "LOWER", "MAX", "MIN", "OCTET_LENGTH", "POSITION",
+					"REGR_AVGX", "REGR_AVGY", "REGR_COUNT", "REGR_INTERCEPT", "REGR_R2", "REGR_SLOPE", "REGR_SXX",
+					"REGR_SXY", "REGR_SYY", "STDDEV_POP", "STDDEV_SAMP", "SUM", "TRIM", "UPPER", "VAR_POP",
+					"VAR_SAMP" );
+		}
+		else {
+			builder.applyReservedWords(
+					"AVG", "CHARACTER_LENGTH", "CHAR_LENGTH", "COUNT", "EXTRACT", "LOWER", "MAX", "MIN", "OCTET_LENGTH",
+					"POSITION", "SUM", "TRIM", "UPPER" );
+		}
 
 		return super.buildIdentifierHelper( builder, dbMetaData );
 	}
@@ -380,6 +498,7 @@ public class FirebirdDialect extends Dialect {
 		return false;
 
 	}
+
 	@Override
 	public boolean supportsCommentOn() {
 		return getVersion() >= 200;
@@ -414,7 +533,7 @@ public class FirebirdDialect extends Dialect {
 
 	@Override
 	public int getInExpressionCountLimit() {
-		// see http://www.firebirdsql.org/file/documentation/reference_manuals/fblangref25-en/html/fblangref25-commons-predicates.html#fblangref25-commons-in
+		// see https://www.firebirdsql.org/file/documentation/html/en/refdocs/fblangref25/firebird-25-language-reference.html#fblangref25-commons-in
 		return 1500;
 	}
 
@@ -465,12 +584,9 @@ public class FirebirdDialect extends Dialect {
 	public String getQuerySequencesString() {
 		return getVersion() < 300
 				? "select rdb$generator_name from rdb$generators"
-				// Note: currently has an 'off by increment' bug, see
+				// Note: Firebird 3 has an 'off by increment' bug (fixed in Firebird 4), see
 				// http://tracker.firebirdsql.org/browse/CORE-6084
-				// May need revision depending on the final solution
-				// The second column might need to be changed to
-				//   rdb$initial_value + rdb$generator_increment
-				: "select rdb$generator_name, rdb$initial_value, rdb$generator_increment from rdb$generators";
+				: "select rdb$generator_name, rdb$initial_value, rdb$generator_increment from rdb$generators where coalesce(rdb$system_flag, 0) = 0";
 	}
 
 	@Override
@@ -493,7 +609,6 @@ public class FirebirdDialect extends Dialect {
 				? SkipFirstLimitHandler.INSTANCE
 				: OffsetFetchLimitHandler.INSTANCE;
 	}
-
 
 	@Override
 	public String getSelectGUIDString() {
@@ -541,6 +656,16 @@ public class FirebirdDialect extends Dialect {
 	}
 
 	@Override
+	public NullOrdering getNullOrdering() {
+		return getVersion() >= 200 ? NullOrdering.SMALLEST : NullOrdering.LAST;
+	}
+
+	@Override
+	public boolean supportsNullPrecedence() {
+		return getVersion() >= 150;
+	}
+
+	@Override
 	public String translateExtractField(TemporalUnit unit) {
 		switch ( unit ) {
 			case DAY_OF_MONTH: return "day";
@@ -567,7 +692,7 @@ public class FirebirdDialect extends Dialect {
 
 	@Override
 	public String translateDatetimeFormat(String format) {
-		throw new NotYetImplementedFor6Exception("format() function not supported on Firebird");
+		throw new NotYetImplementedFor6Exception( "format() function not supported on Firebird" );
 	}
 
 	@Override
@@ -608,8 +733,8 @@ public class FirebirdDialect extends Dialect {
 			switch ( errorCode ) {
 				case 335544336:
 					// isc_deadlock (deadlock, note: not necessarily a deadlock, can also be an update conflict)
-					if (sqlExceptionMessage != null
-							&& sqlExceptionMessage.contains( "update conflicts with concurrent update" )) {
+					if ( sqlExceptionMessage != null
+							&& sqlExceptionMessage.contains( "update conflicts with concurrent update" ) ) {
 						return new LockTimeoutException( message, sqlException, sql );
 					}
 					return new LockAcquisitionException( message, sqlException, sql );
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/annotations/inheritance/joined/JoinedSubclassAndSecondaryTable.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/annotations/inheritance/joined/JoinedSubclassAndSecondaryTable.java
@@ -155,16 +155,17 @@ public class JoinedSubclassAndSecondaryTable {
 	private void assertNumberOfRows(Session s, String tableName, int expectedNumberOfRows, String message) {
 		s.doWork(
 				work -> {
-					PreparedStatement preparedStatement = work.prepareStatement( "select count(*) as row_count from " + tableName );
-					preparedStatement.execute();
-					ResultSet resultSet = preparedStatement.getResultSet();
-					resultSet.next();
-					long rowCount = resultSet.getLong( "row_count" );
-					assertEquals(
-							expectedNumberOfRows,
-							rowCount,
-							message
-					);
+					try (PreparedStatement preparedStatement = work.prepareStatement( "select count(*) as count_of_rows from " + tableName )) {
+						preparedStatement.execute();
+						ResultSet resultSet = preparedStatement.getResultSet();
+						resultSet.next();
+						long rowCount = resultSet.getLong( "count_of_rows" );
+						assertEquals(
+								expectedNumberOfRows,
+								rowCount,
+								message
+						);
+					}
 				}
 		);
 	}
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
@@ -7,6 +7,7 @@
 package org.hibernate.orm.test.query.hql;
 
 import org.hibernate.dialect.DerbyDialect;
+import org.hibernate.dialect.FirebirdDialect;
 
 import org.hibernate.testing.orm.domain.StandardDomainModel;
 import org.hibernate.testing.orm.domain.gambit.EntityOfBasics;
@@ -27,6 +28,7 @@ import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
 
+import static org.hamcrest.CoreMatchers.anyOf;
 import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.nullValue;
@@ -368,7 +370,7 @@ public class FunctionTests {
 							.list();
 					session.createQuery("select cast(e.id as BigInteger(10)), cast(e.theDouble as BigDecimal(10,5)) from EntityOfBasics e")
 							.list();
-					session.createQuery("select cast(e.theString as String(15)), cast(e.theDouble as String(8)) from EntityOfBasics e")
+					session.createQuery("select cast(e.theString as String(15)), cast(e.theDouble as String(17)) from EntityOfBasics e")
 							.list();
 
 					session.createQuery("select cast('1002342345234523.452435245245243' as BigDecimal) from EntityOfBasics")
@@ -384,8 +386,6 @@ public class FunctionTests {
 					session.createQuery("select cast('1234234' as Integer) from EntityOfBasics")
 							.list();
 					session.createQuery("select cast(1 as Boolean), cast(0 as Boolean) from EntityOfBasics")
-							.list();
-					session.createQuery("select cast('ABCDEF' as Character) from EntityOfBasics")
 							.list();
 
 					session.createQuery("select cast('12:13:14' as Time) from EntityOfBasics")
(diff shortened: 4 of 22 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
```diff
@@ -198,18 +237,143 @@ public void initializeFunctionRegistry(QueryEngine queryEngine) {
...
+							hash,
+							"crypt_hash(?1 using " + hash + ")",
+							StandardBasicTypes.BINARY
+					) );
+			functionRegistry.registerAlternateKey( "sha", "sha1" );
+			functionRegistry.registerPattern(
+					"crc32",
+					"hash(?1 using crc32)",
+					StandardBasicTypes.INTEGER
+			);
```
gavinking, line 353: I'm don't think I approve of just adding all these. In the other `Dialects I've deliberately got rid of long lists of native functions, and only registered the ones that seemed directly useful in HQL.

(For example, I can't see why I would ever call `rdb$set_context()` from HQL.)

Remember, you can always call any arbitrary function via the JPAQL `function()` syntax.

But it's more than that: it's essentially impossible for us to really maintain these long lists of functions across all the dialects, and we have historically failed to do so in any sort of consistent way, so I don't think we should try.

gavinking (changes requested):

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
```diff
@@ -231,10 +395,6 @@ public SqlAstTranslatorFactory getSqlAstTranslatorFactory() {
 	public String castPattern(CastType from, CastType to) {
 		String result;
 		switch ( to ) {
-			case FLOAT:
-				return "cast(double(?1) as real)";
-			case DOUBLE:
-				return "double(?1)";
```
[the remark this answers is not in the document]
gavinking: Oh, right, I see, this was a recent change. My original rewrite of `FirebirdDialect` for H6 didn't have those lines, it seems.

hibernate-core/src/test/java/org/hibernate/orm/test/set/SetOperationTest.java
```diff
@@ -91,6 +92,7 @@ public void testUnionAllLimit(SessionFactoryScope scope) {
 
     @Test
     @RequiresDialectFeature(feature = DialectFeatureChecks.SupportsUnion.class)
+    @SkipForDialect(dialectClass = FirebirdDialect.class, reason = "Firebird does not support parenthesized select in UNION")
```
beikov, line 95: We can fix that. I think some other databases are a bit picky there as well and require a select wrapper.

hibernate-core/src/test/java/org/hibernate/orm/test/set/SetOperationJpaCriteriaTest.java
```diff
@@ -108,6 +108,7 @@ public void testUnionAllLimit(SessionFactoryScope scope) {
 
     @Test
     @RequiresDialectFeature(feature = DialectFeatureChecks.SupportsUnion.class)
+    @SkipForDialect(dialectClass = FirebirdDialect.class, reason = "Firebird does not support parenthesized select in UNION")
```
beikov, line 111: Same as https://github.com/hibernate/hibernate-orm/pull/3797/files#r589465018

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/StandardFunctionTests.java
```diff
@@ -888,6 +896,7 @@ public void testAggregateFunctions(SessionFactoryScope scope) {
 
 	@Test
 	@SkipForDialect(dialectClass = DerbyDialect.class, reason = "Derby doesn't support formatting temporal types to strings")
+	@SkipForDialect(dialectClass = FirebirdDialect.class, reason = "Firebird doesn't support formatting temporal types to strings")
```
beikov, line 899: Is there no format function? I mean we could emulate this by using the extraction functions since we require the pattern to be a literal anyway, but I'd rather use a builtin format function if there is one.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
```diff
@@ -441,6 +443,20 @@ public void testCastFunction(SessionFactoryScope scope) {
 		);
 	}
 
+	@Test
+	@SkipForDialect(dialectClass = FirebirdDialect.class, reason = "Firebird cast to string types doesn't truncate")
```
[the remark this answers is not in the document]
mrotteveel: Do you mean something like instead of cast, doing a `substring(?1 from 1 for <length>)` where I somehow need to obtain the length for the cast?

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
```diff
@@ -464,7 +480,7 @@ public void testStrFunction(SessionFactoryScope scope) {
 							.list();
 					assertThat( session.createQuery("select str(69)").getSingleResult(), is("69") );
 					assertThat( session.createQuery("select str(date 1911-10-09)").getSingleResult(), is("1911-10-09") );
-					assertThat( session.createQuery("select str(time 12:13:14)").getSingleResult(), is("12:13:14") );
+					assertThat( session.createQuery("select str(time 12:13:14)").getSingleResult(), anyOf( is( "12:13:14"), is( "12:13:14.0000") ) );
```
beikov, line 483: I'd rather fix the string formatting of firebird to have consistent semantics of the functions.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
```diff
@@ -264,6 +266,7 @@ public void testOverlayFunction(SessionFactoryScope scope) {
 	}
 
 	@Test
+	@SkipForDialect(dialectClass = FirebirdDialect.class, reason = "Firebird requires explicit cast for functions with context sensitive parameters")
```
beikov, line 269: Derby seems to require this as well which is why a rendering mode was introduced. Which functions require that? I can adapt them to make this work.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
mrotteveel: I had somehow assumed that Hibernate would apply the automatic quoting on function names as well if they were not explicitly defined, and then built-in functions will not work (Firebird only supports quoting user-defined function names), but I did some testing and that turns out is not the case. I committed a change where I removed the Firebird specific functions. 

I left the definition of the various SQL standard statistical functions in place. I had considered defining those in `CommonFunctions` like the other statistical functions (e.g. `CommonFunctions.stddevPopSamp(...)`, but I felt that I would then also need to add them to the dialects that also support those functions. 

I also left the mappings for crc32, sha/sha1, md5, sha256 and sha512 in place, as some other dialects define those functions as well. Let me know if you prefer them removed as well.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
```diff
@@ -316,29 +295,6 @@ public void initializeFunctionRegistry(QueryEngine queryEngine) {
...
-					.setInvariantType( StandardBasicTypes.STRING );
-			functionRegistry.namedDescriptorBuilder( "hex_decode" )
-					.setExactArgumentCount( 1 )
-					.setInvariantType( StandardBasicTypes.BINARY );
-			functionRegistry.namedDescriptorBuilder( "base64_encode" )
-					.setExactArgumentCount( 1 )
-					.setInvariantType( StandardBasicTypes.STRING );
-			functionRegistry.namedDescriptorBuilder( "base64_decode" )
-					.setExactArgumentCount( 1 )
-					.setInvariantType( StandardBasicTypes.BINARY );
```
beikov: I think these conversion functions are ok to have. I guess @gavinking only meant that the internal functions `rdb$..` are an overkill

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
```diff
@@ -266,27 +266,6 @@ public void initializeFunctionRegistry(QueryEngine queryEngine) {
...
-		functionRegistry.namedDescriptorBuilder( "gen_uuid" )
-				.setExactArgumentCount( 0 )
-				.setInvariantType( StandardBasicTypes.UUID_BINARY );
-		functionRegistry.namedDescriptorBuilder( "uuid_to_char" )
-				.setExactArgumentCount( 1 )
-				.setInvariantType( StandardBasicTypes.UUID_CHAR );
-		functionRegistry.namedDescriptorBuilder( "gen_id" )
-				.setExactArgumentCount( 2 )
-				.setInvariantType( StandardBasicTypes.BIG_INTEGER )
-				.setArgumentListSignature( "(sequence, step)" );
```
beikov: Same for these functions. What do you think @gavinking?

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
gavinking: Well md5 and sha functions are defined in `CommonFunctionFactory` so you should reuse those definitions.

I’m fine with adding any SQL-standard-defined statistical functions, but they should also be defined in  `CommonFunctionFactory`.

gavinking: > Well md5 and sha functions are defined in CommonFunctionFactory so you should reuse those definitions.

Oh I see, it’s because you’re mapping them to `crypt_hash`. That’s fine, I suppose.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
```diff
@@ -198,18 +237,97 @@ public void initializeFunctionRegistry(QueryEngine queryEngine) {
...
+		functionRegistry.registerAlternateKey( "ascii", "ascii_val" );
+		functionRegistry.namedDescriptorBuilder( "ascii_char" )
+				.setExactArgumentCount( 1 )
+				.setInvariantType( StandardBasicTypes.CHARACTER )
+				.register();
+		functionRegistry.registerAlternateKey( "chr", "ascii_char" );
+		functionRegistry.registerAlternateKey( "char", "ascii_char" );
+		functionRegistry.registerPattern(
+				"radians",
+				"((?1)*PI()/180e0)",
```
gavinking, line 261: is there a reason to write π in shouty-case here?

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
gavinking: I am against including them here unless someone is prepared to go and do the work of including them for *all* the `Dialect`s.

I'm sick of half-assed inconsistencies with this stuff.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
gavinking: These functions aren't generally useful from *queries*.

gavinking: But again, if someone wants to go through *all* the `Dialect`s which support a UUID-like type, and define a portable `uuid_to_char()` function, that would be OK.

On the other hand, it would be much better to just make sure that you can do it using the existing `cast()` function, i..e

    cast(uuid to String)
    cast(char to UUID)

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
beikov: I don't feel strong about this, but the issue with not registering these functions is that they can't be used in the SELECT clause through the JPA `function` wrapper because the expression type is unknown. So to use this, a user would have to use a cast e.g. `cast(function('base64_encode', alias.myBlobField) as String(99999))` vs `base64_encode(alias.myBlobField)`. I just wanted to mention it because this is why I think that supporting these functions is fine. Anyway, I agree though, that it is probably a good idea to either go through all Dialects, or not include it at all.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/StandardFunctionTests.java
mrotteveel: Unfortunately, no: there is http://tracker.firebirdsql.org/browse/CORE-1314 (which is a bit too generic to my taste), and I have now created http://tracker.firebirdsql.org/browse/CORE-6507

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
mrotteveel: Most - if not all - functions of Firebird will need an explicit cast of a parameter, because the engine cannot infer the datatype as most function parameters are contextually typed (I'm not even sure there are cases where the type of a parameter can be inferred, except in a cast).

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
mrotteveel: No reason, will fix it.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
mrotteveel: I could also change it to do a similar check as on line 426:
```java
assertThat( (String) session.createQuery("select cast(datetime 1911-10-09 12:13:14 as String)").getSingleResult(), startsWith("1911-10-09 12:13:14") );
```

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
mrotteveel: I can add them back in if there is a consensus to do so.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
beikov: I didn't check the details, but if the problem is, that we are using a cast function or something that produces the precision, and there is no alternative for Firebird, then we should use `substring` to achieve a consistent output.

gavinking: Rrrrm I dunno about that. It really might be pretty hard to write a robust date parser in SQL.

I'm not sure I think this is a big deal.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
gavinking: Right, the requirement is clear: if someone needs this stuff, they can go through and add it to all the supported `Dialects. Otherwise it's clear they don't really need it _that much_.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
gavinking: Thanks :-) Sorry for being anal, it just saves me going through later and lowercasing things.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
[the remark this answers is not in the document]
gavinking: Hrm, OK my bad, `castPattern()` only has access to the whole type it's casting to, not the length.

So to fix this you would need to add a whole custom `CastFunction`. Might not be worth it.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
mrotteveel: Firebird doesn't actually have a specific UUID type (so mapping to `StandardBasicTypes.UUID_BINARY` might have been the wrong thing). Basically the `char_to_uuid`/`uuid_to_char` functions transform between human 'readable' UUID string (usually `char(36)`) and `binary(16)` (with the assumption that is a binary UUID).

In any case, lets leave them out, we can always revisit this if there turns out to be a strong need for this.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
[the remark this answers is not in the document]
mrotteveel: Ok, lets leave this alone for now.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
mrotteveel: I moved the definition statistical functions into `CommonFunctionFactory` in upcoming commit.

mrotteveel: And I also added (some of) the statistical functions to Oracle, DB2 and PostgreSQL, in as far as I could quickly find references to them in their online documentation.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
mrotteveel: Ok, lets leave them out for now

hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
```diff
@@ -186,6 +186,8 @@ public void initializeFunctionRegistry(QueryEngine queryEngine) {
 		CommonFunctionFactory.median( queryEngine );
 		CommonFunctionFactory.stddev( queryEngine );
 		CommonFunctionFactory.stddevPopSamp( queryEngine );
+		CommonFunctionFactory.regrLinearRegressionAggregates( queryEngine );
+		queryEngine.getSqmFunctionRegistry().registerAlternateKey( "regr_icpt", "regr_intercept" );
```
gavinking, line 190: I guess I would leave off `regr_icpt`, since it's not the standard syntax. As long as people can use `regr_intercept()` they're good, it seems to me.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
```diff
@@ -114,23 +117,39 @@ public FirebirdDialect(int version) {
...
+			registerColumnType( Types.VARBINARY, 32_765, "varbinary($l)" );
+		}
 		registerColumnType( Types.VARBINARY, "blob sub_type binary" );
 
 		registerColumnType( Types.BLOB, "blob sub_type binary" );
 		registerColumnType( Types.CLOB, "blob sub_type text" );
 		registerColumnType( Types.NCLOB, "blob sub_type text" ); // Firebird doesn't have NCLOB, but Jaybird emulates NCLOB support
+		registerColumnType( Types.LONGVARBINARY, "blob sub_type binary" );
+		registerColumnType( Types.LONGVARCHAR, "blob sub_type text" );
+		registerColumnType( Types.LONGNVARCHAR, "blob sub_type text" );
```
gavinking, line 152: I would prefer not to add these. We don't use them anymore, and except for a couple of exceptions the other `Dialect`s generally don't register any mapping for them.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
gavinking: OK, thanks, that's great.

hibernate-core/src/test/java/org/hibernate/orm/test/annotations/inheritance/joined/JoinedSubclassAndSecondaryTable.java
```diff
@@ -155,16 +155,17 @@ public void testSecondaryTableAndJoinedInverse(SessionFactoryScope scope) {
...
-					preparedStatement.execute();
-					ResultSet resultSet = preparedStatement.getResultSet();
-					resultSet.next();
-					long rowCount = resultSet.getLong( "row_count" );
-					assertEquals(
-							expectedNumberOfRows,
-							rowCount,
-							message
-					);
+					try (PreparedStatement preparedStatement = work.prepareStatement( "select count(*) as \"row_count\" from " + tableName )) {
```
gavinking, line 158: Does this work on MySQL?

mrotteveel: Good point; I'll replace it with something that is not a keyword in either.

gavinking: thanks

hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
mrotteveel: Ok, removed in next commit

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
mrotteveel: I wanted to be complete, as Jaybird identifies `blob sub_type binary` and `blob sub_type text` as Types.LONGVARBINARY and Types.LONGVARCHAR in its metadata methods. I haven't found anything where this mattered, but I was wondering if there was a case where some form of metadata inspection would need this info.

If you're not aware of such a case either, I'll remove it again.

gavinking: These column type registrations are only used for DDL generation. We don't have any types that use these.

It is in principle possible for a `UserType` to return `Types.LONGVARBINARY` but it wouldn't make much sense here: it would be more sensible to return `Types.BLOB` if they want a `blob`.

mrotteveel: Ok, I have removed them

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
mrotteveel: The problem is that Firebird's `TIME` type has a fixed precision of 4 (100 micro-seconds) and conversion to string will always return the full precision. It looks like a lot of other database systems either have a fixed precision of 0, or at least a default precision of 0, or maybe do not include trailing zeroes. 

I think that removing the fractional seconds with substring or similar would not be a good solution, as it could lead to loss of information.

gavinking: I’m finding it hard to care about this one.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/StandardFunctionTests.java
beikov: It's up to you if you want to support this by emulation through extraction functions. I'm fine with ignoring this test.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
beikov: AFAIU `java.sql.Time` which is what is backing the TIME type is defined to have second precision, so IMO the cast pattern from `CastType.TIME` to `CastType.STRING` should be `substring(cast(?1 as ?2) from 1 for 8)` and the test changes should be reverted.

hibernate-core/src/test/java/org/hibernate/orm/test/set/SetOperationTest.java
beikov: In the `FirebirdSqlAstTranslator`, please add the following override:

```
	@Override
	protected boolean supportsSimpleQueryGrouping() {
		// Firebird is quite strict i.e. it requires `select .. union all select * from (select ...)`
		// rather than `select .. union all (select ...)`
		return false;
	}
```

hibernate-core/src/test/java/org/hibernate/orm/test/set/SetOperationJpaCriteriaTest.java
beikov: See https://github.com/hibernate/hibernate-orm/pull/3797#discussion_r590040258

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
beikov: Ok so an implementation for this could look like the following in `FirebirdSqlAstTranslator`:

```
        private boolean inFunction;

	@Override
	public void visitSelfRenderingPredicate(SelfRenderingPredicate selfRenderingPredicate) {
		boolean inFunction = this.inFunction;
                this.inFunction = true;
                try {
                    super.visitSelfRenderingPredicate( selfRenderingPredicate );
                }
                finally {
                    this.inFunction = inFunction;
                }
	}

	@Override
	public void visitSelfRenderingExpression(SelfRenderingExpression expression) {
		boolean inFunction = this.inFunction;
                this.inFunction = true;
                try {
                    super.visitSelfRenderingExpression( expression);
                }
                finally {
                    this.inFunction = inFunction;
                }
	}

	@Override
	public void visitParameter(JdbcParameter jdbcParameter) {
		if ( inFunction ) {
			render( jdbcParameter, SqlAstNodeRenderingMode.NO_PLAIN_PARAMETER );
		}
		else {
			super.visitParameter( jdbcParameter );
		}
	}

	@Override
	public void visitJdbcLiteral(JdbcLiteral jdbcLiteral) {
		visitLiteral( jdbcLiteral );
	}

	@Override
	public void visitQueryLiteral(QueryLiteral queryLiteral) {
		visitLiteral( queryLiteral );
	}

	private void visitLiteral(Literal literal) {
		if ( literal.getLiteralValue() == null ) {
			appendSql( SqlAppender.NULL_KEYWORD );
		}
		else {
			renderLiteral( literal, inFunction );
		}
	}
```

One downside to this is, that this could lead to casted parameters also within arithmetic expressions or subqueries that appear within functions. You could fix that though if you override these methods as well and reset the `inFunction` flag for the expression.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
mrotteveel: If someone wants a `java.sql.Time` (note though that `java.time.LocalTime` is now the recommended type), then you wouldn't cast here. The intent is to obtain a string representation of the time value, and in the case of Firebird, that string representation includes the fractional seconds.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
mrotteveel: Thanks, I'll look at that

hibernate-core/src/test/java/org/hibernate/orm/test/set/SetOperationTest.java
mrotteveel: Thanks, I'll do that

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
beikov: >The problem is that Firebird's TIME type has a fixed precision of 4 (100 micro-seconds) and conversion to string will always return the full precision. It looks like a lot of other database systems either have a fixed precision of 0, or at least a default precision of 0, or maybe do not include trailing zeroes.

I think you are right and the current behavior is ok. The problem really seems to be that other databases allow specifying the precision or use a default precision of 0. If other databases would use a different precision, we'd probably get the same output.

As far as I can say, the SQL standard defines that the time type should allow specifying a precision and default to 0. So maybe we need to adapt the Dialects in that regard to allow specifying a precision. What do you think @gavinking? Here is a section from the SQL Standard from 92 section 6.1:

>         <datetime type> ::=
>                DATE
>              | TIME [ <left paren> <time precision> <right paren> ]
>              [ WITH TIME ZONE ]
>              | TIMESTAMP [ <left paren> <timestamp precision> <right paren> ]
>              [ WITH TIME ZONE ]

>         25)If <time precision> is not specified, then 0 is implicit. If
>            <timestamp precision> is not specified, then 6 is implicit.

@mrotteveel It seems to me that using a precision of 3 as default for time is a violation of the specification.

I checked most databases and observed the following:

* Most DBs support `TIME(p)`: MySQL 5.6+, MariaDB, H2, PostgreSQL, HSQL, Ingres, Teradata, SQL Server, Sybase Anywhere, Sybase IQ, Sybase ASE, MimerSQL, CockroachDB
* SAP MaxDB supports the precision but not sure if it supports `TIME(p)` or needs `TIMESTAMP(p)`
* Oracle requires `TIMESTAMP(p)` => currently we map it to `DATE` which has no precision
* DB2 requires `TIMESTAMP(p)` => currently we map it to `TIME` which has no precision
* Informix requires `DATETIME HOUR TO FRACTION(p)` => currently we map it to `datetime hour to second`
* TimesTen requires `TIMESTAMP(p)` => currently we map it to `TT_TIME` which has no precision
* HANA requires `TIMESTAMP(p)` => currently we map it to `TIME` which has no precision
* CUBRID requires `TIMESTAMP(p)` => currently we map it to `TIME` which has no precision
* Derby requires `TIMESTAMP` which has a precision of 6 => currently we map it to `TIME` which has no precision
* Cache requires `TIMESTAMP` which has a precision of 6 => currently we map it to `TIME` which has no precision
* Spanner doesn't allow specifying precision and only supports `TIMESTAMP` with default precision 9

What do you think about registering `time($p)` by default and adding `registerColumnType( Types.TIME, 0, "time" )` to retain the current behavior, but still support a precision if needed @gavinking?

gavinking: Well I thought quite a lot about this, and I decided it's *extremely* difficult to imagine a case where anyone could possibly need a *time* with millisecond precision. (Or even *second* precision, TBH.) It really just doesn't make sense as soon as you think about what a time represents.

(*Timestamps*, of course need it.)

gavinking: FTR, the SQL standard is crap on these points. For example, the notion of a time with a timezone is just broken.

mrotteveel: Unfortunately, Firebird has a fixed precision of 4 for `TIME` and `TIMESTAMP`, and it doesn't support `TIME(p)` (nor `TIMESTAMP(p)`), so this workaround isn't possible.

As an aside, `CURRENT_TIME` and `CURRENT_TIMESTAMP` have a maximum precision of 3 (`CURRENT_TIME` defaults to 0), contrary to the actual datatype, see http://tracker.firebirdsql.org/browse/CORE-4458), which is why it's sometimes thought of having a precision of 3.

gavinking: OK, but IIRC, I concluded that it was perfectly fine for *Hibernate* to treat it has having precision 3. I'm pretty sure I tested this and it was working fine.

gavinking: (Or precision zero in the case of `time` as opposed to `timestamp`.)

mrotteveel: Yes, that will work fine, except for this specific topic: the effects of casting to string, which will include 4 digits for fractional seconds. I think this is a minor concern, that we could ignore for now (that is, accept the status quo), and maybe handle it later if and when a format function (http://tracker.firebirdsql.org/browse/CORE-6507) is available.

beikov: So are you saying that casting from `TIME` to `STRING` should always trim the fractional parts? I never really used `TIME(p)` except for stuff like e.g. delivery time which only has minute precision, so I don't know about use cases either. I can imagine though that someone might want to track elapsed time as `TIME(p)`. The alternative would be to use `INT` with some default unit and do conversions, or use a `TIMESTAMP(p)` and ignore the date part. IMO supporting `TIME(p)` is in the same ball park as supporting `FLOAT(p)` (for which I also can't imagine a use case) which you added @gavinking, so why treat this differently?

gavinking: > So are you saying that casting from `TIME` to `STRING` should always trim the fractional parts?

As I said above, I just don't really care about this.

> I can imagine though that someone might want to track elapsed time as `TIME(p)`.

They should definitely not use `time` for that! Just think about it.

> IMO supporting `TIME(p)` is in the same ball park as supporting `FLOAT(p)`

I don't think that's similar at all. We use the `p` in `float(p)` to indicate whether we're dealing with single precision or double precision floating point types.

There's no reasonable usecase that I can imagine for millisecond-precise times. It just doesn't make sense. You're saying that something happens at time 1:35:12.1234 each day!? No, you're wrong. It doesn't.

gavinking: > You're saying that something happens at time 1:35:12.1234 each day!?

Or even: it's *going to happen at* 1:35:12.1234 next Wednesday?

Still nope. It's won't.

mrotteveel: @beikov I'm not saying that, but if you're insisting that the produced string value should only have second precision, then trimming the fractional parts is the only possibility for Firebird.

This is different from `FLOAT(p)` though. The standard specifies that the binary precision _p_ should lead to a database-specific floating point datatype, though usually 1 <= p <= 24 will result in a 32-bit single precision, while 25 <= p <= 53 in a 64-bit double precision, and some database systems even support higher precision. For Firebird, `FLOAT` (without precision) is a 32-bit single precision (with alias `REAL`). In Firebird 3.0 and earlier, Firebird has an undocumented `FLOAT(p)` with **decimal** precision, while Firebird 4 (based on a discussion I had with Gavin a while back), introduced `FLOAT(p)` with binary precision as defined in the standard.

What a lot of people consider `FLOAT` is actually `FLOAT(p)` with an implementation-specific default precision, which in some database systems could very well be 53 (aka a 64-bit double precision). Where `REAL` and `DOUBLE PRECISION` have a fixed (but still database-specific) precision (though usually 32-bit single and 64-bit double), `FLOAT(p)` is intended to allow for more dynamic ranges. For example, Oracle allows values of _p_ up to 126 (for a 176-bit floating point type!).

gavinking: Right, I mean, `float` and `real` don't really seem to mean anything particularly well-defined. That's why we avoid them in H6.

beikov: >There's no reasonable usecase that I can imagine for millisecond-precise times. It just doesn't make sense. You're saying that something happens at time 1:35:12.1234 each day!? No, you're wrong. It doesn't.

The use case I was imagining was tracking elapsed time e.g. a racing car took X <time> to complete a lap.

gavinking: > trimming the fractional parts is the only possibility for Firebird

The question is: is that actually easy to do?

gavinking: > The use case I was imagining was tracking elapsed time e.g. a racing car took X [time] to complete a lap.

But that's not a time. It's a duration.

And how on earth do you propose  to handle the case where the car starts its lap at 23:59:50?

gavinking: I mean, sure, I have no doubt that programmers abuse the `time` type for all kinds of things because they like having bugs in their programs, but I don't think we should *encourage* it.

mrotteveel: > The use case I was imagining was tracking elapsed time e.g. a racing car took X [time] to complete a lap.

`TIME` is not intended for tracking elapsed time. Consider what happens if the elapsed time is longer than 24 hours.

> > trimming the fractional parts is the only possibility for Firebird
> 
> The question is: is that actually easy to do?

It would require string manipulation, and as such it is 1) complicated and 2) may have undesired side-effects (e.g. what if Hibernate thinks it is the wrong type, e.g. it is actually a TIMESTAMP), or what if the value is a TIME WITH TIME ZONE. 

I'm not sure the added complications are worth it. To be honest, I think wanting to cast a TIME to a string in a query is a pretty niche requirement.

gavinking: > I don't think we need to debate whether sub-second fractions of `TIME` make sense or not, this is simply the reality we're dealing with.

Well that's not clear to me.

Previously the dialect was set up to say that Firebird has 3-digit precision, because that's what is *reliably* supports everywhere (e.g. in `current_timestamp`).

I'm still not convinced that's wrong. *Why* do we need to say that it has 4-digit precision when most likely the 4th digit holds garbage? Was there some particular thing that was broken @mrotteveel? Because if there's not, I would be inclined to leave it like it was.

(Again, I don't case about what casting to string does, that's a minor side issue IMO.)

gavinking: > I find it odd that we don't allow defining the precision for time but allow mapping e.g. a Java float with a precision although nothing other than `float(24)` makes sense.

Well it's because (a) more databases support `float(p)` than support `time(p)`, and (b) we're using that syntax to distinguish single precision from double precision and it's no more work to support it generically.

On the other hand we have _no good reason_ to support `time(p)`. We're always happy with the default precision of 0.

But sure, if somebody wanted to go through all the `Dialects and add support for `time(p)` where supported then I guess I wouldn't object. I just see it as a pretty pointless exercise.

mrotteveel: > Well that's not clear to me.
> 
> Previously the dialect was set up to say that Firebird has 3-digit precision, because that's what it _reliably_ supports everywhere (e.g. in `current_timestamp`).

Specifically what change are you referring to? Are you referring to the change I made to `getFractionalSecondPrecisionInNanos()` from 1_000_000 to 100_000, or `getDefaultTimestampPrecision()` to return 4 (which previous to my test would apply the default value of 6)?

Reverting those won't change the reason why I made the change to the test we're commenting on. I modified that test because otherwise this specific test will fail on Firebird, no matter what precision Hibernate is attempting to apply, as cast from `TIME` to a string type will always include fractional seconds in how Firebird performs that conversion.
 
> I'm still not convinced that's wrong. _Why_ do we need to say that it has 4-digit precision when most likely the 4th digit holds garbage? Was there some particular thing that was broken @mrotteveel? Because if there's not, I would be inclined to leave it like it was.

The fourth digit will either be zero (e.g. when `CURRENT_TIME` is used), or the value that was explicitly stored there. It won't be garbage.

I have the feeling we're discussing two different things here.

hibernate-core/src/test/java/org/hibernate/orm/test/set/SetOperationTest.java
mrotteveel: This also solved two failing tests in `InsertUpdateTests`

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
gavinking: > Specifically what change are you referring to? 

Yeah, sorry, apologies, I was doing five things at once as usual and said the wrong thing. What I'm arguing is we should treat it as having consistently millisecond precision. (I might be wrong.)

> Reverting those won't change the reason why I made the change to the test we're commenting on. 

Yes, sure, I know that.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
```diff
@@ -166,14 +182,28 @@ public int getPreferredSqlTypeCodeForBoolean() {
...
-		return 21; // -> 7 decimal digits
+		return getVersion() < 400
+				? 21 // -> 7 decimal digits (actually 24, but needed for Dialect#binaryToDecimalPrecision(int,size))
+				: 24;
+	}
+
+	@Override
+	public int getDefaultTimestampPrecision() {
+		// Though in some cases values will only have millisecond precision (3)
+		return 4;
```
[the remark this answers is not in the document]
gavinking: It seems to me that we should say that it has millisecond precision. That would be `3` and `1_000_000`.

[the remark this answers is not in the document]
gavinking: But y'know, if you want to say it has 100 microsecond precision I guess that's OK. Your call.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
mrotteveel: The code as suggested doesn't work, because `SqlAstNodeRenderingMode.NO_PLAIN_PARAMETER` does things like changing the parameter to `(?||'')`, which breaks the type inference once again. And that was after fixing a stack overflow.

I found an alternative:

```java
	@Override
	public void visitParameter(JdbcParameter jdbcParameter) {
		if ( inFunction ) {
			inFunction = false;
			try {
				final JdbcMapping jdbcMapping = jdbcParameter.getExpressionType().getJdbcMappings().get( 0 );
				final List<SqlAstNode> arguments = new ArrayList<>( 2 );
				arguments.add( jdbcParameter );
				arguments.add( new CastTarget( (BasicValuedMapping) jdbcMapping ) );
				castFunction().render( this, arguments, this );
			} finally {
				inFunction = true;
			}
		}
		else {
			super.visitParameter( jdbcParameter );
		}
	}
```

However, this has the problem that it can generate things like `cast(cast(? as varchar(255)) as varchar(255))` if there is already an explicit cast present, and I haven't yet found a way to detect that the parameter is rendered immediately contained in a cast expression. Any ideas how to do that?

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
mrotteveel: I changed the dialect to report as if it has millisecond precision.

hibernate-core/src/main/java/org/hibernate/dialect/FirebirdDialect.java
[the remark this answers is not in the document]
mrotteveel: I changed it to report millisecond precision

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
beikov: >The code as suggested doesn't work, because SqlAstNodeRenderingMode.NO_PLAIN_PARAMETER does things like changing the parameter to (?||''), which breaks the type inference once again.

If that doesn't work, it's fine to always do explicit casts. Strange though, that such operations which should IMO have a clear result type, are not helping the type inference. At least for Derby, this did the trick. Did you try if swapping the operands helps i.e. `(''||?)`?

>And that was after fixing a stack overflow.

Something I should know about or was that just an issue with the code I posted?

>Any ideas how to do that?

I don't think you can detect that as there is currently no way to determine what function a `SelfRenderingExpression` represents. This might change though as I need something similar to support certain window function emulations.

mrotteveel: > > The code as suggested doesn't work, because SqlAstNodeRenderingMode.NO_PLAIN_PARAMETER does things like changing the parameter to (?||''), which breaks the type inference once again.
> 
> If that doesn't work, it's fine to always do explicit casts. Strange though, that such operations which should IMO have a clear result type, are not helping the type inference. At least for Derby, this did the trick. Did you try if swapping the operands helps i.e. `(''||?)`?

Unfortunately no, the transformation of `?||''` or `''||?` makes the parameter type once again unknown, because Firebird allows dynamic coercion to string, in other words, the concatenation operator is also contextually typed, which makes the datatype of the parameter unknown again, which was the problem we're trying to address.

> > And that was after fixing a stack overflow.
> 
> Something I should know about or was that just an issue with the code I posted?

IIRC, it was because `render( jdbcParameter, SqlAstNodeRenderingMode.NO_PLAIN_PARAMETER );` will introduce that concatenation, it will call `visitParameter(JdbcParameter jdbcParameter)` again, and so on. I had to switch `inFunction` to `false` and restore it after the call completed.

> > Any ideas how to do that?
> 
> I don't think you can detect that as there is currently no way to determine what function a `SelfRenderingExpression` represents. This might change though as I need something similar to support certain window function emulations.

I was thinking that I can maybe check it by inspecting the current state of the string builder, but that is hardly efficient. I could also create a custom cast function that calls back into the ast parser, but that feels wrong as well, and I'm also not sure that is sufficient for the problem I'm trying to solve here.

I'll do some experiments later this week.

beikov: >Unfortunately no, the transformation of ?||'' or ''||? makes the parameter type once again unknown, because Firebird allows dynamic coercion to string, in other words, the concatenation operator is also contextually typed, which makes the datatype of the parameter unknown again, which was the problem we're trying to address.
>IIRC, it was because render( jdbcParameter, SqlAstNodeRenderingMode.NO_PLAIN_PARAMETER ); will introduce that concatenation, it will call visitParameter(JdbcParameter jdbcParameter) again, and so on. I had to switch inFunction to false and restore it after the call completed.

Makes sense. To avoid the SO you will have to cast explicitly anyway, so that's fine.

>I was thinking that I can maybe check it by inspecting the current state of the string builder, but that is hardly efficient. I could also create a custom cast function that calls back into the ast parser, but that feels wrong as well, and I'm also not sure that is sufficient for the problem I'm trying to solve here.

Do you really think it is worth the hassle to get rid of the second cast? I was thinking about merging the cast during query parsing in `SemanticQueryBuilder` into the parameter, but the parameter type does not necessarily have to be the same as the cast type i.e. it would be ok to have the same named param multiple times within a query, once cast it to e.g. a `Integer` and another time to a `Double` yet the parameter type could be `String` so that wouldn't be a good idea.

I think introducing `CastSqmFunction` and `CastSqlFunction` nodes or something like that would be more appropriate. That way, you could intercept the rendering of a cast and check if the first argument is a parameter, in which case you can render it without another cast.

beikov: With this PR (https://github.com/hibernate/hibernate-orm/pull/3813) you should be able to implement this in the translator like this:

```
	@Override
	public void visitSelfRenderingExpression(SelfRenderingExpression expression) {
		boolean inFunction = this.inFunction;
                this.inFunction = !( expression instanceof FunctionExpression ) || !"cast".equals( ( (FunctionExpression) expression ).getFunctionName() );
                try {
                    super.visitSelfRenderingExpression( expression);
                }
                finally {
                    this.inFunction = inFunction;
                }
	}
```

mrotteveel: I have implemented the suggested change, and it works well. What is the preferred way of updating the PR given I have now a merge commit from wip/6.0. Should I rebase and force push, or push including the merge commit from wip/6.0?

mrotteveel: Also, currently as suggested in https://github.com/hibernate/hibernate-orm/pull/3797#discussion_r590046638 I override `visitJdbcLiteral(JdbcLiteral jdbcLiteral)` and `public void visitQueryLiteral(QueryLiteral queryLiteral)` and add a private method `visitLiteral(Literal literal)`. Would it make sense to not override those methods, and instead make `AbstractSqlAstTranslator.visitLiteral(Literal literal)` protected and override only that?

gavinking: If it were me I would just squash and force push.

mrotteveel: Done

beikov (approved):
Looks good now

mrotteveel:
Is anything more needed before this can be merged?

beikov:
Could you please rebase and resolve the conflicts? After that, I would merge the PR.

gavinking:
Thanks @mrotteveel!

PR #1499, opened 2016-07-29
HHH-10999 Array datatypes module

coladict:
What works:
- Using single-dimensional Object arrays in Entities (no primitive arrays).
- Saving arrays through entities through the `EntityManager` interface.
- Loading arrays into entities through the `EntityManager` interface.
- Tested under PostgreSQL.

What doesn't work:
- Table creation without setting `columnDefinition` on the `javax.persistence.Column` annotation.
- Passing null in the position of an array in manually written queries (bug is in Postgre server, not Hibernate).
- Using arrays of primitives, which are optional for JDBC drivers, even when arrays are supported.

TODO list without changes to core:
- Test more types
- Determine which types should be contributed, and which should be optional.
- Test on HyperSQL (only other database listed in the test suite that supports arrays).
- Annotate the tests to only use compatible dialects or features

TODO list with core changes:
- Add dialect feature checks for arrays to use for testing.
- Change StandardTableExporter to recognize and use arrays when desired.
- Use option to determine to use legacy model (default) or new array model.

PR #3799, opened 2021-03-08
HHH-10999 Add support for SQL array types mapped as Java arrays

beikov:
This work and the tests are based on https://github.com/hibernate/hibernate-orm/pull/1499 for implementing https://hibernate.atlassian.net/browse/HHH-10999

https://hibernate.atlassian.net/browse/HHH-15288

This PR implements support for mapping basic (as in `@Basic`) Java arrays and collections to JDBC array types. It registers special `JavaType` variants for primitive array types, except for `byte[]` and `char[]` since these are mapped differently. For other array types, the java type resolving was updated to specially handle when an array `Class` is resolved, to resolve `ArrayJavaType`.

The resolving for collections works a bit differently and is in general extendible:

1. Initially, an attribute with a subtype of `Collection` is resolved to `CollectionJavaType`
2. `CollectionJavaType#createJavaType` refines this into the new `BasicCollectionJavaType`, which implements `BasicContainerJavaType` (like `ArrayJavaType`)

After this point, we treat all `BasicContainerJavaType` equally, regardless if we have arrays or collections.
`BasicContainerJavaType#resolveType` is called to determine the resulting `BasicType` based on the element `BasicType`.

The idea of `BasicContainerJavaType#resolveType` is, that every container type (array, collection etc.) parameterization will resolve to a dedicated `BasicType` which will contain a properly configured `ArrayJdbcType` instance. Properly configured means, that the `ArrayJdbcType` will know the array component `JdbcType` and in case of Oracle also the SQL type name for the Oracle varray. This is determined through the `ColumnTypeInformation` which essentially exposes the `Column#columnDefinition` with a fallback to asking `Dialect#getArrayTypeName`.

Here an outline of the changes:

* Add `ArrayJdbcType`
* Enable array types for H2 2.0+, HSQL, PostgreSQL, CockroachDB, Spanner and Oracle
* Introduce possibility to customize array DDL via `Dialect#getArrayTypeName`
* Add `ArrayJavaType` for mapping Java arrays `T[]`
* Register `XPrimitiveArrayJavaType` instances for primitive arrays in `JavaTypeBaseline`
* Add `getPreferredJavaTypeClass` to `JdbcType` to enable unwrapping to arrays of appropriate type
* Add array emulation based on XML/JSON types
* Introduce `ColumnTypeInformation` containing DDL information methods from `ColumnInformation` and use that in `Dialect#equivalentTypes` to allow type validation based on proper column type information
  => JDBC reports the ARRAY type code for array columns, so we need more information to do DDL validation, because `varchar[]` != `smallint[]`
  => `org.hibernate.mapping.Column` imlpements `ColumnTypeInformation` to provide necessary information for DDL validation

Here are some of the follow-ups that I have planned:

* Add support for `EnumSet` container type
* Add functions for accessing and replacing array elements (as well as JSON/XML emulation)
* Add functions for checking array contains one more elements (as well as JSON/XML emulation)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
@@ -2557,23 +2557,6 @@ public interface AvailableSettings {
 	String PREFERRED_INSTANT_JDBC_TYPE = "hibernate.type.preferred_instant_jdbc_type";
 
 	/**
-	 * Specifies the preferred JDBC type for storing basic array and collection values. When no
-	 * type is explicitly specified, a sensible
-	 * {@link org.hibernate.dialect.Dialect#getPreferredSqlTypeCodeForArray()
-	 * dialect-specific default type code} is used.
-	 *
-	 * Can be overridden locally using {@link org.hibernate.annotations.JdbcType},
-	 * {@link org.hibernate.annotations.JdbcTypeCode} and friends
-	 *
-	 * Can also specify the name of the {@link org.hibernate.type.SqlTypes} constant field.  E.g.
-	 * {@code hibernate.type.preferred_array_jdbc_type=VARBINARY}
-	 *
-	 * @since 6.1
-	 */
-	@Incubating
-	String PREFERRED_ARRAY_JDBC_TYPE = "hibernate.type.preferred_array_jdbc_type";
-
-	/**
 	 * Specifies a {@link org.hibernate.type.FormatMapper} used for JSON serialization
 	 * and deserialization, either:
 	 * <ul>
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
@@ -2557,6 +2557,23 @@ public interface AvailableSettings {
 	String PREFERRED_INSTANT_JDBC_TYPE = "hibernate.type.preferred_instant_jdbc_type";
 
 	/**
+	 * Specifies the preferred JDBC type for storing basic array and collection values. When no
+	 * type is explicitly specified, a sensible
+	 * {@link org.hibernate.dialect.Dialect#getPreferredSqlTypeCodeForArray()
+	 * dialect-specific default type code} is used.
+	 *
+	 * Can be overridden locally using {@link org.hibernate.annotations.JdbcType},
+	 * {@link org.hibernate.annotations.JdbcTypeCode} and friends
+	 *
+	 * Can also specify the name of the {@link org.hibernate.type.SqlTypes} constant field.  E.g.
+	 * {@code hibernate.type.preferred_array_jdbc_type=VARBINARY}
+	 *
+	 * @since 6.1
+	 */
+	@Incubating
+	String PREFERRED_ARRAY_JDBC_TYPE = "hibernate.type.preferred_array_jdbc_type";
+
+	/**
 	 * Specifies a {@link org.hibernate.type.FormatMapper} used for JSON serialization
 	 * and deserialization, either:
 	 * <ul>
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/AbstractArrayJavaType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/AbstractArrayJavaType.java
@@ -1,0 +1,88 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.type.descriptor.java;
+
+import java.sql.Types;
+
+import org.hibernate.dialect.Dialect;
+import org.hibernate.engine.jdbc.spi.JdbcServices;
+import org.hibernate.tool.schema.extract.spi.ColumnTypeInformation;
+import org.hibernate.type.BasicArrayType;
+import org.hibernate.type.BasicPluralType;
+import org.hibernate.type.BasicType;
+import org.hibernate.type.descriptor.jdbc.ArrayJdbcType;
+import org.hibernate.type.descriptor.jdbc.JdbcType;
+import org.hibernate.type.descriptor.jdbc.JdbcTypeIndicators;
+import org.hibernate.type.spi.TypeConfiguration;
+
+public abstract class AbstractArrayJavaType<T, E> extends AbstractClassJavaType<T>
+		implements BasicPluralJavaType<E> {
+
+	private final JavaType<E> componentJavaType;
+
+	public AbstractArrayJavaType(Class<T> clazz, BasicType<E> baseDescriptor, MutabilityPlan<T> mutabilityPlan) {
+		this( clazz, baseDescriptor.getJavaTypeDescriptor(), mutabilityPlan );
+	}
+
+	public AbstractArrayJavaType(Class<T> clazz, JavaType<E> baseDescriptor, MutabilityPlan<T> mutabilityPlan) {
+		super( clazz, mutabilityPlan );
+		this.componentJavaType = baseDescriptor;
+	}
+
+	@Override
+	public JavaType<E> getElementJavaType() {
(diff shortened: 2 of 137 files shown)


gavinking:
Could we get a write-up of what precisely this does and what were the design issues faced? Perhaps as a Discussion.

(Unless it's already documented elsewhere, but I didn't find anything.)

jwgmeligmeyling:
I know you havent touched the collection types yet, but if you could consider EnumSet together with the other collection types that would be awesome ;-) EnumSets mapped as element collections prove to be very viable replacements with Array types with quite significant performance gains, whereas the other array types I rarely see in the wild.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/GenericArrayTypeDescriptor.java
```diff
@@ -0,0 +1,358 @@
...
+			if ( ! componentDescriptor.areEqual( one[i], another[i] )) {
+				return false;
+			}
+		}
+		return true;
+	}
+
+	@Override
+	public int extractHashCode(T[] value) {
+		return java.util.Arrays.hashCode( value );
```
NathanQingyangXu, line 145: If we implement 'areEqual()' using componentDescriptor's 'areEqual()' internally, we might well use its 'extractHashCode' in the above 'extractHashCode()' implementation as well; otherwise there would be the danger that the contract between 'areEqual()' and 'extractHashCode()' might not be satisfied. A possible downside is we can insert some instance into a Map or Set but we won't be able to retrieve it.
Not 100% sure though.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/GenericArrayTypeDescriptor.java
```diff
@@ -0,0 +1,358 @@
...
+																+ " Outside of quote, but neither whitespace, comma, array end, nor null found." );
+				}
+			}
+			else if ( c == '\\' && i + 2 < len && (string.charAt( i + 1 ) == '\\' || string.charAt( i + 1 ) == '"')) {
+				c = string.charAt( ++i );
+			}
+			// If there is ever a null-pointer here, the if-else logic before is incomplete
+			sb.append( c );
+		}
+		String[] objects = lst.toArray( new String[lst.size()] );
```
NathanQingyangXu, line 254: In modern JVM, `list.toArray( new String[0] )` is always significantly faster than `list.toArray( new String[list.size()] )`, let alone much better code readability and maintainability. We might well go even further by using the various empty constants in ArrayHelper to avoid the performance cost of `new String[0]`.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -2556,6 +2556,23 @@ public interface AvailableSettings {
 	@Incubating
 	String PREFERRED_INSTANT_JDBC_TYPE = "hibernate.type.preferred_instant_jdbc_type";
 
+	/**
```
sebersole, line 2559: Just what I said I do not want to see ;)

gavinking: I agree.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -2556,6 +2556,23 @@ public interface AvailableSettings {
...
+	 * Specifies the preferred JDBC type for storing basic array and collection values. When no
+	 * type is explicitly specified, a sensible
+	 * {@link org.hibernate.dialect.Dialect#getPreferredSqlTypeCodeForArray()
+	 * dialect-specific default type code} is used.
+	 *
+	 * Can be overridden locally using {@link org.hibernate.annotations.JdbcType},
+	 * {@link org.hibernate.annotations.JdbcTypeCode} and friends
+	 *
+	 * Can also specify the name of the {@link org.hibernate.type.SqlTypes} constant field.  E.g.
```
sebersole, line 2568: Just a general thought as we now duplicate this exact sentence multiple times.  We should move this to somewhere central and refer to that from these field Javadocs.  

Not here, to be clear, as I was serious when I said I want to stop this new round of config explosion.  We worked hard in 6 dev to clean tons of settings and we REALLY need to be judicious here

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/BasicPluralJavaType.java
```diff
@@ -0,0 +1,43 @@
...
+ * This is used for array or collection types, that are backed by e.g. SQL array or JSON/XML DDL types.
+ *
+ * The interface can be implemented by a plural java type e.g. {@link org.hibernate.type.descriptor.java.spi.BasicCollectionJavaType}
+ * and provides access to the element java type, as well as a hook to resolve the {@link BasicType} based on the element {@link BasicType},
+ * in order to gain enough information to implement storage and retrieval of the composite data type via JDBC.
+ *
+ * @see org.hibernate.type.descriptor.java.spi.BasicCollectionJavaType
+ */
+@Incubating
+public interface BasicPluralJavaType<T> extends Serializable {
```
sebersole, line 29: Did you end up liking this name?

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/AbstractArrayJavaType.java
```diff
@@ -0,0 +1,88 @@
...
+import org.hibernate.tool.schema.extract.spi.ColumnTypeInformation;
+import org.hibernate.type.BasicArrayType;
+import org.hibernate.type.BasicPluralType;
+import org.hibernate.type.BasicType;
+import org.hibernate.type.descriptor.jdbc.ArrayJdbcType;
+import org.hibernate.type.descriptor.jdbc.JdbcType;
+import org.hibernate.type.descriptor.jdbc.JdbcTypeIndicators;
+import org.hibernate.type.spi.TypeConfiguration;
+
+public abstract class AbstractArrayJavaType<T, E> extends AbstractClassJavaType<T>
```
sebersole, line 22: I'd also rename this one to reflect whatever naming you choose with the "basic plurality".  Just based on name, I have no idea that this is limited to these basic pluralities compared to plural associations.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
beikov: I thought you meant the code for the preferred enum type code, but also the one for arrays? I only introduced that to allow backwards compatibility, you know, the thing we will have to keep attention to in order for EAP to get updates 😉
But seriously, I can certainly remove this as well, though I believe this particular setting is valuable.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/AbstractArrayJavaType.java
sebersole: Not fond of he `AbstractClassJavaType` name either in retrospect.  But probably not worth changing now

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
sebersole: No EAP version has 6.0, which also means that no EAP version supports basic arrays.  Nothing to see here, move along...

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/BasicPluralJavaType.java
beikov: Not really, but I don't really have a preference and also no better idea for a name.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
beikov: Going to be removed anyway

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
beikov: Alright, as discussed on Zulip, we will remove the setting, as we don't think anyone is relying on this and really want that particular behavior. So if someone complains later, we might re-add it, but not now.

PR #3801, opened 2021-03-08
Non-select native query support and fix parameter binding ordering issues for select native queries

beikov:
Factored out from #3799


beikov:
Note that this PR causes `MultiLevelCascadeCollectionIdClassTest` and `MultiLevelCascadeCollectionEmbeddableTest` to fail because now the tests don't fail anymore due to `NotImplementedYetException` in the native query execution which was expected so far. This PR thus depends on https://github.com/hibernate/hibernate-orm/pull/3802 which will restore the `NotImplementedYetException`, but for the right reasone, because non-pk FKs are not yet implemented.

sebersole (approved):
Looks great!

PR #3802, opened 2021-03-08
Implement support for delayed collections with FKs based on non-primary keys

beikov:
Factored out from #3799


dreab8 (approved):
it looks good to me :)

gavinking:
What does "delayed collections" mean?

beikov:
>What does "delayed collections" mean?

@gavinking: Represents a "lazy collection" where we just fetch the FK to be able to create a `PersistenetCollection`. I didn't come up with that term, but I assume that nobody has a problem with changing the name if you come up with something better.

gavinking:
I still don't understand. Is there a test for this, or an example somewhere?

PR #3803, opened 2021-03-08
Switch from java.lang.Class to java.lang.reflect.Type in the metamodel to support parameterized types

beikov:
Factored out from #3799


sebersole (approved):
Much better

PR #3807, opened 2021-03-09
HHH-14482 - Do not discard prior implicit join by key

dreab8:
https://hibernate.atlassian.net/browse/HHH-14482


dreab8:
Merged upstream,

@beikov and @sebersole thanks for the review!

PR #3814, opened 2021-03-10
Use FK descriptor instead of target entity identifier

beikov:
I just saw these uses which were using the PK instead of relying on the FK descriptor.


dreab8:
it looks good to me!

sebersole (approved):

PR #3815, opened 2021-03-10
Wip/6.0 more envers fixes

Naros:
Various fixes to get more Envers tests to pass.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/StandardSortedMapSemantics.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/StandardSortedMapSemantics.java
@@ -49,9 +49,8 @@ public class StandardSortedMapSemantics<K,V> extends AbstractMapSemantics<Sorted
 			Object key,
 			CollectionPersister collectionDescriptor,
 			SharedSessionContractImplementor session) {
-		final PersistentSortedMap<K,V> result = new PersistentSortedMap<>( session );
-		result.setComparator( (Comparator<K>) collectionDescriptor.getSortingComparator() );
-		return result;
+		//noinspection unchecked
+		return new PersistentSortedMap<>( session, (Comparator<K>) collectionDescriptor.getSortingComparator() );
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/collection/internal/StandardSortedMapSemantics.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/internal/StandardSortedMapSemantics.java
@@ -49,7 +49,9 @@ public class StandardSortedMapSemantics<K,V> extends AbstractMapSemantics<Sorted
 			Object key,
 			CollectionPersister collectionDescriptor,
 			SharedSessionContractImplementor session) {
-		return new PersistentSortedMap<>( session );
+		final PersistentSortedMap<K,V> result = new PersistentSortedMap<>( session );
+		result.setComparator( (Comparator<K>) collectionDescriptor.getSortingComparator() );
+		return result;
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -947,7 +947,14 @@ public abstract class Dialect implements ConversionContext {
 	}
 
 	public String getTypeName(int code) throws HibernateException {
-		return getTypeName( code, Size.nil() );
+		switch ( code ) {
+			case Types.TIMESTAMP:
+			case Types.TIMESTAMP_WITH_TIMEZONE:
+				// explicitly enforce dialect's default timestamp precision
+				return getTypeName( code, Size.precision( getDefaultTimestampPrecision() ) );
+			default:
+				return getTypeName( code, Size.nil() );
+		}
 	}
 
 	/**
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -947,10 +947,18 @@ public abstract class Dialect implements ConversionContext {
 	}
 
 	public String getTypeName(int code) throws HibernateException {
+		// explicitly enforce dialect's default precisions
 		switch ( code ) {
+			case Types.DECIMAL:
+			case Types.NUMERIC:
+				return getTypeName( code, Size.precision( getDefaultDecimalPrecision() ) );
+			case Types.FLOAT:
+			case Types.REAL:
+				return getTypeName( code, Size.precision( getFloatPrecision() ) );
+			case Types.DOUBLE:
+				return getTypeName( code, Size.precision( getDoublePrecision() ) );
 			case Types.TIMESTAMP:
 			case Types.TIMESTAMP_WITH_TIMEZONE:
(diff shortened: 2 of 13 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -947,7 +947,14 @@ public String getTypeName(SqlTypeDescriptor sqlTypeDescriptor) throws HibernateE
 	}
 
 	public String getTypeName(int code) throws HibernateException {
-		return getTypeName( code, Size.nil() );
+		switch ( code ) {
+			case Types.TIMESTAMP:
+			case Types.TIMESTAMP_WITH_TIMEZONE:
+				// explicitly enforce dialect's default timestamp precision
+				return getTypeName( code, Size.precision( getDefaultTimestampPrecision() ) );
```
beikov, line 954: Shouldn't we then also use the default precision for `DECIMAL`/`NUMERIC` (`Dialect.getDefaultDecimalPrecision`), `FLOAT`/`REAL` (`Dialect.getFloatPrecision`) and `DOUBLE` (`Dialect.getDoublePrecision`)?

hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/DelayedCollectionInitializer.java
```diff
@@ -76,7 +76,9 @@ public void resolveKey(RowProcessingState rowProcessingState) {
 //						LoggingHelper.toLoggableString( getNavigablePath(), this.collectionKey.getKey() )
 //				);
 //			}
-			parentAccess.registerResolutionListener( owner -> collectionInstance.setOwner( owner ) );
+			if ( collectionInstance != null ) {
```
beikov, line 79: Shouldn't this if move into the lambda instead? The field is only initialized when `resolveInstance` is called.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/DelayedCollectionInitializer.java
```diff
@@ -85,7 +87,9 @@ public void resolveKey(RowProcessingState rowProcessingState) {
 						collectionAttributeMapping.getCollectionDescriptor(),
 						parentKey
 				);
-				parentAccess.registerResolutionListener( owner -> collectionInstance.setOwner( owner ) );
+				if ( collectionInstance != null ) {
```
beikov, line 90: Again, shouldn't this if move into the lambda instead? The field is only initialized when `resolveInstance` is called.

hibernate-core/src/main/java/org/hibernate/collection/internal/StandardSortedMapSemantics.java
```diff
@@ -49,7 +49,9 @@ public CollectionClassification getCollectionClassification() {
 			Object key,
 			CollectionPersister collectionDescriptor,
 			SharedSessionContractImplementor session) {
-		return new PersistentSortedMap<>( session );
+		final PersistentSortedMap<K,V> result = new PersistentSortedMap<>( session );
+		result.setComparator( (Comparator<K>) collectionDescriptor.getSortingComparator() );
```
beikov, line 53: It seems to me that the setter `setComparator` is unnecessary and should be removed. IMO the comparator should be `final` in `PersistentSortedMap` as well as `PersistentSortedSet` and be initialized through the constructor.

hibernate-envers/src/main/java/org/hibernate/envers/internal/entities/mapper/relation/query/AbstractRelationQueryGenerator.java
```diff
@@ -65,7 +67,11 @@ public Query getQuery(SharedSessionContractImplementor session, Object primaryKe
 		final String queryString = getQueryString( session.getFactory(), removed );
 
 		final Query query = session.createQuery( queryString );
-		query.setParameter( DEL_REVISION_TYPE_PARAMETER, RevisionType.DEL );
+		query.setParameter(
+				DEL_REVISION_TYPE_PARAMETER,
+				RevisionType.DEL,
+				new CustomType( new RevisionTypeType(), session.getFactory().getTypeConfiguration() )
```
beikov, line 73: Wouldn't `sessionFactory.getTypeConfiguration().getBasicTypeRegistry().getRegisteredType( RevisionTypeType.class.getName() )` also work instead of creating a new type all the time? The same would apply to `org.hibernate.envers.internal.tools.query.QueryBuilder#toQuery`

beikov (changes requested):

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
Naros: Quite possibly.  I can certainly add those here but the Envers tests don't have test cases for those tmk.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/DelayedCollectionInitializer.java
Naros: I don't think so because the lamda firing is what yielded the NPE in the first place because `collectionInstance` was not yet resolved.  We can certainly rework it that way but you'd be adding a lamda that's firing for absolutely no reason.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/DelayedCollectionInitializer.java
Naros: See my earlier comment.

hibernate-core/src/main/java/org/hibernate/collection/internal/StandardSortedMapSemantics.java
Naros: We can certainly look at passing it via the ctor so long as are willing to remove the ctor that takes the `SessionImplementor` argument that is deprecated.  The setter however needs to remain to support SOAP libraries, as indicated in the no-arg ctor.

hibernate-envers/src/main/java/org/hibernate/envers/internal/entities/mapper/relation/query/AbstractRelationQueryGenerator.java
Naros: I'll give it a look, I simply reused what was being used in QueryBuilder that had been changed for 6.0 earlier.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/DelayedCollectionInitializer.java
Naros: I went ahead and moved it into the lambda for now as requested.

hibernate-envers/src/main/java/org/hibernate/envers/internal/entities/mapper/relation/query/AbstractRelationQueryGenerator.java
Naros: Worked like a charm, thanks @beikov !

beikov (approved):
Looks good now!

PR #3816, opened 2021-03-11
L2 in memory cache implementation using redis for invalidation and redis connect 

nucleussoftwareopen:
Change Summary:
1. L2 in memory cache implementation using redis for invalidation.
2. Provision to connect to redis without authentication
A flag "hib.redis.server.password.required" is added to make it optional to connect to redis with or without authentication.
It is by default true. To connect to redis without authentication the flag needs to to be set to false.
Any other value than false will be treated as true.


nucleussoftwareopen:
This change is based on Hibernate 5.2.8 version

PR #3821, opened 2021-03-12
HHH-14475 - select indices from ternary relation is broken since 5.4.…

drahkrub:
…13 (at least in mariadb)

a simple test case for https://hibernate.atlassian.net/browse/HHH-14475 - succeeds in 5.4.12, fails in 5.4.13 onwards


beikov:
Thanks for the test case. This PR is superseded by https://github.com/hibernate/hibernate-orm/pull/3827

PR #3824, opened 2021-03-15
HHH-11692 Introduce DISTINCT predicate in HQL and JPA Criteria extensions and implement emulations

beikov:
[HHH-11692](https://hibernate.atlassian.net/browse/HHH-11692)

This is a long requested feature. The example criterion in the legacy Criteria API used the same semantics. This PR introduces the SQL standard `DISTINCT` predicate for this purpose. I can imagine adding an operator as well, but that would be a minor addition to the parser and I wanted to get the feature done first.

I also deprecated some methods of `Dialect`:

* `getFromDual()`
* `supportsSelectQueryWithoutFromClause()`
* `supportsRowValueConstructorSyntax()`
* `supportsRowValueConstructorSyntaxInSet()`
* `supportsRowValueConstructorSyntaxInQuantifiedPredicates()`
* `supportsRowValueConstructorSyntaxInInList()`
* `supportsRowValueConstructorSyntaxInInSubquery()`
* `getGroupBySummarizationRenderingStrategy()`
* `getGroupByConstantRenderingStrategy()`
* `requiresCastingOfParametersInSelectClause()`
* `supportsTuplesInSubqueries()`

These were moved to `AbstractSqlAstTranslator` and/or inlined into the `SqlAstTranslator` implementations for the dialects. Since some of these methods were only introduced in 6.0 and are now unused anyway, I would suggest we remove them in 6.0.

The reasoning for this is that I wanted to inline as much of the logic as possible into the `SqlAstTranslator` implementations to avoid cluttering the code with even more if-else chains to handle emulations for the `DISTINCT` predicate. Since these methods were only ever used within the `SqlAstTranslator` implementations anyway, it's IMO nicer to move this "supports" logic there.

### DIFF
--- a/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
+++ b/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
@@ -415,6 +415,8 @@ comparisonOperator
 	| GREATER_EQUAL
 	| LESS
 	| LESS_EQUAL
+	| IS DISTINCT FROM
+	| IS NOT DISTINCT FROM
 	;
 
 inList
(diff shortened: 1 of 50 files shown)


gavinking:
Nice!

sebersole (approved):

PR #3826, opened 2021-03-15
HHH-14499 Explicitly listing mapped classes which are @MappedSuperClass might break narrowing down of generics

dreab8:
https://hibernate.atlassian.net/browse/HHH-14499


hibernate-core/src/test/java/org/hibernate/test/mapping/MappedSuperclassWithGenericsTest.java
```diff
@@ -0,0 +1,92 @@
...
+		}
+	}
+
+	@Entity(name = "BaseEntity")
+	public static class BaseEntity<T> extends IntermediateAbstractMapped<byte[]> {
+
+		String aString;
+
+		/**
+		 * Default constructor required by JPA
```
Sanne, line 86: I'll remove the comment

Sanne:
Removed a comment then merged. Thanks!

PR #3828, opened 2021-03-15
HHH-14503 - Migration of tests from jpa/test to orm/test/jpa

jrenaat:


dreab8:
Hi @jrenaat , it seems there are failures with tests running on MariaDB

beikov (commented):
The `tearDown` methods of `GetLoadTest`, `MergeTest` and `PersistTest` use a delete query for deleting `Node` and it seems MariaDB deletes rows in an order where it violates the FK constraint temporarily. Other databases seem to be smart enough to check FKs by default only after a statement has finished. I suggest you add a statement `entityManager.createQuery( "update from Node set parent = null" ).executeUpdate();` before doing the delete to resolve this.

dreab8 (approved):
It looks good to me

PR #3834, opened 2021-03-17
HHH-14449 Handle explicit onClose registrations on Query.getResultStream() 

juja0:
This PR originally started off with the intent to remove StreamDecorator and related classes based on the discussion over at https://discourse.hibernate.org/t/unexpected-getresultstream-behavior-onclose-not-called/4386/4?u=juja

Based on some test failures after the initial commit, it appears that the intent of StreamDecorator is to ensure that the underlying database resources are closed even if developers forget to explicitly close the java Stream returned from Query.getResultStream() (either via an explicit call to onClose or using try-with-resources). I am not sure if hibernate should do that since if a developer forgets to close a Stream that's probably on the developer. That being said, I do see that an argument can be made that most people won't think of explicitly closing a java Stream since they are usually used with collections which don't need to be closed explicitly. I feel creating a wrapper over java internals may make the code brittle but that's just my opinion. In any case, since StreamDecorator has been around for a while and the current behaviour is to auto-close after terminal operations, removing StreamDecorator may end up being a breaking change. So I've re-instated the classes and changed the PR to instead make sure that explicitly registered onClose callbacks are not ignored.   

This is done by making sure that a call to onClose on StreamDecorator is propagated to the underlying Stream.


sebersole:
No need to say "don't merge; in progress".  Just mark the PR as a draft :)

juja0:
> No need to say "don't merge; in progress". Just mark the PR as a draft :)

Noted :)

sebersole:
> I am not sure if hibernate should do that since if a developer forgets to close a Stream that's probably on the developer

Lots of things are "on the developer" but Hibernate is still expected to handle it.  IMO this is one of them

juja0:
> > I am not sure if hibernate should do that since if a developer forgets to close a Stream that's probably on the developer
> 
> Lots of things are "on the developer" but Hibernate is still expected to handle it. IMO this is one of them

Fair enough. My worry was that hibernate would have to keep playing catch-up with the JDK for every new method that's added to Stream.java, but I guess that's better than having a potential resource leak in a critical application (considering how widespread hibernate usage is).

beikov (commented):
Apart from a few code style issues, this looks good to me.

juja0:
Update: Creating a custom Spliterator works (sort of). 

The implementation automatically closes the underlying database resource after finishing iteration (or any terminal operation for that matter), or if iteration fails (in case of exceptions in user code), or if iteration is cancelled (in case of partial iteration via takeWhile, dropWhile, filter etc).

Caveat: This will only close the underlying database connection that hibernate manages and will not *automatically* call any user-registered onClose callbacks. Note however that user-registered onClose callbacks will still be called if the Stream is closed by the user (either via an explicit call to close or using try-with-resources) which is inline with how java.util.Stream works. I feel this is better because hibernate is still closing the underlying database resource but makes no assumptions about user-registered onClose callbacks. This will also not be a breaking change because onClose never used to work before (even if the stream is explicitly closed).

Considering we now have two potential routes that we can take, I would like to hear some thoughts on which one we should go forward with and if I've missed anything in my analysis. 

Attaching the implementation of AutoClosingSpliterator for reference.

[AutoClosingSpliterator.txt](https://github.com/hibernate/hibernate-orm/files/6162624/AutoClosingSpliterator.txt)

juja0:
@beikov I think we commented at the same time :) Would you mind taking a look at my comment and let me know your thoughts ?

beikov:
IMO the solution you provide in this PR is good. I would just apply it.

> hibernate would have to keep playing catch-up with the JDK

AFAICT the JDK can't just add any method as that would be backwards incompatible. They have to do that through default methods, so IMO that is not a concern.

juja0:
To remain backward compatible, the JDK will make sure not to change signature of existing methods but I think new methods can still be added. For example, takeWhile and dropWhile were introduced in JDK9 and StreamDecorator had to be enhanced to support those two methods (via reflection no less).

And java 16 has introduced a bunch of new methods which then have to be implemented in StreamDecorator as well.

That being said, I am happy to apply this PR in the current state and revisit this discussion later if it ends up being as much of an issue as I'm imagining it would be.

P.S. I've updated the codestyle as well. I used the codeStyle XML from CONTRIBUTING.MD but reformatting the entire file ended up with too many changed lines so I've formatted only lines which I have changed. Hope that's fine.

beikov:
Sure, these methods were introduced, but with default implementations. See here for example: https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/stream/Stream.java#L759

So there is no need for Hibernate to adapt for those new methods. Hibernate can of course add a custom implementation if it can do better than the default implementation, but that's an optimization.

juja0:
Ah I see your point. I missed the 'default' modifier on those methods. 

Feel free to merge the PR if everything else looks good. Thanks.

sebersole:
Squashed, rebased and applied upstream.

Thanks @juja0

PR #3839, opened 2021-03-18
Wip/6.0 Cannot resolve path

dreab8:
Fix Cannot resolve path exception with EmbeddedId and hql queries not using aliases, this fixes another bunch of envers tests

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/EmbeddedIdTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/EmbeddedIdTest.java
@@ -1,0 +1,198 @@
+package org.hibernate.orm.test.query.hql;
+
+import java.io.Serializable;
+import javax.persistence.Embeddable;
+import javax.persistence.EmbeddedId;
+import javax.persistence.Entity;
+import javax.persistence.Id;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/EmbeddedIdTest.java
```diff
@@ -0,0 +1,198 @@
...
+			final Document document = new Document( 1, "DL" );
+			final PersonDocument pd = new PersonDocument( person, document );
+			entityManager.persist( person );
+			entityManager.persist( document );
+			entityManager.persist( pd );
+		} );
+	}
+
+	@AfterEach
+	public void tearDowm(SessionFactoryScope scope) {
```
beikov, line 44: Minor typo

```suggestion
	public void tearDown(SessionFactoryScope scope) {
```

beikov (approved):

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/EmbeddedIdTest.java
dreab8: Thanks @beikov , fixed!

PR #3840, opened 2021-03-18
HHH-14514 Auto evict collection cache not work when use transactional cache concurrency strategy

pabro:
Test case and fix


beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/6127

PR #3841, opened 2021-03-18
Migrate tests and fix query cache stats, native query variable substitution and named query support as loader

beikov:


hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
```diff
@@ -643,6 +646,17 @@ public void contributeTypes(TypeContributions typeContributions, ServiceRegistry
 
 			typeContributions.contributeSqlTypeDescriptor( descriptor );
 		}
+
+		// We need a special type descriptor for Oracle, because that must return BIT as SQL type
+		// in order for PreparedStatement#setNull to work properly
+		typeContributions.contributeSqlTypeDescriptor(
```
sebersole, line 652: I really think we should try `Types#NULL` for the cases where we need to set null values.  No idea if that works for all drivers/databases but if so its a much better option IMO than randomly picking one

gavinking: The other option is to delegate binding nulls to the `WrapperOptions` object. It's what it exists for.

We could either let `WrapperOptions` remap "null" JDBC type codes, or we could let it actually handle the whole call to `setNull()`.

sebersole: Or even delegate back to Dialect through WrapperOptions

gavinking: Yep, right, that's what I mean. Should have been more explicit.

hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
```diff
@@ -643,6 +653,79 @@ public void contributeTypes(TypeContributions typeContributions, ServiceRegistry
 
 			typeContributions.contributeSqlTypeDescriptor( descriptor );
 		}
+		typeContributions.contributeSqlTypeDescriptor( new OracleBooleanTypeDescriptor() );
+	}
+
+	private static final class OracleBooleanTypeDescriptor extends BooleanTypeDescriptor {
+
+		private static final Logger log = CoreLogging.logger( BasicBinder.class );
+
```
sebersole, line 662: We should probably instead create a "subsystem" logger here and use that from the various impls.  More just a mental todo...

sebersole (approved):
Awesome!

sebersole:
I say squash and push

sebersole:
Nice!  And just in time :)

PR #3844, opened 2021-03-18
HHH-10661 - Reduce code duplication in serialize method of StatefulPersistenceContext

beikov:
https://hibernate.atlassian.net/browse/HHH-10661

Supersedes https://github.com/hibernate/hibernate-orm/pull/2357

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
@@ -1572,117 +1572,86 @@ public class StatefulPersistenceContext implements PersistenceContext {
 		oos.writeBoolean( defaultReadOnly );
 		oos.writeBoolean( hasNonReadOnlyEntities );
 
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
```diff
@@ -1572,117 +1572,86 @@ public void serialize(ObjectOutputStream oos) throws IOException {
...
 		entityEntryContext.serialize( oos );
+		writeMapToStream(
+				collectionsByKey,
+				oos,
+				"collectionsByKey",
+				(entry, stream) -> {
+					entry.getKey().serialize( stream );
+					stream.writeObject( entry.getValue() );
+				}
+		);
```
NathanQingyangXu, line 1600: It seems the 'entityKeySerializer' could be reused for 'entitiesByUniqueKey' and 'collectionsByKey' as well, no?

Sanne:
I will apply this, nice cleanup thanks

PR #3850, opened 2021-03-19
cleaner workaround for Oracle setNull(BOOLEAN) bug

gavinking:
Here's what I think is a better way to do this:

I've introduced `doBindNull()` in `BasicBinder` and I'm using `WrapperOptions`, `FastSessionServices`, and `getPreferredSqlTypeCodeForBoolean()` for what they're designed for.

Perhaps one might object that adding `doBindNull()` introduces an additional polymorphic call when binding null values, but if that overhead were measurable, then it would actually be worse in the case of the regular `doBind()` call that exists today for all basic types when the value is not-null, and would be an argument for eliminating `BasicBinder` altogether. So I don't see any reasonable objection to this on performance grounds.


gavinking:
Merging this because the main substantive change is also needed by #3974, and because it lets us delete a whole class.

PR #3853, opened 2021-03-21
HHH-14519 More informative error message for bad logical column name

karge-itestra:
When a logical column name identifier cannot be created, there is only an error message that the "column name must not be null". In the case I experienced, the column name was in fact empty and not null, but the column identifier could not be found. This changes the error message to make the cause more clear and also give information about the context (table name, column name given)

https://hibernate.atlassian.net/browse/HHH-14519


hibernate-core/src/main/java/org/hibernate/boot/internal/InFlightMetadataCollectorImpl.java
```diff
@@ -981,15 +981,20 @@ public void addColumnNameBinding(Table table, Identifier logicalName, Column col
 
 	@Override
 	public String getPhysicalColumnName(Table table, String logicalName) throws MappingException {
-		return getPhysicalColumnName( table, getDatabase().toIdentifier( logicalName ) );
+		Identifier logicalColumnIdentifier = getDatabase().toIdentifier(logicalName);
+		if (logicalColumnIdentifier == null) {
+			throw new MappingException( String.format(
+					Locale.ENGLISH,
+					"Column with logical name \"[%s]\" in table [%s] cannot be mapped to column identifier.",
```
gavinking, line 988:
```suggestion
					"Column with logical name '%s' in table '%s' cannot be mapped to column identifier",
```

gavinking:
Thanks! Since this PR is out of date, I redid it in #9324. I credited you in the commit message.

PR #3857, opened 2021-03-22
HHH-14521 Auto evict collection cache always evict entire region for XML mapping entities

pabro:


gavinking:
`hbm.xml` is deprecated, and we aren't making improvements to it, especially not for things which aren't even enabled by default. Sorry.

PR #3859, opened 2021-03-23
Various fixes for test migrations

beikov:
* Add support for DML with versioned entities
* Resolve return type for `SUM` according to JPA spec
* Specify invariant return types for `SQRT` and `MOD` as required by the JPA spec
* Fix JPA tuple element access support
* Fix join management for JPA related methods
* Handle optional escape character for `LIKE` predicate
* Implement type inference for result arms of `CASE` expressions
* Implement min/max element/index functions as sub-query
* Implement min/max function support
* Implement emptiness, exists and member of predicate for JPA Criteria
* Implement size function as sub-query
* Implement group by entity alias by using FK key

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
@@ -6,11 +6,25 @@
  */
 package org.hibernate.dialect.function;
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.sql.Types;
+import java.util.List;
 import java.util.Arrays;
+import java.util.function.Supplier;
+
+import org.hibernate.metamodel.mapping.BasicValuedMapping;
+import org.hibernate.metamodel.model.domain.AllowableFunctionReturnType;
 
 import org.hibernate.query.spi.QueryEngine;
+import org.hibernate.query.sqm.produce.function.FunctionReturnTypeResolver;
+import org.hibernate.query.sqm.produce.function.StandardFunctionReturnTypeResolvers;
+import org.hibernate.query.sqm.tree.SqmTypedNode;
 import org.hibernate.sql.ast.SqlAstNodeRenderingMode;
+import org.hibernate.sql.ast.tree.SqlAstNode;
+import org.hibernate.type.BasicType;
 import org.hibernate.type.StandardBasicTypes;
+import org.hibernate.type.spi.TypeConfiguration;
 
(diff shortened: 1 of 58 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
```diff
@@ -1476,7 +1490,83 @@ public static void aggregates(QueryEngine queryEngine) {
...
+									return bigIntegerType;
+								}
+								else {
+									return bigDecimalType;
+								}
+						}
+						return impliedTypeAccess.get();
+					}
+
+				} )
```
gavinking, line 1569: Good catch! 

I did not notice that the JPA spec (Section 4.8.5) is not silent on this issue:

> SUM returns Long when applied to state fields of integral types (other than BigInteger); Double when applied to state fields of floating point types; BigInteger when applied to state fields of type BigInteger; and BigDecimal when applied to state fields of type BigDecimal.

Could you do me a favor and copy that spec language into a comment, so we know why this code is necessary.

hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
```diff
@@ -1511,6 +1601,7 @@ public static void math(QueryEngine queryEngine) {
 				.register();
 
 		queryEngine.getSqmFunctionRegistry().namedDescriptorBuilder("mod")
+				.setInvariantType( StandardBasicTypes.INTEGER )
```
gavinking, line 1604: This is defined in spec section 4.6.17.2.2.

hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
```diff
@@ -1524,6 +1615,7 @@ public static void math(QueryEngine queryEngine) {
 				.register();
 
 		queryEngine.getSqmFunctionRegistry().namedDescriptorBuilder("sqrt")
+				.setInvariantType( StandardBasicTypes.DOUBLE )
```
gavinking, line 1618: This is defined in spec section 4.6.17.2.2.

gavinking (changes requested):
Looks. good

And it must have been me who removed the logic about assigning types to `sum()`, `mod()`, and `sqrt()`. So thanks for putting it back.

In my defense, there were no comments explaining why we had those rules, so I would like to see comments pointing to the relevant sections of the JPA spec. It's really important that we document such things.

PR #3860, opened 2021-03-23
HHH-13779 (5.4) - Foreign key schema migrator should be case-insensitive

educhastenier:
This avoids re-creating existing foreign keys with a different name,
after migrating from Hibernate 4 to Hibernate 5 (as implicit naming
convention has changed).
Actually, some RDBMS allow it (PostgreSQL, MySQL, MS SQL Server, ...)
and duplicate the same key, whereas others (Oracle, ...) do not
allow it and Schema update fails.

This fix ignores the case of the table and column name when checking
if a equivalent Foreign Key already exists (whatever its name)

More details of the issue on https://hibernate.atlassian.net/browse/HHH-13779

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/tool/schema/internal/AbstractSchemaMigrator.java
+++ b/hibernate-core/src/main/java/org/hibernate/tool/schema/internal/AbstractSchemaMigrator.java
@@ -463,21 +463,24 @@ public abstract class AbstractSchemaMigrator implements SchemaMigrator {
 		 * Find existing keys based on referencing column and referencedTable. "referencedColumnName" is not checked
 		 * because that always is the primary key of the "referencedTable".
 		 */
-		Predicate<ColumnReferenceMapping> mappingPredicate = m -> {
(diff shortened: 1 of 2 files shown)


sebersole (approved):
LGTM

For future, please submit pull requests against the main branch.

sebersole:
applied upstream

FHannes:
I also recently identified this problem, however I do believe that this fix does break the check in specific cases. When a [delimited identifier](https://crate.io/docs/sql-99/en/latest//chapters/02.html#delimited-identifier) is used, it is possible to create multiple tables with the same name, but different casing. This is obviously not a practice that should be recommended given it would be very ambiguous, but it is covered by the standard and an implementation such as PostgreSQL will happily allow you to do it.

A solution that covers all cases should check if the tablename/columnname is a delimited identifier, which can be achieved by checking if the tablename specified in the JPA annotations is "quoted", as it is referred to in Hibernate. In the method `Identifier.toIdentifier()`, these quotes are removed for example. The check should only be case-insensitive if the identifier is not delimited.

PR #3867, opened 2021-03-24
HHH-6054 add @TenantId multitenancy and remove MultiTenancyStrategy

gavinking:
Use the existence of a `MultiTenantConnectionProvider` to infer that multitenancy is enabled.

Thoughts?

UPDATE: this PR now includes an implementation of discriminator-based multitenancy.


documentation/src/main/asciidoc/userguide/chapters/multitenancy/MultiTenancy.adoc
```diff
@@ -46,17 +46,6 @@ Using this approach, we would have a distinct JDBC Connection pool per-tenant wh
 * Connections could point to the database itself (using some default schema) but the Connections would be altered using the SQL `SET SCHEMA` (or similar) command.
 Using this approach, we would have a single JDBC Connection pool for use to service all tenants, but before using the Connection, it would be altered to reference the schema named by the "tenant identifier" associated with the currently logged in user.
 
-[[multitenacy-discriminator]]
```
sebersole, line 49: Not sure why we are removing this section.  This is still a known/valid form of multi-tenancy.  This section is just discussing the various forms of multi-tenancy.  Just note below that support for this form is not (yet) implemented

sebersole (commented):
I'd like to see the documentation change be done differently as I commented.  Otherwise looks good

gavinking:
> I'd like to see the documentation change be done differently as I commented. Otherwise looks good

OK, no problem.

gavinking:
I've just rebased this PR, fixed conflicts, and added back that doc section that @sebersole wanted to keep.

gavinking:
So there was a failure on Sybase, which I understand is expected.

And there was a failure on Oracle in `LockNoneWarmingTest.testQuerySetLockModeNONEDoNotLogAWarnMessageWhenTheDialectUseFollowOnLockingIsTrue` which I assume cannot possibly be related to this change.

So @sebersole, IMO, this is ready to go.

beikov:
Yeah you can ignore these failures. I think though, that the "discriminator" strategy should be implemented at some point. One user on the issue for [implementing support for the discriminator strategy](https://hibernate.atlassian.net/browse/HHH-6054) wrote:

> I'm currently using Filter to do this, but it's not very practical because you need an interceptor to always activate it.

And followed up with an issue about global activation of a filter: https://hibernate.atlassian.net/browse/HHH-3890

I think this is an approach we could take for the discriminator strategy and IMO this is also a pretty good solution that many people probably use. I agree though that the distinction between SCHEMA and DATABASE is not very useful though, so +1 from me on removing that.

gavinking:
In order to demonstrate that `MultiTenancyStrategy` is entirely unnecessary in order to have discriminator-based multitenancy, I went and added a roughish implementation of discriminator-based multitenancy.

This was even easier than I had expected it to be.

All the user has to do is annotate a field of their entity `@TenantId`.

gavinking:
> And followed up with an issue about global activation of a filter: https://hibernate.atlassian.net/browse/HHH-3890

Well that issue seems to be nonsense. Filters already apply to all queries run in a `Session`.

Not convinced the user actually understands how filters work.

beikov:
> Well that issue seems to be nonsense. Filters already apply to all queries run in a Session.
> Not convinced the user actually understands how filters work.

The user means that the filter is enabled by default i.e. without having to call `session.enableFilter`

hibernate-core/src/main/java/org/hibernate/tuple/TenantIdGeneration.java
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	@Override
+	public ValueGenerator<?> getValueGenerator() {
+		return this;
+	}
+
+	@Override
+	public Object generateValue(Session session, Object owner) {
+		return session.getTenantIdentifier();
```
beikov, line 35: Does this still allow to set the tenant identifier manually?

hibernate-core/src/test/java/org/hibernate/test/tenantid/TenantIdTest.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+)
+public class TenantIdTest implements SessionFactoryProducer {
+
+    String currentTenant;
+
+    @Override
+    public SessionFactoryImplementor produceSessionFactory(MetadataImplementor model) {
+        final SessionFactoryBuilder sessionFactoryBuilder = model.getSessionFactoryBuilder();
+        sessionFactoryBuilder.applyCurrentTenantIdentifierResolver( new CurrentTenantIdentifierResolver() {
```
beikov, line 40: It would be awesome if we could update `org.hibernate.boot.registry.selector.internal.StrategySelectorImpl#resolveDefaultableStrategy(java.lang.Class<T>, java.lang.Object, java.util.concurrent.Callable<T>)` to allow instantiating a "bean" i.e. allow dependency injection to happen into the resolver instance.

hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
```diff
@@ -241,6 +241,18 @@ public SessionImpl(SessionFactoryImpl factory, SessionCreationOptions options) {
 			setHibernateFlushMode( initialMode );
 		}
 
+		if ( factory.getDefinedFilterNames().contains("_tenantId") ) {
```
beikov, line 244: IMO `"_tenantId"` should be moved into a constant, just like the parameter name `tenantId`

hibernate-core/src/main/java/org/hibernate/cfg/annotations/PropertyBinder.java
```diff
@@ -190,6 +197,31 @@ private Property makePropertyAndValue() {
...
+			InFlightMetadataCollector collector = buildingContext.getMetadataCollector();
+			collector.addFilterDefinition(
+					new FilterDefinition(
+							"_tenantId",
+							"",
+							singletonMap(
+									"tenantId",
+									collector.getTypeConfiguration()
+											.getBasicTypeRegistry()
+											.getRegisteredType(String.class)
```
beikov, line 212: It would be cool if we could support other kinds of identifiers as well i.e. integers are very common.

hibernate-core/src/main/java/org/hibernate/annotations/TenantId.java
```diff
@@ -0,0 +1,28 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Identifies a field of an entity that holds a tenant id
+ * in discriminator-based multitenancy.
+ *
+ * @author Gavin King
+ */
+@ValueGenerationType(generatedBy = TenantIdGeneration.class)
+@Target({TYPE, METHOD, FIELD})
```
beikov, line 26: Do you think it makes sense to allow `@TenantId(columnName = "tenant_id")` on the class level or would we require the tenant id to always be mapped?

beikov (commented):
Looks great, although I think we should add a section in the documentation as well.

hibernate-core/src/main/java/org/hibernate/tuple/TenantIdGeneration.java
gavinking: Looks like it gets overwritten, which is fine, IMO.

gavinking: I've improved the API of `ValueGenerator` to allow this sort of validation.

I now enforce that the tenant id be correct.

beikov: The improved API looks great, but since this value generator only runs on INSERT, one could still setup cross tenant associations by doing updates, so this is IMO at best a "limitation". Anyway, preventing to accidently set a wrong value is IMO the task of a developer by controlling the write access to the property through proper encapsulation and we shouldn't mess with that. What if a one tenant is allowed to create other (sub-)tenants? It also seems silly to disallow this since one could still do it through an update i.e. I just don't think this kind of validation at this point is useful.

hibernate-core/src/main/java/org/hibernate/cfg/annotations/PropertyBinder.java
beikov: Unfortunately, we would also need to change a few other APIs that currently expect a String, but I think this should be a separate issues and PR.

hibernate-core/src/test/java/org/hibernate/test/tenantid/TenantIdTest.java
beikov: FTR, moved to https://hibernate.atlassian.net/browse/HHH-14822

hibernate-core/src/main/java/org/hibernate/cfg/annotations/PropertyBinder.java
beikov: FTR, moved to https://hibernate.atlassian.net/browse/HHH-14822

hibernate-core/src/main/java/org/hibernate/tuple/TenantIdGeneration.java
gavinking: Well I was already planning on having `@TenantId` imply `optional=false` and `updateable=false`.

I've implemented that now.

I think it's wrong to call this a "limitation". At most it's a limitation associated with the whole concept of multitenancy based on tenant ids and partitioning. As soon as you have associations between rows with different tenant ids, our filters just aren't going to do the right thing.

> What if a one tenant is allowed to create other (sub-)tenants? 

Well then you would need a whole tree-based tenancy model, not a flat model based on uninterpreted ids and partitioning. And the filters you would need are going to be a *lot* more complicated.

Now, in principle I guess you could now implement something like that on top of the new `AttributeBinder` stuff I just introduced, but you're (quite rightly) on your own.

gavinking:
I've now generalized this a bit, by introducing a new `AttributeBinder` API and `@AttributeBinderType` annotation.

This opens up the possibility of writing your own `@TenantId`-like annotations that add filters, or otherwise mess with the model associated with a property (to a certain limited extent).

This is obviously a sharp knife, and users shouldn't run with it.

gavinking:
> I've now generalized this a bit, by introducing a new `AttributeBinder` API and `@AttributeBinderType` annotation.

As a further demonstration of how this can be used, I reimplemented the `@AttributeAccessor` annotation as an `AttributeBinder`.

gavinking:
I've introduces a notion of a "root" tenant, a sort of superuser, in order to provide a reasonable way to turn this functionality off.

hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java
```diff
@@ -241,6 +243,22 @@ public SessionImpl(SessionFactoryImpl factory, SessionCreationOptions options) {
...
+			}
+			else {
+				CurrentTenantIdentifierResolver resolver = factory.getCurrentTenantIdentifierResolver();
+				if ( resolver==null || !resolver.isRoot(tenantIdentifier) ) {
+					// turn on the filter, unless this is the "root" tenant with access to all partitions
+					getLoadQueryInfluencers()
+							.enableFilter( TenantIdBinder.FILTER_NAME )
+							.setParameter( TenantIdBinder.PARAMETER_NAME, tenantIdentifier );
+				}
+			}
```
beikov, line 259: What do you think about the following instead to support treating `null` as possible root?

```suggestion
			CurrentTenantIdentifierResolver resolver = factory.getCurrentTenantIdentifierResolver();
			String tenantIdentifier = getTenantIdentifier();
			if ( tenantIdentifier == null && ( resolver == null || !resolver.isRoot(tenantIdentifier) ) ) {
				throw new HibernateException( "SessionFactory configured for multi-tenancy, but no tenant identifier specified" );
			}
			else {
				if ( resolver==null || !resolver.isRoot(tenantIdentifier) ) {
					// turn on the filter, unless this is the "root" tenant with access to all partitions
					getLoadQueryInfluencers()
							.enableFilter( TenantIdBinder.FILTER_NAME )
							.setParameter( TenantIdBinder.PARAMETER_NAME, tenantIdentifier );
				}
			}
```

beikov (commented):
Looks great :)

PR #3869, opened 2021-03-24
HHH-12752 Test and fix for secondary select for unique key association due to missing EntityUniqueKey registration

beikov:
https://hibernate.atlassian.net/browse/HHH-12752

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
@@ -75,6 +75,7 @@ import org.hibernate.internal.util.StringHelper;
 import org.hibernate.internal.util.collections.CollectionHelper;
 import org.hibernate.loader.spi.AfterLoadAction;
 import org.hibernate.persister.collection.CollectionPersister;
+import org.hibernate.persister.entity.AbstractEntityPersister;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.persister.entity.Loadable;
 import org.hibernate.persister.entity.UniqueKeyLoadable;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/loader/Loader.java
```diff
@@ -1896,6 +1897,25 @@ private void loadFromResultSet(
 				session.getPersistenceContextInternal().addEntity( euk, object );
 			}
 		}
```
gbadner, line 1899: There is a block above what is displayed here, lines 1877-1899, that deals with the `AssociationType`s returned by `#getOwnerAssociationTypes()`.

It appears that in the case where an entity association is join fetched, and its `AssociationType#getRHSUniqueKeyPropertyName != null`, an entry will be added in both the above and below block.

I think that what has been added below was intended to be accomplished by the block above. The problem seems to be that `ownerAssociationTypes[i]` will be `null` unless the association is join fetched.

hibernate-core/src/main/java/org/hibernate/loader/Loader.java
```diff
@@ -1896,6 +1897,25 @@ private void loadFromResultSet(
...
+		boolean[] propertyUniqueness = ( (AbstractEntityPersister) persister ).getPropertyUniqueness();
+		for ( int index = 0; index < propertyUniqueness.length; index++ ) {
+			if ( propertyUniqueness[index] ) {
+				final Type type = persister.getPropertyTypes()[index];
+				final Object uniqueKey = type.semiResolve( values[index], session, object );
```
gbadner, line 1904: One thing to consider is that in the case of a property-ref that contains a many-to-one, that many-to-one may get initialized, because [`ComponentType#semiResolve`](https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/type/ComponentType.java#L709-L714) will delegate to `#resolve`.

I'm not sure if this is a real issue. Please add test case that has a property-ref with a lazy many-to-one association, so we can see what happens.

hibernate-core/src/main/java/org/hibernate/loader/Loader.java
gbadner: FWIW, there were no test failures commenting out the block above, and it did not change the SQL that got generated by hibernate-core tests.

@Steve, do you see a reason to keep the block above?

PR #3870, opened 2021-03-24
move several new types out of the root package

gavinking:
The namespace `org.hibernate` is already much too cluttered and we should try hard to not make it worse in H6.

Furthermore, some of these things look to be of more interest to developers of Hibernate than to users.

Now, I'm not sure that `org.hibernate.query` is absolutely the best place either, but it's at least better.


sebersole (approved):
The only real regret I have there was moving the main `Query` interface to `o.h.query`.  Anyway, +1 from me

As for a "better" package name... I mean it all relates to queries, so not sure what could fit better ;)  Unless we want to go all out and move `org.hibernate.{x}...` to `org.hibernate.orm.{x}...` :grin:

beikov (approved):

gavinking:
@sebersole 

> The only real regret I have there was moving the main `Query` interface to `o.h.query`.

I actually dislike that too FTR.

It wouldn't be a breaking change to move it *back* y'know.

gavinking:
> It wouldn't be a breaking change to move it back y'know.

Let me clarify: I wouldn't literally move it _back_. But I do think it's worth investigating the possibility of adding a new cleaned-up supertype of `o.h.q.Query` to `o.h` after removing some legacy baggage.

gavinking:
So this and my other PRs are failing the test suite now because of unrelated failures in `PackagedEntityManagerTest`.

beikov:
>So this and my other PRs are failing the test suite now because of unrelated failures in PackagedEntityManagerTest.

Sorry about that, I must have missed something. I pushed another fix now to finally resolve this.

PR #3880, opened 2021-03-28
HHH-14537 : EntityNotFoundException thrown when non-existing association with @NotFound(IGNORE) mapped has proxy in PersistenceContext

gbadner:
https://hibernate.atlassian.net/browse/HHH-14537

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
@@ -362,10 +362,17 @@ public class DefaultLoadEventListener implements LoadEventListener {
 		if ( !options.isAllowProxyCreation() ) {
 			impl = load( event, persister, keyToLoad, options );
 			if ( impl == null ) {
-				event.getSession()
-						.getFactory()
-						.getEntityNotFoundDelegate()
-						.handleEntityNotFound( persister.getEntityName(), keyToLoad.getIdentifier() );
+				if ( options == LoadEventListener.INTERNAL_LOAD_NULLABLE ) {
+					// The proxy is for a non-existing association mapped as @NotFound.
+					// Don't throw an exeption; just return null.
+					return null;
+				}
+				else {
+					event.getSession()
+							.getFactory()
+							.getEntityNotFoundDelegate()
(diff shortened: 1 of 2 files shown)


sebersole (changes requested):
I think the change is reasonable.  

I would prefer to see a "better" test though - the whole double mapping of the FK obfuscates what is being tested.  Just use the (non-existent) Employee proxy as the association when saving the Task.

gbadner:
> I would prefer to see a "better" test though - the whole double mapping of the FK obfuscates what is being tested. Just use the (non-existent) Employee proxy as the association when saving the Task.

@sebersole, I'm not sure what you mean about "the whole double mapping of the FK".

I added `Employee#location` (without `@NotFound(IGNORE)` and test, `#testExistingProxyWithNonExistingAssociation`, to be able to test that when `Task#employee` exists, but `Task#employee#location` FK refers to a non-existing entity, that `EntityNotFoundException` is properly thrown.

Did you mean about making `Task#employee` non-insertable and non-updatable? I thought that was an interesting way to still have access to the ID for the non-existing entity.

sebersole:
```

					// just use proxy to the non-existing Employee#2 to create Task#2
					final Employee emp2 = session.load( Employee.class, 2 );
					final Task task2 = new Task( 2, emp2 );
					session.persist( task2 );
```

gbadner:
@sebersole, I just pushed a commit that removes {{Task#employeeId}} and just assigns the proxy to {{Task#employee}} as you suggested.

I also added a couple of tests that show that lazy associations with the non-existing entity are assigned a proxy, and eager associations with the same non-existing entity mapped {{@NotFound(IGNORE)}} is set to {{null}}.

@sebersole, please confirm if this is the outcome you were expecting.

sebersole (approved):
Looks good to me +1

sebersole:
Do you know the reason for the failure on the Oracle job?

beikov:
>Do you know the reason for the failure on the Oracle job?

Like I was writing in the Zulip topic about GH actions and Oracle EE, this is due to the use of Oracle XE. I will try to deactivate this test for Oracle XE tomorrow and see how I can improve the CI situation as a whole.

gbadner:
Thanks for the review @sebersole!

PR #3881, opened 2021-03-29
Split GH Actions workflow into individual parts, document docker_db.sh and use Oracle EE for CI

beikov:
It's not pretty, but it should work. I split the workflow into individual parts so that we can download reports immediately after a job fails. Also, I switched to Oracle EE for testing so this should finally make the Oracle tests succeed.


beikov:
This does not work as expected. Even with Oracle EE the desired tests don't work as expected. HANA requires at least 8GB RAM which we don't have on GH Actions.

PR #3882, opened 2021-03-30
HHH-14537 : ENFE when non-existing association with @NotFound(IGNORE)  has proxy in PC (5.3)

gbadner:
Do not merge. This PR follows a different process.

https://hibernate.atlassian.net/browse/HHH-14537


sebersole:
Marking as draft is a good protection against it getting merged

gbadner:
@sebersole, I need to remove the "Draft" setting. It is confusing the EAP process that I'm following.

PR #3883, opened 2021-03-31
HHH-14523 Fix for dirty-checking bug

maesenka:


beikov (approved):
Seems ok to me although I don't know if the tests against PostGIS are passing.

PR #3884, opened 2021-04-01
HHH-14529 + HHH-14530 XML mapping improvements to allow for compile-time parsing in Quarkus

yrodiere:
* [HHH-14530](https://hibernate.atlassian.net/browse/HHH-14530): Adding pre-parsed Binding<?> instance to MetadataSources
* [HHH-14529](https://hibernate.atlassian.net/browse/HHH-14529): Move from dom4j to JAXB for orm.xml parsing

So... the first commit is rather straightforward, I think.

The next ones... well, they are massive. Essentially, we need to move from DOM4J to Jaxb for `orm.xml` parsing if we want to make it work in Quarkus, because:

1. dom4j cannot work in native mode
2. Quarkus ideally should parse the XML mapping at compile time, and serialize it somehow (using something fast) to reuse it during static init. The representation of the XML mapping in DOM4J cannot easily be serialized in anything else than XML.

JAXB solves all of that, in particular because it allows representing the XML mapping as POJOs, which can easily be serialized by Quarkus' `BytecodeRecorder`.

Because this is a lot of changes, and thus a lot of risk, I could not bring myself to "just change it". I introduced an alternative JAXB implementation alongside the existing one based on dom4j, and added options to select this alternative implementation. In ORM 5.4.x, we will disable the JAXB implementation by default; in Quarkus, we will enable it. In ORM 5.5, I suppose I could add a commit to enable JAXB by default, or even to remove the DOM4J implementation completely.


yrodiere:
Updated the description to explain what's going on.

@Sanne : I will need this for https://github.com/quarkusio/quarkus/issues/14762 . With these changes I'm able to get a very simple XML mapping to work in Quarkus. I still have to test more complex mappings and native mode, but this is a start.

sebersole:
> * [HHH-14529](https://hibernate.atlassian.net/browse/HHH-14529): Move from dom4j to JAXB for orm.xml parsing

Awesome!


> Because this is a lot of changes, and thus a lot of risk, I could not bring myself to "just change it". I introduced an alternative JAXB implementation alongside the existing one based on dom4j, and added options to select this alternative implementation. In ORM 5.4.x, we will disable the JAXB implementation by default; in Quarkus, we will enable it. In ORM 5.5, I suppose I could add a commit to enable JAXB by default, or even to remove the DOM4J implementation completely.

To me, the choice between dom4j and jaxb here is an implementation detail.  Meaning, I would not be concerned with making this switchable.  We did this for quite a while with Javassist/ByteBuddy but that situation is different because that change does (can) affect user code.

yrodiere:
> To me, the choice between dom4j and jaxb here is an implementation detail

Oh totally. I did not make this switchable so that people can enjoy some dom4j goodness :).
My concern was more to prevent regressions: considering the amount of changes, there is a significant risk of regressions.

I personally don't think we should introduce dangerous changes like this in micros, and yet we need to, for Quarkus. Hence the switch: we'll introduce the dangerous change in Quarkus (where it's not dangerous, since XML mapping is just now being introduced), while leaving other users unaffected.

sebersole:
No problem.  We'll just drop that part in 6.

gavinking:
@yrodiere could you make it just automatically use JAXB if there is no dom4j in the classpath?

yrodiere:
> @yrodiere could you make it just automatically use JAXB if there is no dom4j in the classpath?

Not really, since:

1. I believe we do have dom4j in the classpath in Quarkus, and I'm not sure Hibernate ORM would still work if we removed it (it's used elsewhere)
2. Users may add dom4j to the classpath for some reason, and I wouldn't want the Quarkus Hibernate ORM extension to start blowing up if they do (because it will blow up if we don't use JAXB for XML parsing).

It's just simpler to always enable this alternative implementation in Quarkus.

yrodiere:
They're out of luck, as we didn't make this possible yet. One day we will ([HHH-13703](https://hibernate.atlassian.net/browse/HHH-13703)), but this PR is enough to get there.

yrodiere:
> They're out of luck, as we didn't make this possible yet. One day we will ([HHH-13703](https://hibernate.atlassian.net/browse/HHH-13703)), but this PR is enough to get there.

I meant **not** enough :)

gavinking:
OK.

sebersole:
At the moment that is really just an issue (granted a painful one) with Envers

Sanne:
> Well I was more thinking of this the other way round: what about a plain Hibernate user who doesn't want the dom4j dependency?

We'll do that for Quarkus users as in that case this is about a new feature (XML mapping support is missing there), and because we can automatically re-introduce dom4j on the classpath for the Envers users.

Other users will benefit from it in 6 :)

Sanne:
@yrodiere many thanks, looks very good  - however since it's large I have only read it once.

@sebersole WDYT, can we merge? Seems there's a good chunk written by you and Strong

I propose we merge it in `main`, and experiment a bit more with it by implementing the Quarkus side of the patches working with Hibernate ORM 5.5.

We can then decide if we want to backport this to ORM 5.4 and keep Quarkus on that version, or perhaps to really release a 5.5 build and upgrade.

Quarkus is now in a slower-than-usual window: it will release in double the usual time as Quarkus 2.0 is being prepared - so it could be a good time to apply this and other more aggressive changes on ORM `main` and then bump the minor version.

yrodiere:
@Sanne the backport PR is already done, so it wouldn't cost us much.

@sebersole ... is this good to merge as far as you are concerned?

Sanne:
@yrodiere ok I have no objections to backport.

But I wasn't thinking in terms of cost, more in terms of risk management - there's a series of "larger than usual" patches we want in the 5.x branch, so perhaps it's a good time to call it 5.5 - but we can also backport this one specifically to 5.4.

yrodiere:
Ah well, if the goal is to finally start releasing minors instead of micros, I'm totally in favor :)

yrodiere:
Tests pass, no objection: I merged it. Thanks everyone!
I'll wait a bit more for the backport (#3885), as we may end up using 5.5 in Quarkus in the end. We'll see.

sebersole (commented):
Little late to the party, but +1

PR #3887, opened 2021-04-01
[6.0] Table group rendering support

beikov:
Implement table group rendering for real table groups and get rid of canInnerJoin flags due to that


sebersole (approved):
Looks good to me

PR #3896, opened 2021-04-04
HHH-14546 - Test case for bytecode-enhanced byte[] entity getter unexpectedly loading other non-requested byte[] fields

kawoolutions:
https://hibernate.atlassian.net/browse/HHH-14546


kawoolutions:
Forgot to tag as Jira issue test case - sry

dreab8:
Hi @kawoolutions ,

actually this is the expected behaviour for all singular attributes as stated [here in the doc](https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#BytecodeEnhancement-lazy-loading)


> By default, all singular attributes are part of a single group, meaning that when one lazy singular attribute is accessed all lazy singular attributes are loaded

in order to obtain what you want you need to use `@LazyGroup` annotation.

PR #3898, opened 2021-04-06
HHH-14543: fix gap error when there is an empty binding

JintaoXIAO:
in below case, if the binding variable is (1, [2,3]) (2, [])
```
WHERE c.id IN ?1 AND (n IS EMPTY or n.id IN ?2)
```
then the generated sql will be 
```
WHERE c.id IN (?1, ?3) AND (n IS EMPTY or n.id IN ())
```
if we don't reuse original ordinal parameter, we just get,
which I think is an easy solution
```
WHERE c.id IN (?3, ?4) AND (n IS EMPTY or n.id IN ())
```
(sorry for my poor English)


JintaoXIAO:
maybe need further investigation, just close it

PR #3913, opened 2021-04-10
HHH-14553: test case for byte-enhanced runner persisting Long ID entity failing with IAE

kawoolutions:
Test case for https://hibernate.atlassian.net/browse/HHH-14553

We're currently stuck on a big project using a Java EE 7 environment and we won't be able to switch to Java EE 8 anytime soon. It would be very nice to get this fixed, also given this doesn't happen on the current Hibernate main branch, the fix must somehow be present in 5.3 and/or 5.4.


kawoolutions:
It's OK, we found out that the Hibernate team forgot to publish 5.2.18 to the main website, which included a good deal of fixes that worked for us. We will upgrade to 5.3+ and Java EE 8 or Jakarta within the next two months.

Thanks for the time.

PR #3915, opened 2021-04-12
HHH-11490 Fix for UUID mapping conflict

maesenka:


maesenka:
I tried to use an arbitrary int value, but this throw an exception when saving a null Geometry value. The JDBC driver for Postgresql checks the type code and throws an error on codes it doesn't know about. 

My solution here still feels like a hack, though. With more database types all mapped by the driver to `Types.OTHER` I feel the assumption that there is exactly one database type corresponding to a JDBC type code is getting problematic.

beikov:
You should probably use some arbitrary value instead and instead create a custom Binder instead which uses the JDBC type code `OTHER` for binding null. Binding null for arrays and other complex types in Oracle actually also requires the use of the type name, so this will probably be easier in 6.0, but for 5.x you will have to create a custom `Binder`

PR #3916, opened 2021-04-13
HHH-14549 Collection with default field initializer will always be empty

dreab8:
https://hibernate.atlassian.net/browse/HHH-14549


beikov (approved):
LGTM

sebersole (approved):
Looks reasonable to me :)

PR #3917, opened 2021-04-13
HHH-14557 Connection leaked on rollback with mode DELAYED_ACQUISITION_AND_RELEASE_BEFORE_TRANSACTION_COMPLETION

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14557

Turns out we don't have a "before transaction completion" event in the case of rollbacks, so we need to rely on the "after transaction completion" event to do some more cleanup, even when using `DELAYED_ACQUISITION_AND_RELEASE_BEFORE_TRANSACTION_COMPLETION`.

Related: https://issues.redhat.com/browse/AG-168

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/JdbcCoordinatorImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/JdbcCoordinatorImpl.java
@@ -270,7 +270,8 @@ public class JdbcCoordinatorImpl implements JdbcCoordinator {
 	public void afterTransaction() {
 		transactionTimeOutInstant = -1;
 		if ( getConnectionReleaseMode() == ConnectionReleaseMode.AFTER_STATEMENT ||
-				getConnectionReleaseMode() == ConnectionReleaseMode.AFTER_TRANSACTION ) {
+				getConnectionReleaseMode() == ConnectionReleaseMode.AFTER_TRANSACTION ||
+				getConnectionReleaseMode() == ConnectionReleaseMode.BEFORE_TRANSACTION_COMPLETION ) {
 			this.logicalConnection.afterTransaction();
 		}
 	}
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/JdbcCoordinatorImpl.java
```diff
@@ -269,10 +269,7 @@ public void afterStatementExecution() {
 	@Override
 	public void afterTransaction() {
 		transactionTimeOutInstant = -1;
-		if ( getConnectionReleaseMode() == ConnectionReleaseMode.AFTER_STATEMENT ||
-				getConnectionReleaseMode() == ConnectionReleaseMode.AFTER_TRANSACTION ) {
-			this.logicalConnection.afterTransaction();
-		}
+		this.logicalConnection.afterTransaction();
```
dreab8, line 272: maybe I'm wrong but in this way I think we are releasing the connection after a transaction even when `ConnectionReleaseMode.ON_CLOSE`, probably we should just add `|| getConnectionReleaseMode() == ConnectionReleaseMode.BEFORE_TRANSACTION_COMPLETION`

yrodiere: There's already a check in `this.logicalConnection.afterTransaction();`:

```
		if ( connectionHandlingMode.getReleaseMode() != ConnectionReleaseMode.ON_CLOSE ) {
			// NOTE : we check for !ON_CLOSE here (rather than AFTER_TRANSACTION) to also catch:
			// - AFTER_STATEMENT cases that were circumvented due to held resources
			// - BEFORE_TRANSACTION_COMPLETION cases that were circumvented because a rollback occurred
			//   (we don't get a beforeTransactionCompletion event on rollback).
			log.debug( "Initiating JDBC connection release from afterTransaction" );
			releaseConnection();
		}
```

yrodiere: Ha but you're right, there's a call to super() which release resources (not the connection)... I'll change this.

yrodiere: Fixed, thanks.

yrodiere:
Merged, thanks @dreab8!

PR #3922, opened 2021-04-15
Correction to the QuerySqmImpl constructor used for criteria queries

jrenaat:


sebersole:
Would be better to have a test here that shows the failure before the fix, and passing after the fix.  Could you pull one over here?

PR #3923, opened 2021-04-16
6.0 Cleanup case when handling and implement parameter handling for case result arms

beikov:


hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -2496,6 +2496,31 @@ protected void renderSelectExpression(Expression expression) {
...
+		}
+		else if ( expression instanceof CaseSimpleExpression ) {
+			visitCaseSimpleExpression( (CaseSimpleExpression) expression, true );
+		}
+		else if ( expression instanceof CaseSearchedExpression ) {
+			visitCaseSearchedExpression( (CaseSearchedExpression) expression, true );
+		}
+		else {
+			expression.accept( this );
+		}
```
gavinking, line 2521: would it make sense to get rid of this switch here and dispatch this to a polymorphic method?

beikov: Would you mind elaborating what you mean exactly? This logic really is specific for databases like Derby, DB2, Firebird and HSQL and I wanted these code paths really only for the respective translators to avoid the general overhead. AFAIU you are proposing to add something like `acceptInSelect` or an overload to `Expression` for this purpose and respective `visitXXX` methods (or change just the respective visit methods to accept `boolean inSelect`) in the walker for the specific case of "being in select", correct? I don't have a strong preference here, but it feels like this change would clutter the API.

Please clarify what you mean.

sebersole: I may be wrong, but I think he was just asking about general design - preferring polymorphism over functional.  I think here it is fine.

PR #3924, opened 2021-04-16
Better union subclass support and type discriminator fixes

beikov:
`DiscriminatorType` is now a `BasicType` which contains it's underlying `BasicType` for our internal representation.
To the user, we only ever expose the discriminator as a `Class`, but the internal type is either a number or a string.
Drop `SqmPathEntityType` in favor of `SqmBasicValuedEntityTypePath` which now refers to a `SqmLiteralEntityType`.
Support `TYPE` also as select item by implement `DomainResultProducer`.
Cleanup path type expression code to work the same way as it does for other attributes.

Introduce `originalPropertyOrdering` for `ComponentType` to capture the reordering after sorting.
This is important for foreign key mappings that come from XML which depend on the order.

Resolve UnionTableReference in UnionTableGroup for all subqueries and table expressions within the hierarchy.
Resolve special table references referring to the actual sub-class tables for multi-table DML.
Fix transitive sub-type linking in persisters.


sebersole (approved):

dreab8:
LGTM :+1:

PR #3930, opened 2021-04-19
HHH-14567 Filters are ignored if enabled after query creation but before execution

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14567

No need to backport, as the optimization that causes the bug was not backported.


yrodiere:
Reviewed by @dreab8 ; looks good: https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/topic/Hibernate.20ORM.205.2E5/near/235196525

PR #3934, opened 2021-04-19
Fix wrong creation of new instances for non aggregated id with no IdClass

dreab8:


beikov (approved):
Looks good AFAICT.

dreab8:
Merged, Thanks @beikov !

PR #3935, opened 2021-04-20
6.0 Infer parameter bind type for null parameter values instead of using a default

beikov:


hibernate-core/src/main/java/org/hibernate/query/spi/AbstractQuery.java
```diff
@@ -1314,9 +1314,6 @@ else if ( retType.isArray() ) {
 			type = getParameterMetadata().getQueryParameter( namedParam ).getHibernateType();
 		}
 		if ( type == null ) {
-			if ( retType == null ) {
-				return JavaObjectType.INSTANCE;
-			}
 			type = getSession().getFactory().resolveParameterBindType( retType );
```
dreab8, line 1317: i think here should be
```
if(retType != null) {
				type = getSession().getFactory().resolveParameterBindType( retType );
			}
```

sebersole: Yes, this condition was leading to NPE when `retType` was null which is why I added this that JavaObjectType bit in the first place

gavinking: With this sort of code we need to be very careful that we don't blow up with NPEs in cases where the programmer typed something like `where p.firsName = :name` and so Hibernate couldn't assign a type to the expression `p.firsName`.

Do we have tests for that kind of thing?

sebersole: This is why I added `org.hibernate.query.sqm.sql.internal.TypeHelper#highestPrecedence`. 

Not in the particular case you mention.  I do not believe a parameter will ever imply the type of the "other side".

I cannot think of any practical example where what you suggest might happen.

gavinking: OK; my concern is more about making sure that in general the type inference stuff doesn't cause NPEs even for broken queries.

sebersole: Do you have an example of such a "broken query"?

sebersole: Not an inference thing specifically, but I noticed that queries like this "work":

```
from EntityOfBasics e where e.gender = e.ordinalGender
```

where `gender` is mapped to a CHAR and `ordinalGender` is mapped to a TINYINT.  That does not cause problems on H2 and in this specific case, but I do worry about that as a general issue.

gavinking: Well I mean any case where we had a path expression to which we weren't able to assign a type, because one of the path elements didn't resolve to anything.

(I noticed one problem of this general form a long time ago, but it's probably been fixed by now.)

PR #3938, opened 2021-04-21
Backporting  HHH-13077 and HHH-14567 to 5.4

DavideD:
Backporting these changes from 5.5 will make it possible for Hibernate Reactive to work with 5.4 and 5.5.
Is it possible to apply these changes?

Backports: [HHH-13077](https://hibernate.atlassian.net/browse/HHH-13077) and [HHH-14567](https://hibernate.atlassian.net/browse/HHH-14567).

For Hibernate Reactive only HHH-13077 is required but the other commits fix a regression in Quarkus.


Sanne:
merged! Thanks

DavideD:
Thank you!

PR #3939, opened 2021-04-22
Fix EmbeddedForeignKeyDescriptor creation when o.h.mapping.Value has …

dreab8:
…no columns, Re-enabled additional tests


beikov (approved):
LGTM

PR #3941, opened 2021-04-22
HHH-14571 Hibernate.isPropertyInitialized( someEntity, "id" ) returns false

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14571


yrodiere:
Discussed on Zulip with Andrea; conclusion: it should be safe enough. Merging!

PR #3944, opened 2021-04-23
HHH-14574 Avoid NPEs from LoadContexts.cleanup(ResultSet)

dreis2211:
See https://hibernate.atlassian.net/browse/HHH-14574


educhastenier (approved):
Looks good to me :+1: even if there is no unit test (Just kidding 😃  there would be no value of such a test)

educhastenier:
Why not fix it on a previous maintenance branch?

dreis2211:
I'd appreciate this, @educhastenier - but I'd leave an eventual backport decision to the maintainers. It should apply cleanly.

dreab8:
Thanks @dreis2211 for the PR

PR #3949, opened 2021-04-29
HHH-14513 Move publishing release artifacts from BinTray

dreab8:


Sanne (commented):
Looks good at high level.. @DavideD could you have a look as well based on your experience on HR?

DavideD:
Looks alright. Before doing a complete release, I would suggest to try to create and close the repository without releasing it.
This way we can check on the stage repository that everything is alright.

We can then update the release process to do a complete release.

dreab8:
thnaks @Sanne  and @DavideD  merged upstream

PR #3951, opened 2021-04-29
Support for type coercion for values passed as ids and as query param…

sebersole:
Support for type coercion for values passed as ids and as query parameter bindings

- widening coercions
- valid (no over/under flow) narrowing coercions

NOTE : this leads to (1) test failure which explicitly tests for load-by-id type mismatches to fail.  JPA requires this.  Either we should drop load-by-id as part of this work, or cover this under `JpaCompliance`.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/BigDecimalTypeDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/BigDecimalTypeDescriptor.java
@@ -8,6 +8,7 @@ package org.hibernate.type.descriptor.java;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.util.Locale;
 
 import org.hibernate.dialect.Dialect;
 import org.hibernate.type.descriptor.WrapperOptions;
@@ -78,13 +79,13 @@ public class BigDecimalTypeDescriptor extends AbstractClassTypeDescriptor<BigDec
 		if ( value == null ) {
 			return null;
 		}
-		if ( BigDecimal.class.isInstance( value ) ) {
+		if ( value instanceof BigDecimal ) {
 			return (BigDecimal) value;
 		}
-		if ( BigInteger.class.isInstance( value ) ) {
+		if ( value instanceof BigInteger ) {
 			return new BigDecimal( (BigInteger) value );
 		}
-		if ( Number.class.isInstance( value ) ) {
+		if ( value instanceof Number ) {
 			return BigDecimal.valueOf( ( (Number) value ).doubleValue() );
 		}
 		throw unknownWrap( value.getClass() );
@@ -98,5 +99,31 @@ public class BigDecimalTypeDescriptor extends AbstractClassTypeDescriptor<BigDec
 	@Override
 	public int getDefaultSqlPrecision(Dialect dialect) {
 		return dialect.getDefaultDecimalPrecision();
+	}
+
+	@Override
+	public <X> BigDecimal coerce(X value, CoercionContext coercionContext) {
+		if ( value == null ) {
+			return null;
+		}
+
+		if ( value instanceof Number ) {
+			return BigDecimal.valueOf( ( (Number) value ).doubleValue() );
+		}
+
+		if ( value instanceof String ) {
+			return CoercionHelper.coerceWrappingError(
+					() -> BigDecimal.valueOf( Double.parseDouble( (String) value ) )
+			);
+		}
+
+		throw new CoercionException(
+				String.format(
+						Locale.ROOT,
+						"Unable to coerce value [%s (%s)] to BigDecimal",
+						value,
+						value.getClass().getName()
+				)
+		);
 	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ByteTypeDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ByteTypeDescriptor.java
@@ -5,6 +5,10 @@
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
 package org.hibernate.type.descriptor.java;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.Locale;
+
 import org.hibernate.dialect.Dialect;
 import org.hibernate.type.descriptor.WrapperOptions;
 import org.hibernate.type.descriptor.java.spi.Primitive;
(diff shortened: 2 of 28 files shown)


gavinking:
> cover this under `JpaCompliance`.

This option gets my vote.

beikov (commented):
I'm not a big fan of this, as we would have to consider coercion for every parameter that is ever set, which feels wrong to me. IMO people should just pass the parameter of the expected type and enjoy the fast code paths that is enables. Like I wrote in the discussion (https://github.com/hibernate/hibernate-orm/discussions/3937#discussioncomment-639501), this was not supported in 5, so this is a new feature. I am not totally against it, but -0 from my side as I don't think this is worth the overhead. The performance could be improved by making use of `ClassValue`, but that is something we should consider for all if-instanceof chains in a dedicated PR.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/BigDecimalTypeDescriptor.java
```diff
@@ -99,4 +100,30 @@ public long getDefaultSqlLength(Dialect dialect) {
 	public int getDefaultSqlPrecision(Dialect dialect) {
 		return dialect.getDefaultDecimalPrecision();
 	}
+
+	@Override
+	public <X> BigDecimal coerce(X value, CoercionContext coercionContext) {
+		if ( value == null ) {
+			return null;
+		}
+
```
dreab8, line 109: i would add
```
if ( value instanceof BigDecimal ) {
			return (BigDecimal) value;
		}
```

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ByteTypeDescriptor.java
```diff
@@ -102,4 +106,58 @@ public int getDefaultSqlPrecision(Dialect dialect) {
...
 	}
+
+	@Override
+	public <X> Byte coerce(X value, CoercionContext coercionContext) {
+		if ( value == null ) {
+			return null;
+		}
+
+		if ( value instanceof Byte ) {
+			return (byte) value;
```
dreab8, line 117: not sure but probably it is better to cast the value to Byte

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ByteTypeDescriptor.java
```diff
@@ -102,4 +106,58 @@ public int getDefaultSqlPrecision(Dialect dialect) {
...
+		}
+
+		if ( value instanceof Short ) {
+			return CoercionHelper.toByte( (short) value );
```
dreab8, line 121: here to cast to `Short`

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DoubleTypeDescriptor.java
```diff
@@ -121,4 +122,58 @@ public int getDefaultSqlPrecision(Dialect dialect) {
...
+
+		if ( value instanceof Integer ) {
+			return ( (Integer) value ).doubleValue();
+		}
+
+		if ( value instanceof Long ) {
+			return ( (Long) value ).doubleValue();
+		}
+
+		if ( value instanceof Float ) {
```
dreab8, line 151: here probably better to cast the valut to `Float`

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/IntegerTypeDescriptor.java
```diff
@@ -110,4 +111,58 @@ public int getDefaultSqlPrecision(Dialect dialect) {
...
+		if ( value == null ) {
+			return null;
+		}
+
+		if ( value instanceof Integer ) {
+			return (int) value;
+		}
+
+		if ( value instanceof Short ) {
+			return CoercionHelper.toInteger( (short) value );
```
dreab8, line 126: same here and the following method, probably better not cast to primitives

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/IntegerTypeDescriptor.java
```diff
@@ -110,4 +111,58 @@ public int getDefaultSqlPrecision(Dialect dialect) {
...
 	}
+
+	@Override
+	public Integer coerce(Object value, CoercionContext coercionContext) {
+		if ( value == null ) {
+			return null;
+		}
+
+		if ( value instanceof Integer ) {
+			return (int) value;
```
dreab8, line 122: here cast to `Integer`

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ByteTypeDescriptor.java
sebersole: It already checks for null previously

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/ByteTypeDescriptor.java
sebersole: Like above, already checked for null

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/IntegerTypeDescriptor.java
sebersole: Like above, already checked for null

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/IntegerTypeDescriptor.java
sebersole: Like above, already checked for null

sebersole:
> I'm not a big fan of this, as we would have to consider coercion for every parameter that is ever set, which feels wrong to me. IMO people should just pass the parameter of the expected type and enjoy the fast code paths that is enables. Like I wrote in the discussion ([#3937 (comment)](https://github.com/hibernate/hibernate-orm/discussions/3937#discussioncomment-639501)), this was not supported in 5, so this is a new feature. I am not totally against it, but -0 from my side as I don't think this is worth the overhead. The performance could be improved by making use of `ClassValue`, but that is something we should consider for all if-instanceof chains in a dedicated PR.

But this *is* supported in 5.  We have tests failing because this used to work.

sebersole:
Applied upstream.  Fixes legacy tests expecting such coercions for parameter binds.  We can talk about removing that capability if we want to as part of 6.x, but let's do that as a follow-up

sebersole:
Grr, so some of these primitive casts do fail on the CI machine but not locally.  Any ideas?

beikov:
I have same problem when I run `./gradlew compile` locally. Seems like IntelliJ uses some kind of different compiler internally that allow this kind of cast.

dreab8:
it fails when using java8

sebersole:
> it fails when using java8

Ahh... I use 11 locally

sebersole:
> I have same problem when I run `./gradlew compile` locally. Seems like IntelliJ uses some kind of different compiler internally that allow this kind of cast.

Nothing to do with IJ.  Andrea is on the right track most likely

PR #3952, opened 2021-04-30
HHH-14541 SessionFactoryObserver#sessionFactoryClosing() gets called after the session factory is marked as closed

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14541


yrodiere:
@Sanne FYI I would like to backport this to 5.4, too. No problem with that?

PR #3953, opened 2021-05-01
Publish to Sonatype OSSRH

sebersole:


gradle/publishing-repos.gradle
```diff
@@ -4,29 +4,23 @@
...
+//publishing {
+//	if ( project.ormVersion.isSnapshot ) {
+//		repositories {
+//			maven {
+//				name 'snapshots'
+//				url 'https://oss.sonatype.org/content/repositories/snapshots/'
+//
+//				credentials {
+//					username project.credentials.hibernateSnapshotsUsername
+//					password project.credentials.hibernateSnapshotsPassword
```
DavideD, line 21: Are there separate credentials for the Snapshots? Just curious.
The `sonatype()` repository from the plugin already includes this url, so you could configure it by reading the different credentials when we have to release a SNAPSHOT or not.

Mostly because I don't like having dead commented code, especially in the gradle build files.

DavideD (approved):
Looks good to me.

But do you still need `publishing-repos` though? I'm asking because the it contains only dead code that doesn't seem necessary. `nexusPublishing` configure the sonatype repository for snapshots and releases, if we need to release with different credentials, we could read the right credentials when we check the username and password.

dreab8:
It looks good to me :+1:

gradle/publishing-repos.gradle
dreab8: I think we can remove this file

dreab8 (approved):

gradle/publishing-repos.gradle
sebersole: Yep, I commented it out to test the `nexusPublishing` block and then forgot to delete it.  It can be removed

sebersole:
Superseded by https://github.com/hibernate/hibernate-orm/pull/3975

PR #3954, opened 2021-05-03
Jrenaat jrenaat wip/6.0 jpatest migrations 5

dreab8:


hibernate-core/src/test/java/org/hibernate/orm/test/jpa/transaction/batch/JtaWithStatementsBatchTest.java
```diff
@@ -92,16 +92,18 @@ public void testUnableToReleaseStatementMessageIsNotLogged(EntityManagerFactoryS
 						assertStatementsListIsCleared();
 						assertAllStatementsAreClosed( testBatch.createdStatements );
 					}
-					catch (Exception e) {
+					catch (Throwable t) {
```
jrenaat, line 95: I wonder if it wouldn't be better to catch both Exception + AssertionError, as you did in TransactionUtil.wrapInTransaction?

jrenaat (approved):
LGTM

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/transaction/batch/JtaWithStatementsBatchTest.java
dreab8: changed it, thanks @jrenaat

dreab8:
merged

PR #3956, opened 2021-05-03
HHH-12445 Discriminator column should be nullable when using @DiscriminatorValue( "null" )

yrodiere:
https://hibernate.atlassian.net/browse/HHH-12445


beikov (approved):
LGTM

yrodiere:
Merged, thanks.

PR #3962, opened 2021-05-05
Fix in one of the PersistenceXmlParser.parse() methods used in tests

jrenaat:


jrenaat:
Converted back to draft for now

PR #3965, opened 2021-05-06
Topic/main java library

sebersole:
Attempt to move from the `java` plugin to the `java-library` plugin.  Causes problems with Java 9 module (auto) naming, presumably due to `java-library` not using Jar from depended project within the same build.

DO NOT MERGE!!!

Intended for troubleshooting only


sebersole:
@jjohannes When you said to use Gradle's built in Java module support, you are talking about [this](https://docs.gradle.org/current/samples/sample_java_modules_multi_project.html)?

I'm still not understanding why we would need to move the "code" from `hibernate-integrationtest-java-modules/src/main` to `hibernate-integrationtest-java-modules/src/test`.  Do you have a pointer to that discussion?  If the page above is what you intended and it is in that page I'll find it so no worries.

jjohannes:
@sebersole testing modules is a bit specific and controversial. Ideally, the tests themselves would need to live in a module if you want environment close to a real one. Please have a look at this section: https://docs.gradle.org/current/userguide/java_testing.html#sec:java_testing_modular

sebersole:
@jjohannes About moving code within the `hibernate-integrationtest-java-modules`...

First I did the following:

1)
  a. dropped the 3rd party plugin and instead added `java.modularity.inferModulePath = true` as you suggested
  b. upgraded to JUit 4.13.2 as you suggested
2) moved to JUnit 5

I was just doing the changes in steps. 

After the first change (a & b), the build still failed though interestingly it went from a compile errors to test failures.

After the second change, the tests now pass.

Is that what you would expect?

sebersole:
Superseded by https://github.com/hibernate/hibernate-orm/pull/3975

PR #3967, opened 2021-05-07
Only apply --add-opens to compiler JVM options when using JDK9+

yrodiere:
Without this, I simply cannot run this locally:

```
./gradlew clean :hibernate-integrationtest-java-modules:test -Ptest.jdk.version=11 -Porg.gradle.java.installations.paths=$HOME/tools/java/jdk1.8,$HOME/tools/java/jdk11
```

It fails when the (JDK8) compiler encounters the `--add-opens` option, which is normal because that option isn't supported in JDK8:

```
> Task :hibernate-core:compileJava FAILED
Compiling with '/home/yrodiere/tools/java/jdk8u222-b10'
Unrecognized option: --add-opens
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

FAILURE: Build failed with an exception.
```

For some reason we don't get this failure on CI; I have no idea why. Regardless, it seems to me that there is a problem and it should be fixed.


yrodiere:
As far as I can tell the test failures are unrelated and happen in other PRs as well (e.g. https://github.com/hibernate/hibernate-orm/pull/3968#partial-pull-merging ), so I'm going to go ahead and merge.

PR #3972, opened 2021-05-07
HHH-14585 size() HQL function discards '@Where' clause

mrizzi:
Reproducer for issue https://hibernate.atlassian.net/browse/HHH-14585


mrizzi (commented):
Fixed in https://github.com/hibernate/hibernate-orm/pull/3977

PR #3974, opened 2021-05-08
wip/6.0 verify UUID

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14600


hibernate-core/src/test/java/org/hibernate/orm/test/id/uuid/interpretation/UUIDBasedIdInterpretationTest.java
```diff
@@ -90,6 +92,7 @@ public void testMySQL() {
 
 	@Test
 	@TestForIssue( jiraKey = "HHH-10564")
+	@RequiresDialect( PostgreSQL94Dialect.class )
```
beikov, line 95: In Hibernate 6 we don't use the version specific dialect sub-classes anymore but configure the dialect during construction instead, so please use:

```suggestion
	@RequiresDialect( PostgreSQLDialect.class )
```

hibernate-core/src/test/java/org/hibernate/orm/test/id/uuid/interpretation/UUIDBasedIdInterpretationTest.java
```diff
@@ -78,6 +79,7 @@ private void checkUuidTypeUsed(StandardServiceRegistry ssr, Class<? extends Type
 
 	@Test
 	@TestForIssue( jiraKey = "HHH-10564")
+	@RequiresDialect( MySQL5Dialect.class )
```
beikov, line 82: In Hibernate 6 we don't use the version specific dialect sub-classes anymore but configure the dialect during construction instead, so please use:

```suggestion
	@RequiresDialect( MySQLDialect.class )
```

beikov (changes requested):
Other than a few small issues, this looks good! Thanks

hibernate-core/src/test/java/org/hibernate/orm/test/id/uuid/sqlrep/sqlchar/UUIDCharTest.java
```diff
@@ -0,0 +1,110 @@
...
+			assertThat( node.parent, notNullValue() );
+		} );
+	}
+
+	@Entity(name = "Node")
+	static class Node {
+
+		@Id
+		@Column(name = "id", length = 36)
+		@GeneratedValue( generator = "uuid2" )
```
beikov, line 90: Shouldn't this be the following? Either this, or change the usage in the other test to `@GeneratedValue( generator = "uuid2" )` as well.

```suggestion
		@GeneratedValue(generator = "uuid")
		@GenericGenerator(name = "uuid", strategy = "uuid2")
```

beikov (commented):
Not sure why, but it seems `UUIDBasedIdInterpretationTest` is failing now.

hibernate-core/src/test/java/org/hibernate/orm/test/id/uuid/sqlrep/sqlchar/UUIDCharTest.java
NathanQingyangXu: I changed as you suggested

hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQLDialect.java
```diff
@@ -823,91 +833,7 @@ public void contributeTypes(TypeContributions typeContributions, ServiceRegistry
...
-				@Override
-				protected X doExtract(CallableStatement statement, int position, WrapperOptions wrapperOptions) throws SQLException {
-					return javaTypeDescriptor.wrap( statement.getObject( position ), wrapperOptions );
-				}
-
-				@Override
-				protected X doExtract(CallableStatement statement, String name, WrapperOptions wrapperOptions) throws SQLException {
-					return javaTypeDescriptor.wrap( statement.getObject( name ), wrapperOptions );
-				}
-			};
```
NathanQingyangXu, line 910: I removed the above content to the private static child class in PostgresUUIDType (i.e. PostgresUUIDJdbcTypeDescriptor). See the code changes in PostgresUUIDType below for details.

After we register PostgresUUIDType as below, we also need to register the magical jdbc sql type into Dialect as above.

hibernate-core/src/test/java/org/hibernate/orm/test/id/uuid/sqlrep/sqlchar/UUIDCharTest.java
NathanQingyangXu: By closer debugging, I found the original testing case is buggy. The gist of testing here is to ensure the uuid db column's db type is binary and (var)char. Previously we used `@Column(length = 16)` and `@Column(length=36)` to achieve the goal, but they both end up with 'binary' though with different lengths.
I changed the annotation to use explicit Hibernate type to be in line with the testing goals. The generator related annotations are not mandatory here and for that reason I deleted them.

NathanQingyangXu:
Now I finalized the PR and it is ready for a new review again. The following serious bugs were flushed out and fixed:

- previously we only register uuid jdbc descriptor (with jdbc type tweaking, i.e. using a special sql code) to Postgres Dialect without registering basic UUID type; I copied the new jdbc descriptor code to the old internal static counterpart class in PostgresUUIDType and register it instead;
- previously our new Postgres UUID jdbc descriptor override 'doBind' to ensure 'OTHERS' is passed to driver (rather than the new special sql code), but when we need to set a null value, we still send out our special code to driver. I refactored the Binder class to expose the 'setNull' overriding possibility so we can fix the loophole;
- some previous testing code has serious bug. Especially, we only use '@Column(length=16)` and `@Column(length=36)` to differentiate between binary and (var)char sql column types; but the reality is they both end up with binary types.

@gavinking I would appreciate if you can take a new look at the code changes in this PR, especially the ones solving the two issues mentioned above.

gavinking:
@NathanQingyangXu the changes you made to `BasicBinder` are essentially the same as a change I already proposed in #3850. Which means that this PR creates a conflict with mine. Do you think you could at least change the names from `doSetNull()` to `doBindNull()` as I did, which is more consistent with the naming of other methods in that class?

gavinking:
Actually, @NathanQingyangXu I just went ahead and merged #3850, which means you'll need to adapt this PR.

NathanQingyangXu:
sure. Thanks for letting me know.

NathanQingyangXu:
@gavinking I found another issue related to PostgresUUIDType's new magic sql code. During BasicBinder logging, the following weird statement will be produced:
```
22:22:21,578 TRACE org.hibernate.orm.jdbc.bind:(BasicBinder:60) - binding parameter [3] as [UNKNOWN(3975)] - [ccc92a00-be56-4f0b-9966-f7e79604ee20]
```
The reason is internally BasicBinder has no way to get the real sql type code. I added another separate commit (https://github.com/hibernate/hibernate-orm/pull/3974/commits/ed4bba1d36dc3668616d2b9daef9ed123ccffb83) to solve this issue by adding a new 'getLoggableJdbcTypeCode()' to JdbcTypeDescriptor and then the correct binding result will show up as below:
```
binding parameter [3] as [VARCHAR] - [ccc92a00-be56-4f0b-9966-f7e79604ee20]
```
Let me know whether this commit is necessary. I am ready to have it removed if you wish. Thanks.

NathanQingyangXu:
@gavinking there is still one imperfection due to the magical PostgresUUID sql code. In PostgreSQLDialect, BasicBinder will output weird logging as below:
```
binding parameter [1] as [UNKNOWN(3975] - [743433bf-b0e2-4db7-88f4-f6c01182f4dc]
```
Not sure whether we need to fix this in this PR and/or how to fix it in most elegant way (inserting entry into org.hibernate.type.descriptor.**JdbcTypeNameMapper** or tweak **BasicBinder**).

sebersole (commented):
The new output is much nicer!

I'll take a look tomorrow and get this integrated into the release I plan on doing tomorrow.

One thing off the bat... This needs a dedicated Jira.  There is much more to this than just "moving tests around"

NathanQingyangXu:
> The new output is much nicer!
> 
> I'll take a look tomorrow and get this integrated into the release I plan on doing tomorrow.
> 
> One thing off the bat... This needs a dedicated Jira. There is much more to this than just "moving tests around"

I created a new jira ticket (https://hibernate.atlassian.net/browse/HHH-14600) and changed the commit message accordingly.

beikov (approved):
LGTM

sebersole:
> I created a new jira ticket (https://hibernate.atlassian.net/browse/HHH-14600) and changed the commit message accordingly.

Thanks!  Looking this one over atm


> I also updated the PR to correct typos of 'extracter' at #3561 and created a jira ticket for it (https://hibernate.atlassian.net/browse/HHH-14601) as well.

This PR fails with compile errors?

hibernate-core/src/main/java/org/hibernate/type/descriptor/jdbc/JdbcTypeDescriptor.java
```diff
@@ -124,6 +124,16 @@ default boolean isNumber() {
 		return false;
 	}
 
+	default boolean isBinary() {
```
sebersole, line 127: BLOB?

sebersole (commented):
LGTM^2

NathanQingyangXu:
> > I created a new jira ticket (https://hibernate.atlassian.net/browse/HHH-14600) and changed the commit message accordingly.
> 
> Thanks! Looking this one over atm
> 
> > I also updated the PR to correct typos of 'extracter' at #3561 and created a jira ticket for it (https://hibernate.atlassian.net/browse/HHH-14601) as well.
> 
> This PR fails with compile errors?

The compiler error was fixed.

hibernate-core/src/main/java/org/hibernate/type/descriptor/jdbc/JdbcTypeDescriptor.java
NathanQingyangXu: Seems the original intention here is binary (as the counterpart of 'isString()' method), not BLOB. Do you mean adding 'BLOB' to the method body? I checked the type list in 'isString' and found it does not contain CLOB.

sebersole: Though I also do not see `CLOB` nor `NCLOB` listed under `#isString()`.  Guess it depends how these are used.  But if LOBs are not supposed to be included, that should be either better indicated in the method name or at least mentioned in the Javadoc.  

@beikov ?

beikov: That method only seems to be used for the purpose of deciding how to inline parameter values. I guess `CLOB` and `NCLOB` could be considered as strings as well.

sebersole: Ok, then we should add all 3.  Or document why they are not there.

Let's not hold up the release for that though - https://hibernate.atlassian.net/browse/HHH-14605

PR #3977, opened 2021-05-11
HHH-14585 size() HQL function discards '@Where' clause

fax4ever:
https://hibernate.atlassian.net/browse/HHH-14585

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/CollectionSizeNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/CollectionSizeNode.java
@@ -12,6 +12,7 @@ import org.hibernate.hql.internal.antlr.HqlSqlTokenTypes;
 import org.hibernate.internal.util.StringHelper;
 import org.hibernate.persister.collection.CollectionPropertyMapping;
 import org.hibernate.persister.collection.CollectionPropertyNames;
+import org.hibernate.persister.collection.OneToManyPersister;
 import org.hibernate.persister.collection.QueryableCollection;
 import org.hibernate.type.StandardBasicTypes;
 
@@ -89,6 +90,8 @@ public class CollectionSizeNode extends SqlNode implements SelectExpression {
 			buffer.append( ownerKeyColumns[i] ).append( " = " ).append( collectionKeyColumns[i] );
 		}
 
+		buffer.append( collectionDescriptor.filterFragment( collectionTableAlias,
+				collectionOwnerFromElement.getWalker().getEnabledFilters() ) );
 		buffer.append( ")" );
 
 		if ( scalarName != null ) {
--- a/hibernate-core/src/test/java/org/hibernate/test/hql/size/WhereClauseOrderBySizeTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/hql/size/WhereClauseOrderBySizeTest.java
@@ -7,15 +7,12 @@
 
 package org.hibernate.test.hql.size;
 
-import org.hibernate.annotations.ResultCheckStyle;
-import org.hibernate.annotations.SQLDelete;
-import org.hibernate.annotations.Where;
-import org.hibernate.jpa.test.BaseEntityManagerFunctionalTestCase;
+import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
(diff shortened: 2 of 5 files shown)


sebersole (approved):

hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/CollectionSizeNode.java
```diff
@@ -89,6 +90,9 @@ public String toSqlExpression() {
 			buffer.append( ownerKeyColumns[i] ).append( " = " ).append( collectionKeyColumns[i] );
 		}
 
+		String filterFragment = collectionDescriptor.filterFragment( collectionTableAlias,
```
sebersole, line 93: One minor quibble... either mark this var `final` or get rid of the var completely and just add the filter-fragment directly to the buffer.

hibernate-core/src/test/java/org/hibernate/test/hql/size/WhereClauseOrderBySizeTest.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.List;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
+import static org.junit.Assert.assertEquals;
+
+@TestForIssue(jiraKey = "HHH-14585")
+@SkipForDialects({
+		@SkipForDialect(value = SQLServerDialect.class, comment = "The @SQLDelete expression seems to be not working with this dialect"),
+		@SkipForDialect(value = DB2Dialect.class, comment = "The @SQLDelete expression seems to be not working with this dialect"),
+		@SkipForDialect(value = Oracle12cDialect.class, comment = "The @SQLDelete expression seems to be not working with this dialect")
```
beikov, line 42: Please update the comments to say something like `The database does not support boolean data types`. I think `WhereAnnotatedOneToManySizeTest` works on SQL Server and DB2 because `true`/`false` is an alias for `1`/`0`. For updating, this might not work though as you seem to have discovered.

hibernate-core/src/test/java/org/hibernate/test/hql/size/filter/WhereAnnotatedOneToManySizeTest.java
```diff
@@ -0,0 +1,127 @@
...
+	@After
+	public void after() {
+		inTransaction( session -> {
+			session.createQuery("DELETE FROM City c").executeUpdate();
+			session.createQuery("DELETE FROM Region c").executeUpdate();
+		} );
+	}
+
+	@Test
+	public void orderBy_sizeOf() {
```
beikov, line 89: DB2 does not support correlated subqueries in the ORDER BY clause. A workaround would be to use `select r, size(r.cities) s from Region r order by s desc`

hibernate-core/src/test/java/org/hibernate/test/hql/size/filter/WhereAnnotatedOneToManySizeTest.java
```diff
@@ -0,0 +1,127 @@
...
+		inSession( session -> {
+			QueryImplementor<Region> query = session.createQuery(
+					"FROM Region r ORDER BY size(r.cities) DESC", Region.class );
+			List<Region> regions = query.getResultList();
+			assertThat( regions ).extracting( "name" ).containsExactly( "Lombardy", "Lazio" );
+		} );
+	}
+
+	@Test
+	public void orderBy_dotSize() {
```
beikov, line 99: DB2 does not support correlated subqueries in the ORDER BY clause. A workaround would be to use `select r, r.cities.size s from Region r order by s desc`

hibernate-core/src/test/java/org/hibernate/test/hql/size/WhereClauseOrderBySizeTest.java
dreab8: Thanks @beikov , Fabio and I were about to ask about the db2 issue :)

fax4ever:
@mrizzi @dreab8 @sebersole @beikov thank you all for your efforts!
Is there anything else we can do to improve this contribution or can I merge it as soon as the Ci passes?

beikov (approved):
IMO you can merge it as soon as the tests pass!

PR #3981, opened 2021-05-13
HHH-14603: Removal of the deprecated DTDEntityResolver and ConfigHelper classes

jrenaat:
Also:
- deprecated the EJB3DTDEntityResolver
- created Internal utility ResourcesHelper class so that resources can be loaded when the ClassLoaderService is not available
- Remove deprecated ClassLoaderService.fromConfigSettings method, as well as related AvailableSetting constants


sebersole:
I'll look tomorrow.  But this should absolutely have a related Jira

jrenaat:
> I'll look tomorrow. But this should absolutely have a related Jira

Totally right, done: https://hibernate.atlassian.net/browse/HHH-14603

hibernate-core/src/main/java/org/hibernate/cfg/Environment.java
```diff
@@ -177,28 +178,31 @@ public static void verifyProperties(Map<?,?> configurationValues) {
 		//Set USE_REFLECTION_OPTIMIZER to false to fix HHH-227
 		GLOBAL_PROPERTIES.setProperty( USE_REFLECTION_OPTIMIZER, Boolean.FALSE.toString() );
 
-		try {
-			InputStream stream = ConfigHelper.getResourceAsStream( "/hibernate.properties" );
+		LinkedList<ClassLoader> classLoaders = new LinkedList<>();
```
beikov, line 181: Please use an `ArrayList` or `Arrays.asList( Thread.currentThread().getContextClassLoader(), Environment.class.getClassLoader() )` as `LinkedList` is rarely appropriate.

beikov (approved):
LGTM

hibernate-core/src/main/java/org/hibernate/cfg/Environment.java
jrenaat: It is in this case I think, since I want to make sure to respect the order in which ClassLoaders are interrogated, as before this change.

sebersole: List is inherently ordered

sebersole: Even an array would work here as the number of items is known and constant

jrenaat: > List is inherently ordered

I seem to still confuse Collection.iterator() with List.iterator() ... after all these years, it's just  .... sad :-(((

hibernate-core/src/main/java/org/hibernate/cfg/EJB3DTDEntityResolver.java
```diff
@@ -114,13 +120,45 @@ else if ( systemId.endsWith( "persistence_1_0.xsd" ) ) {
...
-		InputSource source = super.resolveEntity( publicId, systemId );
+		InputSource source = _resolveEntity( publicId, systemId );
 		if ( source != null ) {
 			resolved = true;
 		}
 		return source;
 	}
 
+	// Copied over from DTDEntityResolver that was removed
+	private InputSource _resolveEntity(String publicId, String systemId) {
```
NathanQingyangXu, line 131: Do we feel the necessity of such '_' prefixing naming convention? IMHO, it is not necessary in Java.

hibernate-core/src/main/java/org/hibernate/cfg/EJB3DTDEntityResolver.java
```diff
@@ -137,7 +175,51 @@ private InputSource buildInputSource(String publicId, String systemId, InputStre
...
+			if ( stream == null && hasLeadingSlash ) {
+				stream = classLoader.getResourceAsStream( stripped );
+			}
+		}
+
+		if ( stream == null ) {
+			stream = Environment.class.getClassLoader().getResourceAsStream( path );
+		}
+		if ( stream == null && hasLeadingSlash ) {
+			stream = Environment.class.getClassLoader().getResourceAsStream( stripped );
```
NathanQingyangXu, line 220: Could we move the 'stripped' creation logic into this if block? We can enjoy two benefits:

- better code readability for the variable creation is near its usage;
- potential avoidance of unnecessary performance cost for obviously it is perfectly normal when we don't need to create the 'stripped' string.

I would even go as far as below:
```
if ( stream == null && path.startsWith( "/" ) {
    stream = Environment.class.getClassLoader().getResourceAsStream( path.substring( 1 ) );
}
```

hibernate-core/src/main/java/org/hibernate/cfg/Environment.java
```diff
@@ -177,28 +179,31 @@ public static void verifyProperties(Map<?,?> configurationValues) {
 		//Set USE_REFLECTION_OPTIMIZER to false to fix HHH-227
 		GLOBAL_PROPERTIES.setProperty( USE_REFLECTION_OPTIMIZER, Boolean.FALSE.toString() );
 
-		try {
-			InputStream stream = ConfigHelper.getResourceAsStream( "/hibernate.properties" );
+		List<ClassLoader> classLoaders = new ArrayList<>();
```
NathanQingyangXu, line 182: I think we can specify the size here:
```
List<ClassLoader> classLoaders = new ArrayList<>( 2 );
```

sebersole: I agree, although I'd use varargs

sebersole: > I agree, although I'd use varargs

To clarify, I mean 

```
class ResourcesHelper {
    ...

    public static InputStream locateResourceAsStream(String resourceName, ClassLoader... classLoaders) { ... }
}
```

and then here:

```
final InputStream stream = ResourcesHelper.locateResourceAsStream( 
        "/hibernate.properties",
        Thread.currentThread().getContextClassLoader(),
        Environment.class.getClassLoader()
);
```

jrenaat: > > I agree, although I'd use varargs
> 
> To clarify, I mean
> 
> ```
> class ResourcesHelper {
>     ...
> 
>     public static InputStream locateResourceAsStream(String resourceName, ClassLoader... classLoaders) { ... }
> }
> ```
> 
> and then here:
> 
> ```
> final InputStream stream = ResourcesHelper.locateResourceAsStream( 
>         "/hibernate.properties",
>         Thread.currentThread().getContextClassLoader(),
>         Environment.class.getClassLoader()
> );
> ```

Yes, I did get that, but thanks anyhow for the clarification ;)

beikov (approved):

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -233,41 +233,6 @@
 	 */
 	String TC_CLASSLOADER = "hibernate.classLoader.tccl_lookup_precedence";
 
-	/**
```
sebersole, line 236: :+1:

hibernate-core/src/main/java/org/hibernate/cfg/EJB3DTDEntityResolver.java
```diff
@@ -134,10 +155,46 @@ private InputSource buildInputSource(String publicId, String systemId, InputStre
...
+				LOG.usingOldDtd();
+			}
+		}
 		return dtdStream;
 	}
+
+	// Copied over from ConfigHelper that was removed
+	private InputStream resolveInLocalNamespace(String path) {
+		boolean hasLeadingSlash = path.startsWith( "/" );
+		String stripped = hasLeadingSlash ? path.substring( 1 ) : path;
```
sebersole, line 179: I wonder how often there is a difference between look-up with the leading slash and without.

Anyway, the `ResourceHelper` methods already handle this.  So why not just call `ResourceHelper` here as well?

Also, why are `getStreamFromJpaClasspath` and `getStreamFromNonJpaClasspath` so different?  They both (should)  effectively do the same thing.

hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/PersistenceXmlParser.java
```diff
@@ -62,7 +64,7 @@
 	 */
 	public static List<ParsedPersistenceXmlDescriptor> locatePersistenceUnits(Map integration) {
 		final PersistenceXmlParser parser = new PersistenceXmlParser(
```
sebersole, line 66: This should be changed quite a bit IMO....

`#locatePersistenceUnits` is only ever called from one place in "real code".  That call already has access to the `ClassLoaderService`.  We should just pass that one along.  The other uses are tests.  For those we can just create the CLS, or get it from the `ServiceRegistry` if they have access to one.

`#locateIndividualPersistenceUnit` - None of the overloads are ever called from outside this class.  IMO we should just remove them.  This is an internal class, so changing stuff is kosher.

`#locateNamedPersistenceUnit` - Same as `#locateIndividualPersistenceUnit`

`#parse` - One form is never used.  The other is used from a context that does not (atm) have access to CLS.  So we would need to do something there.  Though I think we want to find an option here that does not involve creating 2 CLS refs.  This method is called when creating an `EntityManagerFactoryBuilder`.  That EMFB will eventually create a CLS (or use one passed in).  So not only is it something I'd prefer to avoid perf wise[1], but it is potentially wrong if we are passed a CLS or ClassLoader to build into a CLS.


[1] Yes this "throw away" CLS is only created and used at boot time.  Still...

hibernate-core/src/main/java/org/hibernate/cfg/EJB3DTDEntityResolver.java
jrenaat: > I wonder how often there is a difference between look-up with the leading slash and without.
> 
> Anyway, the `ResourceHelper` methods already handle this. So why not just call `ResourceHelper` here as well?
> 
> Also, why are `getStreamFromJpaClasspath` and `getStreamFromNonJpaClasspath` so different? They both (should) effectively do the same thing.


So, to recap where this comes from, the point was to remove the deprectated DTDEntityResolver, but since the EJB3DTDEntityResolver wasn't deprecated, functionality from the former that the latter depends on had to be moved over. In the same go, deprecate EJB3DTDEntityResolver so that we can remove it in the near future as well.
My primary concern in this refactoring is to do it with minimal risk of introducing new bugs (there is also no test for either class to be able to double check), so i just copied over whatever was needed with that in mind.
That's the motivation of trying to change this code as little as possible and why there still is some difference between 'getStreamFromJpaClasspath' and 'getStreamFromNonJpaClasspath'. I considered the possibility of taking it further, but I don't know the internals of the old code enough to know why this or that is done the way it is so again, trying to respect what there was as much as possible.
I could of course write a test for this class, but I wonder if it's worth the effort, since this is most likely not being used, and now also deprecated.
That said, I'll take the refactoring as far as you want me to, it's more a matter of me not knowing what the general policy is in such cases.

hibernate-core/src/test/java/org/hibernate/jpa/test/persistenceunit/DuplicatePersistenceUnitNameTest.java
```diff
@@ -75,13 +75,13 @@ public void testDuplicatePersistenceUnitNameLogAWarnMessage() {
...
 		public TestClassLoader() {
 			urls = Arrays.asList(
-				findAsResource(
-					"org/hibernate/jpa/test/persistenceunit/META-INF/persistence.xml"
-				)
-				,
-				findAsResource(
-					"org/hibernate/jpa/test/persistenceunit/META-INF/persistenceUnitForNameDuplicationTest.xml"
-				)
+					ClassLoaderServiceTestingImpl.INSTANCE.locateResource(
```
sebersole, line 78: Hard to tell from just diff... did you change these from tabs to spaces?  Or just change the tab-indentation-depth?  Either way, this looks wrong

sebersole (changes requested):
Left a few comments.

hibernate-core/src/test/java/org/hibernate/jpa/test/persistenceunit/DuplicatePersistenceUnitNameTest.java
jrenaat: this is IntelliJ's auto formatting utility acting up I'm afraid

beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/4020

PR #3983, opened 2021-05-14
HHH-14069 Testcase

waldi5001:
LazyLoading with @Inheritance(SINGLE_TABLE) doesn't honor the discriminator value, which result in a WrongClassException

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/lazyload/Child1Entity.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/lazyload/Child1Entity.java
@@ -1,0 +1,10 @@
+package org.hibernate.test.lazyload;
+
+import javax.persistence.DiscriminatorValue;
+import javax.persistence.Entity;
+
+@Entity
+@DiscriminatorValue(value = "1")
+public class Child1Entity extends ParentManyEntity {
+
+}
--- a/hibernate-core/src/test/java/org/hibernate/test/lazyload/Child2Entity.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/lazyload/Child2Entity.java
@@ -1,0 +1,10 @@
+package org.hibernate.test.lazyload;
+
+import javax.persistence.DiscriminatorValue;
+import javax.persistence.Entity;
+
+@Entity
+@DiscriminatorValue(value = "2")
+public class Child2Entity extends ParentManyEntity {
+
+}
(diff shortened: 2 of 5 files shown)


gavinking:
I understand that this sort of mapping is usually handled using the `@Where` annotations, as described in the documentation here:

<https://docs.jboss.org/hibernate/orm/5.2/userguide/html_single/Hibernate_User_Guide.html#mapping-where-example>

waldi5001:
Hallo @gavinking, 
thanks for the reply. Your suggestion is a workable workaround and in the end, we use it. In my opinion this bug is critical, for the following reasons: 
- First, if you are not familiar with hibernate the behavior is strange and cost you hours of debugging and finding a solution for this problem.
- Second you can run in a situation that you are using unnoticed wrong objects. As a user you see objects of class A, but the objects are of class B.

Is there a realistic chance to fix the Bug? And if not, can you give me a hint to fix the bug and provide it here in the PR?
Thanks in advance.

gavinking:
Well, yes, I believe it's in-principle fixable (perhaps quite _easily_ fixable) by adding a discriminator column to the join.

We discussed it here:

<https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/topic/something.20that.20might.20be.20worth.20fixing/near/236839046>

And I also commented about the issue here:

<https://github.com/eclipse-ee4j/jpa-api/issues/39#issuecomment-830112571>

Now, I think this would be nice to fix, and I might look at it myself if/when I find time, but as I said in those threads, I definitely don't think it's "critical". Mapping two different associations to the same FK column is pretty strange thing to do to begin with, and if you *do* do this strange thing then the `@Where()` workaround isn't so awful.

If you want to try, the trick is to convince Hibernate to add the discriminator column to the table join in single table inheritance mappings. That's probably pretty straightforward. The trickier part, I imagine, is detecting when you need to do this. The problem is that Hibernate simply doesn't know that you mapped two different associations to the same FK column. You need to detect that. Perhaps it's not so hard, I'm not sure.

gavinking:
> The problem is that Hibernate simply doesn't know that you mapped two different associations to the same FK column. You need to detect that. Perhaps it's not so hard, I'm not sure.

And for consistency, you probably even need to detect it when the associations belong to different entities.

gavinking:
> > The problem is that Hibernate simply doesn't know that you mapped two different associations to the same FK column. You need to detect that. Perhaps it's not so hard, I'm not sure.
> 
> And for consistency, you probably even need to detect it when the associations belong to different entities.

Actually that's not right: probably the right thing to do is to detect that the association references a type whose discriminator column is declared by a supertype.

gavinking:
I believe that #4200 demonstrates that this now works in H6.

I don't think we should try to fix this in H5.

So I recommend closing this PR.

PR #3984, opened 2021-05-14
wip/6.0 HHH 14605 Handling for BLOB, CLOB and NCLOB relative to JavaTypeDescriptor `#isString` and `#isBinary`

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14605

Given we have discussed about this, especially we have confirmed with Beikov, we can go ahead to fix it right away. See the discussion in https://github.com/hibernate/hibernate-orm/pull/3974.


beikov:
Thanks!

PR #3985, opened 2021-05-15
wip/6.0 HHH-14607 prefer FetchOptions over FetchStrategy

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14607

The ultimate goal is to avoid instantiating **FetchStrategy** (currently the only implementation of **FetchOptions** interface which is simply a combination of **FetchTimingAccess** interface and **FetchStyleAccess** interface) to eliminate a performance drain.

In the new v6 mapping stuff, it is easy to achieve that by simply creating various overloaded constructors which accepts **FetchTiming** and **FetchStyle** in instead of **FetchOptions** (or even worse, **FetchStrategy**).

In the deprecated v5 loading stuff (mainly in **walker** package), I went even further to delete **FetchStrategy** totally by:

1. substitute **FetchOptions** for **FetchStrategy** in method signature(s);
2. create a static factory method (i.e. **FetchOptions**.**valueOf**(**FetchTiming** timing, **FetchStyle** style)) and use it to return **FetchOptions** instance (internally a package private **FetchOptionsImpl** class was created but the whole codebase needs to know nothing about its existence outside the package where both **FetchOptions** and it reside).

Some related refactoring is also done to make the eliminating of **FetchStrategy** complete. 

TBH, deleting **FetchStrategy** is not mandatory for sooner or later v5 loading stuff would be deleted and currently there is no need of a FetchOptions implementation class in v6 stuff, but I do think it is a big concept simplification for we only need to work on **FetchOptions** interface alone and we can get rid of the potential confusion between **FetchStrategy** and **FetchOptions**(their relationship, lack of naming convention, potential misuse in method signature or using implementation class instead of interface, etc.).

Let me know whether my rationale of eliminating FetchStrategy is sensible and I am always ready to revert it back.


hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -716,7 +715,7 @@ private void applyOrdering(TableGroup tableGroup, OrderByFragment orderByFragmen
 				// 'entity graph' takes precedence over 'fetch profile'
 				if ( entityGraphTraversalState != null ) {
 					traversalResult = entityGraphTraversalState.traverse( fetchParent, fetchable, isKeyFetchable );
-					fetchTiming = traversalResult.getFetchStrategy();
```
NathanQingyangXu, line 719: This method name is confusing esp. when we have a FetchStrategy class already. I changed it to better reflect its purpose.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/DiscriminatedAssociationAttributeMapping.java
```diff
@@ -75,9 +73,8 @@ public DiscriminatedAssociationAttributeMapping(
 				bootProperty.getName(),
 				stateArrayPosition,
 				attributeMetadataAccess,
-				fetchTiming == FetchTiming.IMMEDIATE
-						? new FetchStrategy( FetchTiming.IMMEDIATE, FetchStyle.SELECT )
-						: new FetchStrategy( FetchTiming.DELAYED, FetchStyle.SELECT ),
+				fetchTiming,
+				FetchStyle.SELECT,
 				declaringType,
```
NathanQingyangXu, line 78: The above is a perfect example why this PR is necessary.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/Fetchable.java
```diff
@@ -19,11 +19,6 @@
 
 	FetchOptions getMappedFetchOptions();
 
-	// todo (6.0) : all we need here is (1) FetchTiming and (2) whether the values are available in the current JdbcValuesSource
-	//		Having to instantiate new FetchStrategy potentially multiple times
-	// 		per Fetch generation is performance drain.  Would be better to
-	// 		simply pass these 2 pieces of information
-
 	default Fetch resolveCircularFetch(
```
NathanQingyangXu, line 22: I think the above comment could be deleted after this PR is merged.

beikov (approved):
LGTM but I don't know for sure if this is what @sebersole originally intended, so let's wait for his ok before merging this.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/DiscriminatedAssociationAttributeMapping.java
sebersole: Yes, cases like this are exactly why I created this task

sebersole (approved):
> LGTM but I don't know for sure if this is what @sebersole originally intended, so let's wait for his ok before merging this.

This is more-or-less exactly what I intended.  Relying on the interface instead of the class allows us to save instantiations

PR #3986, opened 2021-05-16
HHH-14608 Merge causes StackOverflow when JPA proxy compliance is enabled

dreab8:
https://hibernate.atlassian.net/browse/HHH-14608

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/proxy/pojo/BasicLazyInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/proxy/pojo/BasicLazyInitializer.java
@@ -62,7 +62,7 @@ public abstract class BasicLazyInitializer extends AbstractLazyInitializer {
 				return System.identityHashCode( proxy );
 			}
 			else if ( isUninitialized() && method.equals( getIdentifierMethod ) ) {
-				return getIdentifier();
+				return getIdentifier( isInitializeProxyWhenAccessingIdentifier() );
 			}
 			else if ( "getHibernateLazyInitializer".equals( methodName ) ) {
 				return this;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/proxy/pojo/BasicLazyInitializer.java
```diff
@@ -89,6 +89,19 @@ else if ( method.equals( setIdentifierMethod ) ) {
 
 	}
 
+	Serializable getIdentifier(boolean initialize) {
+		if ( initialize ) {
+			initialize();
+		}
+		return getIdentifier();
```
NathanQingyangXu, line 96: We might well improve code readability a little bit by changing the method parameter to 'needsToInitialize' or stuff like that, so causal code reader could  grasp the rough meaning solely based on method signature without digging into impl details in method body.

Not sure how to improve the method name of 'isInitializeProxyWhenAccessingIdentifier' though.
Feel free to ignore my comment. Not a big deal.

dreab8: Thanks @NathanQingyangXu , I renamed it `needsToInitializeProxy`

gbadner:
IMO, the logic should remain in `AbstractLazyInitializer` so if/when a new `LazyInitializer` extends `AbstractLazyInitializer` (but not BasicLazyInitializer`), it will work as expected with `hibernate.jpa.compliance.proxy=true`.

With the proposed change, `MapLazyInitializer` is no longer affected by `hibernate.jpa.compliance.proxy=true` since it doesn't extend `BasicLazyInitializer`. Obviously, dynamic maps are not defined by JPA, so if an application uses them, they likely would not set `hibernate.jpa.compliance.proxy=true`.

My preference would be to add a method to `AbstractLazyInitializer` that does not have the initialization logic in it for Hibernate's use only; e.g., `#getIdentifierInternal`.

dreab8:
Thanks @gbadner for the suggestion, what do you think about this https://github.com/hibernate/hibernate-orm/pull/3997 ?

gbadner:
Superseded by https://github.com/hibernate/hibernate-orm/pull/3997.

PR #3987, opened 2021-05-17
wip/6.0 Fix some obvious issues found by IntelliJ IDE's Inspect Code

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14615

Quite some valuable bugs flushed out via IDE's help. Left detailed comment for each of them.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/annotations/reflection/internal/JPAXMLOverriddenAnnotationReader.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/annotations/reflection/internal/JPAXMLOverriddenAnnotationReader.java
@@ -251,10 +251,10 @@ public class JPAXMLOverriddenAnnotationReader implements AnnotationReader {
 		annotationToXml.put( AccessType.class, "access" );
 		annotationToXml.put( AttributeOverride.class, "attribute-override" );
 		annotationToXml.put( AttributeOverrides.class, "attribute-override" );
-		annotationToXml.put( AttributeOverride.class, "association-override" );
-		annotationToXml.put( AttributeOverrides.class, "association-override" );
-		annotationToXml.put( AttributeOverride.class, "map-key-attribute-override" );
-		annotationToXml.put( AttributeOverrides.class, "map-key-attribute-override" );
+		annotationToXml.put( AssociationOverride.class, "association-override" );
+		annotationToXml.put( AssociationOverrides.class, "association-override" );
+		//annotationToXml.put( AttributeOverride.class, "map-key-attribute-override" );
+		//annotationToXml.put( AttributeOverrides.class, "map-key-attribute-override" );
 		annotationToXml.put( Id.class, "id" );
 		annotationToXml.put( EmbeddedId.class, "embedded-id" );
 		annotationToXml.put( GeneratedValue.class, "generated-value" );
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
@@ -281,10 +281,10 @@ public class StatefulPersistenceContext implements PersistenceContext {
 	@Override
 	public void setEntryStatus(EntityEntry entry, Status status) {
 		entry.setStatus( status );
-		setHasNonReadOnlyEnties( status );
+		setHasNonReadOnlyEntities( status );
 	}
 
-	private void setHasNonReadOnlyEnties(Status status) {
+	private void setHasNonReadOnlyEntities(Status status) {
 		if ( status==Status.DELETED || status==Status.MANAGED || status==Status.SAVING ) {
 			hasNonReadOnlyEntities = true;
 		}
@@ -558,7 +558,7 @@ public class StatefulPersistenceContext implements PersistenceContext {
 
 		entityEntryContext.addEntityEntry( entity, e );
 
-		setHasNonReadOnlyEnties( status );
+		setHasNonReadOnlyEntities( status );
 		return e;
 	}
 
@@ -569,7 +569,7 @@ public class StatefulPersistenceContext implements PersistenceContext {
 		((ManagedEntity)entity).$$_hibernate_getEntityEntry().setStatus( status );
 		entityEntryContext.addEntityEntry( entity, ((ManagedEntity)entity).$$_hibernate_getEntityEntry() );
 
-		setHasNonReadOnlyEnties( status );
+		setHasNonReadOnlyEntities( status );
 		return ((ManagedEntity)entity).$$_hibernate_getEntityEntry();
 	}
 
--- a/hibernate-core/src/main/java/org/hibernate/internal/util/collections/ConcurrentReferenceHashMap.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/util/collections/ConcurrentReferenceHashMap.java
@@ -451,7 +451,7 @@ public class ConcurrentReferenceHashMap<K, V> extends AbstractMap<K, V>
 		}
 
 		@SuppressWarnings("unchecked")
-		static final <K, V> HashEntry<K, V>[] newArray(int i) {
+		static <K, V> HashEntry<K, V>[] newArray(int i) {
 			return new HashEntry[i];
 		}
 	}
@@ -560,7 +560,7 @@ public class ConcurrentReferenceHashMap<K, V> extends AbstractMap<K, V>
 		}
 
 		@SuppressWarnings("unchecked")
-		static final <K, V> Segment<K, V>[] newArray(int i) {
+		static <K, V> Segment<K, V>[] newArray(int i) {
 			return new Segment[i];
 		}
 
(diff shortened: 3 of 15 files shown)


hibernate-core/src/main/java/org/hibernate/cfg/annotations/reflection/internal/JPAXMLOverriddenAnnotationReader.java
```diff
@@ -251,10 +251,10 @@
...
 		annotationToXml.put( AttributeOverride.class, "attribute-override" );
 		annotationToXml.put( AttributeOverrides.class, "attribute-override" );
-		annotationToXml.put( AttributeOverride.class, "association-override" );
-		annotationToXml.put( AttributeOverrides.class, "association-override" );
-		annotationToXml.put( AttributeOverride.class, "map-key-attribute-override" );
-		annotationToXml.put( AttributeOverrides.class, "map-key-attribute-override" );
+		annotationToXml.put( AssociationOverride.class, "association-override" );
+		annotationToXml.put( AssociationOverrides.class, "association-override" );
+		//annotationToXml.put( AttributeOverride.class, "map-key-attribute-override" );
+		//annotationToXml.put( AttributeOverrides.class, "map-key-attribute-override" );
```
NathanQingyangXu, line 257: Not sure what we can do for the above two annotations

hibernate-core/src/main/java/org/hibernate/internal/util/config/ConfigurationHelper.java
```diff
@@ -462,17 +462,16 @@ public static String resolvePlaceHolder(String property) {
...
+					while ( x < chars.length && chars[x] != '}' ) {
 						// if we reach the end of the string w/o finding the
 						// matching end, that is an exception
 						if ( x == chars.length - 1 ) {
 							throw new IllegalArgumentException( "unmatched placeholder start [" + property + "]" );
 						}
+						x++;
 					}
-					String systemProperty = extractFromSystem( systemPropertyName );
+					String systemProperty = extractFromSystem( property.substring( pos + 2, x ) );
```
NathanQingyangXu, line 474: Not 100% sure about the correctness, but the original `systemPropertyName += chars[x];` might be too inefficient.

hibernate-core/src/main/java/org/hibernate/loader/ast/spi/SingleEntityLoader.java
```diff
@@ -25,6 +25,6 @@
 	T load(Object key, LockOptions lockOptions, Boolean readOnly, SharedSessionContractImplementor session);
 
 	default T load(Object key, LockOptions lockOptions, SharedSessionContractImplementor session) {
-		return load( key, lockOptions, session );
+		return load( key, lockOptions, Boolean.FALSE, session );
```
NathanQingyangXu, line 28: The default implementation would end up with infinite loop.

hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
```diff
@@ -281,10 +281,10 @@ public boolean hasNonReadOnlyEntities() {
 	@Override
 	public void setEntryStatus(EntityEntry entry, Status status) {
 		entry.setStatus( status );
-		setHasNonReadOnlyEnties( status );
+		setHasNonReadOnlyEntities( status );
 	}
 
-	private void setHasNonReadOnlyEnties(Status status) {
+	private void setHasNonReadOnlyEntities(Status status) {
```
NathanQingyangXu, line 287: typos in code are in the scope of this PR, whereas typos in comment are not.

hibernate-core/src/main/java/org/hibernate/internal/util/collections/ConcurrentReferenceHashMap.java
```diff
@@ -451,7 +451,7 @@ final void setValue(V value, ReferenceType valueType, ReferenceQueue<Object> ref
 		}
 
 		@SuppressWarnings("unchecked")
-		static final <K, V> HashEntry<K, V>[] newArray(int i) {
+		static <K, V> HashEntry<K, V>[] newArray(int i) {
```
NathanQingyangXu, line 454: The combination of 'static' and 'final' is really confusing, for the 'final' keyword is only meaningful in non-static context.

hibernate-core/src/main/java/org/hibernate/cfg/annotations/reflection/internal/JPAXMLOverriddenAnnotationReader.java
beikov: AFAICT the code only uses `annotationToXml.containsKey()` and once `annotationToXml.getOrDefault()` for exception reporting, so this is fine IMO.

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
```diff
@@ -1139,7 +1139,7 @@ public String getPropertyTableName(String propertyName) {
 		return constraintOrderedTableNames;
 	}
 
-	public String[][] getContraintOrderedTableKeyColumnClosure() {
+	public String[][] getConstraintOrderedTableKeyColumnClosure() {
```
beikov, line 1142: Changing these public contracts on the other hand might be problematic as that is not backwards compatible. WDYT @sebersole and @dreab8? Should we just do this? +1 from my side

hibernate-core/src/main/java/org/hibernate/persister/entity/Queryable.java
```diff
@@ -82,7 +82,7 @@
 	 * the first dimension.  If not, that would be a problem ;)
 	 *
 	 */
-	public String[][] getContraintOrderedTableKeyColumnClosure();
+	public String[][] getConstraintOrderedTableKeyColumnClosure();
```
beikov, line 85: Same as for `JoinedSubclassEntityPersister`

hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
```diff
@@ -911,7 +911,7 @@ public boolean isMultiTable() {
 		return constraintOrderedTableNames;
 	}
 
-	public String[][] getContraintOrderedTableKeyColumnClosure() {
+	public String[][] getConstraintOrderedTableKeyColumnClosure() {
```
beikov, line 914: Same as for `JoinedSubclassEntityPersister`

hibernate-core/src/main/java/org/hibernate/persister/entity/UnionSubclassEntityPersister.java
```diff
@@ -558,7 +558,7 @@ public String getPropertyTableName(String propertyName) {
 	}
 
 	@Override
-	public String[][] getContraintOrderedTableKeyColumnClosure() {
+	public String[][] getConstraintOrderedTableKeyColumnClosure() {
```
beikov, line 561: Same as for `JoinedSubclassEntityPersister`

beikov (commented):
Looks good except for the few naming changes which are not backwards compatible. That part requires a team decision. Thanks for the PR!

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
dreab8: +1 for me, probably we can deprecate them in 5.5, adding the new method method that calls the old one

dreab8: I may be wrong but it seems that in 6 we are not using this method anymore

sebersole: Probably.  I added that originally to support bulk manipulation queries, which are handled very differently now.

Still the point here is valid - this is a "public" API and we need to be careful changing those.  I vote we simply deprecate it

hibernate-core/src/main/java/org/hibernate/internal/util/collections/ConcurrentReferenceHashMap.java
djarnis73: That is not quite true. A static final is a static that cannot be reassigned.

NathanQingyangXu:
Yeah, static final is meaningful for variable. But here it is used on
method. In this context, final is meaningless for static method cannot be
overridden by child class.

djarnis73:
Yeah, sorry my bad, did not pay enough attention.

> Yeah, static final is meaningful for variable. But here it is used on method. In this context, final is meaningless for static method cannot be overridden by child class.

NathanQingyangXu:
No worries. I think my original comment is misleading without further
explanation. Thanks for your response though.

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
NathanQingyangXu: I guess I have some knowledge gap here. We don't have freedom to change public method name in wip/v6?

beikov: Yes and no. We try to do as few changes as possible to avoid unnecessary breakage. We might be able to remove some of the methods in the EntityPersister hierarchy so we might as well just rename some of them to fix grammar issues. I just wanted to bring it up so everyone can express their opinion on this topic.

beikov:
Now that we have a more stable code base, you are very welcome to do some targeted refactorings, but this PR is clearly outdated

PR #3988, opened 2021-05-18
HHH-14616 Optimistic Lock throws org.hibernate.exception.SQLGrammarException: could not retrieve version

dreab8:
https://hibernate.atlassian.net/browse/HHH-14616

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
@@ -166,6 +166,7 @@ public abstract class AbstractEntityPersister
 	private static final CoreMessageLogger LOG = CoreLogging.messageLogger( AbstractEntityPersister.class );
 
 	public static final String ENTITY_CLASS = "class";
+	public static final String VERSION_COLUMN_ALIAS = "v";
 
 	private final NavigableRole navigableRole;
 
@@ -1721,7 +1722,7 @@ public abstract class AbstractEntityPersister
 		SimpleSelect select = new SimpleSelect( getFactory().getDialect() )
 				.setTableName( getVersionedTableName() );
 		if ( isVersioned() ) {
-			select.addColumn( versionColumnName );
+			select.addColumn( getVersionColumnName(), VERSION_COLUMN_ALIAS );
 		}
 		else {
 			select.addColumns( rootTableKeyColumnNames );
@@ -1947,7 +1948,7 @@ public abstract class AbstractEntityPersister
 					if ( !isVersioned() ) {
 						return this;
 					}
-					return getVersionType().nullSafeGet( rs, getVersionColumnName(), session, null );
+					return getVersionType().nullSafeGet( rs, VERSION_COLUMN_ALIAS, session, null );
 				}
 				finally {
 					session.getJdbcCoordinator().getLogicalConnection().getResourceRegistry().release( rs, st );
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
@@ -166,7 +166,7 @@ public abstract class AbstractEntityPersister
 	private static final CoreMessageLogger LOG = CoreLogging.messageLogger( AbstractEntityPersister.class );
 
 	public static final String ENTITY_CLASS = "class";
-	public static final String VERSION_COLUMN_ALIAS = "v";
+	public static final String VERSION_COLUMN_ALIAS = "version_";
 
 	private final NavigableRole navigableRole;
 
--- a/hibernate-core/src/test/java/org/hibernate/test/optlock/OptimisticLockWithGloballyQuotedIdentifierTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/optlock/OptimisticLockWithGloballyQuotedIdentifierTest.java
@@ -1,0 +1,92 @@
+package org.hibernate.test.optlock;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.LockModeType;
+import javax.persistence.Version;
+
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.cfg.Configuration;
+
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class OptimisticLockWithGloballyQuotedIdentifierTest extends BaseCoreFunctionalTestCase {
+
+	@Override
+	protected void configure(Configuration configuration) {
+		configuration.setProperty( AvailableSettings.GLOBALLY_QUOTED_IDENTIFIERS, "true" );
+	}
+
+	@Override
+	protected Class<?>[] getAnnotatedClasses() {
+		return new Class[] { Person.class };
+	}
+
+	@Before
+	public void setUp() {
+		inTransaction(
+				session -> {
+					Person person = new Person( "1", "Fabiana" );
+					session.persist( person );
+				}
+		);
+	}
+
+	@After
+	public void tearDown() {
+		inTransaction(
+				session -> {
+					session.createQuery( "delete from Person" ).executeUpdate();
+				}
+		);
+	}
+
+	@Test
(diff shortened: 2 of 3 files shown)


beikov (approved):
LGTM

hibernate-core/src/test/java/org/hibernate/test/locking/OptimisticLockTest.java
```diff
@@ -0,0 +1,92 @@
...
+		);
+	}
+
+	@Entity(name = "Person")
+	public static class Person {
+		@Id
+		private String id;
+
+		@Version
+		private long version1;
```
NathanQingyangXu, line 64: Could we use 'version' directly if `GLOBALLY_QUOTED_IDENTIFIERS` has been set to true?

hibernate-core/src/test/java/org/hibernate/test/locking/OptimisticLockTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.cfg.Configuration;
+
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class OptimisticLockTest extends BaseCoreFunctionalTestCase {
```
NathanQingyangXu, line 16: I found there is already a similarly-named test class of `org.hibernate.test.optlock.OptimisticLockTest` already. If I didn't misunderstand, this test aims to solve the specific bug (jira annotation is needed?), so maybe we should name the testing class to better reflect its purpose? Maybe we could even move this class into the same package as the existing one?

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -1721,7 +1722,7 @@ public String generateSelectVersionString() {
 		SimpleSelect select = new SimpleSelect( getFactory().getDialect() )
 				.setTableName( getVersionedTableName() );
 		if ( isVersioned() ) {
-			select.addColumn( versionColumnName );
+			select.addColumn( getVersionColumnName(), VERSION_COLUMN_ALIAS );
```
NathanQingyangXu, line 1725: Is there possibility that this magical alias might conflict with existing alias? For instance, the existing alias created by session.createNativeQuery(). If so we might end up with new subtle bug when we closed this jira ticket.
Wild guessing and feel free to ignore my comment.

hibernate-core/src/test/java/org/hibernate/test/locking/OptimisticLockTest.java
dreab8: Thanks @NathanQingyangXu I changed the name of the test and moved it to the `o.h.test.optlock` package. I also changed the name of `version1` to `version`

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -1721,7 +1722,7 @@ public String generateSelectVersionString() {
...
 		}
 		else {
```
gbadner, line 1727: I suspect that there will be a similar failure when an entity is not versioned, but it has an ID column that is quoted. I did look into that, but apparently that use case is broken in a different way (NPE is thrown by `EntityVerifyVersionProcess#doBeforeTransactionCompletion`).

It's OK with me if this other use case is not fixed here.

gbadner (approved):

gbadner:
FYI, the same issue arises when a query is executed setting lock mode to `LockModeType.OPTIMISTIC` using `hibernate.globally_quoted_identifiers=false` (the default), if the version column name is quoted; e.g., `@Column(name = "`version`").

The fix works for this case as well.

I pushed a commit with a test for this use case.

dreab8:
@gbadner, @beikov , @NathanQingyangXu thanks a lot for the the reviews !

gbadner:
@dreab8, can I go ahead and merge, or would you prefer to wait for @sebersole to review?

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -166,6 +166,7 @@
 	private static final CoreMessageLogger LOG = CoreLogging.messageLogger( AbstractEntityPersister.class );
 
 	public static final String ENTITY_CLASS = "class";
+	public static final String VERSION_COLUMN_ALIAS = "v";
```
sebersole, line 169: To make this still unique, but more understandable/readable should we make this alias more like `version_`?  I'd suggest the same for `class` but I'm pretty sure places throughout the codebase refer to this name directly.

sebersole (approved):
Looks good to me

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -1721,7 +1722,7 @@ public String generateSelectVersionString() {
 		SimpleSelect select = new SimpleSelect( getFactory().getDialect() )
 				.setTableName( getVersionedTableName() );
 		if ( isVersioned() ) {
-			select.addColumn( versionColumnName );
+			select.addColumn( getVersionColumnName(), VERSION_COLUMN_ALIAS );
 		}
 		else {
 			select.addColumns( rootTableKeyColumnNames );
```
sebersole, line 1728: Assuming this block ever happens, wouldn't you need to apply the same alias here so the read from ResultSet later can find it?

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
sebersole: That code assumes that the entity is versioned.  This comes back to my comment below that I'm not sure this should be done at all if the entity is not versioned

gbadner: @sebersole, I guess it wasn't clear, but I was referring to the same block you mention. I already tried what you suggested, but it fails as I described above.

<  I did look into that, but apparently that use case is broken in a different way (NPE is thrown by EntityVerifyVersionProcess#doBeforeTransactionCompletion).

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
gbadner: To be clear, 

> I did look into that, but apparently that use case is broken in a different way (NPE is thrown by EntityVerifyVersionProcess#doBeforeTransactionCompletion).

sebersole: Yes, my guess is it fails trying to compare the versions right.  I.e. `entry.getVersion()` returns null, so that `entry.getVersion().equals()`.

My point is that none of this should be happening if the entity is not versioned.  If that is what you mean by "broken in a different way", then sure.

gbadner: Yes, you're correct about the NPE, and that it won't happen if the entity is versioned.

What I meant by "broken in a different way", was that the non-versioned use case is broken by the NPE before being broken by the lack of an alias for a quoted column.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
gbadner: `ENTITY_CLASS` has been set to `"class"` since 2007, and the contant is used in a bunch of places.

gbadner:
@sebersole, is the only change that you are requesting just to change:

`public static final String VERSION_COLUMN_ALIAS = "v";`

to:

`public static final String VERSION_COLUMN_ALIAS = "version_";` ?

If so, I'd like to push a commit that does that and get this merged.

Please confirm.

sebersole:
Yes.

The rest is just discussion of something you brought up - the "no version"
case

dreab8:
just pushed a commit that changes `VERSION_COLUMN_ALIAS` to `version_`

Thanks @gbadner @sebersole @beikov @NathanQingyangXu for the review!

PR #3994, opened 2021-05-19
HHH-14620 Do not initialize collections just to take a snapshot of their size

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14620

As far as I can tell this is safe enough for dirty tracking:

* The collection tracker will return a size of -1 for that collection
* Which is exactly the behavior we currently get after
  `$$_hibernane_clearDirtyCollectionNames` is called if a collection has
  been "retrieved" (getter called) but was not initialized.
* This will mainly prevent some optimizations because we will no longer
  be able to tell whether a collection is "dirty" or not.

I think we should be able to restore those optimizations: for
`PersistentCollection` instances, we would store the "initial" size
inside the collection itself upon initialization,
and we would compare THAT size to the current size in implementations
of `$$_hibernate_areCollectionFieldsDirty` (see
`org.hibernate.bytecode.enhance.internal.bytebuddy.CodeTemplates`).

Alternatively we could store the `CollectionTracker` inside the
`PersistentCollection` so that the collection can update the tracker
upon initialization.

However, that's outside the scope of this bug, that would require
significant testing, and that may cause conflicts with ORM 6, so I won't
do it here.

Relates to https://github.com/quarkusio/quarkus/issues/17291, which will be fixed after we release and upgrade the ORM dependency in Quarkus.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/LazyCollectionLoadingTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/LazyCollectionLoadingTest.java
@@ -91,7 +91,6 @@ public class LazyCollectionLoadingTest extends BaseCoreFunctionalTestCase {
             Parent parent = s.load( Parent.class, parentID );
             assertThat( parent, notNullValue() );
             assertThat( parent, not( instanceOf( HibernateProxy.class ) ) );
-            assertThat( parent, not( instanceOf( HibernateProxy.class ) ) );
             assertFalse( isPropertyInitialized( parent, "children" ) );
             checkDirtyTracking( parent );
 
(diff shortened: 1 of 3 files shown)


hibernate-core/src/test/java/org/hibernate/test/bytecode/enhancement/lazy/LazyCollectionLoadingTest.java
```diff
@@ -101,7 +102,40 @@ public void testTransaction() {
...
+
+    @Test
+    @TestForIssue( jiraKey = "HHH-14620" )
+    public void testTransaction_noProxy() {
+        doInHibernate( this::sessionFactory, s -> {
+            // find will not return a proxy, which is exactly what we want here.
+            Parent parent = s.find( Parent.class, parentID );
+            assertThat( parent, notNullValue() );
+            assertThat( parent, not( instanceOf( HibernateProxy.class ) ) );
+            assertThat( parent, not( instanceOf( HibernateProxy.class ) ) );
```
NathanQingyangXu, line 120: I know this might be copied from existing method, but is it intentional that we repeat the same assertion twice (or do we rely on some side effect)?

yrodiere: Fixed, thanks

beikov (approved):
LGTM

yrodiere:
Merged, thanks

PR #3997, opened 2021-05-20
HHH-14608 Merge causes StackOverflow when JPA proxy compliance is enabled

dreab8:
https://hibernate.atlassian.net/browse/HHH-14608

This is a different solution than that of https://github.com/hibernate/hibernate-orm/pull/3986

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/jpa/internal/PersistenceUnitUtilImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/jpa/internal/PersistenceUnitUtilImpl.java
@@ -68,7 +68,7 @@ public class PersistenceUnitUtilImpl implements PersistenceUnitUtil, Serializabl
 		}
 
 		if ( entity instanceof HibernateProxy ) {
-			return ((HibernateProxy) entity).getHibernateLazyInitializer().getIdentifier();
+			return ((HibernateProxy) entity).getHibernateLazyInitializer().getInternalIdentifier();
 		}
 		else if ( entity instanceof ManagedEntity ) {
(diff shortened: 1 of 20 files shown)


hibernate-core/src/main/java/org/hibernate/jpa/internal/PersistenceUnitUtilImpl.java
```diff
@@ -68,7 +68,7 @@ public Object getIdentifier(Object entity) {
 		}
 
 		if ( entity instanceof HibernateProxy ) {
-			return ((HibernateProxy) entity).getHibernateLazyInitializer().getIdentifier();
+			return ((HibernateProxy) entity).getHibernateLazyInitializer().getInternalIdentifier();
```
gbadner, line 71: l'm not sure about the above change, since the method implements `PersistenceUnitUtil#getIdentifier`. I'll check the spec tomorrow.

gbadner: I'll also run the TCK tonight.

Sanne: Looks like we got stuck here - @gbadner could you suggest how to proceed please?

gbadner: Sorry about that. I ran the TCK with this change and there were no failures.

gbadner (approved):

gbadner:
@dreab8, thanks for fixing this and sorry for the delay getting this merged.

dreab8:
@gbadner thanks for looking at it and merging it :)

PR #3999, opened 2021-05-20
HHH-14618 Allow passing multiple script source files separated by comma

beikov:
https://hibernate.atlassian.net/browse/HHH-14618

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/tool/schema/internal/Helper.java
+++ b/hibernate-core/src/main/java/org/hibernate/tool/schema/internal/Helper.java
@@ -12,6 +12,7 @@ import java.io.Writer;
 import java.net.URL;
 import java.sql.SQLException;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 import org.hibernate.boot.model.relational.Namespace;
 import org.hibernate.boot.registry.classloading.spi.ClassLoaderService;
@@ -24,6 +25,7 @@ import org.hibernate.resource.transaction.spi.DdlTransactionIsolator;
 import org.hibernate.service.ServiceRegistry;
 import org.hibernate.tool.schema.extract.spi.DatabaseInformation;
 import org.hibernate.tool.schema.extract.internal.DatabaseInformationImpl;
+import org.hibernate.tool.schema.internal.exec.ScriptSourceInputAggregate;
 import org.hibernate.tool.schema.internal.exec.ScriptSourceInputFromFile;
 import org.hibernate.tool.schema.internal.exec.ScriptSourceInputFromReader;
 import org.hibernate.tool.schema.internal.exec.ScriptSourceInputFromUrl;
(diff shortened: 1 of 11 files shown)


hibernate-core/src/main/java/org/hibernate/tool/schema/internal/exec/ScriptSourceInputAggregate.java
```diff
@@ -0,0 +1,87 @@
...
+
+import org.hibernate.tool.hbm2ddl.ImportSqlCommandExtractor;
+import org.hibernate.tool.schema.spi.ScriptSourceInput;
+
+/**
+ * A script source input that aggregates over multiple other {@link ScriptSourceInput}.
+ *
+ * @author Christian Beikov
+ */
+public class ScriptSourceInputAggregate implements ScriptSourceInput {
```
NathanQingyangXu, line 21: My comment seems frivolous, but I checked existing name convention for such similar stuff and found we usually adopted 'Aggregatedxxxxx' or 'Aggregatexxxx'.
Feel free to ignore my comment though.

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/Helper.java
```diff
@@ -55,22 +54,38 @@ public static ScriptSourceInput interpretScriptSourceSetting(
...
-			//		1) string URL representation (i.e., "file://...")
-			//		2) relative file path (resource lookup)
-			//		3) absolute file path
-
-			log.trace( "Trying as URL..." );
-			// ClassLoaderService.locateResource() first tries the given resource name as url form...
-			final URL url = classLoaderService.locateResource( scriptSourceSettingString );
-			if ( url != null ) {
-				return new ScriptSourceInputFromUrl( url, charsetName );
+			final String[] paths = scriptSourceSettingString.split( "\\s*,\\s*" );
```
NathanQingyangXu, line 57: Using a compiled Pattern might be much more performant, unless the parameter of String#split() is one char long (JVM have optimization to use the char directly without incurring reg compiling).

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/exec/ScriptSourceInputAggregate.java
```diff
@@ -0,0 +1,87 @@
...
+		for ( int i = 0; i < inputs.length; i++ ) {
+			lists[i] = inputs[i].read( commandExtractor );
+			size += lists[i].size();
+		}
+
+		final List<String> list = new ArrayList<>( size );
+		for ( List<String> strings : lists ) {
+			list.addAll( strings );
+		}
+		return list;
```
NathanQingyangXu, line 80: Seems we prefer memory over performance above; otherwise the following statements suffice:
```
final List<String> list = new ArrayList<>();
for ( int i = 0; i < inputs.length; i++ ) {
    list.addAll( inputs[i].read( commandExtractor ) );
}
return list;
```
Given that the resulting list is usually not pretty big, seems the tradeoff in favour of memory is not that important, IMHO.

dreab8:
Thanks @beikov

PR #4001, opened 2021-05-20
HHH-14624  Oracle from version 12 started supporting the `offset ? rows fetch next ? rows only`syntax for pagination

dreab8:
https://hibernate.atlassian.net/browse/HHH-14624

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/pagination/AbstractLimitHandler.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/pagination/AbstractLimitHandler.java
@@ -8,8 +8,13 @@ package org.hibernate.dialect.pagination;
 
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.hibernate.engine.spi.RowSelection;
+
+import static java.util.regex.Pattern.CASE_INSENSITIVE;
+import static java.util.regex.Pattern.compile;
 
 /**
  * Default implementation of {@link LimitHandler} interface. 
@@ -17,6 +22,10 @@ import org.hibernate.engine.spi.RowSelection;
  * @author Lukasz Antoniak (lukasz dot antoniak at gmail dot com)
  */
 public abstract class AbstractLimitHandler implements LimitHandler {
+
+	private static final Pattern FOR_UPDATE_PATTERN =
+			compile("\\s+for\\s+update\\b|\\s*(;|$)", CASE_INSENSITIVE);
+
 
 	protected AbstractLimitHandler() {
 		// NOP
@@ -176,6 +185,33 @@ public abstract class AbstractLimitHandler implements LimitHandler {
 		}
 		else {
 			return maxRows;
+		}
+	}
+
+	/**
+	 * The offset/limit clauses typically must come
+	 * before the {@code FOR UPDATE}ish clauses, so
+	 * we need a way to identify these clauses in
+	 * the text of the whole query.
+	 */
+	protected Pattern getForUpdatePattern() {
+		return FOR_UPDATE_PATTERN;
+	}
+
+	/**
+	 * Insert a fragment of SQL right before the
+	 * {@code FOR UPDATE}ish clauses at the end
+	 * of the query.
+	 */
+	protected String insertBeforeForUpdate(String limitOffsetClause, String sqlStatement) {
+		Matcher forUpdateMatcher = getForUpdatePattern().matcher( sqlStatement );
+		if ( forUpdateMatcher.find() ) {
+			return new StringBuilder( sqlStatement )
+					.insert( forUpdateMatcher.start(), limitOffsetClause )
+					.toString();
+		}
+		else {
+			return sqlStatement;
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


sebersole (approved):
lgtm

gbadner:
Unfortunately, there is an Oracle failure using this improvement with pessimistic locking as demonstrated by {{org.hibernate.userguide.locking.ExplicitLockingTest#testFollowOnLocking

Without the change, the following query is successful:

{code:sql}
    select
        * 
    from
        ( select
            explicitlo0_.id as id1_0_,
            explicitlo0_."name" as name2_0_ 
        from
            Person explicitlo0_ ) 
    where
        rownum <= ? for update
{code}

With the change, the following SQL fails:

{code:sql}
    select
        explicitlo0_.id as id1_0_,
        explicitlo0_."name" as name2_0_ 
    from
        Person explicitlo0_ fetch first ? rows only for update
{code}

with:

{code:java}
15:14:56,315  WARN SqlExceptionHelper:137 - SQL Error: 2014, SQLState: 42000
15:14:56,315 ERROR SqlExceptionHelper:142 - ORA-02014: cannot select FOR UPDATE from view with DISTINCT, GROUP BY, etc.
{code}

hibernate-core/src/main/java/org/hibernate/dialect/pagination/AbstractLimitHandler.java
```diff
@@ -178,4 +187,31 @@ protected final int getMaxOrLimit(RowSelection selection) {
...
+		Matcher forUpdateMatcher = getForUpdatePattern().matcher( sqlStatement );
+		if ( forUpdateMatcher.find() ) {
+			return new StringBuilder( sqlStatement )
+					.insert( forUpdateMatcher.start(), limitOffsetClause )
+					.toString();
+		}
+		else {
+			return sqlStatement;
+		}
+	}
```
NathanQingyangXu, line 216: Is RegExp approach fool-proof? For instance:
```
select a from t where b = 'for update';
```
Seems not

hibernate-core/src/main/java/org/hibernate/dialect/pagination/AbstractLimitHandler.java
```diff
@@ -8,16 +8,25 @@
...
+
 /**
  * Default implementation of {@link LimitHandler} interface. 
  *
  * @author Lukasz Antoniak (lukasz dot antoniak at gmail dot com)
  */
 public abstract class AbstractLimitHandler implements LimitHandler {
 
+	private static final Pattern FOR_UPDATE_PATTERN =
+			compile("\\s+for\\s+update\\b|\\s*(;|$)", CASE_INSENSITIVE);
```
NathanQingyangXu, line 27: Will the above pattern matches any string for any string will match the `\\s*(;|$)`? Seems the following is correct:
```
compile("\\s+for\\s+update\b", CASE_INSENSITIVE);
```
Also see my other comment regarding the robustness of such string pattern matching.

hibernate-core/src/main/java/org/hibernate/dialect/pagination/AbstractLimitHandler.java
```diff
@@ -178,4 +187,31 @@ protected final int getMaxOrLimit(RowSelection selection) {
...
+
+	/**
+	 * Insert a fragment of SQL right before the
+	 * {@code FOR UPDATE}ish clauses at the end
+	 * of the query.
+	 */
+	protected String insertBeforeForUpdate(String limitOffsetClause, String sqlStatement) {
+		Matcher forUpdateMatcher = getForUpdatePattern().matcher( sqlStatement );
+		if ( forUpdateMatcher.find() ) {
+			return new StringBuilder( sqlStatement )
```
NathanQingyangXu, line 209: by default, StringBuilder will allocate 16 additional chars after the constructor string parameter. It is very likely the 16 chars are enough for 'limitOffsetClaus'. To avoid StringBuilder resizing, the following approach might work:
```
return new StringBuilder( sqlStatement.size() + limitOffsetClause.size() ).append( sqlStatement )...;
```
But it sacrifices code readability a little bit, so it might not be worthwhile though. Not very sure here.

hibernate-core/src/main/java/org/hibernate/dialect/pagination/AbstractLimitHandler.java
```diff
@@ -178,4 +187,31 @@ protected final int getMaxOrLimit(RowSelection selection) {
...
+					.insert( forUpdateMatcher.start(), limitOffsetClause )
```
NathanQingyangXu, line 210: I think we need to. use `" " + limitOffsetClause` to avoid the bug that there would be no whitespace delimiter between the preceding text and 'limitOffsetClause'.

hibernate-core/src/main/java/org/hibernate/dialect/pagination/AbstractLimitHandler.java
```diff
@@ -178,4 +187,31 @@ protected final int getMaxOrLimit(RowSelection selection) {
...
+					.toString();
+		}
+		else {
+			return sqlStatement;
```
NathanQingyangXu, line 214: Did you mean:
```
else {
    return sqlStatement + ' ' + limitOffsetClause;
}
```
?  If so we still need to attend to the edge case that 'sqlStatement' might end with ';' and we might need to remove it before appending.

NathanQingyangXu, line 214: Is there any possibility the above else block will be run? It seems the RegExp will match every possible string invariably.

dreab8:
Thanks @gbadner and @NathanQingyangXu for the review, I'm reworking the PR to fix the issues.

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/4005

PR #4002, opened 2021-05-20
Added ability to specify @Where annotation for JoinedSubclass

DariaMitrofanovna:
JIRA issue: HHH-14626

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/persister/entity/WhereJoinedSubclassTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/persister/entity/WhereJoinedSubclassTest.java
@@ -1,0 +1,66 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.persister.entity;
+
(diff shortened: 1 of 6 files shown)


hibernate-core/src/test/java/org/hibernate/persister/entity/WhereJoinedSubclassTest.java
```diff
@@ -0,0 +1,66 @@
...
+	@Test
+	public void testSchemaNotReplacedInCustomSQL() throws Exception {
+
+		String className = ChildEntity.class.getName();
+
+		final AbstractEntityPersister persister = (AbstractEntityPersister) sessionFactory().getEntityPersister(
+				className );
+		assertEquals( persister.getSQLWhereString( "alias" ), "( alias.id = 1) " );
+		assertEquals( persister.getSqlJoinedSubclassWhereString( "alias" ), "( alias.age = 1) " );
+		assertEquals( persister.filterFragment( "alias" ), " and ( alias_1_.id = 1)  and ( alias.age = 1) " );
```
NathanQingyangXu, line 44: It is a common misuse of 'assertEquals()' by replacing:
```
assertEquals( expected, actual );
```
with
```
assertEquals( actual, expected );
```
The misuse can only be noticed when assertion failed ending up with a confusing message.
```

DariaMitrofanovna: Thanks, it's fixed in next commit

PR #4004, opened 2021-05-23
HHH-14631 add Sakila sample database as new testing domain model

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14631

Sakila is a well-known MySQL sample database modelling dvd rental store: https://dev.mysql.com/doc/sakila/en/
A HR example database in Oracle is also included.

We might consider other sample databases as per https://data-xtractor.com/blog/data-modeling/example-databases/.

### DIFF
--- a/hibernate-testing/src/main/java/org/hibernate/testing/orm/domain/StandardDomainModel.java
+++ b/hibernate-testing/src/main/java/org/hibernate/testing/orm/domain/StandardDomainModel.java
@@ -10,7 +10,9 @@ import org.hibernate.testing.orm.domain.animal.AnimalDomainModel;
 import org.hibernate.testing.orm.domain.contacts.ContactsDomainModel;
 import org.hibernate.testing.orm.domain.gambit.GambitDomainModel;
 import org.hibernate.testing.orm.domain.helpdesk.HelpDeskDomainModel;
+import org.hibernate.testing.orm.domain.humanresource.HumanResourceDomainModel;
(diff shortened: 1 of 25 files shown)


beikov (approved):
Great work, thanks!

sebersole:
Do we need to do anything special here with copyrights and/or licenses?

NathanQingyangXu:
> Do we need to do anything special here with copyrights and/or licenses?

I have no idea in this respect. @beikov , @dreab8 @Sanne how do you think?

beikov:
I don't know for sure, but other projects that provide a Sakila sample database provide it under the BSD-2 license which seems to be the original license of the MySQL example model, though these projects provide the actual SQL DDL and Hibernate would just emit compatible SQL DDL. Maybe @Sanne knows more, but from my understanding, it should be be put under BSD-2 license.

sebersole:
We'd definitely want to investigate the implication of BSD on LGPL and make it abundantly clear that the BSD only applies to those classes

Sanne:
@NathanQingyangXu how did you create these classes? I assume via reverse engineering from the actual DB?

NathanQingyangXu:
Nope, I created these classes purely based on the open documentations.

Sanne:
Ok then they are not BSD licensed, and you can choose to contribute them to the Hibernate project as LGPL (like you did in the headers), so it should be all good

PR #4005, opened 2021-05-23
HHH-14624 HHH-14624 Oracle from version 12 started supporting the offset ? rows fetch next ? rows onlysyntax for pagination

dreab8:
https://hibernate.atlassian.net/browse/HHH-14624

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Oracle12cDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Oracle12cDialect.java
@@ -10,6 +10,9 @@ import org.hibernate.boot.model.TypeContributions;
 import org.hibernate.cfg.Environment;
 import org.hibernate.dialect.identity.IdentityColumnSupport;
 import org.hibernate.dialect.identity.Oracle12cIdentityColumnSupport;
+import org.hibernate.dialect.pagination.AbstractLimitHandler;
+import org.hibernate.dialect.pagination.LimitHandler;
+import org.hibernate.dialect.pagination.Oracle12LimitHandler;
 import org.hibernate.engine.config.spi.ConfigurationService;
 import org.hibernate.engine.config.spi.StandardConverters;
 import org.hibernate.service.ServiceRegistry;
@@ -23,6 +26,8 @@ import org.hibernate.type.WrappedMaterializedBlobType;
  */
 public class Oracle12cDialect extends Oracle10gDialect {
 	public static final String PREFER_LONG_RAW = "hibernate.dialect.oracle.prefer_long_raw";
+
+	private static final AbstractLimitHandler LIMIT_HANDLER = Oracle12LimitHandler.INSTANCE;
 
 	public Oracle12cDialect() {
 		super();
@@ -61,5 +66,10 @@ public class Oracle12cDialect extends Oracle10gDialect {
 	@Override
 	public IdentityColumnSupport getIdentityColumnSupport() {
 		return new Oracle12cIdentityColumnSupport();
+	}
+
+	@Override
+	public LimitHandler getLimitHandler() {
+		return LIMIT_HANDLER;
 	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/dialect/pagination/LimitHandler.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/pagination/LimitHandler.java
@@ -9,6 +9,7 @@ package org.hibernate.dialect.pagination;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 
+import org.hibernate.engine.spi.QueryParameters;
 import org.hibernate.engine.spi.RowSelection;
 
 /**
@@ -37,12 +38,24 @@ public interface LimitHandler {
 	/**
 	 * Return processed SQL query.
 	 *
-     * @param sql       the SQL query to process.
+     * @param sql the SQL query to process.
      * @param selection the selection criteria for rows.
      *
 	 * @return Query statement with LIMIT clause applied.
 	 */
 	String processSql(String sql, RowSelection selection);
+
+	/**
+	 * Return processed SQL query.
+	 *
+	 * @param sql the SQL query to process.
+	 * @param queryParameters the queryParameters.
+	 *
+	 * @return Query statement with LIMIT clause applied.
+	 */
+	default String processSql(String sql, QueryParameters queryParameters ){
+		return processSql( sql, queryParameters.getRowSelection() );
+	}
 
 	/**
 	 * Bind parameter values needed by the LIMIT clause before original SELECT statement.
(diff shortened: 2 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/Oracle12cDialect.java
```diff
@@ -62,4 +67,9 @@ public String getNativeIdentifierGeneratorStrategy() {
 	public IdentityColumnSupport getIdentityColumnSupport() {
 		return new Oracle12cIdentityColumnSupport();
 	}
+
+	@Override
+	public LimitHandler getLimitHandler() {
+		return LIMIT_HANDLER;
```
NathanQingyangXu, line 73: Seems 'LIMIT_HANDLER' is only used above once. Why not use 'Oracle12LimitHndler.INSTANCE' directly without an intermediary?

hibernate-core/src/main/java/org/hibernate/dialect/pagination/LimitHandler.java
```diff
@@ -44,6 +45,18 @@
 	 */
 	String processSql(String sql, RowSelection selection);
 
+	/**
+	 * Return processed SQL query.
+	 *
+	 * @param sql       the SQL query to process.
+	 * @param queryParameters the queryParameters.
```
NathanQingyangXu, line 52: Super minor. It would be great to align both descriptions above vertically.

hibernate-core/src/main/java/org/hibernate/dialect/pagination/Oracle12LimitHandler.java
```diff
@@ -0,0 +1,170 @@
...
+		if ( forUpdateLastIndex > -1 ) {
+			if ( lastIndexOfQuote == -1 ) {
+				return forUpdateLastIndex;
+			}
+			if ( lastIndexOfQuote > forUpdateLastIndex ) {
+				return -1;
+			}
+			return forUpdateLastIndex;
+		}
+		return forUpdateLastIndex;
```
NathanQingyangXu, line 153: The above code handles the example case perfectly but is it perfectly enough without ANTLR parsing? IMHO, v6's ANTLR parsing result is the only reliable way to be robust. Maybe this is enough for v5 but when we merge this PR to v6, we might at least leave some TODO comment to remind us to implement it in better way.
Not 100% sure about my remarks, though.

hibernate-core/src/main/java/org/hibernate/dialect/pagination/Oracle12LimitHandler.java
```diff
@@ -0,0 +1,170 @@
...
+		if ( isForUpdate ) {
+			pagingSelect.append( " " );
+			pagingSelect.append( forUpdateClause );
+		}
+
+		return pagingSelect.toString();
+	}
+
+	private int getForUpdateIndex(String sql) {
+		final int forUpdateLastIndex = sql.toLowerCase( Locale.ROOT ).lastIndexOf( "for update" );
```
NathanQingyangXu, line 141: Is there such possibility that the sql string was provided by user and it happens to have two empty chars between 'for' and 'update' (or it doesn't use single quote but double quote)?

hibernate-core/src/main/java/org/hibernate/dialect/Oracle12cDialect.java
dreab8: this method is used by `Loader` and `AbstractLoadPlanBasedLoader`

```
final LimitHandler limitHandler = getFactory().getDialect().getLimitHandler();
```

hibernate-core/src/main/java/org/hibernate/dialect/pagination/Oracle12LimitHandler.java
dreab8: not sure I have understood your point, here we are trying to manage the oracle specific sql code needed to the limit ResultSet

hibernate-core/src/main/java/org/hibernate/dialect/pagination/LimitHandler.java
dreab8: Thanks, I'm going to fix this

hibernate-core/src/main/java/org/hibernate/dialect/pagination/Oracle12LimitHandler.java
dreab8: I think the point related to the spaces between `for` and `update` is true while the single and double quote I think it does not matter here

hibernate-core/src/main/java/org/hibernate/dialect/Oracle12cDialect.java
NathanQingyangXu: My question was not about the public method, but about its internal implementation. Seems we can change its impl to:
```
@Override
public LimitHandler getLimitHandler() {
    return Oracle12LimitHandler.INSTANCE;
}

```
and we can delete the private static variable. It is just a little bit confusing about its necessity.
Feel free to ignore my comment though. Not a big deal.

hibernate-core/src/main/java/org/hibernate/dialect/pagination/Oracle12LimitHandler.java
NathanQingyangXu: I simply gave some example regarding the robustness concern of such SQL string simple tweaking. The possibilities might be much more than we can expect now and ANTLR or semantic clue might be the only reliable way to be fool-proof. Yeah, we have to process a SQL string here, but in v6 we might face a semantic representation and then we can solve the issue in more elegant way.

hibernate-core/src/main/java/org/hibernate/dialect/pagination/Oracle12LimitHandler.java
NathanQingyangXu: My point is SQL string tweaking has its limit and we can easily think of bad cases to defeat our simple string processing. For instance, is it possible that 'do update' might appear in some SQL comment in the middle? Just an example but if you check the SQL grammar, it is not easy to cover all the possibilities. That is one of the selling points of v6: we rely on ANTLR or semantic presentation in favour of SQL string tweaking. 
Feel free to ignore my comment, as always.

NathanQingyangXu:
super minor, but it seems we duplicate the JIRA ticket id twice in the title.

NathanQingyangXu:
Great job! This PR is much nicer than the superseded one. I didn't approve the PR simply because I don't feel qualified due to my knowledge gap, but I am very happy my concerns have been attended to. Thanks.

NathanQingyangXu (approved):
Much nicer than the superseded PR. Some robustness concern is not restrictive to this PR but to the whole v5. Great job!

hibernate-core/src/main/java/org/hibernate/dialect/Oracle12cDialect.java
dreab8: sorry I haven't understood, I'm going to change it, thanks

beikov (approved):

sebersole (approved):

PR #4008, opened 2021-05-25
HHH-14635 Upgrade to latest JUnit and to Log4j 2

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14635


sebersole (approved):

hibernate-testing/src/main/java/org/hibernate/testing/logger/TestableLoggerProvider.java
```diff
@@ -44,57 +46,67 @@ public Logger getLogger(final String name) {
...
+
+	@Override
 	public void removeMdc(String key) {
-		MDC.remove( key );
+		ThreadContext.remove( key );
 	}
 
+	@Override
+	public Map<String, Object> getMdcMap() {
+		return new HashMap<>( ThreadContext.getImmutableContext() );
```
NathanQingyangXu, line 74: There are some subtle differences after the above refactoring. For instance, the following assertion would fail due to additional map would be created for 'getMdcMap()' rather than the underlining map per se:
```
Map<String, Object> mdc = getMdcMap();
mdc.put(key, value);
assertEquals(value, getMdc(key));
```
Also the following assertion would fail due to the value string format transformation.
```
putMdc(key, 10);
assertEquals(10, getMdc(key));
```
I would suggest being more cautious about the code refactoring, though not clear about how.

yrodiere: As stated above:

> 	// Maintainer note:
>	// Except the next method, which is adjusted to return our own Log4DelegatingLogger
>	// this class is a verbatim copy of org.jboss.logging.Log4jLoggerProvider

I just changed this class to be a verbatim copy of `Log4j2LoggerProvider` instead of `Log4jLoggerProvider`. I don't have anything to do with the content of these methods. I'll trust the JBoss logging developers :)

yrodiere:
Merged, thanks!

PR #4009, opened 2021-05-25
Lock handling support

beikov:
* Move locking tests to orm package
* Implement applying locks in SqlAstTranslator and thus make queries that have locks cacheable
* Implement follow on locking
* Update H2 to 1.4.200 which is necessary because some lock tests didn't work with 1.4.199
* Replace `LockMode` in SqlAst and `Fetch`/`DomainResult` with _source alias_ to resolve lock modes right before execution
* Add CTE materialization option to force materialization of the ID select CTE for certain PostgreSQL versions
* Acquire locks for ID select queries if possible
* New `Dialect` methods `getWriteLockRowIdentifier()`, `getReadLockRowIdentifier()` as replacement for `forUpdateOfColumns`
* Add ability to mark function as aggregate and allow SQL FILTER clause being attached
** Needed to determine whether follow on locking is required
* Capture whether a function is an aggregate in descriptor and introduce `SelfRenderingSqmAggregateFunction` for rendering SQL FILTER clause


hibernate-core/src/main/java/org/hibernate/LockOptions.java
```diff
@@ -363,4 +371,14 @@ else if ( this == that ) {
...
 
+	@Override
+	public int hashCode() {
+		int result = lockMode != null ? lockMode.hashCode() : 0;
+		result = 31 * result + timeout;
+		result = 31 * result + ( aliasSpecificLockModes != null ? aliasSpecificLockModes.hashCode() : 0 );
+		result = 31 * result + ( followOnLocking != null ? followOnLocking.hashCode() : 0 );
+		result = 31 * result + ( scope ? 1 : 0 );
+		return result;
+	}
```
NathanQingyangXu, line 382: It would be great to use **Objects** static util methods to simplify, if only from code readability perspective.

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
```diff
@@ -354,4 +358,151 @@ public LimitHandler getLimitHandler() {
...
+
+	@Override
+	public boolean supportsOuterJoinForUpdate() {
+		return false;
+	}
+
+	@Override
+	public boolean supportsNoWait() {
+		return getVersion() >= 2010;
+	}
```
NathanQingyangXu, line 487: maybe we can reuse this method in other methods to centralize the logic.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -3515,6 +3534,15 @@ public boolean supportsNoWait() {
 		return false;
 	}
 
+	/**
+	 * Does this dialect/database support WAIT timeout.
```
NathanQingyangXu, line 3538: super minor, but a '?' seems better alternative at the end of sentence.

PR #4010, opened 2021-05-25
Wip/6.0 re-enabled many to many tests

dreab8:
it seems HHH-11144 is resolved in 6 :)


hibernate-core/src/main/java/org/hibernate/engine/spi/BatchFetchQueue.java
```diff
@@ -352,7 +352,15 @@ else if ( !isCached( ce.getLoadedKey(), collectionPersister ) ) {
 				}
 			}
 		}
-		return keys; //we ran out of keys to try
+		return resize( keys, i ); //we ran out of keys to try
+	}
+
+	private Object[] resize(Object[] keys, int size) {
```
beikov, line 358: What is the purpose of this method? In 5.x we didn't need this either so I am a bit surprised that we need this now.

beikov (commented):
LGTM except for the changes in `BatchFetchQueue`

hibernate-core/src/main/java/org/hibernate/engine/spi/BatchFetchQueue.java
NathanQingyangXu: I think we can reuse the `Arrays.copyOf(T[] original, int newLength)` from JDK without bothering with creating a new method, if I didn't misunderstand your purpose.

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/orphan/onetomany/DeleteSharedOneToManyOrphansTest.java
```diff
@@ -78,7 +78,6 @@ public void cleanupTest(EntityManagerFactoryScope scope) throws Exception {
 
 	@Test
 	@TestForIssue(jiraKey = "HHH-11144")
-	@FailureExpected( jiraKey = "HHH-11144" )
```
NathanQingyangXu, line 81: It is such a wonderful feeling to get rid of such stuff, :).

hibernate-core/src/main/java/org/hibernate/engine/spi/BatchFetchQueue.java
dreab8: removed it

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/orphan/onetomany/DeleteSharedOneToManyOrphansTest.java
dreab8: unfortunately it was solved but because a wrong side effect introduced by the `seize` method i introduced, after removing the method it the issue  is back :(

PR #4013, opened 2021-05-26
HHH-14513 5.3 

dreab8:
Blanket jira for backporting:
https://hibernate.atlassian.net/browse/HHH-14640

Includes the following:
https://hibernate.atlassian.net/browse/HHH-13685
https://hibernate.atlassian.net/browse/HHH-13689
https://hibernate.atlassian.net/browse/HHH-13925
https://hibernate.atlassian.net/browse/HHH-13682
https://hibernate.atlassian.net/browse/HHH-14030
https://hibernate.atlassian.net/browse/HHH-14083
https://hibernate.atlassian.net/browse/HHH-14247
https://hibernate.atlassian.net/browse/HHH-14111
https://hibernate.atlassian.net/browse/HHH-14144
https://hibernate.atlassian.net/browse/HHH-14199
https://hibernate.atlassian.net/browse/HHH-14283
https://hibernate.atlassian.net/browse/HHH-14315
https://hibernate.atlassian.net/browse/HHH-14513

this PR also contains:

- the upgrade to Gradle 6.7.1
- changes to automate the release process


yrodiere (commented):
LGTM, though you might want to change the HHH- prefix of the commits to match yours (I think that's what Gail usually does?)

I also checked the generated OSGi instructions in `MANIFEST.MF`; here's the diff:
https://gist.github.com/yrodiere/234146423b45654cb3e5266c951f6353

As far as I can tell we have the same (expected) differences as in #3080: 

> * In `hibernate-core`, there are two new imported packages (`Import-Package` header): `com.ibm.websphere.jtaextensions` and `org.objectweb.jonas_tm`. The old plugin didn't detect the dependencies, but the new one does: we load classes from these packages dynamically, if they are present, in the transaction manager code. I made the imports optional and removed the version constraints, since I'm not sure what the version constraints are.
> * In `hibernate-core`, there is a new `Private-Package` header that references two packages: `org.hibernate.xsd.cfg` and `org.hibernate.xsd.mapping`. From what I can tell, those packages only contain resources (XSD files). They weren't exported in 5.4.7.Final. The `Private-Package` header is just a byproduct of BND and does not change anything in OSGi runtimes (it's ignored).
> * In `hibernate-gradle-plugin`, the `Import-Package`s for the Groovy APIs no longer point to a specific version (the "version" bit was removed). As far as I can tell, we don't have an explicit dependency to a specific version of Groovy in our own gradle files, so I think it makes sense? I can add a version manually, but I would need to know which version we want to target. Or, you know, we just leave it as is: I might be wrong, but running a Gradle plugin in an OSGi environment looks pretty niche.
> * All OSGi instructions were removed from `javadoc` and `sources` JARs. Apparently they were copied there by mistake. If we wanted to restore them, it would be quite complicated with the new plugin.

build.gradle
```diff
@@ -55,18 +89,18 @@ task release {
 			"the fact that subprojects will appropriately define a release task " +
 			"themselves if they have any release-related activities to perform"
 
-	// Force to release with JDK 8. Releasing with JDK 11 is not supported yet:
-	// - the hibernate-orm-modules tests do not run due to an issue with the ASM version currently used by Gradle
+	// Force to release with JDK 8. It used to not work on JDK11 because of the hibernate-orm-modules module,
+	// but this limitation might be resolved now that this module has been deleted?
```
gbadner, line 93: I have no idea if this matters, but Hibernate ORM 5.3 still has the hibernate-orm-modules tests.

In any case, JDK 1.8 will be used for releasing.

gbadner: I'll update the comment in a new commit.

gradle/published-java-module.gradle
```diff
@@ -10,70 +10,91 @@ apply from: rootProject.file( 'gradle/java-module.gradle' )
...
+		attributes(
+				// Basic JAR manifest attributes
+				'Specification-Title': project.name,
+				'Specification-Version': project.version,
+				'Specification-Vendor': 'Hibernate.org',
+				'Implementation-Title': project.name,
+				'Implementation-Version': project.version,
+				'Implementation-Vendor': 'Hibernate.org',
+				'Implementation-Vendor-Id': 'org.hibernate',
+				'Implementation-Url': 'http://hibernate.org/orm',
```
gbadner, line 27: Should the above be:

`'Implementation-Url': 'https://hibernate.org/orm',`

gradle/published-java-module.gradle
```diff
@@ -10,70 +10,91 @@ apply from: rootProject.file( 'gradle/java-module.gradle' )
...
+
+				// Hibernate-specific JAR manifest attributes
+				'Hibernate-VersionFamily': project.ormVersion.family,
+				'Hibernate-JpaVersion': project.jpaVersion.name,
+
+				// BND Plugin instructions (for OSGi):
+				'Bundle-Name': project.name,
+				'Bundle-SymbolicName': project.java9ModuleName,
+				'Bundle-Vendor': 'Hibernate.org',
+				'Bundle-DocURL': "http://www.hibernate.org/orm/${project.ormVersion.family}",
```
gbadner, line 40: Chrome says `http://www.hibernate.org/` and `https://www.hibernate.org/` are not secure.

Should the above be:

`'Bundle-DocURL': "https://hibernate.org/orm/${project.ormVersion.family}` ?

gradle/publishing-repos.gradle
```diff
@@ -5,20 +5,11 @@
  * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
```
gbadner, line 5: I'm afraid to ask, but should `http://www.gnu.org/licenses/lgpl-2.1.html` be changed to `https://www.gnu.org/licenses/lgpl-2.1.html`?

gradle/java-module.gradle
```diff
@@ -334,6 +340,10 @@ task forbiddenApisSystemOut(type: CheckForbiddenApis, dependsOn: compileJava) {
 	targetCompatibility = project.forbiddenAPITargetJDKCompatibility
 	bundledSignatures += 'jdk-system-out'
 	suppressAnnotations += ['org.hibernate.internal.build.AllowSysOut', 'org.hibernate.internal.build.AllowPrintStacktrace']
+
+	// This slows down the checks a little, but is necessary to avoid the gradle deamon holding on
+	// to class definitions loaded previously - even possibly in a previous build.
+	disableClassloadingCache = true
```
scottmarlow, line 346: :+1:

gradle/published-java-module.gradle
dreab8: i think so

dreab8:
Thanks @yrodiere for great help with this PR!

gradle/publishing-repos.gradle
yrodiere: We have the same problem in `main`. Probably worth updating there first, and then backporting to 5.4 and 5.3?

Though that's really not critical, it's just a comment :)

gradle/published-java-module.gradle
yrodiere: The `http` value is the one we had before these changes.

We can change it, but that would be changing the content of the manifest files.
Not sure what the impact would be, that's why I kept this value at the time.

Also, if we change this, better do it on the `main` branch too (and maybe 5.4).

gradle/published-java-module.gradle
yrodiere: Same as the other comments: if we change this, better do it on the `main` branch too (and maybe 5.4).

yrodiere:
Thanks to you for doing the actual work :)

gbadner:
> LGTM, though you might want to change the HHH- prefix of the commits to match yours (I think that's what Gail usually does?)
> 

I'm happy to edit the commit messages that are missing the "HHH-" prefix. 

IMO, it's not worth the effort to try to figure out the original jira issue for those commits (if there was one). The easiest thing to do would be to create a new jira issue for backporting this work to 5.3. I plan to reference the new jira issue in the edited commit comments for those commits that don't already reference an HHH issue.

@yrodiere, @dreab8, does that sound good to you?

scottmarlow:
Does this PR change the MANIFEST.MF contents to use different values than before?

gradle/published-java-module.gradle
gbadner: @yrodiere, I was just going to check to see what it said in main and 5.4 branches.

I'll create a new jira issue to change "http://" to "https://". Should I target 5.5?

For now, I won't plan to backport to 5.3.

scottmarlow:
> LGTM, though you might want to change the HHH- prefix of the commits to match yours (I think that's what Gail usually does?)
> 
> I also checked the generated OSGi instructions in `MANIFEST.MF`; here's the diff:
> https://gist.github.com/yrodiere/50b0bb95a44f650a41099e8de1f9b867
> 
> As far as I can tell we have the same (expected) differences as in #3080:
> 
> > * In `hibernate-core`, there are two new imported packages (`Import-Package` header): `com.ibm.websphere.jtaextensions` and `org.objectweb.jonas_tm`. The old plugin didn't detect the dependencies, but the new one does: we load classes from these packages dynamically, if they are present, in the transaction manager code. I made the imports optional and removed the version constraints, since I'm not sure what the version constraints are.

I'm not sure if adding optional dependencies on `com.ibm.websphere.jtaextensions` + `org.objectweb.jonas_tm` could cause (surprising) problems in some environments that might find those but previously didn't.  

For example, could `com.ibm.websphere.jtaextensions` cause Hibernate Sessions to auto join JTA transactions automatically in an environment that includes `com.ibm.websphere.jtaextensions` where as previously without this change that didn't happen?

Is it easier to exclude these new dependencies instead of answering my question for this pr?

> > * In `hibernate-core`, there is a new `Private-Package` header that references two packages: `org.hibernate.xsd.cfg` and `org.hibernate.xsd.mapping`. From what I can tell, those packages only contain resources (XSD files). They weren't exported in 5.4.7.Final. The `Private-Package` header is just a byproduct of BND and does not change anything in OSGi runtimes (it's ignored).
> > * In `hibernate-gradle-plugin`, the `Import-Package`s for the Groovy APIs no longer point to a specific version (the "version" bit was removed). As far as I can tell, we don't have an explicit dependency to a specific version of Groovy in our own gradle files, so I think it makes sense? I can add a version manually, but I would need to know which version we want to target. Or, you know, we just leave it as is: I might be wrong, but running a Gradle plugin in an OSGi environment looks pretty niche.
> > * All OSGi instructions were removed from `javadoc` and `sources` JARs. Apparently they were copied there by mistake. If we wanted to restore them, it would be quite complicated with the new plugin.

gbadner:
> IMO, it's not worth the effort to try to figure out the original jira issue for those commits (if there was one). The easiest thing to do would be to create a new jira issue for backporting this work to 5.3. I plan to reference the new jira issue in the edited commit comments for those commits that don't already reference an HHH issue.
> 
> @yrodiere, @dreab8, does that sound good to you?

I've created https://hibernate.atlassian.net/browse/HHH-14640

gradle/published-java-module.gradle
gbadner: I've created https://hibernate.atlassian.net/browse/HHH-14641 to change  "http://" to "https://".

yrodiere:
> > IMO, it's not worth the effort to try to figure out the original jira issue for those commits (if there was one). The easiest thing to do would be to create a new jira issue for backporting this work to 5.3. I plan to reference the new jira issue in the edited commit comments for those commits that don't already reference an HHH issue.
> > @yrodiere, @dreab8, does that sound good to you?
> 
> I've created https://hibernate.atlassian.net/browse/HHH-14640

Sounds perfect, thanks.

> For example, could `com.ibm.websphere.jtaextensions` cause Hibernate Sessions to auto join JTA transactions automatically in an environment that includes `com.ibm.websphere.jtaextensions` where as previously without this change that didn't happen?

No idea. All I can say is that I'm not aware of any report of this problem in the ~1 year that we've released 5.4 with this patch.

> Is it easier to exclude these new dependencies instead of answering my question for this pr?

It's easier. Done. @dreab8 , @gbadner, please note that I added a commit just for this.

gbadner:
Superseded by https://github.com/hibernate/hibernate-orm/pull/4030.

PR #4018, opened 2021-05-28
wip/6.0 HHH-14642 verify/test Criteria  - basic package

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14642

This is the first PR under the big umbrella of 'Criteria verification'. The goal is to create a bunch of PRs incrementally to make code review possible.

During the migration of testing cases, only working testing methods are relocated whereas the broken ones are left in the old location. For that reason, quite some testing classes still remain in 'org.hibernate.jpa.test.criteria', though some of their methods might have been migrated.

This PR aims to ensure some pretty basic features of Criteria. The following issues were flushed out:

- org.hibernate.query.sqm.tree.jpa.**ParameterCollector** was the ANTLR tree walker that aims to collect parameters for Criteria exclusively, but it is by no means complete. It was enriched to cover the migrated testing cases in this PR (still far from being complete and is meant to be enriched further in later PRs) and was renamed to more appropriate **CriteriaParameterCollectQueryWalker** to differentiate it from another org.hibernate.query.sqm.internal.**ParameterCollector** interface (see https://github.com/hibernate/hibernate-orm/pull/4018/files#r642723735);
- we often feel the need to resolve **MappingModelExpressable** from some sqm node. For that purpose, we have the following 3 methods to achieve similar goals: 
1) org.hibernate.query.sqm.sql.**BaseSqmToSqlAstConverter**#determineValueMapping(); 
2) org.hibernate.query.sqm.internal.SqmMappingModelHelper#resolveMappingModelExpressable(); 
3) org.hibernate.metamodel.model.domain.internal.MappingMetamodelImpl#resolveMappingExpressable(). 
In this PR the former two are routed to the third one, which also can handle BasicSqmPathSource, which is super important for Criteria's parameter mapping type resolution (see https://github.com/hibernate/hibernate-orm/pull/4018/files#r642724535);
- fix the nasty overloading resolution error in org.hibernate.query.sqm.NodeBuilder#in() methods (see https://github.com/hibernate/hibernate-orm/pull/4018/files#r642723129);
- revert back the new implementation in v6 which render **false** for in list predicate with empty list. The resulting 'where false' would be broken in quite some dbs (Oracle and MSSQL etc.); this PR reverted it back to the previous approach to render it as either '**in ()**' or '**in (null)**' based on Dialect's **supportsEmptyInList**() feature (see https://github.com/hibernate/hibernate-orm/pull/4018/files#r642726177);
- many other minor bug fixes or improvements.


hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AbstractSqmExpression.java
```diff
@@ -125,7 +125,7 @@ public SqmPredicate in(Expression<?>... values) {
 
 	@Override
 	public SqmPredicate in(Collection<?> values) {
-		return nodeBuilder().in( this, values );
+		return nodeBuilder().in( this, (Collection) values );
 	}
```
NathanQingyangXu, line 129: This is inspiring and error-prone. Without the above seemingly ugly fix, the following statement:
```
@Override
public SqmPredicate in(Collection<?> values) {
	return nodeBuilder().in( this, values );
}
```
would be routed to:
```
@Override
<T> SqmInPredicate<T> in(Expression<? extends T> expression, T... values);
```
not the intended alternative which accepts collection parameter:
```
@Override
<T> SqmInPredicate<T> in(Expression<? extends T> expression, Collection<T> values);
```
You can confirm this by following the link in IDE by mouse clicking. The end result is disastrous and the root cause is our imperfect usage of Java generics.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/jpa/CriteriaParameterCollectQueryWalker.java
```diff
@@ -22,24 +25,25 @@
...
 
 /**
- * todo (6.0) : how is this different from {@link org.hibernate.query.sqm.internal.ParameterCollector}?
+ * A criteria parameter collector to feed parameter collected to a {@link Consumer}
+ * while walking a Semantic Query Model (SQM) tree.
  *
  * @author Steve Ebersole
  */
-public class ParameterCollector extends BaseSemanticQueryWalker {
+public class CriteriaParameterCollectQueryWalker extends BaseSemanticQueryWalker {
```
NathanQingyangXu, line 33: The renaming seems appropriate to avoid the confusion and we can remove a 'todo' item in the Javadoc.

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/MappingMetamodelImpl.java
```diff
@@ -743,6 +743,10 @@ public MappingModelExpressable resolveMappingExpressable(SqmExpressable<?> sqmEx
 			throw new NotYetImplementedFor6Exception( "Resolution of embedded-valued SqmExpressable nodes not yet implemented" );
 		}
 
+		if ( sqmExpressable instanceof EntityTypeImpl ) {
+			throw new NotYetImplementedFor6Exception( "Resolution of entity-valued SqmExpressable nodes not yet implemented" );
+		}
+
 		if ( sqmExpressable instanceof EmbeddableTypeImpl ) {
 			return (MappingModelExpressable) sqmExpressable;
 		}
```
NathanQingyangXu, line 752: This class should be the centralized method to resolve mapping. The above insertion is to ensure regression of the methods annotated with **@NotImplementedYet** in org.hibernate.orm.test.jpa.criteria.components.**ComponentInWhereClauseTest**

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -3587,10 +3603,6 @@ public void visitGroupedPredicate(GroupedPredicate groupedPredicate) {
 
 	@Override
 	public void visitInListPredicate(InListPredicate inListPredicate) {
-		if ( inListPredicate.getListExpressions().isEmpty() ) {
-			appendSql( "false" );
-			return;
-		}
```
NathanQingyangXu, line 3593: The above would not work for the following invalid SQL where clause in many DBSMs:
```
where false;
```
The previous v5 approach was reverted back to, though it seems imperfect for **Derby**, which neither supports '**in ()**' nor '**in (null)**'.

dreab8:
Thanks Nathan,
it looks good to me!

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/basic/BasicCriteriaUsageTest.java
```diff
@@ -164,4 +145,106 @@ public void testDoubleNegation() {
...
+		@ManyToOne
+		@JoinColumn(name = "right_id")
+		public Wall getRight() {
+			return right;
+		}
+
+		public void setRight(Wall right) {
+			this.right = right;
+		}
+	}
```
beikov, line 248: Why did you inline the entity classes here and not just reuse the previous ones?

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/basic/BasicCriteriaUsageTest.java
```diff
@@ -63,31 +69,6 @@ public void testTrivialCompilation() {
 		em.close();
 	}
 
-	@Test
-	@TestForIssue(jiraKey = "HHH-8283")
-	public void testDateCompositeCustomType() {
```
beikov, line 68: Why did you remove the test?

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -2589,29 +2588,7 @@ protected Expression consumeSqmParameter(SqmParameter sqmParameter) {
...
-				return tableGroup.getModelPart().findSubPart(
-						navigablePath.getLocalName(),
-						null
-				);
-			}
-
-			return getFromClauseAccess().getTableGroup( navigablePath ).getModelPart();
-		}
-
-		if ( parameterSqmType instanceof BasicValuedMapping ) {
```
beikov, line 2606: This case is not handled in `org.hibernate.metamodel.model.domain.internal.MappingMetamodelImpl#resolveMappingExpressable` and should probably replace the `BasicType` case.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -2957,6 +2957,11 @@ public void visitTuple(SqlTuple tuple) {
 	}
 
 	protected final void renderCommaSeparated(Iterable<? extends SqlAstNode> expressions) {
+		if ( ! expressions.iterator().hasNext() ) {
+			renderEmptyExpressions();
+			return;
+		}
+
```
beikov, line 2964: What's the reason for this? I don't think it's necessary.

```suggestion
```

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -2966,6 +2971,11 @@ protected final void renderCommaSeparated(Iterable<? extends SqlAstNode> express
 	}
 
 	protected final void renderCommaSeparatedSelectExpression(Iterable<? extends SqlAstNode> expressions) {
+		if ( ! expressions.iterator().hasNext() ) {
+			renderEmptyExpressions();
+			return;
+		}
+
```
beikov, line 2978: What's the reason for this? I don't think it's necessary.

```suggestion
```

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -2979,6 +2989,12 @@ protected final void renderCommaSeparatedSelectExpression(Iterable<? extends Sql
 		}
 	}
 
+	private void renderEmptyExpressions() {
+		if ( ! getDialect().supportsEmptyInList() ) {
+			appendSql( NULL_KEYWORD );
+		}
+	}
+
```
beikov, line 2997: Please remove this

```suggestion
```

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
beikov: This is the only case where we need to handle emptyness, so I would rather render `1=0` instead.

beikov (changes requested):
Looks good overall but there are a few issues I found.

hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/basic/ExpressionsTest.java
```diff
@@ -69,154 +60,6 @@ public void cleanupTestData() {
 		em.close();
 	}
 
-	@Test
```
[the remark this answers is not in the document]
jrenaat: I agree; that's in fact the reason why I hadn't moved any of these tests to orm.test package yet.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
NathanQingyangXu: My concern is whether we need to render it with parenthesis as '**(1=0)**' so SQL operator precedence won't be an issue so:
```
1=0 || 5
```
would be interpreted as:
```
1=(0 || 5)
```
Just an example which might be inappropriate but the point might be valid.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/jpa/CriteriaParameterCollectQueryWalker.java
```diff
@@ -93,4 +97,40 @@ public Object visitNamedParameterExpression(SqmNamedParameter expression) {
...
+		expression.getOtherwise().accept( this );
+		return expression;
+	}
+
+	@Override
+	public Object visitBinaryArithmeticExpression(SqmBinaryArithmetic expression) {
+		expression.getLeftHandOperand().accept( this );
+		expression.getRightHandOperand().accept( this );
+		return expression;
+	}
```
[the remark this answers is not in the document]
NathanQingyangXu: I was torn apart by the decision which method to put into base parent class and which ones to be overridden. For instance, it seems our current base or default implementation aims to explore thoroughly, so we should visit as much as possible there? But we can easily think of child class whose requirement conflicts with this, for instance:

- some logging class that needs to input tab or end-of-line before visiting another method;
- some comment specific class that needs to only focus on comment node and ignore all the others and wanna be performant.

The downside to put such stuff into the base parent class is the logic coupling due to inheritance (Effective Java has a good chapter on the danger of it) and it would be so difficult to change the parent class without ripple effect on all the other child classes which relies on it.
Some random thoughts. I am gonna change as you suggested in this PR.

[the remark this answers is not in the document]
NathanQingyangXu: I move methods to BaseSemanticQueryWalker as you suggested.

hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/basic/PredicateTest.java
```diff
@@ -204,115 +125,4 @@ public void testComplicatedNotAnd() {
...
-		List<Order> orders = em.createQuery( orderCriteria ).getResultList();
-		assertTrue( orders.size() == 0 );
-		em.getTransaction().commit();
-		em.close();
-	}
-
-	@Test
-	@TestForIssue( jiraKey = "HHH-5803" )
-	@SkipForDialect( value = CockroachDialect.class, comment = "https://github.com/cockroachdb/cockroach/issues/41943")
-	public void testQuotientConversion() {
```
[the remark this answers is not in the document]
NathanQingyangXu: Thanks for your great catch!! I added it after non-trivial code refactoring. Please see the code changes in org.hibernate.query.sqm.sql.BaseSqmToSqlAstConverter#getExpressionType(SqmBinaryArithmetic expression) in the new commit.

hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/basic/PredicateTest.java
```diff
@@ -204,115 +125,4 @@ public void testComplicatedNotAnd() {
 		em.getTransaction().commit();
 		em.close();
 	}
-
```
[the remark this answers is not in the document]
NathanQingyangXu: The reason is the testing case fails for the current Predicate implementation. Needs to discuss with Steve on this to see whether we need to tweak the testing case.
In v5 when we negate a compound predicate, we reverse its operand (from OR to AND or from AND to OR). Now in v6 we stop doing this and simply end up with a new wrapper. Not sure whether we wanna do this for immutability reason. Needs to confirm with Steve for he implemented this previously.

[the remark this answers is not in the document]
NathanQingyangXu: @sebersole could you shed some lights on this?
```
orderCriteria.select( orderRoot );
Predicate p1 = builder.equal( orderRoot.get( "id" ), "order-1" );
Predicate p2 = builder.equal( orderRoot.get( "id" ), "order-2" );
Predicate compoundPredicate = builder.and( p1, p2 ).not();
// a negated AND should become an OR
assertEquals( Predicate.BooleanOperator.OR, compoundPredicate.getOperator() );
```
The above assertion would fail because we seem to have changed the implementation without reversing as in v5.

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/basic/AggregationResultTest.java
```diff
@@ -0,0 +1,155 @@
...
+	/**
+	 * Sum of Floats should return a Double
+	 */
+	@Test
+	public void testSumOfFloats(EntityManagerFactoryScope scope) {
+		scope.inTransaction( entityManager -> {
+			CriteriaBuilder builder = entityManager.getCriteriaBuilder();
+			CriteriaQuery<Double> criteria = builder.createQuery( Double.class );
+			Root<Product> productRoot = criteria.from( Product.class );
+			criteria.select( builder.sumAsDouble( productRoot.get( Product_.rating ) ) );
```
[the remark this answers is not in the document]
NathanQingyangXu: It would fail and that is why I asked the question on Zulip. I think we broke the backward compatibility here. As per the JPA interface of javax.persistence.criteria.CriteriaBuilder, it implies we should allow this:
Screen Shot 2021-06-01 at 12 29 25

End user of Criteria would feel so confused for they should rely on the API interface code artifact as opposed to JPA Spec doc. Again, @sebersole what is your input here?

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
NathanQingyangXu: I did as you suggested and I am happy we have a much elegant solution in v6 compared to v5. Yeah, this is the only place where it matters.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
NathanQingyangXu: The code change is not necessary as per your approach

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
NathanQingyangXu: Great catch!! I enriched the `MappingMetamodelImpl#resolveMappingExpressable` to accept BasicValuedMapping as well (or any other MappingModelExpressable) as following:
```
if ( sqmExpressable instanceof MappingModelExpressable ) {
	return (MappingModelExpressable) sqmExpressable;
}
```
it also make the following statement unnecessary so I deleted it:
```
if ( sqmExpressable instanceof BasicType<?> ) {
	return (BasicType) sqmExpressable;
}
```

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/basic/AggregationResultTest.java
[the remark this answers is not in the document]
NathanQingyangXu: I revert back the two broken testing cases into org.hibernate.jpa.test.criteria.basic.AggregationResultTest. The float summing test case will trigger the following excpetion:

Caused by: org.hibernate.query.QueryTypeMismatchException: Specified result type [java.lang.Float] did not match Query selection type [java.lang.Double]

hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/basic/ExpressionsTest.java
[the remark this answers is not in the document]
NathanQingyangXu: There are still some broken testing cases which require more time, including:

- Derby's special concat function processing;
- Predicate impl in v6 removes the reverse operation when it is negated; needs to confirm with Steve whether it is intentional;
- Some testing cases expose backward compatibility issue and we need confirmation further.

I would love to migrate the whole class but as long as we managed not to lose one single testing case during the migration, method based incremental migration is positive for we can end up with PR of manageable size.

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/basic/BasicCriteriaUsageTest.java
NathanQingyangXu: My original thought is it is only used within BasicCriteriaUsageTest and making them inline improves code readability (the testing case doesn't require apt result so inline is acceptable). Now I revert back the changes to make testing code comparison easier.

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/basic/BasicCriteriaUsageTest.java
NathanQingyangXu: The test was removed because Steve has changed the entity class field annotation so it doesn't apply any more. I revert back this testing case so we won't lose one single testing case. I need to confirm with Steve the rationale behind the entity class changing (maybe to make his testing cases pass in his PR and now we might be able to revert back).

hibernate-core/src/main/java/org/hibernate/type/descriptor/jdbc/spi/BasicJdbcLiteralFormatter.java
```diff
@@ -29,6 +29,15 @@ public BasicJdbcLiteralFormatter(JavaTypeDescriptor<?> javaTypeDescriptor) {
 			return (X) value;
 		}
 
+		if ( value.getClass().isEnum() ) {
```
beikov, line 32:
```suggestion
		if ( value instanceof Enum<?> ) {
```

hibernate-core/src/main/java/org/hibernate/dialect/function/DerbyConcatFunction.java
```diff
@@ -0,0 +1,101 @@
...
+ * This last spec is to deal with a limitation on DB2 and variants (e.g. Derby)
+ * where dynamic parameters cannot be used in concatenation unless they are being
+ * concatenated with at least one non-dynamic operand.  And even then, the rules
+ * are so convoluted as to what is allowed and when the CAST is needed and when
+ * it is not that we just go ahead and do the CASTing.
+ *
+ * @author Steve Ebersole
+ * @author Nathan Xu
+ */
+public class DerbyConcatFunction extends AbstractSqmSelfRenderingFunctionDescriptor {
```
beikov, line 37: Could you please explain which cases need this function? I would rather see you register the concat function like this: `CommonFunctionFactory.concat_pipeOperator( queryEngine, SqlAstNodeRenderingMode.NO_PLAIN_PARAMETER )`

beikov (changes requested):

hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/basic/BasicCriteriaUsageTest.java
```diff
@@ -36,33 +32,6 @@
...
-		CriteriaQuery<Wall> criteria = em.getCriteriaBuilder().createQuery( Wall.class );
-		criteria.from( Wall.class );
-		em.createQuery( criteria ).getResultList();
-		em.getTransaction().commit();
-		em.close();
-	}
-
 	@Test
 	@TestForIssue(jiraKey = "HHH-8283")
 	public void testDateCompositeCustomType() {
```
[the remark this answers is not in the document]
NathanQingyangXu: The reason is the following code change on Payment entity class by Steve:
Screen Shot 2021-06-04 at 08 04 13
So the Date field ceases to be a composite custom type, thus this testing doesn't apply.
@sebersole , is it desirable to revert back composite type annotation (seems CompositeUserType is replaced by CompositeType in v6?)?

hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/basic/AggregationResultTest.java
```diff
@@ -88,22 +72,6 @@ public void testSumOfIntegers() {
...
-		Root<Product> productRoot = criteria.from( Product.class );
-		criteria.select( builder.sum( productRoot.get( Product_.price ) ) );
-		Object sumResult = em.createQuery( criteria ).getSingleResult();
-		assertReturnType( Double.class, sumResult );
-		em.getTransaction().commit();
-		em.close();
-	}
-
 	/**
 	 * Sum of Floats should return a Float; note that this is distinctly different than JPAQL
```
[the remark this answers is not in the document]
NathanQingyangXu: This testing will fail because of the code change by you to return double type as opposed to float (An exception will be thrown complaining returning type is not expected). Not sure what to do then.

hibernate-core/src/test/java/org/hibernate/jpa/test/criteria/basic/AggregationResultTest.java
```diff
@@ -56,22 +56,6 @@ public void cleanUpTestData() throws Exception {
...
-		Root<Product> productRoot = criteria.from( Product.class );
-		criteria.select( builder.sum( productRoot.get( Product_.partNumber ) ) );
-		Object sumResult = em.createQuery( criteria ).getSingleResult();
-		assertReturnType( Long.class, sumResult );
-		em.getTransaction().commit();
-		em.close();
-	}
-
 	/**
 	 * Sum of Integers should return an Integer; note that this is distinctly different than JPAQL
```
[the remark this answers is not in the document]
NathanQingyangXu: Same reason as the above.

hibernate-core/src/main/java/org/hibernate/dialect/function/DerbyConcatFunction.java
NathanQingyangXu: It would fail in Derby CI mode for the following testing case (org.hibernate.orm.test.jpa.criteria.basic.ConcatTest:
```
@Test
	public void testConcat() throws Exception {
		EntityManager entityManager = getOrCreateEntityManager();
		entityManager.getTransaction().begin();
		try {
			CriteriaBuilder cb = entityManager.getCriteriaBuilder();
			CriteriaQuery query = cb.createQuery();
			Root<TestEntity> testEntity = query.from( TestEntity.class );

			query.select( testEntity ).where( cb.equal( testEntity.get( "name" ), cb.concat( "test", cb.literal( "_1" ) ) ) );

			final List results = entityManager.createQuery( query ).getResultList();
			entityManager.getTransaction().commit();

			assertThat( results.size(), is( 1 ) );
		}
		catch (Exception e) {
			if ( entityManager.getTransaction().isActive() ) {
				entityManager.getTransaction().rollback();
			}
			throw e;
		}
		finally {
			entityManager.close();
		}
	}
```

The reason is Derby needs special treatment when jdbc parameter is involved, as in the above testing case. See DerbyConcatFunction in v5 for further details.

NathanQingyangXu: Seems this function is not common and tightly coupled with Derby dialect.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmMappingModelHelper.java
```diff
@@ -61,11 +59,10 @@ public static EntityPersister resolveEntityPersister(
 	}
 
 	public static <J> SqmPathSource<J> resolveSqmKeyPathSource(
-			String name,
+			CollectionPart.Nature nature,
 			DomainType<J> valueDomainType,
 			Bindable.BindableType jpaBindableType) {
-		// todo (6.0): the key path source must create a special path for the key
```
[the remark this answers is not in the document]
NathanQingyangXu: reverted it back

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -3588,7 +3588,9 @@ public void visitGroupedPredicate(GroupedPredicate groupedPredicate) {
 	@Override
 	public void visitInListPredicate(InListPredicate inListPredicate) {
 		if ( inListPredicate.getListExpressions().isEmpty() ) {
-			appendSql( "false" );
+			// trick to simulate empty in list expression
+			// many a DBMS refuses SQL like 'x in ()'
+			appendSql( "( 1 = 0 )" );
```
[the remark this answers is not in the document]
NathanQingyangXu: done

hibernate-core/src/main/java/org/hibernate/dialect/function/DerbyConcatFunction.java
NathanQingyangXu: I tried you suggestion on org.hibernate.orm.test.jpa.criteria.basic.ConcatTest and found there is a stackoverflow issue as the following code would trigger infinite loop:
```
<img width="1016" alt="Screen Shot 2021-06-04 at 11 52 43" src="https://user-images.githubusercontent.com/8211458/120829415-61341180-c52b-11eb-9ce7-77c89f3d4864.png">
```
due to the internal implementation to append empty string ending up with infinite invoking loops.

NathanQingyangXu: I added a new testing case to cover this troublesome Derby concat scenario. I think the issue still remains when we use jdbc parameter explicitly as opposed to constant value or literal as in the above testConcat() method.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/select/SqmQuerySpec.java
```diff
@@ -240,7 +241,7 @@ public SqmPredicate getRestriction() {
 		if ( getWhereClause() == null ) {
 			setWhereClause( new SqmWhereClause( nodeBuilder() ) );
 		}
-		getWhereClause().applyPredicates( (SqmPredicate[]) restrictions );
+		getWhereClause().applyPredicates( Arrays.copyOf( restrictions, restrictions.length, SqmPredicate[].class ) );
```
beikov, line 244: Please revert this. This is unnecessary.

NathanQingyangXu:
This is necessary for otherwise it will fail to compile. Java Array doesn't
support downcasting for it is not based on type erasion. One of examples we
should deprecate Array usage in favour of collection.
See
https://stackoverflow.com/questions/12210311/downcasting-of-arrays-in-java/12210355

gavinking:
> Java Array doesn't support downcasting for it is not based on type erasion

Well in principle you shouldn't downcast a list element type either, since that is an _unchecked_ cast that can cause failures further along, potentially far from the point at which the typecast was performed. An unchecked cast is actually much worse than a regular typecast.

The really correct solution is to use generics _properly_, with wildcards and type variables. (Though that might be just too much work in this particular case.)

NathanQingyangXu:
> > Java Array doesn't support downcasting for it is not based on type erasion
> 
> Well in principle you shouldn't downcast a list element type either, since that is an _unchecked_ cast that can cause failures further along, potentially far from the point at which the typecast was performed. An unchecked cast is actually much worse than a regular typecast.
> 
> The really correct solution is to use generics _properly_, with wildcards and type variables. (Though that might be just too much work in this particular case.)

Java generics' wildcards is the solution: see https://docs.oracle.com/javase/tutorial/extra/generics/wildcards.html.
So the following method:
```
void method(Collection<? extends Animal> animals);
```
will accept:
```
List<Dog> dogs = ...
method( dogs );
```

See the good tutorial on why compiling would fail for the following method signature:
```
void method(Collection<Animal> animals);
```
The same reason explains the array downcast compilation failure. The difference between array and generics is one of the most ugly and confusing parts in Java design due to its backward compatibility goal (as compared to C#, etc.).

The root cause is Array is covariant and Java generics is based on type erasion, so the wildcards make the downcast possible.

NathanQingyangXu:
This PR is closed for it becomes impossible to review. Will create a new one to replace this soon.

NathanQingyangXu:
This PR was superseded by https://github.com/hibernate/hibernate-orm/pull/4042

PR #4019, opened 2021-06-01
Fix throw exception when more than one row with the given identifier …

dreab8:
…was found and re-enabled additiona tests


hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
```diff
@@ -25,18 +25,24 @@
...
+			return UNIQUE_INSTANCE;
+		}
 		return uniqueFilter ? UNIQUE_FILTER_INSTANCE : NORMAL_INSTANCE;
 	}
 
 	private final boolean uniqueFilter;
+	private final boolean singleResultExpected;
 
-	public ListResultsConsumer(boolean uniqueFilter) {
+	public ListResultsConsumer(boolean uniqueFilter, boolean singleResultExpected) {
```
beikov, line 43: IMO we should introduce an enum for this purpose instead of 2 flags:

```
enum UniqueSemantics {
  NONE,
  FILTER,
  ASSERT;
}
```

hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
```diff
@@ -68,6 +74,14 @@ public ListResultsConsumer(boolean uniqueFilter) {
 				boolean add = true;
 				if ( uniqueRows ) {
 					if ( results.contains( row ) ) {
```
beikov, line 76: It's not part of your PR, but I guess it might be more appropriate to use the backing `JavaTypeDescriptor` and do the unique check based on that. I also think that we should create another loop for the uniquig part. Something like the following:

```
if ( uniqueSemantics == NONE ) {
	while ( rowProcessingState.next() ) {
		results.add( rowReader.readRow( rowProcessingState, processingOptions ) );
		rowProcessingState.finishRowProcessing();
	}
}
else {
	final JavaTypeDescriptor<R> jtd = rowReader.getResultJavaTypeDescriptor();
	while ( rowProcessingState.next() ) {
		final R row = rowReader.readRow( rowProcessingState, processingOptions );
		boolean add = true;
		for ( R existingRow : results ) {
			if ( jtd.isEqual( existingRow, row, session ) ) {
				if ( singleResultExpected && !rowProcessingState.hasCollectionInitializers() ) {
					throw new HibernateException(
							"More than one row with the given identifier was found: " +
								jdbcValuesSourceProcessingState.getExecutionContext().getEntityId() +
								", for class: " +
								resultJavaType.getName()
					);
				}
				add = false;
				break;
			}
		}
		if ( add ) {
			results.add( row );
		}
		rowProcessingState.finishRowProcessing();
	}
}
```

Wdyt?

dreab8: I'll try it

dreab8: @beikov not sure how to get the result `JavaTypeDescriptor` from the `RowReader` in a clean way, any suggestion?

beikov: You would have to expose it in `RowReader`.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/SingleIdLoadPlan.java
```diff
@@ -73,21 +73,22 @@ T load(
...
-		return load( restrictedValue, lockOptions, null, null, session );
+		return load( restrictedValue, lockOptions, null, null,false, session );
 	}
 
 	T load(
 			Object restrictedValue,
 			LockOptions lockOptions,
 			Object entityInstance,
 			Boolean readOnly,
+			Boolean singleResultExpected,
```
beikov, line 91: I was actually thinking about pulling the `UniqueSemantics` up to here to avoid all the boolean stuff down the line. Wdyt?

hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
```diff
@@ -51,35 +62,57 @@ public ListResultsConsumer(boolean uniqueFilter) {
...
+				}
+			}
+			else {
+				boolean uniqueRows = false;
+				final Class<R> resultJavaType = rowReader.getResultJavaType();
+				if ( resultJavaType != null && !resultJavaType.isArray() ) {
+					final EntityPersister entityDescriptor = session.getFactory().getMetamodel().findEntityDescriptor(
+							resultJavaType );
+					if ( entityDescriptor != null ) {
+						uniqueRows = true;
```
beikov, line 80: Maybe we could do something like this?

```suggestion
			boolean uniqueRows = false;
			if ( uniqueSemantic != UniqueSemantic.NONE ) {
				final Class<R> resultJavaType = rowReader.getResultJavaType();
				if ( resultJavaType != null && !resultJavaType.isArray() ) {
					final EntityPersister entityDescriptor = session.getFactory().getMetamodel().findEntityDescriptor(
							resultJavaType );
					if ( entityDescriptor != null ) {
						uniqueRows = true;
					}
				}
			}
			if ( uniqueRows ) {
				final List<JavaTypeDescriptor> resultJavaTypeDescriptors = rowReader.getResultJavaTypeDescriptors();
				assert resultJavaTypeDescriptors.size() == 1;
				final JavaTypeDescriptor<R> resultJavaTypeDescriptor = resultJavaTypeDescriptors.get( 0 );
				while ( rowProcessingState.next() ) {
					final R row = rowReader.readRow( rowProcessingState, processingOptions );
					boolean add = true;
				        for ( R existingRow : results ) {
					        if ( resultJavaTypeDescriptor.areEqual( existingRow, row ) ) {
						        if ( uniqueSemantic == UniqueSemantic.ASSERT && !rowProcessingState.hasCollectionInitializers() ) {
							        throw new HibernateException(
									        "More than one row with the given identifier was found: " +
											        jdbcValuesSourceProcessingState.getExecutionContext()
													        .getEntityId() +
											        ", for class: " +
											        resultJavaType.getName()
							        );
						        }
						        add = false;
						        break;
					        }
				        }
					if ( add ) {
						results.add( row );
					}
					rowProcessingState.finishRowProcessing();
				}
			}
			else {
				while ( rowProcessingState.next() ) {
					results.add( rowReader.readRow( rowProcessingState, processingOptions ) );
					rowProcessingState.finishRowProcessing();
				}
			}
```

dreab8: done, Thanks @beikov for the review :+1:

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/SingleIdLoadPlan.java
dreab8: done, Thanks @beikov

beikov (approved):

dreab8:
pushed upstream,

Thanks @beikov for your help!

PR #4020, opened 2021-06-01
Sebersole topic/6 dtd xsd resolver and proposal for test failures

jrenaat:


hibernate-core/src/test/java/org/hibernate/orm/test/bootstrap/scanning/ScannerTest.java
```diff
@@ -102,10 +104,5 @@ public void testCustomScanner() throws Exception {
 		emf = Persistence.createEntityManagerFactory( "defaultpar", integration );
 		assertTrue( CustomScanner.isUsed() );
 		emf.close();
-
-		CustomScanner.resetUsed();
-		emf = Persistence.createEntityManagerFactory( "defaultpar", null );
-		assertTrue( ! CustomScanner.isUsed() );
-		emf.close();
```
beikov, line 109: What was the reason again for removing this part?

jrenaat: see https://github.com/sebersole/hibernate-orm/pull/196#discussion_r642446147 (linked for reference)

beikov: Ok let's wait for @sebersole since he is involved here as well.

sebersole: I kind of agree here.  If it is explicitly a problem with passing `null` as the Map, we should fix that - handle null as simply empty.  Or even change the test to pass `Collections#emptyMap`.

jrenaat: It's not a matter of whether or not we pass null as the Map, but rather that we're not passing a Classloader instance in such case, which makes the test fail because it cannot constitute the persistence provider. This is why I mentioned that if the solution rests on passing a CL where it is needed for the tests that failed after Steve's refactor in the Map, passing a null Map defeats that solution.

sebersole: Well part of this is actually discovering the scanner which is done via ClassLoader.  Passing in no ClassLoader is handled differently than passing in a ClassLoader (#analysis!).  

Not passing in this ClassLoader is by far the more common case.  You don't think it makes sense to test that case?  Am I misunderstanding something here?

jrenaat: Not at all; my fix simply is flawed, as I have suspected from the get-go, and I haven't found the correct way of improving it yet (I'll have to revisit this)

sebersole: You mean improve "passing the Map"?  We are calling a static method that by nature has to happen before we create the CLS (or even the ServiceRegistry).  I'm not sure there is going to be a better solution there, unfortunately.  

However, I think there may be at least a minor improvement.  There is just one "real" call to that method from creating the `EntityManagerFactoryBuilder` in "SE" environments.  That does have access to all these details although it is not passing them all along.  The method creating the `EntityManagerFactoryBuilder` ultimately has this args:

```
String persistenceUnitName
Map properties
ClassLoader providedClassLoader
ClassLoaderService providedClassLoaderService
```

So we have all the details we need to do this "better" I think without creating the CLS (if not "provided").  Still not a great solution, but like I said... not sure there is a great solution here

sebersole: Ultimately the code-path we are interested in there though is only related to these "SE" deployments which will never pass those values.  So probably not worth the effort.

jrenaat: I'm not sure I understand what the conclusion is here? (read: I'm pretty sure I don't)

PR #4024, opened 2021-06-03
HHH-14649 Oracle limit handler create wrong sql query when multiple spaces are present in the query

dreab8:
https://hibernate.atlassian.net/browse/HHH-14649


NathanQingyangXu:
I would also suggest a more realistic example involving SQL comment (maybe read from a SQL script):
```
SELECT a
FROM b /* for update scenario, use another table, i.e. b2 */
WHERE c > 0;
```
It might be challenging to deal with all possibilities in v6 using SQM for such vendor specific SQM stuff (function approach doesn't apply).
Not sure about my comment. Again, ignore it when needed. Doesn't hurt to think pessimistically.

beikov (approved):

beikov:
In the SQM translation this is not a problem because we don't have to parse SQL @NathanQingyangXu .

PR #4032, opened 2021-06-06
HHH-14656 - SequenceStyleGenerator must set its LegacyHiLoAlgorithmOptimizer optimizer's incrementSize to (allocationSize - 1)

GreenRecycleBin:
To maintain backward compatibility between
SequenceStyleGenerator + LegacyHiLoAlgorithmOptimizer and
SequenceHiLoGenerator + LegacyHiLoAlgorithmOptimizer.


beikov:
I tried your test and from what I understand, the behavior you seem to be looking for is provided by the `hilo` optimizer already. I have not checked what earlier versions of Hibernate ORM did, but changing a legacy optimizer is not a good idea. At best, a new optimizer can be added to handle what previous versions of Hibernate ORM did, but you can also do that yourself.

PR #4033, opened 2021-06-07
HHH-14659 "join fetch" on mapped-by association is ignored when using bytecode enhancement

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14659

This is not a great fix. But it should work. All the code related to loaders, and especially lazy/eager fetching, is being rewritten in ORM 6, and for good reasons. With that in mind, I think investing more in this fix would be a waste of time.

Here's an explanation of the problem, if someone wants more information:

For mapped-by associations, we generate the correct SQL and fetch the associated entities correctly, and we even add them to the session. But when we decide which value to assign to the corresponding property, we assume that since the property is lazy,  we didn't retrieve it. Which is wrong when `join fetch` is being used.

This behavior was most likely implemented when fixing [HHH-13134](https://hibernate.atlassian.net/browse/HHH-13134) as a way to avoid unnecessary queries for lazy associations. In most cases it's a good thing, but here it's causing us to ignore a `join fetch`.

To be more precise,  when the problem occurs, the `laziness` local variable in `org.hibernate.persister.entity.AbstractEntityPersister#hydrate` (`org/hibernate/persister/entity/AbstractEntityPersister.java:3191`) is `true` for the property representing the association, which causes the `hydrate` method to consider the property "unfetched", even though the associated entity is being loaded.

When the association is not `mapped-by`, the property does get set because `laziness` in this method is `false`, so we set the value to the ID of the non-owning entity, and later another method will retrieve the associated entity from the persistence context and set the property value (in `org.hibernate.engine.internal.TwoPhaseLoad#initializeEntityEntryLoadedState`, which will indirectly call `org.hibernate.type.EntityType#loadByUniqueKey`).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/Loader.java
@@ -65,7 +65,6 @@ import org.hibernate.engine.spi.TypedValue;
 import org.hibernate.event.spi.EventSource;
 import org.hibernate.event.spi.PostLoadEvent;
 import org.hibernate.event.spi.PreLoadEvent;
-import org.hibernate.event.spi.PreLoadEventListener;
 import org.hibernate.hql.internal.HolderInstantiator;
 import org.hibernate.internal.CoreLogging;
 import org.hibernate.internal.CoreMessageLogger;
@@ -136,11 +135,24 @@ public abstract class Loader {
 
 	/**
 	 * An array indicating whether the entities have eager property fetching
-	 * enabled.
+	 * enabled for all of their properties.
+	 * <p>
+	 * Supersedes {@link #getEntityEagerPerPropertyFetches()}.
 	 *
 	 * @return Eager property fetching indicators.
 	 */
 	protected boolean[] getEntityEagerPropertyFetches() {
+		return null;
+	}
+
+	/**
+	 * An array indicating for each entity which specific properties must have eager fetching enabled.
+	 * <p>
+	 * Superseded by {@link #getEntityEagerPropertyFetches()}.
+	 *
+	 * @return Eager property fetching indicators.
+	 */
+	protected boolean[][] getEntityEagerPerPropertyFetches() {
 		return null;
 	}
 
@@ -1799,11 +1811,15 @@ public abstract class Loader {
 		hydratedObjects.add( object );
 	}
 
-	private boolean isEagerPropertyFetchEnabled(int i) {
+	private boolean isAllPropertyEagerFetchEnabled(int i) {
 		boolean[] array = getEntityEagerPropertyFetches();
 		return array != null && array[i];
 	}
 
+	private boolean[] getPerPropertyEagerFetchEnabled(int i) {
+		boolean[][] array = getEntityEagerPerPropertyFetches();
+		return array != null ? array[i] : null;
+	}
 
 	/**
 	 * Hydrate the state an object from the SQL <tt>ResultSet</tt>, into
@@ -1837,7 +1853,7 @@ public abstract class Loader {
 			);
 		}
 
-		boolean fetchAllPropertiesRequested = isEagerPropertyFetchEnabled( i );
+		boolean fetchAllPropertiesRequested = isAllPropertyEagerFetchEnabled( i );
 
 		// add temp entry so that the next step is circular-reference
 		// safe - only needed because some types don't take proper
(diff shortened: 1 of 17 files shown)


NathanQingyangXu:
> https://hibernate.atlassian.net/browse/HHH-14659
> 
> This is not a great fix. But it should work. All the code related to loaders, and especially lazy/eager fetching, is being rewritten in ORM 6, and for good reasons. With that in mind, I think investing more in this fix would be a waste of time.
> 
> Here's an explanation of the problem, if someone wants more information:
> 
> For mapped-by associations, we generate the correct SQL and fetch the associated entities correctly, and we even add them to the session. But when we decide which value to assign to the corresponding property, we assume that since the property is lazy, we didn't to retrieve it. Which is wrong when `join fetch` is being used.
> 
> This behavior was most likely implemented when fixing [HHH-13134](https://hibernate.atlassian.net/browse/HHH-13134) as a way to avoid unnecessary queries for lazy associations. In most cases it's a good thing, but here it's causing us to ignore a `join fetch`.
> 
> To be more precise, when the problem occurs, the `laziness` local variable in `org.hibernate.persister.entity.AbstractEntityPersister#hydrate` (`org/hibernate/persister/entity/AbstractEntityPersister.java:3191`) is `true` for the property representing the association, which causes the `hydrate` method to consider the property "unfetched", even though the associated entity is being loaded.
> 
> When the association is not `mapped-by`, the property does get set because `laziness` in this method is `false`, so we set the value to the ID of the non-owning entity, and later another method will retrieve the associated entity from the persistence context and set the property value (in `org.hibernate.engine.internal.TwoPhaseLoad#initializeEntityEntryLoadedState`, which will indirectly call `org.hibernate.type.EntityType#loadByUniqueKey`).

The description is great

> https://hibernate.atlassian.net/browse/HHH-14659
> 
> This is not a great fix. But it should work. All the code related to loaders, and especially lazy/eager fetching, is being rewritten in ORM 6, and for good reasons. With that in mind, I think investing more in this fix would be a waste of time.
> 
> Here's an explanation of the problem, if someone wants more information:
> 
> For mapped-by associations, we generate the correct SQL and fetch the associated entities correctly, and we even add them to the session. But when we decide which value to assign to the corresponding property, we assume that since the property is lazy, we didn't to retrieve it. Which is wrong when `join fetch` is being used.
> 
> This behavior was most likely implemented when fixing [HHH-13134](https://hibernate.atlassian.net/browse/HHH-13134) as a way to avoid unnecessary queries for lazy associations. In most cases it's a good thing, but here it's causing us to ignore a `join fetch`.
> 
> To be more precise, when the problem occurs, the `laziness` local variable in `org.hibernate.persister.entity.AbstractEntityPersister#hydrate` (`org/hibernate/persister/entity/AbstractEntityPersister.java:3191`) is `true` for the property representing the association, which causes the `hydrate` method to consider the property "unfetched", even though the associated entity is being loaded.
> 
> When the association is not `mapped-by`, the property does get set because `laziness` in this method is `false`, so we set the value to the ID of the non-owning entity, and later another method will retrieve the associated entity from the persistence context and set the property value (in `org.hibernate.engine.internal.TwoPhaseLoad#initializeEntityEntryLoadedState`, which will indirectly call `org.hibernate.type.EntityType#loadByUniqueKey`).

I am not sure whether this fix is great, but definitively the fix explanation is so great!

hibernate-core/src/main/java/org/hibernate/loader/Loader.java
```diff
@@ -136,11 +135,22 @@ public Loader(SessionFactoryImplementor factory) {
...
 	/**
 	 * An array indicating whether the entities have eager property fetching
-	 * enabled.
+	 * enabled for all of their properties.
+	 * <p>
+	 * Supersedes {@link #getEntityPropertyEagerFetches()}.
 	 *
 	 * @return Eager property fetching indicators.
 	 */
-	protected boolean[] getEntityEagerPropertyFetches() {
```
dreab8, line 143: not sure if this renaming can cause any compatibility issue

sebersole: I agree.  As Loader is meant to be extended and this is protected it is at least possible that someone may override it

sebersole (approved):
Looks good to me

yrodiere:
Thanks for the reviews. I amended my fix to avoid backwards-incompatible renamings. I think we can merge as soon as CI passes :)

PR #4034, opened 2021-06-07
HHH-14655 Upper case the information schema sequence identifier again to fix issues with case sensitive databases

beikov:
https://hibernate.atlassian.net/browse/HHH-14655


gavinking:
OOoops, my bad: I did the right thing in H6, but then futzed it up here. There's even a comment explaining that it's needed. I suck.

gavinking (approved):

beikov:
NP Gavin, happens to the best ;)

gavinking:
P.S. @beikov I think the comment should be improved.

Instead of "this should work" which sounds like the person didn't test it, I think it should explain the situation, that is, that the database server can have defined "collation mode", some of which have case-sensitive table names (though this is not the default).

gavinking:
Thanks for cleaning up my mess.

PR #4035, opened 2021-06-07
wip/6.0 HHH 14642 verify Criteria - tuple

NathanQingyangXu:
https://hibernate.atlassian.net/jira/software/c/projects/HHH/issues/HHH-14642

Just another sub task of Criteria verifying to make the grand goal achievable in peace meal manner . Easy to understand and verify.
The root cause is the tuple implementation doesn't maintain the original order. Also, if multiple entry has no alias (as common), the last entry would overwrite the others simply because we are using HashMap.

Other minor changes are provided to have the migrated single testing class succeed.

This PR requires the merging of https://github.com/hibernate/hibernate-orm/pull/4018. Before that, it will be in 'DRAFT' state.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,10 +21,10 @@ While we try to keep requirements for contributing to a minimum, there are a few
 we ask that you mind.
 
 For code contributions, these guidelines include:
-* respect the project code style - find templates for [IntelliJ IDEA](https://community.jboss.org/docs/DOC-15468) or [Eclipse](https://community.jboss.org/docs/DOC-16649)
+* respect the project code style - find templates for [IntelliJ IDEA](https://hibernate.org/community/contribute/intellij-idea/) or [Eclipse](https://hibernate.org/community/contribute/eclipse-ide/)
 * have a corresponding JIRA issue and the key for this JIRA issue should be used in the commit message
 * have a set of appropriate tests.  For bug reports, the tests reproduce the initial reported bug
-	and illustrates that the solution actually fixes the bug.  For features/enhancements, the 
+	and illustrate that the solution actually fixes the bug.  For features/enhancements, the 
 	tests illustrate the feature working as intended.  In both cases the tests are incorporated into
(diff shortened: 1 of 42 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/tree/select/SqmSelectClause.java
```diff
@@ -62,6 +65,11 @@ public void makeDistinct(boolean distinct) {
 	public void addSelection(SqmSelection selection) {
 		if ( selections == null ) {
 			selections = new ArrayList<>();
+			aliases = new HashSet<>();
+		}
+		final String alias = selection.getAlias();
+		if ( alias != null && ! aliases.add( alias )) {
+			throw new IllegalArgumentException( "Multi-select expressions defined duplicate alias : " + alias );
```
NathanQingyangXu, line 72: The above exception messages are copied from v5 without changes.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -1532,17 +1533,18 @@ public SelectClause visitSelectClause(SqmSelectClause selectClause) {
 
 	@Override
 	public Void visitSelection(SqmSelection<?> sqmSelection) {
-		final Map<String, DomainResultProducer<?>> resultProducers;
+		final List<Pair<String, DomainResultProducer>> aliasAndProducerPairs;
```
NathanQingyangXu, line 1536: We might well create a simple class to combine both together. But I often feel the need of a Pair util class here and there and with good variable naming we won't impact code readability.
Let me know your preference and I am ready to revert back the Pair util class and either:

1. create a simple POJO class to combine both alias and producer;
2. use two Lists to contain alias and producer respectively (but code reader needs to keep that in mind that the sync up between them).

PR #4039, opened 2021-06-07
HHH-14657 Use the compile instead of runtime classpath

markusheiden:
The runtime classpath contains the dependencies as jars that are not yet built when the plugin needs them. So use the compile classpath plus the compiled classes of the current project.


Sanne:
Looks reasonable! Thanks

PR #4042, opened 2021-06-08
wip/6.0 HHH-14642 Criteria verifying - basic

NathanQingyangXu:
This PR aims to be the first step of criteria verifying and it mainly covers the 'basic' testing package migration, though other minor packages are also included. It superseded the previous https://github.com/hibernate/hibernate-orm/pull/4018 wherein @beikov left many great comments.

The main code changes are summarized as following:

- simplify the previous 'empty in list' processing in dialect on top of @beikov 's previous work. A expression simulation (`(1 = 0)` is used rather than the tricky 'in (null)' trick which was not universally workable for all dialects (Derby supports neither `in ()` nor `in (null)`, for instance); this change accounts for most of the file changes in this PR for we can delete the obsolete dialect feature (`supportsEmptyInList()`) from everywhere.
- add a missing `DerbyConcatSimulation` for Derby has special requirement when any jdbc parameter is involved (constant value inlining won't solve the issue); compared to previous v5 counterpart, the logic is changed to use 'cast' trick when ANY jdbc parameter is found (v5 counterpart logic is when ALL arguments are jdbc parameters). To cover the error-prone Derby concat processing, a new testing case was added to showcase the `ANY` logic is correct.
- many code improvements and refactoring efforts to ensure testing case migration.

Especially, I wanna share two examples why array usage is so error-prone (but I didn't refactor in favour of Java generics but rely on ugly fixes instead):

- The following array down casting will trigger runtime exception (see https://stackoverflow.com/questions/12210311/downcasting-of-arrays-in-java/12210355)
```
	@Override
	public SqmQuerySpec<T> setRestriction(Predicate... restrictions) {
		if ( getWhereClause() == null ) {
			setWhereClause( new SqmWhereClause( nodeBuilder() ) );
		}
		getWhereClause().applyPredicates( (SqmPredicate[]) restrictions );
		return null;
	}
```
- The following statement
```
	@Override
	public SqmPredicate in(Collection<?> values) {
		return nodeBuilder().in( this, values );
	}
```
will be routed by Java compiler to
```
	@Override
	<T> SqmInPredicate<T> in(Expression<? extends T> expression, T... values);
```
not the intended
```
	@Override
	<T> SqmInPredicate<T> in(Expression<? extends T> expression, Collection<T> values);
```

Two corresponding fixes are provided for they are needed.


beikov:

PR #4044, opened 2021-06-09
HHH-14660 Deprecate component mappings with different attributes for the same class

beikov:
https://hibernate.atlassian.net/browse/HHH-14660

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/log/DeprecationLogger.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/log/DeprecationLogger.java
@@ -250,4 +250,12 @@ public interface DeprecationLogger extends BasicLogger {
 					"5.3 in upgrading.  It will be removed in a later version."
 	)
 	void logUseOfDeprecatedZeroBasedJdbcStyleParams();
+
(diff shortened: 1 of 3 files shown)


dreab8 (approved):
LGTM

hibernate-core/src/main/java/org/hibernate/internal/EntityManagerMessageLogger.java
```diff
@@ -133,4 +133,12 @@ void unableToLocateStaticMetamodelField(
 	)
 	void duplicatedPersistenceUnitName(String name);
 
+	@LogMessage(level = WARN)
```
sebersole, line 136: Please use `DeprecationLogger` for this..

sebersole (approved):

sebersole:
LGTM except for the DeprecationLogger comment

hibernate-core/src/main/java/org/hibernate/internal/EntityManagerMessageLogger.java
beikov: Fixed, thanks!

PR #4045, opened 2021-06-10
Fix many-to-many wrong table for order column

dreab8:


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ordering/ast/ColumnReference.java
```diff
@@ -110,7 +109,11 @@ TableReference getTableReference(TableGroup tableGroup) {
 		ModelPartContainer modelPart = tableGroup.getModelPart();
 		if ( modelPart instanceof PluralAttributeMapping ) {
 			final PluralAttributeMapping pluralAttribute = (PluralAttributeMapping) modelPart;
-			final MappingType elementMappingType = (pluralAttribute).getElementDescriptor().getPartMappingType();
+			if ( !pluralAttribute.getCollectionDescriptor().hasManyToManyOrdering() ) {
+				return tableGroup.getPrimaryTableReference();
+			}
+
+			final MappingType elementMappingType = ( pluralAttribute ).getElementDescriptor().getPartMappingType();
```
beikov, line 116:
```suggestion
			final MappingType elementMappingType = pluralAttribute.getElementDescriptor().getPartMappingType();
```

beikov (approved):
LGTM!

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ordering/ast/ColumnReference.java
dreab8: Thanks @beikov, fixed it

hibernate-core/src/test/java/org/hibernate/orm/test/idbag/IdBagTest.java
```diff
@@ -76,33 +73,33 @@ public void testUpdateIdBag() throws HibernateException {
...
+
+					List l = session.createQuery( "from User u join u.groups g" ).list();
+					assertEquals( 1, l.size() );
+
+					session.clear();
+
+					gavin = (User) session.createQuery( "from User u join fetch u.groups" ).uniqueResult();
+					assertTrue( Hibernate.isInitialized( gavin.getGroups() ) );
+					assertEquals( gavin.getGroups().size(), 2 );
+					assertEquals( ( (Group) gavin.getGroups().get( 0 ) ).getName(), "admins" );
```
NathanQingyangXu, line 96: technically, you messed up the parameter order, though the difference can only be noticed when the testing case is broken.

dreab8: I haven't noticed that the original was using the wrong order, I'm going to fix it.
Thanks

NathanQingyangXu:
Arguably hamcrest assertThat is more readable, consistent and free of such
issue. But it seems not mandatory.

beikov:
Don't worry about the failure for MariaDB. wip/6.0 fails also due to a recent test move by Steve. I fixed that test on my other PR already though. IMO we can merge this.

dreab8:
Applied upstream,

Thanks @beikov and @NathanQingyangXu for the review!

PR #4046, opened 2021-06-10
Wip/6.0 AbstractEntityPersister#findSubpart

dreab8:
Fix AbstractEntityPersister#findSubpart returning the wrong ModelPart when the Entity has an attribute named id that it is not the identifier


beikov (approved):

sebersole (approved):

beikov:
FYI, this is the change we discussed @dreab8 https://github.com/hibernate/hibernate-orm/blob/c33e375bf71ecf1cfec30950daec01a5b32ff600/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java#L6786

dreab8:
Hi @beikov ,

I have changed that method too and it is similar to what you have done, I'm going to try your solution.

dreab8:
Pushed upstream,

Thanks @sebersole and @beikov for the review!

PR #4048, opened 2021-06-11
Test move

beikov:
Move quite a few tests and in the process, fix some issues:

* Fix issues with order by fragments for collections
* Fix quite a lot of caching and related statistics use cases
* Support selecting ToOne- and PluralAttributeMappings in LoaderSelectBuilder
* Fix multiple bag role handling in LoaderSelectBuilder
* Fix dynamic batch id entity loader problems
* Simplify lock handling and cache JdbcSelect for SingleIdLoadPlan
* Add support for, but at the same time deprecation-warn about component mappings with different structure but same Java type. See HHH-14660
* Fix enhancement issues and implement support for property laziness
* Implement association key access for both sides of the FK
* Fix subpart access for nullable + TARGET side ToOneAttributeMapping
* Pass through fetch timing to CircularFetch to respect fetch profiles
* Respect fetch profiles for CollectionLoader and SingleIdLoadPlan
* Support group by and order by for entity and embeddable paths/joins
* Don't apply lock clause for LockMode.READ and less
* Change result type for round to argument type
* Fix inference for tuple parameter elements when comparing against embeddables


hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
```diff
@@ -1109,15 +1109,28 @@ private static String buildOrderByClauseFromHql(String orderByFragment, Persiste
...
+	private static String buildOrderById(PersistentClass associatedClass, String order) {
+		final StringBuilder sb = new StringBuilder();
+		final Iterator<Selectable> columnIterator = associatedClass.getIdentifier().getColumnIterator();
+		while ( columnIterator.hasNext() ) {
+			final Selectable selectable = columnIterator.next();
+			sb.append( selectable.getText() );
+			sb.append( order );
+			sb.append( ", " );
+		}
+		sb.setLength( sb.length() - 2 );
```
NathanQingyangXu, line 1130: I think the below code change would be slightly better
```
while ( columnIterator.hasNext() ) {
    ... ...
    if ( columnIterator.hasNext() ) {
        sb.append(", ");
    }
}
```

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EntityCollectionPart.java
```diff
@@ -173,109 +174,25 @@ public EntityFetch generateFetch(
...
-	}
+		final FromClauseAccess fromClauseAccess = creationState.getSqlAstCreationState().getFromClauseAccess();
+		final TableGroup partTableGroup = fromClauseAccess.resolveTableGroup(
+				navigablePath,
+				np -> {
+					// We need to create one.  The Result will be able to find it later by path
+
+					// first, find the collection's TableGroup
+					final TableGroup collectionTableGroup = fromClauseAccess
+							.getTableGroup( navigablePath.getParent() );
```
NathanQingyangXu, line 185: technically we are supposed to use the `np` lambda parameter here

hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
beikov: I like the branchless loop-body approach more though.

sebersole (commented):
Based on our chat today, aside from the few verification tasks, +1

hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
NathanQingyangXu: Got it. I hate to be focused on such low level details and obviously the PR is so great. But the current approach will trigger exception when the while loop is skipped in the first place. If we can ensure at least one column exists, we might well achieve the goal of branchless coding without any hack:
```
sb.append( columnIterator.next() );
while ( columnIterator.hasNext() ) {
    sb.append(", ");
    sb.append( columnIterator.next() );
}
```
Ignore my comment if needed however.

PR #4049, opened 2021-06-11
wip/6.0 HHH-14642 Critera verifying - mapjoin

NathanQingyangXu:
Another sub task of https://hibernate.atlassian.net/browse/HHH-14642

Currently the EntityCollectionPart class has limitation for not being able to process **{element}** and **{index}** when either part points to external entity (other than the parent owner entity) except for many-to-many case. This PR tries to fix this by migrating all testing cases in the original jpa.test.criteria.**mapjoin** package.

The trick is **PluralAttributeMappingImpl** has stored foreign keys for both internally and we need to expose them to **EntityCollectionPart** and make most of them. One caveat is if **PluralAttributeMappingImpl** is inverse (e.g. in common one-to-many case), the foreign key descriptors should be avoided for they are dummy ones (the **{element}** fk will point to itself per se) and we should use the one pointing to owner side instead (as we have done previously, but for all cases wrongly).


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/PluralAttributeMappingImpl.java
```diff
@@ -269,6 +271,7 @@ public void finishInitialization(
 						indexFkDescriptor = createForeignKeyDescriptor(
 								( (IndexedCollection) bootDescriptor ).getIndex(),
 								(EntityType) collectionDescriptor.getIndexType(),
+								bootDescriptor.getCollectionTable(),
 								creationProcess,
```
NathanQingyangXu, line 275: The {element} foreign key descriptor created in case of embedded identifier case has bug to have wrong key table. I have to expose a new parameter to pass key table explicitly to **MappingModelCreationHelper .buildEmbeddableForeignKeyDescriptor()** method.

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/mapjoin/MapJoinTestWithEmbeddable.java
```diff
@@ -65,12 +65,12 @@ public void testSelectingKeyOfMapJoin() {
...
 			CriteriaBuilder cb = entityManager.getCriteriaBuilder();
-			CriteriaQuery<Node> query = cb.createQuery( Node.class );
+			CriteriaQuery<BatchNodeMetadata> query = cb.createQuery( BatchNodeMetadata.class );
 			Root<Batch> root = query.from( Batch.class );
 
-			MapJoin nodes = (MapJoin) root.join( "batchNodeMetadata" );
+			MapJoin<Batch, Node, BatchNodeMetadata> nodes = (MapJoin) root.join( "batchNodeMetadata" );
 
-			query.select( nodes );
+			query.select( nodes.value() );
```
NathanQingyangXu, line 73: This migrated testing case is wrong technically in v5 for the type of `nodes` doesn't match up with the specified query type (i.e. `Node`). Given that the other testing case has tested `nodes.key()`, this testing case was changed to select `nodes.value()`.
A good example to showcase the power of Java generics type checking. Java compiler will complain type inconsistency as long as we use generics type to `MapJoin`.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/SqmMapJoin.java
```diff
@@ -76,75 +75,81 @@ public SqmMapJoin(
...
+		}
+		registerReusablePath( sqmPath );
+		return sqmPath;
 	}
 
 	@Override
 	@SuppressWarnings("unchecked")
 	public Expression<Map.Entry<K, V>> entry() {
+		key();
+		value();
```
beikov, line 152: What's the reason for this? If this really is necessary, this should be moved to the constructor of `SqmMapEntryReference` to make sure this always happens and add a comment explaining why this is necessary.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/SqmMapJoin.java
```diff
@@ -76,75 +75,81 @@ public SqmMapJoin(
 		final SqmPathSource keyPathSource = getReferencedPathSource().getKeyPathSource();
```
beikov, line 75: Ideally, the method would just use `org.hibernate.query.sqm.tree.domain.AbstractSqmPath#resolvePath(java.lang.String, org.hibernate.query.sqm.SqmPathSource<?>)`, but we need a special path source implementation for the key type as per the todo in `org.hibernate.query.sqm.internal.SqmMappingModelHelper#resolveSqmKeyPathSource`. Apart from that, it would be nice if you would reuse the same approach for `org.hibernate.query.hql.internal.SemanticQueryBuilder#visitCollectionElementNavigablePath`, `org.hibernate.query.hql.internal.SemanticQueryBuilder#visitCollectionIndexFunction` and `org.hibernate.query.hql.internal.SemanticQueryBuilder#visitMapKeyNavigablePath`. In fact, every `createSqmPath` call in `SemanticQueryBuilder` is questionable IMO as that should all be handled through `resolvePath` or some other path related methods.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -2101,9 +2101,12 @@ private void consumeReusablePaths(
...
 						log.tracef( "Starting implicit join handling for `%s`", joinedPath.getNavigablePath() );
 					}
-					// No need to create a table group for this path if no sub-paths exist
 					if ( joinedPath.getReusablePaths().isEmpty() ) {
-						return;
+						final String localName = joinedPath.getNavigablePath().getUnaliasedLocalName();
+						if ( CollectionPart.Nature.fromNameStrictly( localName ) == null ) {
+							// No need to create a table group for this path if it is not special path and no sub-paths exist
+							return;
+						}
```
beikov, line 2109: Shouldn't this be unnecessary when we register reusable paths now anyway?

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/CollectionPart.java
```diff
@@ -53,6 +53,15 @@ public static Nature fromName(String name) {
...
 		}
+
+		public static Nature fromNameStrictly(String name) {
+			for ( Nature nature : Nature.values() ) {
+				if ( nature.getName().equals( name ) ) {
+					return nature;
+				}
+			}
+			return null;
+		}
```
beikov, line 64: Should be obsolete.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EntityCollectionPart.java
```diff
@@ -241,8 +242,12 @@ public String toString() {
 
 	@Override
 	public ForeignKeyDescriptor getForeignKeyDescriptor() {
-		// todo (6.0) : this will not strictly work - we'd want a new ForeignKeyDescriptor that points the other direction
-		return collectionDescriptor.getAttributeMapping().getKeyDescriptor();
+		if ( collectionDescriptor.isInverse() ) {
```
beikov, line 245: I don't know exactly what the reason for this is, but it feels wrong to me that we could return the FK descriptor for the index part of the collection. If there is a problem with the direction of the foreign key, this should be handled by returning a different `ForeignKeyDescriptor.Nature` in the method below. At least this is how `ToOneAttributeMapping`s are handled.

IMO this should always return `return collectionDescriptor.getAttributeMapping().getKeyDescriptor( nature );`

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EntityCollectionPart.java
```diff
@@ -270,16 +275,35 @@ public TableGroupJoin createTableGroupJoin(
...
-				navigablePath,
-				lhs,
-				explicitSourceAlias,
-				sqlAstJoinType,
-				fetched,
-				aliasBaseGenerator,
-				sqlExpressionResolver,
-				creationContext
-		);
+		if ( collectionDescriptor.isInverse() ) {
```
beikov, line 278: I don't think `isInverse` is the proper flag for this. IMO this should be controlled by the FK descriptor directionality. AFAIU the `isInverse` method just signals if collection changes should be flushed to the database, which might by coincidence be true/false in the same cases when the FK is _inverse_, but it not necessarily correct.

beikov (changes requested):

beikov:

PR #4055, opened 2021-06-17
HHH-11817 Allow schema-export commands written to file to truncate in addition to current appending

dreab8:
https://hibernate.atlassian.net/browse/HHH-11817


yrodiere (approved):
LGTM. It's clearly the safest option we have in 5.x.

Now, about 6.0... Maybe we should consider reverting the default behavior to "truncate" in 6.0, since apparently it's causing problems to some people? Though TBH I don't see the use case for appending to an existing file, so in 6.0 we could just set the behavior to "truncate" **and** remove the configuration property completely... ?

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -1588,6 +1588,16 @@
 	@SuppressWarnings("JavaDoc")
 	String HBM2DDL_SCRIPTS_CREATE_TARGET = "javax.persistence.schema-generation.scripts.create-target";
 
+	/**
+	 * For cases where the {@value #HBM2DDL_SCRIPTS_ACTION} value indicates that schema commands should
+	 * be written to DDL script file, specifies specifies if schema commands should be appended to the end of the file rather than written at the beginning of the file.
```
NathanQingyangXu, line 1593: there are two 'specified' here

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -1588,6 +1588,16 @@
...
+	 *
+	 * Values are: {@code true} for appending schema commands to the end of the file, {@code false} for schema writing commands at the beginning.
```
NathanQingyangXu, line 1595: 'for schema writing commands' -> 'for writing schema commands'?

hibernate-core/src/main/java/org/hibernate/tool/hbm2ddl/SchemaExport.java
```diff
@@ -159,6 +160,18 @@ public SchemaExport setOutputFile(String filename) {
...
 	}
 
+	/**
+	 * For generating a export script file, by default the content will be appended at the begin or end of the file.
+	 *
+	 * the sql will be written to the end of the file rather than the beginning
+	 *
+	 * @return this
+	 */
+	public SchemaExport setOverrideOutputFileContent(){
```
NathanQingyangXu, line 170: I hate to focus on such minor details, but adding space delimiter before '{' is definitively better

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/exec/ScriptTargetOutputToFile.java
```diff
@@ -36,10 +37,24 @@
 	 *
 	 * @param file The file to read from
 	 * @param charsetName The charset name
+	 * @param append If true, then bytes will be written to the end of the file rather than the beginning
 	 */
-	public ScriptTargetOutputToFile(File file, String charsetName) {
+	public ScriptTargetOutputToFile(File file, String charsetName, boolean append ) {
```
NathanQingyangXu, line 42: super minor, but I don't think we need space char after 'append' here.

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/exec/ScriptTargetOutputToFile.java
```diff
@@ -72,7 +87,7 @@ public void release() {
 	}
 
 	@SuppressWarnings("ResultOfMethodCallIgnored")
-	static Writer toFileWriter( File file, String charsetName ) {
+	static Writer toFileWriter( File file, String charsetName, boolean append ) {
```
NathanQingyangXu, line 90: As per our convention, we need space delimiter when we invoke some method but we don't need it for method declaration, right?

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/exec/ScriptTargetOutputToUrl.java
```diff
@@ -36,10 +37,23 @@
 	 *
 	 * @param url The url to read from
 	 * @param charsetName The charset name
+	 * @param append If true, then bytes will be written to the end of the file rather than the beginning
 	 */
-	public ScriptTargetOutputToUrl(URL url, String charsetName) {
+	public ScriptTargetOutputToUrl(URL url, String charsetName, boolean append ) {
```
NathanQingyangXu, line 42: space delimiter after 'append' unnecessary and inconsistent here.

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaCreationToOutputScriptTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+	private File output;
+	private ServiceRegistry serviceRegistry;
+	private MetadataImplementor metadata;
+
+	@Before
+	public void setUp() throws Exception {
+		output = File.createTempFile( "creation_script", ".sql" );
+		output.deleteOnExit();
+		try (FileWriter writer = new FileWriter( output )) {
```
NathanQingyangXu, line 54: Technically speaking, we should avoid writing to file without buffering in legacy Java file IO for that is a big performance drain. But given this is a testing case, the concern might be discarded. Alternatively we can use Java NIO file processing without such concern at all:
```
public static Path write(Path path, byte[] bytes, OpenOption... options)
        throws IOException
```
see https://mkyong.com/java/how-to-write-data-to-temporary-file-in-java/

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaMigrationToOutputScriptTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+		SchemaManagementToolCoordinator.process(
+				metadata,
+				serviceRegistry,
+				properties,
+				null
+		);
+		List<String> commands = Files.readAllLines( output.toPath() );
+		assertThat( commands.size(), is( 10 ) );
+		assertTrue( commands.get( 9 ).contains( createTableMySecondEntity ) );
```
NathanQingyangXu, line 160: Not defect, just sharing idea. We might well use 'StringContains' hamcrest matcher residing in the same package as 'Is':
```
assertThat( commands.get( 9 ), containsString( createTableMySecondEntity ) );
```

NathanQingyangXu (approved):

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
dreab8: Thanks @NathanQingyangXu , fixed it!

hibernate-core/src/main/java/org/hibernate/tool/hbm2ddl/SchemaExport.java
dreab8: Thanks @NathanQingyangXu , fixed it!

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
dreab8: Thanks @NathanQingyangXu , fixed it!

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/exec/ScriptTargetOutputToFile.java
dreab8: Thanks @NathanQingyangXu , fixed it!

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/exec/ScriptTargetOutputToUrl.java
dreab8: Thanks @NathanQingyangXu , fixed it!

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaCreationToOutputScriptTest.java
dreab8: Thanks @NathanQingyangXu , used the buffering!

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaMigrationToOutputScriptTest.java
dreab8: Thanks @NathanQingyangXu , the `containsString` is definitely better than the `assertTrue` and `contains`!

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaCreationToOutputScriptTest.java
```diff
@@ -0,0 +1,170 @@
...
+				, "This is the sixth line"
+				, "This is the seventh line"
+				, "This is the eighth line"
+				, "This is the ninth line"
+		);
+
+		try (BufferedWriter bw = new BufferedWriter( new FileWriter( output ) )) {
+			for ( String s : content ) {
+				bw.write( s );
+				bw.write( System.lineSeparator() );
```
NathanQingyangXu, line 72: You might well use another decorator layer if what is written is string
```
try (PrintWriter pw = new PrintWriter(new BufferedWriter( new FileWriter( output ) ) ) ) {
    for ( String s : content ) {
        pw.println( s );
    }
}
```
but let us stop here. What really matters is the testing per se!
Whenever I am in an interview and asked about design pattern, I wanna say "there is good example of design pattern overuse in Java - Legacy IO based on decorator pattern!".

sebersole:
> LGTM. It's clearly the safest option we have in 5.x.
> 
> Now, about 6.0... Maybe we should consider reverting the default behavior to "truncate" in 6.0, since apparently it's causing problems to some people? Though TBH I don't see the use case for appending to an existing file, so in 6.0 we could just set the behavior to "truncate" **and** remove the configuration property completely... ?

IIRC the issue is generating these from multiple "sources".  JPA allows you to request that these files be populated from the model + some other script (aka, our legacy sql.import).  IIRC truncating simply over-wrote whichever source was processed first with the other.

sebersole:
And let's not forget that this truncate behavior is super easy to get... delete the file first

hibernate-core/src/test/java/org/hibernate/test/schemaupdate/SchemaCreationToOutputScriptTest.java
sebersole: Personally not at all concerned with what tests do so long as they verify the expected behavior

sebersole (approved):
LGTM

yrodiere:
> And let's not forget that this truncate behavior is super easy to get... delete the file first

Easy sure, convenient... not so much. Especially when doing iterative development.

sebersole:
:shrug: 

Anyway, you asked about the back-ground of why so I replied about the reason.

PR #4057, opened 2021-06-18
HHH-14654 Fix for schema validation bug

maesenka:


hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/postgis/PostgisPG9Dialect.java
```diff
@@ -48,6 +48,12 @@ public void contributeTypes(TypeContributions typeContributions, ServiceRegistry
 		support.contributeTypes( typeContributions, serviceRegistry );
 	}
 
+	@Override
+	public boolean equivalentTypes(int typeCode1, int typeCode2) {
+		return super.equivalentTypes( typeCode1, typeCode2 ) ||
+				( support.isSpatial( typeCode1 ) && support.isSpatial( typeCode2 ) );
+	}
+
```
NathanQingyangXu, line 56: Could we move the logic into the parent class to avoid duplication in each child dialect class?

hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/postgis/PGSpatialDialectTrait.java
```diff
@@ -125,4 +127,13 @@ default boolean supportsFiltering() {
 	default boolean supports(SpatialFunction function) {
 		return support.supports( function );
 	}
+
+	/**
+	 * Checks whether the typeCode is (potentially) the code for a spatial type
+	 * @param typeCode the JDBC type code
+	 * @return if the typecode corresponds with a spatialt type
```
NathanQingyangXu, line 134: I did some research here. So there is some subtle difference between 'corresponds to' and 'corresponds with'. Seems 'corresponds to' is more appropriate here for 'corresponds with' means 'comply with' or 'in harmony'.

hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/postgis/PostgisPG9Dialect.java
NathanQingyangXu: Saw it relies on 'support', which is not put into parent class. Ignore my above comment if necessary.

beikov:
Thanks!

PR #4058, opened 2021-06-18
wip/6.0 HHH-14642 Critera verifying - tuple

NathanQingyangXu:
another subtask of https://hibernate.atlassian.net/browse/HHH-14642

The fixing is straightforward. Originally we didn't maintain the tuple entry order for we are using Map. Also, the map key is based on nullable alias so if two tuple entries both have no alias, the last one will override the previous one because HashMap accepts null key.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/select/SqmSelectClause.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/select/SqmSelectClause.java
@@ -8,7 +8,9 @@ package org.hibernate.query.sqm.tree.select;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import org.hibernate.internal.util.collections.CollectionHelper;
 import org.hibernate.query.criteria.JpaSelection;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/tree/select/SqmSelectClause.java
```diff
@@ -25,6 +27,7 @@
 public class SqmSelectClause extends AbstractSqmNode implements SqmAliasedExpressionContainer<SqmSelection>, JpaSelection<Object> {
 	private boolean distinct;
 	private List<SqmSelection<?>> selections;
+	private Set<String> aliases;
```
beikov, line 30: IMO this set of aliases should be built on demand when a selection is added/set. We can replace all calls to `add` or `addSelection` that we currently have with some form of `setSelections` that also does this duplicate check, but over the full list of selections and only once. Also, the check should IMO go deeper, as I think that e.g. `DynamicInstantiation` and nested (not sure if possible) `SqmJpaCompoundSelection` are currently not handled. IMO the duplicate check logic should be moved to a utility class and be used in `SqmDynamicInstantiation`, `SqmJpaCompoundSelection` and `SqmSelectClause`.

beikov (changes requested):

beikov:

PR #4061, opened 2021-06-21
Wip/6.0 re-enable tests from org.hibernate.test.join package

dreab8:


hibernate-core/src/test/java/org/hibernate/orm/test/join/Person.java
```diff
@@ -30,7 +30,6 @@
 @Entity
 @Inheritance(strategy = InheritanceType.SINGLE_TABLE)
 @DiscriminatorColumn( name = "person_type" )
-@DiscriminatorValue( value = "P" )
```
beikov, line 33: I guess this change was by accident?

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -3311,6 +3311,25 @@ protected final void renderCommaSeparated(Iterable<? extends SqlAstNode> express
...
+				expression.accept( this );
+				separator = COMA_SEPARATOR;
+			}
+		}
+	}
+
+	private boolean isNullValue(SqlAstNode expression) {
+		if(expression instanceof QueryLiteral){
+			QueryLiteral literal = (QueryLiteral) expression;
+			return literal.getLiteralValue().equals( "null" ) || literal.getLiteralValue().equals( "not null" );
```
beikov, line 3328: Wouldn't this also be triggered if I did a query like `from Entity a where a.name in ('null', 'not null')` which is completely unrelated? IMO we should introduce special `QueryLiteral` objects or a special expression class for which we can do an identity check here.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -4000,13 +4020,46 @@ else if ( !supportsRowValueConstructorSyntaxInInList() ) {
...
+			if ( nullExpressions > 0 && !areOnlyNullExpressions ) {
+				appendSql( OPEN_PARENTHESIS );
+			}
 			inListPredicate.getTestExpression().accept( this );
-			if ( inListPredicate.isNegated() ) {
-				appendSql( " not" );
+
+			if ( nullExpressions > 0 ) {
+				if ( !areOnlyNullExpressions ) {
+				}
```
beikov, line 4039: Unnecessary if?

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -4000,13 +4020,46 @@ else if ( !supportsRowValueConstructorSyntaxInInList() ) {
...
+			if ( !areOnlyNullExpressions ) {
+				if ( negated ) {
+					appendSql( " not" );
+				}
+				appendSql( " in (" );
+				renderNonNullCommaSeparated( listExpressions );
+				appendSql( CLOSE_PARENTHESIS );
+			}
+			if ( nullExpressions > 0 && !areOnlyNullExpressions ) {
+				appendSql( CLOSE_PARENTHESIS );
```
beikov, line 4061: Move this into the if above and eliminate the second conjunction.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -3311,6 +3311,25 @@ protected final void renderCommaSeparated(Iterable<? extends SqlAstNode> express
 		}
 	}
 
+	protected final void renderNonNullCommaSeparated(Iterable<? extends SqlAstNode> expressions) {
+		String separator = NO_SEPARATOR;
+		for ( SqlAstNode expression : expressions ) {
+			if( !isNullValue( expression ) ){
```
beikov, line 3317: Formatting

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -4000,13 +4020,46 @@ else if ( !supportsRowValueConstructorSyntaxInInList() ) {
 			}
 		}
 		else {
+			int nullExpressions = 0;
```
beikov, line 4023: How about we do all of this only when the LHS has a model part type of `java.lang.Class`?

beikov (changes requested):

hibernate-core/src/test/java/org/hibernate/orm/test/join/Person.java
dreab8: yes it was :)

hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
```diff
@@ -983,10 +988,32 @@ private Predicate createDiscriminatorPredicate(
...
+				final NullnessPredicate nullnessPredicate = new NullnessPredicate( sqlExpression );
+				if ( hasNonNull ) {
+					nullnessPredicate.isNegated();
+				}
+				junction.add( nullnessPredicate );
+				junction.add( p );
+				return junction;
 			}
-			return new InListPredicate( sqlExpression, values );
+			return p;
```
beikov, line 1016:
```suggestion
			final Predicate p = new InListPredicate( sqlExpression, values );
			if ( hasNull || hasNonNull ) {
				final Junction junction = new Junction(
						Junction.Nature.DISJUNCTION
				);
				// This essentially means we need to select everything, so we don't need a predicate at all
				if ( hasNull && hasNonNull ) {
					return junction;
				}
				final NullnessPredicate nullnessPredicate = new NullnessPredicate( sqlExpression, hasNonNull );
				junction.add( nullnessPredicate );
				junction.add( p );
				return junction;
			}
			return p;
```

dreab8: Thanks @beikov changed it

beikov (approved):

PR #4063, opened 2021-06-22
HHH-14657 Use the compile instead of runtime classpath (5.4 backport)

markusheiden:
The runtime classpath contains the dependencies as jars that are not yet built when the plugin needs them. So use the compile classpath plus the compiled classes of the current project.

Since Spring Boot 2.6 (which uses Hibernate 5.5) will be released not before November, it would be nice to have this fix available for Spring Boot 2.5 (which uses Hibernate 5.4).


Sanne:
Ok, sounds reasonable I'll apply this. Thanks!

PR #4064, opened 2021-06-22
HHHH-14688 Get IdentifierGenerator from BeanContainer if not registered

quaff:
https://hibernate.atlassian.net/browse/HHH-14688


Sanne:
Nice, looks useful. Could you add a test please to prevent future regressions?

quaff:
> Nice, looks useful. Could you add a test please to prevent future regressions?

Please review.

quaff:
Could someone take a look at this? It's a small improvement.

sebersole (approved):
Looks good.  Thanks!

Sanne:
thanks @quaff & @sebersole !

quaff:
> thanks @quaff & @sebersole !

@Sanne It should be backported to 5.4, then Spring Boot users will be benefit, WDYT?

Sanne:
I would prefer not since it's a new feature; can't I assume that Spring Boot will also upgrade to 5.5 sometime soon?

PR #4068, opened 2021-06-22
Revert to static bytebuddy caches, attempt to reuse generated Class d…

Sanne:
…efinitions across ORM instances

Just an experiment for @scottmarlow to play with


scottmarlow:
Testing with the main branch showed that only one HibernateProxy class was defined for the test Employee entity (instead of the previous 50 Employee HibernateProxy class definitions.

Will try cherry-picking this change onto the 5.3 branch and test with that.

Sanne:
closing this experiment, I'll propose better ways to go forward.

PR #4073, opened 2021-06-23
HHH-14697 TimePropertyTest fails on MySQL 8.0 with a ComparisonFailure

RanabirChakraborty:
Issue: https://hibernate.atlassian.net/browse/HHH-14697

JBEAP Issue: https://issues.redhat.com/browse/JBEAP-22070


RanabirChakraborty:
@gbadner can you check this once :)

RanabirChakraborty:
@gbadner Can you pleas take a look :)

gbadner:
@RanabirChakraborty, thanks for the new PR. I'm confused though, because I am not seeing a failure running the test without your fix using `MySQL8Dialect`.

gbadner:
The issue reproduces using mysql-connector-java-8.0.23, but does not reproduce using the current version used as a dependency in 5.3.

I will merge this issue, since mysql-connector-java-8.0.23 is used for testing EAP 7.4. I will deal with upgrading the mysql-connector-java dependency in a separate jira.

PR #4074, opened 2021-06-24
Wip/6.0 Re-enabled additional tests

dreab8:


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
```diff
@@ -465,10 +465,10 @@ public void resolveInstance(RowProcessingState rowProcessingState) {
 
 		final PersistenceContext persistenceContext = session.getPersistenceContext();
 		final Object proxy = getProxy( persistenceContext );
-
 		// Special case map proxy to avoid stack overflows
 		// We know that a map proxy will always be of "the right type" so just use that object
-		if ( proxy != null && ( proxy instanceof MapProxy || concreteDescriptor.isInstance( proxy ) ) ) {
+		if ( proxy != null && ( proxy instanceof MapProxy
+				|| entityDescriptor.getJavaTypeDescriptor().getJavaTypeClass().isInstance( proxy ) ) ) {
```
beikov, line 471: Why the change from `concreteDescriptor` to `entityDescriptor`? I don't remember the details anymore, but it seemed right to me at the time when I wrote this to use the `concreteDescriptor` instead of the `entityDescriptor`.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/SqmCorrelatedRootJoin.java
```diff
@@ -55,7 +49,7 @@ public SqmCorrelatedRootJoin(
 
 	@Override
 	public SqmRoot<T> getCorrelationParent() {
-		return null;
+		return this;
```
beikov, line 52: Is there a failing test due to how it was before? There are other classes following the same approach, that's why I am asking

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/from/TableGroup.java
```diff
@@ -41,6 +41,8 @@
 	String getSourceAlias();
 
 	List<TableGroupJoin> getTableGroupJoins();
+	
+	boolean hasNonInnerJoins();
```
beikov, line 45: I would prefer a name like `isOptional` or `isOuterJoined`. Also, as far as I can say, the way this is modeled right now is not entirely correct. IMO this should only be `true` if there is a `TableGroupJoin` referring to this `TableGroup` and that  `TableGroupJoin` uses anything but `INNER` or the parent table group is `optional`. This can be determined at `TableGroup` construction time.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -1891,8 +1891,8 @@ protected void consumeFromClauseRoot(SqmRoot<?> sqmRoot) {
 		fromClauseIndex.register( sqmRoot, tableGroup );
 		currentQuerySpec().getFromClause().addRoot( tableGroup );
 
-		consumeExplicitJoins( sqmRoot, tableGroup );
 		consumeReusablePaths( sqmRoot, tableGroup );
+		consumeExplicitJoins( sqmRoot, tableGroup );
```
beikov, line 1895: What's the reason for changing the switch here and below? I'm not sure if this is still correct when we first render reusable paths and only then the explicit joins.

`from Entity a right join a.association1 a1 where a.association2.name = 'abc'`

would render as 

```
select *
from tbl a
inner join association2 a2 on a.a2 = a2.id
right join association1 a1 on a.a1 = a1.id
```

which might be fine, but I just want to bring this up so we can think about the consequences of this again. If the tests run through, it's probably correct, as I think @jwgmeligmeyling tried to explicitly provoke cases where join ordering matters in the tests that you moved.

beikov (changes requested):

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
jwgmeligmeyling: Does Hibernate have implicit joins and is `a.association2.name` actually allowed? 😉 

In the example of derefencing a property from the first join node there isn't much that can go wrong, because either the dereference introduces a left join (which does not remove rows) or it generates an inner join (which might remove rows, but only ones that would already be removed regardless of join order).

Now it gets a bit interesting when dereferencing / implicit joining a property from a join node that is joined by a right or full outer join (`a1` in this example).

However, in the first place it I'd think that any dereference/implicit join can only occur after the join node it references. Otherwise you simply end up with invalid SQL. So if these joins need to appear after the join node they reference already, I think actually it doesn't make much of a difference between join nodes that are inner/left joined and ones dat are right/full outer joined.

jwgmeligmeyling: By the way the tests did indeed try to provike cases where join ordering matters, so as long as the tests are just moved and the assertions are not altered it should give a fair bit of confidence in the change.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
dreab8: `DiscriminatorTest#testLoadSuperclassProxyPolymorphicAccess` was failing without this change

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
dreab8: without these change a test in `OuterJoinTest` (if I remember well `testJoinOrderWithRightJoinWithIdDereference`) was failing with the exception `Could not locate TableGroup exception`

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/SqmCorrelatedRootJoin.java
dreab8: the failing test id `OuterJoinTest#testJoinOrderWithRightJoinWithInnerImplicitJoins`

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/from/TableGroup.java
dreab8: I'm going to change the method name and to determine its value as you suggest

beikov: You should be able to determine this in e.g. `ToOneAttributeMapping#createTableGroupJoin`. It is outer joined if `sqlAstJoinType != INNER || lhs.isOuterJoined()`. You do something similar for the `PluralAttributeMapping` as well and other instantiations of `TableGroup`

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/4080

PR #4075, opened 2021-06-25
Filter clause for aggregate functions in HQL

jrenaat:

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
@@ -3548,8 +3548,9 @@ public class SemanticQueryBuilder<R> extends HqlParserBaseVisitor<Object> implem
 	public SqmExpression<?> visitMaxFunction(HqlParser.MaxFunctionContext ctx) {
 		final SqmExpression<?> arg = (SqmExpression<?>) ctx.expression().accept( this );
 		//ignore DISTINCT
-		return getFunctionDescriptor("max").generateSqmExpression(
-				arg,
+		return getFunctionDescriptor("max").generateAggregateSqmExpression(
+				singletonList( arg ),
+				getFilterExpression( ctx.filterClause() ),
 				(AllowableFunctionReturnType<?>) arg.getNodeType(),
 				creationContext.getQueryEngine(),
 				creationContext.getJpaMetamodel().getTypeConfiguration()
@@ -3560,8 +3561,9 @@ public class SemanticQueryBuilder<R> extends HqlParserBaseVisitor<Object> implem
 	public SqmExpression<?> visitMinFunction(HqlParser.MinFunctionContext ctx) {
 		final SqmExpression<?> arg = (SqmExpression<?>) ctx.expression().accept( this );
 		//ignore DISTINCT
-		return getFunctionDescriptor("min").generateSqmExpression(
-				arg,
+		return getFunctionDescriptor("min").generateAggregateSqmExpression(
+				singletonList( arg ),
+				getFilterExpression( ctx.filterClause() ),
 				(AllowableFunctionReturnType<?>) arg.getNodeType(),
 				creationContext.getQueryEngine(),
 				creationContext.getJpaMetamodel().getTypeConfiguration()
@@ -3575,8 +3577,9 @@ public class SemanticQueryBuilder<R> extends HqlParserBaseVisitor<Object> implem
 				? new SqmDistinct<>(arg, getCreationContext().getNodeBuilder())
 				: arg;
 
-		return getFunctionDescriptor("sum").generateSqmExpression(
-				argument,
+		return getFunctionDescriptor("sum").generateAggregateSqmExpression(
+				singletonList( argument ),
+				getFilterExpression(ctx.filterClause()),
 				null,
 				creationContext.getQueryEngine(),
 				creationContext.getJpaMetamodel().getTypeConfiguration()
@@ -3593,8 +3596,13 @@ public class SemanticQueryBuilder<R> extends HqlParserBaseVisitor<Object> implem
 
 		final SqmExpression<?> argument = (SqmExpression<?>) ctx.predicate().accept( this );
 
-		return getFunctionDescriptor("every").generateSqmExpression(
-				argument,
+		if ( argument instanceof SqmSubQuery<?> && ctx.filterClause() != null ) {
+			throw new SemanticException( "Quantified expression cannot have a filter clause!" );
+		}
+
+		return getFunctionDescriptor("every").generateAggregateSqmExpression(
+				singletonList( argument ),
+				getFilterExpression( ctx.filterClause() ),
 				resolveExpressableTypeBasic( Boolean.class ),
 				creationContext.getQueryEngine(),
 				creationContext.getJpaMetamodel().getTypeConfiguration()
@@ -3611,8 +3619,13 @@ public class SemanticQueryBuilder<R> extends HqlParserBaseVisitor<Object> implem
 
 		final SqmExpression<?> argument = (SqmExpression<?>) ctx.predicate().accept( this );
 
-		return getFunctionDescriptor("any").generateSqmExpression(
-				argument,
+		if ( argument instanceof SqmSubQuery<?> && ctx.filterClause() != null ) {
+			throw new SemanticException( "Quantified expression cannot have a filter clause!" );
+		}
+
+		return getFunctionDescriptor("any").generateAggregateSqmExpression(
+				singletonList( argument ),
+				getFilterExpression( ctx.filterClause() ),
 				resolveExpressableTypeBasic( Boolean.class ),
 				creationContext.getQueryEngine(),
 				creationContext.getJpaMetamodel().getTypeConfiguration()
@@ -3626,8 +3639,9 @@ public class SemanticQueryBuilder<R> extends HqlParserBaseVisitor<Object> implem
 				? new SqmDistinct<>( arg, getCreationContext().getNodeBuilder() )
 				: arg;
 
-		return getFunctionDescriptor("avg").generateSqmExpression(
-				argument,
+		return getFunctionDescriptor("avg").generateAggregateSqmExpression(
+				singletonList( argument ),
+				getFilterExpression( ctx.filterClause() ),
 				resolveExpressableTypeBasic( Double.class ),
 				creationContext.getQueryEngine(),
 				creationContext.getJpaMetamodel().getTypeConfiguration()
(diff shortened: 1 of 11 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/sql/exec/HqlDeleteExecutionTests.java
```diff
@@ -48,8 +48,8 @@ public void testSimpleDelete(SessionFactoryScope scope) {
 	@Test
 	public void testSimpleRestrictedDelete(SessionFactoryScope scope) {
 		scope.inTransaction(
-				session -> session.createQuery( "delete BasicEntity where data = :filter" )
-						.setParameter( "filter", "abc" )
+				session -> session.createQuery( "delete BasicEntity where data = :thefilter" )
```
beikov, line 51: You have to add this token to the `identifier` rule at the end of `HqlParser.g4` to avoid the changes in this file

hibernate-core/src/test/java/org/hibernate/orm/test/sql/exec/HqlUpdateExecutionTests.java
```diff
@@ -85,9 +85,9 @@ public void testSimpleUpdateWithData(SessionFactoryScope scope) {
 	@Test
 	public void testSimpleRestrictedUpdate(SessionFactoryScope scope) {
 		scope.inTransaction(
-				session -> session.createQuery( "update BasicEntity set data = :p where data = :filter" )
+				session -> session.createQuery( "update BasicEntity set data = :p where data = :thefilter" )
```
beikov, line 88: You have to add this token to the `identifier` rule at the end of `HqlParser.g4` to avoid the changes in this file

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/AggregateFilterClauseTest.java
```diff
@@ -0,0 +1,234 @@
...
+							},
+							"The every predicate doesn't support the filter clause"
+					);
+					assertThrows(
+							IllegalArgumentException.class,
+							() -> {
+								Query q2 = session.createQuery( "select all(eob.id > 0) filter(where eob.theBoolean = true) from EntityOfBasics eob" );
+							},
+							"The all predicate doesn't support the filter clause"
+					);
```
beikov, line 229: Actually, the versions that accept predicates instead of subqueries should support the filter clause, as these functions are just special boolean aggregates. Sorry for the confusion before, but I didn't look carefully enough. The negative test can stay though for when a subquery is used as argument.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/AggregateFilterClauseTest.java
```diff
@@ -0,0 +1,234 @@
...
+		);
+	}
+
+	@Test
+	public void testSimpleSum(SessionFactoryScope scope) {
+		scope.inTransaction(
+				session -> {
+					Long expected = 31L;
+					Query q = session.createQuery( "select sum(eob.theInt) from EntityOfBasics eob" );
+					assertEquals( q.getSingleResult(), expected, "expected " + expected + ", got " + q.getSingleResult() );
```
beikov, line 93: I don't know what kind of method you are calling here (I thought there is only a 3 args `assertEquals` method) but what you seem to anticipate is auto-generated anyway. So please switch to using the following pattern in this class.

```
assertEquals( 31L, q.getSingleResult() );
```

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
```diff
@@ -3653,14 +3657,31 @@ public Object visitStrFunction(HqlParser.StrFunctionContext ctx) {
...
 
+	private SqmPredicate getFilterExpression(HqlParser.FilterClauseContext filterClauseCtx) {
+		if (filterClauseCtx == null) {
+			return null;
+		}
+		return (SqmPredicate) filterClauseCtx.whereClause().predicate().accept( this );
+	}
+
+	@Override
+	public SqmExpression<?> visitFilterClause(HqlParser.FilterClauseContext ctx) {
```
beikov, line 3677: You can remove this as I think you handled all uses above.

beikov (changes requested):

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/AggregateFilterClauseTest.java
jrenaat: Not sure I understand the issue; I'm calling the assertEquals that accepts the expected result, the actual result, and a message in case of failing the boolean test to give some more info as to why it failed (w/ respect to "expected 'true', was 'false' the default no-message method gives), respectively.
I did wrongly switch the first 2 params though (corrected), but again, I don't see the problem?

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
jrenaat: Isn't this needed in case 'someone' were to want to walk the tree and drill down into the filter clause?

hibernate-core/src/test/java/org/hibernate/orm/test/sql/exec/HqlDeleteExecutionTests.java
jrenaat: Thanks, did not know this.

hibernate-core/src/test/java/org/hibernate/orm/test/sql/exec/HqlUpdateExecutionTests.java
jrenaat: Thanks, did not know this.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/AggregateFilterClauseTest.java
jrenaat: Just to make sure; are you referring only to the every/all + any/some functions, or are there any more aside those?
Could you clarify what exactly you have in mind for the negative test?

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/AggregateFilterClauseTest.java
beikov: My eyes were too tired apparently 😆 I thought the comma you used in the string was an argument separator. Anyway, in case of a failure, you will get exactly the same message that you are concatenating here, so no need for constructing and passing a message.

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
beikov: Every use of the rule is handled explicitly. This rule is only invoked if the default handler for the AST node is visited, which is never the case.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/AggregateFilterClauseTest.java
beikov: If a predicate is passed as argument to these functions, the filter clause should be allowed i.e. the specific test case you have here, should not fail but actually succeed. On the other hand, an expression like `all(select 1) filter (where ..)` should fail though, because that kind of expression is actually only relevant for quantified predicate comparison where the filter clause is illegal.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/AggregateFilterClauseTest.java
jrenaat: I disagree, the default (no-message) assertEquals only says that true (equality) was expected and false (non-equality) was the actual; which really doesn't say anything, the message provides what the actual expected result is, and what the query returns, so that saves a debugging trip to get this exact information.

beikov: It's fine

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/AggregateFilterClauseTest.java
jrenaat: I finally took this 'negative' test out, because I couldn't get any trial based on your last comment to work; if you want to have a go at it, please do, you prob. have a better idea of what you had in mind.

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
```diff
@@ -3621,8 +3625,9 @@ public Object visitStrFunction(HqlParser.StrFunctionContext ctx) {
 
 		final SqmExpression<?> argument = (SqmExpression<?>) ctx.predicate().accept( this );
 
-		return getFunctionDescriptor("any").generateSqmExpression(
```
beikov, line 3624: What I meant in https://github.com/hibernate/hibernate-orm/pull/4075#discussion_r660577299 is that you add the following:

```
if ( argument instanceof SqmSubQuery<?> && ctx.filterClause() != null ) {
    throw new SemanticException( "Quantified expression can not have a filter clause!" );
}
```

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
```diff
@@ -3603,8 +3606,9 @@ public Object visitStrFunction(HqlParser.StrFunctionContext ctx) {
 
 		final SqmExpression<?> argument = (SqmExpression<?>) ctx.predicate().accept( this );
 
-		return getFunctionDescriptor("every").generateSqmExpression(
```
beikov, line 3606: What I meant in https://github.com/hibernate/hibernate-orm/pull/4075#discussion_r660577299 is that you add the following:

```
if ( argument instanceof SqmSubQuery<?> && ctx.filterClause() != null ) {
    throw new SemanticException( "Quantified expression can not have a filter clause!" );
}
```

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/AggregateFilterClauseTest.java
sebersole: I agree.  I prefer more contextual message here

jrenaat: With whom ? ;)

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/AggregateFilterClauseTest.java
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	@Test
+	public void testIllegalSubquery(SessionFactoryScope scope) {
+		scope.inTransaction(
+				session -> {
+					Exception e = Assertions.assertThrows(
+							IllegalArgumentException.class,
+							() -> {
+								session.createQuery( "select every( eob.theInteger > 0 ) filter ( where select 1 ) from EntityOfBasics eob" );
```
beikov, line 248:
```suggestion
								session.createQuery( "select every( select 1 ) filter ( where eob.theInteger > 0 ) from EntityOfBasics eob" );
```

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/AggregateFilterClauseTest.java
```diff
@@ -0,0 +1,268 @@
...
+					);
+					assertEquals( SemanticException.class, e.getCause().getClass() );
+				}
+		);
+		scope.inTransaction(
+				session -> {
+					Exception e = Assertions.assertThrows(
+							IllegalArgumentException.class,
+							() -> {
+								session.createQuery( "select any( eob.theInteger > 0 ) filter ( where select 1 ) from EntityOfBasics eob" );
```
beikov, line 260:
```suggestion
								session.createQuery( "select any( select 1 ) filter ( where eob.theInteger > 0 ) from EntityOfBasics eob" );
```

beikov (changes requested):
Just a small change to assert the grammar is correct and then this is done

beikov (approved):

PR #4076, opened 2021-06-26
Fix a bug that has existed for more than 15 years. 

cabbage89:
Fix a bug that has existed for more than 15 years. When column names use database reserved words, such as when using `groups` as column names under mysql8, the generated SQL script cannot be executed in the database because the column names are not quoted


gavinking:
Could you please include some test showing in what circumstance this bug arises. Thanks.

cabbage89:
> Could you please include some test showing in what circumstance this bug arises. Thanks.

OK ,I will resubmit this PR according to the contribution requirements

PR #4081, opened 2021-07-05
HHH-14706 Improve error message on incompatible types due to mismatched classloader

Sanne:
- https://hibernate.atlassian.net/browse/HHH-14706


gavinking:
OMG yes, thanks!

PR #4082, opened 2021-07-05
HHH-14707 Upgrade to Byte Buddy 1.11.8

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14707

Build against JDK18: https://ci.hibernate.org/view/Personal%20jobs/job/hibernate-orm-personal-yoann-jdk18/1/

PR #4084, opened 2021-07-06
HHH-14709 Upgrade to Gradle 6.7.1 and move to Gradle's built-in way of testing Java modules

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14709

I did the work while attempting to solve the problems on #4082, but it turned out it wasn't necessary.

Since it's done, we might as well merge it?


yrodiere:
I'll take that as a yes; merging!

PR #4089, opened 2021-07-08
Wip/6.0 re enabling tests 

dreab8:


hibernate-core/src/main/java/org/hibernate/query/sqm/function/SelfRenderingSqmAggregateFunction.java
```diff
@@ -47,10 +48,10 @@ public SelfRenderingFunctionSqlAstExpression convertToSqlAst(SqmToSqlAstConverte
 		return new SelfRenderingAggregateFunctionSqlAstExpression(
 				getFunctionName(),
 				getRenderingSupport(),
-				resolveSqlAstArguments( getArguments(), walker ),
+				resolveSqlAstArguments( getArguments(), (BaseSqmToSqlAstConverter) walker ),
```
beikov, line 51: Seems unnecessary, or am I missing something?

hibernate-core/src/main/java/org/hibernate/query/sqm/function/SelfRenderingSqmAggregateFunction.java
```diff
@@ -47,10 +48,10 @@ public SelfRenderingFunctionSqlAstExpression convertToSqlAst(SqmToSqlAstConverte
...
 				filter == null ? null : (Predicate) filter.accept( walker ),
 				resultType,
-				getMappingModelExpressable( walker, resultType )
+				getMappingModelExpressable( (BaseSqmToSqlAstConverter) walker, resultType )
```
beikov, line 54: Seems unnecessary, or am I missing something?

hibernate-core/src/main/java/org/hibernate/query/sqm/function/SelfRenderingSqmFunction.java
```diff
@@ -60,7 +62,21 @@ public FunctionRenderingSupport getRenderingSupport() {
 
 		final ArrayList<SqlAstNode> sqlAstArguments = new ArrayList<>( sqmArguments.size() );
 		for ( SqmTypedNode<?> sqmArgument : sqmArguments ) {
-			sqlAstArguments.add( (SqlAstNode) ( (SqmVisitableNode) sqmArgument ).accept( walker ) );
+			if ( sqmArgument instanceof SqmParameter ) {
```
beikov, line 65: What's the reason for this? There are legitimate uses of functions where a tuple should be passed in rather than the expanded list.

beikov (changes requested):
Please also rebase against wip/6.0 to fix the issues we had with running on Java 8.

hibernate-core/src/test/java/org/hibernate/orm/test/connections/LazyLoadingConnectionCloseTest.java
```diff
@@ -97,39 +98,44 @@ public void setUp() {
 		} );
 	}
 
+	@AfterAll
+	public void tearDown() {
+		inTransaction( entityManager -> {
+			entityManager.createNativeQuery( "delete from ChildEntity" ).executeUpdate();
+			entityManager.createNativeQuery( "delete from SimpleEntity" ).executeUpdate();
```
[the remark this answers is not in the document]
dreab8: yeap you are right. going to fix it. Thanks

hibernate-core/src/main/java/org/hibernate/query/sqm/function/SelfRenderingSqmAggregateFunction.java
dreab8: you are right, going to remove the cast. Thanks

hibernate-core/src/main/java/org/hibernate/query/sqm/function/SelfRenderingSqmAggregateFunction.java
dreab8: some here, thanks

hibernate-core/src/main/java/org/hibernate/query/sqm/function/SelfRenderingSqmFunction.java
dreab8: I changed this to fix `org.hibernate.orm.test.ordered.HqlOrderByIdsTest` where the following query
```
List<Person> persons = entityManager.createQuery(
					"SELECT p " +
							"FROM Person p " +
							"WHERE p.id IN (:ids) " +
							"ORDER BY FIELD(id, :ids) ", Person.class )
					.setParameter( "ids", Arrays.asList( 3L, 1L, 2L ) )
					.getResultList();
```
was failing, when we try to create the `JdbcParameterBindings` in the `SqmUtil#createJdbcParameterBindings` we have an assertion failure (line 233) `assert expansions.size() == bindValues.size() - 1` because for the second one of the `:ids`   `bindValues.size()` is 0

dreab8: by the way not sure that my solution is the correct /better one

beikov: Hmm, IMO the function should in the rendering expand the tuple manually instead of doing this expansion in the general rendering code.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -785,7 +785,7 @@ public Expression resolveSqlExpression(
 				if ( sqmAssignment.getValue() instanceof SqmParameter ) {
 					final SqmParameter<?> sqmParameter = (SqmParameter<?>) sqmAssignment.getValue();
 
-					consumeSqmParameter(
+					consumeSingleSqmParameter(
```
beikov, line 788: Was this change on purpose?

dreab8: we renamed the method

beikov: I know, but it seems like we might have forgotten to replace this one with `consumeSqmParameter` again. Nevermind though, it seems odd to have a multi-valued parameter in the SET clause.

beikov (approved):

PR #4098, opened 2021-07-15
HHH-14728 Include CamelCaseToUnderscoresNamingStrategy from Spring Boot

Sanne:
Hi @philwebb , @mbhave I assume you'd be cool with this?

Please let me know, as I've heard people are commonly copying this from Spring Boot: it's probably more convenient for all if we distribute this naming strategy with ORM.

 - https://hibernate.atlassian.net/browse/HHH-14728


philwebb:
Adding it here would be great! We can then drop it from our codebase sometime in the future.

The code added here is under the LGPL license where as the original is Apache 2.0. Since It's such a trivial piece of code and @mbhave and myself are the authors I think the license change is fine. Consider this comment permission, but please let me know if you need something more formal.

Sanne:
@philwebb many thanks! No need for more formality :)

PR #4101, opened 2021-07-16
HHH-14730 Failure during lazy loading of bytecode-enhancement proxy triggered by the loading of a collection with eager references to that proxy

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14730

See `org.hibernate.orm.test.mapping.lazytoone.collectioninitializer.InitLazyToOneWithinPaddedCollectionInitializationAllowProxyTest#testTwoReferences` for the reproducer; I think that's the easiest way to understand the scenario (though it's still complex...).


dreab8 (approved):
It looks good to me :+1:

yrodiere:
Thanks @dreab8! Merged.

PR #4103, opened 2021-07-19
HHH-14739 - Implement ILike support

jrenaat:

### DIFF
--- a/documentation/src/main/asciidoc/userguide/chapters/query/hql/HQL.adoc
+++ b/documentation/src/main/asciidoc/userguide/chapters/query/hql/HQL.adoc
@@ -1148,6 +1148,8 @@ The supported aggregate functions are:
 For floating point values (other than `BigDecimal`) the result type is `Double`.
 For `BigInteger` values, the result type is `BigInteger`. For `BigDecimal` values, the result type is `BigDecimal`.
 
+Aggregations often appear with grouping. For information on grouping see <<hql-group-by>>.
+
 [[hql-aggregate-functions-example]]
 .Aggregate function examples
 ====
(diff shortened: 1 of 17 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmTreePrinter.java
```diff
@@ -838,8 +838,9 @@ public Object visitBetweenPredicate(SqmBetweenPredicate predicate) {
 
 	@Override
 	public Object visitLikePredicate(SqmLikePredicate predicate) {
+		String likeType = predicate.isCaseSensitive() ? "like" : "ilike";
```
beikov, line 841:
```suggestion
		final String likeType = predicate.isCaseSensitive() ? "like" : "ilike";
```

beikov (changes requested):

beikov (approved):

gavinking:
How many databases does this work on?

beikov:
> How many databases does this work on?

It will work on all databases because we can emulate it with `LOWER(lhs) LIKE LOWER(rhs)`. Also see `org.hibernate.dialect.Dialect#getLowercaseFunction` which was used in the legacy Criteria implementation.

documentation/src/main/asciidoc/userguide/chapters/query/hql/HQL.adoc
```diff
@@ -1780,6 +1791,20 @@ include::{sourcedir}/HQLTest.java[tags=hql-like-predicate-escape-example]
...
+
+Performs a case-insensitive like comparison on string values. The syntax is:
+
+[[hql-ilike-predicate-bnf]]
+[source, JAVA, indent=0]
+----
+include::{extrasdir}/predicate_ilike_bnf.txt[]
+----
+
+The semantics are identical to those of the aforementioned <<hql-like-predicate>>, with the sole difference that the comparison is now case insensitive. Note that for this to work the database will have to be set to a case sensitive collation.
```
beikov, line 1805: >Note that for this to work the database will have to be set to a case sensitive collation.

Well, that is not really true I think. The case insensitive comparison works regardless of the collation. The case sensitive like is sensible to the collation though.

jrenaat: Yeah, you're right, i'll scratch this.

beikov (approved):

beikov (approved):

PR #4105, opened 2021-07-23
introduce JDBCTypeDescriptor.getDefaultSqlTypeCode()

gavinking:
The goal of this PR is to cleanly separate usages of the JDBC type code in interactions with the JDBC API, with usages of the JDBC type code to identify a SQL column type in schema export.

This should fix a bug I introduced a while back.

Also rename `sqlType` -> `jdbcTypeCodes` in Type hierarchy for consistency.

See https://github.com/hibernate/hibernate-orm/discussions/4088

cc @beikov @sebersole


hibernate-core/src/main/java/org/hibernate/type/Type.java
```diff
@@ -108,7 +108,7 @@
 	 *
 	 * @throws MappingException Generally indicates an issue accessing the passed mapping object.
 	 */
-	int[] sqlTypes(Mapping mapping) throws MappingException;
+	int[] getSqlTypeCodes(Mapping mapping) throws MappingException;
```
beikov, line 111: As much as I like this change, I think this is a backwards compatibility concern. Not sure if that is something we are ok with, but maybe revert the renaming here?

hibernate-core/src/main/java/org/hibernate/type/SingleColumnType.java
```diff
@@ -20,7 +20,7 @@
  */
 public interface SingleColumnType<T> extends Type {
 
-	int sqlType();
+	int getJdbcTypeCode();
```
beikov, line 23: Same backwards compatibility concern here as for `Type`.

beikov (commented):
The PR looks good so far, but I am missing some changes to `JdbcTypeDescriptorRegistry`. IMO the default key for registration should be the sqlTypeCode to be able to have e.g. descriptors for both GIS types and JSON types which both map to the Jdbc type code `Types.OTHER`

gavinking:
Yeah, I guess that's right.

hibernate-core/src/main/java/org/hibernate/type/Type.java
gavinking: Depends what @sebersole thinks.

gavinking:
>> IMO the default key for registration should be the `sqlTypeCode` to be able to have

> Yeah, I guess that's right.

Well, perhaps not, actually, that's unclear.

Currently the only type we have were the SQL and JDBC types differ is `DoubleTypeDescriptor`. And for `DoubleTypeDescriptor` the current implementation does what we want.

Now, naturally we can special-case `DoubleTypeDescriptor` in `JdbcTypeDescriptorBaseline`, but it just seems a bit suspicious that in the one case where there's a difference, the correct rule is opposite of the one you're suggesting.

hibernate-core/src/main/java/org/hibernate/type/Type.java
sebersole: I do think it would be nice to minimize changes as much as we can.

Here though, I think we should just make this change.  This contract is changing already wrt read-from-jdbc , so implementors and consumers are going to have to change.  If we do the whole deprecation paradigm, we are just making them make changes again later if/when we remove the deprecated form)

sebersole:
> IMO the default key for registration should be the sqlTypeCode to be able to have e.g. descriptors for both GIS types and JSON types which both map to the Jdbc type code `Types.OTHER`

Nothing says the "conflicts" have to register under the key for `OTHER`.  In fact, the point of the dynamic registry is that they wouldn't.

beikov (approved):

PR #4106, opened 2021-07-23
HHH-14744 : Refactor contextual information for SchemaManagementTool to be more easily extended by Hibernate Reactive

gbadner:
https://hibernate.atlassian.net/browse/HHH-14744


hibernate-core/src/main/java/org/hibernate/tool/schema/spi/SchemaManagementTool.java
```diff
@@ -32,4 +43,25 @@
...
+				ddlTransactionIsolator,
+				defaultCatalog,
+				defaultSchema,
+				databaseObjectAccess
+		);
+	}
+
+	default InformationExtractor createInformationExtractor(ExtractionContext extractionContext) {
+		return new InformationExtractorJdbcDatabaseMetaDataImpl( extractionContext );
+	}
```
dreab8, line 66: what about creating an interface `org.hibernate.tool.schema.extract.spi.ExtractionTool` and add this two methods there?

gbadner: @dreab8, I was thinking about something like that as well. I think that `#setCustomDatabaseGenerationTarget` could be included as well. I ended up creating a new `ReactiveSchemaManagementTool` in https://github.com/hibernate/hibernate-reactive/pull/903, so I think it would work well to encapsulate this stuff in an interface.

Before making that change, I'd like to take a look at `SchemaValidator` to see what is needed there in case something else needs to be added.

gbadner: @dreab8, thinking more about it, `ExtractionTool#getDatabaseGenerationTarget` doesn't make sense. I'm leaning more towards `SchemaManagementToolContext`. I'm doing some more refactoring and will have an update soon.

gbadner: @dreab8, there are complications with my idea, so I have added a [commit](https://github.com/hibernate/hibernate-orm/pull/4106/commits/e4390a014c8f189e57c20261b8bc30e8edf50958) that moves those methods into `ExtractionTool` as you suggested.

gbadner:
I rebased main and force-pushed.

hibernate-core/src/main/java/org/hibernate/tool/schema/spi/SchemaManagementTool.java
gbadner: @dreab8, I didn't create default methods in `ExtractionTool` because all of the interfaces are annotated with `@Incubating`.  Is that OK with you?

dreab8: Hi @gbadner , it looks good to me :+1:

gbadner:
I added some Javadoc and some other minor changes to be more consistent with `DatabaseMetaData` methods.

I also force-pushed.

gbadner:
@Sanne, I've added you as a reviewer in case @dreab8 is not available.

Sanne:
Thanks @gbadner , this is larger than I expected and is changing some internal services which we use in Quarkus. I'll need to verify the feasibility of upgrading the Quarkus integration code as well.

gbadner:
> Thanks @gbadner , this is larger than I expected and is changing some internal services which we use in Quarkus. I'll need to verify the feasibility of upgrading the Quarkus integration code as well.

@Sanne, I can create default methods. I didn't do that because interfaces were marked `@Incubating`. I wasn't thinking of Quarkus though. I'll check out Quarkus code so I can keep it on my radar as a consumer in the future. I'm happy to look into how Quarkus uses the `SchemaManagementTool` currently, if it would save you some time. Please let me know if I should go ahead and do that.

Sanne:
@gbadner don't worry for the SPI, it seems Quarkus is compiling just fine. But we're having a memory leak which has been preventing me to run all integration tests - sorry this will take a bit longer.

gbadner:
@Sanne, that's fine. I made a couple of changes here that will affect the hibernate-reactive PR, so I'll take care of that now.

Sanne:
It's looking great! I'm inclined to merge it already, but could you confirm why it seems you also wanted to change the PostgreSQL Dialect semantics, and also changed `SchemaUpdateTest` ?

gbadner:
I added `@GeneratedValue` to the IDs in `SchemaUpdateTest` just so that I could step through JDBC code that processes the sequences in the debugger so I could better understand why `SequenceInformationExtractorLegacyImpl` wasn't working with Hibernate Reactive. 

It will be helpful to have at least one sequence ID in this test when we start adding SchemaMigrator/SchemaValidator support for more DBs.

I could change the test so that there are both generated and non-generated IDs.

gbadner:
I just pushed a [commit](https://github.com/hibernate/hibernate-orm/pull/4106/commits/b63d6e82e32f461e4efb79d2df4929bcb39c89e7) that removes `PostgreSQL10Dialect#buildIdentifierHelper`, since it's no longer needed.

hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/InformationExtractorJdbcDatabaseMetaDataImpl.java
```diff
@@ -6,1017 +6,174 @@
...
-			schemaFilter = schema.getText();
-		}
-
-		try {
-			ResultSet resultSet = extractionContext.getJdbcDatabaseMetaData().getIndexInfo(
-					catalogFilter,
-					schemaFilter,
-					tableName.getTableName().getText(),
-					false,        // DO NOT limit to just unique
-					true        // DO require up-to-date results
```
gbadner, line 823: I just noticed that `true` for the `approximate` argument to `DatabaseMetaData#getIndexInfo says:
```
approximate - when true, result is allowed to reflect approximate or out of data values; when false, results are requested to be accurate
```
ORM was passing `true` here, meaning that approximate data is OK, but the comment is inconsistent saying that up-to-date results are required.

Which is correct? 

FWIW, when `DatabaseMetaData` is not used to get this information, there does not seem to be a way to request up-to-date information. 

For example, MySQL documentation says that `information_schema_stats_expiry=0` should be set in order to get up-to-date statistics. This is probably not what is wanted on a production database.

gbadner: @dreab8, any idea what the intention is here? Is this just a documentation bug?

gbadner: This should not delay merging this issue. I just wanted to raise it since I saw the inconsistency.

gbadner:
@Sanne, I've added a [commit](https://github.com/hibernate/hibernate-orm/pull/4106/commits/7f316ae3f930a1595be2ff43ef8958e649078691) that moves a couple of methods from `PostgreSQL10Dialect` into `PostgreSQL81Dialect`. I verified those methods would work in that dialect by looking at the JDBC source code for PostgreSQL 8.1: https://github.com/pgjdbc/pgjdbc/tree/REL8_1_STABLE.

I also restored `SchemaUpdateTest` by removing `@GeneratedValue` annotations.

I will ask about updating `io.vertx.sqlclient.Tuple#getLong` to do the conversion if the value is a `String` (rather than throwing `ClassCastException`. That way, `PostgreSQL10Dialect#getSequenceInformationExtractor` and `SequenceInformationExtractorPostgresSQLDatabaseImpl` could be removed.

Is there anything else that needs to be done before merging this?

gbadner:
I was able to fix this in `ResultSetAdapter`, so I've removed `SequenceInformationExtractorPostgresSQLDatabaseImpl` and `PostgreSQL10Dialect#getSequenceInformationExtractor` in a new [commit](https://github.com/hibernate/hibernate-orm/pull/4106/commits/41801956cfbe6b90fd069f8330d1a29cad0a4c45).

gavinking:
> moves a couple of methods from PostgreSQL10Dialect into PostgreSQL81Dialect. 

We should reproduce those changes in H6 I guess.

Actually, eventually we're going to have to replicate all this work across.

hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/InformationExtractorJdbcDatabaseMetaDataImpl.java
dreab8: not sure but I would say we want to use up-to-date info but better to ask @sebersole for a confirmation

gbadner:
@Sanne, can this be merged?

gbadner:
I've added a [commit](https://github.com/hibernate/hibernate-orm/pull/4106/commits/108353425b3f551157f91ac290a75d14a6e52535) that is required for MySQL. I checked that the setting returned by `MySQLDialect#getNameQualifierSupport` is correct in MySQL 5.0. `MySQLDialect` is deprecated, and the oldest subclass is `MySQL5Dialect`, so this should be fine.

FYI, `#buildIdentifierHelper` was also added to `MySQLDialect`.

gbadner:
I just added a [commit](https://github.com/hibernate/hibernate-orm/pull/4106/commits/64421a889e3950b984c01ec0d4d57a670d61e1d3) that:
1. Adds CockroachDB192Dialect#getNameQualifierSupport and #buildIdentifierHelper.
2. Changes semantics for AbstractInformationExtractorImpl#processIndexInfoResultSet to be more friendly to subclasses.

hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/InformationExtractorJdbcDatabaseMetaDataImpl.java
gbadner: FYI, I don't see a way for Hibernate to force this for MySQL without setting a system variable.

See https://dev.mysql.com/doc/mysql-infoschema-excerpt/8.0/en/information-schema-statistics-table.html.

Sanne (approved):
Thanks @gbadner !

Sanne:
I'll merge this soon but would prefer to branch first.

Sanne:
Merging this in ORM 5.6 (current `main`)

hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
```diff
@@ -566,6 +567,11 @@ public JDBCException convert(SQLException sqlException, String message, String s
 		};
 	}
 
+	@Override
+	public NameQualifierSupport getNameQualifierSupport() {
+		return NameQualifierSupport.CATALOG;
```
yodous, line 572: Copying comment from [initial commit](https://github.com/hibernate/hibernate-orm/commit/f4f4d49da0de10b96cb52829de2315722fb148eb):

Shouldn't it be NameQualifierSupport.SCHEMA and not CATALOG (since MySQL use the concept of schema and not the catalogs).
Whenever there would be an Entity class annotated with @Table(schema = "tech", name = "point"), the class QualifiedObjectNameFormatterStandardImpl would format the name into point and not the tech.point (since it does not define catalog)

PR #4107, opened 2021-07-24
HHH-11413: Native named query creation fails unintuitively when no resultClass is specified

dariush624:


Sanne:
Nice, thanks!

dariush624:
Thanks @Sanne

sebersole:
Per my comment on the Jira, I don't believe this is the correct solution and causes failures in perfectly valid queries

PR #4109, opened 2021-07-26
HHH-14745 Add FunctionContributor contract

maesenka:
See proposal #4108

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
@@ -1,0 +1,34 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+package org.hibernate.boot.model;
+
+import org.hibernate.query.sqm.function.SqmFunctionRegistry;
+import org.hibernate.service.ServiceRegistry;
+
+/**
+ * Contract for contributing functions
+ *
+ * @author Karel Maesen
+ */
+public interface FunctionContributor {
+
+	/**
+	 *  Contribute functions
+	 *
+	 * @param functionContributions The callback for contributing functions
+	 * @param serviceRegistry The service registry
+	 */
+	void contributeTypes(FunctionContributions functionContributions, ServiceRegistry serviceRegistry);
+}
--- a/hibernate-core/src/main/java/org/hibernate/query/spi/QueryEngine.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/spi/QueryEngine.java
@@ -7,6 +7,9 @@
 package org.hibernate.query.spi;
 
 import org.hibernate.Incubating;
+import org.hibernate.boot.model.FunctionContributions;
+import org.hibernate.boot.model.FunctionContributor;
+import org.hibernate.boot.registry.classloading.spi.ClassLoaderService;
 import org.hibernate.boot.spi.BootstrapContext;
 import org.hibernate.boot.spi.MetadataImplementor;
 import org.hibernate.cfg.AvailableSettings;
@@ -25,6 +28,7 @@ import org.hibernate.query.internal.QueryInterpretationCacheDisabledImpl;
 import org.hibernate.query.internal.QueryInterpretationCacheStandardImpl;
 import org.hibernate.query.named.NamedObjectRepository;
 import org.hibernate.query.sqm.NodeBuilder;
+import org.hibernate.query.sqm.function.SqmFunctionDescriptor;
 import org.hibernate.query.sqm.function.SqmFunctionRegistry;
 import org.hibernate.query.sqm.internal.SqmCreationOptionsStandard;
 import org.hibernate.query.sqm.internal.SqmCriteriaNodeBuilder;
@@ -125,6 +129,11 @@ public class QueryEngine {
 		dialect.initializeFunctionRegistry( this );
 		if ( userDefinedRegistry != null ) {
 			userDefinedRegistry.overlay( sqmFunctionRegistry );
+		}
+
+		for ( FunctionContributor contributor : serviceRegistry.getService( ClassLoaderService.class )
+				.loadJavaServices( FunctionContributor.class ) ) {
+			contributor.contributeTypes( sqmFunctionRegistry::register, serviceRegistry );
 		}
 
 		final boolean showSQLFunctions = ConfigurationHelper.getBoolean(
@@ -281,7 +290,9 @@ public class QueryEngine {
 		return new StandardSqmTranslatorFactory();
 	}
 
-	private static QueryInterpretationCache buildInterpretationCache(Supplier<StatisticsImplementor> statisticsSupplier, Map properties) {
+	private static QueryInterpretationCache buildInterpretationCache(
+			Supplier<StatisticsImplementor> statisticsSupplier,
+			Map properties) {
 		final boolean explicitUseCache = ConfigurationHelper.getBoolean(
 				AvailableSettings.QUERY_PLAN_CACHE_ENABLED,
 				properties,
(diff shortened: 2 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public interface FunctionContributor {
+
+	/**
+	 *  Contribute functions
+	 *
+	 * @param functionContributions The callback for contributing functions
+	 * @param serviceRegistry The service registry
+	 */
+	void contributeTypes(FunctionContributions functionContributions, ServiceRegistry serviceRegistry);
```
beikov, line 33: Personally, I would prefer this to look like the following:

```suggestion
	void contributeFunctions(SqmFunctionRegistry sqmFunctionRegistry, ServiceRegistry serviceRegistry);
```

And drop the `FunctionContributions` interface.

hibernate-core/src/main/java/org/hibernate/query/spi/QueryEngine.java
```diff
@@ -127,6 +131,11 @@ public QueryEngine(
 			userDefinedRegistry.overlay( sqmFunctionRegistry );
 		}
 
+		for ( FunctionContributor contributor : serviceRegistry.getService( ClassLoaderService.class )
```
beikov, line 134: I think we should introduce a notion of ordering for such contributors to avoid non-determinism. IMO the `FunctionContributor` interface should have a default method:

```
default int ordinal() {
    return 1000;
}
```

With a comment that system provided contributors have an ordinal from e.g. 0 - 500, library provided ones should have an ordinal from 500 - 1000 and user provided ones should have an ordinal of 1000+.

The loaded services should then be put into a list, and sorted by a comparator that sorts first by ordinal and second by fully qualified class name.

hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
maesenka: The method name was a copy/paste error. That certainly needs fixing.

The `FunctionContributions` interface was introduced in analogy to the `TypeContributor` contract. It breaks the direct coupling to `SqmFunctionRegistry`.

beikov: I get that, but sometimes it is necessary to base a function on another which you could access through the function registry, but not through the custom intermediate type. Note though that the `TypeContributions` gives access to the underlying thing i.e. the `TypeConfiguration`. IMO in this case we don't need a wrapper.

maesenka: I see. I'll change the interface.

beikov (approved):
Except for formatting (put function/method arguments on new lines if too wide), this looks good to me now.

hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * Determines order in which the contributions will be applied (lowest ordinal first).
+	 *
+	 * The range 0-500 is reserved for Hibernate, range 500-1000 for libraries and 1000-Integer.MAX_VALUE for
+	 * user-defined FunctionContributors.
+	 *
+	 * @return the ordinal for this FunctionContributor
+	 */
+	default int ordinal(){
```
sebersole, line 43: Can we just make this an argument to `#contributeFunctions`?

hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.hibernate.query.sqm.function.SqmFunctionRegistry;
+import org.hibernate.service.ServiceRegistry;
+
+/**
+ * Contract for contributing functions
+ *
+ * @author Karel Maesen
+ */
+public interface FunctionContributor {
```
sebersole, line 25: I know it is implicitly, but can we mark this as `@FunctionalInterface`?  It helps when refactoring later

hibernate-core/src/main/java/org/hibernate/query/spi/QueryEngine.java
```diff
@@ -281,7 +290,18 @@ private static SqmTranslatorFactory resolveSqmTranslatorFactory(
 		return new StandardSqmTranslatorFactory();
 	}
 
-	private static QueryInterpretationCache buildInterpretationCache(Supplier<StatisticsImplementor> statisticsSupplier, Map properties) {
+	private static List<FunctionContributor> sortedFunctionContributors(ServiceRegistry serviceRegistry) {
```
sebersole, line 293: I'm missing where this is used

sebersole (commented):
I think the general idea is great.  Left a few comments

hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
beikov: This is a user/library provided service provider contract through the Java ServiceLoader mechanism, so unfortunately we can't.

hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
beikov: It's not necessarily. It's a service provider

hibernate-core/src/main/java/org/hibernate/query/spi/QueryEngine.java
sebersole: Ok, I see it now.  I still think adding it to the `#contributeFunctions` method is better.

However, I wonder why the "position" at all.   Such things tend to be more difficult than useful.

beikov: To avoid possible difficulties with ordering when we run on the class path :)

hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
sebersole: A service provider is implicitly tied to a service contract (here `FunctionContributor`).  So unless you think people will be registering this implementing class under different service contracts, it in can.

But I agree it is probably best to not define it that way

hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
sebersole: No idea what you mean here.  Why would defining the contract this way be different?

```
interface FunctionContributor {
    void contributeFunctions(int relativePosition, SqmFunctionRegistry functionRegistry, ServiceRegistry serviceRegistry);
}
```

sebersole: But again, per my other comment I personally would just drop the positioning.  You say it is deterministic, but it is not necessarily.  What happens when 2 different ones provide the same position?  Its no longer deterministic

beikov: When there are multiple `FunctionContributor` implementations available, the order in which these contribute functions can be controlled through the `ordinal`. To make the ordering fully deterministic, the sorting uses the fully qualified name of the contributor class, but the ordinal helps with the sane ordering issues.

sebersole: Its not deterministic though.

Its the same fundamental issue as registering anything.  We used to majorly have this problem with JTA syncs e.g.

sebersole:
```
class MyContributor1 implements FunctionContributor {
    ...
    @Override
    public int ordinal() { return 50; }
}

class MyContributor2 implements FunctionContributor {
    ...
    @Override
    public int ordinal() { return 50; }
}
```

Not deterministic ^^

hibernate-core/src/main/java/org/hibernate/query/spi/QueryEngine.java
sebersole: I don't think it solves any difficulties

hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
beikov: Yeah, that's why the sorting also includes the fully qualified class name to have a deterministic behavior. See https://github.com/hibernate/hibernate-orm/pull/4109/files#diff-568ca10f5de61f383d66055b763a8109c8d886348f70149b484aceee2294f3b5R297

But the ordinal is still useful to be able to get the ordering right.

PR #4112, opened 2021-07-26
StatelessSession.fetch()

gavinking:
Draft of a `fetch()` operation for `StatelessSession` (https://hibernate.atlassian.net/browse/HHH-15001)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
@@ -15,6 +15,7 @@ import org.hibernate.LockMode;
 import org.hibernate.SessionException;
 import org.hibernate.StatelessSession;
 import org.hibernate.UnresolvableObjectException;
+import org.hibernate.bytecode.enhance.spi.interceptor.EnhancementAsProxyLazinessInterceptor;
 import org.hibernate.bytecode.spi.BytecodeEnhancementMetadata;
 import org.hibernate.cache.spi.access.EntityDataAccess;
 import org.hibernate.collection.spi.PersistentCollection;
@@ -23,6 +24,8 @@ import org.hibernate.engine.internal.Versioning;
 import org.hibernate.engine.spi.EntityKey;
 import org.hibernate.engine.spi.LoadQueryInfluencers;
 import org.hibernate.engine.spi.PersistenceContext;
+import org.hibernate.engine.spi.PersistentAttributeInterceptable;
+import org.hibernate.engine.spi.PersistentAttributeInterceptor;
 import org.hibernate.engine.transaction.internal.jta.JtaStatusHelper;
 import org.hibernate.engine.transaction.jta.platform.spi.JtaPlatform;
 import org.hibernate.id.IdentifierGeneratorHelper;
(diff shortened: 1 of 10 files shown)


sebersole (approved):
LGTM

sebersole:
Just pushed a fix to a breakage I caused in the build.  Sorry.  Can you rebase this from upstream wip/6.0 and push again?

gavinking:
Does anyone have any suggestions about how this should be tested? I would prefer to add tests to something that already exists? But what, precisely?

sebersole:
> how this should be tested

I'm guessing you mean something more than "in a JUnit test".  But not sure what you are asking

> add tests to something that already exists

What kind of "something"?  You mean an existing test class?

gavinking:
Yes, an existing class.

sebersole:
It's a completely new method/feature.  Why would you not want a new test class?

gavinking:
Ooops, I had forgotten about this one .... but I still want it!

hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
```diff
@@ -358,6 +360,56 @@ private Object createProxy(EntityKey entityKey) {
...
+			}
+		}
+		else if ( association instanceof PersistentCollection) {
+			PersistentCollection<?> persistentCollection = (PersistentCollection<?>) association;
+			if ( !persistentCollection.wasInitialized() ) {
+				CollectionPersister persister =
+						getFactory().getMetamodel().collectionPersister( persistentCollection.getRole() );
+				Object key = persistentCollection.getKey();
+				persistenceContext.addUninitializedCollection( persister, persistentCollection, key );
+				persistentCollection.setCurrentSession(this);
```
beikov, line 400: Shouldn't we also unset the session here in the finally block?

beikov (commented):
Looks good except for the collection initialization case, but the formatting could be improved ;)

gavinking:
Don't merge this yet. Something changed in H6 and it broke this PR.

I will fix it when I have time again.

hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
gavinking: Not as far as I am aware: I don't think SS generally cleans up it's temporary PC in a finally. (Remember: failures are unrecoverable in Hibernate, and the session is considered hosed after a failure.)

beikov: Dunno, it just felt odd that you unset it in the above if-branch but not here. If you say it's fine, I'm ok with that. Just wanted to bring your attention to this.

gavinking: Oh you're saying I forget to unset it, not that you want me to unset it in a `finally`. Yeah, sure, I guess that would be more consistent.

gavinking:
See [HHH-15001](https://hibernate.atlassian.net/browse/HHH-15001).

gavinking:
Alright, at last this is fixed and ready!

PR #4114, opened 2021-07-27
[HHH-11755] Drop "Serializable" from the Event Listener interfaces

kdebski85:


kdebski85:
Do I need to do anything before it can be merged?
"Some checks were not successful" is shown above, but failures are rather not caused by my changes.

kdebski85:
Now status is "All checks have passed".

Sanne:
Thanks @kdebski85 ! Merged it

PR #4115, opened 2021-07-28
make @Type more typesafe

gavinking:
Just a little suggestion to kick off discussion.


sebersole:
Love the direction as we discussed on Zulip.  

Not sure this specifically helps though.  I mean I think it is nice from a user perspective.  But from the perspective of the problems we are trying to solve, this does not help much - we'd still need to account for the de-typed (String) `@Type#type`.  

How about adding a `@Type#typeDef` option which can name the `TypeDef` to use?  This would allow us to limit `@Type#type` options to just things that we can immediately resolve/understand (basic-type registration name, FQN).

gavinking:
@sebersole is proposing to add a new `@CustomType` annotation for what this proposal does, so I'm closing this PR.

PR #4120, opened 2021-07-29
HHH-14762 Testsuite: Assert.notNull must not be used on primitive types

boris-unckel:
Fix https://hibernate.atlassian.net/browse/HHH-14762


boris-unckel:
@sebersole yet another PR today... Could you review and approve this one, please?

PR #4123, opened 2021-07-29
HHH-14764 Testsuite: AssertionErrors must not be caught

boris-unckel:
Fixes https://hibernate.atlassian.net/browse/HHH-14764

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/hqlfetchscroll/HQLScrollFetchTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/hqlfetchscroll/HQLScrollFetchTest.java
@@ -257,6 +257,8 @@ public class HQLScrollFetchTest extends BaseCoreFunctionalTestCase {
 		s.close();
 	}
 
+	final String errMsg = "should have failed because data is ordered incorrectly.";
+
 	@Test
 	public void testScrollOrderChildrenDesc() {
 		Session s = openSession();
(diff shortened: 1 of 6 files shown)


boris-unckel:
@sebersole please have a look at this. You might like the change (fixing tests) but unfortunately test are failing now, due to the AssertionError is not suppressed anymore. If a test result should not be counted, I would have used the junit tag `@ignore`.

hibernate-core/src/test/java/org/hibernate/test/hqlfetchscroll/HQLScrollFetchTest.java
```diff
@@ -276,7 +276,7 @@ public void testScrollOrderChildrenDesc() {
 			fail( "should have failed because data is ordered incorrectly." );
 		}
 		catch ( AssertionError ex ) {
-			// expected
+			throw ex;
```
dreab8, line 279: I think this is not correct, probably we should throw the error only if the message is "should have failed because data is ordered incorrectly." and the same for the `testListOrderChildrenDesc` test.

boris-unckel: @dreab8 Hi Andrea, the re-throw depends on the error message now. Could you review/approve, please?

dreab8: it looks good to me

dreab8:
Applied upstream!

Thanks @boris-unckel

PR #4124, opened 2021-07-29
HHH-14765 Null pointers should not be dereferenced

boris-unckel:
Fixes https://hibernate.atlassian.net/browse/HHH-14765

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/InFlightMetadataCollectorImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/InFlightMetadataCollectorImpl.java
@@ -831,7 +831,7 @@ public class InFlightMetadataCollectorImpl implements InFlightMetadataCollector 
 		else {
 			Table table = namespace.locateTable( logicalName );
 			if ( table != null ) {
-				throw new DuplicateMappingException( DuplicateMappingException.Type.TABLE, logicalName.toString() );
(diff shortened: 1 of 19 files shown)


beikov:
I just quickly looked through some of the changes you did and unfortunately you changed the logic in some cases. Right now, the JVM throws a NPE in the respective code due to the use of null references, and this is "fine" as a null in these scenarios is an error. Your changes skip certain blocks when a variable is null, which might change the logic. I would understand if you do something like `if (variable == null) { throw new IllegalArgumentException("..."); }`, but even that is not really that useful in all cases IMO. Ideally we would express the non-null intent through the declaration, but Java does not have something like that.

I'll close this PR and issue due to these reasons I stated above. If you have a concrete bug due to the handling of nulls with a reproducer, then we can talk about a fix that handles nulls somehow.

Please get in touch with us on [Zulip](https://hibernate.zulipchat.com/#narrow/stream/132096-hibernate-user) first before you put effort into an issue as it would be sad if rejections like this would discourage you from contributing in the future.

boris-unckel:
@beikov thanks for your feedback. Out of 40 [https://sonarcloud.io/project/issues?id=hibernate-orm&resolved=false&rules=java%3AS2259&types=BUG](https://sonarcloud.io/project/issues?id=hibernate-orm&resolved=false&rules=java%3AS2259&types=BUG) I have picked up 19 for individual fixing, rest looked false positive/won't fix to me.  I'll follow your suggestion, and if I have a concrete bug and a reproducer and a fix, to come back for a discussion.

PR #4130, opened 2021-08-05
HHH-14770 Optimise access to EntityPersister in Metamodel via Class type

Sanne:
- https://hibernate.atlassian.net/browse/HHH-14770

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
@@ -167,12 +167,12 @@ public class StatelessSessionImpl extends AbstractSharedSessionContract implemen
 
 	@Override
 	public Object get(Class entityClass, Serializable id) {
-		return get( entityClass.getName(), id );
+		return get( entityClass, id, LockMode.NONE );
 	}
 
 	@Override
 	public Object get(Class entityClass, Serializable id, LockMode lockMode) {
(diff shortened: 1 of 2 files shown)


DavideD (approved):
Looks good to me

hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
```diff
@@ -491,7 +494,7 @@ public EntityPersister getEntityPersister(String entityName, Object object)
 			throws HibernateException {
 		checkOpen();
 		if ( entityName == null ) {
-			return getFactory().getMetamodel().entityPersister( guessEntityName( object ) );
```
gavinking, line 494: Hrm. This seems to not change behavior, but only because `guessEntityName()` has a different implementation on `StatelessSession`. I guess it's fine, but it just seems a little anomalous.

Sanne: I agree. In this PR I'm exclusively looking at the performance and am aiming for not making semantic changes, but there's something to look into better as a follow up.

gavinking: Yeah, so my concern was really that deleting that call to `guessEntityName()` might make it harder to notice the difference in future, but really it's no big deal, just a comment.

PR #4136, opened 2021-08-09
Fix alias case sensitive and JPAQL strict compliance

dreab8:


beikov (approved):

dreab8:
Thanks @beikov for the review

PR #4137, opened 2021-08-09
Re-enabled additional tests

dreab8:
Fix load toOne association not referencing a PK with `FetchMode.SELECT`
Fix throw `EntityNotFoundException` when an association not mapped with  `NotFound(IGNORE)`

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
@@ -91,6 +91,7 @@ public class ToOneAttributeMapping
 
 	private final String sqlAliasStem;
 	private final boolean isNullable;
+	private final boolean isIgnoreNotFound;
 	private final boolean unwrapProxy;
 	private final EntityMappingType entityMappingType;
 
@@ -164,6 +165,7 @@ public class ToOneAttributeMapping
 
 		if ( bootValue instanceof ManyToOne ) {
 			final ManyToOne manyToOne = (ManyToOne) bootValue;
+			this.isIgnoreNotFound = ( (ManyToOne) bootValue ).isIgnoreNotFound();
 			if ( manyToOne.isLogicalOneToOne() ) {
 				cardinality = Cardinality.LOGICAL_ONE_TO_ONE;
 			}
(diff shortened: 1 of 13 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -89,6 +89,7 @@
 
 	private final String sqlAliasStem;
 	private final boolean isNullable;
+	private boolean isIgnoreNotFound;
```
beikov, line 92: Make `final`. Also, IMO we should ensure that isNullable is set to `true` if `isIgnoreNotFound` is true.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
```diff
@@ -192,7 +189,7 @@ public void initializeInstance(RowProcessingState rowProcessingState) {
 				entityName,
 				entityIdentifier,
 				true,
-				nullable
+				referencedModelPart.isIgnoreNotFound()
```
beikov, line 192: Is that really enough i.e. is `ignoreNotFound` set to false automatically in case of a non-nullable many-to-one without an explicit `@NotFound` annotation?

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -753,7 +758,6 @@ public EntityFetch generateFetch(
 			return new EntityFetchSelectImpl(
 					fetchParent,
 					this,
-					isNullable,
```
beikov, line 756: We also need to record whether a foreign key constraint is available on the respective ForeignKey instance and based on that and the `ignoreNotFound` flag, adapt `EntityFetchJoinedImpl` to also create a domain result for the foreign key part to be able to check inconsistencies.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
dreab8: I think so but I'll try it

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
dreab8: yeap

dreab8:

PR #4143, opened 2021-08-12
Revert HHH-14770 Optimise access to EntityPersister in Metamodel via Class type

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14770

It causes memory leaks (`SessionFactory` instances are never garbage collected, for some reason. See the comments on the JIRA issue.


yrodiere:
Tests pass, I will merge this.

@Sanne Be aware I reverted your commit because it caused memory leaks.

Feel free to apply your patch again if you know how to fix the memory leak; I don't, and I'm leaving shortly so unfortunately I don't have time to find a solution.

PR #4145, opened 2021-08-16
Re-enabled additional tests

dreab8:


beikov:
Same problem again with the old Oracle version that some identifiers are too long in `org.hibernate.userguide.mapping.embeddable.EmbeddableOverrideTest` :)

dreab8:
Hi @beikov I'm going to fix it, thanks.
What about https://github.com/dreab8/hibernate-orm/commit/18213bdb64c5c46f947024879ff71baf103b7e51 ? not sure about it

jrenaat:
Resolved the conflict with the TransactionTimeoutTest due to myPR; kept some of your things in it.

hibernate-core/src/main/java/org/hibernate/query/results/DomainResultCreationStateImpl.java
```diff
@@ -266,7 +268,8 @@ public Expression resolveSqlExpression(
 		else if ( created instanceof ColumnReference ) {
 			final ColumnReference columnReference = (ColumnReference) created;
 			final String columnExpression = columnReference.getColumnExpression();
-			final int jdbcPosition = jdbcResultsMetadata.resolveColumnPosition( columnExpression );
+
+			final int jdbcPosition = jdbcResultsMetadata.resolveColumnPosition( currentResultBuilder.getColumnAlias( columnExpression ) );
```
beikov, line 272: This and all the related changes shouldn't be necessary. What does that fix?

dreab8: reverted thanks

PR #4156, opened 2021-08-20
HHH-14789 Improve NegatedPredicateWrapper::not

quaff:
return original positive predicate instead of constructing a new NegatedPredicateWrapper

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/NegatedPredicateWrapper.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/internal/predicate/NegatedPredicateWrapper.java
@@ -78,7 +78,7 @@ public class NegatedPredicateWrapper extends ExpressionImpl<Boolean> implements 
 
 	@Override
 	public Predicate not() {
-		return new NegatedPredicateWrapper( this );
+		return predicate;
 	}
 
 	@Override


sebersole:
Unfortunately this return-a-new-node approach is what JPA "expects".  I say that in quotes because the spec does not say this but the TCK explicitly checks that.  I already challenged that with Oracle in previous versions and lost :shrug:  Maybe that can be revisited now with the move to Jakarta.

Now if you want to add an option (under `JpaCompliance`) to do the more sane thing (not generate new nodes) and apply that to all calls I would totally accept that.  But as it stands, I have to reject this

quaff:
@sebersole Actually I wanted is the underlying positive predicate, currently it's not exposed, I'm proposing add a getter for that, I will create a PR if you accept.

My use case is to check Predicate is constant true, then optimize the query to not include unnecessary `true=1` introduced by `cb.isTrue(cb.literal(true))` or other similar

```java
	private static Predicate unwrap(NegatedPredicateWrapper wrapper) {
		return ReflectionUtils.getFieldValue(wrapper, "predicate"); // TODO avoid reflection
	}

	public static boolean isConstantTrue(Predicate predicate) {
		if (predicate instanceof NegatedPredicateWrapper) {
			return isConstantFalse(unwrap((NegatedPredicateWrapper) predicate));
		} else if (predicate instanceof NullnessPredicate) {
			// literal cannot be null
			return false;
		} else if (predicate instanceof BooleanAssertionPredicate) {
			BooleanAssertionPredicate bap = (BooleanAssertionPredicate) predicate;
			Expression<Boolean> exp = bap.getExpression();
			if (exp instanceof LiteralExpression) {
				return bap.getAssertedValue().equals(((LiteralExpression<Boolean>) exp).getLiteral());
			}
		}
		return false;
	}

	public static boolean isConstantFalse(Predicate predicate) {
		if (predicate instanceof NegatedPredicateWrapper) {
			return isConstantTrue(unwrap((NegatedPredicateWrapper) predicate));
		} else if (predicate instanceof NullnessPredicate) {
			return ((NullnessPredicate) predicate).getOperand() instanceof LiteralExpression;
		} else if (predicate instanceof BooleanAssertionPredicate) {
			BooleanAssertionPredicate bap = (BooleanAssertionPredicate) predicate;
			Expression<Boolean> exp = bap.getExpression();
			if (exp instanceof LiteralExpression) {
				return !bap.getAssertedValue().equals(((LiteralExpression<Boolean>) exp).getLiteral());
			}
		}
		return false;
	}

	public static <T> Predicate andExample(Root<T> root, CriteriaBuilder cb, Predicate predicate, Example<T> example) {
		Predicate examplePredicate = QueryByExamplePredicateBuilder.getPredicate(root, cb, example);
		if (isConstantTrue(examplePredicate)) // eliminate unnecessary true=1
			return predicate;
		return cb.and(predicate, examplePredicate);
	}

	public static <T> Predicate orExample(Root<T> root, CriteriaBuilder cb, Predicate predicate, Example<T> example) {
		Predicate examplePredicate = QueryByExamplePredicateBuilder.getPredicate(root, cb, example);
		if (isConstantTrue(examplePredicate))
			return examplePredicate;
		return cb.or(predicate, examplePredicate);
	}

```

PR #4171, opened 2021-08-25
HHH-14792 Remove support for bytecode enhancement via Javassist

Sanne:
- https://hibernate.atlassian.net/browse/HHH-14792


Sanne:
@sebersole would you agree?

I'm about to release 5.6, I think it's a good time to apply some cleanup as it has been deprecated for a while.  Also it would reduce friction for people when they move to 6.

PR #4177, opened 2021-08-31
HHH-14794 : More changes to support SchemaMigrator/SchemaValidator using Hibernate Reactive

gbadner:
https://hibernate.atlassian.net/browse/HHH-14794


Sanne:
Hi @gbadner ! Looks fine - is it intended to stay as a draft? 

We can release more frequently if that helps with the work on the HR side, let me know.

gavinking:
@Sanne I personally would love to see that.

hibernate-core/src/main/java/org/hibernate/dialect/SQLServerDialect.java
```diff
@@ -248,4 +251,10 @@ public String getCreateTemporaryTableColumnAnnotation(int sqlTypeCode) {
 				return "";
 		}
 	}
+
+	@Override
+	public NameQualifierSupport getNameQualifierSupport() {
```
Sanne, line 256: Interesting that this was missing. Perhaps it deserves its own JIRA so to highlight the change better in changelogs?

Sanne:
@gbadner I haven't heard back, but this looks totally fine so I'll start merging this. If you have more changes please send more PRs :)

PR #4180, opened 2021-08-31
Change start_with to start_value for postgres sequences

dempe:
`start_value` is the correct column name.  See here:
https://www.postgresql.org/docs/9.1/infoschema-sequences.html

Note:  sorry I could not create a Hibernate Jira account. I tried nearly
a dozen times.  Kept getting 500s.


gavinking:
`SequenceInformationExtractorHSQLDBDatabaseImpl` has nothing to do with Postgres. It's right there in the name: `HSQLDB`.

PR #4181, opened 2021-08-31
Fixed and reorganized tests in hibernate-jcache

jrenaat:

### DIFF
--- a/hibernate-jcache/src/test/java/org/hibernate/orm/test/jcache/BaseFunctionalTest.java
+++ b/hibernate-jcache/src/test/java/org/hibernate/orm/test/jcache/BaseFunctionalTest.java
@@ -4,28 +4,29 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later
  * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
  */
-package org.hibernate.jcache.test;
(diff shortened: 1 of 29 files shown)


beikov (commented):
`RefreshUpdatedDataTest` is failing on derby, other than that, this looks ok

beikov (approved):

jrenaat:
Still looking into issues with this PR; I removed the extension from BaseUnitTestCase from one (or several) Tests, and that provokes at least one of those to hang, which was previously caught by the Timeout Junit Rule.

dreab8:
Hi @jrenaat, try to add `@BaseUnitTest` to the tests where you have removed `extends BaseUnitTestCase`

jrenaat:
I have ignored the RefreshUpdatedDataTest for the sybase and hsqldb environments, because it simply hangs 8db locking issues?), but now running into another ~170 tests failing in sybase (not apparently related to the changes in hibernate-jcache).
Locally, only two fail (consistently), and this can only be reproduced by running all hibernate-core tests (i.e. not by running the failing test individually); the exception message is "Converting `org.hibernate.exception.JDBCConnectionException` to JPA `PersistenceException` : Error calling Driver#connect" --> connections not being closed somewhere?

beikov:
This should be fine. I guess this is just a temporary issue.

PR #4187, opened 2021-09-03
HHH-14796: Cannot replace an existing JPQL NamedQuery with a native NamedQuery

rmartinc:
Issue: https://hibernate.atlassian.net/browse/HHH-14796

Now the register method also removes the name from the other query map if it exists there.


Sanne:
Very nice, thanks! and I agree with the solution.  Odd that nobody noticed, I guess it's not common to reuse the same name for a different query kind.

I'll apply some minor style fixes and merge it.

Sanne:
I've rebased and merged it. Thanks!

PR #4188, opened 2021-09-03
Test case for HHH-14798 No PostUpdateEvent fired when setting a property with @OneToOne(mappedBy = ...) to null

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14798

No fix at the moment...

See also https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/topic/HHH-14798/near/251876108


yrodiere:
It turns out this behavior was expected; what wasn’t normal was the fact that we got an event when setting the property to a non-null value.

See this comment for details:https://hibernate.atlassian.net/browse/HSEARCH-4708?focusedCommentId=110020

PR #4191, opened 2021-09-06
HHH-14811 org.hibernate.AssertionFailure thrown instead of LazyInitializationException when trying to access a lazy property on a deleted entity

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14811


Sanne:
Thanks!
I do wonder if we shouldn't have a specific exception type for unexpected/illegal operations on deleted entities, but that's probably best left as a separate discussion.

PR #4192, opened 2021-09-06
insert with @DynamicInsert should use generateIdentityInsertString()

gavinking:
See the comment in `generateInsertString()`

This is a very minor change but is conceptually correct.


gavinking:
Hrm, failing on Sybase, but I have no way to figure out if that's related to my change...

dreab8:
Hi @gavinking , I do not thing the failure is caused by you changes.

dreab8:
Thanks @gavinking !

gavinking:
Thanks for merging!

PR #4195, opened 2021-09-06
HHH-14810 @NotBlank and @NotEmpty were used, "NOT NULL" option wasn't add in generated DDL.

include42:
See Jira issue description at [this link](https://hibernate.atlassian.net/browse/HHH-14810)

If there are any mistakes on code, please comment.
thanks :)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/beanvalidation/TypeSafeActivator.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/beanvalidation/TypeSafeActivator.java
@@ -21,6 +21,8 @@ import javax.validation.constraints.Digits;
 import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.Size;
 import javax.validation.metadata.BeanDescriptor;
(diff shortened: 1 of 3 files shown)


Sanne:
It's an interesting point but I suspect there are good reasons for this; sadly I don't remember the details other than having seen such conversations in the past  - we need to ask @gsmet or @gunnarmorling .

include42:
Thank you for your comments 👍 
I was curious about what other contributors thought of this issue. I was also curious if this was intended. 
this is the answer I needed. :)

I'll wait for their answers. thanks a lot.

gsmet:
I think it makes sense for these two.

They are the exception to the rule that `null` is considered a valid value for most constraints. The only thing I'm unsure of is that `@NotEmpty` can be applied to lists, maps and collections too. I have no idea if it requires special handling on the Hibernate ORM side, but I will let you judge of that :).

Sanne:
Note to self: If we decide to merge it (pending further testing on collections like @gsmet suggested) we're not going to backport it as the schema change would be annoying for very little gain.

And perhaps it should be a 6.0 exclusive.

include42:
Hello, @gsmet! Thank you for the comments.
In the code for `@NotEmpty`, I wrote only for the case of String. So I couldn't think about apply `@NotEmpty` to collections.(thanks for find it!) 

However, I have no idea too if it requires special handling about that. 
I agree with your opinion because I couldn't find such a case. :)

PR #4206, opened 2021-09-08
HHH-14814 Inconsistent rendering of table alias in delete queries

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14814

I don't know how to fix this bug in 5.x, but it turns out it's already been fixed in 6.0, somehow.

Let's merge the test and close the ticket as fixed in 6.0?


yrodiere:
Failures seems unrelated. Merging.

PR #4208, opened 2021-09-09
Fix incorrect additional joins being generated when non-fetch joining and selecting a plural association

sebersole:
This first commit is just tests illustrating the problem until we can all discuss the better design approach to fix this - either we stop implicitly "converting" the collection selection as its `{element}` path or later better handle the "mismatched" NavigablePaths in terms of resolving TableGroups


sebersole:
Wrong target branch.  See https://github.com/hibernate/hibernate-orm/pull/4209 instead

PR #4209, opened 2021-09-09
Fix incorrect additional joins being generated when non-fetch joining and selecting a plural association

sebersole:
This first commit is just tests illustrating the problem until we can all discuss the better design approach to fix this - either we stop implicitly "converting" the collection selection as its {element} path or later better handle the "mismatched" NavigablePaths in terms of resolving TableGroups

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmTreePrinter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmTreePrinter.java
@@ -101,6 +101,7 @@ import org.hibernate.query.sqm.tree.select.SqmSubQuery;
 import org.hibernate.query.sqm.tree.update.SqmAssignment;
 import org.hibernate.query.sqm.tree.update.SqmSetClause;
 import org.hibernate.query.sqm.tree.update.SqmUpdateStatement;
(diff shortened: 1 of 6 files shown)


beikov:
I think this is something we "fixed" in 5.x by specifying the force-discriminator option, although I never really liked that, so +1 from me to fix this and maybe we can get rid of the force-discriminator stuff?

dreab8:
I played a little bit with this issue https://github.com/dreab8/hibernate-orm/commit/a746442e345bbc127a7bbd63fa81f3f68d0a1b36

sebersole:
> I think this is something we "fixed" in 5.x by specifying the force-discriminator option, although I never really liked that, so +1 from me to fix this and maybe we can get rid of the force-discriminator stuff?

The inheritance is completely irrelevant here.  The problem is the joins and it happens even without inheritance.

sebersole:
> I played a little bit with this issue [dreab8@a746442](https://github.com/dreab8/hibernate-orm/commit/a746442e345bbc127a7bbd63fa81f3f68d0a1b36)

This is similar to the possible second approach I mentioned initially.

I think we should at least try to change the node used in the selection for this case.  At the moment we implicitly use the element's source-path.  IMO that is the "cleanest" solution

PR #4210, opened 2021-09-09
HHH-14820 Reproducer - Customized EnumType not working with hbm mapping in Hibernate 6

gtoison:
A simple unit test illustrating issue HHH-14820 : since Hibernate ORM 6 the custom NamedEnumUserType is only instantiated once (while there aretwo enums configured with this type).

When trying to execute a query that NamedEnumUserType has the last enum class and `Enum.valueOf(...)` fails.

The test is currently marked with `@Ignore` so it needs to be removed to get a failing result.

See [the JIRA HHH-14820 issue](https://hibernate.atlassian.net/browse/HHH-14820)


beikov:
Thanks a lot for the test case. I'll look into it ASAP.

sebersole:
I'm curious what actually fails for you here.  The test passes for me rebased on top of current 6.0 branch.  And honestly looking at the test.

gtoison:
Did you remove the `@Ignore` ?
I was not sure what was the right way to add a failing unit test

sebersole:
> Did you remove the `@Ignore` ? I was not sure what was the right way to add a failing unit test

Probably you meant to execute the query...

sebersole:
Ok, that was it.  You probably removed the execution at some point

sebersole:
BTW, have a fix already.  Running the full tests and I'll push for you to try

gtoison:
Sorry that was not super clear, I tried again with the latest version of the wip/6.0 branch and test failed with this error:

```
> Task :hibernate-core:test

NamedEnumUserTypeTest > testNamedEnumUserType FAILED
    jakarta.persistence.PersistenceException at ExceptionConverterImpl.java:167
        Caused by: org.hibernate.HibernateException at JdbcValuesResultSetImpl.java:312
            Caused by: java.lang.IllegalArgumentException at Enum.java:238
```

It's great to hear that fix is coming up, thank you very much for looking into it!

gtoison:
Yes sorry I was trying to say that the @Ignore annotation needs to be removed
In some projects the etiquette for submitting a failing unit test is to commit it disabled so you don't break the build.

I've removed it in a new commit for the sake of clarity. Sorry for the noise!

sebersole:
But this is your test method:

```
		doInHibernate( this::sessionFactory, s -> {
			s.createQuery( "from Person p", Person.class );
		} );
```

The query is not executed.    In other words I only got this to fail when I changed that to

```
		doInHibernate( this::sessionFactory, s -> {
			s.createQuery( "from Person p", Person.class ).list();
		} );
```

sebersole:
Fixed - https://github.com/hibernate/hibernate-orm/commit/c799f858013b8043ce2e17b3ba8b5d83b6a4d1c7

PR #4215, opened 2021-09-10
Clean up various legacy "read path" contracts

sebersole:
* Type#nullSafeGet
* Type#hydrate
* Type#resolve
* Type#getSemiResolvedType
* Type#semiResolve
* related


sebersole:
Just a first commit for Type.  More coming to clean up tuple and walking packages

sebersole:
Out of date

PR #4217, opened 2021-09-12
HHH-5249 Add more details on MappingException

lillo42:
Add componentClassName details on  MappingException

See [the JIRA HHH-5249 issue](https://hibernate.atlassian.net/browse/HHH-5249)


Sanne:
Good idea, thanks!

PR #4218, opened 2021-09-13
HHH-13999 Support SQL Server 2016

knutwannheden:
SQL Server 2016 (13.x) and later support the `if exists` clause for most `drop` DDL statements. The new `SQLServer2016Dialect` dialect accounts for this and offers the advantage that no error messages get logged when using `hibernate.hbm2ddl.auto=create-drop`.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Database.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Database.java
@@ -480,7 +480,7 @@ public enum Database {
 	SQLSERVER {
 		@Override
 		public Class<? extends Dialect> latestDialect() {
-			return SQLServer2012Dialect.class;
+			return SQLServer2016Dialect.class;
 		}
 
 		@Override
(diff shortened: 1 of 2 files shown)


Sanne:
thanks!

hibernate-core/src/main/java/org/hibernate/dialect/Database.java
```diff
@@ -501,16 +501,20 @@ public Dialect resolveDialect(DialectResolutionInfo info) {
...
 					case 11:
-					case 12:
-					case 13: {
+					case 12: {
 						return new SQLServer2012Dialect();
 					}
+					case 13:
+					case 14:
+					case 15: {
+						return new SQLServer2016Dialect();
```
Sanne, line 510: is `case 16:` intentionally missing?

I guess it works via the fallback on `latestDialectInstance` but it doesn't seem very intuitive

knutwannheden: AFAIK version 2019 corresponds to 15.x and there is no newer version released yet. But if you think it makes sense, I could of course also add `16` here.

Sanne: ah right, ok I forgot that the major version doesn't match the "product" version name. All good then, thanks!

knutwannheden:
> Shouldn't we patch also method `latestDialect` to return `SQLServer2016Dialect` ?
> 
> If you want I can do it.

Good catch! I completely missed that. Updated now.

Sanne:
Looks great! Merging

PR #4220, opened 2021-09-14
@GeneratedColumn annotation for SQL 'generated always as'

gavinking:
Proposal for a `@GeneratedColumn` annotation.

We need to talk about the name.


beikov:
This looks interesting, though I think we need to discuss how we can make use of "HQL expressions" for this and other usages (column default, check constraint, etc.).

gavinking:
In the latest iteration, this PR adds:

- the `@ColumnGeneratedAlways` annotation, and
- the `@ColumnDefault(fetch)` member.

This lets you write `@ColumnGeneratedAlways(value="...", fetch=true)` and `@ColumnDefault(value="...", fetch=true)` instead of `@Generated(ALWAYS) @ColumnGeneratedAlways("...")` and `@Generated(INSERT) @ColumnDefault("...")`.

It seems to me that `fetch=true` is quite a lot more comfortable.

Of course, this is implemented using `@ValueGenerationType`.

gavinking:
I have added a `name` member to `@ColumnDefault` and `@GeneratedColumn`, in order to be able to handle multi-column mappings. For you @sebersole.

gavinking:
This turned out to be a much deeper rabbit hole than I anticipated.

Closing in favor of https://github.com/hibernate/hibernate-orm/pull/4540

PR #4221, opened 2021-09-15
HHH-14826 Fix regression caused by one-to-one L2 cache support

cristatus:


Sanne:
Thanks for the fix! But could you add an integration test please?

cristatus:
> Thanks for the fix! But could you add an integration test please?

Already attached here https://hibernate.atlassian.net/browse/HHH-14826

cristatus:
> > Thanks for the fix! But could you add an integration test please?
> 
> Already attached here https://hibernate.atlassian.net/browse/HHH-14826



> Thanks for the fix! But could you add an integration test please?

Do you mean to add test in hibernate code itself?

cristatus:
BTW, this assertion is suspicious https://github.com/hibernate/hibernate-orm/blob/f6a03cf7715439ba09bac42afc5380414c545517/hibernate-core/src/test/java/org/hibernate/test/onetoone/cache/OneToOneCacheTest.java#L63

beikov:
> Do you mean to add test in hibernate code itself?

Yes we mean in this PR :)

cristatus:
I was unable to reproduce with hibernate `Session` but can reproduce with `EntityManager`. However, it is breaking the unit tests in `org.hibernate.test.onetoone.cache.OneToOneCacheTest` (which were added during the change that is causing the regression).

But you can see the attached hibernate-test.zip to see the issue.

cristatus:
> I was unable to reproduce with hibernate `Session` but can reproduce with `EntityManager`. However, it is breaking the unit tests in `org.hibernate.test.onetoone.cache.OneToOneCacheTest` (which were added during the change that is causing the regression).
> 
> But you can see the attached hibernate-test.zip to see the issue.

Finally, I am able to re-produce with the hibernate `Session` too. I am wondering why it didn't initially. However, it's failing the other tests. To me, the failures are obvious as it wrongly assumes that one-to-one field value is resolved from cache while it's not. I will push the test soon.

Sanne:
many thanks for the help @cristatus ! As soon as you push it we'll have a better look

dreab8:
I opened a new PR that fixes the failing tests, I have based it on the main branch and extracted the tests relative to this issue out of the `org.hibernate.test.onetoone.cache.OneToOneCacheTest`

Sanne:
This was incorporated in #4228 - closing. Thanks!

deAtog:
@Sanne I apologize for not responding sooner. Even though these changes have already been integrated, I would argue that the issue was not with hibernate, but with the test case provided. The test case calls ProductConfig.setProduct to set the Product of the ProductConfig object and then calls persist() to commit the changes. The test case does not call Product.setConfig() prior to calling persist() so the Product is never associated with the ProductConfig. As a result, the in-memory view of the data is inconsistent prior to calling persist() and results in the cache being inconsistent as well. In my opinion, the inconsistency here is caused by failing to update both sides of the one-to-one relationship prior to calling persist(). The inconsistency can be resolved by modifying ProductConfig.setProduct to call Product.setConfig and Product.setConfig to call ProductConfig.setProduct whenever the values change. There are examples of this in other one-to-one tests.

gavinking:
Ugh.

@Sanne is going to be unavailable for a while.

But @deAtog if your analysis is correctm then yes this should definitely be rolled back ASAP.

I'm not sure if anyone noticed, but this patch changed the implementation of `OneToOneType.assemble()`, which is ancient code that has not changed since the days of Hibernate 2, and should definitely never be changed without really thorough review and testing.

Who can take responsibility for making sure this gets rolled back (if, as I believe, that's indeed that's the right thing to do).

cristatus:
@gavinking @deAtog the code was working fine with earlier versions, so it was a regression. If it's impacting the performance, better to revert. We can adjust our code but than it should be considered as breaking change.

gavinking:
Sure, sometimes incorrect code can work by accident. That's a thing that happens.

Sanne:
FWIW, this was reverted in `main` and in ORM `5.6.5.Final`:
 - https://hibernate.atlassian.net/browse/HHH-15044

cristatus:
Hi @Sanne,

Thanks for the update. My code was like this:

Owning side:

```java
public void setConfig(ProductConfig config) {
  this.config = config;
  if (config != null && config.getProduct() != this) {
    config.setProduct(this);
  }
}
```

Non-owning side:

```java
public void setProduct(Product product) {
  this.product = product;
}
```

So, changing the non-owning side as followed solved the issue:

```java
public void setProduct(Product product) {
  this.product = product;
  if (product != null && product.getConfig() != this) {
    product.setConfig(this);
  }
}
```

Thanks for the investigation. But I think this change should be documented in changelog as it was working fine till 5.5.32.

The code in our app was like this since long till 5.5.33 (almost 5-6 years, we started with some v4.x release). So we thought, it might be a regression.

deAtog:
@cristatus I would recommend the following code:

In your ProductConfig class do:

```java
public void setProduct(Product product) {
    if (this.product != product) {
        if (this.product != null) {
            this.product.setConfig(null);
        }

        this.product = product;

        if (product != null) {
            product.setConfig(this);
        }
    }
}
```
In your Product class do:
```java
public void setConfig(ProductConfig config) {
    if (this.config != config) {
        if (this.config != null) {
            this.config.setProduct(null);
        }

        this.config = config;

        if (config != null) {
             config.setProduct(this);
        }
    }
}
```

This should take care of updating all affected objects in case you move a ProductConfig to another Product or vice versa.

PR #4223, opened 2021-09-15
Re-enable tests

jrenaat:


jrenaat:
WithClauseTest fails for most db's; for the ones I've checked, it appears to be because a table alias is referenced in the select, before it is actually declared, causing an error along the lines of "Unknown column"
snippet:
`from Human h1_0 inner join Animal h1_1 on h1_0.mammal=h1_1.id
                inner join Mammal h1_2 on h1_0.mammal=h1_2.animal
                inner join Animal o1_0 on o1_0.mother_id=o1_0.id and m1_0.father_id=?
                left outer join Animal m1_0 on o1_0.mother_id=m1_0.id`
where the offending alias is m1_0.father_id

beikov (commented):
First of all, like I wrote on Zulip, you shouldn't remove the test that asserts a with clause for fetch joins is illegal, as it ought to stay illegal. Apart from that, it looks like what is failing right now is the following query:

```
list = session.createQuery( "from Human h inner join h.offspring o with o.mother.father = :cousin" )
							.setParameter( "cousin", session.load( Human.class, Long.valueOf( "123" ) ) )
							.list();
```

This is indeed a tough one, because there is an implicit join in the with clause, which would have to be rendered before the current join, or if it is dependent on the current join alias, then as a subquery. I'll look into this and take over this PR.

jrenaat:
> First of all, like I wrote on Zulip, you shouldn't remove the test that asserts a with clause for fetch joins is illegal, as it ought to stay illegal. 

Which is why I asked on zulip ;)

Apart from that, it looks like what is failing right now is the following query:
> 
> ```
> list = session.createQuery( "from Human h inner join h.offspring o with o.mother.father = :cousin" )
> 							.setParameter( "cousin", session.load( Human.class, Long.valueOf( "123" ) ) )
> 							.list();
> ```
> 
> This is indeed a tough one, because there is an implicit join in the with clause, which would have to be rendered before the current join, or if it is dependent on the current join alias, then as a subquery. I'll look into this and take over this PR.

Ok, I had come to this conclusion (sort of) as well, but fixing this it's out of my reach for the time being.

PR #4227, opened 2021-09-16
HHH-13661 | Added case to PostgreSQL Dialect to map client operation cancellation to QueryTimeoutException

bertiepinnock:
Addressing issue [HHH-13661](https://hibernate.atlassian.net/browse/HHH-13661) whereby triggering a client-side query timeout for PostgreSQL (e.g. by setting the `javax.persistence.query.timeout` property) resulted in a generic JDBC exception rather than being mapped to the specific `QueryTimeoutException` provided by Hibernate.

I opted to implement this as part of the PostgreSQL dialect-specific SQL State conversion delegate. As mentioned by the original reporter (who deserves a lot of credit for doing most of the legwork on the reported issue!), there were a few other options, including adding to [some existing DB-specific checks within the generic SQL State conversion delegate](https://github.com/hibernate/hibernate-orm/blob/39e576cea49c62763e651237f3666f97205406f6/hibernate-core/src/main/java/org/hibernate/exception/internal/SQLStateConversionDelegate.java#L132). The commits adding these checks in the generic delegate are quite old, so I was unsure if they had been implemented there for a specific reason, or if they are just there for legacy reasons.

Lastly it looks as though my code formatter has attacked quite a few other lines in the `PostgreSQL81Dialect` class other than the ones I changed - let me know if I should revert those changes!

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java
@@ -18,6 +18,7 @@ import org.hibernate.JDBCException;
 import org.hibernate.LockMode;
 import org.hibernate.LockOptions;
 import org.hibernate.PessimisticLockException;
+import org.hibernate.QueryTimeoutException;
 import org.hibernate.cfg.Environment;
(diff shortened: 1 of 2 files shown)


beikov (approved):
Looks good to me, thanks!

PR #4231, opened 2021-09-20
HHH-14828 Bytecode enhancement generates invalid bytecode for final fields

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14828

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/PersistentAttributeTransformer.java
+++ b/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/PersistentAttributeTransformer.java
@@ -15,10 +15,8 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-
 import javax.persistence.Embedded;
 
-import net.bytebuddy.utility.OpenedClassReader;
 import org.hibernate.bytecode.enhance.internal.bytebuddy.EnhancerImpl.AnnotatedFieldDescription;
 import org.hibernate.bytecode.enhance.spi.EnhancerConstants;
 import org.hibernate.engine.spi.CompositeOwner;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/PersistentAttributeTransformer.java
```diff
@@ -187,15 +192,19 @@ private boolean isEnhanced(String owner, String name, String desc) {
...
-							Visibility.PUBLIC
-					)
-					.withParameters( enhancedField.getType().asErasure() )
-					.intercept( fieldWriter( enhancedField ) );
+					.intercept( fieldReader( enhancedField ) );
+			// Final fields will only be written to from the constructor,
+			// so there's no point trying to replace final field writes with a method call.
+			if ( !enhancedField.getFieldDescription().isFinal() ) {
+				builder = builder
+						.defineMethod(
```
Sanne, line 229: Do you rememer if Panache is taking the existance of such methods for granted?

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/PersistentAttributeTransformer.java
```diff
@@ -42,12 +43,29 @@
...
 import net.bytebuddy.matcher.ElementMatcher.Junction;
 import net.bytebuddy.pool.TypePool;
+import net.bytebuddy.utility.OpenedClassReader;
 
 final class PersistentAttributeTransformer implements AsmVisitorWrapper.ForDeclaredMethods.MethodVisitorWrapper {
 
 	private static final CoreMessageLogger log = CoreLogging.messageLogger( PersistentAttributeTransformer.class );
 
 	private static final Junction<MethodDescription> NOT_HIBERNATE_GENERATED = not( nameStartsWith( "$$_hibernate_" ) );
+	private static final ModifierContributor.ForField REMOVE_FINAL_MODIFIER = new ModifierContributor.ForField() {
```
Sanne, line 53: Interesting, I'm learning a lot here :)

Sanne (approved):
Excellent, thanks!

hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/PersistentAttributeTransformer.java
yrodiere: It is. However, I think Panache only performs extended enhancement, and I think this enhancement supposedly avoids transforming field access within the class declaring the field?

Not sure if it's trying or succeeding, though; ORM is trying but failing, as far as I can see, but that's another problem.

In any case, we'll have to add tests in Quarkus. I'll submit a PR.

PR #4237, opened 2021-09-21
Backport HHH-14828 to branch 5.4 - Bytecode enhancement generates invalid bytecode for final fields

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14828

Backport of #4231 

I did not intend to backport this initially, but I think it will be necessary to fix https://github.com/quarkusio/quarkus/issues/20186 in old versions of Quarkus.


yrodiere:
@Sanne Could you please confirm it's fine to backport this to 5.4?

Sanne:
> @Sanne Could you please confirm it's fine to backport this to 5.4?

yes it's fine but I wouldn't rush it for non-trivial patches. WDYT of leaving this here for a couple weeks and see feedback from 5.5+?

yrodiere:
Works for me. But I'll bet we will just forget about it :D Let's find out!

PR #4238, opened 2021-09-21
HHH-14837 - Move to Jakarta EE

sebersole:
Not stoked about:

1) the `jaxb2-basics` stuff that requires pushing those patched jars into git, but they are only used at build-time so it is not a huge deal.  See https://github.com/highsource/jaxb2-basics/pull/129
2) The JCache/Ehcache one is a bit more troubling.  At the moment, we do not have a working second level cache provider with this move to Jakarta EE.  Ehcache is the hold up - it uses JAXB and would need to update for using "Jakarta XML Binding" (just like we had to do).  See https://github.com/ehcache/ehcache3/issues/2946


sebersole:
The failures are related to spatial depending on tests from hibernate-core.  I am not a fan of this.  These are called test fixtures and really belong in a common place (aka hibernate-testing).

Apparently spatial has been re-enabled since I started this branch

sebersole:
Somehow the rebase messed up all of the transformations on hibernate-core :sigh:

sebersole:
Should never have pulled in the most recent work from upstream wip/6.0 after I started this...

The rebase killed a lot of the transformations and I am now trying to dig through a lot of envers test failures.

Consider this a draft for now

sebersole:
Closing in favor of https://github.com/hibernate/hibernate-orm/pull/4243

PR #4243, opened 2021-09-22
HHH-14837 - Move to Jakarta EE

sebersole:
preliminary transformation;
only `javax.persistence.*` settings supported atm


sebersole:
As mentioned in the second commit, there is still some work to do wrt settings.

I started making `jakarta.persistence.*` take precedence over `javax.persistence.*` and giving deprecation warnings when using `javax.persistence.*`.  Still have more to handle

sebersole:
This one should be ready to go

sebersole:
And I really want to get this integrated today so I can stop constantly rebasing ;)

sebersole:
The test failures are related to Bean Validation integration.  Not quite sure yet.

sebersole:
Strange.  They don't happen here.  :shrug:

Sanne:
Impressive! 

It's hard to get in details, but some questions:
 - What's the reason to committ the jaxb2-basics jars ?
 - `JPA_VALIDATION_FACTORY` seems intentionally reverted to "javax.persistence.validation.factory" .. was it really intentional?
 - javadoc links commented out when pointing to 'https://docs.jboss.org/hibernate/beanvalidation/spec/2.0/api/' and 'https://docs.jboss.org/cdi/api/2.0/' ? Is this handled by `javadocClasspath libraries.jakarta_validation` ?
 - in `gradle/java-module.gradle` you made some changes for the HeapDumpPathProvider ..  you want these changes included here?
 - `CollectionHelper` seems to have some unrelated changes? (They don't look bad, just wondering if it was intentional.)

sebersole:
> * What's the reason to committ the jaxb2-basics jars ?

I had to patch these.  It was a pain to publish them and we only use them at runtime, so...


> * `JPA_VALIDATION_FACTORY` seems intentionally reverted to "javax.persistence.validation.factory" .. was it really intentional?

See `JPA_VALIDATION_FACTORY` and `JAKARTA_JPA_VALIDATION_FACTORY`.  We keep both sets to better support reading both sets


> * javadoc links commented out when pointing to 'https://docs.jboss.org/hibernate/beanvalidation/spec/2.0/api/' and 'https://docs.jboss.org/cdi/api/2.0/' ? Is this handled by `javadocClasspath libraries.jakarta_validation` ?

Those link to the Java EE versions of the Javadoc for those specs.  I was not able to find (yet) Jakarta versions.


> * in `gradle/java-module.gradle` you made some changes for the HeapDumpPathProvider ..  you want these changes included here?

Yes


> * `CollectionHelper` seems to have some unrelated changes? (They don't look bad, just wondering if it was intentional.)

I believe I use them in tests

scottmarlow:
Consider changing `JPA` to `Jakarta Persistence` in documentation as per https://jakarta.ee/legal/acronym_guidelines including `JPA` in Oracle/JCP acronyms list.

sebersole:
Of course, almost immediately after I found the, :laughing: 

- https://javadoc.io/doc/jakarta.enterprise/jakarta.enterprise.cdi-api/latest/
- https://javadoc.io/doc/jakarta.validation/jakarta.validation-api/latest/

sebersole:
> Of course, almost immediately after I found the, 
> 
> * https://javadoc.io/doc/jakarta.enterprise/jakarta.enterprise.cdi-api/latest/
> * https://javadoc.io/doc/jakarta.validation/jakarta.validation-api/latest/

Actually, I take that back... the link I use (https://jakarta.ee/specifications/platform/9/apidocs/) already includes CDI and BV

sebersole:
> Consider changing `JPA` to `Jakarta Persistence` in documentation as per https://jakarta.ee/legal/acronym_guidelines including `JPA` in Oracle/JCP acronyms list.

@Sanne I ran across this section you added in `Wildfly.adoc`, but as this is on the 6.0 branch, the wording needs to be different.  If you could fix that, that would be great...

```
== Hibernate ORM within WildFly

The https://wildfly.org/[WildFly application server] includes Hibernate ORM as the default Jakarta Persistence provider out of the box.

In previous versions of Hibernate ORM, we offered a "feature pack" to enable anyone to use the very latest version in
WildFly as soon as a new release of Hibernate ORM was published.

Unfortunately, since version 5.5 is upgrading to Jakarta Persistence 3.0 and targets integration with components of the Jakarta
EE 9 stack, such feature had to be disabled.

As soon as WildFly releases a Jakarta EE 9 compatible server it might be possible to re-introduce such a feature, but
we can't guarantee that we will do this as the server changed the tooling to define such packs.
```

sebersole:
> Consider changing `JPA` to `Jakarta Persistence` in documentation as per https://jakarta.ee/legal/acronym_guidelines including `JPA` in Oracle/JCP acronyms list.

I'm not sure what you mean by "Oracle/JCP acronyms list", but I updated the references I could find

beikov:
The tests `JakartaSchemaToolingTests` fail for Sybase saying that `Unable to determine Dialect without JDBC metadata (please set 'javax.persistence.jdbc.url', 'hibernate.connection.url', or 'hibernate.dialect')`.

From what I can see, the database name according to the JDBC driver is `ASE` which we don't match yet, but this just leads to matching the JDBC URL later on which normally works, except for this test for some reason. The reason being, we use the `AvailableSettings.URL` property in `org.hibernate.engine.jdbc.connections.internal.DriverManagerConnectionProviderImpl#buildCreator` to determine the URL which is probably not set in these tests for what I can see. Maybe we should also query the JPA and Jakarta URLs there? Apart from that, I suggest we add `|| "ASE".equals( databaseName )` to `org.hibernate.dialect.Database#isASE`

dreab8:
Hi Steve, the `JakartaSchemaToolingTests` is failing on Sybase because the test builds the SF without specifying any dialect but the resolution using the `DataBaseMetadata` fails because `DataBaseMetadata#getDatabaseProductName()` returns `ASE` when we expects `SYBASE`.

Sanne:
wow nice! congratulations

scottmarlow:
+10000

mkurz:
@sebersole I am pretty sure you made a mistake in this PR, look at these lines:
https://github.com/hibernate/hibernate-orm/blob/6746c30275b6600622d82613dc70b1fbf97a2fdf/hibernate-core/src/main/java/org/hibernate/boot/beanvalidation/BeanValidationIntegrator.java#L39-L43
You replaced `javax.validation` in the  `BV_CHECK_CLASS` variable, but that shouldn't happens since this is meant as fallback (like `[JAKARTA_]MODE_PROPERTY` handles that)

I reported that as https://hibernate.atlassian.net/browse/HHH-16538 and opened this pull request as fix: https://github.com/hibernate/hibernate-orm/pull/6489

PR #4258, opened 2021-09-27
HHH-14848 - Move Ant tasks + support from hibernate-core into tooling…

sebersole:
…/hibernate-ant


sebersole:
Integrated upstream

PR #4265, opened 2021-09-29
HHH-14858 - Implement JRT Module Readers

GedMarc:
This PR allows reading the JRT virtual file system, which uses a new URL format structure to reference to files located in the module VFS


sebersole (commented):
I don't know much about the JRT format, but lgtm

yrodiere:
Maybe a test would be a good idea? :)

GedMarc:
Yeah definitely - difficult to do though as you need a 100% module-info chain to create a native runtime, and gradle is not quite there yet with jlink, jpackage and jmod support

Hmm, can fong the test though, as it's just a URL handler with a string extract - will get one in for you @yrodiere, but without building a JLink/JMod package,  you never see the JRT filesystem ;) Would this be acceptable or do we need a non-flaky non-fudged test xD

Can I do something like we did for Jackson? (https://github.com/FasterXML/jackson-jdk11-compat-test)

I had to create a separate project that builds up the JLink artifact, and then executes the tests through the native runtime to replicate all JRT related issues (being a fully read only filesystem)

yrodiere:
That seems awfully complicated; no, I don't think this is worth creating a new module and trying to use jlink.

But `jrt:/` URLs only require your test to run in the modulepath, right? You don't need jlink, native compilation, or anything like that?

There already is a module that runs tests in the modulepath: `hibernate-integrationtest-java-modules`. So you could just add a (unit) test in `hibernate-integrationtest-java-modules/src/test/java`. I would keep it simple: just instantiate `JRTInputStreamBasedArchiveDescriptor`, pass it a `jrt:/` URL to a random module that you know will be in the modulepath, and check that it collects an entry that you know should be there.

Maybe that won't reliably test that your use case is addressed, but at least this will check that your implementation successfully retrieves entries from `jrt:/` URLs.

GedMarc:
I definitely agree with you that's it is awfully complicated :) 
The JRT only exists for native runtimes, so JLink etc, no other way to access that VFS or use/create the jrt:// urls that I know of  ;)
Under class path mode (or loose module mode, a.k.a. not a physical module distribution but running off JAR's instead of compiled JMOD's), the JRT system isn't even included as loaded on a JDK (cause you can't use the new fast-class constructor, the new SPI service, resource finder etc and all the other pretty things that come with native modules and do not operate or even load with classpath/jar files)

What we are looking at here is similar to this https://stackoverflow.com/questions/54140750/java-nio-cant-read-files-from-jrt-image - except it's a jar archive read instead of an Io resource read

I'm creating a ```URLStreamHandlerProvider``` that will fudge the JRT URL and locate accordingly in a jar file, that will at least give us an authentic test on the URL handling and reading :)

hibernate-core/src/main/java/org/hibernate/boot/archive/internal/JRTFileBasedArchiveDescriptor.java
```diff
@@ -0,0 +1,139 @@
...
+	public JRTFileBasedArchiveDescriptor(
+			ArchiveDescriptorFactory archiveDescriptorFactory,
+			URL archiveUrl,
+			String entry) {
+		super( archiveDescriptorFactory, archiveUrl, entry );
+	}
+
+	@Override
+	public void visitArchive(ArchiveContext context) {
+		final String jrtPart = getArchiveUrl().getPath();
```
Sanne, line 44: Ok the main problem here is that you've been using syntax and JDK APIs which won't compile with Java 8: this branch of Hibernate ORM still aims to be compatible with Java 8 users, which implies we run the build with Java 8, only integration tests are run with a matrix of multiple JDKs.

We have some options:

- use a multi-release jar (not my favourite, tools don't really deal well with those)
- we merge this in ORM 6 exclusively, assuming @sebersole and the other developers agree to drop Java 8 compatibility (I was going to open this dicussion anyway as there are more benefits into this, but it implies it won't make it in 5.x version and you'll have to wait a bit longer to have this)
- perhaps it's possible for you to use JDK8 compatible APIs? I only see errors caused by `FileVisitResult`, `SimpleFileVisitor`, `Paths` ... inconvenient but don't seem stricly related to needing Jigsaw related APIs.

Sanne (changes requested):

hibernate-core/src/main/java/org/hibernate/boot/archive/internal/JRTFileBasedArchiveDescriptor.java
GedMarc: Understood

hibernate-core/src/main/java/org/hibernate/boot/archive/internal/JRTFileBasedArchiveDescriptor.java
```diff
@@ -0,0 +1,139 @@
...
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.boot.archive.internal;
+
+import org.hibernate.boot.archive.spi.*;
+
+import java.io.*;
```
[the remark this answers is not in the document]
GedMarc: I'll unlearn and relearn why so at a later point :) making all the changes

hibernate-core/src/main/java/org/hibernate/boot/archive/internal/JRTFileBasedArchiveDescriptor.java
```diff
@@ -0,0 +1,150 @@
...
+			{
+				throw new IOException(
+						String.format(
+								"Unable to access stream from jrt ref [%s] for entry [%s]",
+								lateralPath,
+								lateralPath.toString()
+						)
+				);
+			}
+			final ArchiveEntry entry = new ArchiveEntry()
```
GedMarc, line 118: I'm not really comfortable with this, but it seemed to be the standard way in the other descriptors

GedMarc:
recreating under correct commit strategy with test cases

PR #4268, opened 2021-09-30
HHH-14858 Initial commit for JRT handling

GedMarc:
Adds an archive descriptor for the JRT virtual file system, using the JDK 8 io.nio. package


beikov:
Please create a new PR against the `main` branch and also add tests for this new feature.

PR #4269, opened 2021-09-30
Various improvements

beikov:
- Cancel old pipeline runs if a new commit arrives before the old build finishes
- Temporarily disable testing environments that are also tested on GH actions
- Fix some string literal parsing issues regarding escape sequences
- Improve SQL rendering performance by avoiding intermediate String objects
- Remove some row selection uses
- Get rid of RowSelection, QueryParameters and related deprecated functionality
- Get rid of deprecated dialect methods, JoinFragment, LegacyLimitHandler, LegacySequenceSupport, CaseFragment, SelectFragment etc.


dreab8:
LGTM, I think we only need to create a PR for main where we deprecate the classes you have removed and that were non already drepecated

beikov:
I was hoping that I wouldn't have to do that since Jan is working on the binary compatibility report which we could use for determining what we still need to do.

PR #4273, opened 2021-10-04
Expose `JpaMetamodel` and `MappingMetamodel` through `SessionFactory#getMetamodel()`

yrodiere:
This provides users with a very simple migration path to these new interfaces and away from the deprecated `org.hibernate.Metamodel`/`org.hibernate.metamodel.spi.MetamodelImplementor`.

Without this patch, users have to rely on completely new getters to retrieve a `JpaMetamodel`; they cannot use `SessionFactory#getMetamodel()`. This seems strange, considering that:

* `JpaMetamodel` is an extension of `jakarta.persistence.metamodel.Metamodel`
* `SessionFactory#getMetamodel` is an extension of `EntityManager#getMetamodel()`
* `EntityManager#getMetamodel()` which returns a `jakarta.persistence.metamodel.Metamodel`


sebersole (commented):
Need to run to an appointment and will look more when I get back.  One thing specifically I need to look through is how you are using MetamodelImpl versus JpaMetamodelImpl classes.  Should be only one impl imo one first thought, but I'll look some more

PR #4276, opened 2021-10-06
Wip/6.0 cleanup notyetimplemented

dreab8:


sebersole:
I started trying to resolve these conflicts, but it became non-simple.  Let's look at this after the holidays if you don't get to it by then

dreab8:
I'm going to resolve the conflicts next week, thanks Steve!

beikov (commented):
After fixing the compiler issues and when the tests run through this can be merged IMO.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/SqmPluralPartJoin.java
```diff
@@ -111,4 +114,15 @@ public String toString() {
...
+	@Override
+	public DomainResult<T> createDomainResult(
+			String resultVariable, DomainResultCreationState creationState) {
+		throw new NotYetImplementedFor6Exception( getClass() );
+	}
+
+	@Override
+	public void applySqlSelections(DomainResultCreationState creationState) {
+		throw new NotYetImplementedFor6Exception( getClass() );
+	}
```
beikov, line 127: Just like you did for `SqmSingularJoin`, `SqmPluralPartJoin`, `SqmRoot`, `SqmDynamicInstantiation`, `SqmFunction`, `SqmLiteral`, `SqmCaseSearched`, `SqmJpaCompoundSelection` and `SqmCollectionSize` don't need to implement `DomainResultProducer` because these are always translated to `SqmPathInterpretation`, so you can remove these dummy implementations.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3307,6 +3334,23 @@ public Expression visitMinIndexPath(SqmMinIndexPath<?> path) {
 		return createMinOrMaxIndexOrElement( path, true, false );
 	}
 
+	@Override
+	public Expression visitCorrelation(SqmCorrelation<?, ?> correlation) {
+		return visitRootPath( correlation.getCorrelatedRoot() );
```
beikov, line 3339:
```suggestion

		final TableGroup resolved = getFromClauseAccess().findTableGroup( correlation.getNavigablePath() );
		if ( resolved != null ) {
			log.tracef( "SqmCorrelation [%s] resolved to existing TableGroup [%s]", correlation, resolved );
			return visitTableGroup( resolved, correlation);
		}

		throw new InterpretationException( "SqmCorrelation not yet resolved to TableGroup" );
```

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3453,6 +3497,11 @@ public Object visitMapEntryFunction(SqmMapEntryReference<?, ?> entryRef) {
 						.resolveDescriptor( Map.Entry.class );
 				return new SqmMapEntryResult<>( indexResult, valueResult, resultVariable, mapEntryDescriptor );
 			}
+
+			@Override
+			public void applySqlSelections(DomainResultCreationState creationState) {
+				throw new NotYetImplementedFor6Exception( getClass() );
```
beikov, line 3503:
```suggestion
				throw new UnsupportedOperationException();
```

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -2778,6 +2777,15 @@ private void prepareForSelection(SqmPath<?> joinedPath) {
...
 				navigablePath = ( (SqmTreatedRoot<?, ?>) joinedPath ).getWrappedPath().getNavigablePath();
+
+				final TableGroup tableGroupForTreatedRoot = getTableGroupForTreatedRoot(
+						navigablePath,
+						fromClauseIndex
+				);
+				if ( tableGroupForTreatedRoot != null ) {
+					fromClauseIndex.register( joinedPath, tableGroupForTreatedRoot );
+					return;
+				}
```
beikov, line 2788: I think you should add the following to line 2750 in prepareReusablePath

```
if ( sqmPath instanceof SqmTreatedPath<?, ?> ) {
	fromClauseIndex.register( (SqmPath<?>) sqmPath, tableGroup );
}
```

right after `implicitJoinChecker.accept( newTableGroup );`

beikov (changes requested):

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3054,6 +3052,16 @@ else if ( modelPart instanceof EntityMappingType ) {
 			else {
 				expandToAllColumns = false;
 			}
+
+			final EntityValuedModelPart mapping;
```
beikov, line 3056: Unfortunately, I think we will have to add the type as additional parameter to `EntityValuedPathInterpretation.from` because this change would also change the meaning of `where treat(e.association) = e2` to always match the primary key, although it should match the FK parts.

beikov (approved):

beikov:
Seems there is a problem with `BulkManipulationTest.testInsertWithManyToOne`

dreab8:
@beikov I fixed it :)

PR #4287, opened 2021-10-12
HHH 14873 - PostFlush / PreFlush and lazy loading lead to ConcurrentModificationException

karge-itestra:
See issue description at https://hibernate.atlassian.net/browse/HHH-14873

You may step back one commit (revert the solution) to see that the newly added test really fails.


sebersole:
Add a setting which opts in to this behavior.  Easy.

gavinking:
I don't think we should open this can of worms. We really have no idea what the side-effects of this might be.

And I note that my colleagues have shown a similar lack of enthusiasm for this.

So I'm going to close this. Sorry.

PR #4306, opened 2021-10-16
HHH-14877 - FetchMode.SUBSELECT ignored

sebersole:
Some of the original subselect tests fail and I need to come back and see why.  Those tests are just silly, way complex.

So I added some simple ones which all pass.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/AbstractNaturalIdLoader.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/AbstractNaturalIdLoader.java
@@ -16,7 +16,6 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 
 import org.hibernate.HibernateException;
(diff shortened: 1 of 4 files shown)


sebersole:
Most of your comments are true.  This was essentially a draft so I could see the results on the matrix

beikov:
`SubselectFetchWithFormulaTransactSqlTest` fails on MSSQL and Sybase due to a configuration error, I guess the FQN of an entity changed but is missing somewhere. The Oracle failures are the same as on Jenkins: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-orm-pipeline/detail/PR-4306/4/tests

sebersole:
> > And personally I found that name to be confusing. Which SQL options? I had to physically trace through the method to even find out what it was trying to do.
> 
> Ok, what do you think about `omitLimitAndLockOptions` or something like that?

I think that's better

PR #4310, opened 2021-10-18
HHH-14876 Parameter padding for IN clauses doesn't work in Hibernate 6.0.0.Beta1

dreab8:
https://hibernate.atlassian.net/browse/HHH-14876


beikov:
Also, it seems `InClauseParameterPaddingCriteriaTest#testInClauseParameterPadding` and `InClauseParameterPaddingTest#testInClauseParameterPadding` fail for SQL Server and Sybase for some reason. I guess we will have the same failures when moving the implementation.

sebersole:
This should leverage `BatchLoadSizingStrategy`.  The naming is slanted towards batch fetching and multi-id loading, but the concept is the same

hibernate-core/src/test/java/org/hibernate/orm/test/loading/multiLoad/MultiLoadSubSelectCollectionTest.java
```diff
@@ -88,25 +88,11 @@ public void testSubselect(SessionFactoryScope scope) {
...
-					for ( int i = 50; i < 56; i++ ) {
-						assertFalse( Hibernate.isInitialized( list.get( i ).children ) );
-					}
-
-					// When the 51st collection gets initialized, the remaining collections should
-					// also be initialized.
-					Hibernate.initialize( list.get( 50 ).children );
-
-					for ( int i = 50; i < 56; i++ ) {
+					for ( int i = 0; i < 56; i++ ) {
```
beikov, line 95: How about using a special dialect here that has a limit of 50 to retain the same test logic?

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -3068,7 +3069,22 @@ public NameQualifierSupport getNameQualifierSupport() {
 	protected final BatchLoadSizingStrategy STANDARD_DEFAULT_BATCH_LOAD_SIZING_STRATEGY = new BatchLoadSizingStrategy() {
 		@Override
 		public int determineOptimalBatchLoadSize(int numberOfKeyColumns, int numberOfKeys) {
-			return 50;
+			int paddedSize = MathHelper.ceilingPowerOfTwo( numberOfKeys );
```
beikov, line 3072: IMO it's fine to always do the padding here, and we should also enabled padding (`org.hibernate.cfg.AvailableSettings#IN_CLAUSE_PARAMETER_PADDING`) by default in 6.0, but don't you think this should be controlled by the setting? Or is the setting for IN predicates only so it doesn't need to be respected?

hibernate-core/src/test/java/org/hibernate/orm/test/loading/multiLoad/MultiLoadSubSelectCollectionTest.java
dreab8: I think we have to change `BatchLoadSizingStrategy` implementation, at the moment we are not taking into account any Dialect limit

beikov: It is taken into account, here: https://github.com/hibernate/hibernate-orm/pull/4310/files#diff-db73bbf122356d881edeb9b3d69cc9ad5b469229214d289131fd6c98c6b7b4c0R3077

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
dreab8: probably we have to check the setting

dreab8: done

hibernate-core/src/test/java/org/hibernate/orm/test/loading/multiLoad/MultiLoadSubSelectCollectionTest.java
dreab8: Done!

beikov (approved):
Looks good now. As a follow up, we should IMO enable padding by default in 6.0

hibernate-testing/src/main/java/org/hibernate/testing/orm/junit/Jpa.java
```diff
@@ -108,6 +109,8 @@
 	 */
 	boolean loadByIdComplianceEnabled() default false;
 
+	Class<? extends StatementInspector> statementInspectorClass() default StatementInspector.class;
```
jrenaat, line 112: Ihad already added a `SettingProvider[] settingProviders() default {};` to `@Jpa` in [https://github.com/hibernate/hibernate-orm/pull/4322/](https://github.com/hibernate/hibernate-orm/pull/4322/), since that also covers the need to pass a StatementInspector class, I may simply revert this in my PR, wdyt?

PR #4311, opened 2021-10-19
HHH-14887 forking of the compilation process

ChenZhangg:
According to [Process forking options](https://docs.gradle.org/current/userguide/performance.html#forking_options), Gradle will run all tests in a single forked VM by default. This can be problematic if there are a lot of tests or some very memory-hungry ones.
one option is to fork a new test VM after a certain number of tests have run. So our recommendation is to configure "forkEvery" and we give a specific value of 100

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,6 +61,10 @@ nexusPublishing {
 }
 
 allprojects {
+	tasks.withType(Test).configureEach {
+        	forkEvery = 100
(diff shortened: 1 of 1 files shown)


beikov:
Thanks for the PR, but I think we will need a different approach for this i.e. using JUnit 5 test parallelization, so I'll close this.

PR #4312, opened 2021-10-19
HHH-14888 Running the Compiler as a separate process to shorten the build time

ChenZhangg:
[Compiler daemon](https://docs.gradle.org/current/userguide/performance.html#compiler_daemon),We found that the project had over 1000 Java and Groovy files, so running the Compiler as a separate process shortened the build time of this project

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,6 +61,9 @@ nexusPublishing {
 }
 
 allprojects {
+	tasks.withType(JavaCompile).configureEach {
+        	options.fork = true
+    	}
 	repositories {
 		mavenCentral()
 		//Allow loading additional dependencies from a local path;


beikov:
My guess is, that the compiler JVMs use some kind of automatic memory limit, which is too low, leading GC overhead limit exceeded. The reason for this being too low is that all of a sudden we fork various JVMs, leading to a overall memory pressure situation on that machine. For H2 it even seems that the compilation deadlocks somehow. On Java 11 it seems to work fine. Either way, I think this needs some more fine tuning until we can apply this change. Maybe we can set up a maximum amount of forks? Configure some kind of global memory limit?

beikov:
Thanks, but since this doesn't work, I'll close the PR now.

PR #4318, opened 2021-10-20
HHH-14892 Parallel test with GRADLE

ChenZhangg:
[Parallel test execution maxParallelForks](https://docs.gradle.org/current/userguide/performance.html#parallel_test_execution),running multiple test cases in parallel is useful and helpful when there are several CPU cores.

At the original PR, we added four setting and it failed, we want to know if it can build successfully with the setting of "parallel test"

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,6 +61,9 @@ nexusPublishing {
 }
 
 allprojects {
+	tasks.withType(Test).configureEach {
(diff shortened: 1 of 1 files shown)


beikov:
Thanks for the PR, but I think we will need a different approach for this i.e. using JUnit 5 test parallelization, so I'll close this.

PR #4327, opened 2021-10-26
Various fixes

beikov:
* Handle to-one nullability/optionality correctly and get rid of join type altering
* Implement parameter list expansion for native queries
* Fix empty subselect fetched collection initialization
* Implement support for nested table group joins to allow joins on the map-key
* Replace `getTableReference` with `resolveTableReference` where appropriate to distinguish which calls can cause table reference joins to be created
* Fix some table reference resolving issues with inverse embeddable model parts
* Use a Fetch for entity ids instead of a DomainResult
* Fix bidirectional fetching for collection initializtion
* Implement table reference join pruning for treat usages
* Implement strict JPA compliance for different parameter styles
* Ensure From nodes in Criteria are unique
* Add unique key support to DelayedEntityFetch
* Check if FetchParent is enhanced for lazy loading for DelayedEntityFetch
* Register entity instances under all possible EntityUniqueKey
* Introduce EntityJavaTypeDescriptor that implements equality based on object identity
* Remodel `@MapKey` support to not create subqueries in the on-clause anymore
* Make sure the index table group is reused for the to-one association a `@MapKey` refers to
* Consistently register collection part table groups
* Implement support for FK optimization for EntityCollectionPart

https://hibernate.atlassian.net/browse/HHH-9862
https://hibernate.atlassian.net/browse/HHH-8206


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/EntityMappingType.java
```diff
@@ -195,8 +196,7 @@ default EntityMappingType getRootEntityDescriptor() {
 		return superMappingType.getRootEntityDescriptor();
 	}
 
-	default TableReference locateTableReference(TableGroup tableGroup) {
-		return tableGroup.getPrimaryTableReference();
+	default void pruneForSubclasses(TableGroup tableGroup, Set<String> treatedEntityNames) {
```
dreab8, line 199: Can you please a javadoc to explain the use of this method?

beikov: I'll add a comment

PR #4330, opened 2021-10-27
Order-by nested embedded + to-one fk

fax4ever:
With this pull request I tried to address the case of the test I added here: OrderByEmbeddableToOneTest

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ordering/ast/DomainPathContinuation.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ordering/ast/DomainPathContinuation.java
@@ -10,7 +10,6 @@ import org.hibernate.metamodel.mapping.EmbeddableMappingType;
 import org.hibernate.metamodel.mapping.EmbeddableValuedModelPart;
 import org.hibernate.metamodel.mapping.ModelPart;
 import org.hibernate.metamodel.mapping.internal.AbstractDomainPath;
-import org.hibernate.metamodel.mapping.internal.ToOneAttributeMapping;
 import org.hibernate.metamodel.mapping.ordering.TranslationContext;
 import org.hibernate.query.NavigablePath;
 
@@ -20,9 +19,10 @@ import org.hibernate.query.NavigablePath;
  * @author Steve Ebersole
  */
 public class DomainPathContinuation extends AbstractDomainPath {
-	private final NavigablePath navigablePath;
+	protected final NavigablePath navigablePath;
+	protected final ModelPart referencedModelPart;
+
 	private final DomainPath lhs;
-	private final ModelPart referencedModelPart;
 
 	public DomainPathContinuation(NavigablePath navigablePath, DomainPath lhs, ModelPart referencedModelPart) {
 		this.navigablePath = navigablePath;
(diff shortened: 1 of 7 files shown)


fax4ever:
Thanks @dreab8 and @beikov.
I applied the suggestion.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ordering/ast/DomainPathContinuation.java
```diff
@@ -64,9 +65,28 @@ public SequencePart resolvePathPart(
...
+			ToOneAttributeMapping toOneAttribute = (ToOneAttributeMapping) referencedModelPart;
+			if ( name.equals( toOneAttribute.getTargetKeyPropertyName() ) ) {
+				final ModelPart subPart = toOneAttribute.findSubPart( name );
+				if ( subPart == null ) {
+					throw new PathResolutionException(
+							"Could not resolve path token : " + referencedModelPart + " -> " + name
+					);
+				}
+
+				return new DomainPathContinuation(
```
beikov, line 78: Actually this is not totally correct and to be safe, I would suggest you introduce a new `SequencePart` implementation for `ToOneAttributeMapping` that is able to handle these special paths. The implementation would use the `targetKeyPropertyNames` and for every sub-part, prune possible further paths. If no paths are possible, this is not a simple FK reference and it should fail. The implementation could look roughly like this:

```
class FkDomainPathContinuation extends DomainPathContinuation {
	@Override
	public SequencePart resolvePathPart(
			String name,
			boolean isTerminal,
			TranslationContext translationContext) {
		List<String> newPossiblePaths = new ArrayList<>(possiblePaths.size());
		for (String path : possiblePaths) {
			if (path.startsWith(name)) {
				if (path.length() == name.length()) {
					return new DomainPathContinuation(
						navigablePath.append( name ),
						this,
						((ModelPartContainer) referencedModelPart).findSubPart(name, null)
					);
				}
				else if (path.charAt(name.length()) == '.') {
					newPossiblePaths.add(path.substring(name.length() + 1));
				}
			}
		}
		if (newPossiblePaths.isEmpty()) {
			throw new PathResolutionException(
				"Domain path of type `" +  referencedModelPart.getPartMappingType() +
				"Domain path of type `" + referencedModelPart.getPartMappingType() +
				"` -> `" + name + "`"
			);
		}
		return new FkDomainPathContinuation(
			navigablePath.append( name ),
			this,
			((ModelPartContainer) referencedModelPart).findSubPart(name, null),
			newPossiblePaths                                      
		);
	}
}
```

fax4ever: Interesting. If I am not mistaken, we must also consider here the case in which the key is composed? I tried to address this comment with the commit https://github.com/hibernate/hibernate-orm/pull/4330/commits/d81e3c5c9114ec8dc4783da6d8aedea33b7832b6. I'm on the right track? Many thanks @beikov!

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ordering/ast/PluralAttributePath.java
```diff
@@ -55,21 +56,25 @@ public DomainPath resolvePathPart(
...
-						pluralAttributeMapping.getElementDescriptor()
-				);
-			}
-			else {
+			if ( subPart instanceof EmbeddableValuedModelPart ) {
 				return new DomainPathContinuation( navigablePath.append( name ), this, subPart );
 			}
+			if ( subPart instanceof ToOneAttributeMapping ) {
+				return new FkDomainPathContinuation( navigablePath.append( name ), this,
+													(ToOneAttributeMapping) subPart );
```
beikov, line 64: Minor formatting issue

beikov (approved):

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ordering/ast/PluralAttributePath.java
fax4ever: done, thanks!

fax4ever:
Merged, thanks!

PR #4333, opened 2021-10-29
Log first Exception when MappingModelCreationProcess cannot complete

dreab8:
the aim is to make a little easier to spot what went wrong when the `MappingModelCreationProcess` does not complete

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationProcess.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationProcess.java
@@ -7,6 +7,8 @@
 package org.hibernate.metamodel.mapping.internal;
 
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
@@ -92,8 +94,8 @@ public class MappingModelCreationProcess {
 	private void executePostInitCallbacks() {
 		MappingModelCreationLogger.LOGGER.debugf( "Starting post-init callbacks" );
 
+		Map<PostInitCallbackEntry, Exception> exceptions = new HashMap<>();
 		while ( postInitCallbacks != null && !postInitCallbacks.isEmpty() ) {
-
 			// copy to avoid CCME
 			final ArrayList<PostInitCallbackEntry> copy = new ArrayList<>( postInitCallbacks );
 
@@ -108,6 +110,7 @@ public class MappingModelCreationProcess {
 					if ( completed ) {
 						anyCompleted = true;
 						postInitCallbacks.remove( callbackEntry );
+						exceptions.remove( callbackEntry );
 					}
 				}
 				catch (Exception e) {
@@ -118,6 +121,7 @@ public class MappingModelCreationProcess {
 						);
 						throw e;
 					}
+					exceptions.put( callbackEntry, e );
 
 					final String format = "Mapping-model creation encountered (possibly) transient error : %s";
 					if ( MappingModelCreationLogger.TRACE_ENABLED ) {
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationProcess.java
```diff
@@ -130,20 +134,27 @@ private void executePostInitCallbacks() {
...
-				);
-
-				buff.append( EOL );
-
-				throw new IllegalStateException( buff.toString() );
+				if ( !exceptions.isEmpty() ) {
+					throw new IllegalStateException(
+							"MappingModelCreationProcess$PostInitCallback queue could not be processed...",
+							exceptions.entrySet().iterator().next().getValue()
+					);
```
beikov, line 143: I would prefer if we add all exceptions as suppressed exceptions

NathanQingyangXu: Or create a composite exception as http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/exceptions/CompositeException.html
We might well create a common class to be reused elsewhere.

sebersole (commented):
Is this still relevant?

dreab8:
I think it is, by the way I'm going to rebase it and apply Christian suggestion

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationProcess.java
```diff
@@ -129,20 +133,33 @@ private void executePostInitCallbacks() {
...
+					postInitCallbacks.forEach(
+							callbackEntry -> buff.append( EOL )
+									.append( "        - " ).append( callbackEntry )
+					);
+
+					buff.append( EOL );
+
+					throw new IllegalStateException( buff.toString() );
+				}
```
beikov, line 162:
```suggestion
				final StringBuilder buff = new StringBuilder(
						"PostInitCallback queue could not be processed..."
				);
				postInitCallbacks.forEach(
						callbackEntry -> buff.append( EOL )
								.append( "        - " ).append( callbackEntry )
				);

				buff.append( EOL );
				final IllegalStateException illegalStateException = new IllegalStateException( buff.toString() );
				for ( Map.Entry<PostInitCallbackEntry, Exception> entry : exceptions.entrySet() ) {
					illegalStateException.addSuppressed( entry.getValue() );
				}
				throw illegalStateException;
```

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationProcess.java
```diff
@@ -92,8 +94,8 @@ private void execute() {
 	private void executePostInitCallbacks() {
 		MappingModelCreationLogger.LOGGER.debugf( "Starting post-init callbacks" );
 
+		Map<PostInitCallbackEntry, Exception> exceptions =new HashMap<>();
```
beikov, line 97:
```suggestion
		Map<PostInitCallbackEntry, Exception> exceptions = new HashMap<>();
```

beikov (commented):
Looking good. Just one small change and then it's perfect ;)

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationProcess.java
dreab8: Not sure about that, the
```
else {
					final StringBuilder buff = new StringBuilder(
							"PostInitCallback queue could not be processed..."
					);
					postInitCallbacks.forEach(
							callbackEntry -> buff.append( EOL )
									.append( "        - " ).append( callbackEntry )
					);

					buff.append( EOL );

					throw new IllegalStateException( buff.toString() );
				}
```
should not be necessary, I added it as a safety net in case someone make a mistake, by example not throwing a `NonTransientException` and  with the change you propose the first exception is added as a suppressed one

beikov: My suggested change just reports all callback entries, adding all exceptions as suppressed ones. IMO that's the most natural way to represent this which will include all information and be formatted nicely by default.

dreab8: applied your suggestion

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationProcess.java
dreab8: fixed! Thanks

beikov (approved):

PR #4335, opened 2021-10-29
Updated test to JUnit5 to remove reference to Dialect subclass

jrenaat:
Moved it back for now to the old location, since it fails with a (probably)
non-related error that will have to be looked at as soon as there is an
implementation for composite sub-types


hibernate-core/src/test/java/org/hibernate/test/sql/hand/query/NativeSQLQueriesTest.java
```diff
@@ -0,0 +1,883 @@
...
+					session.persist(jboss);
+					session.persist(gavin);
+					session.persist(emp);
+
+					List l = session.createNativeQuery( getOrgEmpRegionSQL(), "org-emp-regionCode" ).list();
+					assertEquals( l.size(), 2 );
+
+					l = session.createNativeQuery( getOrgEmpPersonSQL(), "org-emp-person" ).list();
+					assertEquals( l.size(), 1 );
+
```
NathanQingyangXu, line 255: It is common defect to mess up the parameter order of assertEquals(). It is not a big deal when it succeeds, but would generate confusing message when it fails.

beikov:
Thanks!

PR #4340, opened 2021-11-04
HHH-14839 allow lazy association properties of detached entities after being persisted or merged to be fully traversed after a flush

busitech:
…r being persisted or merged to be fully traversed after a flush. hibernate search 6 requires this to update indexes.

https://hibernate.atlassian.net/browse/HHH-14839


hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
```diff
@@ -677,6 +677,7 @@ else if ( maybeProxy instanceof PersistentAttributeInterceptable ) {
 			final PersistentAttributeInterceptable interceptable = (PersistentAttributeInterceptable) maybeProxy;
 			final PersistentAttributeInterceptor interceptor = interceptable.$$_hibernate_getInterceptor();
 			if ( interceptor instanceof EnhancementAsProxyLazinessInterceptor ) {
+				( (EnhancementAsProxyLazinessInterceptor) interceptor ).setSession( session );
```
yrodiere, line 680: Copying the [conversation from the original PR](https://github.com/hibernate/hibernate-orm/pull/4339#discussion_r742615080).

From @yrodiere:

> I wonder why we have to do that... I see we already have code whose purpose is to set the session after reassociation, e.g. in `org.hibernate.persister.entity.AbstractEntityPersister#afterReassociate`. But for some reason that code isn't called here...

From @busitech:

> `Persist` and `merge` have not been exposing proxy objects to the application until now, and this helped some failing test cases to pass.  One example is CascadeDeleteCollectionTest-->testDetachedOriginal:149 failing with LazyInitializationException during cascade before delete.

hibernate-core/src/main/java/org/hibernate/tuple/entity/AbstractEntityTuplizer.java
```diff
@@ -569,6 +565,20 @@ protected boolean shouldGetAllProperties(Object entity) {
...
+			if(enhancedForLazyLoading) {
+				final BytecodeLazyAttributeInterceptor interceptor = enhancementMetadata.extractLazyInterceptor(entity);
+				if (interceptor == null && lazyAttributesMetadata.isLazyAttribute( propertyName ) && getters[j].get(entity) == null) {
+					result[j] = LazyPropertyInitializer.UNFETCHED_PROPERTY;
+					continue;
+				}
+			}
+			else {
+				Type type = entityMetamodel.getPropertyTypes()[j];
+				if(type.isCollectionType() && ((CollectionType)type).getReturnedClass().isAssignableFrom(Serializable.class) && !((CollectionType)type).hasHolder() && entityMetamodel.isLazy() && getters[j].get(entity) == null) {
```
yrodiere, line 577: Copying the [conversation from the original PR](https://github.com/hibernate/hibernate-orm/pull/4339#discussion_r742617957).

From @yrodiere:

> Checking that the collection type is serializable? This looks like an awfully specific condition. Can you explain why you're restricting this code to just serializable collections?

From @busitech:

> Hibernate still supports legacy use cases.  LazyPropertyInitializer.UNFETCHED_PROPERTY gets assigned to all collection properties while they are uninitialized, and Serializable is the only interface it implements.  If this marker value cannot be assigned to a collection property, then we can't proceed with marking it unfetched.

From @busitech:

> Do you know of a better test for detecting when any of these collection types are in use, which can not be used with LazyPropertyInitializer?

hibernate-core/src/main/java/org/hibernate/type/CollectionType.java
```diff
@@ -712,7 +712,12 @@ public Object replace(
 			final Object owner,
 			final Map copyCache) throws HibernateException {
 		if ( original == null ) {
-			return null;
+			if(target instanceof PersistentCollection) {
+				return target;
+			}
+			else {
+				return null;
+			}
```
yrodiere, line 720: Copying the conversation from [the original PR](https://github.com/hibernate/hibernate-orm/pull/4339#discussion_r742621019).

From @yrodiere:

> So if someone erased a collection in a entity by setting it to `null`, and then tries to merge that entity, the collection will all of a sudden no longer be erased.
> 
> This looks like you're changing the meaning of a `null` collection from "empty" to "undefined". While I could get behind that in a new ORM implementation, doing this in Hibernate ORM will definitely break many existing application.

From @busitech:

> Your first statement in this comment is incorrect, not only because this PR is passing almost all test cases in Hibernate Core, but also because cascade processing has already occurred by this time.  
> 
> In the above method, if `target` is a `PersistentCollection` and the original value was `null`, we are currently processing a lazy loaded collection property that Hibernate Search may need to traverse.  We need to assign the uninitialized proxy value to the object.  The purpose of this code is to not overlay an uninitialized proxy collection with the originally null value, so Hibernate Search can traverse the relationship, and possibly go deeper into the hierarchical tree by traversing even beyond properties of the present entity instance.

From @busitech:

> It is an incorrect assumption that `null` always means empty, even in ORM today, and this ambiguity is the root cause of our problems.  ORM needs to change this to resolve HHH-14839 with a long term solution.  As discussed previously in the forum, this PR is a workaround pending a fundamental architecture change to ORM.
> 
> ORM is inconsistent in that it _NEVER_ returns a `null` value for a lazy loaded collection property after loading an entity, but _ALWAYS_ returns `null` collection values passed in to `em.persist` and `em.merge`, even when they should have data.  
>
> **The main effect of this PR is to standardize results returned from `em.persist` and `em.merge` so that they are more consistent with those of `em.find`.**
> 
> ORM cannot carry forward bad design forever.  At some point it must be realized that a breaking change is absolutely necessary to purge bad design.  Long term, if a `null` collection will not mean uninitialized, then there needs to be a placeholder for collections that actually means uninitialized in all cases.
> 
> It is our opinion that the meaning of `null` must be changed from a concept of "empty" to "undefined" **_when assigned to a collection property_**, and that only a collection object with length zero should be seen as an empty collection.
> 
> Note also that at the present time, the code in this PR does not affect associations which cascade in any way, to limit the complexity initially.  It might be beneficial to Hibernate Search to have all cascading collections transformed into an uninitialized proxy as well, but we hardly ever use them in our application, so we did not address this yet.
> 
> If we _were_ to apply this logic to collections that use cascade features, applications would simply see a more valid way of representing an empty collection returned instead of `null`.  Applications would see an actual collection object with length of zero, instead of what was previously `null` and ambiguous.
>
> If such an application were depending on seeing null instead of a collection with zero length, it would have to change slightly, but as I already stated, ORM _already_ provides applications with such a value for ALL lazy collection properties during `em.find`, so they are already dealing with them there...

hibernate-core/src/main/java/org/hibernate/tuple/entity/AbstractEntityTuplizer.java
```diff
@@ -569,6 +565,20 @@ protected boolean shouldGetAllProperties(Object entity) {
 			// if the attribute is not lazy (bytecode sense), we can just use the value from the instance
 			// if the attribute is lazy but has been initialized we can just use the value from the instance
 			// todo : there should be a third case here when we merge transient instances
+			if(enhancedForLazyLoading) {
+				final BytecodeLazyAttributeInterceptor interceptor = enhancementMetadata.extractLazyInterceptor(entity);
+				if (interceptor == null && lazyAttributesMetadata.isLazyAttribute( propertyName ) && getters[j].get(entity) == null) {
+					result[j] = LazyPropertyInitializer.UNFETCHED_PROPERTY;
```
yrodiere, line 571: Copying the [conversation from the original PR](https://github.com/hibernate/hibernate-orm/pull/4339#discussion_r742623535).

From @yrodiere:

> I don't understand why this is needed; the pre-existing code below states that we will consider the property unfecteched if it is lazy and not loaded yet according to enhancement metadata. That seems right to me. 
>
> On the other hand, basing your decision on whether the property is null could be completely wrong. A lazy property can be `null` and already loaded, as it can be `null` and not-loaded-yet. Nullness doesn't tell you anything.

From @busitech:

> To understand why, take a look at the `isAttributeLoaded` routine used by the pre-existing code.  > It returns true by default every time the interceptor is null, which is wrong, because all transient/detached objects start out with their interceptors `null` and lazy associations unpopulated.  
> 
> Due to this bad default, all of the `null` values which needed to be lazy loaded were being ignored and staying null.  Now they will be fetched, as they are supposed to be, after the interceptor has been assigned, which happens later.
> 
> Notice line 567 of the pre-existing code:
> `// todo : there should be a third case here when we merge transient instances`
> 
> Yep, transient instances were not being handled well here.

hibernate-core/src/main/java/org/hibernate/action/internal/EntityInsertAction.java
```diff
@@ -150,6 +153,168 @@ public void execute() throws HibernateException {
...
+						}
+					}
+				}
+				if (needsInterceptor) {
+					persister.getBytecodeEnhancementMetadata().injectEnhancedEntityAsProxyInterceptor(instance, getEntityKey(), session);
+					interceptor = interceptable.$$_hibernate_getInterceptor();
+					for (String propertyName : initializedLazyFields) {
+						interceptor.attributeInitialized(propertyName);
+					}
+				}
```
yrodiere, line 214: Copying the [conversation from the original PR](https://github.com/hibernate/hibernate-orm/pull/4339#discussion_r742641143).

From @yrodiere:

> So, here, you're assigning an interceptor to a newly intercepted entity.
> 
> Unless I'm misunderstood, all the code above is trying to isolate `null` associations, so that the interceptor will consider those associations as un-initialized and will try to initialize them the next time we access them.
> 
> Fair enough.
> 
> I think this could be improved, though. This is a lot of code to execute for each property of each newly inserted entity. Maybe it would be a good idea to build a list of "potentially uninitialized properties on persist" on bootstrap, storing it somewhere (e.g. in the entity persister), so that here you only have to go through that list and check for nulls? Then you'll consider non-null properties, as well as all other non-association properties, as initialized.

From @busitech:

> True, but most of this code does not execute if the property is not an association type:
> 
> `if( ! types[i].isAssociationType()) continue;`
> 
> Still, I definitely agree with you, a few iterations can be saved, and I thought about this as I was writing this code.  However, since I am not on the ORM team, I have chosen to avoid creating new class member variables, and changing the signature of existing methods, and just tried to work within the framework.  I wanted the ORM team to decide where and how to optimize, especially since ORM 6.0 is a major release.  I was willing to do 95% of the work to get this up and running in ORM 5.x, but the remaining 5% to fully optimize might double our investment, and then might be thrown out when we see the changes in ORM 6.0?  Ideally, we should try to interrupt the ORM team to look at this while 6.0 is unreleased so they can introduce any needed breaking changes at the major release, rather than at a point release.
> 
> Just as Search is your baby, ORM is someone else's baby, and I was very cognizant of this.  ORM is a huge framework that has been in development for more than 20 years, and after getting into it, it is really a beautiful work of art.  Looking at the code, I saw many loops which iterate over the entity property list, so my impression was that it must not cost enough in terms of performance to require optimization and an increase in memory footprint.
> 
> If we can get the larger issues agreed upon with the ORM team, then we can certainly start to buff and wax the surface.
> 
> It also could be that the ORM team sees the bigger picture right away, and wants to take a different approach entirely, such as solving the issue with a better architectural design, wherein this workaround gets thrown away.  Let's engage them ASAP.

hibernate-core/src/main/java/org/hibernate/action/internal/EntityInsertAction.java
```diff
@@ -150,6 +153,168 @@ public void execute() throws HibernateException {
...
+			if (interceptor == null) {
+				Type[] types = persister.getPropertyTypes();
+				final String[] propertyNames = persister.getPropertyNames();
+				final CascadeStyle[] cascadeStyles = persister.getPropertyCascadeStyles();
+				boolean needsInterceptor = false;
+				Set<String> initializedLazyFields = new HashSet<>();
+				for ( int i = 0; i < types.length; i++) {
+					if(!types[i].isAssociationType()) {
+						continue;
+					}
```
yrodiere, line 172: Copying the [conversation from the original PR](https://github.com/hibernate/hibernate-orm/pull/4339#discussion_r742641548).

From @yrodiere:

> This effectively considers every `*ToOne` association as uninitialized, even if it has a non-null value. That doesn't seem like a good idea?

From @busitech:

> Yes, that is essential.  Just having a value doesn't mean Hibernate Search can traverse any association properties  during indexing unless interceptors are installed.  Assume the value was set by the client, and is "dead" from a perspective of indexing.  Imagine that the present object is only a small part of a huge multidimensional object graph, and that your search indexes depend on being able to traverse both up and down the hierarchical tree, using a proxy installed at every *ToOne and *ToMany property, otherwise your index will be incomplete and corrupt!

yrodiere (commented):
Copying my intiaial review from the original PR (#4339), as well as your answers. Let's continue the discussion here.

Main comment on my initial review:

> Thanks. Here's a first review, before we involve the ORM team (which, as I said before, is pretty busy right now trying to meet deadlines related to ORM 6).

hibernate-core/src/main/java/org/hibernate/tuple/entity/AbstractEntityTuplizer.java
yrodiere: > these collection types are in use

What I was trying to ask was, what are "these" collection types?

Is this code to handle collections mapped to a single column as a serialized blob?

hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
yrodiere: Alright. I think the "ideal" solution would be to remove some of this code and somehow call `org.hibernate.persister.entity.AbstractEntityPersister#afterReassociate` instead, but I suppose that can wait.

hibernate-core/src/main/java/org/hibernate/tuple/entity/AbstractEntityTuplizer.java
```diff
@@ -40,12 +40,7 @@
 import org.hibernate.proxy.ProxyFactory;
 import org.hibernate.tuple.IdentifierProperty;
 import org.hibernate.tuple.Instantiator;
-import org.hibernate.type.AssociationType;
-import org.hibernate.type.BasicType;
-import org.hibernate.type.ComponentType;
-import org.hibernate.type.CompositeType;
-import org.hibernate.type.EntityType;
-import org.hibernate.type.Type;
+import org.hibernate.type.*;
```
yrodiere, line 43: We try to avoid wildcard imports; could you please remove them?

[This IDE setup](https://github.com/hibernate/hibernate-ide-codestyles) might be useful.

yrodiere (commented):
Some more comments and answers.

This is definitely starting to look like something that needs to be in ORM 6 rather than 5. I'll ping the ORM team.

hibernate-core/src/main/java/org/hibernate/tuple/entity/AbstractEntityTuplizer.java
busitech: No, this filter is the opposite of a very specific filter. It is a very broad, inclusive test, only omitting a very select few collection types like ArrayList on legacy test cases. This does not filter a few things in, it kicks a few things out that fail legacy test cases with IllegalArgumentException. This filter lets almost every other type of collection be handled here.

The only ones that get omitted are those to which LazyPropertyInitializer.UNFETCHED_PROPERTY cannot be assigned (and because it implements Serializable as it's only interface, so there's no other good way to test compatibility).

AbstractPersistentCollection implements Serializable, which is the superclass for almost every persistent collection type in ORM, but just to be sure not to break any other test cases, I simply used the interface for LazyPropertyInitializer.UNFETCHED_PROPERTY as the test.

hibernate-core/src/main/java/org/hibernate/type/CollectionType.java
busitech: I heard you say you could get behind it for a new ORM implementation, which I took to mean a new alternative product other than Hibernate ORM.  I thought you meant it would be an eternal dead end trying to get Hibernate ORM improved.  If you meant getting behind it for ORM 6 then that would be terrific.

Let's pursue option 1 on your end, and on this end we are going to run our build of 5 in production, immediately.  We will be able to benefit from our own build while 6 is updated as well.

hibernate-core/src/main/java/org/hibernate/tuple/entity/AbstractEntityTuplizer.java
busitech: Sure, will do.

hibernate-core/src/main/java/org/hibernate/action/internal/EntityInsertAction.java
```diff
@@ -150,6 +153,168 @@ public void execute() throws HibernateException {
...
+					else {
+						if(propertyValue instanceof PersistentAttributeInterceptable) {
+							PersistentAttributeInterceptable interceptableProperty = ((PersistentAttributeInterceptable) propertyValue );
+							PersistentAttributeInterceptor propertyInterceptor = interceptableProperty.$$_hibernate_getInterceptor();
+							if(propertyInterceptor == null) {
+								needsInterceptor = true;
+								continue;
+							}
+							else {
```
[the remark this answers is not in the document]
busitech: Isn't this the only location left where the injector is being set within our changes, here in EntityInsertAction?  I believe it is.

hibernate-core/src/main/java/org/hibernate/action/internal/EntityInsertAction.java
busitech: It actually replaces every *ToOne non-proxy with a HibernateProxy (or injects the interceptor and marks the property unfetched) so that Hibernate Search can go through that node to the next @IndexedEmbedded association that it needs to load in order to reindex.  Every non-null non-proxy like a detached value is replaced with a proxy because all objects have relationships beyond them.

busitech: The only time a property is marked initialized is if we find the interceptor already injected AND the object is initialized.  We assume this means the injector did the init, which means that the next level (up and down the hierarchical tree) can be traversed during indexing.  This scenario happens when server side code does an `em.find` and then assigns that entity to a property of another object which gets persisted.

hibernate-core/src/main/java/org/hibernate/type/CollectionType.java
busitech: I am working on porting our PR to Hibernate 6.

busitech:
This is our PR for Hibernate 6.
https://github.com/hibernate/hibernate-orm/pull/4342

hibernate-core/src/main/java/org/hibernate/action/internal/EntityInsertAction.java
```diff
@@ -150,6 +153,168 @@ public void execute() throws HibernateException {
...
+								}
+							}
+						}
+						if(types[i].isCollectionType()) {
+							if(lazyAttributesMetadata.isLazyAttribute(propertyName)) {
+								initializedLazyFields.add(propertyName);
+							}
+							continue;
+						}
+					}
```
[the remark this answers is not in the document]
busitech: This has been done.

hibernate-core/src/main/java/org/hibernate/tuple/entity/AbstractEntityTuplizer.java
busitech: I tried using the IDE setup files, but they might need updating for IntelliJ IDEA 2021.2.3.  Even after importing I had to change the thresholds manually for imports.  Nevertheless, I got the wildcards removed.

hibernate-core/src/main/java/org/hibernate/action/internal/EntityInsertAction.java
yrodiere: >  Every non-null non-proxy like a detached value is replaced with a proxy

Then that's, IMO, a different problem: you're no longer filling the gaps in associations in a newly inserted object, you're actually correcting wrong references in a graph; you're essentially doing half a merge during a persist.

Right now, applications are not supposed to assign detached values to properties. If all they have is a reference to a detached entity, they must retrieve an attached version of that entity first (e.g. through a merge, or by calling `session.getReference()` to retrieve a proxy), then assign that value to the property.

Knowing that neither a merge nor calling `session.getReference` involves a database call, this time I don't think one can argue this is not doable for performance reasons.

This new behavior you're proposing *could* make sense, but it seems like a different problem you're trying to fix, and I don't think you're fixing it completely. For example, detached entities could be assigned to already-persisted entities, for an update. When will *those* detached entities be replaced with attached proxies?

It also raises the question of cascades, which you saw and sidestepped by not applying this change on associations with any kind of cascade. But not planning for that could result in inconsistent behavior. We need to at least have an idea of the desired behavior when cascades are enabled, before we apply such a change.

If you want to discuss this further, I think the best way to actually make progress would be to talk to the Hibernate ORM team directly, rather than me: https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/

yrodiere (changes requested):
So, I was able to discuss this changeset with the ORM team. You can find the whole discussion here: https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/topic/Intializing.20associations.20in.20persisted.2Fmerged.20entities

What we discussed was the new behavior you're introducing, which makes sure the entity resulting from a `persist`/`merge` has lazy-loaded associations, with all associations that were `null` in the input (of `persist`/`merge`) being considered un-initialized.

The conclusions below. If you want to challenge any of these, please do it in the Zulip thread I linked above: you'll have to convince Steve Ebersole and Gavin King, more than you'll have to convince me.

## Association ownership

This new behavior should only be implemented for the non-owning side of associations.

Hibernate ORM should not try to "fix" the owning side of an association if it's `null`, because that's a user error. The owning side defines the content of the association and thus defines an empty association for a ToOne association, and simply should not be `null` for a ToMany association.

In terms of changes to your patch, this probably means adding a test, and then adapting the changeset as necessary to make sure non-owning sides of associations are never initialized by Hibernate ORM.

## Orphan removal

When the association is marked with `ophanRemoval = true`, and a merge happens with that association being set to `null` in the input entity, then:

1. The resulting object should have its association initialized, and empty (`null` for a ToOne association, empty collection for a ToMany association)
2.  The orphan removal should still happen, as required by the JPA spec.

In terms of changes to your patch, this probably means adding a test, and then adapting the changeset as necessary to make sure orphan removal happens as expected.

## Bi-directional association management

This whole behavior essentially amounts to requiring Hibernate ORM to manage bi-directional associations. The only difference with [bytecode-enhanced bidirectional association management](https://docs.jboss.org/hibernate/stable/orm/userguide/html_single/Hibernate_User_Guide.html#BytecodeEnhancement-dirty-tracking-bidirectional) is that the "management" happens on persist/merge, rather than when calling a setter.

An application with bidirectional association management essentially uses a different programming model than one with this management disabled, with different benefits and constraints. As such, it's important to keep the two programming models strictly separated.

For that reason, the new behavior should only be enabled when bidirectional association management is enabled.

In terms of changes to your patch, this probably means:

1. Adding an `isEnhancedForBidirectionalAssociationManagement()` method to `org.hibernate.bytecode.spi.BytecodeEnhancementMetadata`
2. Calling this method everywhere you made a change in order to decide whether to enable the new behavior.
3. Implementing this method in `org.hibernate.tuple.entity.BytecodeEnhancementMetadataPojoImpl`. We will probably need to rely on a similar trick to the one used for `enhancedForLazyLoading`, i.e. checking that the mapped class is assignable to some interface. See `org.hibernate.tuple.entity.BytecodeEnhancementMetadataPojoImpl#from`. That interface may not exist yet, in which case we'll need to add it. For what it's worth, bidirectional association management is done in `org.hibernate.bytecode.enhance.internal.bytebuddy.BiDirectionalAssociationHandler`, so if we need to add an implemented interface to enhanced entities, it will have to be there. But only if bidirectional association management is enabled.

PR #4341, opened 2021-11-04
Upgrades spatial support for MariaDB, MySQL and H2GIS

maesenka:
Issues HHH-14906, HHH-14804, and HHH-14803


hibernate-spatial/src/test/java/org/hibernate/spatial/testing/dialects/postgis/PostgisTestSupport.java
```diff
@@ -36,7 +36,7 @@ public NativeSQLTemplates templates() {
 	}
 
 	@Override
-	public PredicateRegexes predicateRegexes(){ return new PostgisPredicateRegexes();}
+	public PredicateRegexes predicateRegexes(){ return new PredicateRegexes("st_geomfromtext");}
```
beikov, line 39: Shouldn't it be this?

```suggestion
	public PredicateRegexes predicateRegexes(){ return new PredicateRegexes("st_geomfromewkt");}
```

beikov (changes requested):
Looking ok mostly, but I think we should first wait for the fixes from Steve for the type resolving and replace the BasicType implementations.

hibernate-spatial/src/main/java/org/hibernate/spatial/GeometryLiteralFormatter.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.geolatte.geom.Geometry;
+import org.geolatte.geom.codec.Wkt;
+import org.geolatte.geom.jts.JTS;
+
+public class GeometryLiteralFormatter<T> implements JdbcLiteralFormatter<T> {
+
+	private final JavaType<T> javaType;
+	private final Wkt.Dialect wktDialect;
+	private final String geomFromTextName;
```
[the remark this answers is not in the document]
maesenka: Fixed

hibernate-spatial/src/main/java/org/hibernate/spatial/GeometryLiteralFormatter.java
```diff
@@ -0,0 +1,54 @@
...
+	@Override
+	public void appendJdbcLiteral(
+			SqlAppender appender, T value, Dialect dialect, WrapperOptions wrapperOptions) {
+		Geometry<?> geom;
+		if ( javaType instanceof GeolatteGeometryJavaTypeDescriptor ) {
+			geom = (Geometry<?>) value;
+		}
+		else {
+			geom = jts2Gl( value );
+		}
```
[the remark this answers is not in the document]
maesenka: Yes much better.

hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/mariadb/MariaDB103SpatialDialect.java
```diff
@@ -10,6 +10,7 @@
 import org.hibernate.dialect.MariaDB103Dialect;
 import org.hibernate.spatial.SpatialDialect;
 
+@Deprecated
 public class MariaDB103SpatialDialect extends MariaDB103Dialect implements SpatialDialect {
```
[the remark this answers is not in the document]
maesenka: For backward compatibility reasons. HS users will just get deprecation warning rather than a compile error. I added a comment though

hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/mysql/MySqlSqmFunctionDescriptors.java
```diff
@@ -0,0 +1,37 @@
...
+import org.hibernate.spatial.CommonSpatialFunction;
+
+public class MySqlSqmFunctionDescriptors extends BaseSqmFunctionDescriptors {
+
+	final static private List<CommonSpatialFunction> unsupported = new ArrayList<>();
+
+	static {
+		unsupported.add( CommonSpatialFunction.ST_BOUNDARY );
+		unsupported.add( CommonSpatialFunction.ST_RELATE );
+	}
```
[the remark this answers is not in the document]
maesenka: Done

hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/postgis/PGGeometryType.java
```diff
@@ -52,22 +53,7 @@
...
-		return (appender, value, dialect, wrapperOptions) -> {
-			appender.appendSql( "ST_GeomFromEWKT('" );
-			appender.appendSql( jts2Gl( value ).toString() );
-			appender.appendSql( "')" );
-		};
-	}
-
-	private <T> Geometry<?> jts2Gl(T value) {
-		return JTS.from( (org.locationtech.jts.geom.Geometry) value );
+		return new GeometryLiteralFormatter<T>( javaTypeDescriptor, Wkt.Dialect.SFA_1_1_0, "ST_GeomFromText" );
```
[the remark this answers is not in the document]
maesenka: You're right. It's equivalent for 2D, but not for 3D. Will change it back

hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/mysql/MySQLGeometryType.java
```diff
@@ -28,6 +30,7 @@
 import org.geolatte.geom.codec.Wkb;
 import org.geolatte.geom.codec.WkbDecoder;
 import org.geolatte.geom.codec.WkbEncoder;
+import org.geolatte.geom.codec.Wkt;
 
 /**
  * Descriptor for MySQL Geometries.
```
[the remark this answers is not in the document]
maesenka: That's a good point. Of course at some point MariaDB could choose a different storage format for geometries, but for now it's the same for MySQL and MariaDB. Will remove the MariaDB-related class.

hibernate-spatial/src/main/java/org/hibernate/spatial/GeometryLiteralFormatter.java
```diff
@@ -0,0 +1,48 @@
...
+		appender.appendSql( geomFromTextName );
+		appender.appendSql( Wkt.toWkt( geom, wktDialect ) );
+		appender.appendSql( "'," );
+		appender.appendSql( ( Math.max( geom.getSRID(), 0 ) ) );
+		appender.appendSql( ")" );
+	}
+
+	private Geometry<?> jts2Gl(T value) {
+		return JTS.from( (org.locationtech.jts.geom.Geometry) value );
+	}
```
beikov, line 46: Unused?

hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/postgis/PostgisDialectContributor.java
```diff
@@ -26,8 +28,11 @@ public PostgisDialectContributor(ServiceRegistry serviceRegistry) {
 
 	public void contributeTypes(TypeContributions typeContributions) {
 		HSMessageLogger.LOGGER.typeContributions( this.getClass().getCanonicalName() );
-		typeContributions.contributeType( new GeolatteGeometryType( PGGeometryType.INSTANCE_WKB_2 ) );
-		typeContributions.contributeType( new JTSGeometryType( PGGeometryType.INSTANCE_WKB_2 ) );
+		typeContributions.contributeType( new GeolatteGeometryType( PGGeometryJdbcType.INSTANCE_WKB_2 ) );
+		typeContributions.contributeType( new JTSGeometryType( PGGeometryJdbcType.INSTANCE_WKB_2 ) );
```
beikov, line 32: Shouldn't these BasicType implementations be obsolete now?

hibernate-spatial/hibernate-spatial.gradle
```diff
@@ -14,34 +12,41 @@ apply plugin: 'org.hibernate.matrix-test'
...
 tasks.test {
-    enabled =  ['pgsql', 'cockroachdb'].contains( project.db )
+	// for now we cannot run with 'h2' project.db due to compatability issues H2 requires 4.200 but H2GIS 1.5 needs 4.197
+	enabled = ['pgsql',
+//			   'h2', -- for now we cannot run with 'h2'due to compatability issues we needs H2 at 4.200 but H2GIS only works with 4.197
+//                'pgsql_ci', -- disable running in CI until typing issues are resolved
+			   'cockroachdb',
+			   'mariadb',
+//                'mysql_ci',
+			   'mysql_docker'].contains( project.db )
```
beikov, line 49: IMO you should enable this so we can see the tests running :)

beikov (changes requested):

documentation/src/main/asciidoc/userguide/chapters/query/spatial/Spatial.adoc
```diff
@@ -199,11 +199,11 @@ The dialect `CockroachDB` support the `GEOMETRY` type in CockroachDB v20.2 and l
...
 ====
 
-GeoDB (H2)::
-The `GeoDBDialect` supports the GeoDB a spatial extension of the H2 in-memory database.
+H2GIS::
+The `H2DBDialect` supports the H2GIS, a spatial extension of the H2 in-memory database.
 [NOTE]
 ====
-The dialect has been tested with GeoDB version 0.7
+The dialect has been tested with H2GIS version 1.5
```
NathanQingyangXu, line 206: I would suggest mentioning h2gis not supporting GEOGRAPHY type as well. It used to take one whole day to discover this for me when investigating switching from geodb to h2gis. Other content might need to be updated in this chapter, like the dialect function table.

documentation/src/main/asciidoc/userguide/chapters/query/spatial/Spatial.adoc
```diff
@@ -88,135 +71,99 @@ relevant section.
...
 The dialect does not support DB2 for z/OS or DB2 column-oriented databases.
-[NOTE]
-====
-
++
 In order to use the DB2 Hibernate Spatial capabilities, it is necessary to first execute the following
 SQL statements which will allow DB2 to accept Extended WellKnown Text (EWKT) data and return EWKT data.
 One way to do this is to copy these statements into a file such as ewkt.sql and execute it in a DB2 command window
-with a command like 'db2 -tvf ewkt.sql'.
+with a command sucha as `db2 -tvf ewkt.sql`.
```
beikov, line 166:
```suggestion
with a command such as `db2 -tvf ewkt.sql`.
```

documentation/src/main/asciidoc/userguide/chapters/query/spatial/Spatial.adoc
```diff
@@ -88,135 +71,99 @@ relevant section.
...
+The dialect `CockroachDB` support the `GEOMETRY` type in CockroachDB v20.2 and later. The `GEOGRAPHY` type is currently not supported.
 
-GeoDB (H2)::
-The `GeoDBDialect` supports the GeoDB a spatial extension of the H2 in-memory database.
-[NOTE]
-====
-The dialect has been tested with GeoDB version 0.7
-====
+H2GIS::
+The `H2DBDialect` supports the H2GIS, a spatial extension of the H2 in-memory database. This dialect can be used as a replacement for the `GeoDB` dialect that was supported in previous versions. The major difference with `GeoDB` is that the  `GEOGRAPHY` column type is not present in `H2GIS`.
```
beikov, line 156:
```suggestion
The `H2Dialect` supports H2GIS, a spatial extension of the H2 in-memory database. This dialect can be used as a replacement for the `GeoDB` dialect that was supported in previous versions. The major difference with `GeoDB` is that the  `GEOGRAPHY` column type is not present in `H2GIS`.
```

documentation/src/main/asciidoc/userguide/chapters/query/spatial/Spatial.adoc
```diff
@@ -88,135 +71,99 @@ relevant section.
...
-====
 
 CockroachDB::
-The dialect `CockroachDB` support the `GEOMETRY` type in CockroachDB v20.2 and later.
-
-[NOTE]
-====
-The `GEOGRAPHY` type is not currently supported.
-====
+The dialect `CockroachDB` support the `GEOMETRY` type in CockroachDB v20.2 and later. The `GEOGRAPHY` type is currently not supported.
```
beikov, line 153:
```suggestion
The dialect `CockroachDialect` support the `GEOMETRY` type in CockroachDB v20.2 and later. The `GEOGRAPHY` type is currently not supported.
```

documentation/src/main/asciidoc/userguide/chapters/query/spatial/Spatial.adoc
```diff
@@ -88,135 +71,99 @@ relevant section.
...
-
++
 In may cases, this strategy will suffice.
 If not, you can provide your own implementation of this interface on the classpath, and configure it in the `hibernate.spatial.connection_finder` property.
 Note that implementations must be thread-safe and have a default no-args constructor.
-====
-
 SQL Server::
-The dialect `SqlServer2008Dialect` supports the `GEOMETRY` type in SQL Server 2008 and later.
+Currently, The `GEOGRAPHY` type is not currently supported.
```
beikov, line 149:
```suggestion
Currently, The `GEOGRAPHY` type is not supported.
```

hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/postgis/PostgisDialectContributor.java
```diff
@@ -26,8 +28,11 @@ public PostgisDialectContributor(ServiceRegistry serviceRegistry) {
 
 	public void contributeTypes(TypeContributions typeContributions) {
 		HSMessageLogger.LOGGER.typeContributions( this.getClass().getCanonicalName() );
-		typeContributions.contributeType( new GeolatteGeometryType( PGGeometryType.INSTANCE_WKB_2 ) );
-		typeContributions.contributeType( new JTSGeometryType( PGGeometryType.INSTANCE_WKB_2 ) );
+//		typeContributions.contributeType( new GeolatteGeometryType( PGGeometryJdbcType.INSTANCE_WKB_2 ) );
+//		typeContributions.contributeType( new JTSGeometryType( PGGeometryJdbcType.INSTANCE_WKB_2 ) );
```
beikov, line 32: Please remove the commented out blocks, and ideally, also remove the BasicType implementations.

hibernate-spatial/hibernate-spatial.gradle
```diff
@@ -14,34 +12,42 @@ apply plugin: 'org.hibernate.matrix-test'
...
 sourceSets.test.resources {
-    setSrcDirs(['src/test/java', 'src/test/resources'])
+	setSrcDirs( ['src/test/java', 'src/test/resources'] )
 }
 
 tasks.test {
-    enabled =  ['pgsql', 'cockroachdb'].contains( project.db )
+	// for now we cannot run with 'h2' project.db due to compatability issues H2 requires 4.200 but H2GIS 1.5 needs 4.197
+	enabled = ['pgsql',
+			   'h2',
```
beikov, line 44: Either the comment above is "wrong" or you will have to comment this out.

documentation/src/main/asciidoc/userguide/chapters/query/spatial/Spatial.adoc
NathanQingyangXu: Also, 'support' changed to 'supports'?

documentation/src/main/asciidoc/userguide/chapters/query/spatial/Spatial.adoc
NathanQingyangXu: I think the missing support of geometry type in h2gis might be temporary, see https://github.com/orbisgis/h2gis/issues/864. So we might better tweak verbiage of 'not present' to 'not supported currently'. Otherwise it implies that the decision is by design and be cast in stone.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmCriteriaNodeBuilder.java
```diff
@@ -1451,6 +1451,25 @@ private static TrimSpec convertTrimSpec(Trimspec jpaTs) {
...
+		basicType = getTypeConfiguration().getBasicTypeForJavaType( valueClass );
+		if ( basicType == null ) {
+			final JavaType<Object> javaType = getTypeConfiguration().getJavaTypeDescriptorRegistry()
+					.getDescriptor( valueClass );
+			if ( javaType != null ) {
+				final JdbcType recommendedJdbcType = javaType.getRecommendedJdbcType( getTypeConfiguration().getCurrentBaseSqlTypeIndicators() );
+				if ( recommendedJdbcType != null ) {
+					basicType = (BasicType<T>) getTypeConfiguration().getBasicTypeRegistry().resolve( javaType, recommendedJdbcType);
+				}
+			}
```
beikov, line 1467: I think that's not totally correct as you can see from the test failures for e.g. PostgreSQL, because `JavaTypeRegistry#getDescriptor` creates a `SerializableJavaTypeDescriptor` or `JavaTypeDescriptorBasicAdaptor` on-demand if there it can't find an existing JavaType for the class.
IMO we should introduce a `JavaTypeRegistry#findDescriptor` that returns null if the descriptor does not exist rather than trying to create a new one.

beikov (commented):
I think the H2 test failures are due to H2 only supporting the interval types as of 1.4.198, so we will have to add a conditional registration of the JdbcTypes and DDL types, but I can do that. I'll let you know when I'm done with this so you can rebase.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmCriteriaNodeBuilder.java
beikov: I merged my PR for this, you can rebase and then use `findDescriptor` here instead: https://github.com/hibernate/hibernate-orm/pull/4359

maesenka: Could you have a look at the changes Steve did to in commit 0b4696609 to AbstractJdbcParameter? Since I based my code on his there, maybe it needs changing also?

documentation/src/main/asciidoc/userguide/chapters/query/spatial/Spatial.adoc
```diff
@@ -250,19 +197,12 @@ create transform for db2gse.st_geometry db2_program (
...
 [[spatial-types]]
 === Types
 
 Hibernate Spatial comes with the following types:
 
-jts_geometry::
-	Handled by `org.hibernate.spatial.JTSGeometryType`, it maps a database geometry column type to a `org.locationtech.jts.geom.Geometry` entity property type.
-geolatte_geometry::
-	Handled by `org.hibernate.spatial.GeolatteGeometryType`, it maps a database geometry column type to an `org.geolatte.geom.Geometry` entity property type.
+TODO
```
beikov, line 205: I guess you will retain the same names for something like `SpatialTypes` analogously to `StandardBasicTypes`?

beikov (approved):

documentation/src/main/asciidoc/userguide/chapters/query/spatial/Spatial.adoc
maesenka: I'm not yet sure how to do it, but this whole section needs to be rewritten.

ctabin:
Hi and thanks for this huge PR !
Since GeoDB has been removed and H2GIS added altogether, is there a way to do the migration smoothly ? Typically, it would be great if the H2GIS support was available in Hibernate 5.6.x, so we can avoid the big-bang changes during transition to Hibernate 6.x. Another way would be to keep the GeoDB support in Hibernate 6.x at least temporarily so it would be possible to migrate step-by-step while both versions are available at the same time.
Thanks & best regards.

NathanQingyangXu:
Just a reminder that h2gis feature set is not superset of geodb. E.g., if
you are keen on the usage of Geography data type (for instance,
latitude/longitude and proximity search), you have to stick to geodb.

ctabin:
Hi @NathanQingyangXu and thanks for your response.
Actually, we only use basic `LINESTRING` columns with `INSERSECTS` predicates, so there is no blocker for this migration.
However, how anyone would be able to stick with Hibernate/GeoDB since the latter as been removed by this PR ? :thinking:

NathanQingyangXu:
Good question. Maybe we'd better make both coexist.

ctabin:
> Maybe we'd better make both coexist.

That would be great.
We are using H2 1.4.199 (and also tested 2.0.204) with Geodb and that works fine (even if Geodb is not maintained anymore).
However, we are stuck to upgrade to 2.0.204 for now because of #4524.

beikov:
I'd like to make some things clear here as I think people might have the intention that H2 is a "supported" a.k.a. the Hibernate team guarantees to do maintenance for it. It is not.

It's the Hibernate teams choice to maintain the dialect on a best effort basis because we like testing with the database as it doesn't require a setup and is reasonably fast. Having said that, Geodb is not maintained anymore as you figured and we usually don't want to put effort into stuff that receives no support/maintenance as that usually means people will move away from it. H2GIS seems to be the way forward which we added support for. So unless someone from the community steps up to do the work, Geodb support won't happen.

PR #4343, opened 2021-11-08
HHH-14918 Key-to-one to id-class entity with key-to-one doesn't work anymore

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14918

The most obvious solution seems to fix the problem; let's see if it doesn't break another test...


beikov (approved):
Looks great. If tests pass, this can IMO be merged!

yrodiere:
The failure seems caused by a flaky test. Ignoring and merging.

PR #4345, opened 2021-11-09
HHH-14211 Switch to using oid for CLOB in PostgreSQL to avoid losing data after vacuumlo

beikov:
https://hibernate.atlassian.net/browse/HHH-14211

Also see the discussion here:

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java
@@ -114,7 +114,7 @@ public class PostgreSQL81Dialect extends Dialect {
 		registerColumnType( Types.BINARY, "bytea" );
 		registerColumnType( Types.LONGVARCHAR, "text" );
 		registerColumnType( Types.LONGVARBINARY, "bytea" );
-		registerColumnType( Types.CLOB, "text" );
+		registerColumnType( Types.CLOB, "oid" );
 		registerColumnType( Types.BLOB, "oid" );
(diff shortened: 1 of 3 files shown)


sebersole:
Do not merge this.  I want to look through this one.  See discussion at https://hibernate.zulipchat.com/#narrow/stream/132096-hibernate-user/topic/TOAST.20text.20column.20content.20in.20pg_largeobjects.20in.20pgsql

gavinking:
OK, interesting, so it looks like the issue here was actually the precise opposite of what was reported.

That is, the problem was actually that we used the wrong column type (by default) for things that really are CLOBs and belong in `pg_large_objects`, and that we never realized because the broken-ass postgres drivers tolerate this and work fine with it and make it seem totally legit, right up until your DBA runs some command line tool.

And unfortunately users were reporting this issue as the exact opposite problem: that when they use @Lob they wanted things TOASTed, not stored as oids, which just didn't seem to make any sense at all. Because I guess they were abusing @Lob in order to change the SQL column type to `text`.


Thanks for untangling this @beikov.

beikov:
Exactly @gavinking, you pretty much nailed the gist of this :)

jannic:
> And unfortunately users were reporting this issue as the exact opposite problem: that when they use @lob they wanted things TOASTed, not stored as oids, which just didn't seem to make any sense at all. Because I guess they were abusing @lob in order to change the SQL column type to `text`.

I guess this is because of misleading tutorials like https://www.baeldung.com/jpa-annotation-postgresql-text-type

jannic:
Ping to @Baeldung, perhaps you'd like to comment? If the analysis in this ticket is correct (and as far as I know, it is), the mentioned tutorial should be updated.

PR #4346, opened 2021-11-09
DBZ-14540 Don't share session-scoped interceptors with temp session

Naros:
https://hibernate.atlassian.net/browse/HHH-14540

This needs to be backported to Hibernate 5.3.


Sanne:
hi @Naros , sorry missed the notification. Do you still need this, why is it closed?

PR #4347, opened 2021-11-09
HHH-14540 Don't share session-scoped interceptors with temp session

Naros:
https://hibernate.atlassian.net/browse/HHH-14540

This needs to be backported to Hibernate 5.3.


Sanne (approved):
Thanks @Naros ! 

I've made two super simple changes, then merged:
 - added a backwards compatible (and deprecated) copy of method `configuredInterceptor` so to maintain API drop-in compatibility
 - changed the Debezium copyright to the Hibernate copyright :)

Will now backport these as suggested.

Sanne:
merged already

PR #4348, opened 2021-11-09
HHH-14922 - Inconsistent precedence of "default" versus "implicit" catalog and schema handling

sebersole:
HHH-14922 - Inconsistent precedence of orm.xml implicit catalog/schem a over "default_catalog"/"default_schema"

- proposal based on premise that `hibernate.default_catalog` and `hibernate.default_schema` take precedence over values specified as defaults in JPA and HBM mappings.  See `ImplicitNamespacePrecedenceTests` for assertions

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/relational/Database.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/relational/Database.java
@@ -25,7 +25,13 @@ import org.hibernate.service.ServiceRegistry;
 import org.hibernate.type.spi.TypeConfiguration;
 
 /**
- * @author Steve Ebersole
+ * Keeps track of boot-time descriptors for various database namespaces and objects.
+ *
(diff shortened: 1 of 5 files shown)


yrodiere:
Wow, I didn't expect a PR this fast. In fact I didn't expect one at all: I had assigned the ticket to myself because I was already working on this section of the code (for [HHH-14921](https://hibernate.atlassian.net/browse/HHH-14921)) and had already extensive tests ready; I only needed to decide on which behavior was desired...

Also, I was working on ORM 5.x :/

I'll see if I can take inspiration from this patch to fix this in 5. Unfortunately, the fix will probably have to be different after HHH-14921, since the default schema is applied later.

yrodiere:
Hmm, after this patch the default_schema seems to override any explicitly defined schema (e.g. `@Table(schema = "foo")`).

I'll work on another solution on top of my patch for HHH-14921; it will probably have to be different anyway, unfortunately.

yrodiere:
Superseded by #4351

PR #4352, opened 2021-11-10
Docs: Refer correct entity in "Composite identifiers with associations" section

maciejwalkowiak:
Code sample uses `Book` entity instead of `PersonAddress`.


beikov:
Thanks!

PR #4353, opened 2021-11-11
The JPA TCK tests expect very specific things regarding a negated predicate node

sebersole:
The JPA TCK tests expect very specific things regarding a negated predicate node

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/AbstractNegatableSqmPredicate.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/AbstractNegatableSqmPredicate.java
@@ -33,10 +33,16 @@ public abstract class AbstractNegatableSqmPredicate extends AbstractSqmPredicate
 		this.negated = !this.negated;
 	}
 
+	protected abstract SqmNegatablePredicate createNegatedNode();
+
 	@Override
 	public SqmNegatablePredicate not() {
 		// in certain cases JPA required that this always return
-		// a new instance.  we may need to allow for that here (compliance?)
+		// a new instance.
+		if ( nodeBuilder().isJpaQueryComplianceEnabled() ) {
+			return createNegatedNode();
+		}
+
 		negate();
 		return this;
 	}
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmComparisonPredicate.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmComparisonPredicate.java
@@ -40,6 +40,13 @@ public class SqmComparisonPredicate extends AbstractNegatableSqmPredicate {
 		rightHandExpression.applyInferableType( expressableType );
 	}
 
+	private SqmComparisonPredicate(SqmComparisonPredicate affirmativeForm) {
+		super( true, affirmativeForm.nodeBuilder() );
+		this.leftHandExpression = affirmativeForm.leftHandExpression;
+		this.rightHandExpression = affirmativeForm.rightHandExpression;
+		this.operator = affirmativeForm.operator;
+	}
+
 	public SqmExpression<?> getLeftHandExpression() {
 		return leftHandExpression;
 	}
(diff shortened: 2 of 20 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmComparisonPredicate.java
```diff
@@ -40,6 +40,13 @@ public SqmComparisonPredicate(
 		rightHandExpression.applyInferableType( expressableType );
 	}
 
+	private SqmComparisonPredicate(SqmComparisonPredicate affirmativeForm) {
+		super( true, affirmativeForm.nodeBuilder() );
+		this.leftHandExpression = affirmativeForm.leftHandExpression;
+		this.rightHandExpression = affirmativeForm.rightHandExpression;
+		this.operator = affirmativeForm.operator;
+	}
```
beikov, line 48: Shouldn't this rather be?

```suggestion
	private SqmComparisonPredicate(SqmComparisonPredicate affirmativeForm) {
		super( false, affirmativeForm.nodeBuilder() );
		this.leftHandExpression = affirmativeForm.leftHandExpression;
		this.rightHandExpression = affirmativeForm.rightHandExpression;
		this.operator = affirmativeForm.operator.negated();
	}
```

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/AbstractNegatableSqmPredicate.java
```diff
@@ -33,10 +33,16 @@ public void negate() {
...
 
+	protected abstract SqmNegatablePredicate createNegatedNode();
+
 	@Override
 	public SqmNegatablePredicate not() {
 		// in certain cases JPA required that this always return
-		// a new instance.  we may need to allow for that here (compliance?)
+		// a new instance.
+		if ( nodeBuilder().isJpaQueryComplianceEnabled() ) {
+			return createNegatedNode();
```
beikov, line 43: IMO this should be the default. How else would you spell something like this if we alter the original predicate negation flag on a `not` call?

`predicate OR not(predicate) AND ...`

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmComparisonPredicate.java
sebersole: No.  The problem comes from the `#isNegated` call.  Your form reports that "incorrectly".  I quote "incorrectly" because I personally think the TCK very much oversteps the spec for some of these checks.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/AbstractNegatableSqmPredicate.java
sebersole: Possibly.  See above.  If it were up to me, and we wanted to create new nodes for each negation, I'd simply wrap them all in `SqmNegatedPredicate`.  But like above, that messed with `#isNegated`

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmComparisonPredicate.java
beikov: In that case, shouldn't this be the solution then?

Shouldn't this rather be?

```suggestion
	private SqmComparisonPredicate(SqmComparisonPredicate affirmativeForm) {
		super( !affirmativeForm.isNegated(), affirmativeForm.nodeBuilder() );
		this.leftHandExpression = affirmativeForm.leftHandExpression;
		this.rightHandExpression = affirmativeForm.rightHandExpression;
		this.operator = affirmativeForm.operator;
	}
```

sebersole: The new node is always the negated.  `not(not(true)`.  Each `not` in-and-of-itself is negated form of the wrapped node

PR #4351, opened 2021-11-10
HHH-14921 + HHH-14922 Definition of the default catalog/schema on session factory creation

yrodiere:
* [HHH-14922](https://hibernate.atlassian.net/browse/HHH-14922): Inconsistent precedence of orm.xml implicit catalog/schema over "default_catalog"/"default_schema"
* [HHH-14921](https://hibernate.atlassian.net/browse/HHH-14921): Definition of the default catalog/schema on session factory creation

This should allow us to change the catalog/schema in "runtime" settings in Quarkus: https://github.com/quarkusio/quarkus/issues/19660

I am currently working on porting this to the 6.0 branch. It's going to take some time, considering the size of this patch.

PR #4354, opened 2021-11-11
HHH-14922 - Inconsistent precedence of orm.xml implicit catalog/schema over "default_catalog"/"default_schema"

sebersole:
HHH-14922 - Inconsistent precedence of orm.xml implicit catalog/schema over "default_catalog"/"default_schema"

### DIFF
--- a/gradle/java-module.gradle
+++ b/gradle/java-module.gradle
@@ -66,7 +66,17 @@ dependencies {
 	annotationProcessor( libraries.logging )
 	annotationProcessor( libraries.logging_annotations )
 
-	testCompile( libraries.junit )
+
+	// JUnit dependencies made up of:
+	// 		* JUnit 5
+	// 		* the Jupiter engine which runs JUnit 5 based tests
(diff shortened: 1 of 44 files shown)


sebersole:
The failure here is related to hibernate-core-jakarta.  No idea why the test only fails for Oracle and only in the -jakarta form

beikov (approved):
Looks good to me

yrodiere (changes requested):
As explained on https://github.com/hibernate/hibernate-orm/pull/4351, if we address HHH-14921, we'll need a completely different solution for HHH-14922, unfortunately.

Let's discuss this on Zulip before merging anything? Please?

yrodiere:
Since we merged #4371, I don't think this patch is necessary anymore, so I'm going to close this PR.

Some of the improvements are probably still useful, but that'll need some work, so I'll let Steve get back to it when there's no longer higher priority stuff to be done (that should take a while...).

PR #4358, opened 2021-11-11
Some pre-work for embeddable instantiator work.

sebersole:
initial work on new `EmbeddableInstantiator` contract allowing ctor-injection.  

the main work in this initial commit is just cleaning up where/when the `EmbeddableRepresentationStrategy`, `EmbeddableInstantiator`, etc are created/determined.  was especially important to drop those from the JPA model.

next steps include integrating my work on "cleaning up read paths"

p.s. the test is just for debugging :)


hibernate-core/src/test/java/org/hibernate/orm/test/intg/sqm/HqlTranslationNoFactoryTests.java
```diff
@@ -1,150 +0,0 @@
...
- * Tests making sure that HQL queries can fully be translated without a SessionFactory.
- *
- * todo (6.0) : have a way to directly load these translations into the {@link NamedObjectRepository}
- * 		directly.  For example Quarkus could store its build-time translations and directly here during boot-strap of the SF
- *
- * @author Steve Ebersole
- */
-@ServiceRegistry
-@DomainModel( standardModels = StandardDomainModel.RETAIL )
-public class HqlTranslationNoFactoryTests {
```
beikov, line 52: Was the deletion of this file intentional?

beikov (commented):
Nice work! Looks like you found a way around the need to re-implement the JPA metamodel building based on the mapping model. IMO this looks good and we can merge this or build on top of it in case you want to validate the approach with the real injection first. Wdyt?

hibernate-core/src/test/java/org/hibernate/orm/test/intg/sqm/HqlTranslationNoFactoryTests.java
sebersole: Yeah.  I added it to try to make sure that you could build the HQL translator without reference to a SessionFactory.  The intention was for Quarkus but they do not need this functionality

sebersole:
> Looks like you found a way around the need to re-implement the JPA metamodel building based on the mapping model. 

Its a little hacky, but its better to do minimal changes for now.  I do think it would be better longer term to change this to be based on the mapping model, but that can wait.

> merge this or build on top of it

I have some local changes already based on top of this.  The plan was to get those working first.  Those changes are to actually pass the values to the instantiator, which meant changes to the initializer which is still a work in progress

sebersole:
I went ahead and pushed this.  The work I was doing locally was not working.  I'll come back to this next week.

PR #4363, opened 2021-11-13
HHH-14926 fix ascii error in 'test-case-guide.adoc'

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14926

The first char in `(C)complete` was rendered as copyright symbol (&copy;) as below. Need to escape as https://docs.asciidoctor.org/asciidoc/latest/subs/prevent/#escape-with-backslashes.

38e44d42-9dc1-4bb3-919e-7059002f2bde
@


beikov:
Thanks!

PR #4367, opened 2021-11-15
Query entity with a one to one mapped by to embedded association

fax4ever:


hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -2728,6 +2729,49 @@ protected boolean hasWhere() {
 		return sqlWhereString != null;
 	}
 
+	private ModelPart findNestedSubPart(String attributeName) {
```
beikov, line 2732: I introduced a new method `findByPath` which you should be able to use instead of this.

fax4ever: Nice! Fixed

beikov (approved):

fax4ever:
Merged, thanks!

PR #4368, opened 2021-11-15
Wip/6.0 additional tests 

dreab8:


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntityDelayedFetchInitializer.java
```diff
@@ -121,16 +117,23 @@ public void resolveInstance(RowProcessingState rowProcessingState) {
...
-
-							// If the entity was not in the Persistence Context, but was found now,
-							// add it to the Persistence Context
-							if ( entityInstance != null ) {
-								persistenceContext.addEntity( euk, entityInstance );
+							if ( ( (AbstractEntityInitializer) parentAccess ).getEntityDescriptor()
+									.getBytecodeEnhancementMetadata()
+									.isEnhancedForLazyLoading() ) {
+								entityInstance = persistenceContext.proxyFor( entityInstance );
+							}
```
beikov, line 124: This doesn't make much sense as `entityInstance` is always null at this point. Maybe you left this around while trying out some stuff?

dreab8: right, my bad it should be
```
if ( ((AbstractEntityInitializer) this.parentAccess).getEntityDescriptor()
									.getBytecodeEnhancementMetadata()
									.isEnhancedForLazyLoading()  ) {
								return;
							}
```

beikov (approved):

beikov (approved):

PR #4372, opened 2021-11-18
Wip/6.0 proxy

dreab8:


hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -730,7 +732,14 @@ private void applyOrdering(TableGroup tableGroup, OrderByFragment orderByFragmen
 				fetchablePath = fetchParent.resolveNavigablePath( fetchable );
 			}
 
-			FetchTiming fetchTiming = fetchable.getMappedFetchOptions().getTiming();
+			FetchTiming fetchTiming;
+			if ( fetchParent instanceof EmbeddableFetchImpl && fetchable instanceof BasicAttributeMapping ) {
```
beikov, line 736: Shouldn't we change the fetch timing in the mapping model already if this really is something that we can't support? I guess there are other parts that rely on the default timing as well that would otherwise need similar checks. Apart from that, what's the reason for this? I was of the impression that we do support lazy basic attributes in embeddables as Steve was specially handling this in the initializer for embeddables AFAIR.

dreab8: I moved the fetch timing determination in the mapping model. I do not think we support lazy loading of basic attributes for embeddables but I may be wrong

beikov (commented):
IMO we should at least log a warning when a lazy basic attribute within an embeddable is encountered.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
```diff
@@ -309,8 +309,16 @@ public CascadeStyle getCascadeStyle() {
 			}
 		};
 
-		final FetchTiming fetchTiming = bootProperty.isLazy() ? FetchTiming.DELAYED : FetchTiming.IMMEDIATE;
-		final FetchStyle fetchStyle = bootProperty.isLazy() ? FetchStyle.SELECT : FetchStyle.JOIN;
+		final FetchTiming fetchTiming;
+		final FetchStyle fetchStyle;
+		if ( declaringType instanceof EmbeddableMappingType ) {
+			fetchTiming = FetchTiming.IMMEDIATE;
+			fetchStyle = FetchStyle.JOIN;
```
beikov, line 316: IMO we should emit a warning about this in case fetch timing was `DELAYED`. Are we sure that we must use `JOIN` fetching?

sebersole: I don't think the JOIN really matters here.  But then again, I don't think it matters what the value is so... :)

I guess I'd argue... "what's the other option?"?  SELECT?

sebersole: Also, how would you even denote an embeddable/embedded is DELAYED?  I don't think there is a way

sebersole: Ah, sorry... this is about LAZY for an attribute *on* the embeddable, not the embedded itself.  I misunderstood.

Then I agree that a warning would be nice

sebersole:
I Integrated a form of this adding the log message upstream.  See 

- https://github.com/hibernate/hibernate-orm/commit/9f155a95ab00d68a2d819fc75912be85f96ca0ef
- https://github.com/hibernate/hibernate-orm/commit/8203da2de14bb006978a5e4dee0612a6dc238341

PR #4374, opened 2021-11-21
Generate column order by order class properties

ngthotuan:
Update generate column order in database following the order of the properties
Ref: https://stackoverflow.com/questions/1298322/wrong-ordering-in-generated-table-in-jpa

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/PropertyContainer.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/PropertyContainer.java
@@ -9,21 +9,6 @@
 
 package org.hibernate.cfg;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


quaff:
It would be nice if hibernate generate columns by order of declaration in source file (It's deterministic in bytecode).
```java
class Parent {
String id;
String name;
}
class Child extends Parent{
String gender; 
int age;
}
```
should result in `id,name,gender,age`

beikov:
This simply won't work as the order is unspecified through the reflection API. Also see https://stackoverflow.com/questions/1097807/java-reflection-is-the-order-of-class-fields-and-methods-standardized

We haven an open discussion about the column ordering topic because there might be performance reasons to control the ordering: https://github.com/hibernate/hibernate-orm/discussions/3989

beikov:
By the way, if you care about the DDL representation, you should be using a dedicated tool like Liquibase or Flyway.

PR #4375, opened 2021-11-22
HHH-14936 JdbcConnectionContext in hibernate-testing throws NPE when user/password are not provided in configuration

yrodiere:
https://hibernate.atlassian.net/browse/HHH-14936


yrodiere:
See https://github.com/hibernate/hibernate-test-case-templates/issues/51

beikov:
Thanks!

PR #4376, opened 2021-11-22
Fix loading OnetoOne self referencing entity

dreab8:


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -998,6 +998,22 @@ public EntityFetch generateFetch(
...
+					fetchParent,
+					this,
+					fetchablePath,
+					foreignKeyDescriptor.createDomainResult(
+							fetchablePath,
+							parentTableGroup,
+							ForeignKeyDescriptor.Nature.TARGET,
+							creationState
+					),
+					true,
```
beikov, line 1012: Is this really always a select by unique key?

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -998,6 +998,22 @@ public EntityFetch generateFetch(
...
 
+		if ( cardinality == Cardinality.ONE_TO_ONE && bidirectionalAttributeName != null && fetchParent.getReferencedMappingType() == getEntityMappingType() ) {
+			return new EntityFetchSelectImpl(
+					fetchParent,
+					this,
+					fetchablePath,
+					foreignKeyDescriptor.createDomainResult(
+							fetchablePath,
+							parentTableGroup,
+							ForeignKeyDescriptor.Nature.TARGET,
```
beikov, line 1009: It seems to me, that this is not entirely correct. See how the side nature is determined below.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -998,6 +998,22 @@ public EntityFetch generateFetch(
...
+							creationState
+					),
+					true,
+					creationState
+			);
+		}
+
 		if ( fetchTiming == FetchTiming.IMMEDIATE && selected ) {
```
beikov, line 1017: Maybe it's better to exclude the above case to entering this if-branch instead as the below code block should handle delayed/select fetching?

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
dreab8: if I use the same way, the test fails may be the issue is the wrong value of the `this.sideNature`

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
dreab8: it seems so for the case case of ` cardinality == Cardinality.ONE_TO_ONE && bidirectionalAttributeName != null && fetchParent.getReferencedMappingType() == getEntityMappingType()`

beikov (approved):

PR #4380, opened 2021-11-23
HHH-14837 : Re-enable hibernate-jcache module.

chrisdennis:
Hey @sebersole. My PR against ehcache3 is an enormous ball of Gradle goo, so its still waiting on some poor soul to have the courage to review it. In order to expedite things I've instead cut an informal release with the proposed changes as `3.10.0-alpha0`. This PR is my first cut at re-enabling the `hibernate-jcache` module using that release. I made a cursory attempt at the archaeology required but I probably missed some code that should have been reinstated/uncommented.

Since the proposal for Ehcache 3 is to use Gradle capabilities and feature variants to handle this split it doesn't quite play perfectly with the way Hibernate's build system abstracts its dependencies. Currently I just left the version in `gradle/libraries.gradle` and the capability requirement in the couple of `build.gradle` files. If that's objectionable I imagine more work could be done to clean that split up.

### DIFF
--- a/hibernate-jcache/hibernate-jcache.gradle
+++ b/hibernate-jcache/hibernate-jcache.gradle
@@ -4,12 +4,16 @@ apply from: rootProject.file( 'gradle/published-java-module.gradle' )
 
 
 dependencies {
-    implementation project( ':hibernate-core' )
-    implementation libraries.jcache
+    api project( ':hibernate-core' )
+    api libraries.jcache
 
     testImplementation project( ':hibernate-testing' )
     testImplementation libraries.mockito
     testImplementation libraries.mockito_inline
 
-    testRuntimeOnly libraries.ehcache3
+    testRuntimeOnly(libraries.ehcache3) {
+        capabilities {
+            requireCapability 'org.ehcache.modules:ehcache-xml-jakarta'
+        }
+    }
 }
--- a/hibernate-jcache/src/test/java/org/hibernate/orm/test/jcache/RefreshUpdatedDataTest.java
+++ b/hibernate-jcache/src/test/java/org/hibernate/orm/test/jcache/RefreshUpdatedDataTest.java
@@ -8,12 +8,12 @@ package org.hibernate.orm.test.jcache;
 
 import java.util.ArrayList;
 import java.util.List;
-import javax.persistence.ElementCollection;
-import javax.persistence.Entity;
-import javax.persistence.GeneratedValue;
-import javax.persistence.Id;
-import javax.persistence.Table;
-import javax.persistence.Version;
+import jakarta.persistence.ElementCollection;
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.Id;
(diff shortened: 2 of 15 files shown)


beikov (approved):
Looks good to me, but how did you figure out which tests you need to "update"/revert partially to enable caching again like we had before? The Jakarta commit was pretty big. I just want to make sure we don't miss anything.

sebersole:
If I remember correctly I didn't actually disable tests, I actually removed
the entire modules from settings.xml.  Its hard to check from my phone (and
I'm on PTO 😉), so not sure exactly what the PR does but can look later

chrisdennis:
I ran `git log -Gjcache main..wip/6.0` and reverted all the relevant changes I could find. This mostly amounted to re-enabling the module and restoring a handful of tests under `documentation`.

hibernate-jcache/hibernate-jcache.gradle
```diff
@@ -4,12 +4,16 @@ apply from: rootProject.file( 'gradle/published-java-module.gradle' )
...
+    api project( ':hibernate-core' )
+    api libraries.jcache
 
     testImplementation project( ':hibernate-testing' )
     testImplementation libraries.mockito
     testImplementation libraries.mockito_inline
 
-    testRuntimeOnly libraries.ehcache3
+    testRuntimeOnly(libraries.ehcache3) {
+        capabilities {
```
chrisdennis, line 15: This block should really be colocated with the version in the libraries.gradle. Whether that is worth the re-engineering required is not something I can judge though.

hibernate-jcache/src/test/java/org/hibernate/orm/test/jcache/RefreshUpdatedDataTest.java
```diff
@@ -8,12 +8,12 @@
 
 import java.util.ArrayList;
 import java.util.List;
-import javax.persistence.ElementCollection;
-import javax.persistence.Entity;
-import javax.persistence.GeneratedValue;
-import javax.persistence.Id;
-import javax.persistence.Table;
-import javax.persistence.Version;
+import jakarta.persistence.ElementCollection;
```
chrisdennis, line 11: I've been a bit inconsistent with import ordering... Whats the protocol when choosing between minimized changes and correct import order?

sebersole: We don't really have a hard/fast wrt "minimizing clonflicts".  No one is likely to have touched the files under hibernate-jcache though, so this is fine.

Ordering is defined by our importable code styles for IJ and Eclipse IDEs

hibernate-jcache/hibernate-jcache.gradle
sebersole: libraries.gradle simply defines the versions we want to use.  It does not apply any dependencies

sebersole (commented):
lgtm

sebersole:
Thanks for the work @chrisdennis !

Applied it upstream

PR #4382, opened 2021-11-24
HHH-14938 Upgrade to MySQL Connector/J 8.0.27

Sanne:
- https://hibernate.atlassian.net/browse/HHH-14938

### DIFF
--- a/docker_db.sh
+++ b/docker_db.sh
@@ -2,7 +2,7 @@
 
 mysql_5_7() {
     docker rm -f mysql || true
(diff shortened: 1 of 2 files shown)


beikov:
Ok, so it seems the drivers did something funky until 8.0.26 with respect to charset/collation determination as per https://docs.oracle.com/cd/E17952_01/connector-j-8.0-relnotes-en/news-8-0-26.html which is partly related to the error we are seeing. Drivers apparently didn't take the server setting which I enabled now into account before 8.0.26, but since they now do, this is fine. It seems that previously, the JDBC driver tried to somehow determine some default collation for the server reported charset which seems to have worked so far, although I don't understand why. Anyway, with the new driver, the driver chosen collation is simply wrong and we would either have to set the collation explicitly or rely on the skip handshake setting which I enabled now.

Explicit configuration is cumbersome because MySQL5 and 8 use different charsets/collations which would require new database profiles...

Sanne:
thanks for having a look @beikov !

Regarding:



> [...] we would either have to set the collation explicitly or rely on the skip handshake setting which I enabled now.

Explicitly how, via JDBC properties? What did you change, can I see it?

beikov:
One can set the `connectionCollation` JDBC property to the desired collation.

PR #4388, opened 2021-11-24
Remove 'statementInspectorClass' from @Jpa

jrenaat:
Change 'provider' attribute in @SettingProvider to ' providerClass'
Remove 'statementInspectorClass' from @Jpa and refactor its usage to use a @SettingProvider


hibernate-testing/src/main/java/org/hibernate/testing/orm/junit/SettingProvider.java
```diff
@@ -15,5 +15,5 @@
 	}
 
 	String settingName();
-	Class<? extends Provider<?>> provider();
+	Class<? extends Provider<?>> providerClass();
```
beikov, line 18: Please revert to the shorter name. It doesn't add any value and will reduce the amount of files in the PR.

beikov (commented):
Other than the name of the annotation member this is fine.

beikov (approved):

PR #4389, opened 2021-11-25
Wip/6.0 Support concurrent queries having the same query text

fax4ever:
Made by me and @dreab8!

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/internal/QueryInterpretationCacheStandardImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/internal/QueryInterpretationCacheStandardImpl.java
@@ -52,7 +52,7 @@ public class QueryInterpretationCacheStandardImpl implements QueryInterpretation
 	 */
 	private final BoundedConcurrentHashMap<Key, QueryPlan> queryPlanCache;
 
-	private final BoundedConcurrentHashMap<String, HqlInterpretation> hqlInterpretationCache;
+	private final BoundedConcurrentHashMap<String, ImmutableHqlInterpretation> hqlInterpretationCache;
 	private final BoundedConcurrentHashMap<String, ParameterInterpretation> nativeQueryParamCache;
 	private final Supplier<StatisticsImplementor> statisticsSupplier;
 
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/query/internal/QueryInterpretationCacheStandardImpl.java
```diff
@@ -111,41 +111,49 @@ public HqlInterpretation resolveHqlInterpretation(
...
+			if ( sqmStatement.getSqmParameters().isEmpty() ) {
+				domainParameterXref = DomainParameterXref.empty();
+				parameterMetadata = ParameterMetadataImpl.EMPTY;
+			}
+			else {
+				domainParameterXref = DomainParameterXref.from( sqmStatement );
+				parameterMetadata = new ParameterMetadataImpl( domainParameterXref.getQueryParameters() );
+			}
+
+			immutableHqlInterpretation = new ImmutableHqlInterpretation( sqmStatement, parameterMetadata);
```
NathanQingyangXu, line 131: I think we can use 'final' above if only for consistency; otherwise, simply combine the two statements into one:
```
hqlInterpretationCache.put( queryString, new ImmutableHqlInterpretation( sqmStatement, parameterMetadata ) );
```
We might well insert a space char before the ending ')' above as well, though trivial.

hibernate-core/src/test/java/org/hibernate/orm/test/query/sqm/ConcurrentQueriesByIdsTest.java
```diff
@@ -0,0 +1,103 @@
...
+				.setParameter(
+						"ids",
+						Arrays.asList( base + 0, base + 1, base + 2, base + 3, base + 4, base + 5,
+									   base + 6, base + 7, base + 8, base + 9
+						)
+				)
+				.list();
+	}
+
+	@Entity(name = "simple")
```
NathanQingyangXu, line 78: Usually we use Uppercase for entity name (so it can stand out in HQL string)? Not a big deal for sure.

hibernate-core/src/test/java/org/hibernate/orm/test/query/sqm/ConcurrentQueriesByIdsTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+	private List<SimpleEntity> executeQuery(Session session, int index) {
+		int base = index * 10;
+
+		return session.createQuery( QUERY_STRING, SimpleEntity.class )
+				.setParameter(
+						"ids",
+						Arrays.asList( base + 0, base + 1, base + 2, base + 3, base + 4, base + 5,
+									   base + 6, base + 7, base + 8, base + 9
+						)
```
NathanQingyangXu, line 73: Given that we have decided to choose JDK11 for v6, we might use the following List constant feature (`List.of()`) from JDK9. No need to change above though but it is a good reminder in the future.

fax4ever:
Thank you @beikov ;)

PR #4392, opened 2021-11-26
proposed API for instantiation of detached collections

gavinking:
This is designed to be useful for object-graph deserialization.

It's just a rough draft for discussion.

Summary: right now, you can create an uninitialized detached collection just by doing `new PersistentList()` or `new PersistentSet()`, or whatever, and it seems to work. But that exposes you to internal Hibernate APIs.

This PR adds an API to the `Hibernate` class that let's you do things like:

    Hibernate.<Element>set().createDetachedInstance()
    Hibernate.<Key,Value>Map().createNewInstance()

which is useful for this sort of generic code.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/Hibernate.java
+++ b/hibernate-core/src/main/java/org/hibernate/Hibernate.java
@@ -6,18 +6,38 @@
  */
 package org.hibernate;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.SortedMap;
+import java.util.SortedSet;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import java.util.function.Supplier;
 
 import org.hibernate.bytecode.enhance.spi.interceptor.BytecodeLazyAttributeInterceptor;
 import org.hibernate.bytecode.enhance.spi.interceptor.EnhancementAsProxyLazinessInterceptor;
+import org.hibernate.collection.internal.PersistentBag;
+import org.hibernate.collection.internal.PersistentList;
+import org.hibernate.collection.internal.PersistentMap;
+import org.hibernate.collection.internal.PersistentSet;
+import org.hibernate.collection.internal.PersistentSortedMap;
+import org.hibernate.collection.internal.PersistentSortedSet;
 import org.hibernate.collection.spi.PersistentCollection;
 import org.hibernate.engine.HibernateIterator;
 import org.hibernate.engine.jdbc.LobCreator;
 import org.hibernate.engine.jdbc.spi.JdbcServices;
 import org.hibernate.engine.spi.PersistentAttributeInterceptable;
 import org.hibernate.engine.spi.PersistentAttributeInterceptor;
+import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SessionImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.proxy.HibernateProxy;
 import org.hibernate.proxy.LazyInitializer;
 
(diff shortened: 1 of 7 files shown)


sebersole (commented):
Looks like a good start to me.

gavinking:
So I would like to let you do this for to-one associations as well, that is, I would like a method like `createDetachedProxy(class, id)`. Unfortunately, that requires a `SessionFactory`. So, what to do?

- Move all this stuff to `SessionFactory`, or
- leave it there, but make the signature `createDetachedProxy(class, id, sessionFactory)`?

gavinking:
OK, so I have:

1. now added `Hibernate.createDetachedProxy()` which works fine, I think, and
2. added Javadoc everywhere.

If anyone has feedback on the API, please speak up.

Otherwise, I consider this ready to merge.

hibernate-core/src/main/java/org/hibernate/Hibernate.java
```diff
@@ -239,4 +259,158 @@ public static Object unproxy(Object proxy) {
...
+		}
+		return (E) persister.createProxy(id, null);
+	}
+
+	/**
+	 * Operations for obtaining references to persistent collections of a certain type.
+	 *
+	 * @param <C> the type of collection, for example, <tt>List&lt;User&gt;</tt>
+	 */
+	public static final class CollectionInterface<C> {
```
beikov, line 291: I don't see why it is useful to have this i.e. it doesn't do anything a user would not be able to do through `new WhateverCollection()` or `new PersistentWhateverCollection()`. Creating a collection proxy on the other hand would IMO be something new that would have merit to have, but like I stated before, I'd rather like an API like this:

```
public static <C> C createProxy(PluralAttribute<C, ?, ?> attr) {...}
```

gavinking: > I don't see why it is useful to have this

I agree it's not useful for people handwriting code, but that's not the main usecause here. It might be quite useful for generic code, i.e. serialization frameworks, which is how this is most likely to be used.

> I'd rather like an API like this:

I have absolutely no problem with providing an API like that, and indeed it can be easily implemented on top of what I've proposed. But the API you just showed is _much_ less convenient for framework code, and the extra typesafety is completely useless, since the framework code is going to be working with things in a detyped way.

Summary: I've designed this with the needs of a serialization framework developer in mind.

beikov: But even the serialization framework needs some sort of type structure contract to traverse an object along, and why not use the JPA metmodel for this purpose? Apart from that, I would also add 

```
public static <V> V createDefault(Attribute<?, V> attr) {...}
```

To create whatever default value should be used for an attribute which could be an `ArrayList`, `null` or some primitive value.

gavinking: The serialization framework might not even have access to a `SessionFactory`.

gavinking: And, I mean, it's almost  certainly *not* going to be traversing the Hibernate meta model! It's going to have its whole own logic for walking an object graph. 

The user is just plugging in custom logic for dealing with Hibernate proxies.

beikov: I don't know, I would expect the `SessionFactory` to be available during deserialization, but I know there are some serialization framworks that make this really hard, so yeah, maybe this has merit. IMO we should have at least 1 user of this API (even if it is just us trying it out in some deserializer) that confirms this works great for some purpose before we introduce it. People that write this custom logic could just as well write `new PersistentWhateverCollection()`, so I'm questioning what the particular value of this code is. Is it just convenience?

gavinking: The value is you're not telling people to call what is quite clearly marked as an internal API that their program is not supposed to depend on?

beikov: Hmm, ok, that makes sense. I didn't realize the persistent collections are internal API.

sebersole: > Hmm, ok, that makes sense. I didn't realize the persistent collections are internal API.

We really should have a review of that.  Conceptually I'd say it is a mix - some of it could easily be API/SPI while other parts are clearly internal details.

But I think the PR is fine, as far as it goes.

gavinking: > We really should have a review of that. Conceptually I'd say it is a mix - some of it could easily be API/SPI while other parts are clearly internal details.

That would be worth doing indeed.

PR #4393, opened 2021-11-26
fix the signature of Session.merge()

gavinking:
Previously `Session.merge()` erased the generic type parameter inherited from the supertype, obligating the client code to perform a typecast.

See https://github.com/hibernate/hibernate-orm/discussions/4391


gavinking:
Are there any objections to this change? I think it should simply be merged without further ado.

PR #4399, opened 2021-11-26
Require java11

Sanne:
Marking as "Draft" as I need to have a better look. If anyone wants to have a preview it's here :)


sebersole (commented):
Looking good so far

Sanne:
Updated it to include changes in the github actions setup.

PR #4403, opened 2021-11-30
Testcase for HHH-14943

pb00067:
added method testManyToOneNaturalLoadByNaturalId

unless fixed it fails with a
org.hibernate.exception.SQLGrammarException: could not prepare statement

N.B.: to make things easy, I had to convert the fetch-type to citizen from EAGER to LAZY in NaturalIdOnManyToOne.java 
Otherwise the query in question includes a join statement and thus remained parseable despite being wrong.


dreab8:
Fixed

pb00067:
Thank you very much for solving this issue Andrea!

dreab8:
@pb00067 you are more than welcome and thanks a lot for the reproducer!

pb00067:
@dreab8 May I ask you if you can also take a look at HHH-14944 in the next weeks?
I opened it the same day as HHH-14943

PR #4404, opened 2021-11-30
Fix o.h.Interceptor#onSave call

dreab8:
the issue is 
```
default boolean onSave(Object entity, Object id, Object[] state, String[] propertyNames, Type[] types)
			throws CallbackException {
		if (id instanceof Serializable) {
			return onSave(entity, (Serializable) id, state, propertyNames, types);
		}
		return false;
	}
```
when the deprecated `boolean onSave(Object entity, Serializable id, Object[] state, String[] propertyNames, Type[] types)` is implemented, if `id` is null then `false` is wrongly returned.


beikov (approved):

gavinking:
I'm pretty sure this has to have been the wrong fix for this problem.

gavinking:
The correct fix is _much_ simpler: just use `if (id==null || id instanceof Serializable)`. Really it's only needed for `onSave()` because that's the only case where the `id` can be null.

PR #4406, opened 2021-12-01
TestCase for HHH-14944 Detach removes NaturalIdCrossRef in 2nd level cache too. 

pb00067:
Bug/Regression.
Detaching entity removes NaturalIdCrossReference in second level cache
too. Should operate on first level cache only.


gavinking:
The linked issue was closed as fixed, and this test appears to be in the code base as `CachedMutableNaturalIdTest`.

PR #4407, opened 2021-12-01
Wip/6.0 reenabling test

dreab8:


hibernate-core/src/main/java/org/hibernate/query/results/DomainResultCreationStateImpl.java
```diff
@@ -326,28 +333,47 @@ public DynamicFetchBuilderLegacy resolve(String ownerTableAlias, String fetchedP
...
+						.getModelPart());
+
+				if ( explicitFetchBuilder == null && partName != null ) {
+					currentEntry = new AbstractMap.SimpleEntry<>(
+							currentEntry.getKey() + "." + partName,
+							currentEntry.getValue().append( partName )
+					);
+					explicitFetchBuilder = fetchBuilderResolverStack
+							.getCurrent()
+							.apply( currentEntry.getKey() );
```
beikov, line 370: Not sure if we actually have a test where any of this returns a fetch builder, but I believe that whatever fetch builder we get, it is not the one that we should actually use for this fetchable, as that would represent the FK. Either way, there is something missing here:

```suggestion
					explicitFetchBuilder = fetchBuilderResolverStack
							.getCurrent()
							.apply( currentEntry.getKey() );
					if ( explicitFetchBuilder == null ) {
						fetchBuilderLegacy = legacyFetchResolver.resolve(
							fromClauseAccess.findTableGroup( fetchParent.getNavigablePath() )
									.getPrimaryTableReference()
									.getIdentificationVariable(),
							fetchableName + "." + partName
						);
					}
```

But maybe it's best if we try to write a test case that hits this so we can see what happens. A `@ColumnResult(name = "captain.id"` or something should show the issue in case we actually fetch data.

PR #4411, opened 2021-12-02
HHH-14948 - Reduce the size of the imports cache in the metamodel

ivaylomitrev:
Corresponding task: [HHH-14948](https://hibernate.atlassian.net/browse/HHH-14948)

The original change from [HHH-12485](https://hibernate.atlassian.net/browse/HHH-12485) introduced memory issues (as described in the corresponding task). As such, I am requesting that it is reverted.

I have also added a test to illustrate the issue as is the requirement for bugfix tasks. That said, I don't particularly like using reflection in the way I have so I don't mind dropping the test altogether or accepting advice from more experienced Hibernate developers on how I could perform the test better.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetamodelImpl.java
@@ -632,7 +632,13 @@ public class MetamodelImpl implements MetamodelImplementor, Serializable {
 				return className;
 			}
 			catch ( ClassLoadingException cnfe ) {
-				imports.put( className, INVALID_IMPORT );
+				// This check doesn't necessarily mean that the map can't exceed 1000 elements because
+				// new entries might be added _while_ performing the check (making it 1000+ since size() isn't
+				// synchronized). Regardless, this would pass as "good enough" to prevent the map from growing
+				// above a certain threshold, thus, avoiding memory issues.
+				if ( imports.size() < 1_000 ) {
+					imports.put( className, INVALID_IMPORT );
+				}
 				return null;
 			}
 		}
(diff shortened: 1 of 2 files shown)


Sanne:
Great patch, thank you.

So technically this is a memory leak, the only reason for nobody having noticed so far is likely related with the fact that it requires the "dynamically" generated HQL to be parsed like you mentioned on Jira.

So I'm wondering, while of course a memory leak needs to be fixed as an unbounded memory consumption, I also do remember that the negative caching to avoid CNF exceptions being thrown was very beneficial in the more common use case of having constant HQL, or deterministically generated HQL strings.

How would you feel about changing the patch slightly so that, instead of fully reverting the previous fix, it only stores the empty string if the map size isn't grown too far?

This implies needing to define a threshold; not sure if adding a new configuration option is worth it - we could keep it simple and hardcode a reasonable guess like 1000 - this is generous enough to be unlikely to be reached in most applications, but will ensure an upper bound for use cases like yours.
Maybe a tunable property would be better, but it's never going to be a perfect guess anyway; while having a constant will provide better semantics than fully reverting the patch; I suppose some people might want to tune it but we can wait to see if there's demand for that.

ivaylomitrev:
Thanks for the detailed response!

Your proposal sounds perfectly reasonable. Storing 1000 imported class names shouldn't have any adverse effect on any modern application in terms of memory consumption and the fix would still work for our use case (as pointed out by you).

I will force push a modified version of the code later today in the same branch and write a comment when this has been done.

ivaylomitrev:
I've now pushed the requested change and have modified the not-so-beautiful test accordingly. The cap on the imports map will now be in the range of ~1000. I'm saying "range", because `ConcurrentHashMap#size()` is not guaranteed to reflect the real-time size of the map (due to potential concurrent updates). I think that's still safe enough though.

I've also tried to describe the reasoning you provided earlier, @Sanne , in the commit description.

Let me know if you think there's anything else that needs to be addressed and I'll happily respond.

Sanne:
Looks great, thank you! I'll merge it soon, assuming CI is happy with it.

PR #4413, opened 2021-12-03
initial work on DatabaseVersion#isAfter

sebersole:
Opening as draft as I wanted to get feedback about the contracts here prior to changing `Dialect#getVersion` and its uses


hibernate-core/src/main/java/org/hibernate/dialect/DatabaseVersion.java
```diff
@@ -0,0 +1,253 @@
...
+	 * Determine whether this version before the passed one
+	 */
+	default boolean isBefore(Integer major, Integer minor) {
+		return isBefore( (int) major, minor == null ? NO_VERSION : minor );
+	}
+
+	/**
+	 * Determine whether this version before the passed one
+	 */
+	default boolean isBefore(int otherMajor, int otherMinor, int otherMicro) {
```
beikov, line 238: I would prefer the name `isAtLeast` but YMMV

gavinking (approved):
Yup, all LGTM.

beikov (approved):

dreab8 (approved):
LGTM :+1:

hibernate-core/src/main/java/org/hibernate/dialect/DatabaseVersion.java
sebersole: As an alternative to isBefore?  I assume you mean isSince?

beikov: Yeah. Apparently I clicked on the wrong line

sebersole:
Applied upstream

hibernate-core/src/main/java/org/hibernate/dialect/DatabaseVersion.java
sebersole: I landed on `#isSameOrAfter`

PR #4414, opened 2021-12-04
HHH-14950 - Support mapping of embeddables with no setters (assuming a custom instantiator or repo-strategy is used)

sebersole:
HHH-14950 - Support mapping of embeddables with no setters (assuming a custom instantiator or repo-strategy is used)

Tests


sebersole:
Out of date

PR #4429, opened 2021-12-07
IdentifierGenerator(Factory) related work

sebersole:
HHH-14497 - Drop legacy id-generator settings;
HHH-14718 - Drop deprecated generator implementations;
HHH-14959 - Drop IdentifierGeneratorFactory as a Service


gavinking:
Oh very nice. It's already done.

So, my https://github.com/hibernate/hibernate-orm/discussions/4427#discussioncomment-1771094 WDYT, can it be done?

yrodiere:
>  HHH-14959 - Drop IdentifierGeneratorFactory as a Service

@sebersole this means we urgently need a fix for [HHH-14958](https://hibernate.atlassian.net/browse/HHH-14958), because until now overriding the `IdentifierGeneratorFactory` service was our only workaround. Do you intend to fix HHH-14958 soon? Otherwise, it might be better to keep the `IdentifierGeneratorFactory` service for now...

yrodiere:
> 


> > HHH-14959 - Drop IdentifierGeneratorFactory as a Service
> 
> @sebersole this means we urgently need a fix for [HHH-14958](https://hibernate.atlassian.net/browse/HHH-14958), because until now overriding the `IdentifierGeneratorFactory` service was our only workaround. Do you intend to fix HHH-14958 soon? Otherwise, it might be better to keep the `IdentifierGeneratorFactory` service for now...

Also, even Quarkus needs that service to disable the "instantiate the generator through CDI" behavior. Though I'm not that's for the same reason. @Sanne might have more to say about this.

sebersole:
> > HHH-14959 - Drop IdentifierGeneratorFactory as a Service
> 
> @sebersole this means we urgently need a fix for [HHH-14958](https://hibernate.atlassian.net/browse/HHH-14958), because until now overriding the `IdentifierGeneratorFactory` service was our only workaround. Do you intend to fix HHH-14958 soon? Otherwise, it might be better to keep the `IdentifierGeneratorFactory` service for now...

Have you tried it?

sebersole:
> > 
> 
> > > HHH-14959 - Drop IdentifierGeneratorFactory as a Service
> > 
> > 
> > @sebersole this means we urgently need a fix for [HHH-14958](https://hibernate.atlassian.net/browse/HHH-14958), because until now overriding the `IdentifierGeneratorFactory` service was our only workaround. Do you intend to fix HHH-14958 soon? Otherwise, it might be better to keep the `IdentifierGeneratorFactory` service for now...
> 
> Also, even Quarkus needs that service to disable the "instantiate the generator through CDI" behavior. Though I'm not sure that's for the same reason. @Sanne might have more to say about this.

Yep, have you tried it? ;)

sebersole:
> > HHH-14959 - Drop IdentifierGeneratorFactory as a Service
> 
> @sebersole this means we urgently need a fix for [HHH-14958](https://hibernate.atlassian.net/browse/HHH-14958), because until now overriding the `IdentifierGeneratorFactory` service was our only workaround. Do you intend to fix HHH-14958 soon? Otherwise, it might be better to keep the `IdentifierGeneratorFactory` service for now...

So the "problem" with keeping this a service is that it is kept around until SessionFactory closes while it is only used during this bootstrap phase.  So it won't be made a service again, but if you really find it is still an issue (you have not tried and I do not think it is) we will provide a way for you to supply a custom one

yrodiere:
> > > HHH-14959 - Drop IdentifierGeneratorFactory as a Service
> > 
> > 
> > @sebersole this means we urgently need a fix for [HHH-14958](https://hibernate.atlassian.net/browse/HHH-14958), because until now overriding the `IdentifierGeneratorFactory` service was our only workaround. Do you intend to fix HHH-14958 soon? Otherwise, it might be better to keep the `IdentifierGeneratorFactory` service for now...
> 
> Have you tried it?

No I haven't, I'm still trying to make 5.6 + Search 6 work in WildFly for now.

I'll interpret this as "Yes Yoann, I intend to fix this, in fact I already have". We'll know soon enough if there are still problems for ORM 6 in WildFly or Quarkus, I suppose.

sebersole:
> I'll interpret this as "Yes Yoann, I intend to fix this, in fact I already have". We'll know soon enough if there are still problems for ORM 6 in WildFly or Quarkus, I suppose.

Yes, I think the flag you added could have just as easily been calculated - which is what I did.  

If that does in fact not work, then fine - we will look at allowing you to plug in a custom factory (though not as a service)

sebersole:
I've pushed this work thus far upstream.  We can tweak it from there

gavinking:
Excellent!

PR #4431, opened 2021-12-09
Register sequence-identity strategy

dreab8:


sebersole:
Per our discussion, this should not be integrated

PR #4433, opened 2021-12-09
replace use of <br> with <p> in all the javadoc

gavinking:


gavinking:
I'm merging this, because it's trivial, and I want to avoid conflicts with a different PR I'm about to push.

PR #4434, opened 2021-12-09
fix some warnings and clean up some typing issues

gavinking:
This is a general cleanup of the Session + Query hierarchies, focussing on stuff the IDE was complaining about..

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/SharedSessionContract.java
+++ b/hibernate-core/src/main/java/org/hibernate/SharedSessionContract.java
@@ -185,21 +185,6 @@ public interface SharedSessionContract extends QueryProducer, Closeable, Seriali
 	 * @throws IllegalStateException if the StatelessSession has been closed
 	 */
 	HibernateCriteriaBuilder getCriteriaBuilder();
-//
-//	@Override
-//	<T> org.hibernate.query.Query<T> createQuery(String queryString, Class<T> resultType);
-//
-//	@Override
-//	<T> org.hibernate.query.Query<T> createQuery(CriteriaQuery<T> criteriaQuery);
-//
-//	@Override
-//	<R> org.hibernate.query.Query<R> createQuery(CriteriaUpdate<?> updateQuery);
-//
-//	@Override
-//	<R> org.hibernate.query.Query<R> createQuery(CriteriaDelete<?> deleteQuery);
-//
-//	@Override
-//	<T> org.hibernate.query.Query<T> createNamedQuery(String name, Class<T> resultType);
 
 	/**
 	 * Controller for allowing users to perform JDBC related work using the Connection managed by this Session.
--- a/hibernate-core/src/main/java/org/hibernate/SharedSessionContract.java
+++ b/hibernate-core/src/main/java/org/hibernate/SharedSessionContract.java
@@ -6,30 +6,23 @@
  */
 package org.hibernate;
 
+import java.io.Closeable;
 import java.io.Serializable;
 
-import jakarta.persistence.StoredProcedureQuery;
 import jakarta.persistence.criteria.CriteriaBuilder;
-import jakarta.persistence.criteria.CriteriaDelete;
-import jakarta.persistence.criteria.CriteriaQuery;
-import jakarta.persistence.criteria.CriteriaUpdate;
 
 import org.hibernate.jdbc.ReturningWork;
 import org.hibernate.jdbc.Work;
 import org.hibernate.procedure.ProcedureCall;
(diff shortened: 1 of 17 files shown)


hibernate-core/src/main/java/org/hibernate/SharedSessionContract.java
```diff
@@ -18,18 +18,15 @@
...
- * <p/>
- * NOTE : Poorly named.  "shared" simply indicates that its a unified contract between {@link Session} and
- * {@link StatelessSession}.
- * 
+ * Declares operations that are common between {@link Session} and {@link StatelessSession}.
+ *
  * @author Steve Ebersole
  */
-public interface SharedSessionContract extends QueryProducer, Serializable {
+//NOTE : This interface is poorly named: "shared" simply indicates that it's a unified contract
```
sebersole, line 28: Please remove this line.  We had a very long and drawn out discussion and consensus about the naming here.  Whether you like it or not, you had the chance to chime in when we had that original discussion ;)

hibernate-core/src/main/java/org/hibernate/engine/spi/SessionImplementor.java
```diff
@@ -26,7 +25,8 @@
  * {@link org.hibernate.type.Type}, {@link EntityPersister}
  * and {@link org.hibernate.persister.collection.CollectionPersister} implementations.
  *
- * A Session, through this interface and SharedSessionContractImplementor, implements:<ul>
+ * A Session, through this interface and SharedSessionContractImplementor, implements:
+ * <ul>
```
sebersole, line 29: Minor so don't worry about it here, but in the future this was the preferred syntax for html lists

hibernate-core/src/main/java/org/hibernate/engine/spi/SessionImplementor.java
```diff
@@ -157,4 +117,44 @@ default SessionImplementor getSession() {
...
+	@Override
+	<T> QueryImplementor<T> createQuery(CriteriaQuery<T> criteriaQuery);
+
+	@Override @SuppressWarnings("rawtypes")
+	QueryImplementor createQuery(CriteriaUpdate updateQuery);
+
+	@Override @SuppressWarnings("rawtypes")
+	QueryImplementor createQuery(CriteriaDelete deleteQuery);
+
+//		@Override @SuppressWarnings("rawtypes")
```
sebersole, line 131: Please remove this commented out block

hibernate-core/src/main/java/org/hibernate/query/Query.java
```diff
@@ -640,40 +640,6 @@ default R getSingleResult() {
 	 */
 	<P> Query<R> setParameterList(int position, Collection<? extends P> values, AllowableParameterType<P> type);
 
-//	/**
```
sebersole, line 643: Like you did here ;)

sebersole (commented):
Left a few comments, other than that.. lgtm

hibernate-core/src/main/java/org/hibernate/SharedSessionContract.java
gavinking: I didn't add that line, it was already there, I just moved it out of the Javadoc and into a regular comment.

Do you want me to put it back into the Javadoc? Or just delete it altogether (my preference).

sebersole: Just remove it

hibernate-core/src/main/java/org/hibernate/query/spi/AbstractQuery.java
```diff
@@ -1403,8 +1403,7 @@ else if ( retType.isArray() ) {
 	}
 
 	@Override
-	@SuppressWarnings( "rawtypes" )
-	public QueryImplementor<R> setProperties(Map map) {
+	public QueryImplementor<R> setProperties(Map<?,?> map) {
```
NathanQingyangXu, line 1406: I always think inserting a space char between `?,?` improves code readability (and in line with our existing coding format). 
```
public QueryImplementor<R> setProperties(Map<?, ?> map) {
... ...
}
```
I always hesitated whether nailing down the first type generics to String is better or not
```
public QueryImplementor<R> setProperties(Map<String, ?> map) {
```

sebersole: While we don't have a specific code guideline around this, I personally prefer the no space version as well until these get "nested".

But I'm certainly not going to ask someone to make that change in a PR

PR #4435, opened 2021-12-09
remove left-behind method createSQLQuery() from StatelessSession

gavinking:
It was already annotated `@Remove`.


gavinking:
The CI passes, with just the unrelated Derby failure, so I think I can merge this.

sebersole (commented):
lgtm

sebersole:
Yes, derby and Oracle failures are happening even upstream.  So I agree that this is unrelated

PR #4436, opened 2021-12-09
finally get rid of Session.connection()/disconnect()/reconnect()

gavinking:
This nasty stuff has stuck around for far too long.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/Session.java
+++ b/hibernate-core/src/main/java/org/hibernate/Session.java
@@ -7,7 +7,6 @@
 package org.hibernate;
 
 import java.io.Closeable;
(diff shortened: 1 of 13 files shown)


sebersole:
Yes, I removed the need for these years ago.

Been delaying dropping them because they potentially break user code and really have no harm.

But I agree, with this major, it is time

sebersole:
I'll merge once the actions and checks finish

sebersole:
Sheesh these actions take forever

sebersole:
Merging this as the test failures seem to be consistent with the failures from all jobs.  As far as I know, @beikov said he has a fix for those already.

gavinking:
Thanks!

PR #4437, opened 2021-12-09
Support statistics when criteria(s) are executed

fax4ever:
Running the same test with Hibernate ORM 5,
it seems that statistics are enabled on JPA criteria queries...


fax4ever:
Thank you Steve!

PR #4438, opened 2021-12-09
HHH-14970 make use of new Java language features from v8 to v11

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14970

Mainly focused on the following two new Java language features introduced from v8 to v11 (I referred to https://ondro.inginea.eu/index.php/new-features-in-java-versions-since-java-8/):

- new String's methods in JDK 11 (e.g. String#isBlank());

- new Collection factory methods in JDK 9 (i.e., List.of(), Map.of(), Set.of()); 

Demonstrably `List.of()` is better than `Arrays.asList()` even in immutability context (see https://stackoverflow.com/questions/46579074/what-is-the-difference-between-list-of-and-arrays-aslist, e.g.), but given the latter has been an established pattern, it won't be touched upon. This PR mainly focused on easy migration so it is easy to review and verify.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/spi/AbstractRegionFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/spi/AbstractRegionFactory.java
@@ -6,8 +6,6 @@
  */
 package org.hibernate.cache.spi;
 
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


sebersole (commented):
lgtm.  i'll integrate once the checks complete

NathanQingyangXu:
> lgtm. i'll integrate once the checks complete

seems the two broken testing failures are common ones (same as in https://github.com/hibernate/hibernate-orm/pull/4434).

PR #4439, opened 2021-12-10
Cache and expose column size to get the (default and non-default) scale

fax4ever:
From Hibernate Search we need to have the scale of the column that is finally applied by ORM.
This is the solution we found with @dreab8 and @yrodiere.

I wouldn't have exposed such a method but instead `getColumnSize` without any parameter, but I was one versus two :)


sebersole:
If you need it, you need it.  Curious how you did this with ORM 5 though...

sebersole (commented):
lgtm

fax4ever:
> If you need it, you need it. Curious how you did this with ORM 5 though...

We simply get the value directly from the column.
The very difference is that if the value is not set by the user, that is taken from the Database default, the value now is not present on the Column and we need the its Size...

fax4ever:
Thanks for the feedback!

PR #4441, opened 2021-12-10
proposed solution to "long varchar" problem

gavinking:
I'm really very happy with this, it just makes sense:

- it makes the supported dialects more graceful and much more consistent in their handling of long column lengths
- it makes it easy for people already using JPA's `@Column` or even Hibernate Validators `@Size` annotation
- it nicely distinguishes between 16-bit "long" (MySQL `text`) and really damn 32-bit "long"
- it has an incredibly minimal impact

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/SpannerDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/SpannerDialect.java
@@ -92,17 +92,20 @@ public class SpannerDialect extends Dialect {
 		//there is no time type of any kind
 		registerColumnType( Types.TIME, "timestamp" );
 
-		final int stringMaxLength = 2_621_440;
-		final int bytesMaxLength = 10_485_760;
+		registerColumnType( Types.CHAR, getMaxVarcharLength(), "string($l)" );
+		registerColumnType( Types.CHAR, "string(max)" );
+		registerColumnType( Types.VARCHAR, getMaxVarcharLength(), "string($l)" );
+		registerColumnType( Types.VARCHAR, "string(max)" );
 
-		registerColumnType( Types.CHAR, stringMaxLength, "string($l)" );
-		registerColumnType( Types.VARCHAR, stringMaxLength, "string($l)" );
+		registerColumnType( Types.NCHAR, getMaxNVarcharLength(), "string($l)" );
+		registerColumnType( Types.NCHAR, "string(max)" );
+		registerColumnType( Types.NVARCHAR, getMaxNVarcharLength(), "string($l)" );
+		registerColumnType( Types.NVARCHAR, "string(max)" );
 
-		registerColumnType( Types.NCHAR, stringMaxLength, "string($l)" );
-		registerColumnType( Types.NVARCHAR, stringMaxLength, "string($l)" );
-
-		registerColumnType( Types.BINARY, bytesMaxLength, "bytes($l)" );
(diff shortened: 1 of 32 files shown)


gavinking:
Note that since this PR also cleans up the mappings for `LONGVARCHAR`, this stuff is entirely complementary to the existing `LongVarcharJdbcType` in H6, and allows and improves the `alternative of `@JdbcType(LongVarcharJdbcType.class)`.

hibernate-core/src/main/java/org/hibernate/dialect/SpannerDialect.java
```diff
@@ -114,6 +119,15 @@ public SpannerDialect(DialectResolutionInfo info) {
 		registerKeywords( info );
 	}
 
+	@Override
+	public int getMaxVarcharLength() {
+		return 2_621_440;
```
sebersole, line 124: Wow, I relearned something.  Totally forgot you can format numbers this way! :laughing:

sebersole (commented):
Looks good to me so far.  Were you going to add support for any of the alternatives we discussed?  Or is that left as an exercise for us?

sebersole:
> > Note that since this PR also cleans up the mappings for `LONGVARCHAR`, this stuff is entirely complementary to the existing `LongVarcharJdbcType` in H6, and allows and improves the alternative of `@JdbcType(LongVarcharJdbcType.class)`.
> 
> And, FTR, this also improves the third alternative of `@JdbcTypeCode( Types.LONGVARCHAR )`.

Right, as I said on the Discussion, so long as the Dialects map that type code "properly" this will all JustWork

gavinking:
> Looks good to me so far. Were you going to add support for any of the alternatives we discussed? Or is that left as an exercise for us?

The only thing we've discussed that's not there is the XML side of things.

gavinking:
Now here's a change I'm uncertain about. In our discussion, we said that Oracle we would just use `varchar2(4000)` as the "max" type.

But previously Oracle actually "grew" the varchar type out to `clob`. (Like I said, the `Dialect`s were all totally inconsistent about this stuff in H5.)

So in principle this is a breaking change on Oracle, previously someone who had `@Column(10_000)` would have got a CLOB and it would have worked. With this change that would be an error.

Not sure quite what to think about that.

gavinking:
WTH, I'm getting CI failures due to what look like timeouts contacting the Maven repo!

sebersole:
> Now here's a change I'm uncertain about. In our discussion, we said that Oracle we would just use `varchar2(4000)` as the "max" type.
> 
> But previously Oracle actually "grew" the varchar type out to `clob`. (Like I said, the `Dialect`s were all totally inconsistent about this stuff in H5.)
> 
> So in principle this is a breaking change on Oracle, previously someone who had `@Column(10_000)` would have got a CLOB and it would have worked. With this change that would be an error.
> 
> Not sure quite what to think about that.

To be honest, I was wondering about that.  That scheme (VARCHAR -> CLOB) is really only dangerous on PGSQL as far as I know.

sebersole:
> WTH, I'm getting CI failures due to what look like timeouts contacting the Maven repo!

:shrug: 

I've seen a lot of transient issues with our CI jobs lately

gavinking:
> To be honest, I was wondering about that. That scheme (VARCHAR -> CLOB) is really only dangerous on PGSQL as far as I know.

Yeah, actually I think I'm going to roll back to that. Maxing out at 4000 characters isn't in the spirit of what we're trying to do here. For Oracle, `clob` makes sense.

PR #4442, opened 2021-12-10
Fix test matrix

gavinking:
This fixes some problems that prevented the test suite from running locally.

### DIFF
--- a/hibernate-testing/src/main/java/org/hibernate/testing/cleaner/PostgreSQLDatabaseCleaner.java
+++ b/hibernate-testing/src/main/java/org/hibernate/testing/cleaner/PostgreSQLDatabaseCleaner.java
@@ -52,7 +52,7 @@ public class PostgreSQLDatabaseCleaner implements DatabaseCleaner {
 				statement -> {
 					try {
 						return statement.executeQuery(
-								"SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME <> 'information_schema' AND SCHEMA_NAME <> 'sys' AND SCHEMA_NAME NOT LIKE 'pg_%'" );
+								"SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME <> 'information_schema' AND SCHEMA_NAME <> 'sys' AND SCHEMA_NAME <> 'public' AND SCHEMA_NAME NOT LIKE 'pg_%'" );
 					}
 					catch (SQLException sqlException) {
 						throw new RuntimeException( sqlException );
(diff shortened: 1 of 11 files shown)


hibernate-testing/src/main/java/org/hibernate/testing/cleaner/PostgreSQLDatabaseCleaner.java
```diff
@@ -52,7 +52,7 @@ public void clearAllSchemas(Connection connection) {
 				statement -> {
 					try {
 						return statement.executeQuery(
-								"SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME <> 'information_schema' AND SCHEMA_NAME <> 'sys' AND SCHEMA_NAME NOT LIKE 'pg_%'" );
+								"SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME <> 'information_schema' AND SCHEMA_NAME <> 'sys' AND SCHEMA_NAME <> 'public' AND SCHEMA_NAME NOT LIKE 'pg_%'" );
```
gavinking, line 55: I'm not really sure about this one, but when I created a database and user I didn't have permissions to trunc the `public` schema.

hibernate-testing/src/main/java/org/hibernate/testing/junit4/BaseUnitTestCase.java
```diff
@@ -35,7 +36,8 @@
 @RunWith( CustomRunner.class )
 public abstract class BaseUnitTestCase {
 
-	static {
+	@BeforeClassOnce
+	public void clearSchemas() {
 		DatabaseCleaner.clearSchemas();
```
gavinking, line 41: @beikov it's a bad idea to do this sort of thing in a static initialized, because errors get swallowed and turned into NCDFEs.

hibernate-testing/src/main/java/org/hibernate/testing/cleaner/PostgreSQLDatabaseCleaner.java
gavinking: @beikov thoughts? what permissions are needed to drop the `public` schema?

beikov: Since our testing is usually done with some sort of superuser this didn't come up yet, but I think it is fine to ignore the `public` schema.

hibernate-testing/src/main/java/org/hibernate/testing/junit4/BaseUnitTestCase.java
beikov: I know, sorry you had issues with this :(

hibernate-testing/src/main/java/org/hibernate/testing/cleaner/PostgreSQLDatabaseCleaner.java
gavinking: Ah, right, OK, that explains it.

hibernate-testing/src/main/java/org/hibernate/testing/junit4/BaseUnitTestCase.java
gavinking: Unfortunately this change wasn't enough to eliminate the NCDFEs in all cases. :-(

We need to figure out something better.

PR #4447, opened 2021-12-11
Support multiloading of single item

fax4ever:
Actually this is my attempt to support the case.
I don't know if it is correct, since I don't know too many details of Hibernate ORM 6 code base.
I'm here to fix it!

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/MultiIdLoaderStandard.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/MultiIdLoaderStandard.java
@@ -9,6 +9,7 @@ package org.hibernate.loader.ast.internal;
 import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 import org.hibernate.LockMode;
@@ -234,6 +235,9 @@ public class MultiIdLoaderStandard<T> implements MultiIdEntityLoader<T> {
 		assert ! idsInBatch.isEmpty();
 
 		final int numberOfIdsInBatch = idsInBatch.size();
+		if ( numberOfIdsInBatch == 1 ) {
+			return performSingleMultiLoad( idsInBatch.get( 0 ), lockOptions, session );
+		}
 
 		if ( log.isTraceEnabled() ) {
 			log.tracef( "#loadEntitiesById(`%s`, `%s`, ..)", entityDescriptor.getEntityName(), numberOfIdsInBatch );
@@ -331,6 +335,11 @@ public class MultiIdLoaderStandard<T> implements MultiIdEntityLoader<T> {
 				RowTransformerPassThruImpl.instance(),
 				ListResultsConsumer.UniqueSemantic.FILTER
 		);
+	}
+
+	private List<T> performSingleMultiLoad(Object id, LockOptions lockOptions, SharedSessionContractImplementor session) {
+		T loaded = (T) entityDescriptor.load( id, null, lockOptions, session );
+		return Collections.singletonList( loaded );
 	}
 
 	private List<T> performUnorderedMultiLoad(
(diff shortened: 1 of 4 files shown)


sebersole:
I am making an assumption here as to the original problem based on the solution.  Let me know if my assumption is incorrect.

IIUC the problem really comes down to the fact that you attempt to multi-load a single key, but internally we assume a multi-load will try to load, well, multiple - which is ultimately the point of `numberOfKeysToLoad`.  Internally we make a number of decisions based on `numberOfKeysToLoad`, part of which is to determine if we are processing a multi-load.  Multi-loading a single key (`numberOfKeysToLoad == 1`) confuses that logic and looks like any non-multi load use case.

If my assumption is correct, your solution will not work for composite keys - try writing another test that does the same thing, but using a composite key.  My guess is that that use case still fails.  Effectively, `numberOfKeysToLoad` describes the depth of an array, but you are making a decision based on the "width" of that array - hopefully that analogy makes sense.

Plus, your change "works" for your specific case, but it also affects all non-multi loading cases as well.

IMO there are 2 correct ways to handle this at a high level:

1) More clearly indicate to `LoaderSelectBuilder` that we have a multi-load - passing along a flag or somesuch.
2) Handle this multi-load, not-really-multi-load case specially

Especially important here, imo, is the missed opportunity to cache this JdbcSelect which will be the same each and every time.  In the single-key loaders e.g. we cache these because there we know it will always be the same.  

Not a huge fan of (1) for the same reason I am not a fan of your solution as-is - mainly its just not the right place to address this.

There are quite a few ways to accomplish (2).  

My first thought was to just delegate the loading here to the corresponding single-key loader.  That may or may not work based on the multi-load options.  I *think* it would be ok, but we'd have to verify.  Most of those options are already handled in the `MultiIdEntityLoader` / `MultiNaturalIdEntityLoader` objects, outside of `LoaderSelectBuilder`, which makes me think it would be fine.  Delegating to the corresponding single-key loader after applying those options should be the same effective result.  But again, we'd have to verify this. 

A similar variant of the above would be to have the `MultiIdEntityLoader` / `MultiNaturalIdEntityLoader` themselves make this "yep, we are multi-loading a single key" determination and use a lazily cached `JdbcSelect` for use when they are asked to load a single key.

Either way, IMO, `MultiIdEntityLoader` / `MultiNaturalIdEntityLoader` is the correct place to address this.

fax4ever:
Thanks @sebersole for the review.
I tried to address your suggestion.
Actually I don't know if I did it correctly.
I added tests for single instance load with natural id and composite id.
I'm here to address any other further changes.

hibernate-core/src/main/java/org/hibernate/loader/ast/internal/MultiIdLoaderStandard.java
```diff
@@ -91,6 +95,20 @@ public EntityMappingType getLoadable() {
 		}
 	}
 
+	private List<T> performSingleLoad(Object id, MultiIdLoadOptions loadOptions, SharedSessionContractImplementor session) {
```
sebersole, line 98: EntityPersister already caches the single-id loader.  Is there a reason to not just simply re-use it?

```
entityDescriptor.getIdLoader()...
```

Yes, at the moment it does not expose its `AbstractEntityPersister#singleIdEntityLoader`, but that was initially true for the other loaders which we exposed as needed.

You could even just call `EntityPersister#load` I think

sebersole:
> 

Not saying your change, as is, won't work.  But avoiding generation of a new loader here each and every time would be good.

You could even cache this one you generate here on the multi-load loader.  But since we already have the same loader built on EntityPersister, reusing that one seems like the best option.

fax4ever:
Thanks @sebersole for the comments.
I made some changes to fallback to single id load earlier, this should avoid the recreation of the loader.

fax4ever:
I'm going to add more commit to support a case where some entities are fetched from the second level caches and we need to load a sigle item on the same multiloading...

fax4ever:
Finally I applied what you suggest adding a new test to reproduce another issue that we used to have with Search.

sebersole (approved):
:+1:

fax4ever:
Thanks Steve!

PR #4449, opened 2021-12-12
slightly expand the documentation for Boolean

gavinking:


beikov (approved):

sebersole:
My concern with this (though not limited to just this change) is that we are getting back into a mix of "professional" and "conversational" voices.  One of the good things Vlad did was to unify this.

sebersole:
Also, we've been instructed to not use "JPA" - these refs should be changed to "Jakarta Persistence"

documentation/src/main/asciidoc/userguide/chapters/domain/basic_types.adoc
```diff
@@ -411,9 +412,19 @@ include::{sourcedir}/basic/BooleanMappingTests.java[tags=basic-boolean-example-i
...
 ====
 
-However, it is certainly reasonable to map booleans to other representations in the
-database and Hibernate provides support for mapping booleans to the more well known
-representations - 'Y' / 'N', 'T' / 'F' or 1 (true) / 0 (false).
+But, especially when working with a legacy database, it's quite common to find boolean values
+encoded as a character, or as an integer, even when the platform _does_ support a built-in
+`BOOLEAN` type.
+
+The easiest way to handle this situation is to use a JPA `AttributeConverter`. You can easily
```
sebersole, line 419: This should link to the AttributeConverter section

sebersole, line 419: "You" is an example of what I mean.

```
Such cases are exactly the intention of `AttributeConverter`.  See <<converter-section-id>>.  
For convenience, Hibernate provides 3 boolean-related converters:

- ...
```

documentation/src/main/asciidoc/userguide/chapters/domain/basic_types.adoc
```diff
@@ -411,9 +412,19 @@ include::{sourcedir}/basic/BooleanMappingTests.java[tags=basic-boolean-example-i
...
+write your own `AttributeConverter`, but for your convenience, Hibernate comes with three
+built-in converters for common cases:
+
+- `YesNoConverter` encodes a boolean value as `'Y'` or `'N'`,
+- `TrueFalseConverter` encodes a boolean value as `'T'` or `'F'`, and
+- `NumericBooleanConverter` encodes the value as an integer, `1` for true, and `0` for false.
+
+You may specify the converter to use with the `@Convert` annotation.
```
sebersole, line 427: Again "you", and here, this would already be covered by the link to the converter section

documentation/src/main/asciidoc/userguide/chapters/domain/basic_types.adoc
```diff
@@ -428,11 +439,8 @@ include::{sourcedir}/basic/BooleanMappingTests.java[tags=basic-boolean-example-e
 ----
 ====
 
-Here the built-in `AttributeConverter` implementations provided by Hibernate are used.  A custom
-`AttributeConverter` could be used also.
-
-
-Booleans can also be mapped using `UserType` and `@CustomType`
+Of course, if you need to do something really special, a boolean value can also be mapped using
```
sebersole, line 442: Personally I think my original wording here is better.  There is nothing special or not here - its a user choice

sebersole (changes requested):

sebersole:
> Also, we've been instructed to not use "JPA" - these refs should be changed to "Jakarta Persistence"

Or, in that particular section, just drop "JPA" / "Jakarta Persistence" completely.  It does not add anything, and like other comments in the section, linking to the converter section already covers this

sebersole:
The requested changes were not made and I ran out of time for Beta3

gavinking:
> The requested changes were not made and I ran out of time for Beta3

I'm gunna do it, but I just didn't find the time.

documentation/src/main/asciidoc/userguide/chapters/domain/basic_types.adoc
gavinking: The reason I changed it was because (if I'm completely naive) it reads like this is something special about `Boolean's. But it's not: _any_ type can be mapped using `UserType`. So then I as the naive reader is left wondering what it is that is special about booleans that makes it especially likely that I would want to use `UserType` for them. We don't really say.

So I guess the best thing is to explain exactly *why* booleans are special in this way: and I agree my change did a bad job of that.

We should explain properly that there are all kinds of different encodings that you might find in a legacy database, and that our built-in `AttributeConverters` can't hope to handle all of them. And so if you have something they can't handle, then use a `UserType`.

sebersole:
Applied upstream with the discussed edits

gavinking:
The final result looks good.

PR #4451, opened 2021-12-12
column types rework

gavinking:
DO NOT MERGE!


hibernate-core/src/main/java/org/hibernate/dialect/MariaDBDialect.java
```diff
@@ -34,35 +34,34 @@
...
+				info
 		);
-		this.version = version;
+		this.mariaVersion = mariaVersion;
 	}
 
 	@Override
 	public DatabaseVersion getVersion() {
-		return version;
+		return mariaVersion;
```
beikov, line 64: I believe this could become a problem because that will return null when the `columnType` methods are called from within the `super` constructor call.

gavinking: That's right, but the `mariaVersion` is never (or at least _should_ never be used for anything. It should always be `getMySQLVersion()` that is used. (A bit confusing, I know.)

gavinking:
All checks green. This is ready to merge, IMO.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -249,55 +264,181 @@ protected void registerDefaultColumnTypes() {
...
+	 * This method may be overridden by concrete {@code Dialect}s as an alternative
+	 * to calling {@link #registerColumnType(int,String)}.
+	 *
+	 * @param jdbcTypeCode a JDBC type code
+	 * @return a column type name, with $l, $p, $s placeholders for length, precision, scale
+	 *
+	 * @see SqlTypes
+	 * @see #getSupportedJdbcTypeCodes()
+	 */
+	protected String columnType(int jdbcTypeCode) {
```
beikov, line 362: Maybe it's not worth doing as part of this PR, but I would personally love to see an actual contract here instead of returning a `String` e.g.

```
interface DdlType {
    String getType(Size size);
}
```

This way, we could handle some of the strange types that use bit lengths instead of bytes. Apart from that, such a contract would be extendible in various ways that makes future features easier possible. Something that would be awesome to add is something like:

```
default boolean needsCheckConstraintForSize() {
  return false;
} 
default String generateCheckConstraint(Size size) {
  return null;
}
```

which we could use to allow emitting check constraints for LONGVARCHAR/LONGVARBINARY. Maybe it's even worth passing the `Column` here to allow access to more metadata and enable more use cases?

Apart from that, a previous idea that came to mind was to capture somehow which JDBC API to use for a DDL type e.g. CLOB vs. String APIs to be able to use a clob DDL type even if LONGVARCHAR was requested, so that we can fit the requested size.

hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
```diff
@@ -100,16 +103,11 @@ public MySQLDialect() {
 	}
 
 	public MySQLDialect(DatabaseVersion version) {
-		// Let's be conservative and assume people use a 4 byte character set
-		this( version, 4 );
+		this(version, null);
 	}
 
-	public MySQLDialect(DatabaseVersion version, int characterSetBytesPerCharacter) {
```
beikov, line 107: IMO we should retain this constructor to allow subclasses provide a value to avoid access to the database on boot. The invocation order due to subclassing seems to become a problem here though. How about expecting subclass constructors to invoke some `finishInitialization` instead to avoid some of the potential issues?

hibernate-core/src/main/java/org/hibernate/dialect/MariaDBDialect.java
beikov: Let's continue this here: https://github.com/hibernate/hibernate-orm/pull/4451#discussion_r767914434

hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
gavinking: Well took that approach first, and I didn't like it. Neither approach is quite perfect but I settled on this design as the cleanest option.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
gavinking: Well I'm definitely not going to do anything additional in this PR, which is already more than big enough.

PR #4455, opened 2021-12-13
fix broken float type precision conversion

gavinking:
it is not the case that ln(10) = log_2(10)

@beikov

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -161,6 +161,8 @@ import org.hibernate.type.descriptor.jdbc.spi.JdbcTypeRegistry;
 
 import jakarta.persistence.TemporalType;
 
+import static java.lang.Math.ceil;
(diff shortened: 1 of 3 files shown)


beikov (approved):
Looks good to me. Yeah I apparently didn't think this totally through before, so thanks for making sure this is correct :)

gavinking:
But WTH is up with those HSQLDB failures, surely they can't be related to this change?

beikov:
I'd say don't mind these failures. These seem related to some DST changes. Did DST change 3 hours ago in the TZs we test for? :D 
FWIW, I would love to get rid of the HSQL dialect from hibernate-core, but that's a different discussion.

gavinking:
> I'd say don't mind these failures.

Yeah, agreed. Just ran the tests on a local hsqldb and nothing failed.

> Did DST change 3 hours ago in the TZs we test for? :D

Well the same thought occurred to me, but Dec 13 seems like a strange date to change clocks....

gavinking:
Screw it, I'm merging, `main` was already failing.

PR #4456, opened 2021-12-13
Wip/6.0 polymorphic set

dreab8:


hibernate-core/src/test/java/org/hibernate/orm/test/bytecode/enhancement/lazy/proxy/LazyToOnesProxyWithSubclassesTest.java
```diff
@@ -281,8 +281,6 @@ public void cleanUpTestData() {
 		inTransaction(
 				session -> {
 					session.createQuery( "delete from OtherEntity" ).executeUpdate();
-					session.createQuery( "delete from Human" ).executeUpdate();
-					session.createQuery( "delete from Primate" ).executeUpdate();
```
beikov, line 285: What is this for?

beikov (commented):
The overall change looks ok to me but I'm having a hard time understanding why/when an initializer is called multiple times. Can you elaborate a bit on the case?

beikov (commented):
The overall change looks ok to me but I'm having a hard time understanding why/when an initializer is called multiple times. Can you elaborate a bit on the case?

dreab8:
The assembler can call the initializer, one example is https://github.com/hibernate/hibernate-orm/pull/4456/files#diff-c562e6899ef6d0ae3271173cc4dfc2c54df65de5b10737ce46c19e118c46eac3R258 and also I think for `EmbeddableId` with `toOne`associations.

hibernate-core/src/test/java/org/hibernate/orm/test/bytecode/enhancement/lazy/proxy/LazyToOnesProxyWithSubclassesTest.java
dreab8: I have an annoying failure when I run the tests locally, the issue with temporary tables during deletion, I made a tried and forget to revert this change for the PR

beikov: I found the issue and fixed it as part of https://github.com/hibernate/hibernate-orm/pull/4425

PR #4460, opened 2021-12-13
HHH-14948 Adapt contributed patch to 6.0 branch

Sanne:
This particular test is disabled as it's located in `test_legacy` .. it's also not compiling but it's a 1:1 port of the contribution already applied on 5.6 so I'd first align these to not forget.

I'll try to conver the test soon with @dreab8


Sanne:
fixed the test, let's have CI test it again

PR #4461, opened 2021-12-13
add hsqldb to local test options

gavinking:
because we have it on the CI build

### DIFF
--- a/databases/hsqldb/matrix.gradle
+++ b/databases/hsqldb/matrix.gradle
@@ -1,0 +1,9 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+//databaseProfile {
+	jdbcDependency 'org.hsqldb:hsqldb:2.3.6'
+
--- a/databases/hsqldb/resources/hibernate.properties
+++ b/databases/hsqldb/resources/hibernate.properties
@@ -1,0 +1,25 @@
+#
+# Hibernate, Relational Persistence for Idiomatic Java
+#
+# License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+# See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+#
+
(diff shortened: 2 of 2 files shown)


beikov:
This shouldn't be necessary, except if you use the matrix plugin, which AFAIU is considered "deprecated". You should be selecting the database by using the `-Pdb=...` gradle parameter. Available databases are listed out in `databases.gradle`

gavinking:
Oh really, OK, then there's something I'm not quite understanding here....

...hold on.

Sanne:
To be fair I'm also confused with the many ways we have to run tests on various DBs nowadays. Would be nice if someone could cleanup the various "deprecated" ways and provide some guidance in the readme.

gavinking:
OK, you're right of course, I was just not understanding how this stuff worked.

Yeah @Sanne, a decent README is sorely needed.

beikov:
Maybe @sebersole can help here? I don't know who it is that uses the matrix plugin, but my understanding is, this is needed for the QA team.

gavinking:
What I mean is that we don't have good instructions for:

- how to set up your local databases so that `-Pdb=mysql` (for example) works (you need to create a database and user, grant permissions, etc)
- how to run the tests on Docker (podman doesn't help those of us on Mac)

gavinking:
This is the sort of thing I mean:

<https://github.com/hibernate/hibernate-reactive#running-tests>

beikov:
Well, we have that https://github.com/hibernate/hibernate-orm#using-profiles

but I acknowledge that we currently don't explain how to setup podman/docker. The DB and schema setup is depending on the database, sometimes not so easy, but if you want to setup that stuff locally, the docker_db.sh script should give you enough information what is required.

IMO we should strive to provide good setup instructions for podman/docker on the various OSes. Wdyt?

gavinking:
> Well, we have that https://github.com/hibernate/hibernate-orm#using-profiles

And that document doesn't say anything at all about how to set up my local databases. (Have a look at how I did this in the HR README.)

beikov:
>And that document doesn't say anything at all about how to set up my local databases. (Have a look at how I did this in the HR README.)

Well, this is why:

>The DB and schema setup is depending on the database, sometimes not so easy, but if you want to setup that stuff locally, the docker_db.sh script should give you enough information what is required.

We can surely copy all the steps out of the docker_db.sh file into the readme, but I believe this would just bit-rot and become inconsistent. Would this help you though?

PR #4462, opened 2021-12-13
Fix issue with ResultSetMapping resolution

dreab8:


hibernate-core/src/test/java/org/hibernate/test/jpa/procedure/DateTimeParameterTest.java
```diff
@@ -17,48 +17,51 @@
...
+import jakarta.persistence.TemporalType;
+
 import static org.junit.Assert.assertEquals;
 
 /**
  * @author Steve Ebersole
  */
-public class DateTimeParameterTest extends BaseUnitTestCase {
-	HibernateEntityManagerFactory entityManagerFactory;
+@RequiresDialect(DerbyTenSevenDialect.class)
```
beikov, line 51: AFAICT this can't ever be executed because we don't instantiate these dialect subclasses anymore :)

hibernate-core/src/test/java/org/hibernate/test/jpa/procedure/JpaTckUsageTest.java
```diff
@@ -46,11 +45,17 @@
  *
  * @author Steve Ebersole
  */
-public class JpaTckUsageTest extends BaseUnitTestCase {
+@RequiresDialect(DerbyTenSevenDialect.class)
```
beikov, line 48: AFAICT this can't ever be executed because we don't instantiate these dialect subclasses anymore :)

hibernate-core/src/test/java/org/hibernate/test/jpa/procedure/DateTimeParameterTest.java
dreab8: it is a test that was in the `test_legacy` package, I just made it compilable but it is still disabled, I'm working on it right now :)

PR #4465, opened 2021-12-14
Make TableReference an interface with various sub-classes

beikov:
Split TableReference class into interface and NamedTableReference implementation to allow QueryPartTableReference and ValuesTableReference to fit into the picture


sebersole:
Resolved the conflicts and applied upstream.  

All tests pass, but the conflicts in AbstractEntityPersister and AbstractCollectionPersister were a bit complicated so please double check

PR #1850, opened 2017-03-16
[6.0] HHH-11573 Extended Envers Query API to create type expressions

ffeisst:
Extended the Envers Query API to restrict an alias to a specified type.

[Link to Jira issue](https://hibernate.atlassian.net/browse/HHH-11573)

PR #4471, opened 2021-12-14
HHH-11573 Support restricting an alias to specified type

Naros:
https://hibernate.atlassian.net/browse/HHH-11573

This PR supersedes #1850 and is based on `main` branch for ORM 6.


Naros:
So investigation seems to indicate there is some problems with how the discriminator mappings get built when a composite PK in conjunction with expected table joins that get rendered in 5.6 but aren't present in the 6.0 flavor.  I've reached out to @beikov with details so we can try and resolve the problems with `type( entityAlias )` rendering.

Naros:
Depends on PR #4482

PR #4476, opened 2021-12-15
clean up the Dialect-specific default property handling

gavinking:


sebersole (commented):
Not particularly a fan of still handling these via properties rather than the nice API you just added.  Personally I'd just drop the whole "default properties".  

Sure its "more flexible" in terms of allowing individual Dialects to set additional defaults.  But imo using properties for this is part has always been part of the problem

gavinking:
> Not particularly a fan of still handling these via properties rather than the nice API you just added. Personally I'd just drop the whole "default properties".

Yeah, but I just wanted to do this in baby steps. I don't have time to do more than this right now. But it lays the foundation for completely eliminating the properties as a next step.

sebersole:
> > Not particularly a fan of still handling these via properties rather than the nice API you just added. Personally I'd just drop the whole "default properties".
> 
> Yeah, but I just wanted to do this in baby steps. I don't have time to do more than this right now. But it lays the foundation for completely eliminating the properties as a next step.

Fair enough

hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/IngresDialect.java
```diff
@@ -144,6 +143,30 @@ public IngresDialect(DatabaseVersion version) {
...
+//	protected void initDefaultProperties() {
+//		super.initDefaultProperties();
+//
+//		if ( getVersion().isBefore( 10 ) ) {
+//			// There is no support for a native boolean type that accepts values
+//			// of true, false or unknown. Using the tinyint type requires
+//			// substitutions of true and false.
+//			getDefaultProperties().setProperty( Environment.QUERY_SUBSTITUTIONS, "true=1,false=0" );
+//		}
+//	}
```
beikov, line 166: Why was this not commented back in? Not needed anymore or was that by accident?

gavinking: Well it *shouldnt* be necessary, as far as I can see. But of course I have no way on earth to test that, so I left the commented code there just in case someone comes back and tells us removing it broke something.

beikov: Right, it's not even used anymore AFAICS. How about removing the property altogether?

gavinking: Yes, we should. That one was already on my list.

PR #1966, opened 2017-07-31
HHH-11895 Support traversal of components in audit queries

ffeisst:
This PR adds the possibility to query for properties on components in the envers query API by using the traverseRelation(..) methods.

This PR is based on the open PRs for HHH-11452 ([1849](https://github.com/hibernate/hibernate-orm/pull/1849)) and HHH-11735 ([1904](https://github.com/hibernate/hibernate-orm/pull/1904)). That means that the only the last commit is a new contribution and the other commits are part of the other PRs.

Link to Jira Issue: [HHH-11895](https://hibernate.atlassian.net/browse/HHH-11895)

PR #4479, opened 2021-12-16
HHH-11895 Support traversal of components in audit query API

Naros:
https://hibernate.atlassian.net/browse/HHH-11895

This PR supersedes PR #1966 by being based on ORM 6 `main` branch.


Naros:
Many of the test failures are due to null-value precedence.  I've logged a new issue and opened PR #4481 to add support so that order-by columns with nulls can have deterministic behavior across dialects.

PR #4482, opened 2021-12-16
Fix resolving table references for discriminator path

beikov:


Naros:
Thanks for working on this @beikov.  Is there anything else needed besides me confirming this solves all the needs for the type-based audit queries we're adding in 6?

Naros (approved):
@beikov this fixed all my issues with the `type(alias)` support; safe to merge?

PR #4484, opened 2021-12-16
treat integral types as equivalent for purposes of schema validation

gavinking:
... and update.


gavinking:
@sebersole shall I merge it?

beikov:
Why would we want this? If you have a model:

```
@Entity
public class MyEntity {
    Short value;
}
```

and change it to

```
@Entity
public class MyEntity {
    Integer value;
}
```

hbm2ddl would not report the change.

gavinking:
See the discussion on Zulip. It's a stopgap measure to help with the fact that something changed in how we map enum types.

sebersole (commented):
It looks fine as a stop-gap, as noted

PR #4486, opened 2021-12-17
attempt to fix long-standing bug with collections of a subclass type

gavinking:
- automatically include discriminator in join in HQL query
- will also do the same for filters and @Where conditions

For reference:

1. https://github.com/eclipse-ee4j/jpa-api/issues/39
2. https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/topic/'discriminated'.20collection.20mappings

I'm not sure this fix is perfectly correct, please review carefully.


gavinking:
A bit of further explanation:

- currently, filters, `@Where` conditions, and the discriminator are automatically included in the SQL `on` clause for a HQL `fetch join`, but
- none of them were getting included for an `inner join`.

This change calls ` FilterHelper.createFilterPredicate()` from `consumeAttributeJoin()` from the branch that deals with a `PluralPersistentAttribute`, so that those additional conditions get added in.

sebersole:
I have basically the same code in my working copy improving `@Where` and `@Filter` handling.  Going to close this so to not step on the toes of that much larger work.

As we discussed, I did add your test.

PR #4488, opened 2021-12-17
HHH-14983: clear second-level cache before executing an HQL update

nikolabekcic:
… to prevent loading stale cache entries (in the same session) upon executing the HQL update query. Changed AbstractReadWriteAccess by removing the no-op removeAll method so that  "affected cache regions" can be cleared before executing an HQL, ensuring strong data consistency.

Merged provided test case into existing ReadWriteCacheTest. Performed code polish by moving book creation and saving into a @Before method as it's needed by all tests. Renamed helper test methods to reveal intention in a better way.

https://hibernate.atlassian.net/browse/HHH-14983

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/spi/access/CachedDomainDataAccess.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/spi/access/CachedDomainDataAccess.java
@@ -143,7 +143,7 @@ public interface CachedDomainDataAccess {
 	 * Remove all data for this accessed type
 	 *
 	 * @throws CacheException Propagated from underlying cache provider
-	 * @param session
+	 * @param session Current session.
 	 */
 	void removeAll(SharedSessionContractImplementor session);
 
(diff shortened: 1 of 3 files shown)


gavinking:
Looks to me like this fix is likely correct, and should be integrated.

nikolabekcic:
Thank you for taking the time to review this, @gavinking. 

What are the next steps to merge this patch?

beikov:
I just looked into this as well, because I had recently investigated a different issue that touched the same code. `org.hibernate.cache.spi.support.AbstractCachedDomainDataAccess#unlockRegion` is called after a transaction completes, so you shouldn't ever read stale data in other transactions, but clearly, your test shows a bug since you get stale data within the same transaction. Your change will effectively cause an immediate cache clear which seems necessary to avoid reading stale data within that same transaction.

Can you please rebase your PR on top of the current state of the `main` branch? Then we can run tests again and merge when successful.

gavinking:
Actually, @beikov, after your comment, now I'm questioning whether this is actually the right fix.

Perhaps the problem is really that  `lockRegion()` is a noop and doesn't soft-lock the region. If we just soft locked the region, then we would force readers to the database and not need to actually evict the cache until the tx completes successfully.

WDYT? @sebersole WDYT?

nikolabekcic:
@sebersole - I’m following up regarding the move to the Apache Software License v2. I’ve been tied up with other work recently, and with the holiday season, I couldn't respond sooner.

I will update the subject with the appropriate licensing text so we can proceed with merging the patch and resolving the issue.

Best regards,

PR #4491, opened 2021-12-17
Use JDK11 with LGTM check

Naros:


Naros:
So I've tried everything that I've been able to deduce from the documentation but LGTM check continues to fail that JDK11 isn't being used during the extraction phase and running of Gradle.

sebersole:
As I was mentioning on Zulip, from what @beikov told me, these jobs pick up the job configs for PR actions from the base branch - the PR branch cannot change those.

I cleaned up your PR a bit - mainly I removed the custom build_command as there are a few references that it is not needed by LGTM.  I pushed mine upstream - let's see if that helps

sebersole:
Applied slightly modified form upstream

PR #4496, opened 2021-12-19
HHH-14725 Fix reset handling on BlobProxy

Naros:
https://hibernate.atlassian.net/browse/HHH-14725

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/BlobProxy.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/BlobProxy.java
@@ -31,6 +31,8 @@ import org.hibernate.type.descriptor.java.DataHelper;
 public final class BlobProxy implements Blob, BlobImplementer {
 
 	private final BinaryStream binaryStream;
+	private final int markBytes;
+	private boolean resetAllowed;
 	private boolean needsReset;
 
 	/**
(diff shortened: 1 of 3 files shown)


beikov (approved):
LGTM

beikov:
How about wrapping an `InputStream` to make sure closing only happens at the end of a TX when we do resource cleanup in Hibernate? If we see that the `InputStream` doesn't support marking, we could read the contents first into a `byte[]` and then serve data from there, or alternatively, write into a temporary file and serve data from that.

oppahansi:
Is there maybe an ETA when this will be merged?
We have encountered an issue with exactly this problem recently.

Our workaround for now is:
We `refresh` the `Entity` with the `Blob` using the `EntityManager`  **before** operating on the `Blob's InputStream/Bytes`.

Naros:
> How about wrapping an `InputStream` to make sure closing only happens at the end of a TX when we do resource cleanup in Hibernate? If we see that the `InputStream` doesn't support marking, we could read the contents first into a `byte[]` and then serve data from there, or alternatively, write into a temporary file and serve data from that.

Hi @beikov do we have another proxy type that does this so I can use that as a reference guide?

beikov:
It's fine, let's merge this as soon as tests succeed.

beikov:
The test seems to fail @Naros

gavinking:
@Naros What is the status of this one?

dreab8:
Closing this one in favour of https://github.com/hibernate/hibernate-orm/pull/9402 where I rebased the PR and fixed a problem with the test.

PR #4500, opened 2021-12-20
Upgrade to DB2 container image ibmcom/db2:11.5.7.0

Sanne:
Also setting environment variable `PERSISTENT_HOME=false` as it helps a little with startup times.

### DIFF
--- a/docker_db.sh
+++ b/docker_db.sh
@@ -89,7 +89,7 @@ edb() {
 
 db2() {
     docker rm -f db2 || true
-    docker run --name db2 --privileged -e DB2INSTANCE=orm_test -e DB2INST1_PASSWORD=orm_test -e DBNAME=orm_test -e LICENSE=accept -e AUTOCONFIG=false -e ARCHIVE_LOGS=false -e TO_CREATE_SAMPLEDB=false -e REPODB=false -p 50000:50000 -d ibmcom/db2:11.5.5.0
+    docker run --name db2 --privileged -e DB2INSTANCE=orm_test -e DB2INST1_PASSWORD=orm_test -e DBNAME=orm_test -e LICENSE=accept -e AUTOCONFIG=false -e ARCHIVE_LOGS=false -e TO_CREATE_SAMPLEDB=false -e REPODB=false -p 50000:50000 -d ibmcom/db2:11.5.7.0
     # Give the container some time to start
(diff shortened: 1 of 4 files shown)


Sanne:
Added a commit to upgrade the JCC driver as well

beikov (commented):
I'd be ok with the upgrade but it seems the container fails to start?

Sanne:
That's annoying, it works fine for me locally and I don't understand the error. Wondering if it's a timeout - but even so we wouldn't want to upgrade to an image that will most likely have frequent timeout issues.

Do we have some way to cache the container image between subsequent runs on github actions?

docker_db.sh
```diff
@@ -89,7 +89,7 @@ edb() {
 
 db2() {
     docker rm -f db2 || true
-    docker run --name db2 --privileged -e DB2INSTANCE=orm_test -e DB2INST1_PASSWORD=orm_test -e DBNAME=orm_test -e LICENSE=accept -e AUTOCONFIG=false -e ARCHIVE_LOGS=false -e TO_CREATE_SAMPLEDB=false -e REPODB=false -p 50000:50000 -d ibmcom/db2:11.5.5.0
+    docker run --name db2 --privileged -e DB2INSTANCE=orm_test -e DB2INST1_PASSWORD=orm_test -e DBNAME=orm_test -e LICENSE=accept -e PERSISTENT_HOME=false -e AUTOCONFIG=false -e ARCHIVE_LOGS=false -e TO_CREATE_SAMPLEDB=false -e REPODB=false -p 50000:50000 -d ibmcom/db2:11.5.7.0
```
beikov, line 92: According to https://hub.docker.com/r/ibmcom/db2 `PERSISTENT_HOME` should only be specified on windows. Maybe that's the issue?

Sanne: I've tested PERSISTENT_HOME=false locally (on linux) and all it does is skip the mount points... locally it helps as then the filesystem is cached in memory.

But sure we can try..

Sanne: pushed again with that change omitted

beikov: Seems to work and it is really fast now :)

Sanne: OK but it's suspicious... before merging I'll run it a couple more times to ensure it's not flaky

PR #4502, opened 2021-12-20
Fix instantiation of composite property ref

dreab8:


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EmbeddedAttributeMapping.java
```diff
@@ -406,6 +406,7 @@ private static PropertyAccess getPropertyAccess(
 					true );
 		}
 		else {
+//			throw new NotYetImplementedFor6Exception();
```
beikov, line 409: Unnecessary

beikov:
Looks good overall, but AFAIU these changes are necessary for the `EmbeddableInitializer` to receive the fetch parent access, so that it can inject the state to it's "virtual instance" which is the entity itself, right? I guess we will need access to the fetch parent access anyway even if we model virtual embeddables differently, so I'm fine with the change.

dreab8:
@beikov yeap we need access to the parent  to inject the state to it's "virtual instance" which is the entity itself

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EmbeddedAttributeMapping.java
dreab8: yeap an oversight :) I'm going to delete it thanks

PR #4505, opened 2021-12-22
Re-enabled additional test

dreab8:


hibernate-core/src/test/java/org/hibernate/orm/test/annotations/embeddedid/Dvd.java
```diff
@@ -27,6 +28,7 @@
 	@GeneratedValue(generator = "custom-id")
 	@GenericGenerator(name = "custom-id", strategy = "org.hibernate.test.annotations.type.MyOidGenerator")
 //	@Type(type = "org.hibernate.test.annotations.type.MyOidType")
+	@EmbeddedId
```
beikov, line 31: IMO this should switch to using `@CustomType` to match what the behavior was before.

PR #4510, opened 2021-12-22
Set correct key type if !referenceToPrimaryKey

fax4ever:
wip/6.0 remaining tests: Try to fix stuff for `NaturalIdCachingTest.java`


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -624,6 +627,10 @@ public String getReferencedPropertyName() {
 		return referencedPropertyName;
 	}
 
+	public boolean isReferenceToPrimaryKey() {
+		return referenceToPrimaryKey;
```
beikov, line 631: Wouldn't it be enough to use `referencedPropertyName == null`? AFAIU this should always be null in case the primary key is referenced.

fax4ever: I'm trying it

fax4ever: From my tests you're right. Fixed it. Thanks!

beikov (approved):

beikov:
Thanks!

PR #4516, opened 2021-12-24
proposed fixes to Query API

gavinking:
The main goal of this work is to alleviate the problem of warning in the client program resulting from invocations of raw `Query`s returned by Hibernate `Session` methods.

This is a difficult problem due to the requirements of backward compatibility with ancient versions of Hibernate, and due to the inheritance from `EntityManager` which still uses raw types.

As discussed, I have:

- Preferred the use of `@SuppressWarnings("rawtypes")` to wildcards `<?>` in methods of `Query`.
- `@Deprecated` `createXxxxQuery()` methods which accept no class.
- improved the `createNativeQuery()` methods with additional overloads, and the ability to accept a non-entity result class.

This solution is imperfect, but it's at least an improvement.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/SynchronizeableQuery.java
+++ b/hibernate-core/src/main/java/org/hibernate/SynchronizeableQuery.java
@@ -46,8 +46,8 @@ public interface SynchronizeableQuery<T> {
 	 */
 	default SynchronizeableQuery<T> addSynchronizedQuerySpace(String... querySpaces) {
 		if ( querySpaces != null ) {
-			for ( int i = 0; i < querySpaces.length; i++ ) {
-				addSynchronizedQuerySpace( querySpaces[i] );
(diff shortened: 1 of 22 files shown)


hibernate-core/src/main/java/org/hibernate/SynchronizeableQuery.java
```diff
@@ -46,8 +46,8 @@
 	 */
 	default SynchronizeableQuery<T> addSynchronizedQuerySpace(String... querySpaces) {
 		if ( querySpaces != null ) {
-			for ( int i = 0; i < querySpaces.length; i++ ) {
-				addSynchronizedQuerySpace( querySpaces[i] );
+			for (String querySpace : querySpaces) {
```
sebersole, line 49: We prefer the form you changed from.  This form can have performance impact and for no real gain

PR #4521, opened 2021-12-25
Miscellaneous cleanups

gavinking:
Cleanups to Session-associated interfaces.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/SessionFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/SessionFactory.java
@@ -314,31 +314,4 @@ public interface SessionFactory extends EntityManagerFactory, HibernateEntityMan
 	 */
 	@Deprecated
 	CollectionMetadata getCollectionMetadata(String roleName);
-
-	/**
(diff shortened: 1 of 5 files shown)


gavinking:
These changes are surely uncontentious so I'm just going to merge this one.

hibernate-core/src/main/java/org/hibernate/SessionFactory.java
```diff
@@ -285,7 +282,7 @@ default void inTransaction(Consumer<Session> action) {
 	 * @deprecated Use the descriptors from {@link #getMetamodel()} instead
 	 */
 	@Deprecated
-	ClassMetadata getClassMetadata(Class entityClass);
+	ClassMetadata getClassMetadata(@SuppressWarnings("rawtypes") Class entityClass);
```
beikov, line 285: Stupid question, why not remove this and other deprecated methods?

sebersole: Sorry, missed that.  Its been integrated, so we should just come behind and make that change

PR #4522, opened 2021-12-25
proposed changes to SessionBuilder

gavinking:
Here's one option for getting rid of the raw types.


beikov (commented):
Looks ok to me.

gavinking:
Steve fixed this independently.

PR #4524, opened 2021-12-27
HHH-14985 H2Dialect does not work properly with h2 2.0.202 on inserts

froque:
This should be cherry picked into Hibernate 5.x

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/identity/H2IdentityColumnSupport.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/identity/H2IdentityColumnSupport.java
@@ -28,6 +28,6 @@ public class H2IdentityColumnSupport extends IdentityColumnSupportImpl {
 
 	@Override
 	public String getIdentityInsertString() {
-		return "null";
+		return "default";
 	}
 }


gavinking:
From which version of h2 is this change necessary?

froque:
works with H2 1.3.176

gavinking:
What i'm saying is that it should be written so that it checks the version number and adapts to the version of h2.

vxel:
The change is necessary for all H2 2.0.x versions, i.e. starting at H2 2.0.202
See : 
https://github.com/h2database/h2database/issues/3231

gavinking:
OK, great, then would you please make the change to this PR to reflect that?

Do you understand what I'm asking?

katzyn:
`DEFAULT` should be used for all versions of H2 unconditionally, it works with all versions. Even H2 1.0 (2006-12-17) supports it well.

Attempt to insert `NULL` into column with `NOT NULL` constraint is wrong by itself. H2 1.X accepted it for identity columns only, but H2 2.X follows the Standard here in default configuration and doesn't allow it any more.

gavinking:
> `DEFAULT` should be used for all versions of H2 unconditionally

Ah, OK, then that's different.

hibernate-core/src/test/java/org/hibernate/orm/test/dialect/H2DialectTestCase.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.hibernate.testing.RequiresDialect;
+import org.hibernate.testing.TestForIssue;
+import org.junit.Test;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
+import static org.junit.Assert.assertEquals;
+
+@RequiresDialect(H2Dialect.class)
+public class H2DialectTestCase extends BaseEntityManagerFunctionalTestCase {
```
gavinking, line 27: I don't understand why we need a new test.

Hibernate already has oodles of tests for `GenerationType.IDENTITY`. What is new/different about this one?

katzyn:
Just for information: `getIdentitySelectString()` isn't compatible with H2 2.0 too and there is nothing to return for 2.0.202, 2.0.204 and later versions. 1.4.200 was the last version where current implementation can be used (with caution, because `IDENTITY()` isn't a reliable function, for example, it may return value from another table if trigger performed some operations.)

In Regular mode H2 2.0 provides only two normal ways to query what was actually inserted:
1. Standard data change delta tables: `SELECT ID FROM FINAL TABLE (INSERT INTO TEST(C1, C2) VALUES (1, 2))`
2. `Statement.getGeneratedKeys()`, if JDBC is used.

Few compatibility modes provide additional legacy functions for identity columns.

It looks like `GetGeneratedKeysDelegate` can be used for 2.0.202 and newer versions and perhaps it will be better to use it since H2 1.4.197, but it needs to be tested.

In 1.4.196 and older versions `getGeneratedKeys()` was implemented with `SELECT SCOPE_IDENTITY()` by itself, it means result set doesn't know real name of the column. But `SCOPE_IDENTITY()` at least doesn't return unexpected values from triggers.

hibernate-core/src/test/java/org/hibernate/orm/test/dialect/H2DialectTestCase.java
froque: Nothing special about this test class. 

I just did not see any test for H2 in `org.hibernate.orm.test.dialect` package and wrote one as requested in [CONTRIBUTING.md](https://github.com/hibernate/hibernate-orm/blob/main/CONTRIBUTING.md)

Should I remove this test ? 

Any other changes to be made ?

gavinking: Yes, please remove the test.

froque: removed

gavinking: Thanks.

beikov (approved):
>Just for information: `getIdentitySelectString()` isn't compatible with H2 2.0

If `java.sql.DatabaseMetaData#supportsGetGeneratedKeys` returns `true`, which it does, it will use JDBC `getGeneratedKeys`, so this should be fine.

Changes look good. IMO we can merge this.

gavinking:
@beikov but ... perhaps we should do this instead:

https://github.com/hibernate/hibernate-orm/pull/4569

I'm not sure.

beikov:
I think `supportsInsertSelectIdentity` should be conditional based on the dialect version as I think the `final table` syntax is only supported as of 1.4.200. Is that correct @katzyn ?

gavinking:
> I think `supportsInsertSelectIdentity` should be conditional based on the dialect version as I think the `final table` syntax is only supported as of 1.4.200. Is that correct @katzyn ?

Right, could be. But it's also unclear if it's actually better than using `getGeneratedKeys()`, so Im really not clear on anything here.

gavinking:
Also, I mean, the Hibernate test suite doesn't run at all on H2. *Many* test failures. No idea if lots of things are broken or just one big thing.

katzyn:
> as I think the `final table` syntax is only supported as of 1.4.200

Data change delta tables are supported since H2 1.4.200, but they have some issues in this outdated version. They were resolved in H2 2.0.202.

`Statement.getGeneratedKeys()` uses the same collector as `FINAL TABLE`, but it is currently optimized a little bit better, because it collects values only from requested columns (data change delta tables collect complete rows).

katzyn:
> No idea if lots of things are broken or just one big thing.

Possibly many. `INFORMATION_SCHEMA` in H2 2.0 is compliant with the SQL Standard as it should in every DBMS, but it is very different from what H2 1.x has. Some non-standard additional tables were also modified to represent complete correct information.

It looks like Hibernate can't find indexes any more, it needs to use different queries for H2 2.0 and 1.*.

gavinking (approved):

beikov:
Thanks!

PR #4527, opened 2021-12-27
make @Generated default to INSERT

gavinking:
Currently there was no default value for this annotation member.

I think `INSERT` is a fantastic default value.


gavinking:
Actually I retract this proposal. I think that we should do is make `@ColumnDefault` and `@ColumnGeneratedAlways` imply an appropriate value for `@Generated`.

PR #4529, opened 2021-12-28
HQL documentation rewrite/restructure

gavinking:
- rewrite parts of the section dealing with the Query API + execution
- split out a new chapter about the query language
- remove usage of `java.sql.Timestamp` and `TemporalType`

Look, I don't like to sound harsh, but this section of the docs was simply bad.

1. It makes a huge upfront big deal over the differences between Hibernate and JPA to the point of actually _exaggerating_ the differences.
2. It offered minute-detail coverage of old and/or not-terribly-useful operations like `scroll()` and `stream()`, and old-fashioned versions of `setParameter()` which have not been useful for a long time and which we've now deprecated.
3. It was missing _really basic information_ like `setMaxResults()`.
4. Far, far down the page it actually starts describing the query language. I have not even got up to looking at that bit yet, but now I'm scared.

I really don't understand how anyone is supposed to learn HQL from this. It's in every way totally hostile to a non-expert reader.


gavinking:
> Far, far down the page it actually starts describing the query language. I have not even got up to looking at that bit yet, but now I'm scared.

I mean, just to begin with, discussion of the query language itself is cluttered with `Query` API calls and quotes and string manipulation. I've never seen anyone ever define a language by showing what it looks like when _embedded in a different language_, and, well, perhaps it's OK, but I'm skeptical. I need to decide what to do about that.

sebersole:
Am I understanding correctly that there is now a chapter `HQL.adoc` which describes the `Query` API and this new `QueryLanguage.adoc` chapter that discusses HQL?

documentation/src/main/java/org/hibernate/userguide/model/Person.java
```diff
@@ -42,155 +43,146 @@
...
-        hints = {
-            @QueryHint(
-                name = "org.hibernate.readOnly",
-                value = "true"
-            )
+//tag::sql-scalar-NamedNativeQuery-example[]
+@NamedNativeQuery(
+    name = "find_person_name",
+    query =
+        "SELECT name " +
```
sebersole, line 50: (Not a fan of capitalizing the keywords)*

sebersole (commented):
Still looking through the actual content, but some initial comments

gavinking:
> Am I understanding correctly that there is now a chapter `HQL.adoc` which describes the `Query` API and this new `QueryLanguage.adoc` chapter that discusses HQL?

Yup.

documentation/src/main/java/org/hibernate/userguide/model/Person.java
gavinking: Neither am I, but that's not new. That's just how it was.

sebersole:
> > Am I understanding correctly that there is now a chapter `HQL.adoc` which describes the `Query` API and this new `QueryLanguage.adoc` chapter that discusses HQL?
> 
> Yup.

I mean clearly that is not great.  The chapter on the Query API should be named `Query.adoc` or somesuch - not `HQL.adoc`.  `QueryLanguage.adoc` is fine, though my point was that you could have split the Query content out into a new `Query.adoc` chapter and left `HQL.adoc` discussing, well, HQL

documentation/src/main/java/org/hibernate/userguide/model/Person.java
sebersole: Sure, but its stuff you explicitly changed already.  Why not just fix that at the same time?

gavinking:
> I mean clearly that is not great. The chapter on the Query API should be named `Query.adoc` or somesuch - not `HQL.adoc`. `QueryLanguage.adoc` is fine, though my point was that you could have split the Query content out into a new `Query.adoc` chapter and left `HQL.adoc` discussing, well, HQL

Sure, that would make more sense. `QueryLanguage.adoc` is just a temporary placeholder. I was going to ask others what they wanted to call the files. I have no particular opinion.

documentation/src/main/java/org/hibernate/userguide/model/Person.java
gavinking: Well I think it's an all-or-nothing thing. There's actually rather a lot of example code there. I could go through and lowercase everything, but it will take me some time I don't really have _right now_. I can do it later when I go have time, or someone else can do it later. Here I was much more focussed one the organization/flow of the text.

sebersole:
> > I mean clearly that is not great. The chapter on the Query API should be named `Query.adoc` or somesuch - not `HQL.adoc`. `QueryLanguage.adoc` is fine, though my point was that you could have split the Query content out into a new `Query.adoc` chapter and left `HQL.adoc` discussing, well, HQL
> 
> Sure, that would make more sense. `QueryLanguage.adoc` is just a temporary placeholder. I was going to ask others what they wanted to call the files. I have no particular opinion.

`QueryLanguage.adoc` and `Query.adoc` (or maybe `QueryApi.adoc`) imo

gavinking:
> `QueryLanguage.adoc` and `Query.adoc` (or maybe `QueryApi.adoc`) imo

OK, done.

documentation/src/main/java/org/hibernate/userguide/model/Person.java
gavinking: Oh, excuse me! I thought the existing code was using uppercase keywords everywhere, not just in that one code example.

Fixed. I've lowercased them.

sebersole:
Still looking, but a few comments:


== QueryLanguage.adoc

It's strange in my opinion to jump right into case sensitivity.  Seems like there ought to be at least some discussion of what HQL is, what JPQL is and how they relate.  Yes the old docs hyper-focused on this way to much, but I think this is over-correcting a bit.  E.g.

```
This chapter describes Hibernate Query Language (HQL) and Jakarta Persistence Query Language (JPQL).

[NOTE]
====
HQL is a complete super-set of JPQL.  This chapter focuses on the more complete, more powerful HQL language.

If strict Jakarta Persistence compliance is desired, use the setting `hibernate.jpa.compliance.query=true`.  With this configuration, any attempt to use HQL features beyond the JPQL subset will result in an exception being thrown.
====

HQL (and JPQL) are String-based query languages based loosely on SQL....
```


== Query.adoc

The start of this chapter spends an awful lot of its initial real-estate discussing HQL.  And then also Criteria, etc.  Just link to those chapters.

Then the first meat of the chapter discusses named queries.  It should first discuss the actual `Query` interface imo.

gavinking:
superseded by https://github.com/hibernate/hibernate-orm/pull/4541.

gavinking:
> It's strange in my opinion to jump right into case sensitivity.

Ummm. I can move it later. That was how I found it, and it didn't seem *extremely* bizarre to me that the doc would start discussing lexical structure.

But I don't mind moving it.

gavinking:
> Seems like there ought to be at least some discussion of what HQL is, what JPQL is and how they relate. 

Yes, definitely, I agree. Currently there is no real preamble at all.

gavinking:
> The start of this chapter spends an awful lot of its initial real-estate discussing HQL. And then also Criteria, etc. Just link to those chapters.

> Then the first meat of the chapter discusses named queries. It should first discuss the actual Query interface imo.

I agree with both those comments, and I already struggled a lot with the first part of the chapter, partly because I inherited a total disaster. Let's leave this for now, and come back to it when we have more time.

gavinking:
@sebersole I copied your suggested preamble, with some minor changes.

PR #4533, opened 2021-12-29
propose Query.getSingleResultOrNull()

gavinking:
This is a method we have in Hibernate Reactive.

It's _very_ similar to the existing `uniqueResult()` method, but its name is much clearer, and it throws the JPA exception type.

The main reason for introducing this is so that:

1. users can find it more easily, and
2. people reading the code understand what's going on (it's not clear with `uniqueResult()`).

I doubt that many people appreciate the difference between `getSingleResult()`  and `uniqueResult()`, especially since the Javadoc in Hibernate has been incorrect for a long time.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/procedure/internal/ProcedureCallImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/procedure/internal/ProcedureCallImpl.java
@@ -1002,6 +1002,24 @@ public class ProcedureCallImpl<R>
 	}
 
 	@Override
+	public R getSingleResultOrNull() {
+		final List<R> resultList = getResultList();
+		if ( resultList == null || resultList.isEmpty() ) {
+			return null;
+		}
+		else if ( resultList.size() > 1 ) {
+			throw new NonUniqueResultException(
+					String.format(
+							"Call to stored procedure [%s] returned multiple results",
+							getProcedureName()
+					)
+			);
+		}
+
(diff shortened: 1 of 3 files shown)


sebersole:
Why not just `#uniqueResultOptional`?

gavinking:
> Why not just `#uniqueResultOptional`?

Because `query.uniqueResultOptional().orElse(null)` is just a lot less comfortable than `query.getSingleResultOrNull()`, especially for people who prefer the better-named `query.getSingleResult()` method.

gavinking:
> Why not just `#uniqueResultOptional`?

I mean, to be clear: if you're happy with the name `uniqueResult`, then this new method does almost exactly the same thing as plain `uniqueResult()`, modulo the exception type (which is a minor thing but not nothing).

This is mainly for people who prefer the naming of `getSingleResult()`, and the clarity of purpose of `getSingleResult()` vs `getSingleResultOrNull()`.

It doesn't introduce new functionality.

beikov (approved):
LGTM. Not sure about the failures on Oracle for `StandardFunctionTests#testTimestampAddDiffFunctions` but I'll look into them in case they happen again after a merge.

sebersole:
Rebasing this, and I'll apply before cr1

gavinking:
Excellent, thanks!

sebersole:
Applied upstream

PR #4537, opened 2021-12-29
propose @DialectOverride annotation

gavinking:
This allows the use of native SQL in annotations of portable programs.

We've needed something like this for a long time.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/mapping/formula/FormulaTests.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/mapping/formula/FormulaTests.java
@@ -14,6 +14,8 @@ import jakarta.persistence.criteria.Root;
 
 import org.hibernate.annotations.Formula;
 
+import org.hibernate.annotations.DialectOverride;
+import org.hibernate.dialect.H2Dialect;
 import org.hibernate.testing.orm.junit.DomainModel;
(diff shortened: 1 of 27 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/mapping/formula/FormulaTests.java
```diff
@@ -94,7 +96,8 @@ void tearDown(SessionFactoryScope scope) {
 
 		private Double rate;
 
-		@Formula(value = "credit * rate")
+		@Formula(value = "credit * rate",
+				overrides = @DialectOverride(dialect = H2Dialect.class, value ="rate * credit") )
 		private Double interest;
```
gavinking, line 101: Here's an example of how you might use it, but it's much more general than this.

gavinking:
Currently the API for this looks like:

```java
@Formula(value = "credit * rate",
	overrides = @DialectOverride(dialect = H2Dialect.class, value ="rate * credit") )
```

I guess an alternative way to do it, that has less impact on the existing annotations, would be to write:

```java
@Formula(value = "credit * rate")
@DialectOverride(annotation=Formula.class, dialect = H2Dialect.class, value ="rate * credit")
```

I'm unclear on which option is "better". I guess they're pretty equivalent.

gavinking:
Superseded by #4607.

PR #4538, opened 2021-12-30
remove some obsolete code from AnnotationBinder

gavinking:
This code was resulting in warnings in IntelliJ.


sebersole (commented):
lgtm

no idea why we are back to having failures.  these jobs seem so flaky

gavinking:
> no idea why we are back to having failures. these jobs seem so flaky

It's date time-related tests, as per usual.

PR #4540, opened 2021-12-30
introduce @GeneratedColumn annotation

gavinking:
- make it easy to produce portable DDL with `generated always as` or equivalent
- also add `name` to `@ColumnDefault`, and add some missing tests for basic usage patterns of `@ColumnDefault`


gavinking:
I'm very happy with this stuff now, and I plan to merge it tomorrow.

hibernate-core/src/main/java/org/hibernate/annotations/GeneratedColumn.java
```diff
@@ -0,0 +1,51 @@
...
+ * database.
+ *
+ * @author Gavin King
+ *
+ * @see ColumnDefault
+ */
+@Target( {FIELD, METHOD} )
+@Retention( RUNTIME )
+@ValueGenerationType(generatedBy = GeneratedAlwaysValueGeneration.class)
+public @interface GeneratedColumn {
```
beikov, line 30: Wouldn't it be better to name it `ColumnGenerated` for consistency sake with `ColumnDefault`?

gavinking: I went down that path originally, but eventually settled on this, _after_ I realized that it's really not the same as `@ColumnDefault` at all. They're actually much more different than they seemed at first blush.

PR #4545, opened 2022-01-01
More doc rework

gavinking:


hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -933,7 +933,7 @@ private NonSelectQueryPlan buildInsertQueryPlan() {
 		final EntityPersister entityDescriptor = getSessionFactory().getDomainModel().findEntityDescriptor( entityNameToInsert );
 
 		final SqmMultiTableInsertStrategy multiTableStrategy = entityDescriptor.getSqmMultiTableInsertStrategy();
-		if ( multiTableStrategy == null ) {
+		if ( multiTableStrategy == null || sqmInsert instanceof SqmInsertValuesStatement ) {
```
beikov, line 936: Hmm, was that an issue somehow for you because I don't know if it is safe to simply do this, given that a VALUES clause can have expressions.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
```diff
@@ -147,77 +190,105 @@ include::{sourcedir}/../batch/BatchTest.java[tags=batch-bulk-hql-update-version-
...
 ====
 
+An `insert` statement must be executed by calling `Query#executeUpdate()`.
+
+[IMPORTANT]
+====
+`insert` statements are inherently _not_ polymorphic!
+The `targetFields` list is of fixed length, whereas each subclass of an entity class might declare additional fields.
+For an entity involved in a mapped inheritance hierarchy, only attributes declared directly by the named entity may be used in the list of target fields.
+Superclass attributes are not allowed, and subclass attributes do not make sense.
```
beikov, line 257: >Superclass attributes are not allowed

Not sure this is still the case. I guess you just moved this to here from somewhere? Assigning superclass attributes is definitely possible in the insert statement just as well as assigning fields with columns contained in secondary tables. This is new and I just recently implemented this.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
```diff
@@ -1604,63 +1725,87 @@ These functions are especially useful for reporting:
...
+=== The `limit` and `offset` clauses
+
+The `limit` and `offset` clauses are an alternative to the use of `setMaxResults()` and `setFirstResult()` respectively.
+
+[TIP]
+====
+If the `limit` or `offset` is parameterized, its much easier to use `setMaxResults()` or `setFirstResult()`.
+====
+
+The SQL syntax `fetch first ... rows only` and `fetch next ... rows only` is also allowed.
```
beikov, line 1810: It might be worth adding a paragraph about the `fetch` clause since it also allows to specify fetching a percentage of rows instead of just an absolute amount of rows and has the option _with ties_.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
gavinking: Yes, I was fixing the fact that `insert ... values ...` was broken for the _very first example I tried_.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
gavinking: OK, good, it seemed strange to me that this would be the case.

So can I completely remove that passage or are there still limitations the user needs to know about?

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
gavinking: (Specifically, it was broken, because it was trying to insert `values()` into a temp table, and didn't assign a value to the `rn_` column.)

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
beikov: I don't think that there are any limitations anymore.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
gavinking: I added a BNF last night.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
beikov: Ah, that's a bug :)

I guess your tests will reproduce that, so I'll take a look and see if I can change this logic somehow.

gavinking: Yup, they will.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
gavinking: I presume that it's still not possible to assign to _subclass_ fields.

gavinking: https://github.com/hibernate/hibernate-orm/pull/4558

beikov: > I presume that it's still not possible to assign to subclass fields.

AFAIU it doesn't make sense to do that 😄

PR #4546, opened 2022-01-01
minor cleanups to the grammar

gavinking:
- simplify two rules that were unnecessarily flexible
- remove useless parens
- rename some rules for readability

### DIFF
--- a/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
+++ b/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
@@ -31,20 +31,20 @@ selectStatement
 	: queryExpression
 	;
 
-subQuery
+subquery
 	: queryExpression
 	;
 
-dmlTarget
+targetEntity
 	: entityName identificationVariableDef?
 	;
 
 deleteStatement
-	: DELETE FROM? dmlTarget whereClause?
+	: DELETE FROM? targetEntity whereClause?
 	;
 
 updateStatement
-	: UPDATE VERSIONED? dmlTarget setClause whereClause?
+	: UPDATE VERSIONED? targetEntity setClause whereClause?
 	;
 
 setClause
@@ -56,10 +56,10 @@ assignment
 	;
 
 insertStatement
-	: INSERT INTO? dmlTarget targetFieldsSpec (queryExpression | valuesList)
+	: INSERT INTO? targetEntity targetFields (queryExpression | valuesList)
 	;
 
-targetFieldsSpec
+targetFields
 	: LEFT_PAREN dotIdentifierSequence (COMMA dotIdentifierSequence)* RIGHT_PAREN
 	;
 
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
```diff
@@ -419,12 +419,12 @@ comparisonOperator
...
 	: (ELEMENTS|INDICES) LEFT_PAREN dotIdentifierSequence RIGHT_PAREN				# PersistentCollectionReferenceInList
 	| LEFT_PAREN (expressionOrPredicate (COMMA expressionOrPredicate)*)? RIGHT_PAREN# ExplicitTupleInList
-	| LEFT_PAREN subQuery RIGHT_PAREN												# SubQueryInList
+	| LEFT_PAREN subquery RIGHT_PAREN												# SubqueryInList
 	| parameter 																	# ParamInList
 	;
 
 likeEscape
-	: ESCAPE expression
+	: ESCAPE STRING_LITERAL
```
beikov, line 427: May I ask for a reference for this change? I would assume parameters are valid as well.

gavinking: I can't imagine why anyone would _ever_ need to parameterize the wildcard escape character, when they have total control over the pattern string.

If this were a _function_ with proper delimiters around the argument list, I guess I would probably agree with you and leave it as `expression`. But it's not, it's part of an operator, and could _very_ easily be the source of grammatical ambiguities down the road. It's _probably_ OK right now because it's protected by a prefix keyword that we don't use for anything else. But even so, I just don't see why we need to open ourselves up to that. I would rather just not even need to think about it.

(Sadly, ANTLR4, unlike ANTLR3, does not warn about ambiguities.)

gavinking: Well, actually the grammar _was_ and _is_ strictly-speaking ambiguous, the usual ambiguity you get with nested `if`/`else`, though that's not a real concrete problem here because of the types.

Anyway, no need to make things more complex than they need to be.

beikov: >I can't imagine why anyone would ever need to parameterize the wildcard escape character, when they have total control over the pattern string.

I just don't want to arbitrarily limit the grammar to something because we can't imagine certain scenarios. If there is a real problem we can think about it, but this just seems like an artificial restriction.

gavinking: I think it's like, the most natural restriction on earth. Have you ever seen anyone use anything other than a literal-valued `escape` in the whole history of the Universe?

beikov: I don't know, some generic code might render it this way. In fact, our JPA Criteria implementation in 5.x rendered it to HQL this way by default, so I wouldn't be too surprised that some other library took inspiration from our old implementation ;)

gavinking: Meh, OK, it's true that `CriteriaBuilder.like()` does accept an expression here.

Now, that doesn't matter anymore, because the new `Criteria` implementation works directly with `SqmLikePredicate`, so no problem there.

But OTOH, actually the JPQL spec says:

> The optional escape_character is a single-character string literal or a character-valued input parameter 

So, indeed, we're expected to accept a parameter there.

But you'll note that it _is_ otherwise a literal, according to JPQL.

PR #4548, opened 2022-01-01
id(), version(), and naturalid()

gavinking:
I noticed that the grammar defines `id()`, `version()`, and `naturalid()` functions.

Two of these were already working, but weren't documented, and I could not find where they were tested. So I added documentation and tests.

`naturalid()` was unimplemented, so I (roughly) implemented it, at least for the case of a single`@NaturalId` field.


beikov (approved):
LGTM

PR #4549, opened 2022-01-02
redesign 'collate' and add tests

gavinking:
I started out adding tests, but then I realized that the grammar for `collate` was repetitive and ambiguous, and the _very first thing I tried_ fell into one of the ambiguities and exposed a bad bug.

So I changed this stuff to use a much less ambiguous syntax that's also *much* more consistent with the rest of the language.


gavinking:
I'm going to squash and merge this, because if I don't, then I'm going to get myself into a hell of conflicts.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/select/SortSpecification.java
```diff
@@ -20,18 +20,15 @@
 	private final Expression sortExpression;
 	private final SortOrder sortOrder;
 	private final NullPrecedence nullPrecedence;
+	private final Collation collation;
```
beikov, line 23: You seem to think that collations only matter for sorting but that's not true, the matter for everything that is based on relational operators and sorting is just one aspect. The DISTINCT operator, relational predicates etc. all depend on the collation. Usually, databases are smart enough to infer a proper collation, but sometimes not, which is when the collation feature as it was before came in.

It's great that you now added tests for this, but I would like to revert the implementation and grammar change.

gavinking: I don't think that, and that's more like the _opposite_ of what this change does.

You're looking at one line of code in isolation and don't realize the context of the code you're looking at.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/select/SortSpecification.java
```diff
@@ -20,18 +20,15 @@
...
+		this( sortExpression, null, sortOrder, NullPrecedence.NONE );
 	}
 
 	public SortSpecification(Expression sortExpression, String collation, SortOrder sortOrder, NullPrecedence nullPrecedence) {
-		if ( collation == null ) {
-			this.sortExpression = sortExpression;
-		}
-		else {
-			this.sortExpression = new Collate( sortExpression, collation );
-		}
```
gavinking, line 34: @beikov 

FTR, this is what you were missing: this was what it did before.

The client code which instantiates this class all passes in a collation. I don't know why, and I don't know what that was for. Quite likely, it can all be ripped out now, but I certainly was not going to rip out code I don't understand.

PR #4556, opened 2022-01-03
better explanation of query language concepts

gavinking:
and a bit of a cleaner-looking css

### DIFF
--- a/documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
+++ b/documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
@@ -69,14 +69,14 @@ Most of the complexity here arises from the interplay of set operators (`union`,
 
 We'll describe the various clauses of a query later in this chapter, but to summarize, a query might have:
 
-* a `select` list, specifying a projection (the things to return from the query),
-* a `from` clause and joins, specifying the entities involved in the query,
-* a `where` clause, specifying a restriction,
-* a `group by` clause, for aggregation,
-* a `having` clause, specifying a restriction to apply _after_ aggregation,
-* set operators applied to the results of multiple subqueries,
-* an `order by` clause, for ordering the results, and even
-* a `limit`/`offset` clause, for limiting the results.
+* a `select` list, specifying a <<hql-select-clause,projection>> (the things to return from the query),
+* a `from` clause and joins, <<hql-from-clause,specifying>> the entities involved in the query,
+* a `where` clause, specifying a <<hql-where-clause,restriction>>,
+* a `group by` clause, for <<hql-group-by,aggregation>>,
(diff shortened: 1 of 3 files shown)


documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
```diff
@@ -1910,7 +1997,16 @@ include::{sourcedir}/HQLTest.java[tags=hql-order-by-example]
...
-=== The `limit` and `offset` clauses
+=== Limits and offsets
+
+It's often useful to place a hard upper limit on the number of results that may be returned by a query.
+
+[NOTE]
+====
+_Limiting_ certainly _isn't_ a well-defined relational operation, and must be used with care.
+
+A limit can easily break the semantics of certain other features of HQL, including <<hql-explicit-fetch-join,fetch joins>>.
```
beikov, line 2008: In fact it is the other way around, fetch joins cause limits to be applied in memory 😆

gavinking: Sorry? Are you saying we now _don't_ pass through the `limit` clause when we have a fetch join, materialize the whole result set and apply the limit in Java!?

gavinking: I tested it. What's described in the documentation is the true behavior.

beikov: Take a look into `org.hibernate.query.sqm.internal.QuerySqmImpl#doList` which throws an exception if configured, or emits a warning. If that doesn't happen, there is a bug somewhere.

gavinking: You can try it for yourself.

https://github.com/hibernate/hibernate-orm/pull/4568/files

All those queries result in a limit clause being sent to the database.

beikov: Right, because you are fetch joining a to-one which preserves the cardinality. Try fetch joining a collection.

gavinking: Oh gawd I am too. Sorry.

gavinking: Well, changing it to a to-many collection made to difference. I still got the limit clause, and verified that I can get an incomplete collection returned.

So if you tell me that's a bug, then it's a bug.

Honestly I think both behaviors are broken, just in different ways.

beikov: I'll look into your other PR then and try to figure out what goes wrong here. There should be an exception.

PR #4557, opened 2022-01-03
allow parameter in like escape

gavinking:
as required by JPQL

### DIFF
--- a/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
+++ b/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
@@ -424,7 +424,7 @@ inList
 	;
 
 likeEscape
-	: ESCAPE STRING_LITERAL
+	: ESCAPE (STRING_LITERAL | parameter)
 	;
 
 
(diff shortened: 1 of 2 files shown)


beikov (commented):
I still don't understand why you want to restrict the grammar unnecessarily, but if you do so, please also add a note into the migration guide about dropping support for arbitrary expressions.

gavinking:
> I still don't understand why you want to restrict the grammar unnecessarily

It's I guess really because I don't view the syntax `string like pattern escape '/'` as a ternary operator in three strings.

I view the `escape` keyword as introducing something that belongs to a class of language elements containing:

- `trimCharacter` and `padCharacter`, 
- things that come after `as`: a `castTarget`, `collation`, or `format`, and
- even a bit like the literal values that follow `date`, `time`, or `datetime`.

None of those things are treated as generic expressions in the grammar, and it seems to me that `pattern escape 'c'` is most like those things.

In fact, I dislike the syntax of `like ... escape` quite a lot. It seems to me that  something like 

    matches(string like pattern escape '/')

would be easer to parse (for the human reader).

But of course 99.9% of the time `escape` is not needed and in that case the infix binary operator `like` is pretty good and readable.

PR #4561, opened 2022-01-03
test showing alias matching is case-sensitive

gavinking:
this is wrong according to JPQL


gavinking:
Ugh, this is because I wasn't in JPA compliance mode, and I was mislead by the old docs.

PR #4568, opened 2022-01-03
better documentation for limit and friends

gavinking:
with code example

### DIFF
--- a/documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
+++ b/documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
@@ -2868,6 +2868,54 @@ public class HQLTest extends BaseEntityManagerFunctionalTestCase {
 	}
 
 	@Test
+	public void test_hql_limit_example() {
+
+		doInJPA(this::entityManagerFactory, entityManager -> {
+			//tag::hql-limit-example[]
+			List<Call> calls1 = entityManager.createQuery(
+				"select c " +
+				"from Call c " +
+				"join c.phone p " +
+				"order by p.number " +
+				"limit 50",
+				Call.class)
+			.getResultList();
+
+			// same thing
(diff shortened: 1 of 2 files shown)


documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -2867,6 +2867,54 @@ public void test_hql_order_by_example_2() {
...
+				"select c " +
+				"from Call c " +
+				// join fetch should not be used with limit
+				"join fetch c.phone " +
+				// but if you insist, at least sort by the collection owner
+				"order by c, c.phone, c.timestamp " +
+				// this won't be the final number of results!
+				"limit 50",
+				Call.class)
+			.getResultList();
```
gavinking, line 2913: cc @beikov

documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -2867,6 +2867,54 @@ public void test_hql_order_by_example_2() {
...
+			List<Phone> wrongCalls = entityManager.createQuery(
+				"select p " +
+				"from Phone p " +
+				// join fetch should not be used with limit
+				"join fetch p.calls " +
+				// but if you insist, at least sort by the collection owner
+				"order by p " +
+				// this won't be the final number of results!
+				"limit 50",
+				Phone.class)
```
gavinking, line 2912: @beikov fixed the example, but same behavior.

gavinking:
I'm going to merge this when the tests succeed. If the behavior documented here changes before the release, please let me know so that I can update this text to reflect that the new behavior in 6 is differently-broken. ;-)

gavinking:
>  to reflect that the new behavior in 6 is differently-broken. ;-)

(Because, to be clear, applying a `limit` _in memory_ is still pretty broken.

PR #4569, opened 2022-01-03
use 'select id from final table' to retrieve generated key

gavinking:
see https://github.com/hibernate/hibernate-orm/pull/4524#issuecomment-1002055088

Is this correct @froque and @katzyn?

Does it actually work?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/identity/H2IdentityColumnSupport.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/identity/H2IdentityColumnSupport.java
@@ -23,11 +23,22 @@ public class H2IdentityColumnSupport extends IdentityColumnSupportImpl {
 
 	@Override
 	public String getIdentitySelectString(String table, String column, int type) {
+		//not supported since H2 1.4.200
 		return "call identity()";
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/identity/H2IdentityColumnSupport.java
```diff
@@ -23,11 +23,22 @@ public String getIdentityColumnString(int type) {
...
+	}
+
+	@Override
+	public boolean supportsInsertSelectIdentity() {
+		return true;
+	}
+
+	@Override
+	public String appendIdentitySelectToInsert(String insertString) {
+		return "select id from final table (" + insertString + ")";
```
froque, line 42: I explored something like this. This does not work if the column is not named `id`

```
19:03:24.089 [Time-limited test] ERROR org.hibernate.engine.jdbc.spi.SqlExceptionHelper - Column "ID" not found; SQL statement:
select id from final table (insert into Event (id2, active, createdOn) values (default, ?, ?)) [42122-202]
```

gavinking: OK thanks.

PR #4570, opened 2022-01-03
Fix collation handling for ordering expressions. Fix SqmQuerySpec#setRestriction

beikov:


hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -933,7 +933,7 @@ private NonSelectQueryPlan buildInsertQueryPlan() {
 		final EntityPersister entityDescriptor = getSessionFactory().getDomainModel().findEntityDescriptor( entityNameToInsert );
 
 		final SqmMultiTableInsertStrategy multiTableStrategy = entityDescriptor.getSqmMultiTableInsertStrategy();
-		if ( multiTableStrategy == null || sqmInsert instanceof SqmInsertValuesStatement ) {
```
gavinking, line 936: Did you also fix the bug with `insert .... values`?

beikov: I didn't encounter issues when running tests locally, so I wanted to check how it runs on CI. I'll look into that though if I see errors.

gavinking: Well yeah, of course, you probably didn't run the documentation tests.

gavinking:
So you said you fixed something. And there's new code there that looks like it's doing something new.

But you have not added any tests.

So what have you fixed that does not require tests?

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
beikov: I did run the full suite with H2 and it seems H2 also works on CI, but there are errors for other DBs which I will look into tomorrow.

gavinking: And yes: if you check the build, `HQLTest.hql_insert_example()` is failing.

It's like you think I'm an incompetent idiot who just reports and fixes imaginary bugs.

I just don't understand.

beikov:
What I fixed is, remove the collation from SortSpecification and instead wrap the expression if a collation should be applied for all OrderingExpressions. I didn't test that because it also wasn't tested before, and I first wanted to see if there might be existing tests that fail. I'll add a test though.

gavinking:
>  I didn't test that because it also wasn't tested before

That's not how any of this works.

The most basic principle is: if I fix something, then if possible I add a test for it.

gavinking:
> Add table to readme showing how to test with every database

👏👏👏 Thank you!!

documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -219,6 +222,15 @@ public void hql_insert_example() {
 				"values (100L, 'Jane Doe')")
 			.executeUpdate();
 
+			//end::hql-insert-example[]
+		});
+	}
+
+	@Test
+	@RequiresDialectFeature(feature = DialectFeatureChecks.SupportsValuesListForInsert.class)
```
gavinking, line 230: This is going to conflict with #4574 but I'm just going to let you sort that out.

PR #4579, opened 2022-01-05
createStatement() methods for QueryProducer

gavinking:
So, I kinda messed up, and forgot that `createQuery()` is used for running things that aren't queries (insert, update, and delete statements).

But it's OK, because this is much better anyway.


gavinking:
We need these operations for sure. You guys can try it out, and if you can think of better names, we'll change them.

PR #4580, opened 2022-01-05
Wip/6.0 reenable tests 5

dreab8:


sebersole:
Do we still need this?  Sorry it took me so long to look

dreab8:
the PR makes a lazy collection with  `fetch=join` (<set name="bids" cascade="all" fetch="join" ....>) loaded lazily when a Criteria or HQL query does not explicitly fetch it

beikov:
Could you rebase this PR @dreab8? I pushed my fixes to main already for retaining laziness even if fetch="join" is used.

dreab8:
@beikov rebased :)

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -6215,6 +6216,12 @@ else if ( getLoadQueryInfluencers().hasEnabledFetchProfiles() ) {
 						}
 					}
 				}
+				else if ( fetchable instanceof PluralAttributeMapping && fetchable.getMappedFetchOptions()
+						.getStyle() == FetchStyle.JOIN ) {
+					if ( ( (PluralAttributeMapping) fetchable ).isLazy() ) {
+						fetchTiming = FetchTiming.DELAYED;
+					}
+				}
```
beikov, line 6224: I don't think this is necessary anymore

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/PluralAttributeMappingImpl.java
```diff
@@ -100,6 +100,8 @@
 
 	private final String bidirectionalAttributeName;
 
+	private final boolean isLazy;
+	private final CascadeStyle cascadeStyle;
```
beikov, line 104: I don't think this is necessary anymore

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/PluralAttributeMapping.java
```diff
@@ -116,4 +116,6 @@ default void applyWhereRestrictions(
 			SqlAstCreationState creationState) {
 		getCollectionDescriptor().applyWhereRestrictions( predicateConsumer, tableGroup, useQualifier, creationState );
 	}
+
+	boolean isLazy();
```
beikov, line 120: I don't think this is necessary anymore

beikov (commented):
Looks good mostly, but I think a few changes have now become unnecessary

beikov (approved):

PR #4581, opened 2022-01-05
typecheck HQL function arguments

gavinking:
This is my second attempt to implement type checking for HQL function arguments.

This is a rather successful attempt so far, though I'll still surely have a couple of test failures on some DBs.

The design still isn't quite complete: this checking will still let me try to `extract` a year from a time, for example, but additional `ArgumentsValidator`s like that are easy to add later.

In general, I've erred on the side of being permissive. But one question I have is: should I really check that the arguments of `concat` are strings, as the JPQL spec says, or is it OK to trust the DB to do the implicit typecast? Not sure. I'm no great lover of implicit typecasts but this one seems sorta safe/useful.

There's still some work to go here, but I want to see the damage across different dbs.


gavinking:
Alright! I'm pretty happy with this now.

gavinking:
Yay, finally! I've been promising to do this for > 2 years.

PR #4583, opened 2022-01-06
quote column names with initial _

gavinking:
DB2 and Oracle don't allow `_identifiers` and it looks like this is the ANSI standard.

So maybe we should just auto quote identifiers that start with an _.

(I wish I could make this depend on the `Dialect` but that would be a much more impactful change for such a small thing.)


gavinking:
On reflection, this isn't the right way to do this at all: it's much too breaky, since people might have `@Column("_id")` mapping to a column named `_ID`.

PR #4584, opened 2022-01-06
HHH-15013 H2 2.0.202 no longer supports nextval, replacing with ANSI SQL

froque:


beikov:
Nice, thanks for this. I had a fix for this in the pipeline as well, but you beat me to it :)

PR #4588, opened 2022-01-06
H2 2.0 support

beikov:
https://hibernate.atlassian.net/browse/HHH-15014
https://hibernate.atlassian.net/browse/HHH-15009
https://hibernate.atlassian.net/browse/HHH-15035

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
@@ -9,20 +9,28 @@ package org.hibernate.dialect.function;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.sql.Types;
+import java.time.format.DateTimeFormatter;
 import java.util.Date;
 import java.util.List;
 import java.util.Arrays;
+import java.util.Locale;
 import java.util.function.Supplier;
 
+import org.hibernate.QueryException;
 import org.hibernate.dialect.Dialect;
 import org.hibernate.metamodel.mapping.BasicValuedMapping;
 import org.hibernate.query.ReturnableType;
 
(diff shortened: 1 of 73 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
```diff
@@ -2379,12 +2384,33 @@ public static void format_toVarchar(QueryEngine queryEngine) {
...
 						queryEngine.getTypeConfiguration().getBasicTypeRegistry().resolve( StandardBasicTypes.STRING )
 				)
-				.setExactArgumentCount( 2 )
-				.setParameterTypes(TEMPORAL, STRING)
+				.setArgumentsValidator( formatValidator() )
 				.setArgumentListSignature( "(TEMPORAL datetime as STRING pattern)" )
 				.register();
 	}
 
+	public static ArgumentsValidator formatValidator() {
```
gavinking, line 2392: This is a good idea, but have fun with the conflicts.

beikov: I fixed them already :)

gavinking: Oh, I see, no conflicts.

gavinking (commented):
Wait, @beikov, I thought this validation was a good idea, but then I realized that it actually duplicates what already happens in `SemanticQueryBuilder.visitFormat()`. It _is_ a good idea, and that's why I already had it. :-)

Now, doing that validation in an `ArgumentsValidator` is, I dunno, _not obviously better_? SQB already has the job of interpreting/validating literal values, so I don't see this validation as being out of place where it is right now.

PR #4589, opened 2022-01-06
use instanceof instead of isInstance() on the hot path

gavinking:
That's surely NBD, or it would have shown up in Sanne's performance work long ago, but still, every time I see it I'm just left with that nagging doubt.

And it's such an easy thing to fix in IntelliJ..


gavinking:
Does anyone object to this change? @sebersole?

gavinking:
Nobody has objected, and all those warnings just mess with my OCD.

PR #4595, opened 2022-01-07
remove the layer-breaking getLobCreator() methods from Hibernate class

gavinking:
This is in principle a breaking change, but one I consider pretty innocuous, since I can't imagine why anyone was calling `Hibernate.getLobCreator(session)` instead of `session.getLobHelper()`.

It's important that we not leak internal APIs like `SessionImplementor` into the signature of public APIs in `org.hibernate`.


gavinking:
P.S. if someone is very concerned about the breakage, we can put back the version which accepts `Session`. But I can't quite see it's purpose.

PR #4602, opened 2022-01-08
test showing maxindex() / maxelement() broken

gavinking:
This test as written only fails on some DBs, but looking more closely I just noticed that the generated SQL is actually wrong for _all_ of them, since it used both a join _and_ a subquery, resulting in duplicated results.


beikov:
I messed up with Git apparently, but the change was just to remove the `@RequiresDialect` uses in 2 tests. I'll create a new PR with a fix.

beikov:
See https://github.com/hibernate/hibernate-orm/pull/4789

PR #4603, opened 2022-01-08
test showing selecting enum value causes NPE

gavinking:


sebersole:
Not particularly useful imo, but it wasn't difficult to implement.

sebersole:
integrated upstream with fix

gavinking:
Great, thanks!

> Not particularly useful imo

Sure, it wasn't particularly a question of _usefulness_, more a matter of not blowing up for things that are sort perfectly-legal expressions, just because it _looks bad_, aside from anything else.

PR #4606, opened 2022-01-08
Fix converter issues

dreab8:


sebersole:
What is the fix?  Seems like this is trying to deal with exceptions thrown from the converter calls?

dreab8:
@sebersole especially testing that the converter is applied! without the change to `SqmUpdateStatement.java ` the converter is not applied at all, because we resolve the type of the parameter based on the value and not on the attribute, so later on we do not know that we need a converter.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/jpa/AbstractJpaTupleElement.java
```diff
@@ -47,7 +47,7 @@ protected void setAlias(String alias) {
 		return expressableType;
 	}
 
-	protected final void setExpressableType(SqmExpressable<?> expressableType) {
+	public final void setExpressableType(SqmExpressable<?> expressableType) {
```
beikov, line 50: This change is a bit concerning. Why is it necessary to set the expressable type to null for `JpaCriteriaParameter`? Wouldn't it be better to have some sort of special method on `JpaCriteriaParameter` to take care of this?

beikov (commented):
LGTM overall, but the change of visibility of `setExpressableType` is a bit concerning.

beikov (approved):

PR #4607, opened 2022-01-09
different approach to overriding SQL in annotations

gavinking:
Introduces the `@ForDialect` annotation, which is really a perfectly elegant thing if you're careful not to look at its implementation.

This approach has the advantage that it doesn't impact the existing annotations at all: it's completely self-contained.

See https://github.com/hibernate/hibernate-orm/discussions/4528.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/annotations/DialectOverride.java
+++ b/hibernate-core/src/main/java/org/hibernate/annotations/DialectOverride.java
@@ -1,0 +1,300 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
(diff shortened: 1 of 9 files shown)


gavinking:
Note, with this new annotation, you can write things like:

```java
@Entity(name = "Person")
@Check(constraints = "code > 0")
@ForDialect(dialect = H2Dialect.class, check = @Check(constraints = "sign(code) = 1"))
public static class Person { ... }
```

gavinking:
Does anyone know what's up with MySQL and Maria? Why they keep timing out?

beikov:
Can you rebase? I fixed the hanging yesterday evening.

gavinking:
>  I fixed the hanging yesterday evening.

Thanks, that worked.

gavinking:
I have merged this. It's marked `@Incubating`, and it's very under-tested, since some of this is quite hard to write tests for, hell, I noticed that we have no tests for `@Check` at all. Let's see if people find it useful.

PR #4611, opened 2022-01-09
Move FetchMode to org.hibernate.mapping

gavinking:
See https://github.com/hibernate/hibernate-orm/discussions/4610.


gavinking:
Steve doesn't want to do this, so I'm closing this for now.

PR #4614, opened 2022-01-10
Fix postLoad callback method not called

dreab8:


dreab8:
this is not the correct solution

PR #4626, opened 2022-01-12
When JPA list compliance is enabled interpret List not annotated with…

dreab8:
… @OrderColumn with BAG semantic


dreab8:
discarded

PR #4628, opened 2022-01-12
massive cleanout of the transformers stuff

gavinking:
- add meaningful generic types to stuff
- remove things which weren't used/tested, and seem obsolete
  (and which could not be propertly generified)

@sebersole WDYT?


gavinking:
Note, this is breaky. But: removal of `Criteria` was much more breaky in this area, and `ResultTransformer` was already marked deprecated, so we have the right.

sebersole:
Looked good to me.  Made some Javadoc changes and integrated upstream

PR #4635, opened 2022-01-12
Fix native query alias validation

dreab8:
may be I'm wrong but provably we can just remove the check for duplicate alias and not throwing the`NonUniqueDiscoveredSqlAliasException` , the test added when this exception was introduced seems not failing anymore even with duplicate aliases.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/jpa/compliance/nativequery/NativeQueryTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/jpa/compliance/nativequery/NativeQueryTest.java
@@ -1,0 +1,193 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.orm.test.jpa.compliance.nativequery;
+
+import java.util.List;
+import java.util.Objects;
+
+import org.hibernate.testing.orm.junit.EntityManagerFactoryScope;
+import org.hibernate.testing.orm.junit.Jpa;
+import org.junit.jupiter.api.Test;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.EntityResult;
+import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.OneToOne;
+import jakarta.persistence.SqlResultSetMapping;
+import jakarta.persistence.Table;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+@Jpa(
+		annotatedClasses = { NativeQueryTest.Item.class, NativeQueryTest.Order.class }
+)
+public class NativeQueryTest {
+
+	@Test
+	public void testIt(EntityManagerFactoryScope scope) {
+
+		scope.inTransaction(
+				entityManager -> {
+					final Item item1 = new Item( 1, "WaterShoes" );
+					final Item item2 = new Item( 2, "FlipFlops" );
+					final Item item3 = new Item( 3, "Sandals" );
+
+					entityManager.persist( item1 );
+					entityManager.persist( item2 );
+					entityManager.persist( item3 );
+
+					Order order1 = new Order( 1, 25.0D, item1 );
+					entityManager.persist( order1 );
+					Order order2 = new Order( 2, 125.0D, item2 );
+					entityManager.persist( order2 );
+					Order order3 = new Order( 3, 150.0D, item3 );
+					entityManager.persist( order3 );
+
+					List<Object[]> results = entityManager.createNativeQuery(
+							"Select o.ID, o.ORDER_PRICE, o.ORDER_ITEM , i.ID, i.ITEM_NAME from ORDER_TABLE o, ITEM_TABLE i "
+									+ " WHERE (o.ORDER_PRICE > 140) AND (o.ORDER_ITEM = i.ID)",
+							"OrderItemResult"
+					).getResultList();
+
+					assertEquals( 1, results.size(), "Wrong result size" );
+
+					for ( Object[] result : results ) {
+						assertEquals( 2, result.length,  "Wrong number of objects in the array result" );
+						for ( Object o : result ) {
+							if ( o instanceof Order ) {
+								assertEquals( order3, o );
+							}
+							else if ( o instanceof Item ) {
+								assertEquals( item3, o );
+							}
+							else {
+								fail( "Received an unexpected object result:" + o );
+							}
+						}
+					}
+				}
+		);
+	}
+
+	@Entity(name = "Item")
+	@Table(name = "ITEM_TABLE")
+	public static class Item {
+
+		@Id
+		@Column(name = "ID")
+		private Integer id;
+
+		@Column(name = "ITEM_NAME")
+		private String itemName;
+
+		@OneToOne(mappedBy = "item")
+		private Order order;
+
+		public Item() {
+		}
+
+		public Item(Integer id, String itemName) {
+			this.id = id;
+			this.itemName = itemName;
+		}
+
+		public Integer getId() {
+			return id;
+		}
+
+		public Order getOrder() {
+			return this.order;
+		}
+
+		@Override
+		public boolean equals(Object o) {
+			if ( this == o ) {
+				return true;
+			}
+			if ( o == null || getClass() != o.getClass() ) {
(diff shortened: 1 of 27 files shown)


beikov (commented):
I'm fine with not throwing the error anymore but it is a kind of safety net for users.

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/compliance/nativequery/NativeQueryTest.java
```diff
@@ -0,0 +1,193 @@
...
+					entityManager.persist( item1 );
+					entityManager.persist( item2 );
+					entityManager.persist( item3 );
+
+					Order order1 = new Order( 1, 25.0D, item1 );
+					entityManager.persist( order1 );
+					Order order2 = new Order( 2, 125.0D, item2 );
+					entityManager.persist( order2 );
+					Order order3 = new Order( 3, 150.0D, item3 );
+					entityManager.persist( order3 );
```
beikov, line 51: IMO we should challenge the TCK test as this only works by "accident". If we change the id values of `Order` or `Item`, the test would fail because the JDBC driver will always return position 1 for the column alias "ID", whereas we should clearly use position 4 when trying to lookup the "ID" for `Item`.

I can see us reading out the whole `String[]` aliases from the `ResultSetMetaData` and assume that every "result entity" in the `@SqlResultSetMapping#entities` array "consumes" a whole block i.e. we start searching for the next position by alias for `Item` at index 4, because `Order` "consumed" everything from index 1 to 3.

I don't think the spec says anything in particular about this, but if that is the intended behavior, I'm fine with it, though it should definitely be clarified.

@sebersole or @scottmarlow can you create a challenge for this?

scottmarlow: @beikov could you please link to the failing TCK test result that you are asking if we could challenge.  I'm not yet understanding what the challenge would be exactly and for which TCK test.  

Also, is there an improvement that should be made to the Persistence TCK tests to better support a particular part of the Persistence specification?

dreab8: @scottmarlow the test to be challanged is `com/sun/ts/tests/jpa/core/annotations/nativequery/Client.java#nativeQueryTest1`

scottmarlow: Thanks @dreab8, for reference, the test source link is https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/src/com/sun/ts/tests/jpa/core/annotations/nativequery/Client.java#L69

Is there a test change that we would like to see made for Persistence 3.1?  We are pretty close to finishing the TCK changes for `3.1` but if a change is made via a pull request this week, that likely could be merged (after review + some testing).  Whomever creates that pull request, doesn't have to build the TCK (I can handle the building part on https://ci.eclipse.org/jakartaee-tck and let you know the download link for Persistence 3.1 TCK that we build).  

Or you could also try your change locally by running the Persistence 3.0 TCK  locally, update the test source, change into the test source folder (e.g. `cd src/com/sun/ts/tests/jpa/core/annotations/nativequery`) and build via `ant compile build package`.  Then run the tests in that folder (e.g. `ant runclient`).  This assumes that you have the environment setup to run tests (e.g. likely via `gradle` instead of `ant` if your using the same scripting as before).

Copied from the test source, the test assertions are a combination of Persistence SPEC + JavaDoc requirements:
```
@assertion_ids: PERSISTENCE:SPEC:1009; PERSISTENCE:JAVADOC:199;
   * PERSISTENCE:JAVADOC:64; PERSISTENCE:JAVADOC:41; PERSISTENCE:SPEC:736;
```

```
PERSISTENCE:SPEC:1009:
<assertion required="true" impl-spec="false" defined-by="technology"
                   status="active" testable="true">
            <id>PERSISTENCE:SPEC:1009</id>
            <description>The SqlResultSetMapping annotation is used to specify
                the mapping of the result of a
                native SQL query or stored procedure.
            </description>
            <location chapter="10" section="4.4"/>
        </assertion>
```

```
PERSISTENCE:SPEC:736
       <assertion required="true" impl-spec="true" defined-by="technology"
                   status="active" testable="true">
            <id>PERSISTENCE:SPEC:736</id>
            <description>The SQL query facility is intended to provide support
                for those cases where it is necessary to
                use the native SQL of the target database in use (and/or where
                Java Persistence QL cannot be used).
                Native SQL queries are not expected to be portable across
                databases.
            </description>
            <location chapter="3" section="10.16"/>
        </assertion>
```

```
PERSISTENCE:JAVADOC:199
<assertion required="true" impl-spec="false" status="active" testable="true">
      <id>PERSISTENCE:JAVADOC:199</id>
      <description>Specifies the result set mapping to entities.</description>
      <package>jakarta.persistence</package>
      <class-interface>SqlResultSetMapping</class-interface>
      <method name="entities" return-type="jakarta.persistence.EntityResult[]" />
    </assertion>
```

```
PERSISTENCE:JAVADOC:64
 <assertion required="true" impl-spec="false" status="removed" testable="false">
      <id>PERSISTENCE:JAVADOC:640</id>
      <description>if the query is found not to be a Java Persistence query language SELECT query or a Criteria API query</description>
      <package>jakarta.persistence</package>
      <class-interface>TypedQuery.{jakarta.persistence.Query}</class-interface>
      <method name="getLockMode" return-type="jakarta.persistence.LockModeType">
        <throw>java.lang.IllegalStateException</throw>
      </method>
    </assertion>
```

```
PERSISTENCE:JAVADOC:41
 <assertion required="true" impl-spec="false" status="active" testable="true">
      <id>PERSISTENCE:JAVADOC:41</id>
      <description>Create an instance of Query for executing a native SQL query.</description>
      <package>jakarta.persistence</package>
      <class-interface>EntityManager</class-interface>
      <method name="createNativeQuery" return-type="jakarta.persistence.Query">
        <parameters>
          <parameter>java.lang.String</parameter>
          <parameter>java.lang.String</parameter>
        </parameters>
      </method>
    </assertion>
```

Regarding the process of how to file TCK challenges, guidance is given via [the TCK Process doc](https://jakarta.ee/committees/specification/tckprocess/).  See sections `Who can file a challenge?`, `Valid Challenges`, `Invalid Challenges`, `Filing a Challenge`.

Also see TCK Challenges list from the past few years](https://github.com/eclipse-ee4j/jakartaee-tck/projects/2).

beikov: So this is kind of a question that I would like to ask or ask for clarification in the spec/tck.

In 3.10.16.1. the spec mentions this particular example that the test refers to and says

>When the column names of the SQL result do not correspond to those of the object/relational mapping
>metadata, more explicit SQL result mapping metadata must be provided to enable the persistence
>provider runtime to map the JDBC results into the expected objects. This might arise, for example,
>when column aliases must be used in the SQL SELECT clause when the SQL result would otherwise
>contain multiple columns of the same name or when columns in the SQL result are the results of
>operators or functions. The FieldResult annotation element within the EntityResult annotation is used
>to specify the mapping of such columns to entity attributes.

The example being

```
Query q = em.createNativeQuery(
  "SELECT o.id, o.quantity, o.item, i.id, i.name, i.description " +
  "FROM Order o, Item i " +
  "WHERE (o.quantity > 25) AND (o.item = i.id)",
  "OrderItemResults");
@SqlResultSetMapping(name="OrderItemResults", entities={
  @EntityResult(entityClass=com.acme.Order.class),
  @EntityResult(entityClass=com.acme.Item.class)
})
```

Both `o.id` and `i.id` select items would get the column alias `id`, so AFAIU, according to the specification, this would require the introduction of an alias in the SQL and the use of `@FieldResult` to disambiguate which result set index should be for which attribute. Yet this TCK test expects this to work and I think that this might be just by accident, because in the test data the corresponding item for an order has the same id. So the test doesn't necessarily clarify or assert how this case should be supported.

Should JPA providers assume that a result set column can only be "consumed" for a single attribute mapping i.e. the first encounter of a column label `id` is consumed for the `Order#id` attribute and the second one for `Item#id`?

beikov: FYI I created https://github.com/eclipse-ee4j/jpa-api/issues/350

sebersole: @beikov I believe we discussed this one and decided to look at the table name from the ResultSetMetaData as well to help solve this?

beikov:
Closing this as we will wait for the challenge to be accepted: https://github.com/eclipse-ee4j/jpa-api/issues/350

PR #4638, opened 2022-01-13
Throw IllegalArgumentException for null Criteria literal value

dreab8:


hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmCriteriaNodeBuilder.java
```diff
@@ -914,7 +914,7 @@ void checkMultiselect(List<? extends JpaSelection<?>> selections) {
 	@Override
 	public <T> SqmLiteral<T> literal(T value) {
 		if ( value == null ) {
-			return new SqmLiteralNull<>( this );
+			throw new IllegalArgumentException( "literal value cannot be null" );
```
sebersole, line 917: What do you think about protecting this with `JpaCompliance#isQueryComplianceEnabled`?

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmCriteriaNodeBuilder.java
```diff
@@ -914,6 +914,9 @@ void checkMultiselect(List<? extends JpaSelection<?>> selections) {
 	@Override
 	public <T> SqmLiteral<T> literal(T value) {
 		if ( value == null ) {
+			if ( jpaComplianceEnabled ) {
+				throw new IllegalArgumentException( "literal value cannot be null" );
+			}
```
beikov, line 919: Shouldn't we maybe throw an error in general since there is a special `nullLiteral` method?

sebersole: I mean what is the real reason to not accept null here?  Aside from the spec saying we should not and the TCK checking it

beikov: To avoid potential erroneous passing of a null value and due to that producing an untyped null literal. It's no big deal maybe, but I'd prefer to align with the JPA spec here.

PR #4650, opened 2022-01-17
Fix CriteriaBuidler#locate() arguments order rendering

dreab8:


beikov (commented):
Great catch, but wouldn't it be better to align with the signature of the JPA spec and change the function definitions in the dialects to match that?

beikov (approved):
Looks great!

dreab8:
@beikov thanks for the help!

gavinking:
Whooaaah, wait, this change doesn't look correct to me. 

According to the JPA spec:

> The LOCATE function returns the position of a given string within a string, starting the search at a spec- ified position. It returns the first position at which the string was found as an integer. _The first argument is the string to be located; the second argument is the string to be searched;_

That is: the _first argument_ is the pattern, the second argument is the string we're searching for the pattern in.

That's what I had implemented, I believe.

beikov:
Yeah, we are reverting this already with https://github.com/hibernate/hibernate-orm/pull/4657

IMO the spec text is a bit hard to read for a non-native speaker. I had assumed it was the other way around.

PR #4657, opened 2022-01-18
Wip/6.0 tck locate 2

dreab8:


hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/IngresDialect.java
```diff
@@ -279,10 +279,16 @@ public void initializeFunctionRegistry(QueryEngine queryEngine) {
...
 
 		final BasicType<Integer> integerType = queryEngine.getTypeConfiguration().getBasicTypeRegistry()
 				.resolve( StandardBasicTypes.INTEGER );
+		queryEngine.getSqmFunctionRegistry().registerBinaryTernaryPattern(
+				"locate",
+				integerType,
+				"position(?1 in ?2)",
+				"(position(?1 in substring(?2 from ?3))+(?3)-1)",
+				STRING, STRING, INTEGER
+		).setArgumentListSignature("(pattern, string[, start])");
```
beikov, line 291: IMO this change was correct. According to the docs(https://docs.actian.com/ingres/10.2/index.html#page/SQLRef/String_Functions.htm#ww417057) the position function returns a 1 based index, just like the JPA spec requires it.

hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQLDialect.java
```diff
@@ -452,10 +452,10 @@ public void initializeFunctionRegistry(QueryEngine queryEngine) {
 		queryEngine.getSqmFunctionRegistry().registerBinaryTernaryPattern(
 				"locate",
 				queryEngine.getTypeConfiguration().getBasicTypeRegistry().resolve( StandardBasicTypes.INTEGER ),
-				"position(?2 in ?1)",
-				"(position(?2 in substring(?1 from ?3))+(?3)-1)",
+				"position(?1 in ?2)",
+				"(position(?1 in substring(?2 from ?3))+(?3)-1)",
 				FunctionParameterType.STRING, FunctionParameterType.STRING, FunctionParameterType.INTEGER
-		).setArgumentListSignature("(string, pattern[, start])");
+		).setArgumentListSignature("(pattern, string[, start])");
```
beikov, line 458: I think this one here should use `CommonFunctionFactory.locate_positionSubstring` as well

beikov: I just tried this out on PostgreSQL and the pattern `(position(?1 in substring(?2 from ?3))+(?3)-1)` is correct, so I actually think that `CommonFunctionFactory.locate_positionSubstring` might be wrong because it misses the `-1` at the end.

hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
```diff
@@ -1591,10 +1580,10 @@ public static void locate_positionSubstring(QueryEngine queryEngine) {
 		queryEngine.getSqmFunctionRegistry().registerBinaryTernaryPattern(
 						"locate",
 						queryEngine.getTypeConfiguration().getBasicTypeRegistry().resolve( StandardBasicTypes.INTEGER ),
-						"position(?2 in ?1)", "(position(?2 in substring(?1 from ?3))+?3)",
+						"position(?1 in ?2)", "(position(?1 in substring(?2 from ?3))+?3)",
```
beikov, line 1583:
```suggestion
						"position(?1 in ?2)", "(position(?1 in substring(?2 from ?3))+(?3)-1)",
```

PR #4658, opened 2022-01-18
HHH-15022 Test deleting all entities of a given type in a table-per-class hierarchy with a default schema

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15022

It turns out the problem only affects 5.6, not 6.0, but I figured I would still add a non-regression test in 6.0


yrodiere:
Merging, since tests pass and  #4659 was approved.

PR #4659, opened 2022-01-18
Backport HHH-15022 to branch 5.6 - Test and fix deleting all entities of a given type in a table-per-class hierarchy with a default schema

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15022

It turns out the problem only affects 5.6, not 6.0. So the fix is only present in this PR, not in #4658


beikov (approved):
LGTM!

yrodiere:
I changed the test to execute with postgresql only, because the bug report was for postgresql and because we need to specify a schema name, which is inherently vendor-specific.

beikov:
Thanks!

PR #4663, opened 2022-01-18
Fix expression type for arithmetic operations

dreab8:
for `SqmBinaryArithmetic` i think we do not need the following code
```
final SqmExpressable<?> sqmExpressable = QueryHelper.highestPrecedenceType(
				expression.getLeftHandOperand().getNodeType(),
				expression.getRightHandOperand().getNodeType()
		);
```
to determine the correct expression type, when we built it we already resolve the correct type 
```
(SqmExpressable<T>) domainModel.getTypeConfiguration().resolveArithmeticType(
						lhsOperand.getNodeType(),
						rhsOperand.getNodeType(),
						operator
				)
```


beikov:
`ASTParserLoadingTest#testArithmetic` is failing for MariaDB

PR #4666, opened 2022-01-19
HHH-15042: add offset-clause usage for DB2zDialect

Atgoogat:
DB2z in Version 12 and beyond supports the "offset"-clause.

Current behavior:
Hibernate uses the "fetch first ..."-clause but does not use "offset" for paging requests.

Expected behavior:
For DB2z Databases in Version 12 and beyond the dialect should use the now supported "offset"-clause

IBM Docs:
https://www.ibm.com/docs/en/db2-for-zos/12?topic=subselect-offset-clause
https://www.ibm.com/docs/en/db2-for-zos/12?topic=subselect-fetch-clause

Test:
Because there is no DB2z Docker-DB testing this comes down to testing against an z/OS DB2 Instance.
I tested it against the DB2z Instance (v12.1.0) at work without any issues.


beikov:
Awesome, thanks!

PR #4670, opened 2022-01-20
HHH-15044 Revert HHH-14826 fix because the provided test was wrong

dreab8:
https://hibernate.atlassian.net/browse/HHH-15044


hzerai:
https://hibernate.atlassian.net/browse/HHH-15486

PR #4671, opened 2022-01-20
clean up very old Dialect javadoc

gavinking:
and squash a couple of warnings


gavinking:
The failure on Sybase can’t possibly be related:

PR #4674, opened 2022-01-20
remove check for deprecated setting

gavinking:
I don't think we need to look for `org.hibernate.ejb.HibernatePersistence` anymore.


gavinking:
Anyone object to this change??

beikov (approved):

PR #4678, opened 2022-01-20
kill ExternalSessionFactoryConfig

gavinking:
this is no longer used


gavinking:
This class was removed by @sebersole in another commit.

PR #4681, opened 2022-01-20
Some deprecation removals

jrenaat:


hibernate-core/src/main/java/org/hibernate/tool/schema/internal/AbstractSchemaValidator.java
```diff
@@ -187,13 +187,13 @@ protected void validateSequence(Sequence sequence, SequenceInformation sequenceI
 			);
 		}
 
-		if ( sequenceInformation.getIncrementSize() > 0
-				&& sequence.getIncrementSize() != sequenceInformation.getIncrementSize() ) {
+		if ( sequenceInformation.getIncrementValue() > 0
```
beikov, line 190: Since the original method handled null by returning 0, I think you should use the following here to be safe.

```suggestion
		if ( sequenceInformation.getIncrementValue() != null && sequenceInformation.getIncrementValue() > 0
```

beikov (commented):
Other than the little issue I found, this looks good to me!

gavinking:
Yup, big +1 on cleaning up deprecated things.

sebersole (commented):
:+1:

PR #4685, opened 2022-01-21
HHH-15046/fix db2z i dialect init error

Atgoogat:
During DB2zDialect and DB2iDialect initialization the private property “version” might be 
used resulting in an NullpointerException.

For DB2zDialect this is happening when the ColumnType “TIMESTAMP_WITH_TIMEZONE” is being registered in the overriden Method columnType().

For DB2iDialect this is happing because the createUniqueDelegate() Method is override and uses the private property. 
 
Expected Behavior is the correct initialization of both Dialects without errors.

As a fix I merged the version property of the class DB2Dialect and the private version of DB2zDialect/DB2iDialect.
I added two TestCases which reproduce the issues.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
@@ -6,7 +6,6 @@
  */
 package org.hibernate.dialect;
 
-import jakarta.persistence.TemporalType;
 import org.hibernate.LockOptions;
 import org.hibernate.boot.model.TypeContributions;
 import org.hibernate.dialect.function.CommonFunctionFactory;
@@ -55,7 +54,13 @@ import org.hibernate.type.descriptor.java.PrimitiveByteArrayJavaType;
 import org.hibernate.type.descriptor.jdbc.*;
 import org.hibernate.type.descriptor.jdbc.spi.JdbcTypeRegistry;
 
-import java.sql.*;
+import java.sql.CallableStatement;
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Types;
+
+import jakarta.persistence.TemporalType;
 
 import static org.hibernate.type.SqlTypes.*;
 
--- a/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.dialect;
 
+import jakarta.persistence.TemporalType;
 import org.hibernate.LockOptions;
 import org.hibernate.boot.model.TypeContributions;
 import org.hibernate.dialect.function.CommonFunctionFactory;
@@ -54,13 +55,7 @@ import org.hibernate.type.descriptor.java.PrimitiveByteArrayJavaType;
 import org.hibernate.type.descriptor.jdbc.*;
 import org.hibernate.type.descriptor.jdbc.spi.JdbcTypeRegistry;
 
-import java.sql.CallableStatement;
-import java.sql.DatabaseMetaData;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Types;
-
-import jakarta.persistence.TemporalType;
+import java.sql.*;
 
 import static org.hibernate.type.SqlTypes.*;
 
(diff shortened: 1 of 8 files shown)


gavinking:
Yeah, um, I guess that makes sense.

beikov:
I guess it might be better to introduce a `getDB2Version` in `DB2Dialect` which also returns version, but in the subclasses `DB2zDialect` and `DB2iDialect` it returns a constant version.

Atgoogat:
Can you explain to me why it is better to use an constant version for db2z?
And which constant version to use?

gavinking:
@beikov well, I'm not sure: depends how different DB2 is on the various platforms. If features generally get introduced in same-numbered versions, what's proposed here is OK.

beikov:
The versions for DB2i, DB2z and LUW are completely different and have different features, so I think that the minimum overlap really is the version 9 of LUW.

Atgoogat:
I changed the behavior as you suggested.

The `DB2SqlAstTranslator` now has to check if the Dialect is actually a DB2Dialect.
If it is the version is checked via getDB2Version() otherwise with getVersion().

gavinking:
> The versions for DB2i, DB2z and LUW are completely different and have different features

Ah, OK, fair enough.

hibernate-core/src/main/java/org/hibernate/dialect/DB2SqlAstTranslator.java
```diff
@@ -296,4 +290,12 @@ protected void visitReturningColumns(MutationStatement mutationStatement) {
...
 	}
 
+	private DatabaseVersion getDialectVersion() {
+		final var dialect = getDialect();
+		if (dialect instanceof DB2Dialect) {
+			return ((DB2Dialect) dialect).getDB2Version();
+		}
+		return dialect.getVersion();
+	}
+
```
gavinking, line 300: This approach looks a little bit uglier than the one we took for `MySQLDialect` vs `MariaDBDialect`/`TiDBDialect`.

I would prefer it if you could just follow that pattern mainly just for consistency.

(Assuming they're functionally equivalent, of course.)

gavinking (changes requested):

hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
```diff
@@ -54,13 +55,7 @@
...
 import org.hibernate.type.descriptor.jdbc.spi.JdbcTypeRegistry;
 
-import java.sql.CallableStatement;
-import java.sql.DatabaseMetaData;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Types;
-
-import jakarta.persistence.TemporalType;
+import java.sql.*;
```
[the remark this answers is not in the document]
Atgoogat: sorry missed that. I will fix it soon.

hibernate-core/src/main/java/org/hibernate/dialect/DB2SqlAstTranslator.java
Atgoogat: The problem for MariaDB and MySql is differently in that both dialects use a separate `AbstractSqlAstTranslator` Implementation so there are aware of the version differences between MySql and MariaDB.

Currently the `DB2SqlAstTranslator` is used as the parent for the `DB2zSqlAstTranslator` and `DB2iSqlAstTranslator`.
It might be a better alternative to set the version for `DB2SqlAstTranslator`  in the constructor and not get it from the dialect directly.
What do you think of that?

gavinking: Um, OK, that sounds fine to me, not sure what @beikov thinks of that.

Atgoogat: I keep this as unresolved until we have his feedback.

beikov: >It might be a better alternative to set the version for DB2SqlAstTranslator in the constructor and not get it from the dialect directly.
>What do you think of that?

Sounds good to me.

Atgoogat: Alright. I will change this later this day.

gavinking: Thanks @Atgoogat, apologies for nitpicking.

Atgoogat: @gavinking No problem! I learned a lot and had fun :)

Atgoogat: @beikov I rethought this and saw that the problem might be solved exactly like in the belonging dialects by overriding the `getDB2Version()` method.

I now have changed it like that and think this is better than setting this in the constructor. Is this acceptable?

hibernate-core/src/main/java/org/hibernate/dialect/DB2iDialect.java
```diff
@@ -45,17 +43,17 @@ public DB2iDialect() {
...
-		super();
-		this.version = version;
+		super(version);
 	}
 
-	public DatabaseVersion getIVersion() {
-		return version;
+	@Override
+	public DatabaseVersion getDB2Version() {
+		return DatabaseVersion.make(9, 0);
```
beikov, line 51: I would appreciate if you could move these `DatabaseVersion.make(9, 0)` into a static final constant

hibernate-core/src/main/java/org/hibernate/dialect/DB2SqlAstTranslator.java
beikov: Yeah, that's fine

beikov (approved):

beikov:
Thanks a lot!

PR #4686, opened 2022-01-21
remove hibernate.jdbc.use_streams_for_binary config property

gavinking:
This should be handled by the `Dialect`s.


gavinking:
I've now also removed `hibernate.bytecode.enforce_legacy_proxy_classnames` and some unused methods from `Environment`.

gavinking:
The failure on HSQL is not related, so merging.

PR #4689, opened 2022-01-21
api, spi, internal, deprecation

sebersole:
`org.hibernate.query`

- improved organization
- Javadoc


sebersole:
Integrated upstream

PR #4690, opened 2022-01-21
Remove several deprecations in org.hibernate.boot

jrenaat:


jrenaat:
Checked the failures locally on hsqldb and mariadb, and at these for those db's they don't seem related to my commit (ran the tests without it).

gavinking (approved):
LGTM, merge it!

beikov (approved):
LGTM. You just need to resolve the conflicts.

sebersole (commented):
:+1: as well.  Let's apply it as soon as the jobs complete after your rebase

hibernate-core/src/main/java/org/hibernate/cfg/Settings.java
```diff
@@ -194,7 +194,7 @@ public boolean isJtaTrackByThread() {
 	}
 
 	public boolean isStrictJPAQLCompliance() {
-		return sessionFactoryOptions.isStrictJpaQueryLanguageCompliance();
+		return sessionFactoryOptions.getJpaCompliance().isJpaCacheComplianceEnabled();
```
gavinking, line 197: @jrenaat pretty sure this should have been `isJpaQueryComplianceEnabled()` instead of `isJpaCacheComplianceEnabled()`, but don't worry: nothing calls this method so I think I'm just going to remove it.

PR #4697, opened 2022-01-24
Implement SQM copying for JPA Criteria API uses

beikov:
This is necessary for JPA TCK compliance. We might put the copying of the SQM behind a new config property or the JPA compliance flag, but that needs more discussion.

https://hibernate.atlassian.net/browse/HHH-12563


hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -248,11 +250,11 @@ public QuerySqmImpl(
 			SharedSessionContractImplementor producer) {
 		super( producer );
 		this.hql = CRITERIA_HQL_STRING;
-		this.sqm = criteria;
+		this.sqm = criteria.copy( SqmCopyContext.simpleContext() );;
```
sebersole, line 253: This should be opted-in to.  As a user, I'd hate to incur this overhead just because the JPA spec says this should be supported.  It is an awful expectation imo

sebersole (commented):
I had the one comment which we already spoke about on Zulip

documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
```diff
@@ -79,6 +79,15 @@ Traditionally, Hibernate has considered the names locally scoped.
...
 with the same name will cause a `java.lang.IllegalArgumentException` to be thrown at boot time.
 
+`*hibernate.jpa.compliance.criteria_copy*` (e.g. `true` (default value) or `false` )::
+The Jakarta Persistence spec says that mutations done to `CriteriaQuery`, `CriteriaUpdate` and `CriteriaDelete`
+after such objects were used to create a `jakarta.persistence.Query` may not affect that query.
+This requirement makes it necessary to copy these objects because the APIs allow mutations.
++
+If disabled, it is assumed that users do not mutate the criteria query afterwards
+and due to that, no copy will be created, which will improve performance.
+By default, the copy compliance is enabled because the Jakarta Persistence Criteria API is mutable.
```
sebersole, line 89: This is a tough one actually in terms of default.  I hate implicitly opting a user in to behavior we know will be worse performance.  Not to mention, this would be the first compliance setting we enable by default.

And really, how many people do this?  Create a criteria, use it to create a Query and then mutate the Criteria...

hibernate-core/src/main/java/org/hibernate/jpa/internal/MutableJpaComplianceImpl.java
```diff
@@ -92,6 +93,12 @@ public MutableJpaComplianceImpl(Map configurationSettings, boolean jpaByDefault)
 				configurationSettings,
 				jpaByDefault
 		);
+
+		criteriaCopyCompliance = ConfigurationHelper.getBoolean(
+				AvailableSettings.JPA_CRITERIA_COPY_COMPLIANCE,
+				configurationSettings,
+				true
```
sebersole, line 100: Right, here is the crux of my comment about the default.  All of the others default to `jpaByDefault`.  I believe this should be the same

hibernate-core/src/main/java/org/hibernate/jpa/spi/JpaCompliance.java
```diff
@@ -153,4 +153,20 @@
 	 * @since 6.0
 	 */
 	boolean isLoadByIdComplianceEnabled();
+
+	/**
+	 * JPA says that mutations done to {@link jakarta.persistence.criteria.CriteriaQuery},
```
sebersole, line 158: This should be reworded in my opinion based on the default change

sebersole (commented):
Looks really good aside from my comment/question about the default

hibernate-core/src/main/java/org/hibernate/jpa/internal/MutableJpaComplianceImpl.java
beikov: Hmm, if you think this the right thing to do, then I'll do it. I'll direct the users with complaints about this to you then 😜

PR #4699, opened 2022-01-24
Remove some deprecations in org.hibernate.cfg

jrenaat:


gavinking:
Excellent, I noticed that one.

hibernate-core/src/test/java/org/hibernate/orm/test/mappingexception/MappingExceptionTest.java
```diff
@@ -111,15 +111,6 @@ public void testNotFound() throws MappingException, MalformedURLException {
...
 
-		try {
-			cfg.addResource( "nothere", getClass().getClassLoader() );
-			fail();
-		}
-		catch (org.hibernate.boot.MappingNotFoundException e) {
-			assertEquals( SourceType.RESOURCE, e.getOrigin().getType() );
-			assertEquals( "nothere", e.getOrigin().getName() );
-		}
-
```
beikov, line 122: Why was this test removed?

hibernate-core/src/test/java/org/hibernate/orm/test/mappingexception/MappingExceptionTest.java
```diff
@@ -211,15 +202,6 @@ public void testInvalidMapping() throws MappingException, IOException {
...
 
-		try {
-			cfg.addResource( resourceName, getClass().getClassLoader() );
-			fail();
-		}
-		catch (InvalidMappingException inv) {
-			assertEquals( "resource", inv.getType() );
-			assertEquals( resourceName, inv.getPath() );
-		}
-
```
beikov, line 222: Why was this test removed?

hibernate-core/src/test/java/org/hibernate/orm/test/mappingexception/MappingExceptionTest.java
jrenaat: The Configuration.addResource() method that accepts a ClassLoader was removed since it was deprecated, so testing this call didn't seem to make sense any more

hibernate-core/src/test/java/org/hibernate/orm/test/mappingexception/MappingExceptionTest.java
jrenaat: See the comment for the removal above

hibernate-core/src/test/java/org/hibernate/orm/test/mappingexception/MappingExceptionTest.java
beikov: Right, but the previous implementation just ignored the passed class loader, so you could just omit the class loader in these two tests as well.

jrenaat: If you mean that there should be a call to Configuration.addResource (String resourceName), that's still there in the test ... unless i'm misunderstanding your intention ... ?

beikov: You are right, I didn't see the call in the lines before this. It's fine!

beikov (approved):

PR #4700, opened 2022-01-25
Fix race condition that allowed Component.getType() to return null

TheCodePope:
This fix prevents an NPE in org.hibernate.mapping.SimpleValue.isValid() and likely elsewhere


beikov:
Thanks for the fix!

PR #4701, opened 2022-01-25
actually remove EntityTuplizer and friends

gavinking:
I really can't think of a good reason to leave these zombie classes hanging around if they don't actually *work* anymore.

@sebersole?


beikov (approved):
LGTM

sebersole (approved):

PR #4702, opened 2022-01-25
change getNationalizationSupport() on Sybase to use UNSUPPORTED

gavinking:
`getNationalizationSupport()` currently returns `IMPLICIT` for the jTDS driver. But from the comment I _think_ it should probably return `UNSUPPORTED`.

WDYT @sebersole @beikov ?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/SybaseDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/SybaseDialect.java
@@ -196,7 +196,9 @@ public class SybaseDialect extends AbstractTransactSQLDialect {
 	@Override
 	public NationalizationSupport getNationalizationSupport() {
 		// At least the jTDS driver doesn't support this
-		return jtdsDriver ? NationalizationSupport.IMPLICIT : super.getNationalizationSupport();
+		return jtdsDriver
+			? NationalizationSupport.UNSUPPORTED
+			: super.getNationalizationSupport();
 	}
 
 	@Override


beikov:
The comment meant to say that the jTDS driver doesn't support the `getNString`/`setNString` methods, but `getString`/`setString` seem to work fine since the database does support the `NCHAR`/`NVARCHAR` types. I thought that was exactly the use case for `IMPLICIT`, no?

gavinking:
Well if you turn on `hibernate.use_nationalized_character_data` Hibernate _uses_ `setNString()` / `getNString()`, so I think that counts as "not supported", right?

I mean, I guess we could put a `catch` in `NVarcharJdbcType`, but it's not there today.

gavinking:
> I thought that was exactly the use case for `IMPLICIT`, no?

Not exactly, no. With `IMPLICIT` it will use `VARCHAR` and `CHAR` for DDL generation, which on SQL Server don't (by default) accommodate Unicode.

beikov:
Hmm, then it might be better to register special JdbcTypes for NChar, NVarchar and NClob when we detect the jTDS driver.

gavinking:
Well I'm not sure that's quite right either: at least it's quite inconsistent.

I think what you're reaching for is a new enumeration value `IMPLICIT_FOR_JDBC_EXPLICIT_FOR_SQL`.

gavinking:
See https://github.com/hibernate/hibernate-orm/pull/9193.

PR #4707, opened 2022-01-26
get rid of last usages of getMetamodel()

gavinking:
and remove some unnecessary methods with no callers.

cc @sebersole


hibernate-core/src/main/java/org/hibernate/engine/spi/SessionFactoryDelegatingImpl.java
```diff
@@ -195,34 +193,6 @@ public IdentifierGenerator getIdentifierGenerator(String rootEntityName) {
...
-				.getCollectionDescriptor( role );
-	}
-
-	@Deprecated
-	public Map<String, CollectionPersister> getCollectionPersisters() {
-		// for the time being, leave this calling `MetamodelImplementor` to avoid
-		// creating the map
-		return delegate.getMetamodel().collectionPersisters();
-	}
-
```
gavinking, line 225: @sebersole IMO, `SessionFactoryDelegatingImpl` doesn't need to declare new public methods :)

gavinking: (I mean "new" in the sense of not inherited from its supertypes.)

gavinking: OTOH, perhaps I'm wrong. Not sure. If they're needed, they need docs explaining why.

gavinking:
I have split this PR into two bits and moved the stuff I wasn't sure about to #4710 so Steve can check it.

PR #4711, opened 2022-01-26
CriteriaQuery#getParameters() should not return parameters internally …

dreab8:
…created because value handling mode is bind

Not sure about the name of the new class I have introduced `ValueBindJpaCriteriaParameter` neither if it is necessary, we can obtain the same just filering the `JpaCriteriaParameter` having `name == null`


beikov (commented):
IMO we should move the `value` from `JpaCriteriaParameter` to `ValueBindJpaCriteriaParameter` which should be pretty easy.

dreab8:
@dreab8 what do you think about https://github.com/hibernate/hibernate-orm/pull/4715 ?

beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/4715

PR #4717, opened 2022-01-26
Fix various failures with inline value handling mode

dreab8:


sebersole:
Rebasing this to resolve the conflict and then I'll apply upstream

sebersole:
Applied this upstream

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -4042,46 +4033,41 @@ else if ( expressible instanceof EmbeddableValuedModelPart ) {
...
-			}
-			else {
-				final EntityIdentifierMapping identifierMapping = entityValuedModelPart.getEntityMappingType()
-						.getIdentifierMapping();
-				associationKeyPart = identifierMapping;
-				associationKey = identifierMapping.getIdentifier(
-						literal.getLiteralValue(),
-						null
-				);
-			}
```
beikov, line 4065: Note to self, check why removing this is necessary. The LHS of a comparison `cb.equal( root.get( "assoc" ), cb.literal( myEntity ) )` will use the FK part so this change IMO might lead to comparing wrong columns.

PR #4726, opened 2022-01-27
TCK - Fix String type parameters coercion

dreab8:


beikov (approved):
LGTM but I'd personally like to remove the coercion to avoid errors such as these :D

dreab8:
@beikov I'm closing this PR, the issue is not what I thought, the "substring" function accpets an Integer as second parameter, so it seems the issue is the function parameter type determination.

PR #4727, opened 2022-01-27
HHH-15057 Reproducer for exception about order on many-to-many list

yrodiere:
See https://hibernate.atlassian.net/browse/HHH-15057
See https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/topic/List.20are.20no.20longer.20bags.20by.20default.3F


yrodiere:
Superseded by https://github.com/hibernate/hibernate-orm/pull/4737

PR #4730, opened 2022-01-27
Fix function parameters type determination

dreab8:


hibernate-core/src/main/java/org/hibernate/query/sqm/tree/jpa/ParameterCollector.java
```diff
@@ -95,6 +99,28 @@ public Object visitNamedParameterExpression(SqmNamedParameter<?> expression) {
...
+						( (SqmVisitableNode) e ).accept( this );
+					}
+				}
+		);
+		if ( sqmFunction instanceof SqmAggregateFunction<?> ) {
+			final SqmPredicate filter = ( (SqmAggregateFunction<?>) sqmFunction ).getFilter();
+			if ( filter != null ) {
+				filter.accept( this );
+			}
+		}
```
beikov, line 118:
```suggestion
		super.visitFunction( sqmFunction );
```

beikov (commented):
With the small fixes I proposed this looks good to me

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/jpa/ParameterCollector.java
dreab8: done

beikov (approved):

PR #4733, opened 2022-01-28
HHH-15058 Add unwrap() method to SelectionQuery/MutationQuery

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15058

I did it for `SelectionQuery` as well mostly for consistency: there's no real use to unwrap such a query right now, except maybe for integrators (but those can just cast the query anyway).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/CommonQueryContract.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/CommonQueryContract.java
@@ -17,6 +17,7 @@ import org.hibernate.Session;
 
 import jakarta.persistence.FlushModeType;
 import jakarta.persistence.Parameter;
+import jakarta.persistence.PersistenceException;
 import jakarta.persistence.TemporalType;
 
 /**
@@ -544,4 +545,15 @@ public interface CommonQueryContract {
 	 * @return {@code this}, for method chaining
 	 */
 	CommonQueryContract setProperties(@SuppressWarnings("rawtypes") Map bean);
+
+	/**
+	 * Allows unwrapping this query as another, more specific type.
+	 *
+	 * @param type The class of the object to be returned.
+	 * This is normally an interface implemented by the underlying query implementation.
+	 * @return The unwrapped query.
+	 * @throws RuntimeException If the given type is not supported.
+	 */
+	<T> T unwrap(Class<T> type);
+
 }
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmSelectionQueryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmSelectionQueryImpl.java
@@ -18,6 +18,7 @@ import java.util.function.Supplier;
 import jakarta.persistence.FlushModeType;
 import jakarta.persistence.LockModeType;
 import jakarta.persistence.Parameter;
+import jakarta.persistence.PersistenceException;
 import jakarta.persistence.TemporalType;
 
 import org.hibernate.CacheMode;
@@ -26,6 +27,7 @@ import org.hibernate.HibernateException;
 import org.hibernate.LockMode;
 import org.hibernate.LockOptions;
 import org.hibernate.ScrollMode;
+import org.hibernate.engine.query.spi.EntityGraphQueryHint;
 import org.hibernate.engine.spi.LoadQueryInfluencers;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
@@ -555,6 +557,18 @@ public class SqmSelectionQueryImpl<R> extends AbstractSelectionQuery<R> implemen
 		putIfNotNull( hints, HINT_FOLLOW_ON_LOCKING, getQueryOptions().getLockOptions().getFollowOnLocking() );
 	}
 
+	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	// unwrap
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public <T> T unwrap(Class<T> cls) {
+		if ( cls.isInstance( this ) ) {
+			return (T) this;
+		}
+
+		throw new PersistenceException( "Unrecognized unwrap type [" + cls.getName() + "]" );
+	}
 
 	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 	// covariance
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/query/sqm/BasicSelectionQueryTests.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/query/sqm/BasicSelectionQueryTests.java
@@ -6,16 +6,19 @@
  */
 package org.hibernate.orm.test.query.sqm;
 
-import jakarta.persistence.Basic;
-import jakarta.persistence.Entity;
-import jakarta.persistence.Id;
-import jakarta.persistence.NamedQuery;
-import jakarta.persistence.Table;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.fail;
 
 import org.hibernate.ScrollMode;
 import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.graph.spi.AppliedGraph;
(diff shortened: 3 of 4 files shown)


gavinking:
I don't really understand the issue description, and I'm not clear why this is necessary. I mean, we don't declare `unwrap()` on other Hibernate interfaces, that I know of, we always just inherit it from the JPA interfaces. The intent of `unwrap()` is always to go from JPA API to Hibernate API for programs which require portability across JPA providers.

But for a program that is already using *Hibernate* interfaces, what's the advantage in using `unwrap()` compared to just typecasting? I mean, your implementation of `unwrap()` is quite literally just a typecast.

gavinking:
> go from JPA API to Hibernate API for programs which require portability across JPA providers.

I expressed that very nonsensically, but I think you know what I mean. It's a JPA thing, not a Hibernate thing.

yrodiere:
> But for a program that is already using Hibernate interfaces, what's the advantage in using unwrap() compared to just typecasting?

I agree, and I mentioned it's not very useful. The real point was exposing `unwrap` on `MutationQuery`, which is a `Query` under the hood. Its implementation supports converting to `Query` (basically just a typecast), which is why I needed and thus what I tested, but [other things](https://github.com/hibernate/hibernate-orm/blob/da00adf0da1d427b1bd320f2eccdbb6e6d073b2a/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java#L1167-L1198) too.

As to the advantage of `unwrap` over typecasting... personally I've always interpreted it as "if I cast and it works, there's no guarantee it will work in the next version (the implementation could change); if I unwrap and it works, there *is* a guarantee it will work". But unfornately, that's not how it's implemented in Hibernate ORM (anymore?).

FWIW, that's the approach I took in Hibernate Search, though: https://github.com/hibernate/hibernate-search/blob/9372a6788f033f6ec68182df28ecb0543a861afd/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ElasticsearchClientImpl.java#L91-L98

> I mean, we don't declare unwrap() on other Hibernate interfaces, that I know of, we always just inherit it from the JPA interfaces. The intent of unwrap() is always to go from JPA API to Hibernate API for programs which require portability across JPA providers.

Counter-examples:

* `org.hibernate.service.spi.Wrapped` and `org.hibernate.engine.jdbc.connections.spi.ConnectionProvider`
* `org.hibernate.boot.MetadataBuilder#unwrap`
* `org.hibernate.boot.SessionFactoryBuilder#unwrap`
* `org.hibernate.engine.jdbc.dialect.spi.DialectResolutionInfo#unwrap`
* `org.hibernate.sql.ast.tree.expression.Expression#unwrap`

So, we do use this pattern in non-JPA code; the fact that `SelectionQuery`/`MutationQuery` are native interfaces should not prevent us from adding an `unwrap` method.

> I don't really understand the issue description, and I'm not clear why this is necessary

I mentioned it's mostly for use in legacy code.

An example in Hibernate Search: I have something like below. It's used in a test util that deletes entities; sometimes a HQL delete won't work so we do a select and then iterate on the results to delete them one by one. This method creates the `Query`.

```
	private static Query<?> createSelectOrDeleteAllOfSpecificTypeQuery(EntityType<?> entityType, Session session,
			QueryType queryType) {
		StringBuilder builder = new StringBuilder( QueryType.SELECT.equals( queryType ) ? "select e " : "delete " );

		builder.append( "from " ).append( entityType.getName() ).append( " e" );
		Class<?> typeArg = null;
		if ( hasEntitySubclass( session.getSessionFactory(), entityType ) ) {
			// We must target the type explicitly, without polymorphism,
			// because subtypes might have associations pointing to the supertype,
			// in which case deleting subtypes and supertypes in the same query
			// may fail or not, depending on processing order (supertype before subtype or subtype before supertype).
			builder.append( " where type( e ) in (:type)" );
			typeArg = entityType.getJavaType();
		}
		Query<?> query = QueryType.SELECT.equals( queryType )
				? session.createQuery( builder.toString(), entityType.getJavaType() )
				: session.createQuery( builder.toString() );
		if ( typeArg != null ) {
			query.setParameter( "type", typeArg );
		}
		return query;
	}
```

It's a bit ugly, but the point is to have the hql definition and parameter definition in the same place, and to reuse it for both select and delete queries.

With the API we have in ORM 6, I can no longer create a mutation query that implements `Query`, unless I use deprecated methods.

With `unwrap`, we can.

gavinking:
> I've always interpreted it as "if I cast and it works, there's no guarantee it will work in the next version (the implementation could change); if I unwrap and it works, there is a guarantee it will work". 

I have to say that this is the first time I've heard that. So that's interesting.

IIRC, when we introduced `unwrap()` in JPA 1 it was there because some containers needed to proxy the EM and EMF when injecting them, so direct typecasts weren't going to be possible. I don't think there was anything more behind it than that.

But that reasoning doesn't apply to `Query`, which is never proxied.

Now, on the other hand, I'm not sure on the history of how `unwrap()` ended up on those other interfaces you pointed to. I will have to take a look at who calls those unwrap() methods and for what purpose. (Perhaps they're designed to be called generically or something.) It would not surprise me if at least some of those `unwrap()` methods are unnecessary, so I will check.

In defense of your interpretation, `unwrap()` does have the advantage that some internal refactorings could in principle be hidden from the user, but I question how useful that is in practice.

Really, if you want the client to be able to get a Foo from a Bar in a safe way, and you're committing to support that in future, then you should add a `getFoo()` method to `Bar`. `unwrap()` is just a pretty trashy way to implement that. 

(OK, I'll give you that *perhaps* it's a defensible design for layer breaking.)

I mean, if you ever decide to remove the ability to get a `Foo`, it's much more friendly to your users to remove the `getFoo()` method so that their compiler will tell them what you've done, than let them find out at runtime!

yrodiere:
> Really, if you want the client to be able to get a Foo from a Bar in a safe way, and you're committing to support that in future, then you should add a getFoo() method to Bar. unwrap() is just a pretty trashy way to implement that.
> 
> (OK, I'll give you that perhaps it's a defensible design for layer breaking.)
> 
> I mean, if you ever decide to remove the ability to get a Foo, it's much more friendly to your users to remove the getFoo() method so that their compiler will tell them what you've done, than let them find out at runtime!

Fair enough.

In this specific case, I don't think we have to worry about layers since `MutationQuery` and `Query` (the Hibernate one) are pretty much in the same layer.

So, should I add a `MutationQuery#getQuery()` method instead of the `unwrap` here? The name seems a bit confusing, maybe `asQuery()`/`getAsQuery()`? Though the name `Query` is so generic, I'm afraid it will always be confusing :/

That being said, if the goal of introducing `MutationQuery` was to stop exposing DML as a `Query` where you can do nonsensical stuff such as calling `setFirstResult()`, then maybe I'm committing a heresy by introducing a way to converting `MutationQuery` to `Query` again :)

In that case the use case I provided above would simply be deprecated and ultimately won't be supported anymore, except through JPA interfaces.

gavinking:
> `org.hibernate.service.spi.Wrapped ` and `org.hibernate.engine.jdbc.connections.spi.ConnectionProvider`

Yeah, OK, the Javadoc of the methods you pointed to says  _"Unproxy the service proxy"_, so this appears to be about proxying, like what I was saying above.

> `org.hibernate.boot.MetadataBuilder#unwrap`

This one seems to be entirely pointless, it's only used to typecast to `MetadataBuilder` to `MetadataBuilderImplementor` in tests the tests for one feature of Hibernate. The tests can kust use a typecast.

> `org.hibernate.boot.SessionFactoryBuilder#unwrap`

This one has no callers.

> `org.hibernate.engine.jdbc.dialect.spi.DialectResolutionInfo#unwrap`

This one is used in three places to cast `DialectResolutionInfo` to `DatabaseMetaData`. I think it would be better as a simple getter method. I might propose that change.

> `org.hibernate.sql.ast.tree.expression.Expression#unwrap`

This seems is the least useful of the lot. It's callers should be using typesafe getters.

So look, these really don't seem to be fantastic examples.

gavinking:
> In this specific case, I don't think we have to worry about layers since `MutationQuery` and `Query` (the Hibernate one) are pretty much in the same layer.

Yep.

gavinking:
Why do you want to get a `Query` from a `MutationQuery`?

I still don't understand.

If you want a `Query`, why not just call the `createQuery()` method which returns a `Query`.

(If you're worried about the `@Deprecated`, just cast to `EntityManager` first`.)

gavinking:
> maybe I'm committing a heresy by introducing a way to convert `MutationQuery` to `Query` again :)

Yeah, I guess that's the idea. I mean, if `MutationQuery` has value, its value is that it doesn't expose operations that make no sense.

yrodiere:
Alright, wrapping it up then. Cast it is. Thanks for having a look.

PR #4737, opened 2022-01-28
HHH-15057 JdbcSQLException: NULL not allowed for column "BOOKS_ORDER" when inserting into many-to-many list

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15057

Explanation (copied from the ticket):

> See https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/topic/List.20are.20no.20longer.20bags.20by.20default.3F
> 
> The problem seems to be that we add an order column for both sides of the association, which is not supported (even when using `@OrderColumn` explicitly, even in ORM 5.x).
>
> So the non-owning side of a many-to-many association will have to remain a bag, independently from whatever is configured as the default list semantics.
> 
> From what I’ve seen this will have to be implemented in `org.hibernate.cfg.annotations.CollectionBinder#determineCollectionClassification(java.lang.Class<?>, org.hibernate.annotations.common.reflection.XProperty, org.hibernate.boot.spi.MetadataBuildingContext)`

I decided to keep applying the default list semantics to the owning side of the many-to-many association.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/annotations/CollectionBinder.java
@@ -557,6 +557,11 @@ public abstract class CollectionBinder {
 					|| property.isAnnotationPresent( OrderBy.class ) ) {
 				return CollectionClassification.BAG;
 			}
+			ManyToMany manyToMany = property.getAnnotation( ManyToMany.class );
+			if ( manyToMany != null && ! StringHelper.isEmpty( manyToMany.mappedBy() ) ) {
+				// We don't support @OrderColumn on the non-owning side of a many-to-many association.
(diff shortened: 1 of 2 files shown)


yrodiere:
@dreab8 Hopefully this solves the TCK failure with a bi-directional many-to-many :)

dreab8:
@yrodiere not super happy that we cannot support the order on both side for a bidirectional manytomany but not sure if/how we can implement it, so I would say the change is ok. I think we have to document this!
@sebersole what do you think?

yrodiere:
@dreab8 Yes unfortunately that's the conclusion we reached there: https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/topic/List.20are.20no.20longer.20bags.20by.20default.3F/near/269744314

beikov (approved):

PR #4738, opened 2022-01-28
Fix Criteria Subquery#getSelection() returns null when it has not any…

dreab8:
… select expression


beikov (approved):

dreab8:
Thanks @beikov for the review

PR #4739, opened 2022-01-28
javadoc about XML format mappings

gavinking:


hibernate-core/src/main/java/org/hibernate/boot/Metadata.java
```diff
@@ -30,9 +30,12 @@
 
 /**
  * Represents the ORM model as determined from all provided mapping sources.
+ * An instance may be obtained by calling {@link MetadataSources#buildMetadata()}.
  *
- * NOTE : for the time being this is essentially a copy of the legacy Mappings contract, split between
- * reading the mapping information exposed here and collecting it via InFlightMetadataCollector
+ * @apiNote For the time being this is essentially a copy of the legacy
```
sebersole, line 35: This blurb should just go away

hibernate-core/src/main/java/org/hibernate/boot/jaxb/internal/XmlSources.java
```diff
@@ -27,7 +27,11 @@
 import org.w3c.dom.Document;
 
 /**
- * Helper for building and handling {@link XmlSource} references
+ * Helper for building and handling {@link XmlSource} references.
```
sebersole, line 30: Either add a note to `#addJar` stating that `orm.xml` files are not handled within the jar file, or change `#addJar` to recognize `orm.xml` files also.

hibernate-core/src/main/java/org/hibernate/boot/registry/BootstrapServiceRegistryBuilder.java
```diff
@@ -24,10 +24,17 @@
 import org.hibernate.service.ServiceRegistry;
 
 /**
- * Builder for {@link BootstrapServiceRegistry} instances.  Provides registry for services needed for
- * most operations.  This includes {@link Integrator} handling and ClassLoader handling.
- *
- * Additionally responsible for building and managing the {@link org.hibernate.boot.registry.selector.spi.StrategySelector}
+ * Builder for {@link BootstrapServiceRegistry} instances.
+ * <p>
+ * An instance of this class may be obtained by instantiations, simply by
```
sebersole, line 29: s/instantiations/instantiation

sebersole: In fact I'd reword this sentence 

`An instance of this class may be obtained by {@linkplain #BootstrapServiceRegistryBuilder instantiation}.`

or 

`An instance of this class may be obtained by instantiation; see  {@link #BootstrapServiceRegistryBuilder}.`

hibernate-core/src/main/java/org/hibernate/boot/registry/BootstrapServiceRegistryBuilder.java
```diff
@@ -24,10 +24,17 @@
...
+ * calling {@code new BootstrapServiceRegistryBuilder()}. Then a new
+ * {@code BootstrapServiceRegistry} may be obtained by calling {@link #build()}.
+ * It should be later destroyed by calling {@link #destroy(ServiceRegistry)}.
```
sebersole, line 32: This should also mention `#enableAutoClose`.

hibernate-core/src/main/java/org/hibernate/cfg/Configuration.java
```diff
@@ -64,21 +64,31 @@
  * using {@link MetadataBuilder} and {@link StandardServiceRegistryBuilder}
  * under the covers.
  * <p>
+ * An instance of {@code Configuration} may be obtained simply by
```
sebersole, line 67: Like earlier, this can be condensed - developers understand that new == instantiation.

hibernate-core/src/main/java/org/hibernate/service/ServiceRegistry.java
```diff
@@ -7,7 +7,22 @@
 package org.hibernate.service;
 
 /**
- * A registry of {@linkplain Service services}.
+ * A registry of {@linkplain Service services}. This interface abstracts
```
sebersole, line 10: 1. You are just listing known implementations.  Javadoc already does this for you.
2. In general, not a fan of linking the same thing multiple times within the same javadoc.  Here it is not as terrible as doing it within the actual javadoc text as `@see` is rendered separately.  But it still adds no value imo

sebersole (commented):
Few comments

gavinking:
OK, I've fixed those things.

hibernate-core/src/main/java/org/hibernate/boot/jaxb/internal/XmlSources.java
sebersole: Unless I am mistaken we never use `#addJar` from JPA bootstrapping.  So this would already be a Hibernate-specific series of calls.

But anyway, it was an either/or statement I made.  I do not think it is a good idea to change the semantic of this - aka just add a note saying `orm.xml` files are not picked up.  

If we want to support `#addJar` + `orm.xml`, probably better at this point to add an overload accepting a boolean.

gavinking: Well, I was wondering if we should add a method something approximately like `addOrmXmlFromJar()` - don't know what to call it, but it would know where to look (in `META-INF`).

PR #4741, opened 2022-01-29
cleanup obsolete code and data structures in EntityPersisters

gavinking:
This is a potential performance gain.


gavinking:
Merging this. I will clean up the commenter code in a couple of weeks when we’re sure about all this.

PR #4742, opened 2022-01-29
Streams are no longer closed by terminal operations

gavinking:
This is no longer true, according to H6 migration guide, so undocument it.

@sebersole is this right?


gavinking:
Folks, is this change correct? Because according to what I understand from the [migration guide](https://github.com/hibernate/hibernate-orm/blob/main/migration-guide.adoc#query-stream), what's currently written in the docs is wrong and has changed. But I don't know how any of this stuff works.

beikov:
Yeah, the docs need to be updated. The change, to not automatically close the underlying resources on terminal operations, was done on purpose as you figured through the migration guide.

gavinking:
Sorry, I only asking whether the change to the docs proposed here was correct.

I wasn't questioning the underlying change.

beikov:
I'd say the doc change is correct, but I prefer deleting stuff instead of commenting it out.

gavinking:
Fine.

PR #4746, opened 2022-01-30
make the contexts passed along during cascading actions typesafe

gavinking:
I have wanted to do this for decades.


gavinking:
Woohoo!

PR #4752, opened 2022-01-31
Throw UnknownSqlResultSetMappingException for non existing StoreProce…

dreab8:
…dure result mapping


beikov (approved):

dreab8:

PR #4755, opened 2022-01-31
EntityManager#find() throws IllegalArgumentException for wrong arguments

dreab8:


beikov (commented):
Shouldn't the test use jpa compliance instead? Maybe we should update the tck-runner with the new compliance settings.

sebersole:
> Shouldn't the test use jpa compliance instead? Maybe we should update the tck-runner with the new compliance settings.

JPA bootstrapping enables compliance by default.  `@Jpa` uses JPA boostrapping (its the main difference ultimately with `@SessionFactory`)

sebersole (commented):
I'm not sure that `CoercionException` should be wrapped as `IllegalArgumentException` in all cases, but I also cannot think of specific cases where `CoercionException` would be thrown that should not be.

dreab8:
My bad it seems the `hibernate.jpa.compliance.load_by_id` is not set to `true` when using `@Jpa` this is causing the issue, so @beikov we have to set this value for the tck-runner

sebersole:
`@Jpa` should really just set the top-level value.

PR #4756, opened 2022-01-31
Implemented NamedQuery creation from Criteria

dreab8:
Not sure this is the correct approach and if it is I think we need to clone the `SqmStatement` used to create the `NamedSqmQueryMemento`

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/internal/NamedObjectRepositoryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/internal/NamedObjectRepositoryImpl.java
@@ -26,6 +26,7 @@ import org.hibernate.query.named.NamedResultSetMappingMemento;
 import org.hibernate.query.spi.QueryEngine;
 import org.hibernate.query.spi.QueryInterpretationCache;
 import org.hibernate.query.sql.spi.NamedNativeQueryMemento;
+import org.hibernate.query.sqm.spi.NamedSqmQueryMemento;
 
 import org.jboss.logging.Logger;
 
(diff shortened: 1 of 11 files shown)


hibernate-core/src/main/java/org/hibernate/query/internal/NamedObjectRepositoryImpl.java
```diff
@@ -36,22 +37,44 @@
 	private static final Logger log = Logger.getLogger( NamedObjectRepository.class );
 
 	private final Map<String, NamedHqlQueryMemento> hqlMementoMap;
+	private final Map<String, NamedSqmQueryMemento> sqmMementoMap;
```
beikov, line 40: IMO it would be better to rename the `NamedHqlQueryMemento` to `NamedSqmQueryMemento` and only have a single map.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -1136,9 +1137,25 @@ public R getSingleResult() {
 	// Named query externalization
 
 	@Override
-	public NamedHqlQueryMemento toMemento(String name) {
+	public NamedQueryMemento toMemento(String name) {
 		if ( CRITERIA_HQL_STRING.equals( getQueryString() ) ) {
-			throw new UnsupportedOperationException( "Criteria-based Query cannot be saved as a named query" );
+			return new NamedSqmQueryMementoImpl(
+					name,
+					getSqmStatement(),
```
beikov, line 1144: Dunno how "safe" we want this to be, but maybe we should copy the statement here?

```suggestion
					getSession().isJpaCriteriaCopyComplianceEnabled() ? getSqmStatement() : getSqmStatement().copy( SqmCopyContext.simpleContext(),
```

hibernate-core/src/main/java/org/hibernate/query/internal/NamedObjectRepositoryImpl.java
dreab8: not sure about this, the two implementations do slightly different things. The code seems a little more readable with the two impl

beikov: You can have two implementations, but the interface can be the same and hence you only need a single map and registration code.

dreab8: what do you think of https://github.com/hibernate/hibernate-orm/pull/4782

beikov: IMO the new PR is much better. Want to close this one now?

dreab8:
Superseded by https://github.com/hibernate/hibernate-orm/pull/4782

PR #4759, opened 2022-02-01
HHH-15063 Add test for orphan removal of OneToOne inside Embeddable

phalasOhpen:
This is an example of failing test for issue https://hibernate.atlassian.net/browse/HHH-15063


beikov:

PR #4762, opened 2022-02-01
EntityManager#find() throws IllegalArgumentException for wrong arguments

dreab8:


hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
```diff
@@ -66,8 +67,8 @@ public void onLoad(
 		}
 
 		final Class<?> idClass = persister.getIdentifierType().getReturnedClass();
-		if ( idClass != null &&
-				!idClass.isInstance( event.getEntityId() ) &&
+		persister.getIdentifierMapping().getJavaType().isInstance( idClass );
```
beikov, line 70: Please remove this line

beikov (approved):
LGTM except for the one line

PR #4769, opened 2022-02-02
mention order set aggregates in docs

gavinking:


documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -1454,6 +1454,22 @@ public void test_hql_aggregate_functions_filter_example() {
 		});
 	}
 
+	@Test
+	public void test_hql_aggregate_functions_within_group_example() {
```
beikov, line 1458: Might be worth adding `@RequiresDialectFeature(DialectFeatureChecks.SupportsStringAggregation.class)` to avoid derby test failures

beikov (approved):
Looks great!

documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
gavinking: Well, we don't use `DialectFeatureChecks` in the userguide tests, so I wanted to see where the failures were and add a `@SkipForDialect`.

PR #1077, opened 2015-09-08
HHH-7355 - Added support to join with of many-to-many relation

Janario:

PR #4772, opened 2022-02-02
HHH-15067 Set add() method as public to allow non-nullable associations

blafond:
See hibernate-reactive issue [#1077](https://github.com/hibernate/hibernate-reactive/issues/1077) 

non-nullable association requires reactive to determine non-nullable transient dependencies in order to perform the insert.  The non-visible `NonNullableTransientDependencies.add(..)` method is preventing access to using this class.


DavideD:
By the way  we would need this in `5.6`

PR #4778, opened 2022-02-03
Fix mutation queries using temporary table not adding discriminator c…

dreab8:
…olumn

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/MultiTableSqmMutationConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/MultiTableSqmMutationConverter.java
@@ -29,6 +29,7 @@ import org.hibernate.query.sqm.tree.predicate.SqmWhereClause;
 import org.hibernate.query.sqm.tree.update.SqmAssignment;
 import org.hibernate.query.sqm.tree.update.SqmSetClause;
 import org.hibernate.sql.ast.spi.SqlAstCreationContext;
+import org.hibernate.sql.ast.spi.SqlAstHelper;
 import org.hibernate.sql.ast.spi.SqlAstProcessingState;
 import org.hibernate.sql.ast.spi.SqlExpressionResolver;
 import org.hibernate.sql.ast.tree.Statement;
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/MultiTableSqmMutationConverter.java
```diff
@@ -220,7 +222,14 @@ public Expression resolveSqlExpression(
...
 		try {
-			return (Predicate) sqmWhereClause.getPredicate().accept( this );
+			final Predicate accept = (Predicate) sqmWhereClause.getPredicate().accept( this );
+			if ( discriminatorPredicate != null ) {
+				Junction junction = new Junction( Junction.Nature.CONJUNCTION );
+				junction.add( accept );
+				junction.add( discriminatorPredicate );
+				return junction;
+			}
+			return accept;
```
beikov, line 232: IMO this should use the existing `org.hibernate.sql.ast.spi.SqlAstHelper#combinePredicates` helper method.

```
return org.hibernate.sql.ast.spi.SqlAstHelper.combinePredicates(
    (Predicate) sqmWhereClause.getPredicate().accept( this ),
    discriminatorPredicate 
);
```

beikov (commented):
Other than the small improvement I suggested, this looks good to me

hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/MultiTableSqmMutationConverter.java
```diff
@@ -220,7 +223,11 @@ public Expression resolveSqlExpression(
 
 		pushProcessingState( restrictionProcessingState, getFromClauseIndex() );
 		try {
-			return (Predicate) sqmWhereClause.getPredicate().accept( this );
+			final Predicate predicate = (Predicate) sqmWhereClause.getPredicate().accept( this );
+			if ( discriminatorPredicate != null ) {
```
beikov, line 227: The null check shouldn't be necessary, the `combinePredicates` method handles this properly

dreab8:

PR #4782, opened 2022-02-03
Implemented NamedQuery creation from Criteria

dreab8:


hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
```diff
@@ -906,13 +909,28 @@ public QueryImplementor createNamedQuery(String name) {
...
+			);
+		}
+		else {
+			selectionQuery = new SqmSelectionQueryImpl<>(
+					(NamedCriteriaQueryMementoImpl) memento,
+					expectedResultType,
+					this
+			);
+
+		}
```
beikov, line 930: Might be worth adding a method `toSelectionQuery` on `NamedSqmQueryMemento` similar to the `toQuery` method we have.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -1008,9 +1007,32 @@ public R getSingleResult() {
 	// Named query externalization
 
 	@Override
-	public NamedHqlQueryMemento toMemento(String name) {
+	public NamedQueryMemento toMemento(String name) {
 		if ( CRITERIA_HQL_STRING.equals( getQueryString() ) ) {
-			throw new UnsupportedOperationException( "Criteria-based Query cannot be saved as a named query" );
+			final SqmStatement sqmStatement ;
+			if ( getSession().isJpaCriteriaCopyComplianceEnabled() ) {
```
beikov, line 1013: Since we already copy in the constructor of this class if the compliance is enabled, my suggestion was more that we copy this if compliance is disabled, to be safe. 

```suggestion
			if ( !getSession().isJpaCriteriaCopyComplianceEnabled() ) {
```

hibernate-core/src/main/java/org/hibernate/query/sqm/spi/NamedSqmQueryMemento.java
```diff
@@ -47,68 +41,68 @@
...
-		public Builder setMaxResults(Integer maxResults) {
+		public NamedSqmQueryMemento.Builder setMaxResults(Integer maxResults) {
 			this.maxResults = maxResults;
 			return this;
 		}
 
-		public NamedHqlQueryMemento createNamedQueryDefinition() {
-			return new NamedHqlQueryMementoImpl(
+		public NamedSqmQueryMemento createNamedQueryDefinition() {
+			return new NamedCriteriaQueryMementoImpl(
```
beikov, line 103: Shouldn't this return either a `NamedHqlQueryMementoImpl` or `NamedCriteriaQueryMementoImpl` depending on whether the `sqmSelectStatement` is set?

beikov (commented):
Other than what I have found, this looks good!

hibernate-core/src/main/java/org/hibernate/query/sqm/spi/NamedSqmQueryMemento.java
dreab8: this Builder is never used I'm going to remove it

beikov (approved):

PR #4784, opened 2022-02-04
Fix Attribute#getJavaType() returs also primitive types

dreab8:


hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/PrimitiveBasicTypeImpl.java
```diff
@@ -0,0 +1,23 @@
...
+package org.hibernate.metamodel.model.domain.internal;
+
+import org.hibernate.type.descriptor.java.JavaType;
+
+public class PrimitiveBasicTypeImpl<J> extends BasicTypeImpl<J> {
+	private final Class<J> javaTypeClass;
+
+	public PrimitiveBasicTypeImpl(JavaType<J> javaType, Class<J> javaTypeClass) {
+		super( javaType );
+		this.javaTypeClass = javaTypeClass;
```
sebersole, line 16: Minor, but would be nice to add an assertion here that `javaTypeClass.isPrimitive()`

sebersole (commented):
LGTM

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/PrimitiveBasicTypeImpl.java
dreab8: Done!

karge-itestra:
@dreab8 Is there an issue associated to this? I am seeing this issue in 5.2 as well, will it be backported? It appeared to me in a curious place: when extracting the values in a tuple, the implementation makes an instanceof check for the given element type and the requested type (the requested type comes from a metamodel attribute), which fails for example for `Long` as actual type and `long` as requested type.

beikov:
There is no issue for this @karge-itestra. If you face a problem in 5.x, please create a JIRA issue with a test case against 5.6, as 5.4 and 5.5 are not maintained anymore.

PR #4785, opened 2022-02-04
Use a simpler solution to limit concurrent Jenkins builds

yrodiere:
See https://issues.jenkins.io/browse/JENKINS-43353?focusedCommentId=414847&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-414847


beikov:
Thanks

PR #4789, opened 2022-02-04
Test and fix for maxindex() / maxelement() issues in select clause

beikov:


gavinking:
@beikov this did not fix the problem. There is still an unnecessary join.

For example:

    select maxindex(eol.listOfBasics) from EntityOfLists eol

results in:

```
select
        (select
            max(l2_0.listOfBasics_ORDER) 
        from
            EntityOfLists_basic l2_0 
        where
            e1_0.id=l2_0.EntityOfLists_id) 
    from
        EntityOfLists e1_0 
    join
        EntityOfLists_basic l1_0 
            on e1_0.id=l1_0.EntityOfLists_id
```

beikov:
Thanks for the notice. I'll look into that.

beikov:
See https://github.com/hibernate/hibernate-orm/pull/4798 for the fix

PR #4790, opened 2022-02-04
HHH-15069 Backwards-incompatible changes in SequenceStyleGenerator (and others) following default_schema changes

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15069

I checked locally and this seems to fix https://github.com/liquibase/liquibase/issues/2456. I don't know if *all* these changes are necessary, but they don't hurt.


beikov (approved):
LGTM

PR #4800, opened 2022-02-08
Reproducer for HHH-15073 Criteria query built from JPA metamodel throws PersistenceException: Specified result type [long] did not match Query selection type [java.lang.Long]

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15073


dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/4801

PR #4805, opened 2022-02-10
Fix LocalTime and OffsetTime parameter binding

dreab8:


beikov (approved):

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/EntityMappingType.java
```diff
@@ -53,7 +53,8 @@
  *
  * @author Steve Ebersole
  */
-public interface EntityMappingType extends ManagedMappingType, EntityValuedModelPart, Loadable, Restrictable, Discriminatable {
+public interface
+EntityMappingType extends ManagedMappingType, EntityValuedModelPart, Loadable, Restrictable, Discriminatable {
```
beikov, line 57: ?

beikov:
Looks like this works now

PR #4813, opened 2022-02-14
HHH-15060 Associations with @NotFound should always be left joined when de-referenced in HQL/Criteria

dreab8:
https://hibernate.atlassian.net/browse/HHH-15060

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
@@ -393,6 +393,7 @@ public class DotNode extends FromReferenceNode implements DisplayableNode, Selec
 		String property = propertyName;
 		final boolean joinIsNeeded;
 
+
 		if ( isDotNode( parent ) ) {
 			// our parent is another dot node, meaning we are being further dereferenced.
 			// thus we need to generate a join unless the association is non-nullable and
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java
```diff
@@ -401,8 +402,14 @@ private void dereferenceEntity(
 			property = parentAsDotNode.propertyName;
 			joinIsNeeded = generateJoin && (
 					entityType.isNullable() ||
-					!isPropertyEmbeddedInJoinProperties( parentAsDotNode.propertyName )
+							!isPropertyEmbeddedInJoinProperties( parentAsDotNode.propertyName )
 			);
+			if ( generateJoin
+					&& implicitJoin
+					&& entityType.isNullable()
+					&& isReferenceToPrimaryKey( parentAsDotNode.propertyName, entityType ) ) {
```
sebersole, line 410: Trying to understand why `isReferenceToPrimaryKey` is needed here.  Given the other conditions, don't we always want to use LEFT?

Specifically, doesn't is-nullable itself indicate we need a left-join?

sebersole: Specifically, I mean isn't `generate && implicitJoin && entityType.isNullable()` enough?

dreab8: i think it can cause issues for some queries, I'm going to add a test for this

PR #4816, opened 2022-02-15
Fix BasicTypeResolution for collections annotated with @MapKeyTemporal

dreab8:


hibernate-core/src/main/java/org/hibernate/cfg/annotations/BasicValueBinder.java
```diff
@@ -521,8 +521,10 @@ private void prepareMapKey(
 			}
 
 			final MapKeyClass mapKeyClassAnn = mapAttribute.getAnnotation( MapKeyClass.class );
-			if ( mapKeyClassAnn != null ) {
-				return (BasicJavaType<?>) typeConfiguration.getJavaTypeRegistry().getDescriptor( mapKeyClassAnn.value() );
+			final MapKeyTemporal mapKeyTemporal = mapAttribute.getAnnotation( MapKeyTemporal.class );
+			if ( mapKeyClassAnn != null && mapKeyTemporal == null ) {
```
beikov, line 525: This change seems a bit ad-hoc. Could the problem maybe be that the temporalType is not taken into account?

beikov (approved):
Awesome, that looks aligned with the other code now :)

dreab8:
Thanks @beikov !

PR #4817, opened 2022-02-15
HHH-15080 Session Refresh can cause StaleObjectStateException when entry in 2LC is reread

wburns:


sebersole:
Minor - added a missing space.

Looks good!  The Oracle job timed-out from the initial run.  Let's let all the jobs run again and see.

sebersole:
The JIRA key problem was my fault.  All tests pass now, so pushing upstream.  Thanks!

PR #4826, opened 2022-02-18
HHH-15084 Remove use of BeanManager.createInjectionTarget(AnnotatedType)

bstansberry:
Issue: https://hibernate.atlassian.net/browse/HHH-15084


beikov (approved):
Seems the method exists since CDI 1.1 so IMO this is fine.

PR #4831, opened 2022-02-22
Add sql type name and size information to BasicType

beikov:
The main purpose of this is to have information about the column sql type and size available at runtime which helps to generate casts that work in more situations. See the added tests for scenarios that would fail if we didn't have the proper size available.

This is also a preliminary step to support array types and maybe also beneficial for supporting struct types. Also see the discussion https://github.com/hibernate/hibernate-orm/discussions/3810 related to this.


dreab8:
LGTM

dreab8 (approved):

dreab8:
once concern is the number of `BasicType` we create

PR #4832, opened 2022-02-23
Make sure Hibernate works with the Jakarta Persistence 3.1 API

beikov:


dreab8:
do not understand the change to `EnhancingClassTransformerImpl`

beikov:
See https://github.com/eclipse-ee4j/jpa-api/pull/313 for the details @dreab8

gradle/base-information.gradle
```diff
@@ -32,8 +32,9 @@ ext {
 		ormVersion = HibernateVersion.fromFile(ormVersionFile, project)
 	}
 
+	def version = System.getProperty( 'jakartaJpaVersion', '3.0.0' )
```
scottmarlow, line 35: Should this be `3.1`?  Note that there is no Persistence `3.0.0` release, that would be `3.0`.

hibernate-core/src/main/java/org/hibernate/boot/xsd/ConfigXsdSupport.java
```diff
@@ -31,11 +31,12 @@
...
 	 * 4: JPA 2.2 (default)
 	 * 5: Jakarta Persistence 3.0
+	 * 6: Jakarta Persistence 3.1
 	 */
-	private static final XsdDescriptor[] xsdCache = new XsdDescriptor[6];
+	private static final XsdDescriptor[] xsdCache = new XsdDescriptor[7];
 
 	public XsdDescriptor latestJpaDescriptor() {
-		return getJPA22();
+		return getJPA30();
```
scottmarlow, line 39: Should this return getJPA31() instead?

hibernate-core/src/main/java/org/hibernate/boot/xsd/LocalXsdResolver.java
```diff
@@ -31,7 +31,7 @@
 public class LocalXsdResolver {
 
 	public static String latestJpaVerison() {
-		return "2.2";
+		return "3.0";
```
scottmarlow, line 34: Should this return "3.1" instead?

gradle/base-information.gradle
beikov: https://search.maven.org/artifact/jakarta.persistence/jakarta.persistence-api/3.0.0/jar

scottmarlow: I'm wrong, the SPEC version is `3.0` but the maven artifact is definitely `3.0.0` as also shown on https://jakarta.ee/specifications/persistence/3.0:
```
Maven coordinates

    [jakarta.persistence:jakarta.persistence-api:jar:3.0.0](https://search.maven.org/artifact/jakarta.persistence/jakarta.persistence-api/3.0.0/jar)
```

beikov:
I'll merge this for now, but we will probably need another small PR when the final 3.1.0 version is released.

PR #4833, opened 2022-02-25
HHH-15060 - Associations with @NotFound should always be left joined when de-referenced in HQL/Criteria

dreab8:
@sebersole this changes the cross join into a left one for associations with `@NotFound`


sebersole:
How come you closed this @dreab8 ?  I am playing with this locally and it at least partially fixes the problem.  I changed it locally to more fully fix the problem

dreab8:
because it works on h2 but not for other dbs, if we really want to change the type of join this is what i think we have to do https://github.com/hibernate/hibernate-orm/pull/4836

PR #4835, opened 2022-02-25
Jandex, Classmate and commons annotations shouldn't be listed as API

Sanne:
A little cleanup of the dependency metadata: let's treat Jandex, Classmate and HCANN as internal dependencies.

### DIFF
--- a/hibernate-core/hibernate-core.gradle
+++ b/hibernate-core/hibernate-core.gradle
@@ -27,9 +27,10 @@ configurations {
 dependencies {
     api libraries.jakarta_jpa
     api libraries.jakarta_jta
-    api libraries.jandex
-    api libraries.classmate
-    api libraries.commons_annotations
+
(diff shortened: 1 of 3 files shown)


Sanne:
hi @sebersole - I suppose this is trivial, just checking if I'm missing a deeper reason? Any objection?

beikov:
I don't know if this is possible. `org.hibernate.boot.spi.MetadataContributor` is an SPI and exposes Jandex. No idea about Classmate or HCANN, but at least Jandex can't be excluded.

beikov:
If you think it is ok that a user will see a compile error when trying to implement the `MetadataContributor` SPI unless the user adds an explicit Jande dependency, then I'm fine with removing that as well. Just wanted to mention it since you were seeking for feedback. I agree that this is an advanced use case, but still not unusual. I've had plenty of user questions in the past where the answer was to use that SPI.

PR #4840, opened 2022-03-01
Add sql type name and size information to selectable mapping

beikov:


hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/MimerSQLDialect.java
```diff
@@ -97,6 +97,21 @@ public String getTypeName(int code, Size size) throws HibernateException {
 //		return 45;
 //	}
 
+	@Override
```
sebersole, line 100: This is a great change!

PR #4843, opened 2022-03-01
Enable additional tests and add changes to query result cache to migration guide

dreab8:


migration-guide.adoc
```diff
@@ -329,6 +329,10 @@ The `Query#iterate()` method has been removed. The alternative is to use one of
 * `Query#getResultStream()`
 * Get the `Iterator` from `List` returned by `Query#list()` / `Query#getResultList()`
 
+[[query-result-cache]]
+== Query result cache
+
+Another change in 6.0 is related to the content of the query result cache, it now holds the entity data values and not just the entity identifier information.
```
sebersole, line 335: The impact is more than this though atm.  It also still does not populate the corresponding "data caches" right?

sebersole: Right, pretty sure it is not.  That is worth mentioning here imo

migration-guide.adoc
```diff
@@ -289,6 +289,12 @@ The `Query#iterate()` method has been removed. The alternative is to use one of
 * `Query#getResultStream()`
 * Get the `Iterator` from `List` returned by `Query#list()` / `Query#getResultList()`
 
+[[query-result-cache]]
+== Query result cache
+
+Another change in 6.0 is related to the query result cache.
+
+When the query result cache is enabled and a query is executed only the query result cache will be populated, not the corresponding entity and collection caches, and the content of the query result cache will contain the entity data values and not just the entity identifier information.
```
sebersole, line 297: The structure of this sentence is strange in English.  I also think it could be a little more descriptive.  How about this...

```suggestion
In previous versions, when the query-cache is enabled and a query returning entities is executed, only the entity identifiers are stored in the query-cache.  If second-level caching is enabled for a returned entity, the entity data is stored in its second-level cache region.

Storing just the identifiers in the query-cache has a major drawback when fetching is defined for the query (dynamic fetch, entity-graph, etc) as it can, and often does, lead to N+1 selects.

Starting in 6.0, we now store the complete set of data for the entity into the query-cache.  This also can have a drawback related to the size of caches.  We plan to address this further in later 6.x releases to allow storing just the identifiers along the lines of the previous behavior.

A related change in 6.0 is the fact that queries no longer populate the second-level cache region for any returned entities or collections.  This is again something that will be addressed in later 6.x releases.
```

Maybe even an example query and using that to better describe what happens?

PR #4848, opened 2022-03-03
Fix @ForeignKeyey(ConstraintMode.NO_CONSTRAINT) throwing EntityNotFou…

dreab8:
…ndException

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
@@ -136,7 +136,7 @@ public class ToOneAttributeMapping
 	 hence this flag is also controlling the default join type.
 	 */
 	private final boolean isKeyTableNullable;
-	private final boolean isConstrained;
+	private final boolean isInternalLoadNullable;
 	private final NotFoundAction notFoundAction;
 	private final boolean unwrapProxy;
 	private final boolean isOptional;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -132,6 +132,7 @@
 	 */
 	private final boolean isKeyTableNullable;
 	private final boolean isConstrained;
+	private final boolean isForeignKeyEnabed;
```
beikov, line 135: Actually, `isConstrained` should be enough. I don't think you need another field, or am I missing something?

dreab8: unfortunately for a `OneToOne` is not, I tried to use it but a I had a lot of failures :(

dreab8: @beikov by the way it seems this solution is not working on top of the latest changes... I'm working on a slightly different solution

beikov (approved):
LGTM

sebersole:
Is this still necessary given the work I did on `@NotFound`?  That work also updated how `ConstraintMode.NO_CONSTRAINT` is handled in certain respects.

dreab8:
I reverted my changes and run the tests, the 2 tests I removed the `@NotImplementedYet` still fail.

beikov:
The fix looks good to me and the description of what happens make sense. Do you still have concerns @sebersole?

sebersole (approved):
LGTM

PR #4849, opened 2022-03-03
Implement support for TimeZoneStorageType.COLUMN and add SqlTypes.TIMESTAMP_UTC as jdbc type for Instant

beikov:


sebersole (approved):
lgtm

PR #4854, opened 2022-03-04
HHH-15107 Update build to make tests pass against JDK19

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15107

The upgrade to Mockito 4 turned out not to be necessary after all, but I guess it doesn't hurt? Mockito 2.x hasn't seen a release since 2.28.2 on May 29th, 2019, so it's probably time to upgrade.


beikov (approved):

Sanne:
Mockito is also based on ByteBuddy. Is it shaded, or do we need to check for any dependency alignment requirements in such cases?

yrodiere:
> Is it shaded

No.

> or do we need to check for any dependency alignment requirements in such cases?

In theory we should, but in practice the dependencies haven't been aligned for years. Mockito 2.19.1 was using 1.8.13...

At least Mockito 4.3.1 uses 1.12.7... which happens to be exactly the version we're using right now.

PR #4855, opened 2022-03-04
Work on HHH-15099

sebersole:
https://hibernate.atlassian.net/browse/HHH-15099

`@NotFound` work


sebersole:
Mainly this is to check cross-db testing, but if y'all wanted to check...

sebersole:
Integrated upstream

PR #4856, opened 2022-03-04
Replace typeNames in Dialect with SqlType and SqlTypeRegistry that dialects contribute into

beikov:


hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
```diff
@@ -127,47 +131,33 @@ public int getDefaultStatementBatchSize() {
...
+		// no tinyint
+		sqlTypeRegistry.addDescriptor( new SqlTypeImpl( TINYINT, "smallint", this ) );
+		// HHH-12827: map them both to the same type to avoid problems with schema update
+		// Note that 31 is the maximum precision DB2 supports
+		sqlTypeRegistry.addDescriptor( new SqlTypeImpl( NUMERIC, "decimal($p,$s)", this ) );
+		sqlTypeRegistry.addDescriptor( new SqlTypeImpl( BLOB, "blob($l)", this ) );
+		sqlTypeRegistry.addDescriptor( new SqlTypeImpl( CLOB, "clob($l)", this ) );
+
+		sqlTypeRegistry.addDescriptor( new SqlTypeImpl( TIMESTAMP_WITH_TIMEZONE, "timestamp($p)", this ) );
+		sqlTypeRegistry.addDescriptor( new SqlTypeImpl( TIME_WITH_TIMEZONE, "time", this ) );
```
gavinking, line 147: Wait, what, what's going on here?

After I went to rather a lot of effort to get rid of all these verbose registrations and replace them by much nicer `switch` statements, you're now rolling all that back?!

sebersole (approved):
lgtm

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
```diff
@@ -95,56 +92,36 @@ public CockroachDialect(DatabaseVersion version, PostgreSQLDriverKind driverKind
...
 			case NVARCHAR:
+			case LONGVARCHAR:
+			case LONGNVARCHAR:
 				return "string($l)";
-
-			case NCLOB:
 			case CLOB:
+			case NCLOB:
+			case LONG32VARCHAR:
+			case LONG32NVARCHAR:
```
gavinking, line 109: _Noooo_, please roll back changes like these.

These are *intentionally* not defined in every `Dialect`, and instead have sensible default interpretations defined elsewhere.

Please stop changing stuff like this.

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -105,78 +107,107 @@ public DerbyDialect() {
...
+				return "long varchar for bit data";
+			case NCHAR:
+				return columnType( CHAR );
+			case NVARCHAR:
+				return columnType( VARCHAR );
+			case LONGNVARCHAR:
+				return columnType( LONGVARCHAR );
+			case LONGVARCHAR:
+			case LONG32VARCHAR:
+				return "long varchar";
```
gavinking, line 141: Like, are you copy/pasting this code, which was previously defined on the superclass (DRY) down to the subclasses?

Please don't do that.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -588,19 +508,6 @@ public int resolveSqlTypeLength(
...
-	 * precision of a float(p) using p expressed in decimal
-	 * digits instead of the usual (standard) binary digits.
-	 */
-	protected static Size binaryToDecimalPrecision(int code, Size size) {
-		return code == Types.FLOAT
-				&& size != null
-				&& size.getPrecision() != null
-				? Size.precision( (int) Math.ceil( size.getPrecision() / LOG_BASE2OF10 ) )
-				: size;
-	}
```
gavinking, line 602: @beikov why is this method no longer needed?

beikov: Moved to `BinaryFloatSqlType`

gavinking: Ah, OK, fine, np. The patch is so big I couldn't find it.

PR #4857, opened 2022-03-07
Fix subquery throwing SqlTreeCreationException( Found un-correlated p…

dreab8:
…ath usage in sub query)


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -2851,7 +2851,8 @@ private TableGroup prepareReusablePath(
 			if ( parentTableGroup == null ) {
 				final TableGroup parent = fromClauseIndex.findTableGroupOnParents( parentPath.getNavigablePath() );
 				if ( parent != null ) {
-					throw new SqlTreeCreationException( "Found un-correlated path usage in sub query - " + parentPath );
```
beikov, line 2854: Is this required by a TCK test? Because if it isn't, I would like to avoid trying to "handle" this scenario automatically. I'm assuming that a `Path` from an outer query can't be used directly within a subquery and must be correlated explicitly.

beikov: Ughh, ok I just saw the failing TCK tests... I can't believe the TCK allows this, but ok, if it requires that.

PR #4858, opened 2022-03-08
HHH-13694 fix numeric overflow exception for large sequence min values

beikov:
https://hibernate.atlassian.net/browse/HHH-13694

6.0 version of https://github.com/hibernate/hibernate-orm/pull/3169


sebersole (commented):
lgtm

PR #4861, opened 2022-03-08
HHH-15105 Fix NPE when access default query cache region statistics

beikov:
https://hibernate.atlassian.net/browse/HHH-15105


dreab8:
LGTM

PR #4867, opened 2022-03-11
HHH-15115 Deleting an entity with Joined inheritance and default schema set is throwing and error

dreab8:
https://hibernate.atlassian.net/browse/HHH-15115

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/test/inheritance/JoinedInheritanceDeletionTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/test/inheritance/JoinedInheritanceDeletionTest.java
@@ -1,0 +1,124 @@
+package org.hibernate.test.inheritance;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Inheritance;
+import javax.persistence.InheritanceType;
+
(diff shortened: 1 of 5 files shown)


yrodiere (approved):
Right, this is the same problem as [HHH-15022](https://hibernate.atlassian.net/browse/HHH-15022), with the same solution.

Sorry about that, and thanks for looking into it! LGTM.

hibernate-core/src/test/java/org/hibernate/test/inheritance/JoinedInheritanceDeletionTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.hibernate.dialect.PostgreSQL81Dialect;
+
+import org.hibernate.testing.RequiresDialect;
+import org.hibernate.testing.TestForIssue;
+import org.hibernate.testing.junit4.BaseCoreFunctionalTestCase;
+import org.junit.Before;
+import org.junit.Test;
+
+@RequiresDialect(PostgreSQL81Dialect.class)
+@TestForIssue( jiraKey = "HHH-15115")
```
Sanne, line 19: ouch these whitespaces are unbalanced - my eyes! :smile:

PR #4868, opened 2022-03-11
HHH-15118 Fix duplicate ids with PooledOptimizer when sequence value is initialValue

GChapel:
https://hibernate.atlassian.net/browse/HHH-15118


beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/4884

PR #4870, opened 2022-03-12
HHH-15115 Fix PostgreSQL SqmMultiTableMutationStrategy

dreab8:
https://hibernate.atlassian.net/browse/HHH-15115


beikov:
I'll look into the failure. I recently fixed something in this area that isn't merged yet.

PR #4871, opened 2022-03-14
HHH-15117 ConstraintViolationException is thrown using same @SecondaryTable on two entities

dreab8:
https://hibernate.atlassian.net/browse/HHH-15117


beikov:
Do not merge this yet please. I worked on something similar and would like to look into this before.

beikov (approved):

PR #4872, opened 2022-03-14
HHH-14993 - EAGER non-inverse collection throws StackOverflowError if `max_fetch_depth` not set

sebersole:
HHH-14993 - EAGER non-inverse collection throws StackOverflowError if `max_fetch_depth` not set


beikov (approved):
LGTM

PR #4878, opened 2022-03-15
HHH-15067: Set add() method as public

DavideD:
Backport for `5.6` (it's already been merged in `main`)

See https://github.com/hibernate/hibernate-orm/pull/4772


DavideD:
Merged.
It shouldn't have an impact on anything but it's very easy to revert if necessary

PR #4891, opened 2022-03-21
HHH-13135 Use FOR NO KEY UPDATE when locking in PostgreSQL

beikov:
https://hibernate.atlassian.net/browse/HHH-13135

As can be seen from the test runs, most databases implement the semantics that PostgreSQL introduced with `FOR NO KEY UPDATE`. I don't know if it makes sense to introduce new `LockMode` entries, but if so, we need the following:

* `WRITE_KEY` or `WRITE_ALL` which will block inserts for FKs pointing to the locked row
* `READ_KEY` which will prevent row deletions or changes of the PK pointed to through FKs

We could also implement the `FOR NO KEY UPDATE` locking for HANA by using `... for update of col1, col2, ...` i.e. by listing all non-primary key columns. Maybe it's even enough to list just a single non-PK column. Also see https://help.sap.com/viewer/4fe29514fd584807ac9f2a04f6754767/2.0.06/en-US/20fcf24075191014a89e9dc7b8408b26.html#loio20fcf24075191014a89e9dc7b8408b26__sql_select_1sql_select_for_update

Wdyt about new `LockMode` entries? And support for this in HANA?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQLSqlAstTranslator.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQLSqlAstTranslator.java
@@ -66,7 +66,13 @@ public class PostgreSQLSqlAstTranslator<T extends JdbcOperation> extends Abstrac
 	}
 
 	@Override
+	protected String getForUpdate() {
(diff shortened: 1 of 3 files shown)


beikov:
We'll look into new lock modes when we have the time to analyze compatibility of locking configurations in a few database to see if WRITE_KEY or READ_KEY could be implemented on databases other than PostgreSQL

PR #4896, opened 2022-03-22
Fixes for final

beikov:
While working on the my talk, I found a few "problems" that are worth fixing before final.


beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/4904 and https://github.com/hibernate/hibernate-orm/pull/4905

A third PR is following for handling select item deduplication in query groups

PR #4898, opened 2022-03-22
Updated documentation as per Jira HHH-15049

asutosh936:


NathanQingyangXu:
It would be great to tweak your git commit message to start with "HHH-15049".

NathanQingyangXu (approved):

asutosh936:
@NathanQingyangXu - Updated commit message to start with `HHH-15049`

NathanQingyangXu:
> @NathanQingyangXu - Updated commit message to start with `HHH-15049`

@beikov could you take a look? Seems a good catch!

beikov:
This is a bit more complicated I think as the default value depends on whether Hibernate is bootstrapped via JPA or via the Hibernate native boot API.

The default value for JPA boot is `false` as you can see here: https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java#L256

Whereas `true` for Hibernate native boot as you can see here: https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java#L320

So IMO we should adapt the documentation to make this clear. Wdyt @sebersole and @dreab8?

dreab8:
@beikov yeap I agree with you that it would be better if the documentation mentioned that for JPA boot this setting is ignored and the value is always false.

asutosh936:
Make sense @beikov @dreab8. In that case should the current be updated as below?
`*hibernate.transaction.flush_before_completion* (e.g. `true` or `false` (Hibernate bootstrap with JPA value will be false) ):: `

NathanQingyangXu:
> This is a bit more complicated I think as the default value depends on whether Hibernate is bootstrapped via JPA or via the Hibernate native boot API.
> 
> The default value for JPA boot is `false` as you can see here: https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java#L256
> 
> Whereas `true` for Hibernate native boot as you can see here: https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java#L320
> 
> So IMO we should adapt the documentation to make this clear. Wdyt @sebersole and @dreab8?

I saw it is set to TRUE by default in https://github.com/hibernate/hibernate-orm/blob/9e9a8135dc705a7fc5e5154e89fd11465f5f6589/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java#L320

Is it true that the default value depends on whether it is in JPA boot mode or Hibernate mode?

NathanQingyangXu:
> Make sense @beikov @dreab8. In that case should the current be updated as below? `*hibernate.transaction.flush_before_completion* (e.g. `true`or`false`(Hibernate bootstrap with JPA value will be false) )::`

Hold on. I think the configuration we are talking about is Hibernate specific and it should be TRUE by default; for JPA compliance configuration, we have a separate suite. Maybe we need to add a new entry as following in the JPA Compliance section (https://docs.jboss.org/hibernate/orm/current/userguide/html_single/Hibernate_User_Guide.html#configurations-jpa-compliance) and set its default value to FALSE.

* hibernate.jpa.compliance.transaction.flush_before_completion

@sebersole how do you think?

dreab8:
@NathanQingyangXu for Jpa bootsrapping the [`EntityManagerFactoryBuilderImpl`](https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java#L256) is executed first and then [`SessionFactoryOptionsBuilder`](https://github.com/hibernate/hibernate-orm/blob/9e9a8135dc705a7fc5e5154e89fd11465f5f6589/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java#L320) so the value of `flushBeforeCompletionEnabled` is always `false`

documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
```diff
@@ -709,7 +709,7 @@ A configuration value key used to indicate that it is safe to cache.
 `*hibernate.jta.cacheUserTransaction*` (e.g. `true` or `false` (default value))::
 A configuration value key used to indicate that it is safe to cache.
 
-`*hibernate.transaction.flush_before_completion*` (e.g. `true` or `false` (default value))::
+`*hibernate.transaction.flush_before_completion*` (e.g. `true` (default value) or `false` )::
 Causes the session be flushed during the before completion phase of the transaction. If possible, use built-in and automatic session context management instead.
```
beikov, line 713: I think the following would be better

```suggestion
`*hibernate.transaction.flush_before_completion*` (e.g. `true` (Hibernate boot default value) or `false` (JPA boot default value) )::
Causes the session be flushed during the before completion phase of the transaction. If possible, use built-in and automatic session context management instead. When Hibernate is link:https://docs.jboss.org/hibernate/orm/current/quickstart/html_single/#hibernate-gsg-tutorial-basic-test[booted through the native API], the default value is `true`, otherwise `false`.
```

gavinking:
We now generate these docs from the javadoc, where this setting is documented correctly.

PR #4901, opened 2022-03-24
Timestampdiff and duration fixes

beikov:
* Introduce an option to control the storage for `Duration` but default to `INTERVAL_SECOND` (with fallback to `NUMERIC`)
* Fix the resolved return type for `timestampdiff(second)` to match JPA 3.1 expectation for `extract(second)`
* Switch back to storing amount of nanoseconds instead of seconds with fractional parts to retain backwards compatibility

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaType.java
@@ -54,7 +54,7 @@ public class DurationJavaType extends AbstractClassJavaType<Duration> {
 
 	@Override
 	public JdbcType getRecommendedJdbcType(JdbcTypeIndicators context) {
-		return context.getTypeConfiguration().getJdbcTypeRegistry().getDescriptor( SqlTypes.INTERVAL_SECOND );
+		return context.getTypeConfiguration().getJdbcTypeRegistry().getDescriptor( context.getPreferredSqlTypeCodeForDuration() );
 	}
 
 	@Override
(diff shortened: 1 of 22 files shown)


gavinking:
As I commented on a different one of these PRs: I think this is really moving us in the wrong direction.

We should not be having OR mapping information specified in config properties. Hibernate has _too many_ toggles and switches. Adding _new ones_&mdash;right after we worked so hard to remove a bunch of the legacy ones&mdash;is, IMO a really really bad idea.

I propose removing this new setting (moving it to an annotation if necessary).

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/DurationJavaType.java
```diff
@@ -120,11 +112,17 @@ public Duration fromString(CharSequence string) {
 		}
 
 		if (value instanceof BigDecimal) {
-			return fromDecimal( value );
+			final BigDecimal decimal = (BigDecimal) value;
+			final BigDecimal[] bigDecimals = decimal.divideAndRemainder( BILLION );
+			return Duration.ofSeconds(
+					bigDecimals[0].longValueExact(),
+					bigDecimals[1].longValueExact()
+			);
```
gavinking, line 120: @beikov why didn't you go back and look at the original version of this code, which had the very informative comment that you deleted:

```
			return Duration.ofSeconds(
					secondsAndNanos[0].longValueExact(),
					// use intValue() not intValueExact() here, because
					// the database will sometimes produce garbage digits
					// in a floating point multiplication, and we would
					// get an unwanted ArithmeticException
					secondsAndNanos[1].intValue()
			);
```

beikov: Thanks for the fix for this. I don't know what the comment means exactly, an example of when anything but `intValue()` fails would be helpful, but I merged your change regardless.

beikov:
>I propose removing this new setting (moving it to an annotation if necessary).

Like I wrote on another PR, let's move the discussion to Zulip which you have started with Steve already.

PR #4910, opened 2022-03-25
Add support for specifying the SqlTypes field name when a type code is expected

beikov:


gavinking:
This is good. Even better would be to have a package-level annotation to specify this default, and deprecate the config property which I really hate.

sebersole:
I am not tied to settings particularly.  But I (dis)like annotations here exactly the same

sebersole:
At the moment, users have 2 options -

1) These settings
2) Supply a `TypeContributor`

(2) is not as trivial as it sounds at first though. `TypeContributor` let's you register a `JavaType`, `JdbcType` or `UserType`.  However, it does not expose a trivial means for registering a "default JDBC type (code) for a given Java type".  To achieve that, as `TypeContributor` is currently defined, users would need to either:

1) Supply a custom UserType that does this mapping
2) Replace the `JavaType` to be able to override `#getRecommendedJdbcType` - and we'd have to argue the "purpose" of this method

sebersole:
Another option would be something like -

```
public interface TypeContributions {
    ...
    void registerJavaJdbcTypeOverride(Class<?> javaType, int jdbcTypeCode);
}
```

+

```
public interface JdbcTypeIndicators {
    ...
    JdbcType getOverridenJdbcTypeFallback(Class<?> javaType);
}
```

PR #4912, opened 2022-03-26
HHH-14672 - Allow specifying CHAR-based storage for UUID mappings

sebersole:


dreab8:
LGTM :+1:

beikov (commented):
IMO we should keep the naming and implementation of the `preferred_xxx_jdbc_type` settings aligned, so I think we should still do some renaming. I took the opportunity and aligned all this here: https://github.com/hibernate/hibernate-orm/pull/4915

Let me know what you think

sebersole:
> Any tips on how I apply your forced commit @beikov ? I pulled from my repo branch but your commit is not there

Oh, interesting... the commit is accessible by commit-id, just not applied to my branch

beikov:
I created a new branch since I didn't want to mess with yours in case you wanted to do something differently

sebersole:
> I created a new branch since I didn't want to mess with yours in case you wanted to do something differently

https://github.com/hibernate/hibernate-orm/commit/c8010a2fd6b2a3c2dfe3e869f7b98c901cf78385

You pushed that branch upstream?

beikov:
It does containyour work and is against upstream directly, though I aligned it with the way we did things for duration and boolean

sebersole:
It seems like you rebased your work and mine together into one commit?  That's odd

beikov:
Yeah sorry for that. I kind of messed up the order of commits that I cherry picked and had to squash in the end

PR #4919, opened 2022-03-28
HHH-15150 Hibernate 6 EmbeddedId CacheKeys are no longer Serializable

wburns:
[Hibernate 6 EmbeddedId CacheKeys are no longer Serializable](https://hibernate.atlassian.net/browse/HHH-15150)

This is resurrecting the old test that was present before to ensure cache keys were serializable. However, the bug is still present in Hibernate 6.


dreab8:
Hi @wburns,
 
I'm going to close this in favour of https://github.com/hibernate/hibernate-orm/pull/4988
Thanks a lot for the test case!

wburns:
> Hi @wburns,
> 
> I'm going to close this in favour of #4988 Thanks a lot for the test case!

Awesome, thanks @dreab8 ! I only made it so it can be used later :)

PR #4920, opened 2022-03-29
Generator increment

dreab8:


beikov (commented):
LGTM

hibernate-core/src/main/java/org/hibernate/mapping/SimpleValue.java
```diff
@@ -434,7 +437,27 @@ public RootClass getRootClass() {
...
+		final ConfigurationService cs = metadata.getMetadataBuildingOptions().getServiceRegistry()
+				.getService( ConfigurationService.class );
+
+		final String idNamingStrategy = cs.getSetting(
+				AvailableSettings.ID_DB_STRUCTURE_NAMING_STRATEGY,
+				StandardConverters.STRING,
+				null
+		);
+
+		if ( LegacyNamingStrategy.STRATEGY_NAME.equals( idNamingStrategy )
```
sebersole, line 449: To me, this just does not feel right.  Shouldn't this be based on (1) `hbm.xml` versus (2) annotations/`orm.xml`?

dreab8: I added tests,may be I'm missing something case, but when using `LegacyNamingStrategy.STRATEGY_NAME` I tried to obtain the same increment as we have in 5.X

sebersole: So do you have a test that shows any combination of settings that causes annotation / `orm.xml` mappings to use increment of 1 on 5.x?

dreab8: for a few cases https://github.com/hibernate/hibernate-orm/pull/4920/files#diff-3fc2829c3bdfbe2d7f4b9a2206d286f3fe586a145df31b773d444ce3e614f1ca

dreab8: not sure it covers enough cases

sebersole: Can you point me to the specific test that uses annotations (or `orm.xml`) and achieves increment = 1 (without actually configuring increment) on 5.x?

dreab8: https://github.com/hibernate/hibernate-orm/pull/4920/files#diff-3fc2829c3bdfbe2d7f4b9a2206d286f3fe586a145df31b773d444ce3e614f1caR63
and
https://github.com/hibernate/hibernate-orm/pull/4920/files#diff-3fc2829c3bdfbe2d7f4b9a2206d286f3fe586a145df31b773d444ce3e614f1caR83
and
https://github.com/hibernate/hibernate-orm/pull/4920/files#diff-3fc2829c3bdfbe2d7f4b9a2206d286f3fe586a145df31b773d444ce3e614f1caR143

sebersole:
So then, :+1:

PR #4922, opened 2022-03-29
Do some renaming for consistency and update documentation for new SQL types

beikov:
Rename `MapKeyCustomCompositeType` to `MapKeyCompositeType` and `MapKeyCustomType` to `MapKeyType` i.e. remove the `Custom` part from the name to be consistent with other annotations.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
@@ -144,9 +144,9 @@ public class DB2Dialect extends Dialect {
 				// Note that 31 is the maximum precision DB2 supports
 				return columnType( DECIMAL );
 			case BLOB:
-				return "blob($l)";
+				return "blob";
 			case CLOB:
-				return "clob($l)";
+				return "clob";
 			case TIMESTAMP_WITH_TIMEZONE:
 				return "timestamp($p)";
 			case TIME_WITH_TIMEZONE:
--- a/documentation/documentation.gradle
+++ b/documentation/documentation.gradle
@@ -61,6 +61,8 @@ dependencies {
 			requireCapability 'org.ehcache.modules:ehcache-xml-jakarta'
 		}
 	}
+	// Needed for JSON tests
+	testRuntimeOnly libraries.jackson
 }
 
 
--- a/documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
+++ b/documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
@@ -392,6 +392,10 @@ Can also specify the name of the constant in `org.hibernate.type.SqlTypes` inste
 Global setting identifying the preferred JDBC type code for storing duration values.
 Can also specify the name of the constant in `org.hibernate.type.SqlTypes` instead.
 
+`*hibernate.type.preferred_instant_jdbc_type*` (e.g. `93` for `java.sql.Types.TIMESTAMP` or `3003` for `org.hibernate.types.SqlTypes.TIMESTAMP_UTC` (default value))::
+Global setting identifying the preferred JDBC type code for storing instant values.
+Can also specify the name of the constant in `org.hibernate.type.SqlTypes` instead.
+
 ==== Bean Validation options
 `*jakarta.persistence.validation.factory*` (e.g. `jakarta.validation.ValidationFactory` implementation)::
 Specify the  `javax.validation.ValidationFactory` implementation to use for Bean Validation.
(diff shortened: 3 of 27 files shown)


gavinking:
@beikov 

I'm against adding global configuration parameters that affect ORM mappings.

IMO, `hibernate.type.preferred_instant_jdbc_type` is a bad addition, and is a good example of the sort of thing we've been actively trying to move away from.

hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
```diff
@@ -144,9 +144,9 @@ protected String columnType(int sqlTypeCode) {
 				// Note that 31 is the maximum precision DB2 supports
 				return columnType( DECIMAL );
 			case BLOB:
-				return "blob($l)";
+				return "blob";
 			case CLOB:
-				return "clob($l)";
+				return "clob";
```
gavinking, line 149: @beikov Why did you make this change.

The commit message doesn't explain why you did this.

Please at least separate it into its own commit.

gavinking:
Wait, you already merged this?!

beikov:
>I'm against adding global configuration parameters that affect ORM mappings.
>IMO, hibernate.type.preferred_instant_jdbc_type is a bad addition, and is a good example of the sort of thing we've been actively trying to move away from.

I guess we can discuss this on Zulip later, but it looks like you already did with Steve. We can certainly switch to a more generic variant for configuring the JDBC/SQL types per Java type, though I think it only makes sense to override the preferred types for certain Java types which can be meaningfully represented through different JDBC types.

hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
beikov: `DB2Dialect`, `DerbyDialect` and `MimerSQLDialect` were the only dialects that specified a length parameter, even though I'm pretty sure other dialects also allow specifying a length. In the end, the length is not really helpful anyway IMO, but the real problem is that we don't have a default length or a way to determine the LOB max size and we ended up with an unresolved variable in a test that I added. So I removed that to be consistent with the other dialects and fix the test I added.

gavinking: > but the real problem is that we don't have a default length or a way to determine the LOB max size and we ended up with an unresolved variable in a test that I added. 

OK, that's a big surprise to me, and I don't know how the existing tests didn't pick it up. But it seems like we should fix _that_ problem, not remove the ability to specify the length of a `blob` in DB2.

Because with the change you've made, if I'm not mistaken, it seems that any length the user explicitly specifies via the `@Length` annotation will now be ignored!

> `DB2Dialect`, `DerbyDialect` and `MimerSQLDialect` were the only dialects that specified a length parameter

Sure, this is all very platform-specific stuff. But if the platform _supports_ a length parameter then we shouldn't just ignore it.

So this looks like the wrong "fix" for the problem you encountered.

beikov: >OK, that's a big surprise to me, and I don't know how the existing tests didn't pick it up. But it seems like we should fix that problem, not remove the ability to specify the length of a blob in DB2.
>Because with the change you've made, if I'm not mistaken, it seems that any length the user explicitly specifies via the @Column annotation will now be ignored!

How about we support length in a unified way across all databases in a follow up fix? We could generate a check constraint to assert the length limit if the type doesn't support specifying capacity.

gavinking: Well I think some databases (Postgres, if I'm not mistaken) simply don't support a maximum length. But  I don't recall any details or this.

gavinking: For MySQL it surely already works correctly.

beikov: >Well I think some databases (Postgres, if I'm not mistaken) simply don't support a maximum length. But I don't recall any details or this.

You are right, but I think these databases usually support check constraints, so we could enforce a length here if we think it is reasonable to do so.

>For MySQL it surely already works correctly.

Not sure what you mean. AFAICS MySQL has no possibility to specify the length for LOB types.

beikov: Let's continue that discussion on this issue I just created: https://hibernate.atlassian.net/browse/HHH-15155

gavinking: > Not sure what you mean. AFAICS MySQL has no possibility to specify the length for LOB types.

It has `blob`, `mediumblob`, `longblob`, etc. I'm sure we already pick the right one based on the specified `length`.

beikov: Right, but there is no length variable in the DDL type "name"/pattern that can specify exact capacity.

PR #4924, opened 2022-03-29
HHH-15133 - Use specified result-type to better infer "shape" of query results with implicit selections

sebersole:
HHH-15133 - Use specified result-type to better infer "shape" of query results with implicit selections
    
    Alternate solution for de-duplication handling

### DIFF
--- a/documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
+++ b/documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
@@ -2447,7 +2447,7 @@ public class HQLTest extends BaseEntityManagerFunctionalTestCase {
 				Object[].class)
 			.getResultList();
 			//end::hql-relational-comparisons-example[]
-			assertEquals(2, phonePayments.size());
+			assertEquals(3, phonePayments.size());
 		});
 	}
 
--- a/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java
@@ -658,6 +658,10 @@ public abstract class AbstractSharedSessionContract implements SharedSessionCont
 
 	@Override
 	public SelectionQuery<?> createSelectionQuery(String hqlString) {
+		return internalCreateSelectionQuery( hqlString, null );
+	}
+
+	private <R> SelectionQuery<R> internalCreateSelectionQuery(String hqlString, Class<R> expectedResultType) {
 		checkOpen();
 		pulseTransactionCoordinator();
 		delayedAfterCompletion();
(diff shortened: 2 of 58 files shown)


sebersole:
Alternate solution to https://github.com/hibernate/hibernate-orm/pull/4890 specific to de-duplication handling

romainmoreau:
Maybe not only it but at a minimum this merge introduced a huge performance loss on HQL queries (and probably Criteria queries if I understood correctly Hibernate 6 new SQM structure) because to my understanding it always checks for row duplication when the result type is an entity which is a performance killer especially when the result set is big.

beikov:
You (@romainmoreau) are right that the filtering will significantly affect performance when the result set is very big, as the worst case performance currently is `O(N!)` which I think is what you might experience when your result set is sorted. Please [create a JIRA issue](https://hibernate.atlassian.net) for this and explain your exact use case there. I think we should make the filtering configurable such that people can opt-out of it. Apart from that, I also think that it might be better to assume data is sorted and start doing the filtering check from the end rather than from the start. On sorted result sets this would lead to a `O(N)` ( `N-1` checks ) complexity.

romainmoreau:
@beikov I agree the complexity is quadratic but I don't think it's just the worst case, it'll always have this complexity based on my understanding of how `withDuplicationCheck` works (and to be sure I also tried to sort my result set both ascending and descending and didn't notice any difference in performance).

So my use case isn't very specific, it's a pretty wide issue I think.

By the way, if anyone's interested, a possible workaround is to use a `Stream` instead of a `List` for the result and call the `toList` method on it (though I'm using a Spring `Repository`, not sure if it's possible for every Hibernate use case).

So that's not a blocking issue unlike another one found after the Hibernate 6 migration (persisting ~10M entities using 16 threads results in the database in ~14M rows, no idea what's the exact reason, probably some concurrency issue somewhere).

beikov:
We have a few options to improve the situation, but the cost is always more memory usage and a higher allocation rate. We could use an identity hash map or a sorted `int[]` where we put the identity hash codes into to deduplicate entries, but overall, this will increase the space complexity by at least `O(N)`, but decrease compute complexity to `O(N)` or `O(N * log(N))`.

Let's see what @sebersole says, but I think ultimately, we will have to benchmark this a bit to see which strategy is best.

>So that's not a blocking issue unlike another one found after the Hibernate 6 migration (persisting ~10M entities using 16 threads results in the database in ~14M rows, no idea what's the exact reason, probably some concurrency issue somewhere).

Either way, please create a JIRA issue for this and also for the other issue you mentioned with some further details.

romainmoreau:
@beikov I pushed some demos for the performance issue and was going to file an issue with those but I'm sure if it's still needed with @dreab8's PR.

- https://github.com/romainmoreau/hibernate-5-performance-demo
- https://github.com/romainmoreau/hibernate-6-performance-demo

For the other issue, I'll need to take some time to investigate.

dreab8:
@romainmoreau 
tried your demo
with orm 6.1.2.Final _Performance test done in 37875 ms_
with my last version of https://github.com/hibernate/hibernate-orm/pull/5201 modified with Christian suggestions _Performance test done in 3192 ms_

ChristianCiach:
@beikov We're affected heavily by this and we've created a Jira issue a while ago: https://hibernate.atlassian.net/browse/HHH-17307

We've attached a reproducer. Is there anything else we can provide to help you to take a look at it?

beikov:
I'm sorry that this is taking so long, but as you can imagine, everyone who reports a problem says that this is critical for them. If the community based best effort support that we provide you for free is not fast enough for you, then you will have to become a Red Hat customer and create a support case so that we can prioritize this task.

ChristianCiach:
It's alright. I was just wondering if there is anything we did wrong in reporting it.

PR #4925, opened 2022-03-30
Documentation updates

beikov:


hibernate-core/src/main/java/org/hibernate/type/StandardBasicTypes.java
```diff
@@ -539,9 +539,8 @@ private StandardBasicTypes() {
 	/**
 	 * The standard Hibernate type for mapping {@link Byte Byte[]} to JDBC {@link org.hibernate.type.SqlTypes#VARBINARY VARBINARY}.
 	 */
-	public static final BasicTypeReference<Byte[]> WRAPPER_BINARY = new BasicTypeReference<>(
-			//TODO find a decent name before documenting
-			"wrapper-binary",
+	public static final BasicTypeReference<Byte[]> BINARY_WRAPPER = new BasicTypeReference<>(
```
sebersole, line 542: I'm not following these changes...

Users would use these names in their mappings.  But their old mappings would no longer match.  Surely I am missing something right?

beikov: Oh yes, you are right. That one was renamed by "mistake", thanks for pointing it out.

sebersole: Its not just one though.  Just look throughout the diff for this file

sebersole: Bottom line, while you may not like the names - they are exposed to users and cannot be changed.  Add a new "synonym" if you wish, but do not change them

PR #4932, opened 2022-04-02
HHH-8713 generated model misses <T>

yafna:
added implementation for the metamodel generator : HHH-8713 generated model misses <T> .


yafna:
Could somebody have a look on this ?

dmitrys-odysseus:
Here, at [OHDSI](https://github.com/OHDSI) we have a few inheritance cases that block us from using metamodel generator.
Can somebody from the maintainers provide feedback for this solution?

YaroslavTir:
I need that as well. What is the status?

gavinking:
The static metamodel consists of `static` field declarations. They simply cannot make use of type variables, not under any circumstances, so this proposal makes no sense. I have commented further on the referenced issues.

PR #4934, opened 2022-04-04
HHH-15162 Add STRING_OR_CLOB FunctionParameterType and use that for various length functions

beikov:
https://hibernate.atlassian.net/browse/HHH-15162


sebersole (commented):
lgtm

PR #4936, opened 2022-04-04
HHH-15167  IllegalArgumentException occurred while calling setter for property 

dreab8:
https://hibernate.atlassian.net/browse/HHH-15167

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
@@ -43,6 +43,7 @@ import org.hibernate.metamodel.mapping.EntityMappingType;
 import org.hibernate.metamodel.mapping.ForeignKeyDescriptor;
 import org.hibernate.metamodel.mapping.JdbcMapping;
 import org.hibernate.metamodel.mapping.ManagedMappingType;
+import org.hibernate.metamodel.mapping.MappingType;
 import org.hibernate.metamodel.mapping.ModelPart;
 import org.hibernate.metamodel.mapping.ModelPartContainer;
 import org.hibernate.metamodel.mapping.PluralAttributeMapping;
@@ -947,12 +948,13 @@ public class ToOneAttributeMapping
 			hasBidirectionalFetchParent = true;
 		}
 		else if ( CollectionPart.Nature.fromNameExact( parentNavigablePath.getLocalName() ) != null ) {
-			referencedNavigablePath = parentNavigablePath.getParent().getParent();
+			referencedNavigablePath = getReferencedNavigablePath( creationState, parentNavigablePath.getParent() );
 			hasBidirectionalFetchParent = fetchParent instanceof Fetch
 				&& ( (Fetch) fetchParent ).getFetchParent() instanceof Fetch;
 		}
 		else {
-			referencedNavigablePath = parentNavigablePath.getParent();
+			referencedNavigablePath = getReferencedNavigablePath( creationState, parentNavigablePath );
+
 			hasBidirectionalFetchParent = fetchParent instanceof Fetch;
 		}
 		// The referencedNavigablePath can be null if this is a collection initialization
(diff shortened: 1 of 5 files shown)


scottmarlow:
This change resolves the WildFly + Jakarta EE 10 Platform TCK test failure jpa/core/criteriaapi/CriteriaBuilder/Client.java\#abs_from_pmservlet + jpa/core/criteriaapi/CriteriaBuilder/Client.java\#abs_from_stateless3 mentioned on https://gist.github.com/scottmarlow/cc287c646faee509b5dc56c916045ea6!  Thanks for addressing this @dreab8 !

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -947,12 +948,13 @@ private Fetch createCircularBiDirectionalFetch(
 			hasBidirectionalFetchParent = true;
 		}
 		else if ( CollectionPart.Nature.fromNameExact( parentNavigablePath.getLocalName() ) != null ) {
-			referencedNavigablePath = parentNavigablePath.getParent().getParent();
+			referencedNavigablePath = getReferencedNavigablePath( creationState, parentNavigablePath );
```
beikov, line 951: Will there be test failures if you use this?

```suggestion
			referencedNavigablePath = getReferencedNavigablePath( creationState, parentNavigablePath.getParent().getParent() );
```

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -1033,6 +1035,23 @@ else if ( CollectionPart.Nature.fromNameExact( parentNavigablePath.getLocalName(
...
+	private NavigablePath getReferencedNavigablePath(
+			DomainResultCreationState creationState,
+			NavigablePath parentNavigablePath) {
+		NavigablePath referencedNavigablePath = parentNavigablePath.getParent();
+		MappingType partMappingType = creationState.resolveModelPart( referencedNavigablePath ).getPartMappingType();
+
+		while ( !( partMappingType instanceof EntityPersister )
+				|| ( partMappingType != entityMappingType
+				&& !entityMappingType.isSubclassEntityName( partMappingType.getMappedJavaType().getJavaType().getTypeName() )
+				&& !( (EntityPersister) partMappingType ).isSubclassEntityName( entityMappingType.getEntityName() ) ) ) {
```
beikov, line 1047: Would be awesome if you could add a little comment with an example to make it easier to understand what is going on here. I'm actually a bit surprised that the model shown in the test triggers an issue. I would assume we have plenty of similar models.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
dreab8: yes this change causes failures

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
dreab8: I'll add a comment

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
dreab8: probably it should be
```
 getReferencedNavigablePath( creationState, parentNavigablePath.getParent())
```

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -140,7 +141,7 @@
 	private final NotFoundAction notFoundAction;
 	private final boolean unwrapProxy;
 	private final boolean isOptional;
-	private final EntityMappingType entityMappingType;
+	private final EntityPersister entityMappingType;
```
sebersole, line 145: This could have been left as EntityMappingType (which EntityPersister extends btw).  I'd rather we start relying on the new contracts when possible.  The persister is also available when needed via `EntityMappingType#getEntityPersister`

dreab8: Reverted back to `EntityMappingType`!

dreab8:
@sebersole @beikov thanks for the review!
@scottmarlow thanks for checking!

PR #4938, opened 2022-04-05
HHH-15165 Add support for SQL geography type

beikov:
https://hibernate.atlassian.net/browse/HHH-15165

### DIFF
--- a/hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/postgis/PGGeographyJdbcType.java
+++ b/hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/postgis/PGGeographyJdbcType.java
@@ -1,0 +1,168 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+package org.hibernate.spatial.dialect.postgis;
+
+import java.sql.CallableStatement;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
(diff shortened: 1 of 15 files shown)


maesenka (commented):
I think we need to remove the code duplication. Apart from that, looks good.

hibernate-spatial/src/main/java/org/hibernate/spatial/dialect/postgis/PGGeographyJdbcType.java
```diff
@@ -0,0 +1,168 @@
...
+				buffer = ByteBuffer.from( pgValue );
+				final WkbDecoder decoder = Wkb.newDecoder( wkbDialect );
+				return decoder.decode( buffer );
+			}
+			else {
+				return parseWkt( pgValue );
+			}
+
+		}
+		throw new IllegalStateException( "Received object of type " + object.getClass().getCanonicalName() );
```
NathanQingyangXu, line 82: should the exception be `IllegalArgumentException`?

hibernate-spatial/src/test/java/org/hibernate/spatial/mapping/GeographyMappingTest.java
```diff
@@ -0,0 +1,118 @@
...
+									(Point<G2D>) Wkt.fromWkt(
+											"SRID=4326;POINT(48.2083736 16.3724441)"
+									)
+							)
+					);
+					s.flush();
+					s.clear();
+
+					PointEntity pointEntity = s.find( PointEntity.class, 1 );
+					assertThat( pointEntity.location, is( notNullValue() ) );
```
NathanQingyangXu, line 71: Is the `notNullValue()` assertion a little bit weak? Should we assert the 2D coordinates instead?

maesenka (approved):

PR #4944, opened 2022-04-06
HHH-15184: Component optimisations

Sanne:
- https://hibernate.atlassian.net/browse/HHH-15184


beikov (approved):
LGTM

sebersole:
Column is a Selectable and both of these ultimately delegate to Value.  It would be possible to build both of these at once and avoid iterating the Selectables twice.

Sanne:
That's fair but I prefer keeping the initialization of both as lazy, as we might never need them.

PR #4952, opened 2022-04-07
Use default Locale for lowercasing.

andygithubmf:
For i18n, the default Locale should be used for lowercasing.  Locale.ROOT will not correctly lowercase Turkish "I"s.


sebersole:
This needs a Jira issue.  That Jira issue (the key at least) should be referenced in the commit message(s).

Thanks.

PR #4961, opened 2022-04-11
Gradle changed uploadDocumentation rsync url and port

dreab8:


hibernate-orm-build/src/main/java/org/hibernate/orm/docs/DocumentationPublishing.java
```diff
@@ -41,7 +41,7 @@ public DocumentationPublishing(Project project) {
 				.convention( project.getLayout().getBuildDirectory().dir( "documentation" ) );
 		docServerUrl = project.getObjects()
 				.property( String.class )
-				.convention( "filemgmt.jboss.org:/docs_htdocs/hibernate/orm" );
+				.convention( "filemgmt-prod-sync.jboss.org:/docs_htdocs/hibernate/orm" );
```
yrodiere, line 44: From what I can see, this URL is used for both `scp` and `rsync`.

I just checked, and it seems you can do `scp` on `filemgmt-prod-sync.jboss.org` too, so... I suppose this will work.

yrodiere (approved):

PR #4970, opened 2022-04-13
HHH-15142 Conversion of CriteriaQuery to TypedQuery with Like predicate and escape character fails when repeated.

dreab8:
https://hibernate.atlassian.net/browse/HHH-15142

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/query/ReuseCriteriaWithMixedParametersTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/query/ReuseCriteriaWithMixedParametersTest.java
@@ -1,0 +1,222 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.jpa.test.query;
+
+import java.time.Instant;
+import java.util.Date;
+import java.util.List;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.Query;
+import javax.persistence.TypedQuery;
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.ParameterExpression;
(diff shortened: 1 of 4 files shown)


beikov (approved):
LGTM

hibernate-core/src/test/java/org/hibernate/jpa/test/query/ReuseCriteriaWithMixedParametersTest.java
```diff
@@ -0,0 +1,140 @@
...
+			final CriteriaQuery<Wallet> criteriaQuery = criteriaBuilder.createQuery( Wallet.class );
+			final Root<Wallet> root = criteriaQuery.from( Wallet.class );
+
+			final ParameterExpression<String> stringValueParameter = criteriaBuilder.parameter( String.class );
+
+			criteriaQuery.where(
+					criteriaBuilder.like(
+							root.get( Wallet_.model ),
+							stringValueParameter,
+							'/'
```
NathanQingyangXu, line 82: I might be wrong; but it seems the reported issue is preconditioned by 'escaped char' (`'\\'), not normal '/' char, no?
Below is the original code snippet:
```
CriteriaQuery<Person> criteriaQuery = personQuery      .where(cb.like(
            root.get("name"),
            pattern,
            cb.literal('\\')
      ));
```

ettavolt: It wasn't a syntax error caused by Criteria→JPQL→internal conversion, it happened during parameter binding.
Two other test cases don't have anything about escaping (the use Instant literals), but still fail before the fix.

dreab8: @NathanQingyangXu , I do not think the gist of the issues is this.

NathanQingyangXu: got it, thanks for letting me know. 
It might be an idea to modify the ticket title and description to perfectly removing all misleading factors, for future bug reporters might get confused if he searched for similar bugs.

dreab8: by the way I added an additional test

kguelzau:
Is there any planned release date for 5.6.9?

PR #4974, opened 2022-04-15
HHH-15211 Fix Embeddable element collection converter bug

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-15211

The root cause is we did not sync up '**equalness**' with '**hash code**' for such edge case (collection of **Embeddable** element to be converted)


dreab8:
LGTM

beikov (approved):
LGTM as well since `isEqual` is also non-final.

dreab8:
Thanks @NathanQingyangXu :+1:

PR #4976, opened 2022-04-18
HHH-15216 Cannot change MetadataProvider implementation because JPAXMLOverriddenMetadataProvider is final and precisely expected by a cast operator

reda-alaoui:
https://hibernate.atlassian.net/browse/HHH-15216

This PR moves `JPAXMLOverriddenMetadataProvider` out of package `internal` and makes it non final.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/BootstrapContextImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/BootstrapContextImpl.java
@@ -31,7 +31,7 @@ import org.hibernate.boot.spi.BootstrapContext;
 import org.hibernate.boot.spi.ClassLoaderAccess;
 import org.hibernate.boot.spi.MetadataBuildingOptions;
 import org.hibernate.cfg.AvailableSettings;
-import org.hibernate.cfg.annotations.reflection.internal.JPAXMLOverriddenMetadataProvider;
+import org.hibernate.cfg.annotations.reflection.JPAXMLOverriddenMetadataProvider;
 import org.hibernate.engine.config.spi.ConfigurationService;
 import org.hibernate.jpa.internal.MutableJpaComplianceImpl;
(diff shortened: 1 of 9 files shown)


reda-alaoui:
Hello?

beikov:
Please rebase the PR and make sure you annotate the classes which you move with `@Internal`.

beikov (commented):
Can you explain why you had to move `PropertyMappingElementCollector`? Also, please squash the commits into a single one.

reda-alaoui:
Hello @beikov ,

> Please rebase the PR and make sure you annotate the classes which you move with @Internal.

This is done.

> Can you explain why you had to move PropertyMappingElementCollector?

I wanted to avoid changing its visibility to public. But as you can see now, I reverted the move and changed its visibility to public.

beikov:
>But as you can see now, I reverted the move and changed its visibility to public.

But why do you need this in the first place? The class is only used inside of `JPAXMLOverriddenAnnotationReader` and you can't interfere with the usage.

Also, there seem to be compilation errors

reda-alaoui:
> But why do you need this in the first place? The class is only used inside of JPAXMLOverriddenAnnotationReader and you can't interfere with the usage.

I need to make it public because the class resides in package internal while I moved one of its consumer to the parent package.
`org.hibernate.cfg.annotations.reflection.JPAXMLOverriddenAnnotationReader` can't access `org.hibernate.cfg.annotations.reflection.internal.PropertyMappingElementCollector` otherwise.

reda-alaoui:
The compilation failure should be fixed.

beikov:
Thanks

reda-alaoui:
Thank you!
Could this be back ported to 5.6.x?

beikov:
If you provide a PR for it, then yeah.

reda-alaoui:
Ok I will asap 👍

PR #4979, opened 2022-04-21
fix some typos and delete some unused imports

NathanQingyangXu:
let me know if JIRA ticket is needed

part of the ongoing endeavor. will create new PRs later.


dreab8:
Hi @NathanQingyangXu , for this one I would say that a Jira is not strictly necessary!

NathanQingyangXu:
> Hi @NathanQingyangXu , for this one I would say that a Jira is not strictly necessary!

I created the ticket here: https://hibernate.atlassian.net/browse/HHH-15220

hibernate-core/src/main/java/org/hibernate/loader/entity/CacheEntityLoaderHelper.java
```diff
@@ -109,7 +109,7 @@ public PersistenceContextEntry loadFromSessionCache(
 	 * of entity hierarchies.
 	 * <p/>
 	 * If checkDeleted was set to true, then if the entity is found in the
-	 * session-level cache, it's current status within the session cache
+	 * session-level cache, its current status within the session cache
```
jrenaat, line 112: this one was actually correct ;)  (in this context the possessive form is required, i.e its, not the shorthand for 'it is')

hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetadataContext.java
```diff
@@ -636,7 +636,7 @@ private <X> void registerAttribute(Class<?> metamodelClass, Attribute<X, ?> attr
 				// mismatch in how the annotation processor interpreted the attribute and how our metamodel
 				// and/or annotation binder did.
 
-//              This is particularly the case as arrays are nto handled propery by the StaticMetamodel generator
+//              This is particularly the case as arrays are nto handled property by the StaticMetamodel generator
```
jrenaat, line 640: I would say that nto should become 'not', and i think that 'propery' was meant to be 'properly'

NathanQingyangXu: thanks. I updated.

PR #4980, opened 2022-04-21
HHH-15219 Re-introduce the ability to stop and then reactivate a ServiceRegistry

Sanne:
https://hibernate.atlassian.net/browse/HHH-15219

This is essentially re-introducing a method which has been "cleaned up" during ORM 6 refactorings. We still need this method to integrate with Quarkus, I'd appreciate simply restoring it as-is.
Clarified the related comments a bit.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/registry/internal/StandardServiceRegistryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/registry/internal/StandardServiceRegistryImpl.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.boot.registry.internal;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/boot/registry/internal/StandardServiceRegistryImpl.java
```diff
@@ -94,6 +95,24 @@ private void applyServiceRegistrations(List<StandardServiceInitiator<?>> service
 		}
 	}
 
+	/**
+	 * Not intended for general use. We need the ability to stop and "reactivate" a registry to allow
+	 * experimentation with technologies such as GraalVM, Quarkus and Cri-O.
+	 */
+	public synchronized void resetAndReactivate(BootstrapServiceRegistry bootstrapServiceRegistry,
+												List<StandardServiceInitiator<?>> serviceInitiators,
+												List<ProvidedService<?>> providedServices,
```
NathanQingyangXu, line 104: should we use generics wildcard to allow collection of child class instances to be able to pass in? Like:
```
public synchronized void resetAndReactivate(BootstrapServiceRegistry bootstrapServiceRegistry,
												List<? extends StandardServiceInitiator<?>> serviceInitiators,
												List<? extends ProvidedService<?>> providedServices,
```

PR #4982, opened 2022-04-22
HHH-15223 - Embeddables in the Query Result, the embeddable instance returned by the query will not be in the managed state 

dreab8:
https://hibernate.atlassian.net/jira/software/c/projects/HHH/issues/HHH-15223


hibernate-core/src/test/java/org/hibernate/orm/test/querycache/PersonName.java
```diff
@@ -76,4 +76,13 @@ public int hashCode() {
...
 	}
+
+	@Override
+	public String toString() {
+		return "PersonName{" +
+				"first='" + first + '\'' +
+				", middle='" + middle + '\'' +
+				", last='" + last + '\'' +
+				'}';
+	}
```
NathanQingyangXu, line 87: Is the above code change to `PersonName` necessary for this PR? Seems it was not used in the testing case.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/embeddable/internal/EmbeddableResultImpl.java
```diff
@@ -41,6 +42,12 @@ public EmbeddableResultImpl(
 			DomainResultCreationState creationState) {
 		super( modelPart.getEmbeddableTypeDescriptor(), navigablePath );
 		this.resultVariable = resultVariable;
+		/*
+			This path is needed to be sure to use a different initializer in case for this path there is registered also an EmbeddableFetchImpl,
+			this because Jakarta spec requires that if the result of a query corresponds to an embeddable,
+			the embeddable instance should not be part of the state of any managed entity. So we need different initializers in order to have different instances.
```
NathanQingyangXu, line 48: seems some minor verbiage issues. E.g.
```
there is registered also an EmbeddableFetchImpl
```
seems to have grammatical error if I didn't misunderstand. Could be changed to:
```
there is also an registered EmbeddableFetchImpl
```

another:
```
this because Jakarta spec ...
```
could be changed to
```
because Jakarta spec ...
```

Overall I tried to create a more compact comment as following:
```
a new path is required for a new initializer for Jakarta spec dictates that if the result of a query corresponds to an embeddable, the embeddable instance should not be part of the state of any managed entity
```
Feel free to improve in any way;  the current comment seems to have room for improvement

sebersole: I'd phrase it like this:

```suggestion
An `{embeddable_result}` sub-path is created for the corresponding initializer to differentiate it from a fetch-initializer if this embedded is also fetched. The Jakarta Persistence spec says that any embedded value selected in the result should not be part of the state of any managed entity.  Using this `{embeddable_result}` sub-path avoids this situation.
```

hibernate-core/src/test/java/org/hibernate/orm/test/querycache/PersonName.java
dreab8: reverted, thanks @NathanQingyangXu

hibernate-core/src/main/java/org/hibernate/sql/results/graph/embeddable/internal/EmbeddableResultImpl.java
dreab8: Thnaks @sebersole , I really like your version :+1:

beikov (approved):
LGTM

dreab8:
applied upstream!

hibernate-core/src/test/java/org/hibernate/orm/test/querycache/PersonName.java
sebersole: :shrug: often these kinds of things are useful in debugger

PR #4983, opened 2022-04-22
HHH-15225 - Jta transactions,  name queries with errors throw IllegalStateException instead of IllegalArgumentException when transaction is not active

dreab8:
https://hibernate.atlassian.net/browse/HHH-15225


beikov (approved):
LGTM

PR #4985, opened 2022-04-25
HHH-15178 Backport Jenkinsfile and GH actions

beikov:
Backport for https://hibernate.atlassian.net/browse/HHH-15178 also requires backports of
* https://hibernate.atlassian.net/browse/HHH-14434
* https://hibernate.atlassian.net/browse/HHH-14938
* https://hibernate.atlassian.net/browse/HHH-3326
* https://hibernate.atlassian.net/browse/HHH-13779


hibernate-core/src/main/java/org/hibernate/cfg/Configuration.java
```diff
@@ -721,7 +721,14 @@ public SessionFactory buildSessionFactory(ServiceRegistry serviceRegistry) throw
...
 		log.debug( "Building session factory using internal StandardServiceRegistryBuilder" );
 		standardServiceRegistryBuilder.applySettings( properties );
-		return buildSessionFactory( standardServiceRegistryBuilder.build() );
+		StandardServiceRegistry serviceRegistry = standardServiceRegistryBuilder.build();
+		try {
+			return buildSessionFactory( serviceRegistry );
+		}
+		catch (Throwable t) {
+			serviceRegistry.close();
+			throw t;
```
yrodiere, line 730:
```suggestion
			try {
					serviceRegistry.close();
			}
			catch (RuntimeException e) {
				t.addSuppressed( e );
			}
			throw t;

File: hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java [hunk 2]
```diff
@@ -199,74 +202,81 @@ private EntityManagerFactoryBuilderImpl(
...
+			}
 
-		// for the time being we want to revoke access to the temp ClassLoader if one was passed
-		metamodelBuilder.applyTempClassLoader( null );
+			// for the time being we want to revoke access to the temp ClassLoader if one was passed
+			metamodelBuilder.applyTempClassLoader( null );
+		}
+		catch (Throwable t) {
+			bsr.close();
+			cleanup();
```
yrodiere, line 277: Here too, I think a try/catch + addSuppressed would be better; especially on 5.3 where we don't want backwards-incompatible changes.

yrodiere (commented):
I found those two potential problems, though I guess it's not very critical. I'd be careful about those things on 5.3, though.

About the backported fixes, if I were you I'd track down the original authors/reviewers and ask them directly.

Sanne:
Sorry, I'm not comfortable backporting these changes to 5.3 at this time. We can revisit in future if my assesment of benefits vs risks changes.

PR #4987, opened 2022-04-25
HHH-15228 ("RoundFunction" which extends StandardSQLFunction to return BigDecimal)

mayur-bhindi:
HHH-15228

"RoundFunction" which extends StandardSQLFunction to return BigDecimal

- support for round() function in H2 database to return BigDecimal if argument is BigDecimal
- By default, it would return Double

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
@@ -34,6 +34,7 @@ import org.hibernate.tool.schema.extract.internal.SequenceInformationExtractorH2
 import org.hibernate.tool.schema.extract.internal.SequenceInformationExtractorLegacyImpl;
 import org.hibernate.tool.schema.extract.spi.SequenceInformationExtractor;
 import org.hibernate.type.StandardBasicTypes;
+
 import org.jboss.logging.Logger;
 
 import java.sql.SQLException;
@@ -53,8 +54,8 @@ public class H2Dialect extends Dialect {
 	private static final AbstractLimitHandler LIMIT_HANDLER = new AbstractLimitHandler() {
 		@Override
 		public String processSql(String sql, RowSelection selection) {
-			final boolean hasOffset = LimitHelper.hasFirstRow(selection);
-			return sql + (hasOffset ? " limit ? offset ?" : " limit ?");
+			final boolean hasOffset = LimitHelper.hasFirstRow( selection );
+			return sql + ( hasOffset ? " limit ? offset ?" : " limit ?" );
 		}
 
 		@Override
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
```diff
@@ -85,19 +81,19 @@ public H2Dialect() {
 		SequenceInformationExtractor sequenceInformationExtractor = SequenceInformationExtractorH2DatabaseImpl.INSTANCE;
 		try {
 			// HHH-2300
-			final Class h2ConstantsClass = ReflectHelper.classForName( "org.h2.engine.Constants" );
-			final int majorVersion = (Integer) h2ConstantsClass.getDeclaredField( "VERSION_MAJOR" ).get( null );
-			final int minorVersion = (Integer) h2ConstantsClass.getDeclaredField( "VERSION_MINOR" ).get( null );
-			final int buildId = (Integer) h2ConstantsClass.getDeclaredField( "BUILD_ID" ).get( null );
-			if ( buildId < 32 ) {
```
NathanQingyangXu, line 92: Seems your format change doesn't align with Hibernate's convention, which uses extra spaces to improve code readability. Please revert back the format changes (if you are using Intellij IDE, refer to the "Code Style" section in https://hibernate.org/community/contribute/intellij-idea/) so your meaningful semantic changes stand out.

hibernate-core/src/main/java/org/hibernate/dialect/function/RoundFunction.java
```diff
@@ -0,0 +1,38 @@
...
+	public RoundFunction(String name) {
+		super(name);
+	}
+
+	@Override
+	public Type getReturnType(Type firstArgumentType, Mapping mapping) {
+		return firstArgumentType == StandardBasicTypes.BIG_DECIMAL ? firstArgumentType : StandardBasicTypes.DOUBLE;
+	}
+
+}
```
NathanQingyangXu, line 38: As a general rule, PR without testing case has slim chance to be approved (with the exception of doc PR), let alone merged.
Please consider creating a compelling testing case(s) to prove your code changes.

gavinking:
Not sure I think we should be making piecemeal changes like this to Hibernate 5.

This is all already fixed Hibernate 6 where all this is *much* more well-defined, and if you're stuck on 5 you can pretty do this customization in your own project.

katzyn:
`ROUND()` in modern versions of H2 (2.0.202 and newer) actually returns value of exactly the same data type as data type of its first argument, if it is a `TINYINT`, `SMALLINT`, `INTEGER`, `BIGINT`, `REAL`, `DOUBLE PRECISION` or `DECFLOAT`. If it is a `NUMERIC` (`DECIMAL`), this function returns a `NUMERIC` value with adjusted precision and scale.

In unsupported (on H2 side) H2 1.4.200 it returns value of the same data type for `REAL` and `DOUBLE PRECISION` arguments and `DECIMAL` (`NUMERIC`) for arguments of other data types.

In unsupported (on H2 side) H2 1.4.199 and older versions it returns a `DOUBLE PRECISION` value unconditionally.

Old implementation looks like correct for H2 1.4.199 and older versions and incorrect for newer versions. Implementation proposed here looks like incorrect for all versions of H2, but it may work better anyway.

gavinking:
But the correct (portable) behavior is now defined by JPA, and is not something that is supposed to vary between databases.

Again, I don't think we need to be messing with the existing behavior in H5. This was all already carefully reviewed for H6.

mayur-bhindi:
@gavinking but it behaves differently with postgres. We are using postgres database production environment, it returns the data type same as argument, but H2 doesn't for same version of hibernate.

So currently, there are different behaviours for H2 and postgres (I assume others database as well).

gavinking:
Again: if you're talking about Hibernate 5 here, then yes, correct, there are very many things that are inconsistent between dialects on H5. These have been fixed in Hibernate 6, and I don't think it makes much sense to go back and try to fix them all in H5.

People who want better portability between databases should update to H6.

gavinking:
#9323 shows that this works (it has worked since at least H6 and maybe earlier).

PR #4988, opened 2022-04-27
HHH-15150 Hibernate 6 EmbeddedId CacheKeys are no longer Serializable

dreab8:
https://hibernate.atlassian.net/browse/HHH-15150

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/ComponentType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/ComponentType.java
@@ -21,15 +21,14 @@ import org.hibernate.NotYetImplementedFor6Exception;
 import org.hibernate.PropertyNotFoundException;
 import org.hibernate.boot.spi.MetadataBuildingContext;
 import org.hibernate.bytecode.enhance.spi.LazyPropertyInitializer;
-import org.hibernate.cfg.Environment;
-import org.hibernate.engine.config.spi.ConfigurationService;
+import org.hibernate.cache.internal.CacheKeyValueDescriptor;
+import org.hibernate.cache.internal.ComponentCacheKeyValueDescriptor;
 import org.hibernate.engine.spi.CascadeStyle;
 import org.hibernate.engine.spi.Mapping;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.internal.util.StringHelper;
 import org.hibernate.internal.util.collections.ArrayHelper;
-import org.hibernate.internal.util.config.ConfigurationHelper;
 import org.hibernate.mapping.Component;
 import org.hibernate.mapping.Property;
 import org.hibernate.metamodel.spi.EmbeddableInstantiator;
@@ -64,7 +63,6 @@ public class ComponentType extends AbstractType implements CompositeTypeImplemen
 
 	private final boolean isKey;
 	private boolean hasNotNullProperty;
-	private final boolean createEmptyCompositesEnabled;
 	private final CompositeUserType<Object> compositeUserType;
 
 	private EmbeddableValuedModelPart mappingModelPart;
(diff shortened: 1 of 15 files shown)


hibernate-core/src/main/java/org/hibernate/type/ComponentType.java
```diff
@@ -64,10 +64,9 @@ public class ComponentType extends AbstractType implements CompositeTypeImplemen
 
 	private final boolean isKey;
 	private boolean hasNotNullProperty;
-	private final boolean createEmptyCompositesEnabled;
```
dreab8, line 67: I removed this field because it seems that it is not used anymore

hibernate-core/src/main/java/org/hibernate/type/ComponentType.java
```diff
@@ -64,10 +64,9 @@ public class ComponentType extends AbstractType implements CompositeTypeImplemen
...
 	private final CompositeUserType<Object> compositeUserType;
 
-	private EmbeddableValuedModelPart mappingModelPart;
+	private transient EmbeddableValuedModelPart mappingModelPart;
```
dreab8, line 69: not sure if we want to Serialize also this field, the `ComponentType` is used by the cache key for checking the equality so the involved fields are just `propertyTypes` and `compositeUserType`.

sebersole (commented):
You asked about what to Serialize - I think we need to see how this is used after it is serialized.  Is it really truly just to perform equals/hashCode, which you seemed to indicate?  Is this (caching) the only time a ComponentType can be serialized?  These answers affect the possible solutions.

As possible alternatives, there were 2 options I was thinking of to address this, though I had not yet dug into implementing them:

1. Keep a registry of either `EmbeddableValuedModelPart` or `EmbeddableMappingType` by `NavigableRole`, serialize only the `NavigableRole`  and use that `NavigableRole` to resolve from the registry.
2. Use a memento pattern.

For the memento option, we would generate a memento of some form to use for cache keys.  E.g.

```
interface Type {
    default CacheKeyValueDescriptor toCacheKeyDescriptor() {
        return (CacheKeyValueDescriptor) this;
    }
}

interface CacheKeyValueDescriptor {
    int determineHashCode(Object key);
    boolean areEqual(Object key1, Object key2);
}

class ComponentType ... {
    @Override
    public CacheKeyValueDescriptor toCacheKeyDescriptor() {
        // generate the memento
        return ...;
    }
}
```

The only ramifications AFAICS would be `CacheKeyImplementation` using `CacheKeyValueDescriptor` rather than `Type`.

Or, if we want to be a little less "invasive"[1] we could type the memento itself as a `Type`:


```
interface Type {
    default Type toCacheKeyDescriptor() {
        return this;
    }
}

class ComponentType ... {
    @Override
    public Type toCacheKeyDescriptor() {
        // generate the memento
        return ...;
    }
}
```

[1] Personally I do not find this particularly invasive.  But it would be changes to a few contracts  during a bugfix release; though this is classified as a bug, so...

hibernate-core/src/main/java/org/hibernate/type/ComponentType.java
```diff
@@ -855,4 +845,35 @@ public EmbeddableValuedModelPart mappingModelPart() {
...
+		oos.writeObject( originalPropertyOrder );
+		oos.writeInt( propertySpan );
+		oos.writeObject( cascade );
+		oos.writeObject( joinedFetch );
+		oos.writeBoolean( isKey );
+		oos.writeBoolean( hasNotNullProperty );
+		oos.writeObject( compositeUserType );
+	}
+
+	private void readObject(ObjectInputStream ois) throws IOException, ClassNotFoundException, SQLException {
```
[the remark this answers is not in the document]
dreab8: checked and you are right, it is not enough

hibernate-core/src/test/java/org/hibernate/orm/test/serialization/entity/PK.java
```diff
@@ -30,4 +32,24 @@ protected Long getValue() {
...
+		if ( this == o ) {
+			return true;
+		}
+		if ( o == null || getClass() != o.getClass() ) {
+			return false;
+		}
+
+		PK pk = (PK) o;
+
+		return !( value != null ? !value.equals( pk.value ) : pk.value != null );
```
NathanQingyangXu, line 47: we might well use the new util method `Objects#equals(value, pk.value)` from JDK8 and `Objects#hashCode(value)` below to save some chore and ensure nullness correctness, but it would shine more for multiple-field case.

dreab8:
@sebersole I tried to apply what we discussed, not sure about `Type extends CacheKeyValueDescriptor` but it avoids a toCacheKeyDescriptor in the `toCacheKeyDescriptor`.

sebersole:
Applied upstream with some minor changes

PR #4989, opened 2022-04-27
HHH-15212 SchemaExport.execute does not replace the ${schema}-placeholder in HBM database-object with configured schema

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15212

cc @NathanQingyangXu 

Backport to branch 5.6: #4990

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/relational/SqlStringGenerationContext.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/relational/SqlStringGenerationContext.java
@@ -25,8 +25,24 @@ public interface SqlStringGenerationContext {
 	 * @return The helper for dealing with identifiers in the current JDBC environment.
 	 * <p>
 	 * Note that the Identifiers returned from this helper already account for auto-quoting.
+	 *
+	 * @deprecated Use {@link #toIdentifier(String)} instead.
 	 */
+	@Deprecated
 	IdentifierHelper getIdentifierHelper();
+
+	/**
+	 * Generate an Identifier instance from its simple name as obtained from mapping
+	 * information.
+	 * <p/>
+	 * Note that Identifiers returned from here may be implicitly quoted based on
+	 * 'globally quoted identifiers' or based on reserved words.
+	 *
+	 * @param text The text form of a name as obtained from mapping information.
+	 *
+	 * @return The identifier form of the name.
+	 */
+	Identifier toIdentifier(String text);
 
 	/**
 	 * @return The default catalog, used for table/sequence names that do not explicitly mention a catalog.
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/boot/database/qualfiedTableNaming/DefaultCatalogAndSchemaTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/boot/database/qualfiedTableNaming/DefaultCatalogAndSchemaTest.java
@@ -80,7 +80,7 @@ import jakarta.persistence.TableGenerator;
 import static org.assertj.core.api.Assertions.assertThat;
 
 @RunWith(CustomParameterized.class)
-@TestForIssue(jiraKey = { "HHH-14921", "HHH-14922" })
+@TestForIssue(jiraKey = { "HHH-14921", "HHH-14922", "HHH-15212" })
 public class DefaultCatalogAndSchemaTest {
 
 	private static final String SQL_QUOTE_CHARACTER_CLASS = "([`\"]|\\[|\\])";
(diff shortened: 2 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/relational/SqlStringGenerationContext.java
```diff
@@ -25,9 +25,25 @@ public interface SqlStringGenerationContext {
...
+	 * <p/>
+	 * Note that Identifiers returned from here may be implicitly quoted based on
+	 * 'globally quoted identifiers' or based on reserved words.
+	 *
+	 * @param text The text form of a name as obtained from mapping information.
+	 *
+	 * @return The identifier form of the name.
+	 */
+	Identifier toIdentifier(String text);
+
```
beikov, line 46: How about this instead? Also, is it really necessary to deprecate the `getIdentifierHelper` method?

```suggestion
	 */
	IdentifierHelper getIdentifierHelper();

	/**
	 * Generate an Identifier instance from its simple name as obtained from mapping
	 * information.
	 * <p/>
	 * Note that Identifiers returned from here may be implicitly quoted based on
	 * 'globally quoted identifiers' or based on reserved words.
	 *
	 * @param text The text form of a name as obtained from mapping information.
	 *
	 * @return The identifier form of the name.
	 */
	default Identifier toIdentifier(String text) {
		return getIdentifierHelper() != null ? getIdentifierHelper().toIdentifier( text ) : Identifier.toIdentifier( text );
	}

```

beikov (approved):
LGTM. Just for confirmation, the issue is that `getIdentifierHelper()` returns `null` which is why it replaces the placeholder with an empty string, right?

hibernate-core/src/main/java/org/hibernate/boot/model/relational/SqlStringGenerationContext.java
yrodiere: > Just for confirmation, the issue is that getIdentifierHelper() returns null which is why it replaces the placeholder with an empty string, right?

No, actually `getIdentifierHelper()` wasn't used at all before this PR. I believe it was introduced by mistake, and recently at that (in my PR that delayed the resolution of default catalog/schema until session factory creation).

The actual fix is the fact that we call `context.catalogWithDefault`/`context.schemaWithDefault` in `injectCatalogAndSchema`. It just turns out that we need to turn a `String` into an `Identifier` in order to do that, so I had to add such a feature to `SqlStringGenerationContext`, and I noticed problems with `getIdentifierHelper`.

> How about this instead? Also, is it really necessary to deprecate the getIdentifierHelper method?

Necessary, no. But as stated above, I believe this method was introduced by mistake. There are no identified use cases for that method, and it can return null in some cases, which is not documented. I'd rather not expose things that we don't

hibernate-core/src/test/java/org/hibernate/orm/test/boot/database/qualfiedTableNaming/DefaultCatalogAndSchemaTest.java
```diff
@@ -512,6 +514,15 @@ private void verifyDDLQualifiers(String sql) {
...
 		verifyOnlyQualifier( sql, SqlType.DDL, EntityWithExplicitQualifiersWithEnhancedSequenceGenerator.NAME, expectedExplicitQualifier() );
+
+		if ( dbSupportsCatalogs && expectedDefaultCatalog != null ) {
+			verifyOnlyQualifier( sql, SqlType.DDL, "catalogPrefixedAuxObject",
+					expectedQualifier( expectedDefaultCatalog, null ) );
+		}
+		if ( dbSupportsSchemas && expectedDefaultSchema != null ) {
+			verifyOnlyQualifier( sql, SqlType.DDL, "schemaPrefixedAuxObject",
+					expectedQualifier( null, expectedDefaultSchema ) );
+		}
```
NathanQingyangXu, line 525: it would be great to add this JIRA ticket to the `@TestForIssue` annotation as well

hibernate-core/src/test/resources/org/hibernate/orm/test/boot/database/qualfiedTableNaming/database-object-using-catalog-placeholder.hbm.xml
```diff
@@ -0,0 +1,26 @@
...
+	<database-object>
+		<create>
+			CREATE OR REPLACE FUNCTION ${catalog}.catalogPrefixedAuxObject()
+			RETURNS varchar AS
+			$BODY$
+			BEGIN
+			SELECT 'test';
+			END;
+			$BODY$
+			LANGUAGE plpgsql
```
NathanQingyangXu, line 22: I assume this DDL is vendor specific, so maybe we need to specify dialect (Postgres) in testing code?

yrodiere: I just need to put something here. This DDL never gets executed. I'll change the code to make this clearer.

I want to test `${schema}` replacement with every dialect, so moving this to a postgres-specific test is not the right solution.

yrodiere:
Thanks for the reviews; I think I addressed the comments. Will merge as soon as CI goes green.

NathanQingyangXu:
> Thanks for the reviews; I think I addressed the comments. Will merge as soon as CI goes green.

I hate to be so picky, but there is a typo introduced previously in the package name of org/hibernate/orm/test/boot/database/qualfiedTableNaming. Should be .../qualifiedTableNaming. Not a big deal for it is only meant for testing.

NathanQingyangXu (approved):

PR #4991, opened 2022-04-28
Add missing blank in Dialect.cast SQL expression

urld:
The missing blank before 'as' resultet in an invalid SQL expression.


dreab8:
Hi @urld ,

can you please create a Jira ticket and if possible add to the PR a test case (it will be useful to prevent regressions).

Thanks

beikov:
I understand the desire to fix this, but I just wanted to note that the only usages of this method are passing in `Types.CHAR` so the faulty branch will never be executed in regular Hibernate code. Only user code calling this method would run into the issue.

beikov:
Closing the PR now because it was stale for a while and the bug doesn't affect ORM code. You can create a new PR for a JIRA issue, after you created that, along with a test case.

PR #4995, opened 2022-04-29
HHH-15233 Query NPE when using ":param IS NULL" and the parameter is an Entity 

dreab8:
https://hibernate.atlassian.net/browse/HHH-15233


sebersole (commented):
:+1:

PR #4996, opened 2022-04-29
HHH 15201 Composite ID with dynamic-map entities fails with an exception 

dreab8:
https://hibernate.atlassian.net/browse/HHH-15201


sebersole (commented):
I made a small change to this, and I'll push the combined changes upstream

PR #4998, opened 2022-04-29
HHH-15238 - Documentation: update JDBC logging configuration for 6.0

hanohrs:
https://hibernate.atlassian.net/browse/HHH-15238

Correct an appender setting and separate them into the bind and the extract settings.


beikov:
Thanks!

PR #4370, opened 2021-11-16
HHH-14927 bump current hibernate version from 5.5 to 5.6 in user guide

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-14927

Fixed some obvious defects I've found in user guide along the way. See self-comments for details.

PR #4999, opened 2022-04-30
HHH-15239 add 5.6 link in user guide after v6 became the new "current"

NathanQingyangXu:
So I added the missing "previous current" or v5.5 link when v5.6 became the new "current" (see https://github.com/hibernate/hibernate-orm/pull/4370); now we need to do similar thing when v6 became the new "current".

It would be great to keep that in mind when new version got released: the previous "current" needs to be added explicitly in documentation/src/main/style/js/toc.js

Or could we make it automatic so there is zero delay before the "previous version" doc was added?


beikov:
Thanks

PR #5000, opened 2022-04-30
HHH-15241 fix broken references in javadoc

NathanQingyangXu:
It is common to focus on code refactoring without updating Javadoc and the broken link reference is the most obvious implication from user side. This PR aims to eliminate all the broken link references.


hibernate-core/src/main/java/org/hibernate/boot/model/source/spi/PluralAttributeMapKeySource.java
```diff
@@ -39,7 +39,7 @@ enum Nature {
 	 * (relevant only for one-to-many and many-to-many associations)?
 	 *
 	 * If this method returns {@code true}, then this object can safely
-	 * be cast to {@link PluralAttributeMapKeySourceEntityAttribute}.
+	 * be cast to {@link PluralAttributeMapKeyManyToAnySource}.
 	 *
 	 * @return true, if this plural attribute index source for an attribute of the referenced
 	 * entity; false, otherwise.
```
NathanQingyangXu, line 45: The above change is the most difficult one and I am not sure it is correct. Let me know if the change is erroneous.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/JpaCriteriaParameter.java
```diff
@@ -21,7 +21,7 @@
 /**
  * {@link JpaParameterExpression} created via JPA {@link jakarta.persistence.criteria.CriteriaBuilder}.
  *
- * Each occurence of a JpaParameterExpression results in a unique SqmParameter
+ * Each occurrence of a JpaParameterExpression results in a unique SqmParameter
```
NathanQingyangXu, line 24: still found a typo my last PR overlooked. Correct it along the way.

beikov:
Please fix the checkstyle issue

hibernate-core/src/main/java/org/hibernate/annotations/BatchSize.java
```diff
@@ -19,19 +19,19 @@
...
- * <pre>{@code
- * @OneToMany
- * @BatchSize(size = 5) /
- * Set<Product> getProducts() { ... };
- * }</pre>
+ * <pre>
+ *    &#64;OneToMany
+ *    &#64;BatchSize(size = 5) /
+ *    Set&lt;Product&gt; getProducts() { ... };
+ * </pre>
```
NathanQingyangXu, line 34: we need to escape HTML special chars (including '<', '>' and '&'). See the current erroneous rending effect at https://docs.jboss.org/hibernate/orm/6.0/javadocs/org/hibernate/annotations/BatchSize.html

NathanQingyangXu:
> Please fix the checkstyle issue

Fixed. Seems the checkstyle checking is not dialect specific? Then it makes more sense to refactor it into its own step rather than duplicating it in each dialect checking. Anyway.

beikov:
Thanks

PR #5001, opened 2022-05-01
HHH-15232 exception thrown when empty entity collection is provided for a query param

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-15232


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -6433,6 +6433,7 @@ private Predicate processInSingleCriteriaParameter(
 		return processInSingleParameter( sqmPredicate, sqmWrapper, jpaCriteriaParameter, domainParamBinding );
 	}
 
+	@SuppressWarnings( "rawtypes" )
```
NathanQingyangXu, line 6436: guys, we have so many Java Generics warnings, :(.

beikov:
Thanks!

PR #5003, opened 2022-05-02
HHH-15240 Defining a UserVersionType causes AssertionError from VersionResolution if running in JVM with -ea flag set

dreab8:
https://hibernate.atlassian.net/browse/HHH-15240

The issue is that `legacyType` is an instance of `CustomType` and its `JdbcType` is an instance of `UserTypeSqlTypeAdapter` while `recommendedJdbcType` is an instcance of `BigIntTypeDescriptor` so the assertion falis.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/process/internal/VersionResolution.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/process/internal/VersionResolution.java
@@ -83,7 +83,7 @@ public class VersionResolution<E> implements BasicValue.Resolution<E> {
 		final BasicType<?> basicType = typeConfiguration.getBasicTypeRegistry().resolve( jtd, recommendedJdbcType );
 		final BasicType legacyType = typeConfiguration.getBasicTypeRegistry().getRegisteredType( jtd.getJavaType() );
 
-		assert legacyType.getJdbcType().equals( recommendedJdbcType );
+		assert legacyType.getJdbcType().getJdbcTypeCode() == recommendedJdbcType.getJdbcTypeCode();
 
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/process/internal/VersionResolution.java
```diff
@@ -22,6 +22,8 @@
 import org.hibernate.type.descriptor.java.MutabilityPlan;
 import org.hibernate.type.descriptor.jdbc.JdbcType;
 import org.hibernate.type.descriptor.jdbc.JdbcTypeIndicators;
+import org.hibernate.type.internal.UserTypeJavaTypeWrapper;
+import org.hibernate.type.internal.UserTypeVersionJavaTypeWrapper;
```
NathanQingyangXu, line 26: Are the above two imports really used in this class? Seems not from your code changes below.

hibernate-core/src/test/java/org/hibernate/orm/test/version/UserVersionTest.java
```diff
@@ -0,0 +1,201 @@
...
+
+		@Override
+		public CustomVersion replace(CustomVersion original, CustomVersion target, Object owner)
+				throws HibernateException {
+			return original;
+		}
+
+		@Override
+		public CustomVersion seed(SharedSessionContractImplementor session) {
+			return new CustomVersion( ++versionValue );
```
NathanQingyangXu, line 188: seems the `versionValue` is only used in this entity class, so relocating the variable declaration within this class makes more sense. not a big deal though.

beikov (approved):
LGTM

NathanQingyangXu (approved):

PR #5004, opened 2022-05-02
Skip signing of artifacts when the key file is missing

beikov:


gradle/published-java-module.gradle
```diff
@@ -99,9 +99,21 @@ publishing {
 signing {
 	var signingKey = resolveSigningKey()
 	var signingPassword = findSigningProperty( "signingPassword" )
-	useInMemoryPgpKeys( signingKey, signingPassword )
 
-	sign publishing.publications.publishedArtifacts
+	gradle.taskGraph.whenReady {tg ->
```
sebersole, line 103: Technically does not matter, but this should be outside the `signing {}` block.

gradle/published-java-module.gradle
```diff
@@ -99,9 +99,21 @@ publishing {
...
+		if (tg.hasTask(project.tasks.releaseChecks)) {
```
sebersole, line 104: What if I try to execute signing by itself?  This is not the proper task to check for - they only happen when doing a release.  If we re-configure signing, it would be nice to test signing without doing a release right?

beikov:
Fixed via c43fa6b4d9be8235f8d34e450fae7cedb87acd25

PR #5005, opened 2022-05-02
HHH-15154 tweak verbiage in user guide on FetchMode.SUBSELECT

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-15154


beikov:
Thanks

PR #5006, opened 2022-05-02
HHH-15149 add a new 'SelectionQuery#uniqueList()' method

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-15149

straightforward implementation. The subtleties include:
* uniqueList() should be based directly on top of list() result, as opposed to some common internal method shared by both (so overriden list() would take care of uniqueList() automatically)
* null value
* retain the first element in face of duplication as per the original order in list() result
* avoid the common trap to use `new HashSet<>(expectedSize)` (as in both ANTR and byteBuddy libraries, but we have corresponding CollectionHelper method already).

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/spi/AbstractSelectionQuery.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/spi/AbstractSelectionQuery.java
@@ -8,6 +8,7 @@ package org.hibernate.query.spi;
 
 import java.sql.Types;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Collection;
 import java.util.Date;
@@ -15,6 +16,7 @@ import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.stream.Stream;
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/query/spi/AbstractSelectionQuery.java
```diff
@@ -378,6 +381,22 @@ public List<R> list() {
...
+			return list;
+		}
+		Set<R> dict = CollectionHelper.linkedSetOfSize( list.size() );
+		List<R> uniqueList = new ArrayList<>( list.size() );
+		for ( R listElement : list ) {
+			if ( dict.add( listElement ) ) {
+				uniqueList.add( listElement );
+			}
+		}
+		return uniqueList;
```
beikov, line 397: I was thinking about using a new (to be added) method `AbstractSelectionQuery#doList(ListResultsConsumer.UniqueSemantic)` that delegates to `SelectQueryPlan#performList(DomainQueryExecutionContext, ListResultsConsumer.UniqueSemantic)` (additional method or alter the existing one since the contract is incubating) so we can use our existing `ListResultsConsumer` code for handling this. Wdyt @sebersole and @dreab8?

NathanQingyangXu:
I didn't assign the ticket to me for I am not sure of my implementation is
the best. Feel free to take it and I can close mine if needed.

hibernate-core/src/main/java/org/hibernate/query/spi/AbstractSelectionQuery.java
NathanQingyangXu: I think I misunderstood the requirements. The gist of the ticket is to use the `select distinct xxx' from db server side rather than such superficial post processing. For that reason, I closed this PR. Thanks for the feedback.

PR #5008, opened 2022-05-03
HHH-15200 Add support for the SQLXML type

beikov:
https://hibernate.atlassian.net/browse/HHH-15200

### DIFF
--- a/documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
+++ b/documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
@@ -430,8 +430,22 @@ Can reference a
 `FormatMapper` instance,
 `FormatMapper` implementation `Class` reference,
 `FormatMapper` implementation class name (fully-qualified class name) or
-one of the following short hand constants `jackson` or `jsonb`.
-By default the first of the possible providers that is available in the runtime is used, according to the listing order.
+one of the following shorthand constants `jackson` or `jsonb`.
+By default, the first of the possible providers that is available in the runtime is used, according to the listing order.
++
+Note that the default serialization format of collections can differ depending on the serialization library.
+
+`*hibernate.type.xml_format_mapper*` (e.g. A fully-qualified class name, an instance, or a `Class` object reference)::
+Names a https://docs.jboss.org/hibernate/orm/{majorMinorVersion}/javadocs/org/hibernate/type/FormatMapper.html[`FormatMapper`] implementation to be applied to the `SessionFactory` for XML serialization and deserialization.
++
+Can reference a
+`FormatMapper` instance,
+`FormatMapper` implementation `Class` reference,
+`FormatMapper` implementation class name (fully-qualified class name) or
+one of the following shorthand constants `jackson-xml` or `jaxb`.
+By default, the first of the possible providers that is available in the runtime is used, according to the listing order.
++
+Note that the default serialization format of collections can differ depending on the serialization library.
 
 [[configurations-bytecode-enhancement]]
 === Bytecode Enhancement Properties
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
@@ -2562,14 +2562,30 @@ public interface AvailableSettings {
 	 *     <li>an instance of {@code FormatMapper},
 	 *     <li>a {@link Class} representing a class that implements {@code FormatMapper},
 	 *     <li>the name of a class that implements {@code FormatMapper}, or
-	 *     <li>one of the short hand constants {@code jackson} or {@code jsonb}.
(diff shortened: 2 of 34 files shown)


documentation/src/main/asciidoc/userguide/appendices/Configurations.adoc
```diff
@@ -430,8 +430,22 @@ Can reference a
 `FormatMapper` instance,
 `FormatMapper` implementation `Class` reference,
 `FormatMapper` implementation class name (fully-qualified class name) or
-one of the following short hand constants `jackson` or `jsonb`.
-By default the first of the possible providers that is available in the runtime is used, according to the listing order.
+one of the following shorthand constants `jackson` or `jsonb`.
+By default, the first of the possible providers that is available in the runtime is used, according to the listing order.
```
NathanQingyangXu, line 434: This is not related directly to the scope. "in the runtime" is better phrased as "at the runtime"?

hibernate-core/src/main/java/org/hibernate/type/JaxbXmlFormatMapper.java
```diff
@@ -0,0 +1,258 @@
...
+
+/**
+ * @author Christian Beikov
+ */
+public class JaxbXmlFormatMapper implements FormatMapper {
+
+	public static final String SHORT_NAME = "jaxb";
+	public static final JaxbXmlFormatMapper INSTANCE = new JaxbXmlFormatMapper();
+
+	public JaxbXmlFormatMapper() {
```
NathanQingyangXu, line 43: Did you intend private constructor to enforce singleton pattern?

hibernate-core/src/main/java/org/hibernate/type/JaxbXmlFormatMapper.java
```diff
@@ -0,0 +1,258 @@
...
+					context = JAXBContext.newInstance( MapWrapper.class, keyClass, valueClass );
+				}
+				else {
+					keyClass = Object.class;
+					valueClass = Object.class;
+					context = JAXBContext.newInstance( MapWrapper.class );
+				}
+				final Unmarshaller unmarshaller = context.createUnmarshaller();
+				final MapWrapper mapWrapper = (MapWrapper) unmarshaller
+						.unmarshal( new StringReader( charSequence.toString() ) );
```
NathanQingyangXu, line 66: Could we use `CharArrayReader` directly without String transforming overhead?

hibernate-core/src/main/java/org/hibernate/type/JaxbXmlFormatMapper.java
```diff
@@ -0,0 +1,258 @@
...
+				final List<Object> elements = mapWrapper.elements;
+				final Map<Object, Object> map = new LinkedHashMap<>( elements.size() >> 1 );
```
NathanQingyangXu, line 68: This is not correct. Similar to `HashMap` and `HashSet`, we should not use `new LinkedHashMap<>(expectedSize)` for it would lead to rehashing when the size surpasses `expectedSize * 0.75`, thus defeating the purpose.
Our CollectionHelper class has a util method already: `CollectionHelper#linkedMapOfSize(expectedSize)`.
Such expected size misuse is so common. Not sure why JDK introduces some new method to avoid such confusion.

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -2571,6 +2571,22 @@ public interface AvailableSettings {
...
 	String JSON_FORMAT_MAPPER = "hibernate.type.json_format_mapper";
 
+	/**
+	 * Specifies a {@link org.hibernate.type.FormatMapper} used for XML serialization
+	 * and deserialization, either:
+	 * <ul>
+	 *     <li>an instance of {@code FormatMapper},
+	 *     <li>a {@link Class} representing a class that implements {@code FormatMapper},
+	 *     <li>the name of a class that implements {@code FormatMapper}, or
+	 *     <li>one of the short hand constants {@code jackson} or {@code jaxb}.
```
NathanQingyangXu, line 2581: short hand -> shorthand

hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
```diff
@@ -2562,15 +2562,31 @@ public interface AvailableSettings {
 	 *     <li>an instance of {@code FormatMapper},
 	 *     <li>a {@link Class} representing a class that implements {@code FormatMapper},
 	 *     <li>the name of a class that implements {@code FormatMapper}, or
-	 *     <li>one of the short hand constants {@code jackson} or {@code jsonb}.
+	 *     <li>one of the shorthand constants {@code jackson} or {@code jsonb}.
 	 * </ul>
-	 * By default, the first of the possible providers that is available in the runtime is
+	 * By default, the first of the possible providers that is available at the runtime is
```
jrenaat, line 2567: at runtime (without the 'the') ?

PR #5013, opened 2022-05-04
HHH-4384 Allow join column override if @JoinColumn is absent on @OneToOne(mappedBy = "")

dreab8:
https://hibernate.atlassian.net/browse/HHH-4384


beikov (approved):

hibernate-core/src/test/java/org/hibernate/orm/test/annotations/onetoone/OverrideOneToOneJoinColumnTest.java
```diff
@@ -0,0 +1,206 @@
...
+			fail( "Should disallow @JoinColumn override on side with mappedBy" );
+		}
+		catch (AnnotationException ex) {
+			assertTrue(
+					"Should disallow exactly because of @JoinColumn override on side with mappedBy",
+					ex
+							.getMessage()
+							.startsWith( "Illegal attempt to define a @JoinColumn with a mappedBy association:" )
+			);
+		}
```
NathanQingyangXu, line 80: The above boilerplate could be simplified by JUnit 5's `assertThrows()` method as following:
```
final AnnotationException ex = Assertions.assertThrows(AnnotationException.class, () -> new MetadataSources( ssr )
					.addAnnotatedClass( Employee.class )
					.addAnnotatedClass( PartTimeEmployee.class )
					.addAnnotatedClass( Desk.class )
					.buildMetadata());
assertTrue(
					"Should disallow exactly because of @JoinColumn override on side with mappedBy",
					ex.getMessage().startsWith( "Illegal attempt to define a @JoinColumn with a mappedBy association:" )
			);
```
The benefits include:
* more compact and elegant
* avoid the common error to forget the `fail()` statement (happy path would slip through)

One good example of lambda usage shines in JUnit 5.
Need not to be changed; just for communication purpose. In my current big bank contract, so many testing cases lack the 'fail()' statement ending up with failure to catch the happy paths.

NathanQingyangXu:
Congratulations for fixing such **antient** ticket finally! It must be a good feeling, :).

hibernate-core/src/test/java/org/hibernate/orm/test/annotations/onetoone/OverrideOneToOneJoinColumnTest.java
dreab8: Thanks @NathanQingyangXu , applied your suggestion!

dreab8:
> Congratulations for fixing such **antient** ticket finally! It must be a good feeling, :).

thanks to @ettavolt for the fix!

PR #5017, opened 2022-05-06
HHH-15159 Orphan removal of an entity with an @ElementCollection causes a ConstraintViolationException

dreab8:
https://hibernate.atlassian.net/browse/HHH-15159

Not sure if executing `CollectionRemoveAction` first can cause any issue.


beikov (approved):
LGTM. AFAICT we can always do removals of collection items as a pre-step, because no FK could ever point to these elements, so this is a good fix!

sebersole (commented):
Very reasonable

PR #5018, opened 2022-05-06
HHH-15253 Enable JMH performance benchmark

NathanQingyangXu:
A performance benchmark capability is needed to verify various performance improvement, and JMH is an obvious choice in this regard. Without such tool, we might fall prey of wishful thinking, tunnel vision and missing the performance bottleneck.

Might provide some interesting use cases to showcase the usage of such tool, maybe on some of my previous PRs.


sebersole:
We develop JMH benchmarks at https://github.com/sebersole/hibernate-hql-benchmark

PR #5026, opened 2022-05-07
HHH-15255 fix org.hibernate.orm.test.batch.BatchTest

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-15255


hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchingBatch.java
```diff
@@ -30,8 +30,6 @@ public class BatchingBatch extends AbstractBatchImpl {
 			BatchingBatch.class.getName()
 	);
 
-	// IMPL NOTE : Until HHH-5797 is fixed, there will only be 1 statement in a batch
```
NathanQingyangXu, line 33: I delete the above outdated comment along the way; [HHH-](https://hibernate.atlassian.net/browse/HHH-5797) has been fixed since 5.2.15.

beikov:
Thanks!

PR #5027, opened 2022-05-08
Remove redundant check for `columns.size() > 0` in JPAXMLOverriddenAnnotationReader

missingdays:
There's always at least one element in `columns` because of `columns.add( getColumn( element, false, nodeName ) );`


Sanne:
Good catch! Unfortunately I see that after resolving conflicts, this patch is no longer neccesary, so I'll close - but thanks anyway :)

PR #5029, opened 2022-05-09
HHH-15256 HQL Query with left join throws NPE when using :param IS NULL

dreab8:
https://hibernate.atlassian.net/browse/HHH-15256


beikov (approved):
LGTM

PR #5030, opened 2022-05-09
Copy only the content of the documentation staging directory

dreab8:
By adding "/"  at the end of the stagingDirPath only the content of the stagingDirPath is copied,
so on the server it will be created the following dir structure :

```
hibernate /
	orm/
		${releaseFamily}/
			incubating/
			integrationguide/
			...
			userguide/
```

without the dir structure would be :

```
hibernate/
	orm/
		${releaseFamily}/
			documentation/
				incubating/
				integrationguide/
				...
				userguide/
```

see 
https://docs.jboss.org/hibernate/orm/6.0/
or 
https://docs.jboss.org/hibernate/orm/5.5/
created by the previous releases
and the wrong
https://docs.jboss.org/hibernate/orm/6.0/documentation/
created during the last release

### DIFF
--- a/local-build-plugins/src/main/java/org/hibernate/orm/docs/PublishTask.java
+++ b/local-build-plugins/src/main/java/org/hibernate/orm/docs/PublishTask.java
@@ -54,11 +54,12 @@ public abstract class PublishTask extends DefaultTask {
 		final String url = normalizedBase + releaseFamily;
 
 		final String stagingDirPath = stagingDirectory.get().getAsFile().getAbsolutePath();
+		final String stagingDirPathContent =  stagingDirPath.endsWith( "/" ) ? stagingDirPath : stagingDirPath + "/";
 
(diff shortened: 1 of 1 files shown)


beikov (approved):
LGTM

local-build-plugins/src/main/java/org/hibernate/orm/docs/PublishTask.java
```diff
@@ -54,11 +54,12 @@ public void uploadDocumentation() {
 		final String url = normalizedBase + releaseFamily;
 
 		final String stagingDirPath = stagingDirectory.get().getAsFile().getAbsolutePath();
+		final String stagingDirPathContent =  stagingDirPath.endsWith( "/" ) ? stagingDirPath : stagingDirPath + "/";
```
NathanQingyangXu, line 57: It is likely that `stagingDirPath` never ends '/' for internally it will go through a normalizing processing, though it is not a big deal and doing the redundant checking doesn't hurt.

PR #5033, opened 2022-05-09
HHH-15258 Test-case for orphan removal issues in 6.x

theigl:
See https://hibernate.atlassian.net/browse/HHH-15258

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/orphan/onetomany/DeleteOneToManyOrphansTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/orphan/onetomany/DeleteOneToManyOrphansTest.java
@@ -17,6 +17,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 /**
  * @author Gail Badnmer
@@ -151,6 +152,19 @@ public class DeleteOneToManyOrphansTest {
 					Product productQueried = (Product) results.get( 0 );
 					assertEquals( 1, productQueried.getFeatures().size() );
 					assertEquals( featureQueried, productQueried.getFeatures().get( 0 ) );
+				}
+		);
+	}
+
+	@Test
+	@TestForIssue(jiraKey = "HHH-15258")
+	public void testPersistAndQuery(SessionFactoryScope scope) {
+		scope.inTransaction(
+				session -> {
+					Product product = new Product();
+					session.persist( product );
+					assertNotNull( session.createQuery( "from Product" ).list() );
+					assertNotNull( session.createQuery( "from Product" ).list() );
 				}
 		);
 	}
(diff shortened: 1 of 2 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/orphan/onetomany/DeleteOneToManyOrphansTest.java
```diff
@@ -155,4 +156,17 @@ public void testReplacedWhileManaged(SessionFactoryScope scope) {
...
 
+	@Test
+	@TestForIssue(jiraKey = "HHH-15258")
+	public void testPersistAndQuery(SessionFactoryScope scope) {
+		scope.inTransaction(
+				session -> {
+					Product product = new Product();
+					session.persist( product );
+					assertNotNull( session.createQuery( "from Product" ).list() );
+					assertNotNull( session.createQuery( "from Product" ).list() );
```
NathanQingyangXu, line 167: If I didn't misunderstand, your concern is the second `list()` statement would throw exception. It would be great to convey your assertion logic more cleanly, e.g.:
```
session.persist( product );
session.createQuery( "from Product" ).list();
assertDoesNotThrow(session.createQuery( "from Product" ).list(), "without fixing, an exception would be thrown for the second list() invocation");
```
or simply:
```
session.persist( product );
session.createQuery( "from Product" ).list();
// without fixing, the following statement would throw exception
session.createQuery( "from Product" ).list();
```
It would save unnecessary effort for future code maintainer to figure out the irrelevant `assertNotNull()` logic. IMHO.

hibernate-core/src/test/java/org/hibernate/orm/test/orphan/onetomany/Product.java
```diff
@@ -19,7 +20,7 @@
 @Table(name="product")
 public class Product {
 	@Id
-	@GeneratedValue(strategy= GenerationType.AUTO)
+	@GeneratedValue(strategy= GenerationType.IDENTITY)
```
sebersole, line 23: Its generally frowned-upon to change the mapping of entities used in existing tests to match the needs of your new test(s).  If you need a specific mapping, create a model with that mapping

hibernate-core/src/test/java/org/hibernate/orm/test/orphan/onetomany/DeleteOneToManyOrphansTest.java
```diff
@@ -155,4 +157,20 @@ public void testReplacedWhileManaged(SessionFactoryScope scope) {
 		);
 	}
 
+	@Test
+	@TestForIssue(jiraKey = "HHH-15258")
```
sebersole, line 161: Since IDENTITY is required to trigger this (according to Jira), you'll also need to skip this test for databases which do not support IDENTITY. 

```suggestion
	@TestForIssue(jiraKey = "HHH-15258")
	@RequiresDialectFeature( DialectFeatureChecks.SupportsIdentityColumns.class )
```

sebersole (commented):
Thanks for the request!

Left a few comments.

NathanQingyangXu:
I created a PR here: https://github.com/hibernate/hibernate-orm/pull/5035.

dreab8:
Hi @theigl ,

thanks for the test, I'm closing this PR in favour of https://github.com/hibernate/hibernate-orm/pull/5039

PR #5035, opened 2022-05-10
HHH-15258 Orphan removal for OneToMany relations is broken when used with GenerationType.IDENTITY

NathanQingyangXu:

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
@@ -281,6 +281,18 @@ public class ActionQueue {
 			executeInserts();
 			LOG.debug( "Executing identity-insert immediately" );
 			execute( insert );
+
+			addAction(	// HHH-15258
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -281,6 +281,18 @@ private void addResolvedEntityInsertAction(AbstractEntityInsertAction insert) {
...
+					OrphanRemovalAction.class,
+					new OrphanRemovalAction( insert.getId(),
+											 insert.getState(),
+											 null,
+											 insert.getInstance(),
+											 insert.getPersister(),
+											 false,
+											 (SessionImplementor) insert.getSession()
+					)
+			);
```
NathanQingyangXu, line 295: The above bizarre action addition is only to ensure the query spaces are added, so the insertion event can be flushed.
Not elegant solution, but it seems ActionQueue related stuff is not elegant per se and other hack related to OrphanRemoval exists in the class as well.

PR #5037, opened 2022-05-10
HHH-15258 -Orphan removal for OneToMany relations is broken when used with GenerationType.IDENTITY

NathanQingyangXu:

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
@@ -101,6 +101,8 @@ public class ActionQueue {
 	private AfterTransactionCompletionProcessQueue afterTransactionProcesses;
 	private BeforeTransactionCompletionProcessQueue beforeTransactionProcesses;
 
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java
```diff
@@ -257,6 +260,7 @@ private void addInsertAction(AbstractEntityInsertAction insert) {
 			// TODO: find out why this is necessary
 			LOG.tracev( "Executing inserts before finding non-nullable transient entities for early insert: [{0}]", insert );
 			executeInserts();
+			earlyInsertQuerySpaces = insert.getPersister().getQuerySpaces();
```
NathanQingyangXu, line 263: The root cause seems to be `EntityIdentityInsertAction` doesn't leave trace of the table to be updated. I made some exception for it to solve the issue.

dreab8:
the root cause seems to be that a new collection instance is created during the load, so in the PC we have 2 collections intead of one, the first added during the persist and a different one added during the load, this cause the failure during the flush. 

I'm working on a PR

PR #5039, opened 2022-05-10
HHH-15258 Orphan removal for OneToMany relations is broken when used with GenerationType.IDENTITY

dreab8:

The issue is fixed by this https://github.com/hibernate/hibernate-orm/compare/main...dreab8:HHH-15258?expand=1#diff-68e234b1b90c3573e2efd14e73cb756666049314c0f886884a8f0b88dee704f8R118

The `persistentCollection` is added by the persist of the `Parent` that happpens in the same Session of the query.

then I just refactored the common code of the `DelayedCollectionInitializer#resolveInstance` and `SelectEagerCollectionInitializer#resolveInstance` methods into the `AbstractCollectionInitializer`.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/action/internal/AbstractEntityInsertAction.java
+++ b/hibernate-core/src/main/java/org/hibernate/action/internal/AbstractEntityInsertAction.java
@@ -7,16 +7,22 @@
 package org.hibernate.action.internal;
 
 import org.hibernate.LockMode;
+import org.hibernate.collection.spi.PersistentCollection;
 import org.hibernate.engine.internal.ForeignKeys;
 import org.hibernate.engine.internal.NonNullableTransientDependencies;
 import org.hibernate.engine.internal.Nullability;
 import org.hibernate.engine.internal.Versioning;
 import org.hibernate.engine.spi.CachedNaturalIdValueSource;
+import org.hibernate.engine.spi.CollectionKey;
 import org.hibernate.engine.spi.EntityEntry;
 import org.hibernate.engine.spi.EntityKey;
+import org.hibernate.engine.spi.PersistenceContext;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.engine.spi.Status;
 import org.hibernate.metamodel.mapping.NaturalIdMapping;
+import org.hibernate.metamodel.mapping.PluralAttributeMapping;
+import org.hibernate.persister.collection.CollectionPersister;
+import org.hibernate.persister.entity.AbstractEntityPersister;
 import org.hibernate.persister.entity.EntityPersister;
 
 /**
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/AbstractCollectionInitializer.java
```diff
@@ -85,6 +91,69 @@ public void resolveKey(RowProcessingState rowProcessingState) {
...
+			final Object key = collectionKey.getKey();
+
+			persistenceContext.forEachCollectionEntry(
+					(persistentCollection, collectionEntry) -> {
+						if ( collectionEntry.getCurrentPersister() == collectionDescriptor &&
+								key.equals( collectionEntry.getCurrentKey() ) ) {
+							collectionInstance = persistentCollection;
+						}
+					}, true
+			);
```
beikov, line 125: I'm not sure I understand how this code can be relevant. AFAIU, this code should be doing the same as `persistenceContext.getCollection( collectionKey );` on line 107, but if there were an existing collection, we wouldn't even reach this code I marked. Can you explain, maybe also as Java comment, how this special condition can occur?

dreab8: yeap

sebersole: It's delayed insertion of the collection declarer, right?  Assuming so, we should look to circumvent this when the owner is not using IDENTITY generation and/or delayed insertion is by-passed

sebersole: > while for IDENTITY the autoflush is not executed before the query.

Right.  That's the delayed insertion.


> Basically we have to be sure the insert has not yet happened, but not sure how to check this

IIRC the generated-id has some well-defined form and you should be able to tell from the value itself.  Been a long time since I wrote all that though and maybe I misremember

sebersole: To be clear, this is on the EntityKey and/or EntityEntry, not the entity itself.

BTW, you should be able to easily test this theory of delayed insertion being the culprit - change from `#persist` to `#save`.  Because the latter is expected to return the generated id, we are not able to delay the insert there.  But otherwise, the behavior would be the same

NathanQingyangXu: > I debugged a little more and what I noticed id that with an assigned ID value before executing the query an autoflush is executed that for all `persistencecontext.collectionEntries` calls a `persistenceContext.addCollectionByKey` while for IDENTITY the autoflush is not executed before the query.

I traced the same root cause; then I further found the root cause is the "early insertion" in ActionQueue leaves no trace of  the table to be updated. As long as we fixed that, the flush would happen for Identity type. If you were interested, you could refer to my PR closed for details.

dreab8: paired with @beikov and reworked the solution.

@NathanQingyangXu the method you changed I think is used to tell if there are some data to be sync with the db in this case there are not, the insert had already happened.

NathanQingyangXu: > paired with @beikov and reworked the solution.
> 
> @NathanQingyangXu the method you changed I think is used to tell if there are some data to be sync with the db in this case there are not, the insert had already happened.

thanks for letting me know. obviously this bug is too complex to me, :(.

hibernate-core/src/main/java/org/hibernate/action/internal/AbstractEntityInsertAction.java
```diff
@@ -132,6 +138,27 @@ public final void makeEntityManaged() {
 				getPersister(),
 				isVersionIncrementDisabled
 		);
+		if ( isEarlyInsert() ) {
+			final SharedSessionContractImplementor session = getSession();
+			final PersistenceContext persistenceContext = session.getPersistenceContextInternal();
+			Object[] objects = getState();
+			for ( int i = 0; i < objects.length; i++ ) {
+				if ( objects[i] instanceof PersistentCollection<?> ) {
```
NathanQingyangXu, line 146: It is uncommon to specify generic type in `instanceof` expression for generic type has been erased during runtime, but Java compiler would not bother warning about this for it is purely unharmful.

PR #5041, opened 2022-05-11
HHH-15263 - @NamedQuery is not supported with UPDATE statement

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-15263

Not sure my code changes are totally correct. Feel free to work on top of it or even close it.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
@@ -817,14 +817,14 @@ public class QuerySqmImpl<R>
 
 	@Override
 	public SqmQueryImplementor<R> setLockOptions(LockOptions lockOptions) {
-		verifySelect();
+		// No verifySelect call, because in Hibernate we support locking in subqueries
 		getQueryOptions().getLockOptions().overlay( lockOptions );
 		return this;
 	}
 
 	@Override
 	public SqmQueryImplementor<R> setLockMode(String alias, LockMode lockMode) {
-		verifySelect();
+		// No verifySelect call, because in Hibernate we support locking in subqueries
 		getQueryOptions().getLockOptions().setAliasSpecificLockMode( alias, lockMode );
 		return this;
 	}
@@ -869,6 +869,7 @@ public class QuerySqmImpl<R>
 	@Override
 	public SqmQueryImplementor<R> setLockMode(LockModeType lockMode) {
 		if ( lockMode != LockModeType.NONE ) {
+			// JPA requires an exception to be thrown when this is not a select statement
 			verifySelect();
 		}
 		getSession().checkOpen( false );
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -817,14 +817,18 @@ public LockOptions getLockOptions() {
 
 	@Override
 	public SqmQueryImplementor<R> setLockOptions(LockOptions lockOptions) {
-		verifySelect();
+		if ( lockOptions.getLockMode() != LockMode.NONE ) {
+			verifySelect();
```
beikov, line 821:
```suggestion
			// This is required by the JPA spec
			verifySelect();
```

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -817,14 +817,18 @@ public LockOptions getLockOptions() {
...
 		getQueryOptions().getLockOptions().overlay( lockOptions );
 		return this;
 	}
 
 	@Override
 	public SqmQueryImplementor<R> setLockMode(String alias, LockMode lockMode) {
-		verifySelect();
+		if ( lockMode != LockMode.NONE ) {
+			verifySelect();
+		}
```
beikov, line 831: IMO we should remove this restriction, as that is a Hibernate method and we actually do support locking in subqueries. Wdyt @dreab8, @sebersole ?

dreab8: I'm fine with removing this check here

NathanQingyangXu:
Steve, is it because JPA has the feature that only select query can lock at
the same time?

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
NathanQingyangXu: I can add it, but there is similar comment within `verifySelect()` already:
```
	protected void verifySelect() {
		try {
			SqmUtil.verifyIsSelectStatement( getSqmStatement(), hql );
		}
		catch (IllegalQueryOperationException e) {
			// per JPA
			throw new IllegalStateException( "Expecting a SELECT query : `" + hql + "`", e );
		}
	}
```

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
NathanQingyangXu: I made the changes to be in line with another method in the same class:
```
	@Override
	public SqmQueryImplementor<R> setLockMode(LockModeType lockMode) {
		if ( lockMode != LockModeType.NONE ) {
			verifySelect();
		}
		getSession().checkOpen( false );
		getQueryOptions().getLockOptions().setLockMode( LockMode.fromJpaLockMode( lockMode ) );
		return this;
	}
```
@sebersole , saw the code was written recently. Should we remove this check (maybe everywhere in this class)?

beikov:
Thanks

PR #5045, opened 2022-05-12
HHH-15153 - @OneToOne does not support @EmbeddedId

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-15153

The root cause is the following code snippet in 
```
boolean referencesDerivedId = false;
try {
	referencesDerivedId = otherSide.getIdentifier() instanceof Component
			&& ( (Component) otherSide.getIdentifier() ).getProperty( mappedBy ) != null;
}
catch ( MappingException e ) {
	// ignore
}
boolean referenceToPrimaryKey  = referencesDerivedId || mappedBy == null;
value.setReferenceToPrimaryKey( referenceToPrimaryKey );
```
We reuse the `Component#getProperty(String)` to test whether `mappedBy` property exists or not; however, when the property cannot be found, the `MappingException` would be caught and then ignored, ending up with `referencesDerivedId ` remaining `false`.

Another noticeable performance optimization trick is if `mappedBy` has been null, we don't need to go about such property checking at all!  

Relying on exception side effect to go about existence testing works, but error-prone and terribly slow. So I introduced a new `hasProperty(String)` method for this purpose.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/OneToOneSecondPass.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/OneToOneSecondPass.java
@@ -223,15 +223,7 @@ public class OneToOneSecondPass implements SecondPass {
 				// HHH-6813
 				// Foo: @Id long id, @OneToOne(mappedBy="foo") Bar bar
 				// Bar: @Id @OneToOne Foo foo
-				boolean referencesDerivedId = false;
-				try {
-					referencesDerivedId = otherSide.getIdentifier() instanceof Component
-							&& ( (Component) otherSide.getIdentifier() ).getProperty( mappedBy ) != null;
-				}
-				catch ( MappingException e ) {
-					// ignore
-				}
-				boolean referenceToPrimaryKey  = referencesDerivedId || mappedBy == null;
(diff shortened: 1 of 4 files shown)


NathanQingyangXu:
I created some basic benchmark code to compare the two approaches:
```
	String[] properties = { "id", "value", "entity2" };
	String mappedBy = "entity1";

	@Test
	void test() {
		boolean found = false;

		long then = System.nanoTime();
		try {
			hasPropertyBasedOnException();
			found = true;
		} catch ( IllegalStateException ignored) {
			found = false;
		}
		long time1 = System.nanoTime() - then;

		then = System.nanoTime();
		found = hasPropertySimple();
		long time2 = System.nanoTime() - then;

		System.out.println("Time spent for approach based on exception: " + time1);
		System.out.println("Time spent for simple approach: " + time2 );
	}

	private void hasPropertyBasedOnException() {
		for ( String property : properties ) {
			if ( property.equals( mappedBy ) ) {
				return;
			}
		}
		throw new IllegalStateException("impossible");
	}

	private boolean hasPropertySimple() {
		for ( String property : properties ) {
			if ( property.equals( mappedBy ) ) {
				return true;
			}
		}
		return false;
	}
```

This is one example output:
Time spent for approach based on exception: **59700**
Time spent for simple approach:                        **2100**

beikov (approved):
LGTM

davidmoten:
Can the fix be backported to 5.x as well please. My workplace is still on Java 8.

PR #5050, opened 2022-05-13
HHH-15267 Make ParameterBindingsMemento extend Serializable so that cache keys can be serialized

beikov:
[…](https://hibernate.atlassian.net/browse/HHH-15267)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/spi/QueryKey.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/spi/QueryKey.java
@@ -26,7 +26,7 @@ public class QueryKey implements Serializable {
 	/**
 	 * todo (6.0) : integrate work from original 6.0 branch
 	 */
-	public interface ParameterBindingsMemento {
+	public interface ParameterBindingsMemento extends Serializable {
 	}
 
 	public static QueryKey from(


hibernate-core/src/main/java/org/hibernate/cache/spi/QueryKey.java
```diff
@@ -26,7 +26,7 @@ public class QueryKey implements Serializable {
 	/**
 	 * todo (6.0) : integrate work from original 6.0 branch
 	 */
-	public interface ParameterBindingsMemento {
+	public interface ParameterBindingsMemento extends Serializable {
```
NathanQingyangXu, line 29: Is it possible similar issue still exists if the following `values` array contains object not implementing `Serializable`? Hopefully not.
```
	private static class ParameterBindingsMementoImpl implements QueryKey.ParameterBindingsMemento {
		private final Object[] values;
```

beikov: Of course this can happen, but there is no way to magically make parameter values serializable which are provided by a user.

NathanQingyangXu: yeah, makes sense.

NathanQingyangXu: My original concern is whether we can refactor the `values` type to `Serializable[]`, but it might be complex and has ripple effect. The current approach is good enough.

PR #5051, opened 2022-05-13
HHH-15269 Update NativeQuery unable to use h-schema placeholder, just emits placeholder contents into SQL

dreab8:
https://hibernate.atlassian.net/browse/HHH-15269


sebersole (approved):
:+1:

yrodiere (approved):

PR #5061, opened 2022-05-16
HHH-15045 avoid dirty-flush for OneToOne type

NathanQingyangXu:

For the following simple OneToOne case:
```
@Entity
class User {
  @Id int id;

  @OneToOne(mappedBy = "user")
  Profile profile;
}

@Entity
class Profile {
  @Id int id;

  @OneToOne
  User user;
}
```
Internally the 'user' field in `Profile` would be treated as `@ManyToOne`, not `@OneToOne`. the `mappedBy` side is treated as the only `@OneToOne` type. For this reason, OneToOneType is always clean, never **dirty**.

This is a regression issue after https://github.com/hibernate/hibernate-orm/pull/3590 (for https://hibernate.atlassian.net/browse/HHH-14216). After reverting back the changes in `OneToOneType` class, the testing case is still fine in that PR.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
@@ -106,17 +106,12 @@ public class OneToOneType extends EntityType {
 
 	@Override
 	public boolean isDirty(Object old, Object current, SharedSessionContractImplementor session) {
-		if ( isSame( old, current ) ) {
-			return false;
-		}
-
-		return getIdentifierType( session )
-				.isDirty( getIdentifier( old, session ), getIdentifier( current, session ), session );
+		return false;
 	}
 
 	@Override
 	public boolean isDirty(Object old, Object current, boolean[] checkable, SharedSessionContractImplementor session) {
-		return isDirty(old, current, session);
+		return false;
 	}
 
 	@Override
@@ -159,6 +154,12 @@ public class OneToOneType extends EntityType {
 
 	@Override
 	public Object assemble(Serializable oid, SharedSessionContractImplementor session, Object owner) throws HibernateException {
+		if ( oid == null ) {
+			if ( uniqueKeyPropertyName != null ) {
+				return resolve( session.getContextEntityIdentifier( owner ), session, owner );
+			}
+			return null;
+		}
 
 		//the owner of the association is not the owner of the id
 		Object id = getIdentifierType( session ).assemble( oid, session, null );
@@ -169,9 +170,14 @@ public class OneToOneType extends EntityType {
 
 		return resolveIdentifier( id, session );
 	}
-	
+
+	/**
+	 * We don't need to dirty check one-to-one because of how
+	 * assemble/disassemble is implemented and because a one-to-one
+	 * association is never dirty
+	 */
 	@Override
 	public boolean isAlwaysDirtyChecked() {
-		return true;
+		return false;
 	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/type/EntityType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/EntityType.java
@@ -385,7 +385,7 @@ public abstract class EntityType extends AbstractType implements AssociationType
 	/**
 	 * Resolve an identifier or unique key value
 	 */
-	private Object resolve(Object value, SharedSessionContractImplementor session, Object owner) {
+	protected Object resolve(Object value, SharedSessionContractImplementor session, Object owner) {
 		if ( value != null && !isNull( owner, session ) ) {
 			if ( isReferenceToPrimaryKey() ) {
 				return resolveIdentifier( value, session, null );
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/onetoone/flush/DirtyFlushTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/onetoone/flush/DirtyFlushTest.java
@@ -1,105 +1,87 @@
 package org.hibernate.orm.test.onetoone.flush;
 
-import java.util.List;
-import java.util.Map;
-
-import org.hibernate.CallbackException;
-import org.hibernate.Interceptor;
 import org.hibernate.cfg.AvailableSettings;
-import org.hibernate.jpa.boot.spi.Bootstrap;
-import org.hibernate.orm.test.jpa.SettingsGenerator;
-import org.hibernate.type.Type;
 
 import org.hibernate.testing.TestForIssue;
-import org.hibernate.testing.orm.jpa.PersistenceUnitDescriptorAdapter;
-import org.hibernate.testing.orm.junit.DialectContext;
-import org.hibernate.testing.orm.transaction.TransactionUtil;
+import org.hibernate.testing.orm.junit.EntityManagerFactoryScope;
+import org.hibernate.testing.orm.junit.Jpa;
+import org.hibernate.testing.orm.junit.Setting;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import jakarta.persistence.Entity;
-import jakarta.persistence.EntityManagerFactory;
 import jakarta.persistence.Id;
 import jakarta.persistence.OneToOne;
+import jakarta.persistence.Version;
 
 /**
  * @author Nathan Xu
  */
 @TestForIssue( jiraKey = "HHH-15045" )
+@Jpa(
+	annotatedClasses = {
+		DirtyFlushTest.User.class,
+		DirtyFlushTest.Profile.class
+	},
+	properties = @Setting(name = AvailableSettings.GLOBALLY_QUOTED_IDENTIFIERS, value = "true")
+)
 class DirtyFlushTest {
 
-	List<Class> getAnnotatedClasses() {
-		return List.of( User.class, Profile.class );
-	}
-
-	Map basicSettings() {
-		return SettingsGenerator.generateSettings(
-				AvailableSettings.HBM2DDL_AUTO, "create-drop",
-				AvailableSettings.DIALECT, DialectContext.getDialect().getClass().getName(),
-				AvailableSettings.LOADED_CLASSES, getAnnotatedClasses(),
-				AvailableSettings.GLOBALLY_QUOTED_IDENTIFIERS, "true"
-		);
-	}
-
-	EntityManagerFactory buildEntityManagerFactory(Map settings) {
-		return Bootstrap
-				.getEntityManagerFactoryBuilder( new PersistenceUnitDescriptorAdapter(), settings )
-				.build();
-	}
-
-	EntityManagerFactory entityManagerFactory;
-
 	@BeforeEach
-	void setUp() {
-		DirtyFlushInterceptor.dirtyFlushedForUser = false;
+	void setUp(EntityManagerFactoryScope scope) {
+		scope.inTransaction( em -> {
+			final User user = new User();
+			user.id = 1;
+			user.version = 1;
+
+			final Profile profile = new Profile();
+			profile.id = 1;
+			profile.version = 1;
+
+			em.persist( user );
+			em.persist( profile );
+		} );
 	}
 
 	@Test
-	void testDirtyFlushNotHappened() {
-		var settings = basicSettings();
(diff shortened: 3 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
```diff
@@ -169,9 +164,14 @@ public Object assemble(Serializable oid, SharedSessionContractImplementor sessio
 
 		return resolveIdentifier( id, session );
 	}
-	
+
+	/**
+	 * We don't need to dirty check one-to-one because of how
+	 * assemble/disassemble is implemented and because a one-to-one
+	 * association is never dirty
+	 */
```
NathanQingyangXu, line 172: The reverted original comment says all

hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
```diff
@@ -159,6 +154,12 @@ public Serializable disassemble(Object value, SharedSessionContractImplementor s
 
 	@Override
 	public Object assemble(Serializable oid, SharedSessionContractImplementor session, Object owner) throws HibernateException {
+		if ( oid == null ) {
+			if ( uniqueKeyPropertyName != null ) {
+				return resolve( session.getContextEntityIdentifier( owner ), session, owner );
+			}
+			return null;
+		}
```
NathanQingyangXu, line 162: @dreab8 , I saw there were quite some back and forth (see https://hibernate.atlassian.net/browse/HHH-15044); but I think after we revert back the dirty flush related code changes, the above code seems to become necessary again. I am not sure about this, though. Feel free to take over and work on top of my PR or even close it.

NathanQingyangXu: Now there are quite some broken testing cases, but I believe it is due to the above code. Maybe it needs more polishing.

deAtog:
Please do not change the behavior here. The one-to-one is dirty whenever it is associated becomes associated because it needs to track when it is associated. The association does not change the way the object is stored in the database, but does change how it is stored in the cache. If you revert this code you you will cause a cache miss and a database hit whenever `resolve` is called. This negatively impacts performance and is the reason it was changed.

NathanQingyangXu:
But it seems your testing case was not broken when it was reverted back. Am
I missing something?

NathanQingyangXu:
> Looking back at those tests, they do not really seem to test the dirty aspect of the OneToOneType with respect to caching. I think you should implement some other tests to ensure that the behavior is as expected before and after your changes.
> 
> Before my changes, `disassemble` always returned null. Now it returns the id of the associated object or null if there is nothing associated. Likewise, `assemble` previously always called `resolve` using the identifier of the owner. The new implementation returns null whenever the assembled identifier is null since the null value guarantees there was no associated object.
> 
> The changes in these implementations necessitate a dirty check because adding an associated object to an existing one when using a foreign key relationship should update the cache of the existing object, but not the database.

I am not sure we should change dirtiness to suit the cache purpose. To my feelings, the dirtiness should be retained to db sync up issue only. We should strive to achieve both goals (1. no touch on dirtiness or dirtiness should always be false; 2. make second-level cache working). Seems the current PR has satisfied the goals.

hibernate-core/src/main/java/org/hibernate/type/OneToOneType.java
NathanQingyangXu: Thanks. I agreed with Gavin King that most of the time when we touched OneToOne, we are making mistakes. I thought the original code changes by @deAtog introduced some bug in terms of misunderstanding of 'dirtiness' (without touching the dirtiness logic, his testing case still works). What puzzled me is the above new testing case added later. Seems the testing case is invalid one, but it seems to help to prove something.

deAtog:
As is right now, if it's marked dirty and there are no database changes the cache will be updated and the database will not. Unfortunately, there is not a way to determine whether or not it is dirty with respect to the cache or respect to the database. If you change the behavior here, performance will undoubtedly suffer whenever you have a local cache and a remote database. The purpose of my changes were to improve the cache performance while having a very minimal impact otherwise. Prior to my changes, whenever assemble was called with `fetch="join"` and a related object did not exist in the cache, a database query was performed to verify that there was indeed no related object. Given that the purpose of a one-to-one is to store extra data that is typically sparsely populated, these queries were very expensive to execute. 

I don't fault you trying to optimize the little bit of overhead incurred here, but I do not think you are going about it the right way. Also, I think there are far bigger fish out there than this.

deAtog:
I reviewed the above regression. I think we both agree that some mechanism is need to differentiate the whether the one-to-one is dirty from a caching perspective vs a database perspective. Right now the automatic versioning seems to treat the dirtiness of the one-to-one as the only condition required to increment the version, which ultimately is what causes the database update to occur. Without versioning, there isn't an issue. None of the existing tests cover this scenario, hence the regression.

NathanQingyangXu:
Thanks. Makes sense.

NathanQingyangXu:
> I reviewed the above regression. I think we both agree that some mechanism is need to differentiate the whether the one-to-one is dirty from a caching perspective vs a database perspective. Right now the automatic versioning seems to treat the dirtiness of the one-to-one as the only condition required to increment the version, which ultimately is what causes the database update to occur. Without versioning, there isn't an issue. None of the existing tests cover this scenario, hence the regression.

I debugged your testing case and set breakpoints into `isDirty()` method and found it always return false, never returning true. Given that these methods are purely designed for db dirtiness checking, it didn't surprise me (you didn't change the mechanism to use it for cache dirtiness purpose). I think your code changes in 'dirtiness' methods are unnecessary (actually it ends up hurting performance for repeated `isDirty()` method invocation always returning false).

I think we have good reason to revert back your code changes involving dirtiness checking, though yours in assemble() and disassemble() are valid.

NathanQingyangXu:
I found a good testing case and updated. When the OneToOne owner side update its association, the `mappedBy` side should not be marked as "dirty" and hence update it by issuing an unnecessary SQL update statement. The testing case uses `@Version` to prove that its version got bumped erroneously. If you check the SQL output in console, you could also saw the user table was updated erroneously.

Obviously the bug was because of misunderstanding of the dirtiness related method in `OneToOneTyp` in previous PR. Also the code changes was not justified for the testing case didn't prove we need to touch 'dirtiness' to make it work.

deAtog:
As I stated previously, the test cases I added during my initial PR did not test the dirtiness of the OneToOneType. I do not believe the changes you've made to `assemble` are correct. Calling `resolve` whenever the id is null will result in a cache miss and database hit in certain cases. If you go back to my initial PR, you'll see that I did the following:

1. Added failing tests based on existing behavior
2. Changed the way `assemble` and `disassemble` were implemented.
3. Added implementation for `IsDirty`.

Prior to adding implementations for `IsDirty` all my tests failed.

As I stated before, I think more tests are needed here. The one test you've added demonstrates a problem, but  you have not included enough tests to verify you are not breaking caching of the OneToOneType with `fetch = "join"` specified.

NathanQingyangXu:
This is complicated. The code I reverted back was from another PR to cover
another testing case. Your previous dirtiness code eclipsed it, but now it
was needed to cover the testing case. It was not added to cover my own
testing case. I didn't mean to make it perfect but simply correct the issue
we have confirmed.

NathanQingyangXu:
> As I stated previously, the test cases I added during my initial PR did not test the dirtiness of the OneToOneType. I do not believe the changes you've made to `assemble` are correct. Calling `resolve` whenever the id is null will result in a cache miss and database hit in certain cases. If you go back to my initial PR, you'll see that I did the following:
> 
> 1. Added failing tests based on existing behavior
> 2. Changed the way `assemble` and `disassemble` were implemented.
> 3. Added implementation for `IsDirty`.
> 
> Prior to adding implementations for `IsDirty` all my tests failed.
> 
> As I stated before, I think more tests are needed here. The one test you've added demonstrates a problem, but you have not included enough tests to verify you are not breaking caching of the OneToOneType with `fetch = "join"` specified.

I think we'd better limit the scope of this PR to dirtiness exclusively, not caching. If you were interested, maybe you can take over. But we'd better separate two issues: 1) cache; 2) dirtiness. The code I added was from @dreab8 to cover the later testing case after yours (with `mappedBy` scenario your testing case didn't cover). His code was reverted back but the testing case remained and I reverted back his code to cover that testing case exclusively.
I think the commonest usage of `@OneToOne` is the one on `mappedBy` side (see hibernate user guide for example), so the issue we talked about in this ticket is more serious. Could we fix this issue first and you can create a cache ticket later?
@dreab8 @beikov I wanna stop here for I think I've exposed the issue successfully and feel free to work on top of it or even close it.

NathanQingyangXu:
> As I stated previously, the test cases I added during my initial PR did not test the dirtiness of the OneToOneType. I do not believe the changes you've made to `assemble` are correct. Calling `resolve` whenever the id is null will result in a cache miss and database hit in certain cases. If you go back to my initial PR, you'll see that I did the following:
> 
> 1. Added failing tests based on existing behavior
> 2. Changed the way `assemble` and `disassemble` were implemented.
> 3. Added implementation for `IsDirty`.
> 
> Prior to adding implementations for `IsDirty` all my tests failed.
> 
> As I stated before, I think more tests are needed here. The one test you've added demonstrates a problem, but you have not included enough tests to verify you are not breaking caching of the OneToOneType with `fetch = "join"` specified.

I am confused for I did inspect your commit history carefully. Especially, I found prior to step 3, your testing case has succeeded already. Could you double check this? Given dirtiness is well-defined meaning and is not directly related to (second-level) cache, it would surprise me if it is the key step to make your testing case succeed.

liangbodeng:
We need this PR desperately, which reverts the change made to @OneToOne dirty check in PR #3590. PR #3590 has been blocking our upgrading path from 5.4.x for months. It's nice to see caching improvement attempt, but we should ensure correctness first, which in this case is dirty check. There are other people like me eagerly waiting for it reverted back to old behavior.

dreab8:
@NathanQingyangXu Unfortunately this change seems to introduce a regression, I ported a [test](https://github.com/dreab8/hibernate-orm/commit/1d8101d715f1e87610282eb917a1894bd9093878) we have in 5.x but not in 6 and it fails.

beikov:
After some discussion, we came to the conclusion that the fix which was introduced with https://github.com/hibernate/hibernate-orm/pull/3590 for HHH-14216 is invalid, and we will revert it based on what this PR here does. The proper performance improvement for one-to-one associations is IMO to introduce "tombstone" cache entries for the entity type of the one-to-one association, which serve as null markers. This way, we would hit the cache instead of the database and see the tombstone cache entry, which we interpret as null.

gavinking:
> "tombstone" cache entries

OMG "tombstone" I love it.

liangbodeng:
> After some discussion, we came to the conclusion that the fix which was introduced with #3590 for HHH-14216 is invalid, and we will revert it based on what this PR here does. The proper performance improvement for one-to-one associations is IMO to introduce "tombstone" cache entries for the entity type of the one-to-one association, which serve as null markers. This way, we would hit the cache instead of the database and see the tombstone cache entry, which we interpret as null.

Really appreciate it! I can't wait to upgrade to the latest version of hibernate in our projects.

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/5330

NathanQingyangXu:
But I am not sure your reusing "dirtiness" (which is for db purpose) for
cache implementation detail is a good idea. It might hurt in the other way.
Also, I don't think reverting dirtiness back would hurt performance, on the
contrary, it might save a lot in other scenarios. I will investigate
further to demonstrate we can achieve both goals at the same time.

deAtog:
Maybe, and maybe not. I dug into the code for versioning and right now all one-to-one relationships are versioned. For a one-to-one relationship, there are the following scenarios to consider. 

1. One-to-one association that assumes both the parent and the child have the same primary key values.
2. One-to-one association that maps a foreign key column.
3. One-to-one association from an embeddable class to another entity.

In the first case, the relationship from the parent to the child should not be versioned as the child is dependent upon the parent.  In the reverse mapping from the child to the parent, the Id of the child should be versioned, but the one-to-one should not.

In the second case, the relationship from the parent to the child should not be versioned as the child is once again dependent upon the parent. In the reverse mapping from the child to the parent the one-to-one should be versioned as the parent Id is stored in a separate column.

In the third case, the relationship from the parent to the child should be versioned as the parent stores the Id of the child in a separate column. In the reverse mapping from the child to the parent, the one-to-one should not be versioned as  the child contains no information about the parent.

If the versionability of the one-to-one is corrected then the version will not be incremented no database update will occur. It is therefore my opinion that this issue can be solved by addressing the versionability of the one-to-one.

PR #5064, opened 2022-05-19
HHH-15283 - fix NPE for NamedNativeQuery + SqlResultSetMapping (columns)

dreab8:
https://hibernate.atlassian.net/browse/HHH-15283

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/results/complete/CompleteResultBuilderBasicValuedStandard.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/results/complete/CompleteResultBuilderBasicValuedStandard.java
@@ -27,7 +27,7 @@ import static org.hibernate.query.results.ResultsHelper.impl;
 /**
  * ResultBuilder for scalar results defined via:<ul>
  *     <li>JPA {@link jakarta.persistence.ColumnResult}</li>
- *     <li>`<return-scalar/>` as part of a `<resultset/>` stanza in `hbm.xml`</li>
+ *     <li>`&lt;return-scalar/&gt;` as part of a `&lt;resultset/&gt;` stanza in `hbm.xml`</li>
  * </ul>
  *
  * @author Steve Ebersole
@@ -60,7 +60,7 @@ public class CompleteResultBuilderBasicValuedStandard implements CompleteResultB
 
 	@Override
 	public Class<?> getJavaType() {
-		return explicitJavaType.getJavaTypeClass();
+		return explicitJavaType == null ? null : explicitJavaType.getJavaTypeClass();
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/query/resultmapping/NamedNativeQueryWithCountColumnTest.java
```diff
@@ -0,0 +1,84 @@
...
+						entityManager.persist( sample );
+					}
+				}
+		);
+	}
+
+	@Test
+	void testNamedNativeQuery(EntityManagerFactoryScope scope) {
+		scope.inTransaction( em -> {
+			Long count = em.createNamedQuery( "sample.count", Long.class ).getSingleResult();
```
NathanQingyangXu, line 53: currently two dialects testing failed due to the above statement (Class casting exception). Seems some db returns different types. Now I saw this ticket is such an edge case.

hibernate-core/src/test/java/org/hibernate/orm/test/query/resultmapping/NamedNativeQueryWithCountColumnTest.java
```diff
@@ -0,0 +1,84 @@
...
+			assertThat( count, is( 3l ) );
```
NathanQingyangXu, line 54: this is minor. But I think given the count is long type, `assertThat( count, is( 31L ) )` seems less confusing to me.

hibernate-core/src/main/java/org/hibernate/query/results/complete/CompleteResultBuilderBasicValuedStandard.java
```diff
@@ -166,14 +166,14 @@ public boolean equals(Object o) {
...
+		return Objects.equals( explicitJavaType, that.explicitJavaType );
 	}
 
 	@Override
 	public int hashCode() {
 		int result = explicitColumnName != null ? explicitColumnName.hashCode() : 0;
 		result = 31 * result + ( explicitType != null ? explicitType.hashCode() : 0 );
-		result = 31 * result + explicitJavaType.hashCode();
+		result = 31 * result + ( explicitJavaType != null ? explicitJavaType.hashCode() : 0 );
 		return result;
```
NathanQingyangXu, line 177: My initial change is to use Objects util methods aggressively as following:
```
		return Objects.equals( explicitColumnName, that.explicitColumnName )
				&& Objects.equals( explicitType, that.explicitType )
				&& Objects.equals( explicitJavaType, that.explicitJavaType );
```
and
```
return Objects.hash( explicitColumnName, explicitType, explicitJavaType );
```

but later on I reverted back the above changes for I think maybe you have some special reason to avoid the approach (consistent and error-free to me) for performance reason?

NathanQingyangXu:
I just realized the PR title might be imperfect. The condition is not normal real column (not tested though) but seudo column like 'count(*)', so renaming the title as "fix NPE for NamedNativeQuery + SqlResultSetMapping (seudo column like 'count(*)').

hibernate-core/src/test/java/org/hibernate/orm/test/query/resultmapping/NamedNativeQueryWithCountColumnTest.java
dreab8: restricted to h2

PR #5066, opened 2022-05-19
HHH-15286 Query with a Composite Identifier (@IdClass) throws an exception

dreab8:
https://hibernate.atlassian.net/browse/HHH-15286

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
@@ -153,6 +153,9 @@ public class DefaultLoadEventListener implements LoadEventListener {
 						return;
 					}
 				}
+				else if ( idClass.isInstance( event.getEntityId() ) ) {
+					return;
+				}
 			}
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
```diff
@@ -153,9 +153,12 @@ private void checkIdClass(
 						return;
 					}
 				}
+				else if ( idClass.isAssignableFrom( event.getEntityId().getClass() ) ) {
```
beikov, line 156: This is the fix, right? IMO we should stick to the `isInstance` check if possible instead of using `isAssignableFrom`

beikov (approved):
Apart from what I commented on, this looks good

beikov (approved):

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
dreab8: Thanks @beikov , changed it

PR #5067, opened 2022-05-20
HHH-15292 escape html chars in JavaDoc

NathanQingyangXu:
https://hibernate.atlassian.net/browse/HHH-15292

An example overlooking of html char escaping leads to wrong JavaDoc:
screenshot

The corresponding JavaDoc is:
```
 * ResultBuilder for scalar results defined via:<ul>
 *     <li>JPA {@link jakarta.persistence.ColumnResult}</li>
 *     <li>`<return-scalar/>` as part of a `<resultset/>` stanza in `hbm.xml`</li>
 * </ul>
```

After html char (most of time, only '<', '>' and '&' are needed to escape in JavaDoc), the warning number in gradle's javadoc task drops from 100 to 37.

Also fixed some obvious other JavaDoc issues along the way.


beikov:
Thanks

PR #5075, opened 2022-05-23
HHH-15281 INSERTs/UPDATEs no longer executed as JDBC Batch statements if hibernate.temp.use_jdbc_metadata_defaults is set to false

dreab8:
https://hibernate.atlassian.net/browse/HHH-15281


hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/ExtractedDatabaseMetaDataImpl.java
```diff
@@ -169,7 +169,8 @@ public static class Builder {
 		private boolean supportsNamedParameters;
 		private boolean supportsScrollableResults;
 		private boolean supportsGetGeneratedKeys;
-		private boolean supportsBatchUpdates;
+		// In absence of DatabaseMetaData batching updates is considered supported
```
Sanne, line 172:
```suggestion
		// In absence of DatabaseMetaData batching updates is assumed to be supported
```

dreab8: changed! Thanks!

beikov (approved):
LGTM

PR #5077, opened 2022-05-25
HHH-15281 INSERTs/UPDATEs no longer executed as JDBC Batch statements if hibernate.temp.use_jdbc_metadata_defaults is set to false

dreab8:
https://hibernate.atlassian.net/browse/HHH-15281


Sanne (approved):

Sanne:
thanks!

PR #5078, opened 2022-05-26
HHH-15300 - Lazy evaluation to statement.toString() when logging a slow query

koisyu:
Enabling logging of slow queries increases memory usage.
As a result of checking, statement.toString() is called even though the slow query threshold is not exceeded, and it seems to waste resources.
It would be good to change to call toString() only when the slow query threshold is exceeded.

[hibernate-orm/SqlStatementLogger.java at 6.0.2 · hibernate/hibernate-orm](https://github.com/hibernate/hibernate-orm/blob/6.0.2/hibernate-core/src/main/java/org/hibernate/engine/jdbc/spi/SqlStatementLogger.java#L144)

Jira Ticket : https://hibernate.atlassian.net/browse/HHH-15300


sebersole:
LGTM  Thanks @koisyu

koisyu:
Thanks 😄 @sebersole

PR #5079, opened 2022-05-26
HHH-15294 cockroachdb fixes

maesenka:
https://hibernate.atlassian.net/browse/HHH-15294

Some fixes that fix a number of test failures when running against CockroachDB.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
@@ -14,6 +14,7 @@ import java.util.Date;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.TimeZone;
+
 import jakarta.persistence.TemporalType;
 
 import org.hibernate.LockMode;
@@ -21,6 +22,8 @@ import org.hibernate.LockOptions;
 import org.hibernate.boot.model.TypeContributions;
 import org.hibernate.dialect.function.CommonFunctionFactory;
 import org.hibernate.dialect.function.FormatFunction;
+import org.hibernate.dialect.identity.CockroachDBIdentityColumnSupport;
+import org.hibernate.dialect.identity.IdentityColumnSupport;
 import org.hibernate.dialect.pagination.LimitHandler;
 import org.hibernate.dialect.pagination.OffsetFetchLimitHandler;
 import org.hibernate.dialect.sequence.PostgreSQLSequenceSupport;
@@ -92,9 +95,10 @@ import static org.hibernate.type.descriptor.DateTimeUtils.appendAsTimestampWithM
  */
 public class CockroachDialect extends Dialect {
 
+	private static final CockroachDBIdentityColumnSupport IDENTITY_COLUMN_SUPPORT = new CockroachDBIdentityColumnSupport();
 	// KNOWN LIMITATIONS:
-
 	// * no support for java.sql.Clob
+
 
 	private final PostgreSQLDriverKind driverKind;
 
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
```diff
@@ -223,6 +239,23 @@ public void contributeTypes(TypeContributions typeContributions, ServiceRegistry
 				jdbcTypeRegistry.addDescriptorIfAbsent( PostgreSQLJsonJdbcType.INSTANCE );
 			}
 		}
+
+		// Force Blob binding to byte[] for CockroachDB
+		jdbcTypeRegistry.addDescriptor( Types.BLOB, VarbinaryJdbcType.INSTANCE );
+		jdbcTypeRegistry.addDescriptor( Types.CLOB, VarcharJdbcType.INSTANCE );
```
beikov, line 245:
```suggestion
		jdbcTypeRegistry.addDescriptor( Types.BLOB, BlobJdbcType.PRIMITIVE_ARRAY_BINDING );
		jdbcTypeRegistry.addDescriptor( Types.CLOB, ClobJdbcType.STRING_BINDING );
```

I think this would be more in-line with what we have elsewhere.

beikov (approved):
Other than what I commented, this looks good to me. You can squash and push a rebased commit when you feel confident. Also, please backport this to the 6.0 branch or let me know if I should do that for you.

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
maesenka: I tried that, but it doesn't work on CockroachDB. I changed the descriptor to varbinary because in CockroachDB BLOB is an alias for the BYTES type

beikov: What's the problem then when using what I proposed?

maesenka: The postgresdriver expects the Blob column to contain a Long (which is the OID for the blob), but CockroachDB stored the Blob object in the column as a byte array. This then leads to this exception: 
```
Caused by: org.postgresql.util.PSQLException: Bad value for type long : \xaced0005737200116a6176612e7574696c2e486173684d61700507dac1c31660d103000246000a6c6f6164466163746f724900097468726573686f6c6478703f4000000000000c7708000000100000000274000464617461740009696d616765646174617400046d6574617400086d6574616461746178
	at app//org.postgresql.jdbc.PgResultSet.toLong(PgResultSet.java:3043)
	at app//org.postgresql.jdbc.PgResultSet.getLong(PgResultSet.java:2216)
	at app//org.postgresql.jdbc.PgResultSet.getBlob(PgResultSet.java:428)
```
(this is from the o.h.o.test.jpa.lob.BlobTest)

beikov: Yeah, I get that, but when you use the code Suggestion I give it should use the same JDBC methods as for Varchar and Varbinary

maesenka: Apparently, it does not. After applying the suggested code changes, I still get the exact same exception.

maesenka: Btw, your suggested code is what I also initially did. But, as I said, it didn't fix the issue. So that's when I went with the solution of using VarbinaryJdbcType

PR #5080, opened 2022-05-26
HHH-14614 - Better support for JUnit 5 on-failure handling 

sebersole:
HHH-14614 - Better support for JUnit 5 on-failure handling

- this commit is just the "infrastructure" piece.
- still need to come back and use it.  

The benefit is this allows us to perform some action when tests fail.  E.g., atm we are forced to create and drop test data for each test method - `@BeforeEach` and `@AfterEach`.  This support allows us to use `@BeforeAll` and `@AfterAll`.  Previously this would be a potential problem when tests fail.  This support though let's us hook into the "failure process" to rebuild the SessionFactory when we have these failures

### DIFF
--- a/hibernate-testing/src/main/java/org/hibernate/testing/orm/junit/JUnitHelper.java
+++ b/hibernate-testing/src/main/java/org/hibernate/testing/orm/junit/JUnitHelper.java
@@ -6,11 +6,17 @@
  */
 package org.hibernate.testing.orm.junit;
 
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.util.List;
+
 import org.hibernate.Internal;
 
 import org.junit.jupiter.api.extension.Extension;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ParameterContext;
+
+import org.jboss.logging.Logger;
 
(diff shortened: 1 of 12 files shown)


hibernate-testing/src/main/java/org/hibernate/testing/orm/junit/JUnitHelper.java
```diff
@@ -32,6 +42,73 @@ public static ExtensionContext.Store locateExtensionStore(
...
+		if ( callbacks == null ) {
+			return;
+		}
+
+		callbacks.forEach( (callback) -> {
+			try {
+				context.getExecutableInvoker().invoke( callback, context.getRequiredTestInstance() );
+			}
+			catch (Exception e) {
+				log.warnf( e, "Error invoking FailureExpectedCallback handling : `%`", callback.getName() );
```
NathanQingyangXu, line 107: not 100% sure, but should we use `%s` placeholder above?

yrodiere:
It seems this was never merged (cannot find this code or a commit mentioning HHH-14614), but the corresponding issue was closed as fixed: https://hibernate.atlassian.net/browse/HHH-14614

I'll tentatively close this PR, feel free to reopen if that's wrong.

PR #5085, opened 2022-06-01
Add minimum dialect version

jrenaat:


beikov (commented):
Yes, but every dialect should now implement this method and return the version which they currently pass to `super`/`this` in their respective default constructors. Then, we can also check in the `Dialect` constructor if the passed version is greater or equal to the minimum version, and if it isn't, log a warning.

jrenaat:
> Yes, but every dialect should now implement this method and return the version which they currently pass to `super`/`this` in their respective default constructors. Then, we can also check in the `Dialect` constructor if the passed version is greater or equal to the minimum version, and if it isn't, log a warning.

I know, I just issued this PR too quickly

hibernate-core/src/main/java/org/hibernate/internal/CoreMessageLogger.java
```diff
@@ -1792,4 +1792,8 @@ void attemptToAssociateProxyWithTwoOpenSessions(
 	@Message(value = "Association with `fetch=\"join\"`/`@Fetch(FetchMode.JOIN)` and `lazy=\"true\"`/`FetchType.LAZY` found. This will be interpreted as lazy: %s", id = 510)
 	void fetchModeJoinWithLazyWarning(String role);
 
+	@LogMessage(level = WARN)
+	@Message(value = "This version for [%s] is no longer supported, certain features may not work properly. Check the community dialects project for available legacy versions.", id = 511)
```
beikov, line 1796: Please also log the actual version and minimum version.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -259,17 +264,25 @@ protected Dialect() {
...
 	protected Dialect(DialectResolutionInfo info) {
 		this.version = info.makeCopy();
+		checkVersion();
 		registerDefaultKeywords();
 		registerKeywords(info);
 		initDefaultProperties();
 	}
 
+	protected void checkVersion() {
+		if(this.version != null && this.version.isBefore( getMinimumSupportedVersion() )) {
```
beikov, line 281: Let's give people a chance to use an overridden version by calling the `getVersion` method instead.

```suggestion
		final DatabaseVersion version = getVersion();
		if (version != null && version.isBefore( getMinimumSupportedVersion() ) ) {
```

beikov (commented):
I like the approach. Could you please, in a separate commit, alter all dialects so that they make proper use of this?

jrenaat:
> I like the approach. Could you please, in a separate commit, alter all dialects so that they make proper use of this?

I would actually like to keep the basic implementation separate from the specific minimun database versions. I can add those as I go with the legacy removals. I'm not even sure we already have a consensus for minimum supported version for all dialects right now?

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
jrenaat: :+1:

hibernate-core/src/main/java/org/hibernate/internal/CoreMessageLogger.java
jrenaat: :+1:

jrenaat:
I wonder if the minimum version should also include the micro-part. What do y'all think?

beikov:
>I wonder if the minimum version should also include the micro-part. What do y'all think?

Sure, add the micro version also. At least for HANA that one matters.

PR #5088, opened 2022-06-04
HHH-13741 : Logging SQL query when using LOG_SLOW_QUERY

jai-yadav:
[HHH-13741](https://hibernate.atlassian.net/browse/HHH-13741)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/ResultSetReturnImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/ResultSetReturnImpl.java
@@ -11,6 +11,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.function.Supplier;
 
 import org.hibernate.dialect.Dialect;
 import org.hibernate.engine.jdbc.spi.JdbcCoordinator;
@@ -58,7 +59,7 @@ public class ResultSetReturnImpl implements ResultSetReturn {
 			}
 			finally {
 				jdbcExecuteStatementEnd();
-				sqlStatementLogger.logSlowQuery( statement, executeStartNanos );
+				sqlStatementLogger.logSlowQuery( getStatementSql( statement ), executeStartNanos );
 			}
 			postExtract( rs, statement );
 			return rs;
@@ -91,7 +92,7 @@ public class ResultSetReturnImpl implements ResultSetReturn {
 			}
 			finally {
 				jdbcExecuteStatementEnd();
-				sqlStatementLogger.logSlowQuery( callableStatement, executeStartNanos );
+				sqlStatementLogger.logSlowQuery( getStatementSql( callableStatement ), executeStartNanos );
 			}
 			postExtract( rs, callableStatement );
 			return rs;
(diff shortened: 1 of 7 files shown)


ibre5041:
Hi,
thanks for this PR. I'd like to ask: Will there be a way how to override getStatement() method?
So returned string could be formatter or extended?

As Oracle DBA I'm looking for a way how to relate SQL statement logged by application with SQL statement seen in the database. And it is still very hard because SQLs logged by Hibernate are formatted, truncated and moreover JDBC driver internally replaces bind parameters '?' with enumerated bind parameters ':1'. 

So for me id would be great if I could call OracleStatement.getOriginalSQL() somewhere and I could log exact SQL statement which was sent into the database (rather that some similar, truncated fragment of SQL)

Moreover all SQLs in Oracle are identified by SQL_ID which is MD5 of original SQL sent into the database. 
I wrote piece of code which can compute this SQL_ID in Java, so slow SQLs could be logged as SQL_IDs, and the rest of diagnostic information could be found in the database.

https://tanelpoder.com/2009/02/22/sql_id-is-just-a-fancy-representation-of-hash-value/

beikov:
Hey @jai-yadav and thanks for your contribution. A few comments:

* Sometimes you are trying to access the SQL through the registry even when the SQL is given as argument, that should be fixed
* I don't think having a registry is for the SQL strings is necessary, as you can unwrap statements like `statement.unwrap(Statement.class)` and call `toString` on that
* For Oracle, I think I'd personally prefer accessing `oracle.jdbc.driver.OracleStatement#getOriginalSql` reflectively through a static final method instead. Alternatively, we could add `org.hibernate.engine.jdbc.spi.ResultSetReturn#extract(java.sql.PreparedStatement, java.lang.String)` variants which will use the passed in SQL string just for logging purposes

Any thoughts on that @sebersole ?

sebersole:
Sorry, but I have no idea what you are talking about...

```
package org.hibernate.query;

public interface Query<R> ... {

	/**
	 * Set the comment for this query.
	 * <p>
	 * If SQL commenting is enabled, the comment will be added to the SQL
	 * query sent to the database, which may be useful for identifying the
	 * source of troublesome queries.
	 * <p>
	 * SQL commenting may be enabled using the configuration property
	 * {@value org.hibernate.cfg.AvailableSettings#USE_SQL_COMMENTS}.
	 *
	 * @param comment The human-readable comment
	 *
	 * @return {@code this}, for method chaining
	 *
	 * @see #getComment()
	 */
	Query<R> setComment(String comment);
}
```

Criteria queries must be "compiled" into a Query (aka ^^) to execute

jai-yadav:
Hi @beikov Thanks for your feedback.
Agreed on 1st, Added changes to use SQL directly when available instead of using registry in those cases.

But I'm not clear on 2nd and 3rd points how they can help in logging plain SQL text.
Using statement.unwrap(Statement.class) doesn't seems to provide SQL on toString().
Unwrap

Enixes:
Hi,
I am also waiting for [HHH-13741](https://hibernate.atlassian.net/browse/HHH-13741). Any plans to merge this PR?

hojongs:
Hi, is there any update? I'm looking forward to this great work!

gksxodnd007:
Hi, is there any update? I'm looking forward to this great work!!!

sebersole:
See https://github.com/hibernate/hibernate-orm/pull/7030 for (imo) a better fix

beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/7030

PR #5089, opened 2022-06-05
HHH-15317 DeepCopy fields when marking an entity non-readonly

ahus1:
Otherwise a modification might get unnoticed in the next flush.

This is a draft PR for now as it doesn't yet contain a test case, and I'm not sure which branch to submit this one to.

I'd be happy for some advice: 
* if submitting the PR for 5.6 is ok, as I'd like to see a fix for 5.6 - or should I rather submit it for 6.0?
* could you point me to a test case example I could extend that already uses some custom types?
* could the test cases be more minimized?
* between 6.0 and 5.6 some classes about user defined types changed - see migration guide. Should I eventually submit two PRs? 

Thanks!


ahus1:
Obsolete due to the work in #7355

PR #5094, opened 2022-06-07
very experimental impl of Hibernate.size()

gavinking:
Get the size of a collection without fetching it.


gavinking:
Test failures don't appear to be related to my change. They are in `HibernateOrmPluginTest`.

beikov (approved):
I don't see a use case for this, but the implementation seems fine to me. It would be nice though to have a JIRA issue that describes the motivation a bit.

gavinking:
@beikov we discussed it here: https://github.com/hibernate/hibernate-orm/discussions/3779

It's a better alternative to "extra lazy" collections.

sebersole:
Looks great to me.

Though +1 to Jira

gavinking:
> Though +1 to Jira

Yeah yeah I know, I'm on it.

gavinking:
@sebersole what `@since` tag should I put on these methods?

sebersole:
> @sebersole what `@since` tag should I put on these methods?

Good question....

I think these are "niche" enough to put in a bug fix, so 6.1.1.  Sound reasonable?

gavinking:
> I think these are "niche" enough to put in a bug fix, so 6.1.1. Sound reasonable?

Sure.

PR #5096, opened 2022-06-08
Test inheritance with  MySQL 8.0.29

DavideD:
Test for latest MySQL connector: https://bugs.mysql.com/bug.php?id=107447


DavideD:
I've just realized that I need to test this on 5.6

PR #5097, opened 2022-06-09
Use --enable-preview when testing against JDK 19

yrodiere:
Plus a few minor improvements in the Jenkinsfile to make adding that `--enable-preview` arg easier.

Not sure referencing `DEFAULT_JDK_VERSION` from the `BuildEnvironment` class will work correctly, so let's see if CI passes...


beikov (approved):
Nice simplification work. You can tell by my style that I mostly use Java :D

yrodiere:
> Not sure referencing `DEFAULT_JDK_VERSION` from the `BuildEnvironment` class will work correctly, so let's see if CI passes...

Well it didn't work, so I changed that part.

PR #5110, opened 2022-06-15
HHH-15325 Avoid allocations from BitSet.stream() in AbstractEntityPersister

dreis2211:
Backport of https://hibernate.atlassian.net/browse/HHH-15325


beikov:
Thanks!

PR #5116, opened 2022-06-15
HHH-15180 - Remove support for SQL Server versions older than 2008

jrenaat:
https://hibernate.atlassian.net/browse/HHH-15180

### DIFF
--- a/hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/SQLServerLegacyDialect.java
+++ b/hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/SQLServerLegacyDialect.java
@@ -4,7 +4,7 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
-package org.hibernate.dialect;
+package org.hibernate.community.dialect;
 
 import org.hibernate.*;
 import org.hibernate.boot.Metadata;
@@ -12,6 +12,11 @@ import org.hibernate.boot.model.TypeContributions;
 import org.hibernate.boot.model.relational.QualifiedSequenceName;
 import org.hibernate.boot.model.relational.Sequence;
 import org.hibernate.boot.model.relational.SqlStringGenerationContext;
+import org.hibernate.dialect.AbstractTransactSQLDialect;
+import org.hibernate.dialect.DatabaseVersion;
+import org.hibernate.dialect.Dialect;
+import org.hibernate.dialect.Replacer;
+import org.hibernate.dialect.TimeZoneSupport;
 import org.hibernate.dialect.function.CommonFunctionFactory;
 import org.hibernate.dialect.function.CountFunction;
 import org.hibernate.dialect.function.SQLServerFormatEmulation;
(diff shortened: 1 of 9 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/SQLServerDialect.java
```diff
@@ -82,12 +82,13 @@
  * @author Gavin King
  */
 public class SQLServerDialect extends AbstractTransactSQLDialect {
+	private final static DatabaseVersion MINIMUM_VERSION = DatabaseVersion.make( 8, 0 );
```
beikov, line 85:
```suggestion
	private final static DatabaseVersion MINIMUM_VERSION = DatabaseVersion.make( 10, 0 );
```

hibernate-core/src/main/java/org/hibernate/dialect/SQLServerDialect.java
```diff
@@ -118,50 +124,46 @@ protected String columnType(int sqlTypeCode) {
 		if ( sqlTypeCode == DOUBLE ) {
 			return "float";
 		}
```
beikov, line 126: You can now move this into the switch block as an additional case

hibernate-core/src/main/java/org/hibernate/dialect/SQLServerDialect.java
```diff
@@ -677,9 +653,8 @@ public String extractPattern(TemporalUnit unit) {
 				return "(datepart(second,?2)+datepart(nanosecond,?2)/1e9)";
 			case WEEK:
 				// Thanks https://www.sqlservercentral.com/articles/a-simple-formula-to-calculate-the-iso-week-number
-				if ( getVersion().isBefore( 10 ) ) {
-					return "(DATEPART(dy,DATEADD(dd,DATEDIFF(dd,'17530101',?2)/7*7,'17530104'))+6)/7)";
-				}
+				return "(datepart(iso_week,?2))";
+				// DATEPART( dy, )
```
beikov, line 657: IMO, you should just remove the WEEK case

beikov (commented):
Other than what I commented on, this looks good to me

hibernate-core/src/main/java/org/hibernate/dialect/SQLServerDialect.java
jrenaat: Totally right, lapsus, good catch.

beikov (approved):

jrenaat:
Changed a couple of tests for sql server 2005 to the community project, and renamed a couple of others (left out the '2008' part from their name)

beikov (commented):
Can you make a copy of the moved test classes and leave one adapted to the new minimum database version in the hibernate-core module please?

jrenaat:
> Can you make a copy of the moved test classes and leave one adapted to the new minimum database version in the hibernate-core module please?

Sure, in this case I think it would only be about copying the `SQLServer2005DialectTestCase`. I don't know exactly what in this test was so '2005-version-specific', so i can only copy the test, have it run against sql server 2008 and hope that it works out (as it does in this case). Is that ok with you?

hibernate-core/src/test/java/org/hibernate/orm/test/type/SQLServerNVarCharTypeTest.java
```diff
@@ -33,7 +32,7 @@
  */
 @TestForIssue(jiraKey = "HHH-10529")
 @RequiresDialect(value = SQLServerDialect.class, majorVersion = 10)
```
beikov, line 34: Since version 10 is for SQL Server 2008, you can now drop the major version 10.

```suggestion
@RequiresDialect(value = SQLServerDialect.class)
```

hibernate-core/src/test/java/org/hibernate/orm/test/query/SQLServerNationalizedScalarQueryTest.java
```diff
@@ -31,10 +31,10 @@
 @TestForIssue(jiraKey = "HHH-10183")
 @RequiresDialect(value = SQLServerDialect.class, majorVersion = 10)
```
beikov, line 32: Since version 10 is for SQL Server 2008, you can now drop the major version 10.

```suggestion
@RequiresDialect(value = SQLServerDialect.class)
```

hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/SQLServerLegacyDialect.java
```diff
@@ -0,0 +1,948 @@
...
+			buffer.append( ";" );
+		}
+		sql = buffer.toString();
+
+		return sql;
+	}
+
+	@Override
+	public boolean supportsNullPrecedence() {
+		return getVersion().isBefore( 10 );
```
beikov, line 598: Looks like something got lost after the rebase:

```suggestion
		return false;
```

beikov (changes requested):
Seems that some of the original `SQLServerDialect` code was wrongly dropped in the legacy copy. Please fix that. Then we can merge this.

PR #5117, opened 2022-06-16
HHH-15340 Repair or skip test cases in `hibernate-core` and `document` for TiDB.

Icemap:
Jira Link: https://hibernate.atlassian.net/browse/HHH-15340

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/jpa/lock/RepeatableReadTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/jpa/lock/RepeatableReadTest.java
@@ -10,11 +10,9 @@ import java.math.BigDecimal;
 
 import org.hibernate.LockMode;
 import org.hibernate.StaleObjectStateException;
-import org.hibernate.TransactionException;
 import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
 import org.hibernate.cfg.AvailableSettings;
 import org.hibernate.dialect.SQLServerDialect;
-import org.hibernate.dialect.TiDBDialect;
 import org.hibernate.engine.jdbc.connections.spi.ConnectionProvider;
 import org.hibernate.exception.SQLGrammarException;
 import org.hibernate.orm.test.jpa.model.AbstractJPATest;
@@ -24,7 +22,6 @@ import org.hibernate.orm.test.jpa.model.Part;
 import org.hibernate.testing.jdbc.SQLServerSnapshotIsolationConnectionProvider;
 import org.hibernate.testing.orm.junit.DialectFeatureChecks;
 import org.hibernate.testing.orm.junit.RequiresDialectFeature;
-import org.hibernate.testing.orm.junit.SkipForDialect;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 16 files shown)


Icemap:
PTAL @sebersole @beikov

beikov:
I'm not sure I understand exactly why this is necessary. Are you a colleague of Cong Wang from PingCAP? The TiDB tests usually succeed right now already. The standalone instance just times out from time to time for some reason, but @bb7133 is looking into this.

Icemap:
Hello @beikov, I am Wang Cong's colleague and we both work at PingCAP. The current Hibernate test cases are adapted to TiDB c5.2, but we have upgraded to v6.1.

And we are building a Hibernate integration test pipeline for each PR of TiDB, and will do our best to make TiDB fully support Hibernate.

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/lock/RepeatableReadTest.java
```diff
@@ -239,41 +243,48 @@ public void testStaleNonVersionedInstanceFoundOnLock() {
...
-					assertTrue( part == part2 );
-					assertEquals( check, part2.getName(), "encountered non-repeatable read" );
-				}
-		);
+			);
+		} catch (TransactionException e) {
+			if (!(getDialect() instanceof TiDBDialect)) {
+				throw e;
+			}
+		}
```
beikov, line 287: Please revert these changes since you added a `@SkipDialect` anyway

beikov (changes requested):
Alright. So some of the test changes (related to constraint violations and optimistic locking) that you did would actually be unnecessary when you properly implement `org.hibernate.dialect.Dialect#buildSQLExceptionConversionDelegate` and `org.hibernate.dialect.Dialect#getViolatedConstraintNameExtractor` in the `TiDBDialect`. Could you please do that instead and revert the test changes? I'd hate to change the test code and sprinkle `instanceof TiDBDialect` checks around if that's not necessary in the long run.

Icemap:
Sorry @beikov , I ran the tests on a Mac with an M1 chip before. 
After revert the changes, I switched to CentOS 7 and the test was able to pass TiDB 6.1.0 with only a few tests skipped.

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/lock/RepeatableReadTest.java
Icemap: Reverted

beikov:
Can you explain a bit what this `TiDB returned record number same as MySQL` means? I guess the problem might be that you are using a case insensitive collation. Maybe you could switch to a case sensitive one? Also see how we start MySQL with the options ` --character-set-server=utf8mb4 --collation-server=utf8mb4_0900_as_cs` to achieve this here: https://github.com/hibernate/hibernate-orm/blob/main/docker_db.sh#L44

Maybe you don't have to skip tests after all :)

Icemap:
@beikov You're right. We used the `utf8mb4_general_ci` collate to test in the config file. Now I needn't to skip tests. Thank you so much for helping me! I'll close this PR.

beikov:
You're welcome :)

PR #5127, opened 2022-06-22
HHH-15342 Inappropriate variation of HQL left join to SQL inner join

dreab8:
https://hibernate.atlassian.net/browse/HHH-15342

This happens when we have association annotated with `@NotFound`

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
@@ -1196,9 +1196,33 @@ public class ToOneAttributeMapping
 				|| fetchParent.getNavigablePath() instanceof TreatedNavigablePath
 				&& parentNavigablePath.equals( fetchParent.getNavigablePath().getRealParent() );
 
+		/*
+		 In case of @NotFound we are going to add fetch for the `fetchablePath` only if there is not already a `TableGroupJoin`.
 
-		if ( (hasNotFoundAction()
-				|| ( fetchTiming == FetchTiming.IMMEDIATE && selected )) ) {
+		 e.g. given :
+		 	public static class EntityA {
+				...
+
+			@ManyToOne(fetch = FetchType.LAZY)
+			@NotFound(action = NotFoundAction.IGNORE)
+			private EntityB entityB;
+		 	}
+
+			@Entity(name = "EntityB")
+			public static class EntityB {
+				...
+
+				private String name;
+			}
+
+		 and the HQL query :
+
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3265,6 +3265,7 @@ private TableGroup createTableGroup(TableGroup parentTableGroup, SqmPath<?> join
 							false,
 							this
 					);
+					tableGroupJoin.setImplicit();
```
beikov, line 3268: I think it should be possible within `createTableGroupJoin` to determine whether a join is implicit, by checking if `sqlAstJoinType == null`, so I'd prefer to remove `setImplicit` and instead make the field final in `TableGroupJoin`.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -1191,8 +1191,8 @@ public EntityFetch generateFetch(
 				&& parentNavigablePath.equals( fetchParent.getNavigablePath().getRealParent() );
 
 
-		if ( (hasNotFoundAction()
-				|| ( fetchTiming == FetchTiming.IMMEDIATE && selected )) ) {
+		if ( ( ( hasNotFoundAction() && doesNotExistATableGroupJoin( parentTableGroup, fetchablePath ) )
```
beikov, line 1194: Would you mind explaining why we need `doesNotExistATableGroupJoin( parentTableGroup, fetchablePath )`? Maybe even as code comment?

beikov (commented):
Since the SQL we generate here is the same in all databases, I would prefer if you assert that the SQL string matches some expected SQL string, as that is way easier to understand IMO.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
dreab8: Unfortunately it seems not so easy :( I tried but stumbled in a lot of issues.
I tried different ways but without any success

dreab8: Let me know what do you think about the comment I added

beikov:
Here is my alternative suggestion which seems to work fine without the flag: https://github.com/hibernate/hibernate-orm/pull/5138

dreab8:
closing in favour of https://github.com/hibernate/hibernate-orm/pull/5138

PR #5129, opened 2022-06-23
HHH-15346 @ManyToOne associations not loaded correctly with default EAGER and batch fetch property set

dreab8:
https://hibernate.atlassian.net/browse/HHH-15346

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/BatchEntitySelectFetchInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/BatchEntitySelectFetchInitializer.java
@@ -6,7 +6,9 @@
  */
 package org.hibernate.sql.results.graph.entity.internal;
 
-import java.util.LinkedHashMap;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
 
@@ -47,7 +49,7 @@ public class BatchEntitySelectFetchInitializer extends AbstractFetchParentAccess
 	protected Object entityInstance;
 	private EntityKey entityKey;
 
-	private Map<EntityKey, Object> toBatchLoad = new LinkedHashMap<>();
+	private Map<EntityKey, List<Object>> toBatchLoad = new HashMap<>();
 
 	private boolean isInitialized;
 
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/BatchEntitySelectFetchInitializer.java
```diff
@@ -147,7 +148,13 @@ public void initializeInstance(RowProcessingState rowProcessingState) {
...
 
 		persistenceContext.getBatchFetchQueue().addBatchLoadableEntityKey( entityKey );
-		toBatchLoad.put( entityKey, parentAccess.getInitializedInstance() );
+		List<Object> objects = toBatchLoad.get( entityKey );
+		if ( objects == null ) {
+			objects = new ArrayList<>();
+		}
+		objects.add( parentAccess.getInitializedInstance() );
+
+		toBatchLoad.put( entityKey, objects );
```
frode-carlsen, line 157: this statement could be moved inside the if(objects==null) block

hibernate-core/src/test/java/org/hibernate/orm/test/batch/EagerManyToOneBatchTest.java
```diff
@@ -0,0 +1,123 @@
...
+							.setParameter( "name", "a" )
+							.getResultList();
+
+					assertEquals( 2, children.size() );
+
+					Child child1 = children.get( 0 );
+					assertNotNull( child1.getParent() );
+
+					Child child2 = children.get( 1 );
+					assertNotNull( child2.getParent() );
```
beikov, line 68: Shouldn't we maybe also assert the number of SQL queries that are executed?

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/BatchEntitySelectFetchInitializer.java
dreab8: Thanks @frode-carlsen for the suggestion

beikov (approved):

hibernate-core/src/test/java/org/hibernate/orm/test/batch/EagerManyToOneBatchTest.java
dreab8: I added the assertions and also an additional test class for a Lazy association

PR #5131, opened 2022-06-23
HHH-15191 - Remove support for Derby versions older than 10.14.2.0

jrenaat:

https://hibernate.atlassian.net/browse/HHH-15191

### DIFF
--- a/hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/DerbyLegacyDialect.java
+++ b/hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/DerbyLegacyDialect.java
@@ -4,15 +4,24 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
-package org.hibernate.dialect;
+package org.hibernate.community.dialect;
+
+import java.sql.DatabaseMetaData;
+import java.sql.SQLException;
+import java.sql.Types;
 
 import org.hibernate.boot.model.TypeContributions;
+import org.hibernate.dialect.DB2Dialect;
+import org.hibernate.dialect.DatabaseVersion;
+import org.hibernate.dialect.Dialect;
+import org.hibernate.dialect.NationalizationSupport;
+import org.hibernate.dialect.RowLockStrategy;
+import org.hibernate.dialect.function.CaseLeastGreatestEmulation;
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -115,7 +115,7 @@ public DerbyDialect(DialectResolutionInfo info) {
 	protected String columnType(int sqlTypeCode) {
 		switch ( sqlTypeCode ) {
 			case BOOLEAN:
-				return getVersion().isBefore( 10, 7 ) ? "smallint" : super.columnType( sqlTypeCode );
+				return super.columnType( sqlTypeCode );
```
beikov, line 118: You can remove the whole case for boolean instead

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -212,9 +212,7 @@ public int getDefaultDecimalPrecision() {
 
 	@Override
 	public int getPreferredSqlTypeCodeForBoolean() {
-		return getVersion().isBefore( 10, 7 )
-				? Types.SMALLINT
-				: Types.BOOLEAN;
+		return Types.BOOLEAN;
 	}
```
beikov, line 216: I think you can just remove the method as the super implementation will do the right thing

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
```diff
@@ -438,33 +436,22 @@ public String timestampdiffPattern(TemporalUnit unit, TemporalType fromTemporalT
...
 	@Override
 	public void appendBooleanValueString(SqlAppender appender, boolean bool) {
-		if ( getVersion().isBefore( 10, 7 ) ) {
-			appender.appendSql( bool ? '1' : '0' );
-		}
-		else {
-			appender.appendSql( bool );
-		}
+		appender.appendSql( bool );
 	}
```
beikov, line 440: I think you can just remove the method as the super implementation will do the right thing

beikov (commented):
Can you please copy the original version of DerbyDialectTest into the community-dialects module?

hibernate-core/src/main/java/org/hibernate/dialect/DerbyDialect.java
beikov: Yeah you are right, I misremembered the base implementation

PR #5137, opened 2022-06-27
HHH-15358  @Where annotation with globally_quoted_identifiers causes Unable to determine TableReference Exception

dreab8:
https://hibernate.atlassian.net/browse/HHH-15358


beikov (approved):
LGTM!

PR #5143, opened 2022-07-01
HHH-15331 Constructor expressions in the SELECT clause ignores attribute converter

dreab8:
https://hibernate.atlassian.net/browse/HHH-15331


dreab8:
merged

PR #5144, opened 2022-07-01
HHH-15382 CockroachDB doesn't support value propagation

maesenka:
Minor fix


maesenka:
It's related to all CockroachDB versions

PR #5150, opened 2022-07-05
HHH-15218 @OptimisticLocking(DIRTY) leads to wrong query during delete of circular reference

dreab8:
https://hibernate.atlassian.net/browse/HHH-15218


Sanne (approved):
nice, thanks!

PR #5159, opened 2022-07-08
HHH-15390 Calling  SqmQuery#getSqmStatement()#toHqlString() causes ClassCastExpection when using distinct in Criteria or HQL query

dreab8:
https://hibernate.atlassian.net/browse/HHH-15390


sebersole (commented):
lgtm

PR #5162, opened 2022-07-12
Backport snapshot-publish.Jenkinsfile to branch 5.6

yrodiere:
I'm not *completely* sure that's the right command/settings, but from what I gathered from disabled (and probably obsolete...) Jenkins jobs, that should be close. Let's try, worst case the CI run will just fail?

### DIFF
--- a/ci/snapshot-publish.Jenkinsfile
+++ b/ci/snapshot-publish.Jenkinsfile
@@ -1,0 +1,51 @@
+/*
+ * See https://github.com/hibernate/hibernate-jenkins-pipeline-helpers
+ */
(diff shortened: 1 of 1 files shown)


yrodiere:
Also, I intentionally did not backport all the other changes related to publishing snapshots to OSSRH or to publishing Gradle plugin snapshots. I don't need those right now, so that can be an improvement for later (if we end up needing that).

yrodiere:
Merging to try this...

yrodiere:
It worked. Backported to 5.5, 5.4, 5.3.

PR #5164, opened 2022-07-12
HHH-15394- memory optimisation and performance improvements

csfercoci:
If approved I will create another pull request for main 6 branch

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/InFlightMetadataCollectorImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/InFlightMetadataCollectorImpl.java
@@ -1718,7 +1718,7 @@ public class InFlightMetadataCollectorImpl implements InFlightMetadataCollector 
 		Set<FkSecondPass> dependencies = isADependencyOf.get( currentTable );
 
 		// bottom out
-		if ( dependencies == null || dependencies.size() == 0 ) {
+		if ( dependencies == null || dependencies.isEmpty()  ) {
 			return;
 		}
 
@@ -1756,7 +1756,7 @@ public class InFlightMetadataCollectorImpl implements InFlightMetadataCollector 
 					}
 				}
 			}
-			stopProcess = failingSecondPasses.size() == 0 || failingSecondPasses.size() == endOfQueueFkSecondPasses.size();
+			stopProcess = failingSecondPasses.isEmpty()  || failingSecondPasses.size() == endOfQueueFkSecondPasses.size();
 			endOfQueueFkSecondPasses = failingSecondPasses;
 		}
 		if ( endOfQueueFkSecondPasses.size() > 0 ) {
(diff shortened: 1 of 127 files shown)


hibernate-core/src/main/java/org/hibernate/boot/internal/InFlightMetadataCollectorImpl.java
```diff
@@ -1718,7 +1718,7 @@ private void buildRecursiveOrderedFkSecondPasses(
 		Set<FkSecondPass> dependencies = isADependencyOf.get( currentTable );
 
 		// bottom out
-		if ( dependencies == null || dependencies.size() == 0 ) {
+		if ( dependencies == null || dependencies.isEmpty()  ) {
```
Sanne, line 1721: This is a good idea. Could you maybe extract all these specific changes into a separate PR for `main` only?

hibernate-core/src/main/java/org/hibernate/cfg/CollectionPropertyHolder.java
```diff
@@ -319,7 +319,7 @@ public Join addJoin(JoinTable joinTableAnn, boolean noDelayInPkColumnCreation) {
 
 	@Override
 	public String toString() {
-		return super.toString() + "(" + collection.getRole() + ")";
+		return super.toString() + '(' + collection.getRole() + ')';
```
Sanne, line 322: This kind of changes are also fine, and I love you're being thourough with them. But please isolate them into a separate PR and target `main` only.

hibernate-core/src/main/java/org/hibernate/cfg/annotations/reflection/internal/JPAXMLOverriddenAnnotationReader.java
```diff
@@ -221,7 +221,7 @@ private enum PropertyType {
 	private static final Map<Class, String> annotationToXml;
 
 	static {
-		annotationToXml = new HashMap<>();
+		annotationToXml = new HashMap<>(78);
```
Sanne, line 224: When sizing with such precision please add a comment. Also try to consider how an HashMap is implemented: the size is an approximation that helps to compute a certain number of initial buckets, but the exact needed number will depend on the hash distribution;
 - underestimating might save a little memory (with luck) but might require an array resize to be triggered;
 - overestimating might end up having a couple of buckets which aren't strictly necessary

So hitting the exact number isn't very easy nor super important. If you want to really trim memory in such cases, such blocks of code might be better off by refactoring the code to remove the map altogether and think of something smarter to do - but I'm not sure if it's worth spending time on this particular map as there likely are more impactful changes you could make in the same time.

Sanne (changes requested):
I've suggested some changes.
Generally I like it :) But let's split it in smaller bites please, as I can't apply them all.

PR #5176, opened 2022-07-20
HHH-15403 Likely Statement leak on invoking a stored procedure

dreab8:
https://hibernate.atlassian.net/browse/HHH-15403


hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java
```diff
@@ -132,9 +132,6 @@ public void release(ResultSet resultSet, Statement statement) {
 			}
 			else {
 				resultSets.remove( resultSet );
-				if ( resultSets.isEmpty() ) {
-					xref.remove( statement );
```
Sanne, line 136: Is this safe?

dreab8: I think it is not safe to leave it, if the statement is not closed then later on when `ResourceRegistryStandardImpl# releaseResources()` is called, the statement will not be closed... and this is what was causing the leak .

PR #5180, opened 2022-07-21
HHH-15412 Cleanup: remove reflection keeping JDK8 compatibility afloat in AggregatedServiceLoader

Sanne:
@yrodiere could you have a quick look?

 - https://hibernate.atlassian.net/browse/HHH-15412

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/AggregatedServiceLoader.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/AggregatedServiceLoader.java
@@ -6,8 +6,6 @@
  */
 package org.hibernate.boot.registry.classloading.internal;
 
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/AggregatedServiceLoader.java
```diff
@@ -216,22 +141,16 @@ private Collection<S> loadAll() {
...
-		private Stream<? extends Supplier<S>> providerStream(ServiceLoader<S> serviceLoader) {
-			try {
-				return ( (Stream<? extends Supplier<S>>) SERVICE_LOADER_STREAM_METHOD.invoke( serviceLoader ) );
-			}
-			catch (RuntimeException | IllegalAccessException | InvocationTargetException e) {
-				throw new AssertionFailure( "Error calling ServiceLoader.stream()", e );
-			}
+		private Stream<ServiceLoader.Provider<S>> providerStream(ServiceLoader<S> serviceLoader) {
+			return serviceLoader.stream();
 		}
```
yrodiere, line 154: I don't see a reason not to inline this?

yrodiere (approved):
LGTM, thanks.

Just a comment about a method, because it seems odd not to inline it, but it's fine to merge as is.

hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/AggregatedServiceLoader.java
Sanne: right, thanks - clearly forgot

PR #5181, opened 2022-07-21
Jackson and Jakarta Json custom types to not break GraalVM

Sanne:
- https://hibernate.atlassian.net/browse/HHH-15415


hibernate-core/src/main/java/org/hibernate/type/JaxbXmlFormatMapper.java
```diff
@@ -44,7 +44,6 @@
 public class JaxbXmlFormatMapper implements FormatMapper {
```
beikov, line 44: Since you moved other classes to dedicated packages, what about moving this one here as well into e.g. `jakartaxml`?

beikov (approved):
LGTM

hibernate-core/src/main/java/org/hibernate/type/JaxbXmlFormatMapper.java
Sanne: good idea, thanks

PR #5183, opened 2022-07-22
HHH-15391 StackOverflow when applying a load entity graphs to a query

dreab8:
https://hibernate.atlassian.net/browse/HHH-15391

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -56,6 +56,7 @@ import org.hibernate.loader.MultipleBagFetchException;
 import org.hibernate.metamodel.CollectionClassification;
 import org.hibernate.metamodel.MappingMetamodel;
 import org.hibernate.metamodel.mapping.Association;
+import org.hibernate.metamodel.mapping.AssociationKey;
 import org.hibernate.metamodel.mapping.AttributeMapping;
 import org.hibernate.metamodel.mapping.BasicEntityIdentifierMapping;
 import org.hibernate.metamodel.mapping.BasicValuedMapping;
@@ -485,6 +486,8 @@ public abstract class BaseSqmToSqlAstConverter<T extends Statement> extends Base
 	private SqmExpressible<?> adjustedTimestampType; //TODO: remove this once we can get a Type directly from adjustedTimestamp
 	private Expression adjustmentScale;
 	private boolean negativeAdjustment;
+
+	private final Set<AssociationKey> visitedAssociationKeys = new HashSet<>();
 
 	public BaseSqmToSqlAstConverter(
 			SqlAstCreationContext creationContext,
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -7383,4 +7406,50 @@ public void addCteStatement(CteStatement cteStatement) {
...
+
+				TypedQuery<Sample> query = entityManager.createQuery(
+							"select s from Sample s",
+							Sample.class
+					);
+				query.setHint( SpecHints.HINT_SPEC_LOAD_GRAPH, entityGraph );
+
+			 */
+		return creationContext.getMaximumFetchDepth() == null
+				&& ( entityGraphTraversalState != null || getLoadQueryInfluencers().hasEnabledFetchProfiles() );
```
beikov, line 7452: Isn't this also a problem if there is no entity graph or fetch profile? Maybe I'm missing something, but it seems to me that this should only depend on the nullness of maximum fetch depth i.e. use

```java
return creationContext.getMaximumFetchDepth() == null;
```

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -6914,9 +6917,12 @@ public void addFetch(List<Fetch> fetches, FetchParent fetchParent, Fetchable fet
 							fetchable,
 							isKeyFetchable
 					);
+
 					fetchTiming = traversalResult.getFetchTiming();
 					joined = traversalResult.isJoined();
-					explicitFetch = true;
+					if ( !canFetchCreateALoop( maxDepth, fetchable ) ) {
+						explicitFetch = true;
+					}
```
beikov, line 6925: I'd prefer something like this instead:

```java
explicitFetch = isExplicitFetch( maxDepth, fetchable );
```

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -7067,6 +7073,23 @@ public List<Fetch> visitFetches(FetchParent fetchParent) {
...
+			if ( fetchable instanceof ToOneAttributeMapping ) {
+				return this.isAssociationKeyVisited(
+						( (ToOneAttributeMapping) fetchable ).getForeignKeyDescriptor().getAssociationKey()
+				);
+			}
+			else if ( fetchable instanceof PluralAttributeMapping ) {
+				return this.isAssociationKeyVisited(
+						( (PluralAttributeMapping) fetchable ).getKeyDescriptor().getAssociationKey()
+				);
+			}
```
beikov, line 7087: Can you please add a comment here about why the `explicitFetch` should be set to false when "a fetch can create a loop"?

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
dreab8: no the problem does not occur when no load entity graph is specified,

beikov: But why not? If we have a simple eager association, shouldn't it run into the same issue?

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
dreab8: we just do not change the `explicitFetch` value, when `explicitFetch` is true we do not do the circulatiry check

dreab8: so the method check if the fetch can create a look, in such a case we do not want to set `explicitFetch` to true in case it was not explicitly fetched though an explicit hql fetch

beikov (approved):

PR #5185, opened 2022-07-25
HHH-15417 UUID with AttributeConverter, merge throws ClassCastException

dreab8:
https://hibernate.atlassian.net/browse/HHH-15417

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/converter/AttributeConverterTypeAdapter.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/converter/AttributeConverterTypeAdapter.java
@@ -9,6 +9,7 @@ package org.hibernate.type.descriptor.converter;
 import java.sql.CallableStatement;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
+import java.util.Map;
 
 import jakarta.persistence.AttributeConverter;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/type/descriptor/converter/AttributeConverterTypeAdapter.java
```diff
@@ -143,4 +145,32 @@ private Object getConvertedValue(AttributeConverter<T, Object> converter, T valu
...
+			SharedSessionContractImplementor session,
+			Object owner,
+			Map<Object, Object> copyCache) {
+		if ( original == null && target == null ) {
+			return null;
+		}
+
+		return domainJtd.getReplacement( (T) original, (T) target, session );
+	}
+
```
beikov, line 175: Shouldn't we instead call the super constructor with `super( std, domainJtd );` instead of with `super( std, (JavaType<T>) relationalJtd );`?

beikov (commented):
Looks good to me if there is no other way to do this, but maybe it's possible to change just the constructor of `AttributeConverterTypeAdapter` where the super constructor call seems a bit suspicious.

beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/5195

PR #5191, opened 2022-08-01
HHH-15425 org.hibernate.QueryException: could not resolve property is thrown when Hibernate criteria tries to select the id of an association annotated with @NotFound

dreab8:
https://hibernate.atlassian.net/browse/HHH-15425


beikov (approved):
LGTM

PR #5193, opened 2022-08-02
HHH-15423 @Any discriminated association fails to be implicitly fetched due to UnsupportedOperationException

dreab8:
https://hibernate.atlassian.net/browse/HHH-15423

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/results/implicit/ImplicitFetchBuilderDiscriminatedAssociation.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/results/implicit/ImplicitFetchBuilderDiscriminatedAssociation.java
@@ -8,12 +8,14 @@ package org.hibernate.query.results.implicit;
 
 import java.util.function.BiFunction;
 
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/query/results/implicit/ImplicitFetchBuilderDiscriminatedAssociation.java
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	private ImplicitFetchBuilderDiscriminatedAssociation(ImplicitFetchBuilderDiscriminatedAssociation original) {
+		this.fetchPath = original.fetchPath;
+		this.fetchable = original.fetchable;
+	}
+
+	@Override
+	public FetchBuilder cacheKeyInstance() {
+		return new ImplicitFetchBuilderDiscriminatedAssociation( this );
```
beikov, line 45: You can simply return `this` here, as the type is immutable anyway.

beikov (approved):
LGTM

PR #5194, opened 2022-08-02
HHH-15418  Composite primary key with NativeQuery & alias not working

dreab8:
https://hibernate.atlassian.net/browse/HHH-15418

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
@@ -244,6 +244,7 @@ import org.hibernate.sql.results.graph.DomainResult;
 import org.hibernate.sql.results.graph.DomainResultCreationState;
 import org.hibernate.sql.results.graph.Fetch;
 import org.hibernate.sql.results.graph.Fetchable;
+import org.hibernate.sql.results.graph.embeddable.EmbeddableResultGraphNode;
 import org.hibernate.sql.results.graph.entity.internal.EntityResultImpl;
 import org.hibernate.sql.results.internal.SqlSelectionImpl;
 import org.hibernate.stat.spi.StatisticsImplementor;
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -1867,10 +1869,15 @@ public String selectFragment(String alias, String suffix) {
 								FetchTiming fetchTiming = fetchable.getMappedFetchOptions().getTiming();
 								final boolean selectable;
 								if ( fetchable instanceof AttributeMapping ) {
-									final int propertyNumber = ( (AttributeMapping) fetchable ).getStateArrayPosition();
-									final int tableNumber = getSubclassPropertyTableNumber( propertyNumber );
-									selectable = !isSubclassTableSequentialSelect( tableNumber )
-											&& propertySelectable[propertyNumber];
+									if ( fetchParent instanceof EmbeddableFetchImpl && ( (EmbeddableFetchImpl) fetchParent ).getReferencedMappingContainer() instanceof EntityIdentifierMapping ) {
```
beikov, line 1872: How about this instead?

```suggestion
									if ( fetchParent instanceof EmbeddableResultGraphNode && ( (EmbeddableResultGraphNode) fetchParent ).getReferencedMappingContainer() instanceof EntityIdentifierMapping ) {
```

beikov (approved):
LGTM

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
dreab8: @beikov,  I changed it to `if ( fetchParent instanceof EmbeddableResultGraphNode && ( (EmbeddableResultGraphNode) fetchParent ).getReferencedMappingContainer() == getIdentifierMapping() )`, what do you think?

beikov: Even better :)

PR #5200, opened 2022-08-08
HHH-15442 Allow for setParameter usage with a queries for a specific type in an Any-discriminated relationship 

dreab8:
https://hibernate.atlassian.net/browse/HHH-15442


hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/AnyDiscriminatorSqmPath.java
```diff
@@ -67,4 +66,15 @@ public SqmTreatedPath treatAs(Class treatJavaType) throws PathException {
 	public SqmTreatedPath treatAs(EntityDomainType treatTarget) throws PathException {
 		throw new UnsupportedMappingException( "Cannot apply TREAT operator to discriminator path" );
 	}
+
+	@Override
+	public SqmPath<T> getLhs() {
+		return (SqmPath<T>) super.getLhs().getLhs();
+	}
```
beikov, line 73: This feels a bit wrong to me, as the lhs of a `type(root.anyAttribute)` should be the path for `root.anyAttribute`, but this seems to return `root` instead.

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/AnyDiscriminatorSqmPathSource.java
```diff
@@ -19,18 +20,17 @@
...
+			AnyDiscriminatorDomainTypeImpl domainType,
 			BindableType jpaBindableType) {
 		super( localPathName, domainType, jpaBindableType );
 	}
 
 	@Override
 	public SqmPath<D> createSqmPath(SqmPath<?> lhs, SqmPathSource<?> intermediatePathSource) {
 		final NavigablePath navigablePath;
-		if ( intermediatePathSource == null ) {
+		if ( intermediatePathSource == null || intermediatePathSource.getPathName().equals( CollectionPart.Nature.ELEMENT.getName() ) ) {
```
beikov, line 33: I feel like we might be missing a table group registration if this is really required. Can you explain what this change fixes?

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/AnyMappingDomainTypeImpl.java
```diff
@@ -19,10 +19,15 @@
 public class AnyMappingDomainTypeImpl<T> implements AnyMappingDomainType<T> {
 	private final AnyType anyType;
 	private final JavaType<T> baseJtd;
+	private final AnyDiscriminatorDomainTypeImpl<?> anyDiscriminatorType;
 
 	public AnyMappingDomainTypeImpl(AnyType anyType, JavaType<T> baseJtd) {
 		this.anyType = anyType;
 		this.baseJtd = baseJtd;
+		final MetaType discriminatorType = (MetaType) anyType.getDiscriminatorType();
+		anyDiscriminatorType = new AnyDiscriminatorDomainTypeImpl<>(
```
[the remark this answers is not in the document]
dreab8: `AnyDiscriminatorDomainTypeImpl` does not implement `BasicType`.

By the way not sure here how to use the `ConvertedBasicTypeImpl`

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/AnyDiscriminatorSqmPath.java
dreab8: I have to dig into this

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/AnyDiscriminatorSqmPathSource.java
dreab8: I have to dig into this too

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/AnyMappingDomainTypeImpl.java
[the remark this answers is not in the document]
beikov: Oh ok, `AnyDiscriminatorDomainTypeImpl` looked so similar to `DiscriminatorType` that I thought you copied that. It was just an idea, not sure if this actually works, but it would be instantiated like this:

```
anyDiscriminatorType = new ConvertedBasicTypeImpl<>(
  null, // no name
  discriminatorType.getBaseType().getJdbcType(),
  new AnyDiscriminatorConverter<>( discriminatorType )
);
```

where `AnyDiscriminatorConverter` would roughly look like this:

```
public class AnyDiscriminatorConverter<T> implements BasicValueConverter<T, Object> {
	private final MetaType modelPart;

	public AnyDiscriminatorDomainTypeImpl(MetaType modelPart) {
		this.modelPart = modelPart;
	}

	@Override
	public T toDomainValue(Object discriminatorValue) {
		if ( discriminatorValue == null ) {
			return null;
		}
		final String entityName = modelPart.getDiscriminatorValuesToEntityNameMap().get( discriminatorValue );
		if ( entityName == null ) {
			throw new HibernateException( "Unable to resolve discriminator value [" + discriminatorValue + "] to entity name" );
		}
		final EntityPersister entityPersister = persister.getFactory()
				.getRuntimeMetamodels()
				.getMappingMetamodel()
				.getEntityDescriptor( entityName );
		//noinspection unchecked
		return entityPersister.getRepresentationStrategy().getMode() == RepresentationMode.POJO
				? (T) entityPersister.getJavaType().getJavaTypeClass()
				: (T) entityName;
	}

	@Override
	public Object toRelationalValue(T domainForm) {
		if ( domainForm == null ) {
			return null;
		}

		if ( domainForm instanceof Class ) {
			return modelPart.getEntityNameToDiscriminatorValueMap().get( ( (Class) domainForm ).getName() );
		}
		else {
			return modelPart.getEntityNameToDiscriminatorValueMap().get( (String) domainForm );
		}
	}

	@Override
	public JavaType<T> getDomainJavaType() {
		return (JavaType<T>) ClassJavaType.INSTANCE;
	}

	@Override
	public JavaType<Object> getRelationalJavaType() {
		return underlyingType.getExpressibleJavaType();
	}
}
```

I kept the `T` type parameter for the domain type, but from what I understand, it always has to be a `Class`.

beikov (approved):

PR #5201, opened 2022-08-09
Improve ListresultConsumer duplication check method performance

dreab8:
just an attempt to address the potential perf issue mentioned in the https://github.com/hibernate/hibernate-orm/pull/4924 discussion.

https://hibernate.atlassian.net/browse/HHH-15479

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
@@ -7,14 +7,13 @@
 package org.hibernate.sql.results.spi;
 
 import java.util.ArrayList;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Locale;
-import java.util.Set;
 
 import org.hibernate.HibernateException;
 import org.hibernate.engine.spi.PersistenceContext;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
-import org.hibernate.internal.util.collections.IdentitySet;
 import org.hibernate.query.ResultListTransformer;
 import org.hibernate.query.spi.QueryOptions;
 import org.hibernate.sql.results.internal.RowProcessingStateStandardImpl;
@@ -115,13 +114,21 @@ public class ListResultsConsumer<R> implements ResultsConsumer<List<R>, R> {
 			this.resultJavaType = resultJavaType;
 		}
 
-		public boolean contains(R result) {
+		private boolean contains(R result) {
 			for ( int i = 0; i < results.size(); i++ ) {
 				if ( resultJavaType.areEqual( results.get( i ), result ) ) {
 					return true;
 				}
 			}
 			return false;
+		}
+
+		public boolean addUnique(R result) {
+			if ( contains( result ) ) {
+				return false;
+			}
+			results.add( result );
+			return true;
 		}
 
 		public void add(R result) {
@@ -134,20 +141,22 @@ public class ListResultsConsumer<R> implements ResultsConsumer<List<R>, R> {
 	}
 
 	private static class EntityResult<R> extends Results<R> {
-		private final Set<R> added = new IdentitySet<>();
+		private static final Object DUMP_VALUE = new Object();
+
+		private final IdentityHashMap<R, Object> added = new IdentityHashMap<>();
 
 		public EntityResult(JavaType resultJavaType) {
 			super( resultJavaType );
 		}
 
-		public boolean contains(R result) {
-			return added.contains( result );
+		public boolean addUnique(R result) {
+			if ( added.put( result, DUMP_VALUE ) == null ) {
+				super.add( result );
+				return true;
+			}
+			return false;
 		}
 
-		public void add(R result) {
-			super.add( result );
-			added.add( result );
-		}
 	}
 
 	@Override
@@ -263,19 +272,14 @@ public class ListResultsConsumer<R> implements ResultsConsumer<List<R>, R> {
 			Results<R> results,
 			RowProcessingStateStandardImpl rowProcessingState,
 			boolean throwException) {
-		if ( results.contains( result ) ) {
-			if ( throwException && !rowProcessingState.hasCollectionInitializers ) {
-				throw new HibernateException(
-						String.format(
-								Locale.ROOT,
-								"Duplicate row was found and `%s` was specified",
-								UniqueSemantic.ASSERT
-						)
-				);
-			}
-		}
-		else {
-			results.add( result );
+		if ( !results.addUnique( result ) && throwException && !rowProcessingState.hasCollectionInitializers ) {
+			throw new HibernateException(
+					String.format(
+							Locale.ROOT,
+							"Duplicate row was found and `%s` was specified",
+							UniqueSemantic.ASSERT
+					)
+			);
 		}
 	}
 
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
@@ -9,10 +9,12 @@ package org.hibernate.sql.results.spi;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.Set;
 
 import org.hibernate.HibernateException;
 import org.hibernate.engine.spi.PersistenceContext;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.internal.util.collections.IdentitySet;
 import org.hibernate.query.ResultListTransformer;
 import org.hibernate.query.spi.QueryOptions;
 import org.hibernate.sql.results.internal.RowProcessingStateStandardImpl;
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
```diff
@@ -105,6 +107,49 @@ else if ( uniqueSemantic == UniqueSemantic.ASSERT ) {
...
+		}
+
+		public boolean contains(R result) {
+			for ( int i = 0; i < results.size(); i++ ) {
+				if ( resultJavaType.areEqual( results.get( i ), result ) ) {
+					return true;
+				}
+			}
+			return false;
+		}
```
beikov, line 125: I think I'd prefer having a `boolean addUnique(R result)` instead which returns whether the object was added or not. Enables implementing it as `if ( added.add( result ) ) { getResults().add( result ); }` instead which is a bit faster.

hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
```diff
@@ -131,17 +182,25 @@ public List<R> consume(
...
+			else {
+				results = new Results<>( domainResultJavaType );
+			}
+
 			while ( rowProcessingState.next() ) {
 				final R row = rowReader.readRow( rowProcessingState, processingOptions );
-				resultHandlerToUse.handle( row, domainResultJavaType, results, rowProcessingState );
+				resultHandlerToUse.handle( row, results, rowProcessingState );
 				rowProcessingState.finishRowProcessing();
 			}
```
beikov, line 205: Overall, I think I would prefer getting rid of `resultHandler` as I believe the megamorphic call (3 implementors) to `ResultHandler#handle` may be another bottleneck when processing big result lists, but that's just a thought. I understand that the following introduces a bit code duplication, but maybe you can try that and see if it improves the situation with the 1M rows even further.

```
		final UniqueSemantic uniqueSemantic = this.uniqueSemantic == UniqueSemantic.ALLOW && isEnityResultType ? UniqueSemantic.FILTER : this.uniqueSemantic;
		switch ( uniqueSemantic ) {
			case UniqueSemantic.FILTER: {
				while ( rowProcessingState.next() ) {
					final R row = rowReader.readRow( rowProcessingState, processingOptions );
					results.addUnique( result );
					rowProcessingState.finishRowProcessing();
				}
				break;
			}
			case UniqueSemantic.ASSERT: {
				while ( rowProcessingState.next() ) {
					final R row = rowReader.readRow( rowProcessingState, processingOptions );
					if ( !results.addUnique( result ) && !rowProcessingState.hasCollectionInitializers ) {
						throw new HibernateException(
							String.format(
									Locale.ROOT,
									"Duplicate row was found and `%s` was specified",
									UniqueSemantic.ASSERT
							)
						);
					}
					rowProcessingState.finishRowProcessing();
				}
				break;
			default: {
				while ( rowProcessingState.next() ) {
					final R row = rowReader.readRow( rowProcessingState, processingOptions );
					results.add( result );
					rowProcessingState.finishRowProcessing();
				}
				break;
			}
		}
```

dreab8:
my concern is the impact of the new `IdentityHashMap` for small/medium results where the original solution had not real perf issues

hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
```diff
@@ -131,17 +182,25 @@ public List<R> consume(
...
+			boolean isEnityResultType = domainResultJavaType instanceof EntityJavaType;
+			if ( uniqueSemantic == UniqueSemantic.ALLOW && isEnityResultType ) {
 				resultHandlerToUse = ListResultsConsumer::deDuplicationHandling;
 			}
 			else {
 				resultHandlerToUse = this.resultHandler;
 			}
 
+			final Results<R> results;
+			if ( isEnityResultType ) {
```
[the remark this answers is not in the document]
dreab8: we want to apply the check optimisation even when duplication is not allowed

beikov:
>my concern is the impact of the new `IdentityHashMap` for small/medium results where the original solution had not real perf issues

We could try to allocate the `IdentityHashMap` only when exceeding a certain row count threshold, but not sure if that makes things a lot better.

I think, if we pass through a maxResults estimate into `org.hibernate.sql.results.spi.ResultsConsumer#consume`, we could reduce the possible negative effects to only HQL/Criteria queries that select entities where we have no good estimate. In the end, I think we simply need guidance from the user. The user should be able to provide a hint that says "don't deduplicate", in which case we can avoid the `IdentityHashMap`.

hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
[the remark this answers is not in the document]
beikov: My point is, that there is no need to use a `EntityResult` when the `uniqueSemantic` has a value, which will always result in the use of `Result#add` instead of `addUnique`. So IMO the only cases when it makes sense to use `EntityResult` is when the `uniqueSemantic` is `ALLOW` or `FILTER` (both will use `Result#addUnique`) and the result type is an entity. `ASSERT` also uses `Result#addUnique`, but is only used for `SingleIdLoadPlan` when the list is guaranteed to have only a single element. Other `uniqueSemantic` use `Result#add`, so a `EntityResult` would only mean overhead for these scenarios.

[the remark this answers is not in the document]
dreab8: it seems strange the use of ASSERT for  SingleIdLoadPlan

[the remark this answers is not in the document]
dreab8: i would have chosen NEVER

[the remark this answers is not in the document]
beikov: You were the one who chose `ASSERT` originally ;)
I think I would also prefer `NEVER` and do the "assertion" in the `SingleIdLoadPlan` code instead, but not sure if we can remove `ASSERT` now (since it will be unused then) as this is in SPI.

Sanne:
IMO conceptually we'd need a https://en.wikipedia.org/wiki/Bloom_filter
Not keen to add a dependency though - could we check how close we get approximating it via an identity hashmap ?

beikov:
>IMO conceptually we'd need a https://en.wikipedia.org/wiki/Bloom_filter
>Not keen to add a dependency though - could we check how close we get approximating it via an identity hashmap ?

Yeah, either add a dependency or implement/integrate this ourselves in hibernate-core. It's not trivial to implement this, but I think it would be worth it to add a custom implementation. Standard Bloom filters AFAIU need a fixed size, but since we keep the data around, we could grow the bit set on-demand similar to what a hash table does, but use a lot less memory (64 times memory savings i.e. for 1M elements we'd use ~2KB instead of ~1MB). Growing a bit set also consumes a lot less memory, though rehashing is still necessary.

hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
Sanne: yes, if we could avoid the megamorphic calls here that would be great.

sebersole (commented):
Looks reasonable

hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
dreab8: Thanks! done :)

hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
```diff
@@ -116,33 +158,52 @@ public List<R> consume(
...
+			if ( this.uniqueSemantic == UniqueSemantic.FILTER || this.uniqueSemantic == UniqueSemantic.ALLOW && isEnityResultType ) {
+				while ( rowProcessingState.next() ) {
+					results.addUnique( rowReader.readRow( rowProcessingState, processingOptions ) );
+					rowProcessingState.finishRowProcessing();
+				}
+			}
+			else if ( this.uniqueSemantic == UniqueSemantic.ASSERT ) {
+				while ( rowProcessingState.next() ) {
+					if ( !results.addUnique( rowReader.readRow( rowProcessingState, processingOptions ) )
+							&& !rowProcessingState.hasCollectionInitializers ) {
```
beikov, line 190: When `uniqueSemantics == ASSERT && rowProcessingState.hasCollectionInitializers` the effect is the same as `FILTER` in the above if-block. IMO we should add that condition to the upper if, like:

```
if ( uniqueSemantic == UniqueSemantic.FILTER || uniqueSemantics == ASSERT && rowProcessingState.hasCollectionInitializers || uniqueSemantic == UniqueSemantic.ALLOW && isEnityResultType ) {
```

and remove the check for `&& !rowProcessingState.hasCollectionInitializers` here.

hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
```diff
@@ -89,19 +90,60 @@ public enum UniqueSemantic {
...
+					return true;
+				}
+			}
+			return false;
+		}
+
+		public boolean addUnique(R result) {
+			if ( contains( result ) ) {
+				return false;
+			}
```
beikov, line 118: Small nitpick, I'd prefer to inline the `contains` method.

```suggestion
			for ( int i = 0; i < results.size(); i++ ) {
				if ( resultJavaType.areEqual( results.get( i ), result ) ) {
					return false;
				}
			}
```

beikov (approved):
LGTM, just some small nitpicks. Feel free to ignore them.

PR #5206, opened 2022-08-12
HHH-15446 - allow Dialect version to be specified separately from the dialect class name

jrenaat:
See https://hibernate.atlassian.net/browse/HHH-15446

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
@@ -11,20 +11,18 @@ import java.sql.DatabaseMetaData;
 import java.sql.SQLException;
 import java.util.Map;
 
-import org.hibernate.HibernateException;
 import org.hibernate.boot.registry.StandardServiceInitiator;
 import org.hibernate.cfg.AvailableSettings;
-import org.hibernate.dialect.Dialect;
 import org.hibernate.engine.jdbc.connections.spi.ConnectionProvider;
 import org.hibernate.engine.jdbc.connections.spi.JdbcConnectionAccess;
 import org.hibernate.engine.jdbc.connections.spi.MultiTenantConnectionProvider;
-import org.hibernate.engine.jdbc.dialect.spi.DatabaseMetaDataDialectResolutionInfoAdapter;
 import org.hibernate.engine.jdbc.dialect.spi.DialectFactory;
 import org.hibernate.engine.jdbc.dialect.spi.DialectResolutionInfo;
 import org.hibernate.engine.jdbc.env.spi.JdbcEnvironment;
 import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.internal.log.DeprecationLogger;
 import org.hibernate.internal.util.NullnessHelper;
+import org.hibernate.internal.util.StringHelper;
 import org.hibernate.internal.util.config.ConfigurationHelper;
 import org.hibernate.service.spi.ServiceRegistryImplementor;
 
@@ -63,8 +61,8 @@ public class JdbcEnvironmentInitiator implements StandardServiceInitiator<JdbcEn
 				true
 		);
 
-		final Object dbName = NullnessHelper.coalesceSuppliedValues(
-				() -> configurationValues.get( AvailableSettings.JAKARTA_HBM2DDL_DB_NAME ),
+		String explicitDatabaseName = NullnessHelper.coalesceSuppliedValues(
+				() -> (String) configurationValues.get( AvailableSettings.JAKARTA_HBM2DDL_DB_NAME ),
 				() -> {
 					final Object value = configurationValues.get( AvailableSettings.DIALECT_DB_NAME );
 					if ( value != null ) {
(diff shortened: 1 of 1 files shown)


jrenaat:
Still need to add a test to verify the dialect returned corresponds to what is expected

jrenaat:
@beikov It looks a bit different from when we paired, because i didn't want to repeat that stack of nested try-catch in the getMetadataResolutionInfo method we created, when creating the JdbcEnvironment  (HQLTest showed that we need to pass in the dbmd to the JdbcEnvironment when it is allowed, otherwise it fails due to scrollable result sets not being enabled)

beikov (commented):
Looks better now, but some small things still need to be adapted.

hibernate-core/src/main/java/org/hibernate/dialect/SybaseDialect.java
```diff
@@ -60,7 +60,7 @@
  */
 public class SybaseDialect extends AbstractTransactSQLDialect {
 
-	protected boolean jtdsDriver;
+	protected boolean jtdsDriver = true;
```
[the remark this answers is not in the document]
jrenaat: See https://github.com/hibernate/hibernate-orm/pull/5206#discussion_r963133652

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
```diff
@@ -239,6 +224,27 @@ else if ( useJdbcMetadata ) {
...
+								() -> new DialectResolutionInfoImpl(
+										null,
+										explicitDatabaseName,
+										explicitDatabaseVersion != null ? explicitDatabaseVersion : "0",
+										explicitDatabaseMajorVersion != null ? explicitDatabaseMajorVersion : 0,
+										explicitDatabaseMinorVersion != null ? explicitDatabaseMinorVersion : 0,
+										null,
+										0,
+										0,
+										null
```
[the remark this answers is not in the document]
jrenaat: Not sure I understand your intention, what explicit configurations for driver etc. should i use, as compared to how it was until now?

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
```diff
@@ -203,23 +159,52 @@ else if ( useJdbcMetadata ) {
...
+									() -> new DialectResolutionInfoImpl(
+											dbmd,
+											databaseName,
+											databaseVersion,
+											databaseMajorVersion,
+											databaseMinorVersion,
+											null,
+											0,
+											0,
+											null
```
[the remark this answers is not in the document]
jrenaat: See https://github.com/hibernate/hibernate-orm/pull/5206#discussion_r963136170

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
```diff
@@ -239,6 +224,27 @@ else if ( useJdbcMetadata ) {
 				log.unableToObtainConnectionToQueryMetadata( e );
 			}
 		}
+		else {
+			if (someExplicitVersionSet) {
```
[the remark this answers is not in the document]
jrenaat: I'd have to chew on this a bit; I think it's risky to introduce the condition based on `AvailableSettings.DIALECT` because until now this was only evaluated in the`DialectFactory`, not in the initiator (which is something that I still don't understand tbh, it makes the logic hard to follow and messy), so imo this could well have some unwanted side-effects.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
```diff
@@ -73,105 +71,63 @@ public JdbcEnvironment initiateService(Map<String, Object> configurationValues,
 								AvailableSettings.JAKARTA_HBM2DDL_DB_NAME
 						);
 					}
-					return value;
+					return (String) value;
+				}
+		);
+
+		// todo check this after major/minor version, and set only if none of those are set?
```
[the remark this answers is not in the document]
jrenaat: Assuming you mean we move this evaluation below the major/minor ones, I think you mean as first argument, otherwise the additional supplier will not be considered?

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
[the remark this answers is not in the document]
beikov: You can split up the predicates to make it more readable. Not sure about side effects. I mean this only is an issue if people disable JDBC metadata usage, and explicitly configure database information.
If people configure a version but don't configure a database name or dialect, then the current logic would just ignore the configured version. The change I propose would respect the explicitly configured version when a dialect or database name is also configured, so IMO this change is safe. Please explain what you think could cause unwanted side effects.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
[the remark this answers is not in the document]
beikov: `org.hibernate.cfg.AvailableSettings#JAKARTA_JDBC_DRIVER` etc. should be read and passed in here

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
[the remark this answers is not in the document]
beikov: Yes, move below major/minor ones and write this:

```java
String explicitDatabaseVersion = NullnessHelper.coalesceSuppliedValues(
		() -> (String) configurationValues.get( AvailableSettings.JAKARTA_HBM2DDL_DB_VERSION ),
		() -> {
			final Object value = configurationValues.get( AvailableSettings.DIALECT_DB_VERSION );
			if ( value != null ) {
				DeprecationLogger.DEPRECATION_LOGGER.deprecatedSetting(
						AvailableSettings.DIALECT_DB_VERSION,
						AvailableSettings.JAKARTA_HBM2DDL_DB_VERSION
				);
			}
			return (String) value;
		},
		() -> {
			if ( explicitDatabaseMajorVersion != null ) {
				return explicitDatabaseMinorVersion == null ? explicitDatabaseMajorVersion.toString() : (explicitDatabaseMajorVersion.toString() + "." explicitDatabaseMinorVersion.toString());
			}
			return null;
		}
);
```

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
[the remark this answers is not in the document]
beikov: I had wrongly assumed that we have `AvailableSettings` for configuring explicit driver metadata like driver name and version, which we otherwise determine through `DatabaseMetaData` when `useJdbcMetadata` is true. Since we don't have that, you can ignore this comment.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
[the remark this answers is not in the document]
beikov: Also, please add the following code after that code block I posted:

```java
if ( explicitDatabaseMajorVersion == null && explicitDatabaseMinorVersion == null && explicitDatabaseVersion != null ) {
	final String[] parts = explicitDatabaseVersion.split( "\\." );
	try {
		final int potentialMajor = Integer.parseInt( parts[0] );
		if ( parts.length > 1 ) {
			explicitDatabaseMinorVersion = Integer.parseInt( parts[1] );
		}
		explicitDatabaseMajorVersion = potentialMajor;
	}
	catch (NumberFormatException e) {
		// Ignore
	}
}
```

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
```diff
@@ -73,105 +71,84 @@ public JdbcEnvironment initiateService(Map<String, Object> configurationValues,
...
+		);
+
+		Integer explicitDatabaseMajorVersion;
+		Integer explicitDatabaseMinorVersion;
+		if ( _explicitDatabaseMajorVersion == null && explicitDatabaseVersion != null ) {
+			final String[] parts = explicitDatabaseVersion.split( "\\." );
+			final int potentialMajor = Integer.parseInt( parts[0] );
+			if ( parts.length > 1 ) {
+				explicitDatabaseMinorVersion = _explicitDatabaseMinorVersion == null ? Integer.parseInt( parts[1] ) : _explicitDatabaseMinorVersion;
+			}
```
beikov, line 140: You forgot the most important parts, which is the exception handling. What was wrong with the code I posted before? We never use this kind of variable naming with prefixing by `_`, so I'd rather not do this here. It's fine to have these variables (`explicitDatabaseMajorVersion` and `explicitDatabaseMinorVersion`) be mutable here.

hibernate-core/src/main/java/org/hibernate/dialect/SybaseDialect.java
```diff
@@ -73,13 +73,10 @@ public SybaseDialect() {
 
 	public SybaseDialect(DatabaseVersion version) {
 		super(version);
-		jtdsDriver = true;
 	}
 
 	public SybaseDialect(DialectResolutionInfo info) {
 		super(info);
-		jtdsDriver = info.getDriverName() != null
-				&& info.getDriverName().contains( "jTDS" );
```
jrenaat, line 82: This change is problematic; if left in, a couple of hundred tests fail on Sybase due to `net.sourceforge.jtds.jdbc.ProtocolException`.
Any suggestions?

beikov: Revert the change?

jrenaat: hahah, yes i did that and all works again, but i meant any suggestion as to a type of fix that doesn't involve deaulting to jtds always or finding the cause (i could go and debug the driver, if you think it makes sense)

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
jrenaat: i had to adapt it because the lambda's complain about variables not being final, and not being properly initialized and such ... and one of the causes of that was the try/catch. My change may not have been the best way to resolve these issues though, if you want to have a go it putting your original code block in there and resolving those issues more adequately i would certainly appreciate and learn from it.
For the record, i didn't change your code just for the sake of changing it

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
```diff
@@ -239,6 +245,30 @@ else if ( useJdbcMetadata ) {
...
+		else {
+			if (
+				(StringHelper.isNotEmpty( explicitDatabaseVersion ) || explicitDatabaseMajorVersion != null || explicitDatabaseMinorVersion != null)
+				&& ( StringHelper.isNotEmpty( explicitDatabaseName ) || StringHelper.isNotEmpty( (String) configurationValues.get( AvailableSettings.DIALECT ) ) )
+			) {
+				return new JdbcEnvironmentImpl(
+						registry,
+						dialectFactory.buildDialect(
+								configurationValues,
+								() -> new DialectResolutionInfoImpl(
```
beikov, line 257: Create a new variable for `DialectResolutionInfoImpl` to avoid the lambda issue: https://github.com/hibernate/hibernate-orm/pull/5206#discussion_r966116210

hibernate-core/src/main/java/org/hibernate/dialect/SybaseDialect.java
jrenaat: This:
```
public SybaseDialect(DialectResolutionInfo info) {
		super(info);
		jtdsDriver = info.getDriverName() != null
				&& info.getDriverName().contains( "jTDS" );
}
```
causes CI failures on Sybase for this PR, not sure why. Remove it and no failures.
I'm ok with leaving the jtds compatible mode, just pointing out that CI complains.

jrenaat: Solved that one now ... I think / hope

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
```diff
@@ -73,105 +71,82 @@ public JdbcEnvironment initiateService(Map<String, Object> configurationValues,
...
-								configurationValues
+								AvailableSettings.JAKARTA_HBM2DDL_DB_MINOR_VERSION
+						);
+					}
+					return value;
+				}
+		);
+
+		Integer _explicitDatabaseMajorVersion = explicitDatabaseMajorVersion;
+		Integer _explicitDatabaseMinorVersion = explicitDatabaseMinorVersion;
```
beikov, line 113:
```suggestion
		Integer configuredDatabaseMajorVersion = explicitDatabaseMajorVersion;
		Integer configuredDatabaseMinorVersion = explicitDatabaseMinorVersion;
```

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
```diff
@@ -239,11 +244,40 @@ else if ( useJdbcMetadata ) {
 				log.unableToObtainConnectionToQueryMetadata( e );
 			}
 		}
+		else {
+			if (
+				(StringHelper.isNotEmpty( explicitDatabaseVersion ) || explicitDatabaseMajorVersion != null || explicitDatabaseMinorVersion != null)
+				&& ( StringHelper.isNotEmpty( explicitDatabaseName ) || isNotEmpty( configurationValues.get( AvailableSettings.DIALECT ) ) )
+			) {
+				DialectResolutionInfo dri = new DialectResolutionInfoImpl(
```
beikov, line 252:
```suggestion
				final DialectResolutionInfo dialectResolutionInfo = new DialectResolutionInfoImpl(
```

hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java
```diff
@@ -203,23 +178,53 @@ else if ( useJdbcMetadata ) {
...
+						databaseMajorVersion = explicitDatabaseMajorVersion;
+					}
+					if ( explicitDatabaseMinorVersion == null ) {
+						databaseMinorVersion = dbmd.getDatabaseMinorVersion();
+					}
+					else {
+						databaseMinorVersion = explicitDatabaseMinorVersion;
+					}
+
+					DialectResolutionInfo dri = new DialectResolutionInfoImpl(
```
beikov, line 211:
```suggestion
					final DialectResolutionInfo dialectResolutionInfo = new DialectResolutionInfoImpl(
```

beikov (commented):
One last naming change and then this should be good to go

PR #5210, opened 2022-08-16
Fix dead link in documentation section "Java API for HQL and JPQL"

yrodiere:
See the link "next chapter" at the beginning of this section:

https://docs.jboss.org/hibernate/orm/current/userguide/html_single/Hibernate_User_Guide.html#hql


beikov:
Thanks

PR #5211, opened 2022-08-16
HHH-15440 @OneToOne and @OptimisticLock(excluded = true) not working correctly

dreab8:
https://hibernate.atlassian.net/browse/HHH-15440


beikov (approved):
LGTM

PR #5213, opened 2022-08-16
HHH-15449 @ManyToOne associations not loaded correctly with default EAGER and batch fetch property set when using TypedQuery.resultStream

dreab8:
https://hibernate.atlassian.net/browse/HHH-15449

When `hibernate.default_batch_fetch_size` is set a `BatchEntitySelectFetchInitializer` is used that initialise the instanced only when the `endLoading` method is called and in case of a Steam this happens only when the stream is closed.

In case of stream a `FORWARD_ONLY` scroll mode is used, so one solution for forward only scroll mode can be to close the stream when the last element is reached.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/spi/ForwardOnlyScrollableResultConsumer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/spi/ForwardOnlyScrollableResultConsumer.java
@@ -1,0 +1,88 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.sql.results.spi;
+
+import java.util.List;
+
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.internal.FetchingScrollableResultsImpl;
+import org.hibernate.internal.ScrollableResultsImpl;
+import org.hibernate.query.spi.ScrollableResultsImplementor;
+import org.hibernate.sql.results.graph.DomainResult;
+import org.hibernate.sql.results.graph.Fetch;
+import org.hibernate.sql.results.graph.collection.internal.EagerCollectionFetch;
+import org.hibernate.sql.results.graph.entity.EntityResult;
+import org.hibernate.sql.results.internal.RowProcessingStateStandardImpl;
+import org.hibernate.sql.results.jdbc.internal.JdbcValuesSourceProcessingStateStandardImpl;
+import org.hibernate.sql.results.jdbc.spi.JdbcValues;
+import org.hibernate.sql.results.jdbc.spi.JdbcValuesMapping;
+import org.hibernate.sql.results.jdbc.spi.JdbcValuesSourceProcessingOptions;
+
+public class ForwardOnlyScrollableResultConsumer<R> implements ResultsConsumer<ScrollableResultsImplementor<R>, R> {
+	/**
+	 * Singleton access to the standard scrollable-results consumer instance
+	 */
+	public static final ForwardOnlyScrollableResultConsumer INSTANCE = new ForwardOnlyScrollableResultConsumer();
+
+	@SuppressWarnings("unchecked")
+	public static <R> ForwardOnlyScrollableResultConsumer<R> instance() {
+		return INSTANCE;
+	}
+
+	@Override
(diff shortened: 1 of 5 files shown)


frode-carlsen:
If only calling enLoading (which initializes the references) when stream is closed or last element read, it would mean the whole stream must be read before any of the objects are usable? This sort of defeats the purpose of using a stream?

hibernate-core/src/test/java/org/hibernate/orm/test/batchfetch/StreamTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+	@Test
+	public void testGetResultStream(SessionFactoryScope scope) {
+		scope.inTransaction(
+				session -> {
+					QueryImplementor<Child> query = session
+							.createQuery( "select c from Child as c where c.parent.someField=:someField", Child.class )
+							.setParameter( "someField", FIELD_VALUE );
+					Stream<Child> resultStream = query.getResultStream();
+					List<Child> children = resultStream.collect( Collectors.toList() );
```
frode-carlsen, line 57: maybe test directly on stream rather than doing a collect?  if a collect can trigger a close or other event, then it may behave differently than when processing individual stream entries(?)

beikov: Streams are not implicitly closed, so this is fine AFAICT as the `Stream#close` method and hence `endLoading` are not called.

hibernate-core/src/main/java/org/hibernate/sql/results/spi/ForwardOnlyScrollableResultConsumer.java
```diff
@@ -0,0 +1,88 @@
...
+import org.hibernate.sql.results.jdbc.internal.JdbcValuesSourceProcessingStateStandardImpl;
+import org.hibernate.sql.results.jdbc.spi.JdbcValues;
+import org.hibernate.sql.results.jdbc.spi.JdbcValuesMapping;
+import org.hibernate.sql.results.jdbc.spi.JdbcValuesSourceProcessingOptions;
+
+public class ForwardOnlyScrollableResultConsumer<R> implements ResultsConsumer<ScrollableResultsImplementor<R>, R> {
+	/**
+	 * Singleton access to the standard scrollable-results consumer instance
+	 */
+	public static final ForwardOnlyScrollableResultConsumer INSTANCE = new ForwardOnlyScrollableResultConsumer();
```
beikov, line 29: I'd prefer introducing an instance variable `boolean forwardOnly` in `ScrollableResultConsumer` and add another static final variable `ScrollableResultConsumer FORWARD_ONLY_INSTANCE = new ScrollableResultConsumer( true );` to `ScrollableResultConsumer` instead of duplicating the code. Also, this change would make some callsites megamorphic (starting with 3+ subclasses).

beikov (commented):
>If only calling enLoading (which initializes the references) when stream is closed or last element read, it would mean the whole stream must be read before any of the objects are usable? This sort of defeats the purpose of using a stream?

Correct me if I am wrong, but batch loading is triggered behind the scenes when doing lazy loading. So if you access an association object, the lazy loading code will look into the `BatchFetchQueue` and retrieve N - 1 ids from there, and load all of them together. The `BatchFetchQueue` is fed during the row processing.

So if you access every association of every row while processing, then you won't see any batching, as the `BatchFetchQueue` for this entity always has the size 1. 

If you want batching to happen properly, you will have to scroll past a few elements, to fill up the `BatchFetchQueue`, and only then access the association object. It might be worth doing some sort of buffering of N elements and process these in a batch or wrap the buffer in another stream, but I am not sure that code like this is something that should belong to Hibernate, though I could be convinced that something like a `hibernate.scroll_buffer_size` config option could be useful here, to control these buffer sizes in a new implementation of `ScrollableResultsImplementor`

hibernate-core/src/main/java/org/hibernate/sql/results/spi/ForwardOnlyScrollableResultConsumer.java
dreab8: yeap, this was just a draft to check if this was a viable solution.

dreab8:
Hi @frode-carlsen,

you are right unfortunately my solution does not work when processing individual stream entries.

What I'm noticing is that in ORM 5.x with streams we are loading the association after each row so not really batching but probably this is the only solution that works with stream. 
Not sure if wrapping the Stream and doing the batch after all the stream terminal operations can work !

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/5214

PR #5214, opened 2022-08-17
HHH-15449 @ManyToOne associations not loaded correctly with default EAGER and batch fetch property set when using TypedQuery.resultStream

dreab8:
https://hibernate.atlassian.net/browse/HHH-15449

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/batchfetch/EagerManyToOneStreamTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/batchfetch/EagerManyToOneStreamTest.java
@@ -1,0 +1,220 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.orm.test.batchfetch;
+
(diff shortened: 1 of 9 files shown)


beikov (approved):
LGTM

hibernate-core/src/test/java/org/hibernate/orm/test/batchfetch/EagerManyToOneStreamTest.java
```diff
@@ -0,0 +1,200 @@
...
+		final SQLStatementInspector sqlStatementInterceptor = (SQLStatementInspector) scope.getStatementInspector();
+		sqlStatementInterceptor.clear();
+		scope.inTransaction(
+				session -> {
+					QueryImplementor<Child> query = session
+							.createQuery( "select c from Child as c where c.parent.someField=:someField", Child.class )
+							.setParameter( "someField", FIELD_VALUE );
+					Stream<Child> resultStream = query.getResultStream();
+
+					List<Child> children = resultStream.collect( Collectors.toList() );
```
frode-carlsen, line 78: would propose checking the entries as they are streamed, rather than doing a collect first. Would catch issues related to differences in streams being handled as entries are read vs. after everything is read (which would be like getResultList)

dreab8: added `testGetResultStreamForEach`

frode-carlsen (approved):

PR #5215, opened 2022-08-17
HHH-15452 Predicate add expression results in UnsupportedOperationException

dreab8:
https://hibernate.atlassian.net/browse/HHH-15452

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmBooleanExpressionPredicate.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmBooleanExpressionPredicate.java
@@ -6,14 +6,13 @@
  */
 package org.hibernate.query.sqm.tree.predicate;
 
-import java.util.Collections;
+import java.util.ArrayList;
 import java.util.List;
 
 import org.hibernate.query.sqm.NodeBuilder;
 import org.hibernate.query.sqm.SemanticQueryWalker;
 import org.hibernate.query.sqm.tree.SqmCopyContext;
 import org.hibernate.query.sqm.tree.expression.SqmExpression;
-import org.hibernate.sql.ast.tree.predicate.NegatedPredicate;
 
 import jakarta.persistence.criteria.Expression;
 
@@ -73,7 +72,9 @@ public class SqmBooleanExpressionPredicate extends AbstractNegatableSqmPredicate
 
 	@Override
 	public List<Expression<Boolean>> getExpressions() {
-		return Collections.singletonList( booleanExpression );
+		final List<Expression<Boolean>> expressions = new ArrayList<>( 1 );
+		expressions.add( booleanExpression );
+		return expressions;
 	}
 
 	@Override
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmBooleanExpressionPredicate.java
```diff
@@ -73,7 +72,7 @@ public <T> T accept(SemanticQueryWalker<T> walker) {
 
 	@Override
 	public List<Expression<Boolean>> getExpressions() {
-		return Collections.singletonList( booleanExpression );
+		return Arrays.asList( booleanExpression );
```
beikov, line 75: The list which `Arrays.asList` returns can not be added to unfortunately, so we need to use this:

```suggestion
		final List<Expression<Boolean>> list = new ArrayList<>( 1 );
		list.add( booleanExpression );
		return list;
```

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmGroupedPredicate.java
```diff
@@ -59,7 +59,7 @@ public boolean isNegated() {
 
 	@Override
 	public List<Expression<Boolean>> getExpressions() {
-		return Collections.singletonList( subPredicate );
+		return Arrays.asList( subPredicate );
```
beikov, line 62: The list which `Arrays.asList` returns can not be added to unfortunately, so we need to use this:

```suggestion
		final List<Expression<Boolean>> list = new ArrayList<>( 1 );
		list.add( subPredicate );
		return list;
```

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmJunctionPredicate.java
```diff
@@ -88,8 +86,9 @@ public boolean isNegated() {
 
 	@Override
 	public List<Expression<Boolean>> getExpressions() {
-		//noinspection unchecked
-		return (List<Expression<Boolean>>) (List<?>) predicates;
+		ArrayList<Expression<Boolean>> expressions = new ArrayList<>( predicates.size() );
+		expressions.addAll( predicates );
+		return expressions;
```
beikov, line 91:
```suggestion
		return new ArrayList<>( predicates );
```

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmNegatedPredicate.java
```diff
@@ -58,7 +58,7 @@ public SqmPredicate getWrappedPredicate() {
 
 	@Override
 	public List<Expression<Boolean>> getExpressions() {
-		return Collections.singletonList( wrappedPredicate );
+		return Arrays.asList( wrappedPredicate );
```
beikov, line 61: The list which `Arrays.asList` returns can not be added to unfortunately, so we need to use this:

```suggestion
		final List<Expression<Boolean>> list = new ArrayList<>( 1 );
		list.add( wrappedPredicate );
		return list;
```

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/predicate/SqmBooleanExpressionPredicate.java
dreab8: Thanks @beikov

PR #5216, opened 2022-08-18
HHH-14387 Fix flush issues with bytecode enhanced lazy collections

beikov:

I experimented a lot to reach a condition where the initialization of the collection during deletion could cause issues, but ultimately couldn't come up with anything. Cancelling removes of entities (that refer to uninitialized collection) through cascades also didn't show any issues.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/Cascade.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/Cascade.java
@@ -13,11 +13,9 @@ import java.util.List;
 
 import org.hibernate.HibernateException;
 import org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor;
-import org.hibernate.bytecode.spi.BytecodeEnhancementMetadata;
 import org.hibernate.collection.spi.PersistentCollection;
 import org.hibernate.engine.spi.CascadeStyle;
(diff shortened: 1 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/engine/internal/Cascade.java
```diff
@@ -196,6 +202,30 @@ else if ( action.performOnLazyProperty() && type.isEntityType() ) {
 		}
 	}
 
+	private static void registerUninitializedLazyCollectionEntries(EventSource eventSource, EntityPersister persister, Object parent) {
+		if ( !persister.hasCollections() || !persister.hasUninitializedLazyProperties( parent ) ) {
```
dreab8, line 206: what about checking `persister.getBytecodeEnhancementMetadata().isEnhancedForLazyLoading()`?

beikov: That's the first thing `org.hibernate.tuple.entity.BytecodeEnhancementMetadataPojoImpl#hasUnFetchedAttributes` does, which is called through `hasUninitializedLazyProperties`

dreab8 (approved):
trying to think if there can be cases when not having the uninitialised collection in the PC when using  bytecode enhancement can cause problems, but I cannot figure out any so :+1:

sebersole (approved):

PR #5218, opened 2022-08-18
HHH 15454

karge-itestra:
https://hibernate.atlassian.net/browse/HHH-15454

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/jpa/spi/CriteriaQueryTupleTransformer.java
+++ b/hibernate-core/src/main/java/org/hibernate/jpa/spi/CriteriaQueryTupleTransformer.java
@@ -10,6 +10,7 @@ import java.util.List;
 import javax.persistence.Tuple;
 import javax.persistence.TupleElement;
 
+import org.hibernate.internal.util.type.PrimitiveWrapperHelper;
 import org.hibernate.query.criteria.internal.ValueHandlerFactory;
 import org.hibernate.transform.BasicTransformerAdapter;
 
@@ -101,7 +102,7 @@ public class CriteriaQueryTupleTransformer extends BasicTransformerAdapter {
 		public <X> X get(String alias, Class<X> type) {
 			final Object untyped = get( alias );
 			if ( untyped != null ) {
-				if ( !type.isInstance( untyped ) ) {
+				if (!elementTypeMatches(type, untyped)) {
 					throw new IllegalArgumentException(
 							String.format(
 									"Requested tuple value [alias=%s, value=%s] cannot be assigned to requested type [%s]",
@@ -126,17 +127,25 @@ public class CriteriaQueryTupleTransformer extends BasicTransformerAdapter {
 
 		public <X> X get(int i, Class<X> type) {
 			final Object result = get( i );
-			if ( result != null && !type.isInstance( result ) ) {
-				throw new IllegalArgumentException(
-						String.format(
-								"Requested tuple value [index=%s, realType=%s] cannot be assigned to requested type [%s]",
-								i,
-								result.getClass().getName(),
-								type.getName()
-						)
-				);
+			if (result != null) {
+				if (!elementTypeMatches(type, result)) {
+					throw new IllegalArgumentException(
+							String.format(
+									"Requested tuple value [index=%s, realType=%s] cannot be assigned to requested type [%s]",
+									i,
+									result.getClass().getName(),
+									type.getName()
+							)
+					);
+				}
 			}
 			return (X) result;
+		}
+
+		private <X> boolean elementTypeMatches(Class<X> type, Object untyped) {
+			return type.isInstance(untyped)
+					|| type.isPrimitive()
+					&& PrimitiveWrapperHelper.getDescriptorByPrimitiveType(type).getWrapperClass().isInstance(untyped);
 		}
 
 		public Object[] toArray() {
--- a/hibernate-core/src/test/java/org/hibernate/jpa/test/query/TupleQueryRetrievePrimitiveTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/jpa/test/query/TupleQueryRetrievePrimitiveTest.java
@@ -1,0 +1,85 @@
+package org.hibernate.jpa.test.query;
+
+import org.hibernate.jpa.test.metamodel.AbstractMetamodelSpecificTest;
+import org.hibernate.jpa.test.metamodel.ThingWithQuantity;
+import org.hibernate.jpa.test.metamodel.ThingWithQuantity_;
+import org.hibernate.testing.TestForIssue;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import javax.persistence.EntityManager;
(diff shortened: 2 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/jpa/spi/CriteriaQueryTupleTransformer.java
```diff
@@ -100,6 +101,9 @@ public Object get(String alias) {
 
 		public <X> X get(String alias, Class<X> type) {
 			final Object untyped = get( alias );
+			if (type.isPrimitive() && untyped != null && !untyped.getClass().isPrimitive()) {
+				type = PrimitiveWrapperHelper.getDescriptorByPrimitiveType(type).getWrapperClass();
+			}
 			if ( untyped != null ) {
 				if ( !type.isInstance( untyped ) ) {
 					throw new IllegalArgumentException(
```
beikov, line 109:
```suggestion
			if ( untyped != null ) {
				if ( !type.isInstance( untyped ) && ( !type.isPrimitive()  || !PrimitiveWrapperHelper.getDescriptorByPrimitiveType(type).getWrapperClass().isInstance( untyped ) ) {
					throw new IllegalArgumentException(
```

hibernate-core/src/main/java/org/hibernate/jpa/spi/CriteriaQueryTupleTransformer.java
```diff
@@ -112,7 +116,7 @@ public <X> X get(String alias, Class<X> type) {
 					);
 				}
 			}
-			return (X) untyped;
+			return type.cast(untyped);
```
beikov, line 119: The change seems unnecessary, please revert it

hibernate-core/src/main/java/org/hibernate/jpa/spi/CriteriaQueryTupleTransformer.java
```diff
@@ -125,18 +129,21 @@ public Object get(int i) {
...
 
 		public <X> X get(int i, Class<X> type) {
-			final Object result = get( i );
-			if ( result != null && !type.isInstance( result ) ) {
+			final Object untyped = get( i );
+			if (type.isPrimitive() && untyped != null && !untyped.getClass().isPrimitive()) {
+				type = PrimitiveWrapperHelper.getDescriptorByPrimitiveType(type).getWrapperClass();
+			}
+			if ( untyped != null && !type.isInstance( untyped ) ) {
 				throw new IllegalArgumentException(
```
beikov, line 137:
```suggestion
			if ( result != null && !type.isInstance( result ) && ( !type.isPrimitive()  || !PrimitiveWrapperHelper.getDescriptorByPrimitiveType(type).getWrapperClass().isInstance( result ) ) {
				throw new IllegalArgumentException(
```

hibernate-core/src/main/java/org/hibernate/jpa/spi/CriteriaQueryTupleTransformer.java
```diff
@@ -125,18 +129,21 @@ public Object get(int i) {
...
 								"Requested tuple value [index=%s, realType=%s] cannot be assigned to requested type [%s]",
 								i,
-								result.getClass().getName(),
+								untyped.getClass().getName(),
 								type.getName()
 						)
 				);
 			}
-			return (X) result;
+			return type.cast(untyped);
```
beikov, line 146: The change seems unnecessary, please revert it

beikov (changes requested):
Hey there, thanks for the PR. I have a few improvements though which I would like to ask you to implement. Instead of penalizing every call to `get` with additional checks, we should instead only check if a type is a primitive when `isInstance` returns false. If you could please do the changes I requested, I'd merge this.

hibernate-core/src/main/java/org/hibernate/jpa/spi/CriteriaQueryTupleTransformer.java
```diff
@@ -70,8 +71,8 @@ public <X> X get(TupleElement<X> tupleElement) {
 						"Requested tuple element did not correspond to element in the result tuple"
 				);
 			}
-			// index should be "in range" by nature of size check in ctor
-			return (X) tuples[index];
+			Class<? extends X> type = tupleElement.getJavaType();
+			return get(index, type);
```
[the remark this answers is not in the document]
karge-itestra: I would have to apply the same fix as in `get(index, type)` to that method. Would you prefer that I duplicate the primitive type check?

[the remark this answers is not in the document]
beikov: There would be something seriously wrong if the java type of the tuple element is not compatible with the objects we have in the array, so you don't need to change anything in this method.

hibernate-core/src/main/java/org/hibernate/jpa/spi/CriteriaQueryTupleTransformer.java
karge-itestra: The cast to `X` generates a warning that there is an unchecked cast, while `type.cast(untyped)` does not. The `NativeTupleImpl#get(String, Class<X>)` method does it like that as well. I can revert it but that brings back the warning.

karge-itestra:
@beikov thanks for the suggestions, after this force-push all your requirements are satisfied, plus redundancy is reduced, and the tests are more thorough.

hibernate-core/src/main/java/org/hibernate/jpa/spi/CriteriaQueryTupleTransformer.java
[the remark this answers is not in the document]
beikov: Please revert this, then I can merge the PR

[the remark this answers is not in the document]
karge-itestra: Sorry, I thought I had done this already, but it has slipped through. Now it is reverted.

beikov:
Thanks! After CI finishes, I'll merge this

beikov:
Looks like there is a checkstyle issue. Could you please run `./gradlew :hibernate-core:checkstyleMain` and fix the problems?

karge-itestra:
> Looks like there is a checkstyle issue. Could you please run `./gradlew :hibernate-core:checkstyleMain` and fix the problems?

Yep, looks like spaces to fine-tune indentation is not allowed.

karge-itestra:
The spaces for indentation detected by checkstyle have been fixed

beikov:
Thanks for the fix!

karge-itestra:
Shall I port the test to the current main branch and see whether that is still an issue, and make a PR for that if it is?

beikov:
>Shall I port the test to the current main branch and see whether that is still an issue, and make a PR for that if it is?

That would be nice, yeah :)

>you´re welcome, looking forward to the next maintainance release to have it in my project ;)

I'll try to get it released by the end of this or next week. I have one more backport that I want to do before finishing this 5.6 release

PR #5219, opened 2022-08-18

reda-alaoui:
https://hibernate.atlassian.net/browse/HHH-15100


Sanne:
Hi @reda-alaoui , many thanks for the PR - it looks like a good patch, but we'd highly prefer to not introduce such configuration properties; we're working hard to reduce configuration properties as they make it much harder to test all combinations, which leads to maintenance and quality concerns.

But I agree we should improve on how this is handled; dicussed it with the team this morning, and the most interesting outcome is that Hiberante ORM 6 shouldn't be affected as the parser design is entirely different.

For version 5 what I'd propose is to separate the map holding valid entries from the negative cache: we would apply a cap on the negative cache, but store all valid entity names - that would imply it naturally scales better with the complexity of the application. What do you think?

reda-alaoui:
Hi @Sanne ,

I don't know how many negative entries we need on our project. We would have to log this in our production environments and it would have little value for us.
Therefore, we are going to keep our fork until we move to Hibernate 6.

Feel free to adapt the code of this PR as you wish.

Sanne:
Superseded by https://github.com/hibernate/hibernate-orm/pull/5224

PR #5221, opened 2022-08-22
 java.lang.NullPointerException: Cannot invoke "org.hibernate.engine.spi.EntityEntry.getEntityKey()" because "ownerEntry" is nullHHH-15453 

dreab8:
https://hibernate.atlassian.net/browse/HHH-15453

The hql query in translated into

select u1_0.USER_ID, u1_0.AGENCY_ID, u1_0.USER_NAME 
    from
        USER_TABLE u1_0 
    where
        u1_0.USER_ID=2

with the following Initialized
        
        - EntitySelectFetchInitializer [User.agency] 
        - SelectEagerCollectionInitializer [User.roles] << has no owner
        - EntityResultInitializer [User]
        
EntitySelectFetchInitializer#initializeInstance() triggers: internalLoad of Agency before the User.roles collection owner has been set, the following query is executed:
        
select a1_0.AGENCY_ID,  a1_0.AGENCY_NAME 
    from
        AGENCY_TABLE a1_0 
    where
        a1_0.AGENCY_ID=?

with the following Initialized:
        
        - SelectEagerCollectionInitializer [Agency.users]
        - EnityResultInitializer [Agency]  
        
ListReusltsConsumers calls persistenceContext.initializeNonlazyCollections() that triggers
        

select u1_0.AGENCY_ID, u1_0.USER_ID, u1_0.USER_NAME 
    from
        USER_TABLE u1_0 
    where
        u1_0.AGENCY_ID=?
        
       with the following Initialized:
        
        - EntitySelectFetchInitializer [Agency.users.{element}.agency]
        - SelectEagerCollectionInitialzier [Agency.users.{element}.roles]  ---> the collection has no owner
        - EntityJoinedFetchInitializer [Agency.users.{element}]
        - SetInitializer [Agency.users]
        
       
 select r1_0.USER_ID, r1_1.ROLE_ID,  r1_1.ROLE_NAME 
    from
        USER_ROLE r1_0 
    join
        ROLE_TABLE r1_1 
            on r1_1.ROLE_ID=r1_0.ROLE_ID 
    where
        r1_0.USER_ID in(select
            u1_0.USER_ID 
        from
            USER_TABLE u1_0 
        where
            u1_0.AGENCY_ID=?)    
            
 after initializeNonlazyCollections() the connection is closed causing the failures of the executions of the remaining initilizer initializeInstance() calls.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/collection/spi/AbstractPersistentCollection.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/spi/AbstractPersistentCollection.java
@@ -282,7 +282,11 @@ public abstract class AbstractPersistentCollection<E> implements Serializable, P
 				}
 			}
 			else {
-				if ( !session.isTransactionInProgress() ) {
+				/*
+				 Whenever the collection lazy loading is triggered during the loading process,
+				 closing the connection will cause an error when RowProcessingStateStandardImpl#next() will be called.
+				 */
+				if ( !session.isTransactionInProgress() && session.getPersistenceContext().isLoadFinished() ) {
 					session.getJdbcCoordinator().afterTransaction();
 				}
 			}
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/fetch/FetchModeSubselectTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/fetch/FetchModeSubselectTest.java
@@ -1,0 +1,288 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.orm.test.fetch;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.hibernate.annotations.Fetch;
+import org.hibernate.annotations.FetchMode;
+
+import org.hibernate.testing.TestForIssue;
+import org.hibernate.testing.orm.junit.DomainModel;
+import org.hibernate.testing.orm.junit.SessionFactory;
+import org.hibernate.testing.orm.junit.SessionFactoryScope;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import jakarta.persistence.CascadeType;
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.JoinTable;
+import jakarta.persistence.ManyToMany;
+import jakarta.persistence.ManyToOne;
+import jakarta.persistence.OneToMany;
+import jakarta.persistence.Table;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+@DomainModel(
+		annotatedClasses = {
+				FetchModeSubselectTest.User.class,
+				FetchModeSubselectTest.Role.class,
+				FetchModeSubselectTest.Agency.class
+		}
+)
+@SessionFactory
+@TestForIssue(jiraKey = "HHH-15453")
+public class FetchModeSubselectTest {
+
+	@BeforeEach
+	public void setUp(SessionFactoryScope scope) {
+		scope.inTransaction(
+				session -> {
+					User ann = new User( 1, "Ann" );
+					User bob = new User( 2, "Bob" );
+
+					Role manager = new Role( 1, "manager" );
+					Role developer = new Role( 2, "developer" );
+
+					ann.addRole( manager );
+
+					Agency agency = new Agency( 100, "Test Agency" );
+
+					agency.addUser( ann );
+					agency.addUser( bob );
+
+					session.persist( ann );
+					session.persist( bob );
+					session.persist( manager );
+					session.persist( developer );
+					session.persist( agency );
+				}
+		);
+	}
+
+	@AfterEach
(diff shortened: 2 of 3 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/fetch/FetchModeSubselectTest.java
```diff
@@ -0,0 +1,266 @@
...
+		);
+	}
+
+	@Test
+	public void testSelectUserWithRole(SessionFactoryScope scope) {
+		scope.inSession(
+				session -> {
+					List<User> userList = session.createQuery( "FROM User WHERE id = 2", User.class ).list();
+					assertThat( userList.size() ).isEqualTo( 1 );
+					assertThat( userList.get( 0 ).getRoles().size() ).isEqualTo( 0 );
```
beikov, line 92: Looks like this should rather be:

```suggestion
					List<User> userList = session.createQuery( "FROM User WHERE id = 1", User.class ).list();
					assertThat( userList.size() ).isEqualTo( 1 );
					assertThat( userList.get( 0 ).getRoles().size() ).isEqualTo( 1 );
```

hibernate-core/src/main/java/org/hibernate/collection/spi/AbstractPersistentCollection.java
```diff
@@ -282,8 +283,8 @@ else if ( !session.isConnected() ) {
 				}
 			}
 			else {
-				if ( !session.isTransactionInProgress() ) {
-					session.getJdbcCoordinator().afterTransaction();
+				if ( !session.isTransactionInProgress() && !isPhysicallyConnected ) {
+					session.getJdbcCoordinator().getLogicalConnection().afterTransaction();
```
beikov, line 287: Can you add a comment here about why we only have to call `afterTransaction` if this is physically connected? Also, I'm having a hard time understanding why you changed from `session.getJdbcCoordinator().afterTransaction();` to `session.getJdbcCoordinator().getLogicalConnection().afterTransaction();` as the former code seemed to invoke some other code as well, see `org.hibernate.engine.jdbc.internal.JdbcCoordinatorImpl#afterTransaction`.

Overall, I am simply not sure I understand how this change is necessary for the subselect fetch issue and I think a code comment might help here.

dreab8: `session.getJdbcCoordinator().afterTransaction()` it is fine, I can revert this.

About the check to `isPhysicallyConnected`, not able to find a better way. The issue is that an `SelectEagerCollectionInitializer#initilaizeInstance` calls `persistenceContext.initializeNonlazyCollections()` that triggers this code that closes the connection but then when we iterate over the others initializers calling initializeInstance() they will not able to iterate the ResultSet. 
Hope my explanation make sense :)

hibernate-core/src/test/java/org/hibernate/orm/test/fetch/FetchModeSubselectTest.java
dreab8: :+1:

hibernate-core/src/main/java/org/hibernate/collection/spi/AbstractPersistentCollection.java
```diff
@@ -282,7 +283,11 @@ else if ( !session.isConnected() ) {
 				}
 			}
 			else {
-				if ( !session.isTransactionInProgress() ) {
+				/*
+				 Whenever the collection lazy loading is triggered while consuming a ResultSet the logical connection will be physical connected
+				 and closing the connection will cause an error to occur when RowProcessingStateStandardImpl#next() will be called.
+				 */
+				if ( !session.isTransactionInProgress() && !isPhysicallyConnected ) {
```
beikov, line 290: Ok, as far as I understand, it is only valid to call `afterTransaction` when the connection was acquired through this method. In that case, wouldn't it be better to check whether the stack in `PersistenceContext#getLoadContexts` is empty, and only then run this? i.e. it is safe to run `afterTransaction` when no load contexts are available, as that means this method acquired a connection in the first place.

hibernate-core/src/main/java/org/hibernate/collection/spi/AbstractPersistentCollection.java
```diff
@@ -282,7 +282,11 @@ else if ( !session.isConnected() ) {
 				}
 			}
 			else {
-				if ( !session.isTransactionInProgress() ) {
+				/*
+				 Whenever the collection lazy loading is triggered during the loading process,
+				 closing the connection will cause an error when RowProcessingStateStandardImpl#next() will be called.
+				 */
+				if ( !session.isTransactionInProgress() && session.getPersistenceContext().isLoadFinished() ) {
```
beikov, line 289: Wow, that is so much better and more comprehensible if that really works. Fingers crossed that tests run through 🤞

hibernate-core/src/main/java/org/hibernate/collection/spi/AbstractPersistentCollection.java
dreab8: `persistenceContext#getLoadContexts()` cannot be used because it creates one if null, so I used `session.getPersistenceContext().isLoadFinished()` instead, what do you think?

hibernate-core/src/main/java/org/hibernate/collection/spi/AbstractPersistentCollection.java
dreab8: Thanks for all your suggestions they led to this solution , I run it locally and for H2 all tests passed so :crossed_fingers:

dreab8: @beikov can I merge it?

dreab8: @beikov Thanks

cbush06:
Yes, I am encountering this too! Huge pain. Thanks for fixing.

PR #5222, opened 2022-08-22
HHH-15455 User Guide Contains a reference to hibernate-ehcache, no longer supported

dreab8:
https://hibernate.atlassian.net/browse/HHH-15455


beikov (commented):
There is a reference to the removed section on line 30 of this file which should probably be rewritten a bit.

beikov (approved):

PR #5227, opened 2022-08-23
HHH-15443 Allow JdbcType to wrap read and write expressions

beikov:
# JdbcType can now wrap read and write expressions for columns
The main motivation for this is to allow implementing support for SQL types, like JSONB, when the JDBC driver lacks support for binding/extracting values of that type or the driver classes are not visible.
Use cases are:

1. Wildfly might not expose JDBC driver classes, so we would have to use these special JDBC types
2. Driver does not support binding/extracting values of such types (e.g. Vertx or federation driver)

## Write expression wrapping
To wrap the write expression through a JdbcType, we have to determine the JdbcMapping for columns
 => hence Value#getSelectableType was introduced
 
Column writer wrapping is only necessary for parameters in inserts and when inferring the type from context
 
## Read expression wrapping
For top level SqlSelection (DomainResult), we have to apply column read expression wrapping, so SqlAstQueryPartProcessingStateImpl#resolveSqlSelection
will now call Expression#createDomainResultSqlSelection for top level selections, which does the wrapping.

https://hibernate.atlassian.net/browse/HHH-15443

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/jdbc/JdbcType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/jdbc/JdbcType.java
@@ -15,11 +15,13 @@ import org.hibernate.Incubating;
 import org.hibernate.dialect.Dialect;
 import org.hibernate.engine.jdbc.Size;
 import org.hibernate.query.sqm.CastType;
+import org.hibernate.sql.ast.spi.SqlAppender;
+import org.hibernate.sql.ast.spi.StringBuilderSqlAppender;
+import org.hibernate.sql.ast.tree.expression.Expression;
 import org.hibernate.type.SqlTypes;
 import org.hibernate.type.descriptor.ValueBinder;
 import org.hibernate.type.descriptor.ValueExtractor;
 import org.hibernate.type.descriptor.WrapperOptions;
-import org.hibernate.type.descriptor.java.BasicJavaType;
 import org.hibernate.type.descriptor.java.JavaType;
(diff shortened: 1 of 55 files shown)


gavinking:
I'm not sure I like what's going on here, but that could just be because I don't understand it. Could we talk it over sometime?

gavinking:
OK, so after discussing offline, my suggestion is that it would be better if this worked by having the `Type` double-dispatch back to the `Dialect`, and let the `Dialect` render the SQL.

The justification for this is: the generated SQL in principle depends on both the `Type` and the `Dialect`, and the SQL syntax in principle is a responsibility of the dialect.

hibernate-core/src/main/java/org/hibernate/type/descriptor/jdbc/JdbcType.java
```diff
@@ -160,6 +162,35 @@ default String getCheckCondition(String columnName, JavaType<?> javaType, Dialec
 		return null;
 	}
 
+	/**
+	 * Wraps the read expression to be able to read values with this JdbcType's ValueExtractor.
+	 * @since 6.2
+	 */
+	@Incubating
+	default Expression wrapReadExpression(Expression expression) {
```
beikov, line 170: Rename to something like `wrapTopLevelSelectionExpression`

sebersole:
Marked as non-draft to run the Actions.  Though they are still not running - presumably because of the changes to the CI set up being just in this PR?

beikov:
Merged via 7564551f6e85f52076a7b7fcb7e68c507393c261 and 16ca1a0595eb304348ebe2a2a126b0362f30d1ef

PR #5229, opened 2022-08-24
HHH-15454 Retrieve primitives from tuples

karge-itestra:
Same as #5218 for main branch (the same issue existed here as well)


beikov:
Great work, thanks!

PR #5230, opened 2022-08-24
HHH-15359 The entity returned by a merge doesn't contain @ManyToMany relation when the collection resides in @Embeddable

dreab8:
https://hibernate.atlassian.net/browse/HHH-15359


beikov (approved):
LGTM

PR #5237, opened 2022-08-25
HHH-15463 Adjust JdbcType based on DDL capacity for varchar/varbinary like types

beikov:
https://hibernate.atlassian.net/browse/HHH-15463


hibernate-core/src/main/java/org/hibernate/cfg/annotations/BasicValueBinder.java
```diff
@@ -204,7 +205,22 @@ public EnumType getEnumeratedType() {
...
 	public boolean isLob() {
-		return isLob;
+		if ( isLob ) {
+			return true;
+		}
+
+		if ( explicitJdbcTypeAccess != null ) {
+			final JdbcType type = explicitJdbcTypeAccess.apply( getTypeConfiguration() );
+			if ( type != null ) {
+				switch ( type.getJdbcTypeCode() ) {
```
sebersole, line 215: Maybe this instead?

```suggestion
				return type.isLob();
```

hibernate-core/src/main/java/org/hibernate/cfg/annotations/BasicValueBinder.java
```diff
@@ -239,7 +255,23 @@ public int getPreferredSqlTypeCodeForArray() {
...
+					case SqlTypes.NCHAR:
+					case SqlTypes.NVARCHAR:
+					case SqlTypes.LONGNVARCHAR:
+					case SqlTypes.LONG32NVARCHAR:
+					case SqlTypes.NCLOB:
+						return true;
+				}
```
sebersole, line 271: Following the suggestion above, how about adding `JdbcType#isNationalized` and using that here?

```suggestion
				return type.isNationalized();
```

hibernate-core/src/main/java/org/hibernate/cfg/annotations/BasicValueBinder.java
```diff
@@ -204,7 +205,22 @@ public EnumType getEnumeratedType() {
...
+
+		if ( explicitJdbcTypeAccess != null ) {
+			final JdbcType type = explicitJdbcTypeAccess.apply( getTypeConfiguration() );
+			if ( type != null ) {
+				switch ( type.getJdbcTypeCode() ) {
+					case SqlTypes.BLOB:
+					case SqlTypes.CLOB:
+					case SqlTypes.NCLOB:
+						return true;
+				}
```
sebersole, line 220: How about this instead?

```suggestion
				return type.isLob();
```

hibernate-core/src/main/java/org/hibernate/type/descriptor/sql/internal/Scale6IntervalSecondDdlType.java
```diff
@@ -18,11 +18,12 @@ public Scale6IntervalSecondDdlType(Dialect dialect) {
...
 		super( SqlTypes.INTERVAL_SECOND, typeNamePattern, dialect );
 	}
+
 	@Override
 	public String getTypeName(Long size, Integer precision, Integer scale) {
-		// The maximum scale for `interval second` is 6 unfortunately, so we have to use numeric by default
+		// The maximum scale for `interval second` is 6 unfortunately
 		if ( scale == null || scale > 6 ) {
-			return DdlTypeImpl.replace( "numeric($p,$s)", size, precision, scale );
+			throw new IllegalStateException( "Using interval second type with scale > 6 should not be possible!" );
```
sebersole, line 26: Yelling at the user is a bit much ;)

```suggestion
			throw new IllegalStateException( "Illegal attempt to use interval second type with scale > 6 " );
```

sebersole (commented):
Made a few suggestions

sebersole (approved):

beikov:
Will be merged as part of https://github.com/hibernate/hibernate-orm/pull/5239

PR #5238, opened 2022-08-25

Sanne:
"forward port" from 5.6 branch of #5224 - slightly different as this changed


beikov (approved):
LGTM

Sanne:
Now with fixed test..

PR #5240, opened 2022-08-26
HHH-15465 Upgrade to Jandex 3.0.0

dreab8:
https://hibernate.atlassian.net/browse/HHH-15465


Sanne:
thanks @dreab8 !

cc/ @Ladicek could you have a look too please?

Ladicek:
Looks good to me.

As I mentioned on #5243, I don't mind which one goes to `main`. I'd really like Jandex 3.0 (HHH-15466) compatibility on 5.6 and 6.0 too, for which this PR likely isn't acceptable and I've submitted my own (#5241 and #5242).

dreab8:
Hi @Ladicek,

at least for 6 it is seems that is is also necessary this change https://github.com/hibernate/hibernate-orm/pull/5240/files#diff-f2e4aaedb9ade73c4b7ce0e4575d4fdbd3b87b0cff18e3fa86ab05d506dd8a6fL118

Ladicek:
But that is just a local build plugin, that's never used at runtime I believe? So just to achieve runtime compatibility, while still building with Jandex 2.x, my changes in the other PRs should suffice?

beikov:
Looks good to me, though I think we should do the upgrade in 6.2 maybe.

PR #5241, opened 2022-08-26

Ladicek:
that is ever needed.


Sanne:
> Jandex 3.0.0 has a synthetic bridge
it always return null. 

Just curious, why would you add the method bridge if it's not actually useful towards achieving backwards compatibility?

Ladicek:
It achieves binary compatibility, and that allows achieving backward compatibility in a straightforward manner (this PR). Without that, it would be impossible e.g. for Hibernate to be compatible with both Jandex 2.x and 3.0 at the same time.

There's actually a lot of callers of `Indexer.index()` out there that outright ignore the return value -- those are compatible immediately.

Sanne:
ah, ok right that makes sense. Thanks!

PR #5242, opened 2022-08-26

Ladicek:
that is ever needed.


beikov:
Can you please create a new PR against the main branch? The 6.0 branch is not maintained anymore. Other than that, your change looks good to me

Ladicek:
PR against main is #5243. If 6.0 is not maintained anymore, I hope WildFly upgrades to 6.1 _real soon_, because it's on 6.0.2 currently :-) I see there's WFLY-16713, so that's probably fine.

PR #5243, opened 2022-08-26

Ladicek:
that is ever needed.


Ladicek:
I personally don't mind if this PR is used for `main`, or #5240 is used instead. I submitted an equivalent PR to this against 5.6 and 6.0, because those branches, I assume, can't afford breaking changes.

beikov (approved):

PR #5245, opened 2022-08-30
Improve the merge replace operation when original is equal to target

dreab8:


hibernate-core/src/main/java/org/hibernate/type/TypeHelper.java
```diff
@@ -90,6 +91,33 @@ else if ( target[i] == LazyPropertyInitializer.UNFETCHED_PROPERTY ) {
...
+			final Object owner,
+			final Map<Object, Object> copyCache) {
+		final Object[] values = persister.getValues( entity );
+		final Type[] types = persister.getPropertyTypes();
+		for ( int i = 0; i < types.length; i++ ) {
+			if ( values[i] != LazyPropertyInitializer.UNFETCHED_PROPERTY
+					&& values[i] != PropertyAccessStrategyBackRefImpl.UNKNOWN
+					&& !( types[i] instanceof AbstractStandardBasicType ) ) {
+				persister.setValue( entity, i, types[i].replace( values[i], values[i], session, owner, copyCache ) );
+			}
```
beikov, line 117: I guess the check `!( types[i] instanceof AbstractStandardBasicType )` is to determine if replace will actually do something? I don't think this is correct for mutable types like `java.util.Date`, so IMO the more correct approach is this:

```suggestion
			final Object oldValue = values[i];
			final Object newValue;
			if ( oldValue != LazyPropertyInitializer.UNFETCHED_PROPERTY
					&& oldValue != PropertyAccessStrategyBackRefImpl.UNKNOWN
					&& ( newValue = types[i].replace( values[i], values[i], session, owner, copyCache ) ) != oldValue ) {
				persister.setValue( entity, i, newValue );
			}
```

sebersole (approved):

PR #5246, opened 2022-08-30
HHH-15381 Try to fix gradle plugin publishing and platform publishing

beikov:
https://hibernate.atlassian.net/browse/HHH-15381


sebersole (commented):
If you verified that these work, then +1 - the changes seem reasonable

PR #5248, opened 2022-08-31
HHH-15216 Cannot change MetadataProvider implementation because JPAXMLOverriddenMetadataProvider is final and precisely expected by a cast operator

reda-alaoui:
Backport of #4976 to branch 5.6


beikov:
Thanks

PR #5250, opened 2022-09-01
HHH-13485 FilterJoinTable does not to use the defualt condition

dreab8:
https://hibernate.atlassian.net/browse/HHH-13485


beikov (approved):
LGTM

PR #5251, opened 2022-09-01
HHH-15356 Unable to get primary information on some dialects (ex: Oracle) when the PK column's position doesn't match the alphabetical name's order

dreab8:
https://hibernate.atlassian.net/browse/HHH-15356

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/AbstractInformationExtractorImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/AbstractInformationExtractorImpl.java
@@ -996,11 +996,15 @@ public abstract class AbstractInformationExtractorImpl implements InformationExt
 
 			final int columnPosition = resultSet.getInt( getResultSetColumnPositionColumn() );
 
+			final int index = columnPosition - 1;
+			// Fill up the array list with nulls up to the desired index, because some JDBC drivers don't return results ordered by column position
+			while ( pkColumns.size() <= index ) {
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/AbstractInformationExtractorImpl.java
```diff
@@ -999,9 +1002,10 @@ private PrimaryKeyInformation extractPrimaryKeyInformation(
 			final Identifier columnIdentifier = DatabaseIdentifier.toIdentifier(
 					resultSet.getString( getResultSetColumnNameLabel() )
 			);
-			final ColumnInformation column = tableInformation.getColumn( columnIdentifier );
-			pkColumns.add( columnPosition-1, column );
+			fkColumnsByPosition.put( columnPosition - 1, tableInformation.getColumn( columnIdentifier ) );
```
beikov, line 1005: Can we please use something like this instead of switching to a `TreeMap`?

```suggestion
			final int index = columnPosition - 1;
			// Fill up the array list with nulls up to the desired index, because some JDBC drivers don't return results ordered by column position
			while ( pkColumns.size() <= index ) {
				pkColumns.add( null );
			}
			pkColumns.set( index, tableInformation.getColumn( columnIdentifier ) );
```

dreab8: done, Thanks

beikov (approved):

PR #5252, opened 2022-09-02
HHH-15473 Instantiate collections in the default fetch group by default

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15473

Note the first two commits could be (and IMO, should be) applied to the `main` branch regardless of whether we merge the third commit. See the commit messages for more information.

As to whether getting rid of this specific performance optimization is a good idea:

1. I think we can all agree that performance optimizations should be balanced against usability; at least in this case [Gavin](https://github.com/hibernate/hibernate-orm/pull/3558#issuecomment-695003875) and [Sanne](https://github.com/hibernate/hibernate-orm/pull/3558#issuecomment-701440788) agree.
2. The optimization was not about avoiding database calls, just memory allocations.
3. As far as I know, we don't have numbers telling us whether the optimization was worth it.
4. As far as I can see, Hibernate ORM has always instantiated collections in the default fetch group for non-enhanced proxies, and [people have been relying on that](https://github.com/quarkusio/quarkus/discussions/27657).
5. We do have an example in this very pull request where the optimization is actually detrimental: see the assertion change in `OnDemandLoadTest`, where instantiating collections ahead of time allows us to initialize the collection with one temporary session instead of two.
6. I'm not completely sure, but I suspect that this optimization was the cause for several bugs that we had to fix, in particular [HHH-11766](https://hibernate.atlassian.net/browse/HHH-11766) and [HHH-14811](https://hibernate.atlassian.net/browse/HHH-14811). At the very least, the reproducers we added for these tickets do not reproduce the problem anymore after I set `enableCollectionInDefaultFetchGroup` to `true`.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/SessionFactoryBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/SessionFactoryBuilder.java
@@ -638,7 +638,8 @@ public interface SessionFactoryBuilder {
 	 * Should collections be included in the default fetch group when bytecode
 	 * enhancement is used?
 	 *
-	 * @param enabled {@code true} collections should be included
+	 * @param enabled {@code true} collections should be included, {@code false} they should not.
+	 * Default is {@code true}.
 	 *
 	 * @return {@code this}, for method chaining
 	 */
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java
@@ -196,7 +196,7 @@ public class SessionFactoryOptionsBuilder implements SessionFactoryOptions {
 	private boolean orderUpdatesEnabled;
 	private boolean orderInsertsEnabled;
 	private boolean postInsertIdentifierDelayed;
-	private boolean collectionsInDefaultFetchGroupEnabled;
+	private boolean collectionsInDefaultFetchGroupEnabled = true;
 
 	// JPA callbacks
 	private boolean callbacksEnabled;
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/bytecode/enhancement/cascade/CascadeDeleteCollectionTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/bytecode/enhancement/cascade/CascadeDeleteCollectionTest.java
@@ -72,7 +72,7 @@ public class CascadeDeleteCollectionTest extends BaseCoreFunctionalTestCase {
                     .setParameter( "name", "PARENT" )
                     .uniqueResult();
             checkInterceptor( loadedParent, false );
-            assertFalse( Hibernate.isPropertyInitialized( loadedParent, "children" ) );
+            assertFalse( Hibernate.isInitialized( loadedParent.getChildren() ) );
             s.delete( loadedParent );
         } );
         // If the lazy relation is not fetch on cascade there is a constraint violation on commit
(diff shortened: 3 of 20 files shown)


beikov (commented):
I would prefer if you leave tests untouched and instead configure the setting for all tests that depend on it, like you did for some tests already.

yrodiere:
The thing is, testing with `enableCollectionInDefaultFetchGroup` set to `false` has low value, as that's no longer the default and (as far as I can't tell) won't be the default in any know framework, nor will it be easily configurable by users.

That's why I only set `enableCollectionInDefaultFetchGroup` to `false` in a few tests: the tests that can only make sense with that setup.

The other tests (those that I changed) **do** make sense with `enableCollectionInDefaultFetchGroup` set to `true`, so it does offer some value to test that. And the only thing I had to change was the assumption that "property initialized = collection initialized", which is wrong, as explained in the commit message.

If you really want those tests to run unchanged with `enableCollectionInDefaultFetchGroup = false`, IMO the only sensible approach would be to be to copy-paste them, so that we test both `enableCollectionInDefaultFetchGroup = true` and `enableCollectionInDefaultFetchGroup = false`. But I really doubt that's useful.

yrodiere:
Interestingly, we got report that the problem we're fixing here might have been the cause for other issues in Quarkus, some of them really old: https://github.com/quarkusio/quarkus/discussions/27657#discussioncomment-3543599

beikov:
>If you really want those tests to run unchanged with enableCollectionInDefaultFetchGroup = false, IMO the only sensible approach would be to be to copy-paste them, so that we test both enableCollectionInDefaultFetchGroup = true and enableCollectionInDefaultFetchGroup = false. But I really doubt that's useful.

Yeah, probably we should do that. I just want to catch possible regressions of this code path in ORM rather than in Hibernate Reactive. I assume that debugging such issues through HR is a lot harder.

yrodiere:
> > If you really want those tests to run unchanged with enableCollectionInDefaultFetchGroup = false, IMO the only sensible approach would be to be to copy-paste them, so that we test both enableCollectionInDefaultFetchGroup = true and enableCollectionInDefaultFetchGroup = false. But I really doubt that's useful.
> 
> Yeah, probably we should do that. I just want to catch possible regressions of this code path in ORM rather than in Hibernate Reactive. I assume that debugging such issues through HR is a lot harder.

HR sets `enableCollectionInDefaultFetchGroup` to `true`, so I don't really see how that would help.

Anyway, I'll copy-paste the tests.

beikov:
Ok, I just read through some of the history and indeed, it seems like `enableCollectionInDefaultFetchGroup` being set to `false` by default might have been an optimization for lazy proxies i.e. `session.getReference(Entity.class, id)` when byte code enhancement is enabled. I guess that, by `enableCollectionInDefaultFetchGroup` being set to `false` we avoid creating `PersistentCollection` for these proxies. Can you please verify if `PersistentCollection` is created for lazy proxies when setting `enableCollectionInDefaultFetchGroup` to `true`?

If `PersistentCollection` instances are created, that would be a problem. If not, then I don't see any problem with the change of the default.

yrodiere:
> Anyway, I'll copy-paste the tests.

I feel like I'm adding technical debt for no reason, but... done.

> Can you please verify if `PersistentCollection` is created for lazy proxies when setting `enableCollectionInDefaultFetchGroup` to `true`?

I added `LazyProxyBytecodeEnhancementCollectionInitializationTest`, is that what you had in mind? It passes, btw.

beikov:
>I added LazyProxyBytecodeEnhancementCollectionInitializationTest, is that what you had in mind? It passes, btw.

Yes, that's what I had in mind. Good to see it passes, so +1 from me then

>I feel like I'm adding technical debt for no reason, but... done.

Given that "reference proxies" don't instantiate collection proxies, the benefits of not initializing `PersistentCollection` fields is probably very minimal. Let's add @sebersole who initially implemented this and @Sanne to the mix to get their approval on this :)

yrodiere:
I confirmed that having collection *not* instantiated immediately *did* cause a regression a long time ago, which was never addressed: https://github.com/quarkusio/quarkus/issues/4644. Turns out people serializing entities directly (ugh...) rely on being able to call a getter to then check if the collection is initialized or not.

This PR fixes the problem (after almost 3 years :D)

Now I just need an approval from @sebersole or @Sanne to merge this :)

Sanne (approved):
:+1: Thanks!

PR #5253, opened 2022-09-03
HHH-15480: Add Test Case

artbasov:
Please, find description in [HHH-15480](https://hibernate.atlassian.net/browse/HHH-7948)


beikov:

PR #5254, opened 2022-09-05
HHH-14387 Fix flush issues with bytecode enhanced lazy collections

beikov:

A backport of https://github.com/hibernate/hibernate-orm/pull/5216


beikov:
Can't backport this, as it would require a change to `org.hibernate.boot.spi.SessionFactoryOptions#isCollectionsInDefaultFetchGroupEnabled` to return `true` by default.

PR #5256, opened 2022-09-06
HHH-15483  allow @TenantId properties of type UUID, Long, Integer

gavinking:
See https://hibernate.atlassian.net/browse/HHH-15483.


gavinking:
I've very much improved the implementation, and now I think this is good to go.

beikov (approved):
LGTM. It's the best we can do without breaking the API/SPI

PR #5257, opened 2022-09-06
HHH-14387 Fix regression when bytecode enhancement enabled

dreab8:

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/WrapVisitor.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/WrapVisitor.java
@@ -9,7 +9,9 @@ package org.hibernate.event.internal;
 import org.hibernate.HibernateException;
 import org.hibernate.bytecode.enhance.spi.LazyPropertyInitializer;
 import org.hibernate.bytecode.enhance.spi.interceptor.EnhancementAsProxyLazinessInterceptor;
+import org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor;
 import org.hibernate.collection.spi.PersistentCollection;
+import org.hibernate.engine.spi.CollectionEntry;
 import org.hibernate.engine.spi.PersistenceContext;
 import org.hibernate.engine.spi.PersistentAttributeInterceptable;
 import org.hibernate.engine.spi.PersistentAttributeInterceptor;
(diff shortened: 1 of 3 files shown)


beikov (commented):
The change looks reasonable, but do I understand it right, that this can only happen when `SessionFactoryOptions#isCollectionsInDefaultFetchGroupEnabled` returns false? Maybe we should add that condition to the else if block?

yrodiere (approved):
It was a bit hard to test, because the upgrade to Jandex 3 *also* breaks the integration into Hibernate Search... I ended up reverting the Jandex 3 upgrade on top of your patch and testing that. As far as I can tell, it works.

I really need to finish working on Jandex 3... That's a separate problem, though.

> The change looks reasonable, but do I understand it right, that this can only happen when SessionFactoryOptions#isCollectionsInDefaultFetchGroupEnabled returns false? Maybe we should add that condition to the else if block?

If that's true, I'll need to update my PR (#5252) to set `collectionsInDefaultFetchGroupEnabled` to `false` explicitly in this test; otherwise it won't be a valid reproducer.

dreab8:
@yrodiere I checked and @beikov is right, the regression happens only when `collectionsInDefaultFetchGroupEnabled` is `false`

yrodiere:
@dreab8 My PR (https://github.com/hibernate/hibernate-orm/pull/5252) was merged before yours... you'll have to add something like this to your test:

https://github.com/hibernate/hibernate-orm/pull/5252/files#diff-2193473347b56b89e89a32cd4c8673b67e854cf9c6fb4657290414fda4d5f5b6R71-R86

hibernate-core/src/main/java/org/hibernate/event/internal/WrapVisitor.java
```diff
@@ -104,10 +108,31 @@ final Object processArrayOrNewCollection(Object collection, CollectionType colle
 					if ( attributeInterceptor instanceof EnhancementAsProxyLazinessInterceptor ) {
 						return null;
 					}
+					else if ( attributeInterceptor != null
+							&& !session.getFactory().getSessionFactoryOptions().isCollectionsInDefaultFetchGroupEnabled() ) {
```
beikov, line 112: I think it would be more correct to do the following instead:

```suggestion
							&& !attributeInterceptor.isAttributeLoaded( persister.getNavigableRole().getLocalName() ) ) {
```

dreab8:
@yrodiere 
I added
```
@Override
	protected void configureSessionFactoryBuilder(SessionFactoryBuilder sfb) {
		super.configureSessionFactoryBuilder( sfb );
		sfb.applyCollectionsInDefaultFetchGroup( true );
	}
```

hibernate-core/src/main/java/org/hibernate/event/internal/WrapVisitor.java
dreab8: you mean `&& ((LazyAttributeLoadingInterceptor)attributeInterceptor).isAttributeLoaded( persister.getAttributeMapping().getAttributeName() )`?

PR #5258, opened 2022-09-07
HHH-15485 Tests for Collections in default FetchGroup are not enabling this property

dreab8:
https://hibernate.atlassian.net/browse/HHH-15485


beikov (commented):
LGTM, but I think one test should be changed a bit.

beikov (approved):
LGTM

PR #5266, opened 2022-09-09
HHH-15481 NaturalId and Lazy create a class cast Exception

dreab8:
https://hibernate.atlassian.net/browse/HHH-15481


beikov (approved):
LGTM

PR #5269, opened 2022-09-09
Refactored NoCachingTransactionSynchronizationImpl to avoid timestamp…

dreab8:
… creation, Renamed CacheTransactionSynchronization#getCurrentTransactionStartTimestamp method to getCachingTimestamp and removed SharedSessionContractImplementor#getTransactionStartTimestamp method

https://hibernate.atlassian.net/browse/HHH-15502

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/internal/NoCachingTransactionSynchronizationImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/internal/NoCachingTransactionSynchronizationImpl.java
@@ -6,14 +6,35 @@
  */
 package org.hibernate.cache.internal;
 
-import org.hibernate.cache.spi.AbstractCacheTransactionSynchronization;
-import org.hibernate.cache.spi.RegionFactory;
+import org.hibernate.cache.spi.CacheTransactionSynchronization;
 
 /**
  * @author Steve Ebersole
  */
-public class NoCachingTransactionSynchronizationImpl extends AbstractCacheTransactionSynchronization {
-	public NoCachingTransactionSynchronizationImpl(RegionFactory regionFactory) {
-		super( regionFactory );
+public class NoCachingTransactionSynchronizationImpl implements CacheTransactionSynchronization {
+	public static final NoCachingTransactionSynchronizationImpl INSTANCE = new NoCachingTransactionSynchronizationImpl();
+
+	private NoCachingTransactionSynchronizationImpl() {
+
+	}
+
+	@Override
+	public long getCachingTimestamp() {
+		throw new UnsupportedOperationException("Method not supported when 2LC is not enabled");
+	}
+
+	@Override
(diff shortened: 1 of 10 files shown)


Sanne:
Thanks @dreab8 , I think that's a good approach.
I'll comment inline on a couple of clarifications, but I think we could then merge it.

hibernate-core/src/main/java/org/hibernate/cache/spi/CacheTransactionSynchronization.java
```diff
@@ -48,8 +48,10 @@ public interface CacheTransactionSynchronization {
 	 *
 	 * @implSpec This "timestamp" need not be related to timestamp in the Java
 	 * Date/millisecond sense.  It just needs to be an incrementing value.
+	 *
+	 * An UnsupportedOperationException is thrown for No Caching
```
Sanne, line 52: I'm not sure if it's clear what "No Caching" means. Could you clarify that?

hibernate-core/src/main/java/org/hibernate/cache/internal/NoCachingTransactionSynchronizationImpl.java
```diff
@@ -6,14 +6,35 @@
...
+public class NoCachingTransactionSynchronizationImpl implements CacheTransactionSynchronization {
+	public static final NoCachingTransactionSynchronizationImpl INSTANCE = new NoCachingTransactionSynchronizationImpl();
+
+	private NoCachingTransactionSynchronizationImpl() {
+
+	}
+
+	@Override
+	public long getCachingTimestamp() {
+		throw new UnsupportedOperationException("No Caching does not support Timestamp generation");
```
Sanne, line 23: I'm not sure if it's clear what "No Caching" means. Could you clarify that?

Sanne (changes requested):

Sanne:
> Thanks @dreab8 , I think that's a good approach. I'll comment inline on a couple of clarifications, but I think we could then merge it.

we'll also need a JIRA, of course.

Sanne:
@dreab8 could you also add this change please:
 - https://github.com/hibernate/hibernate-orm/commit/ab58157cae10da81c421ffe453983a9179d6a50c

hibernate-core/src/main/java/org/hibernate/cache/internal/NoCachingTransactionSynchronizationImpl.java
dreab8: changed, what do you think? any better?

hibernate-core/src/main/java/org/hibernate/cache/spi/CacheTransactionSynchronization.java
dreab8: changed, what do you think?

PR #5275, opened 2022-09-12
HHH-15497 Count query when counting polymorphic subclasses by type fails when run twice

dreab8:
https://hibernate.atlassian.net/browse/HHH-15497,

When a query is executed a second time the binding type adjustment is not executes, so aligning the SqmParameter expressible type with the binding type while converting SQM->AST will ensure that even for the following query execution the binding type will be correct one

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AbstractSqmExpression.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AbstractSqmExpression.java
@@ -156,4 +156,8 @@ public abstract class AbstractSqmExpression<T> extends AbstractJpaSelection<T> i
 	public JavaType<T> getJavaTypeDescriptor() {
 		return getNodeType() == null ? null : getNodeType().getExpressibleJavaType();
 	}
+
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AbstractSqmExpression.java
```diff
@@ -156,4 +156,8 @@ public JpaSelection<T> alias(String name) {
 	public JavaType<T> getJavaTypeDescriptor() {
 		return getNodeType() == null ? null : getNodeType().getExpressibleJavaType();
 	}
+
+	public void forceInferableType(SqmExpressible<?> type){
```
beikov, line 160: I think we should move the logic from `org.hibernate.query.internal.QueryParameterBindingImpl#setType` to `AbstractSqmParameter#applyInferableType` and call that method instead. Changing the type of the parameter binding is a no-go AFAIU from our last discussion, so we shouldn't do it and deprecate the `setType` method.

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/5277

PR #5277, opened 2022-09-13
HHH-15497 Count query when counting polymorphic subclasses by type fails when run twice

dreab8:
https://hibernate.atlassian.net/browse/HHH-15497

This is an alternative solution to https://github.com/hibernate/hibernate-orm/pull/5275


beikov (approved):
LGTM

PR #5279, opened 2022-09-13
HHH-15505 Getter of loaded entity returns null when using bytecode enhancement on entity whose field is defined both in mapped superclass and concrete entity

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15505

Backports: #5280, #5281


beikov (approved):
LGTM

PR #5284, opened 2022-09-13
HHH-15495/HHH-15578 Consider UUID to be comparable

beikov:
https://hibernate.atlassian.net/browse/HHH-15495
https://hibernate.atlassian.net/browse/HHH-15578

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/produce/function/ArgumentTypesValidator.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/produce/function/ArgumentTypesValidator.java
@@ -26,7 +26,9 @@ import org.hibernate.type.descriptor.jdbc.JdbcType;
 import org.hibernate.type.descriptor.jdbc.JdbcTypeIndicators;
 
 import java.lang.reflect.Type;
+import java.sql.Types;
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/produce/function/ArgumentTypesValidator.java
```diff
@@ -154,7 +155,11 @@ private int validateArgument(int count, JdbcMappingContainer expressionType, Str
 	private void checkType(int count, String functionName, FunctionParameterType type, int code, Type javaType) {
 		switch (type) {
 			case COMPARABLE:
-				if ( !isCharacterType(code) && !isTemporalType(code) &&!isNumericType(code) ) {
+				if ( !isCharacterType(code) && !isTemporalType(code) &&!isNumericType(code) && code != UUID ) {
+					if ( javaType == java.util.UUID.class && isVarbinaryType( code ) || isCharacterType( code ) ) {
+						// We also consider UUID to be comparable when it's a character or binary type
```
dreab8, line 160: I think instead of `isVarbinaryType( code )` it should be `isBinary(code)`

PR #5286, opened 2022-09-14
HHH-14943 byNaturalId API creates unparseable query (AND keyword instead of WHERE)

dreab8:
https://hibernate.atlassian.net/browse/HHH-14943


beikov (approved):
LGTM

PR #5291, opened 2022-09-14
HHH-15511 - fix version determination for CockroachDB

jrenaat:

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
@@ -7,6 +7,7 @@
 package org.hibernate.dialect;
 
 import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
 import java.time.temporal.TemporalAccessor;
@@ -15,6 +16,8 @@ import java.util.Date;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.TimeZone;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import jakarta.persistence.TemporalType;
 
@@ -35,6 +38,7 @@ import org.hibernate.engine.jdbc.env.spi.IdentifierHelper;
 import org.hibernate.engine.jdbc.env.spi.IdentifierHelperBuilder;
 import org.hibernate.engine.jdbc.env.spi.NameQualifierSupport;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.metamodel.mapping.EntityMappingType;
 import org.hibernate.metamodel.spi.RuntimeModelCreationContext;
 import org.hibernate.query.spi.QueryEngine;
@@ -63,6 +67,8 @@ import org.hibernate.type.descriptor.jdbc.spi.JdbcTypeRegistry;
 import org.hibernate.type.descriptor.sql.internal.DdlTypeImpl;
 import org.hibernate.type.descriptor.sql.internal.Scale6IntervalSecondDdlType;
 import org.hibernate.type.descriptor.sql.spi.DdlTypeRegistry;
+
+import org.jboss.logging.Logger;
 
 import static org.hibernate.query.sqm.TemporalUnit.DAY;
 import static org.hibernate.query.sqm.TemporalUnit.NATIVE;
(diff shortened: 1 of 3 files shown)


jrenaat:
CI will fail for CockroachDB because now that the version is (or should be) correctly determined, the `NativeQueryOrdinalParametersTest.testCteNativeQueryOrdinalParameter()` is no longer ignored, and actually fails so that will have to be fixed

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
```diff
@@ -111,8 +115,42 @@ public CockroachDialect() {
...
+				}
+			}
+			catch (SQLException ex) {
+				// Ignore
+			}
+		}
+		if ( databaseVersion == null ) {
+			throw new HibernateException("Could not determine the CockroachDB version");
+		}
+		return databaseVersion;
```
beikov, line 139: Just like we don't throw an exception in `Dialect.checkVersion()` we also shouldn't throw here. Please adapt that to return the minimum version and emit a warning log message.

```suggestion
		if ( databaseVersion == null ) {
			// todo: log a warning that the version can't be determined
			return MINIMUM_VERSION;
		}
		return databaseVersion;
```

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
```diff
@@ -111,8 +115,42 @@ public CockroachDialect() {
...
+		if ( m.find() ) {
+			String[] versionParts = m.group().substring( 1 ).split( "\\." );
+			// if we got to this point, there is at least a major version, so no need to check [].length > 0
+			int majorVersion = Integer.parseInt( versionParts[0] );
+			int minorVersion = versionParts.length > 1 ? Integer.parseInt( versionParts[1] ) : 0;
+			int microVersion = versionParts.length > 2 ? Integer.parseInt( versionParts[2] ) : 0;
+
+			return new SimpleDatabaseVersion( majorVersion, minorVersion, microVersion);
+		}
+		return null;
```
beikov, line 153:
```suggestion
	protected static DatabaseVersion parseVersion(String versionString) {
		try {
			final String[] versionParts = versionString.split( "\\." );
			// if we got to this point, there is at least a major version, so no need to check [].length > 0
			int majorVersion = Integer.parseInt( versionParts[0].substring( 1 ) );
			int minorVersion = versionParts.length > 1 ? Integer.parseInt( versionParts[1] ) : 0;
			int microVersion = versionParts.length > 2 ? Integer.parseInt( versionParts[2] ) : 0;

			return new SimpleDatabaseVersion( majorVersion, minorVersion, microVersion );
		}
		catch ( NumberFormatException e ) {
			// ignore
		}
		return null;
```

hibernate-core/src/test/java/org/hibernate/orm/test/dialect/unit/CockroachDialectVersionTest.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class CockroachDialectVersionTest {
+
+	@Test
+	@TestForIssue(jiraKey = "HHH-15511")
+	public void testCockroachDialectVersionParsing() {
+		DatabaseVersion dv;
+
+		String versionString = "CockroachDB CCL v21.2.10 (x86_64-unknown-linux-gnu, built 2022/05/02 17:38:58, go1.16.6)";
+		dv = CockroachDialect.parseVersion( versionString );
```
beikov, line 29: Instead of making that version parsing method public, please test this code by creating instances of `CockroachDialect` with `TestingDialectResolutionInfo`. You'll have to adapt that `TestingDialectResolutionInfo` a bit though by introducing a field `String databaseVersion` which should be assigned to what `getDatabaseVersion()` returns now by default, but a new `forDatabaseInfo` variant will additionally take the String explicitly.

beikov (changes requested):

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
jrenaat: What happened to the regular expression? I assume that removing it was a lapsus (just in case I tested your change and as-is it definitely won't work).

jrenaat: I'm ok with the the/catch (obviously), but I'm going to push back on removing the regex, sorru Christian ;)

beikov: I don't like the `Pattern` stuff, but it works. It's either try-catch or the `Pattern` thing you did. It might not matter in this specific case, but I generally prefer avoiding to create unnecessary intermediate objects like a `Pattern` if it's not helpful for readability or anything.

hibernate-core/src/test/java/org/hibernate/orm/test/dialect/unit/CockroachDialectVersionTest.java
jrenaat: Not sure if my approach is what you had in mind, let me know your thoughts

PR #5295, opened 2022-09-15
HHH-15500 Cache key is huge since migration to 6

dreab8:
https://hibernate.atlassian.net/browse/HHH-15500

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/internal/CollectionCacheKeyValueDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/internal/CollectionCacheKeyValueDescriptor.java
@@ -1,0 +1,29 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.cache.internal;
+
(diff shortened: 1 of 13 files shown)


Sanne:
hi @dreab8, sorry but I don't understand the relation between these changes and the issue - and there's no test?

Could you explain what the idea is? And how is Redisson encoding the keys? The example made on JIRA is horrible :)

dreab8:
Hi @Sanne, this change tries to reduce the size of the `CacheKeyValueDescriptor` (that is not anymore the entire `org.hibernate.type.BasicType` ), an being the `CacheKeyValueDescriptor`  a field of `CacheKeyImplementation`  this should reduce the size of the key.

Not sure how to create a proper test, any idea?

Sanne:
But how is the `CacheKeyValueDescriptor` converted into a Redis representation? Serialized?

dreab8:
@Sanne good question but I really don't know :(

beikov:
>But how is the CacheKeyValueDescriptor converted into a Redis representation? Serialized?

Yes, it is serialized. I'm working on something to improve this.

Sanne:
Ok, thanks for the info. But I'm still confused about what this PR is attempting to achieve - I admit I was primarily looking for a test in hope that this would shed some light :)

What's the idea @dreab8 ? And why was `CacheKeyValueDescriptor` introduced - it seems to re-introduce the requirement for each such element to be Serializable, which we removed before AFAIK for good reasons, so it feels like a step back.

beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/5324

PR #5296, opened 2022-09-15
HHH-15499 Introduce StandardFunctions constants class and align documentation

beikov:
https://hibernate.atlassian.net/browse/HHH-15499

This is not done yet. Also, I'll propose making other functions standard in follow up PRs, so this is not complete yet.

### DIFF
--- a/documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
+++ b/documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
@@ -1000,6 +1000,9 @@ First, functions for working with strings:
 |===
 | HQL Function | Purpose | Syntax | Notes on syntax
 
+| `chr()` | Create a character from ASCII code point | `chr(integer)` | Standard ANSI SQL
+| `char()` | Synonym to `chr()` | `char(integer)` | Standard ANSI SQL
+| `character_length()` | Synonym to `length()` | `character_length(string)` | Standard ANSI SQL
 | `position()` | Similar to `locate()` | `position(pattern in string)` | Standard ANSI SQL
 | `substring()` | An alternative syntax for JPQL's `substring()`
 | `substring(string from start)`, `substring(string from start for length)` | Standard ANSI SQL
@@ -1022,13 +1025,33 @@ Next, functions for working with numeric values:
 | `exp()` | Exponential function | `exp(x)` | Universal in SQL dialects
 | `power()` | Exponentiation | `power(x,y)` | Universal in SQL dialects
 | `ln()` | Natural logarithm | `ln(x)` | Very common in SQL dialects
+| `log()` | Base-N logarithm | `log(x,x)` | Very common in SQL dialects
+| `log2()` | Base-2 logarithm | `log2(x)` | Very common in SQL dialects
 | `log10()` | Base-10 logarithm | `log10(x)` | Very common in SQL dialects
-| `sin()`, `cos()`, `tan()`, `asin()`, `acos()`, `atan()`, `atan2()` | Basic trigonometric functions | `sin(theta)`, `cos(theta)`, `atan2(opposite, adjacent)` | Very common in SQL dialects
+| `sin()`, `cos()`, `tan()`, `cot()`, `asin()`, `acos()`, `atan()`, `atan2()`, `acot()` | Basic trigonometric functions | `sin(theta)`, `cos(theta)`, `atan2(opposite, adjacent)` | Very common in SQL dialects
+| `sinh()`, `cosh()`, `tanh()`, `coth()`, `asinh()`, `acosh()`, `atanh()`, `acoth()` | Basic trigonometric functions | `sinh(theta)`, `cosh(theta)` | Very common in SQL dialects
 | `round()` | Numeric rounding | As usual: `round(number, places)` | Very common in SQL dialects
 | `floor()` | Floor function | `floor(x)` | Universal in SQL dialects
 | `ceiling()` | Ceiling function | `ceiling(x)` | Very common in SQL dialects
 | `least()` | Return the smallest of the given arguments | `least(x, y, z)` | Very common in SQL dialects
 | `greatest()` | Return the largest of the given arguments | `greatest(x, y, z)` | Very common in SQL dialects
+| `degrees()` | Converts radians to degrees | `degrees(x)` | Very common in SQL dialects
+| `radians()` | Converts degrees to radians | `radians(x)` | Very common in SQL dialects
+| `pi()` | The pi constant | `pi()` | Very common in SQL dialects
+| `rand()` | Return random double between 0 and 1 | `rand()` | Very common in SQL dialects
+|===
+
+Also, some bit operation functions:
+
+|===
+| HQL Function | Purpose | Syntax | Notes on syntax
+
+| `bitand()` | Bitwise and i.e. `&` | `bitand(x,y)` | Standard ANSI SQL
+| `bitor()` | Bitwise or i.e. `\|` | `bitor(x,y)` | Standard ANSI SQL
+| `bitxor` | Bitwise xor i.e. `^` | `bitxor(x,y)` | Standard ANSI SQL
+| `bitnot()` | Bitwise not i.e. `~` | `bitnot(x)` | Standard ANSI SQL
+| `octet_length()` | The length in octets/bytes | `octet_length(x)` | Standard ANSI SQL
+| `bit_length()` | The length in bits | `bit_length(x)` | Standard ANSI SQL
 |===
 
 Finally, functions that evaluate the id, version, or natural id of an entity, or the foreign key of a to-one association:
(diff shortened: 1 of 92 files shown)


gavinking:
@beikov please roll back the changes you made to the docs. If I didn't document certain functions, it's, as usual, *for a good reason*. In this case, I assume it's because they're redundant and/or not completely portable. There is a canonical list of portable functions in a comment in the code for `Dialect`. If you wish to make changes to that canonical list, *please talk to me about it first*.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
```diff
@@ -1000,6 +1000,9 @@ First, functions for working with strings:
 |===
 | HQL Function | Purpose | Syntax | Notes on syntax
 
+| `chr()` | Create a character from ASCII code point | `chr(integer)` | Standard ANSI SQL
+| `char()` | Synonym to `chr()` | `char(integer)` | Standard ANSI SQL
+| `character_length()` | Synonym to `length()` | `character_length(string)` | Standard ANSI SQL
```
gavinking, line 1005: Rollback please.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
```diff
@@ -1022,13 +1025,33 @@ Next, functions for working with numeric values:
 | `exp()` | Exponential function | `exp(x)` | Universal in SQL dialects
 | `power()` | Exponentiation | `power(x,y)` | Universal in SQL dialects
 | `ln()` | Natural logarithm | `ln(x)` | Very common in SQL dialects
+| `log()` | Base-N logarithm | `log(x,x)` | Very common in SQL dialects
+| `log2()` | Base-2 logarithm | `log2(x)` | Very common in SQL dialects
```
gavinking, line 1029: Rollback please.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
```diff
@@ -1022,13 +1025,33 @@ Next, functions for working with numeric values:
...
 | `log10()` | Base-10 logarithm | `log10(x)` | Very common in SQL dialects
-| `sin()`, `cos()`, `tan()`, `asin()`, `acos()`, `atan()`, `atan2()` | Basic trigonometric functions | `sin(theta)`, `cos(theta)`, `atan2(opposite, adjacent)` | Very common in SQL dialects
+| `sin()`, `cos()`, `tan()`, `cot()`, `asin()`, `acos()`, `atan()`, `atan2()`, `acot()` | Basic trigonometric functions | `sin(theta)`, `cos(theta)`, `atan2(opposite, adjacent)` | Very common in SQL dialects
+| `sinh()`, `cosh()`, `tanh()`, `coth()`, `asinh()`, `acosh()`, `atanh()`, `acoth()` | Basic trigonometric functions | `sinh(theta)`, `cosh(theta)` | Very common in SQL dialects
```
gavinking, line 1032: Rollback please.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
```diff
@@ -1022,13 +1025,33 @@ Next, functions for working with numeric values:
...
 | `round()` | Numeric rounding | As usual: `round(number, places)` | Very common in SQL dialects
 | `floor()` | Floor function | `floor(x)` | Universal in SQL dialects
 | `ceiling()` | Ceiling function | `ceiling(x)` | Very common in SQL dialects
 | `least()` | Return the smallest of the given arguments | `least(x, y, z)` | Very common in SQL dialects
 | `greatest()` | Return the largest of the given arguments | `greatest(x, y, z)` | Very common in SQL dialects
+| `degrees()` | Converts radians to degrees | `degrees(x)` | Very common in SQL dialects
+| `radians()` | Converts degrees to radians | `radians(x)` | Very common in SQL dialects
+| `pi()` | The pi constant | `pi()` | Very common in SQL dialects
+| `rand()` | Return random double between 0 and 1 | `rand()` | Very common in SQL dialects
```
gavinking, line 1041: Rollback please.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
```diff
@@ -1022,13 +1025,33 @@ Next, functions for working with numeric values:
...
+
+|===
+| HQL Function | Purpose | Syntax | Notes on syntax
+
+| `bitand()` | Bitwise and i.e. `&` | `bitand(x,y)` | Standard ANSI SQL
+| `bitor()` | Bitwise or i.e. `\|` | `bitor(x,y)` | Standard ANSI SQL
+| `bitxor` | Bitwise xor i.e. `^` | `bitxor(x,y)` | Standard ANSI SQL
+| `bitnot()` | Bitwise not i.e. `~` | `bitnot(x)` | Standard ANSI SQL
+| `octet_length()` | The length in octets/bytes | `octet_length(x)` | Standard ANSI SQL
+| `bit_length()` | The length in bits | `bit_length(x)` | Standard ANSI SQL
```
gavinking, line 1054: Rollback this change too, please.

gavinking (changes requested):

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
gavinking: 1. AFAIK, neither `chr()` nor `char()` are in the ANSI spec. And for sure we don't need to add _both of them_ to the list of standard HQL functions. And I'm not sure why we would `char()` in these days of Unicode. Should we really be encouraging the use of ASCII?
2. I don't see why we need to add the ugly (with a yucky `_`) and verbose synonym `character_length()` as an alternative to JPA-standard `length()`.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
gavinking: 1. `log2()` is only supported on MySQL. It definitely doesn't belong in the list of standard HQL functions. I mean, we _could_ emulate it using `ln(x)/ln(2)` but that is not currently implemented. I don't think we need it.
2. `log(b,x)` _is_ supported natively on _most_ of the built-in dialects, but not all, and so if you want to add it to the list, you need to add code to emulate it using `ln(x)/ln(b)`.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
gavinking: 1. Hyperbolic functions aren't supported on all of the built-in SQL dialects, so they don't belong on this list.
2. `acot()` doesn't even exist.
3. Looks like you could add `cot()` if you feel you really need it.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
gavinking: 1. `pi()` is not supported on all built-in dialects, but it can be emulated using `asin(1.0)`.
2. `degrees()` and `radians()` aren't supported on all built-in dialects, but I guess they can be emulated using `/180*asin(1.0)`.
3. `rand()` is not supported on all built-in dialects, and making it portable might be tricky, I'm not sure.

None of these functions belong on the list unless someone does the work to make them work portably.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
gavinking: 1. I have no idea if `bitxor()` and `bitnot()` are implemented in a portable way, but for sure we have no tests for them. Are these *really* part of standard ANSI SQL??
2. I don't see why HQL needs `octet_length()` and `bit_length()` functions, and I can't imagine what you would use them for. There do seem to be a couple of very minimal tests for these, so maybe they are already portable. OTOH, they have ugly `_`s in their names, unlike our other functions. Ultimately, I'm against adding them to the list.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
gavinking: FTR, the hyperbolic functions can be emulated, of course, using `exp()`. But someone would need to do that work.

documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
gavinking: Also, `chr()` and `ascii()` are apparently not supported on Derby, and it doesn't look easy to emulate them.

PR #5298, opened 2022-09-16
HHH-15512 Querying for an entity persisted but not flushed to the database causes the exception 'A collection with cascade="all-delete-orphan" was no longer referenced by the owning entity instance'

dreab8:
https://hibernate.atlassian.net/browse/HHH-15512


beikov:
>When an entity has been persisted but the session has not been flushed and then a query is executed

Shouldn't the execution of the query, since it touches query spaces that have pending flushes, cause a partial flush before the query can execute? This is usually handled by `session.autoFlushIfRequired( affectedTableNames );`

Maybe I'm missing something, but IMO this fix shouldn't be necessary i.e. I suspect the bug is somewhere else.

dreab8:
yes the method is called but the db has been hit so in the `DefaultAutoFlushEventListener` the method `flushIsReallyNeeded(...)` returns false because `source.getActionQueue().areTablesToBeUpdated( event.getQuerySpaces() )` returns false.

beikov (approved):
That's how I imagined it should look :)

PR #5299, opened 2022-09-16
HHH-15487 - HHH-15488 (Remove support for PostgreSQL and PostgreSQL Advanced Server older than v10)

jrenaat:


jrenaat:
fails on pgsql 9.5, which makes sense if we just bumped the minimum version to 10; i guess we'd have to remove this CI task.

beikov (approved):
After rebasing, you can merge this

PR #5302, opened 2022-09-17
HHH-15522 Hibernate.isInitialized method not working for Envers Collections

fabrii:
https://hibernate.atlassian.net/browse/HHH-15522

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/Hibernate.java
+++ b/hibernate-core/src/main/java/org/hibernate/Hibernate.java
@@ -34,6 +34,7 @@ import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.proxy.HibernateProxy;
 import org.hibernate.proxy.LazyInitializer;
+import org.hibernate.collection.spi.LazyInitializable;
 
 /**
  * Various utility functions for working with proxies and lazy collection references.
@@ -85,8 +86,8 @@ public final class Hibernate {
 		if ( proxy instanceof HibernateProxy ) {
 			( (HibernateProxy) proxy ).getHibernateLazyInitializer().initialize();
 		}
-		else if ( proxy instanceof PersistentCollection ) {
-			( (PersistentCollection<?>) proxy ).forceInitialization();
+		else if ( proxy instanceof LazyInitializable ) {
+			( (LazyInitializable) proxy ).forceInitialization();
 		}
 		else if ( proxy instanceof PersistentAttributeInterceptable ) {
 			final PersistentAttributeInterceptable interceptable = (PersistentAttributeInterceptable) proxy;
@@ -111,8 +112,8 @@ public final class Hibernate {
 			final PersistentAttributeInterceptor interceptor = ( (PersistentAttributeInterceptable) proxy ).$$_hibernate_getInterceptor();
 			return !(interceptor instanceof EnhancementAsProxyLazinessInterceptor);
 		}
-		else if ( proxy instanceof PersistentCollection ) {
-			return ( (PersistentCollection<?>) proxy ).wasInitialized();
+		else if ( proxy instanceof LazyInitializable ) {
+			return ( (LazyInitializable) proxy ).wasInitialized();
 		}
 		else {
 			return true;
--- a/hibernate-core/src/main/java/org/hibernate/collection/spi/LazyInitializable.java
+++ b/hibernate-core/src/main/java/org/hibernate/collection/spi/LazyInitializable.java
@@ -1,0 +1,33 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.collection.spi;
+
+import org.hibernate.Incubating;
+
+/**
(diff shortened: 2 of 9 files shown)


hibernate-core/src/main/java/org/hibernate/Hibernate.java
```diff
@@ -34,32 +34,31 @@
...
- * the program to intervene in the process of serialization and of deserialization.
+ * Graphs of Hibernate entities obtained from a {@link Session} are usually in an amputated form, with associations and
+ * collections replaced by proxies and lazy collections. (That is, by instances of the internal types
+ * {@link HibernateProxy} and {@link PersistentCollection}.) These objects are fully serializable using Java
+ * serialization, but can cause discomfort when working with custom serialization libraries. Therefore, this class
+ * defines operations that may be used to write code that completely removes the amputated leaves of the graph (the
+ * proxies) during serialization, and rematerializes and reattaches them during deserialization. It's possible, in
+ * principle, to use these operations, together with reflection, or with the Hibernate metamodel, to write such generic
+ * code for any given serialization library, but the details depend on what facilities the library itself offers for the
+ * program to intervene in the process of serialization and of deserialization.
```
yrodiere, line 56: Please don't reformat code that you don't need to change, especially code unrelated to your patch.

Can you revert all those unnecessary reformattings, new lines, etc.?

hibernate-core/src/main/java/org/hibernate/Hibernate.java
```diff
@@ -88,6 +86,9 @@ public static void initialize(Object proxy) throws HibernateException {
 		else if ( proxy instanceof PersistentCollection ) {
 			( (PersistentCollection<?>) proxy ).forceInitialization();
 		}
+		else if ( proxy instanceof LazyInitializable ) {
+			( (LazyInitializable) proxy ).forceInitialization();
+		}
```
yrodiere, line 91: You don't need two `if` branches, as `PersistentCollection` extends `LazyInitializable`:

```suggestion
		else if ( proxy instanceof LazyInitializable ) {
			( (LazyInitializable) proxy ).forceInitialization();
		}
```

hibernate-core/src/main/java/org/hibernate/Hibernate.java
```diff
@@ -109,81 +110,80 @@ public static boolean isInitialized(Object proxy) {
...
 			final PersistentAttributeInterceptor interceptor = ( (PersistentAttributeInterceptable) proxy ).$$_hibernate_getInterceptor();
-			return !(interceptor instanceof EnhancementAsProxyLazinessInterceptor);
+			return !( interceptor instanceof EnhancementAsProxyLazinessInterceptor );
 		}
 		else if ( proxy instanceof PersistentCollection ) {
 			return ( (PersistentCollection<?>) proxy ).wasInitialized();
 		}
+		else if ( proxy instanceof LazyInitializable ) {
+			return ( (LazyInitializable) proxy ).wasInitialized();
+		}
```
yrodiere, line 120: You don't need two `if` branches, as `PersistentCollection` extends `LazyInitializable`:

```suggestion
		else if ( proxy instanceof LazyInitializable ) {
			return ( (LazyInitializable) proxy ).wasInitialized();
		}
```

yrodiere (changes requested):
Thanks, the tests and fix are looking good overall (just one suggestion).

You will need to get rid of all the unnecessary formatting changes, though: they will only mess up the git history.

hibernate-core/src/main/java/org/hibernate/Hibernate.java
fabrii: Sure

hibernate-core/src/main/java/org/hibernate/Hibernate.java
fabrii: I thought for a moment that leaving the two if's might simplify the understanding for everyone who is used to working with PersistenCollection. I agree that it is correct to eliminate double checking. I will make the change.

hibernate-core/src/main/java/org/hibernate/Hibernate.java
fabrii: It is ok to make a new commit reverting changes? Or is prefered to make a new "clean" history and force push the new commits to my branch?

yrodiere: You can make a new commit and squash everything together? `git rebase -i main` from the commandline, and follow the instructions in the editor that will open. I'm sure graphical interfaces to git have similar features. Otherwise just push a new commit and I'll squash everything for you.

yrodiere (approved):
LGTM! Thanks for your work.

yrodiere:
(I just rebased and force-pushed to your branch to solve problems with the build system; let's see how it goes)

yrodiere:
I just noticed that test classes don't have the `-Test` suffix in their name, which might prevent them from being executed... I'll fix that.

fabrii:
I noticed that the build failed because of checkstyle. Might be related to the "No new line at end of file" of [Hibernate class](https://github.com/hibernate/hibernate-orm/pull/5302/files#diff-98c491eec3c4f777cb90fe3f7d7b35ad07e838f7eb11b23a0f9d1b2fd075a714) ?

yrodiere:
> I noticed that the build failed because of checkstyle. Might be related to the "No new line at end of file" of [Hibernate class](https://github.com/hibernate/hibernate-orm/pull/5302/files#diff-98c491eec3c4f777cb90fe3f7d7b35ad07e838f7eb11b23a0f9d1b2fd075a714) ?

Yep, that was it. I had to fix some things in backports, so I fixed that as well and force-pushed.

PR #5317, opened 2022-09-20
HHH-15336 Deprecate SharedSessionContractImplementor#getTransactionStartTimestamp() and CacheTransactionSynchronization#getCurrentTransactionStartTimestamp()

dreab8:
https://hibernate.atlassian.net/browse/HHH-15536

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
@@ -167,7 +167,10 @@ public interface SharedSessionContractImplementor
 	 * @apiNote This "timestamp" need not be related to timestamp in the Java Date/millisecond
 	 * sense.  It just needs to be an incrementing value.  See
 	 * {@link CacheTransactionSynchronization#getCurrentTransactionStartTimestamp()}
+	 *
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/SharedSessionContractImplementor.java
```diff
@@ -167,7 +167,10 @@ default void checkOpen() {
 	 * @apiNote This "timestamp" need not be related to timestamp in the Java Date/millisecond
 	 * sense.  It just needs to be an incrementing value.  See
 	 * {@link CacheTransactionSynchronization#getCurrentTransactionStartTimestamp()}
+	 *
+	 * @deprecated no longer supported, when 2LC is enabled {{@link CacheTransactionSynchronization#getCachingTimestamp()}} can be used.
```
gavinking, line 171: It took me a while to figure out that "2LC" meant second-level cache. Could we use spell it out here please?

Sanne:
Do you think it's safe to be backported? I didn't expect that

dreab8:
Hi @Sanne, 
just deprecated the methods we removed in 6.2 and added the new one, why do you think it can be unsafe?

beikov (approved):
LGTM

Sanne:
> Hi @Sanne, just deprecated the methods we removed in 6.2 and added the new one, why do you think it can be unsafe?

ah, ok then :+1:  Sorry I assumed it was a backport.

PR #5320, opened 2022-09-20
HHH-15489 - Remove support for Oracle versions older than 11.2

jrenaat:


hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
```diff
@@ -581,12 +583,10 @@ protected String columnType(int sqlTypeCode) {
 			case TIMESTAMP:
```
beikov, line 583: You must remove this line here to fix the failing tests

beikov (changes requested):
After fixing what I requested, tests should be green again and we can merge this.

beikov (approved):

PR #5323, opened 2022-09-21
HHH-15235  and HHH-15045

dreab8:
https://hibernate.atlassian.net/browse/HHH-15235


beikov (commented):
Seems Derby fails on `org.hibernate.test.annotations.onetoone.EmbeddedIdTest`.

I don't quite understand why a `OneToOneType` where `isReferenceToIdentifierProperty() == false` is never "dirty". Could you explain that a little? Also, is this this also relevant for main or just an issue in 5.6?

PR #5324, opened 2022-09-21
HHH-15500 Cache key is huge since migration to 6

beikov:
https://hibernate.atlassian.net/browse/HHH-15500

Alternative to https://github.com/hibernate/hibernate-orm/pull/5295

The main motivation for the alternative is to not change existing contracts and thus keep backwards compatibility. In https://github.com/hibernate/hibernate-orm/pull/5295 `BasicType` does not implement `CacheKeyValueDescriptor` anymore and all subtypes of `AbstractStandardBasicType` now return a `JavaTypeCacheKeyValueDescriptor`.

This implementation instead implements delegation to a new `JavaType#toCacheKeyDescriptor` method in the concrete `BasicType` implementations, as well as some racy caching. Certain java types that might be used as primary key types have been adapted to return constant `CacheKeyValueDescriptor` instances, to reduce the serialization footprint of cache keys.

Another optimization has been implemented which reduces serialization footprint even further and also removes a megamorphic call site in the most common case, which is when the `CacheKeyValueDescriptor` is `DefaultCacheKeyValueDescriptor.INSTANCE`, in which case we initialize `null` for `CacheKeyImplementation#cacheKeyValueDescriptor`. The code was adapted to use standard `equals`/`hashCode` methods when that is `null`.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/internal/CacheKeyImplementation.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/internal/CacheKeyImplementation.java
@@ -50,10 +50,19 @@ public final class CacheKeyImplementation implements Serializable {
 			final String tenantId,
 			final SessionFactoryImplementor factory) {
 		this.id = id;
-		this.cacheKeyValueDescriptor = type.toCacheKeyDescriptor( factory );
 		this.entityOrRoleName = entityOrRoleName;
 		this.tenantId = tenantId;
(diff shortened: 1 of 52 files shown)


beikov:
I'll go ahead and merge this, as this is a general improvement, not just for Redis as cache provider. We can do further improvements in additional PRs in case you have specific ideas @Sanne

PR #5330, opened 2022-09-22
HHH-15045 + HHH-15235

dreab8:
https://hibernate.atlassian.net/browse/HHH-15235

This PR reverts [HHH-14216](https://hibernate.atlassian.net/browse/HHH-14216), see [discussion](https://github.com/hibernate/hibernate-orm/pull/5061) and [HHH-15540](https://hibernate.atlassian.net/browse/HHH-15540)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/EntityType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/EntityType.java
@@ -385,7 +385,7 @@ public abstract class EntityType extends AbstractType implements AssociationType
 	/**
 	 * Resolve an identifier or unique key value
 	 */
-	private Object resolve(Object value, SharedSessionContractImplementor session, Object owner) {
+	protected Object resolve(Object value, SharedSessionContractImplementor session, Object owner) {
 		if ( value != null && !isNull( owner, session ) ) {
 			if ( isReferenceToPrimaryKey() ) {
 				return resolveIdentifier( value, session, null );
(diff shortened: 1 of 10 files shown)


beikov (approved):

deAtog:
I'm fairly certain that the issue here can be solved by making correcting the versioning of the one-to-one relationships. Right now, all one-to-one relationships are versioned. This does not make sense to me as I do not see why you would version a property that is not stored in the database. The fact that a flush occurs was in my opinion correct. Consider a simple case where you load the parent and add a child. The state of the parent is modified and a flush to the underlying cache is required, but a database update is not. The existing code currently breaks because the one-to-one relationship from the parent to the child in this case is versioned. This causes the version of the parent to be incremented when the child is added to the parent. I think overriding the `isOptimisticLocked` method of the OneToOne type to return true only when the value is to be stored in the database would correct this issue.

The [OneToOne](https://docs.jboss.org/hibernate/jpa/2.1/api/javax/persistence/OneToOne.html) documentation provides the following examples for defining one-to-one relationships:
1. One-to-one association that maps a foreign key column.
    * This uses a OneToOne from the parent to the child and a ManyToOne from the child to the parent.
2. One-to-one association that assumes both the source and target share the same primary key values.
    * This uses a OneToOne from the parent and a OneToOne from the child to the parent with constrained set to true.
3. One-to-one association from an embeddable class to another entity.
    * This uses a ManyToOne from the parent to the child and a OneToOne from the child to the parent? (inferred from reading the code only).

In the above cases, the only time the OneToOne should be versioned is when the OneToOne is constrained in the second case. Therefore, the fix here may be to only return true from OneToOne.isOptimisticLocked() when OneToOne.isConstrained() returns true.

PR #5335, opened 2022-09-22
HHH-987 / HHH-992 

jrenaat:
I believe the problems described in the aforementioned issues are solved in v6.
Providing a test to cover these use cases.
Some cleanup done also.


beikov:
Great work! Thanks.

PR #5340, opened 2022-09-23
HHH-15509 deletion of unloaded entity

gavinking:
See https://hibernate.atlassian.net/browse/HHH-15509.


gavinking:
Unfortunately, but not surprisingly, this change breaks envers.

gavinking:
> Unfortunately, but not surprisingly, this change breaks envers.

Worked around this.

PR #5343, opened 2022-09-24
HHH-19841 test demoing bug in test or fix for HHH-11209

gavinking:
Either the test or the fix for HHH-11209 is bogus, because it leaves stuff hanging around in the action queue.

The test only passes by accident.


yrodiere:
@gavinking This looks like something that should lead to reopening the issue or creating another one? Do you have more context?

yrodiere:
I found this context: https://github.com/hibernate/hibernate-orm/commit/17e8b727e9f01121f1c0b3f6dd982a6c7808ae03

It was merged as part of https://hibernate.atlassian.net/browse/HHH-15509 / https://github.com/hibernate/hibernate-orm/pull/5340

None of which seems directly related to https://hibernate.atlassian.net/browse/HHH-11209, so I assume the problem is still valid and should be filed/fixed.

gavinking:
Should open an issue.

yrodiere:
Created https://hibernate.atlassian.net/browse/HHH-19841

beikov:

PR #5344, opened 2022-09-26
HHH-15509 correctly support @NotFound @ManyToMany

gavinking:
1. error if a non-@ManyToMany collection is @NotFound
2. disable constraint generation for @NotFound @ManyToMany
3. allow lazy fetching for @NotFound @ManyToMany
4. rework a completely bogus test so it makes sense


gavinking:
This is a good any rather low-impact change and I'm going to merge it.

PR #5345, opened 2022-09-26
HHH-15549 remove mapping of NUMBER(1,0) to BOOLEAN on Oracle

gavinking:
This just seems wrong to me. We have no way to say that a `NUMBER(1,0)` column isn't a single-digit number.

I mean, we might just as well say that every `CHAR(1)` column is a boolean, but we don't.


sebersole (commented):
Add a note to the migration guide, then +1

PR #5347, opened 2022-09-26
Backporting HHH-15484

Sanne:
The GraalVM dependency changed from `org.graalvm.nativeimage:svm` to `org.graalvm.sdk:graal-sdk`.

We don't typically allow dependency changes in a micro release, but consider that this is a module clearly flagged as experimental and the `svm` artifact is deprecated we'll need this.

Also good to know: this used to be an "AutomaticFeature" (aka it would activate automatically by just having it on classpath; but automatic features are also deprecated, so now the features included in our `hibernate-graalvm` module will need to be explicitly activated.

### DIFF
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -46,7 +46,7 @@ ext {
     jakartaJaxbRuntimeVersion = '3.0.0'
 
     //GraalVM
-    graalvmVersion = '21.3.0'
+    graalvmVersion = '22.2.0'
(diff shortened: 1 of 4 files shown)


Sanne:
Turns out `org.graalvm.sdk:graal-sdk` is not compatible with JDK8, so we can't do a straightforward backport. Following up via https://github.com/hibernate/hibernate-orm/pull/5350

PR #5352, opened 2022-09-27
HHH-1134 - tests to verify this use case

jrenaat:

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/mapping/inheritance/discriminator/AccountOwner.hbm.xml
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/mapping/inheritance/discriminator/AccountOwner.hbm.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0"?>
+<!--
+  ~ Hibernate, Relational Persistence for Idiomatic Java
+  ~
+  ~ License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+  ~ See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+  -->
+<!DOCTYPE hibernate-mapping PUBLIC
+        "-//Hibernate/Hibernate Mapping DTD 3.0//EN"
+        "http://www.hibernate.org/dtd/hibernate-mapping-3.0.dtd">
+
+<hibernate-mapping
+        package="org.hibernate.orm.test.mapping.inheritance.discriminator">
+
+    <class name="AccountOwner" table="AccOwner">
+        <id name="id">
+            <generator class="native"/>
+        </id>
+        <property name="description"/>
+
+        <set name="creditAccounts"
+             cascade="persist">
+            <key column="ownerId"/>
+            <one-to-many class="CreditAccount"/>
+        </set>
+
+        <set name="debitAccounts"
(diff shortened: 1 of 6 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/mapping/inheritance/discriminator/MultiInheritanceDiscriminatorTest.java
```diff
@@ -0,0 +1,225 @@
...
+		@GeneratedValue
+		private Long id;
+
+		private String name;
+
+		@OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, mappedBy = "owner")
+		private Set<AbstractAcc> creditAccs = new HashSet<>();
+
+		@OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, mappedBy = "owner")
+		private Set<AbstractAcc> debitAccs = new HashSet<>();
```
beikov, line 141: Actually, this should use the following mapping:

```suggestion
		@OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, mappedBy = "owner")
		private Set<CreditAcc> creditAccs = new HashSet<>();

		@OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, mappedBy = "owner")
		private Set<DebitAcc> debitAccs = new HashSet<>();
```

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/inheritance/discriminator/AccountOwner.hbm.xml
```diff
@@ -0,0 +1,54 @@
...
+            <one-to-many class="DebitAccount"/>
+        </set>
+    </class>
+
+    <class name="AbstractAccount" table="Accounts" abstract="true" dynamic-update="true">
+        <id name="id">
+            <generator class="native"/>
+        </id>
+
+        <discriminator column="DISC" force="true" insert="true" type="java.lang.String" not-null="true"/>
```
beikov, line 39: The hbm version works, because you are using `force="true"`, which is the workaround I mentioned to you

```suggestion
        <discriminator column="DISC" insert="true" type="java.lang.String" not-null="true"/>
```

jrenaat: True, I forgot to remove that. I should add though, that it *does* actually work without the force.

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/inheritance/discriminator/MultiInheritanceDiscriminatorTest.java
jrenaat: I resolved to adding `targetEntity = CreditAcc.class`on the mapping, which is more inline with the hbm mapping, and it boils down to the same thing, right? It allwos me to keep the parent reference as with the hbm test.

beikov: Yeah, it's essentially the same thing.

PR #5362, opened 2022-09-30
HHH-13439 Encode Hibernate ORM version into build time enhanced entities

beikov:
https://hibernate.atlassian.net/browse/HHH-13439


scottmarlow:
Thanks @beikov!

PR #5366, opened 2022-10-01
HHH-15564 add @SecondaryRow and HHH-15565 remove zombie code

gavinking:
Apparently, the code underlying `@Table(fetch=SELECT)` was removed in H6, but the `fetch` member was not deprecated, and some cruft was left lying around.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/annotations/SecondaryRow.java
+++ b/hibernate-core/src/main/java/org/hibernate/annotations/SecondaryRow.java
@@ -1,0 +1,48 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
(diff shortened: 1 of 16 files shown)


gavinking:
See https://github.com/hibernate/hibernate-orm/discussions/5359.

gavinking:
I think I'm going to go ahead and merge this. But folks, please take a look at this new API, and let me know if there's something you dislike about it.

sebersole:
What was the logic behind the name "SecondaryRow"?  Oddly worded to me.

gavinking:
Im not super attached to the name, but basically these two options control how the *row* lifecycle is managed. They don't affect the schema of or mapping to the *table*.

PR #5370, opened 2022-10-03
HHH 15554 + HHH-15572

dreab8:
https://hibernate.atlassian.net/browse/HHH-15554 Merge of an Entity with an immutable composite user type throws NPE 
https://hibernate.atlassian.net/browse/HHH-15572 Emulation of tuple comparison produces a wrong SQL query, missing a closing parenthesis
https://hibernate.atlassian.net/browse/HHH-15576 Emulation of tuple comparison produces a wrong SQL query for NOT EQUAL operator


beikov (approved):
LGTM

PR #5380, opened 2022-10-05
HHH-15581 Changes for the integration between Hibernate Reactive and ORM 6

DavideD:
A list of changes I need to make ORM 6 work with Hibernate Reactive.

I had to add a new interface so that I can create different actions for the insert operations: https://github.com/hibernate/hibernate-orm/commit/d098526e3d78f7f223b76e843d8ae7705526723d and https://github.com/hibernate/hibernate-orm/commit/076006120149b80ce97db435f0ecf8cb9613d32f

That's because in Hibernate Reactive we have an  interface `ReactiveEntityInsertAction` and it wasn't possible to add it to the queue when there was only using `Comparable<EntityAction>` (if somebody has a solution, I'm listening).

All the other changes make it possible for Hibernate Reactive to reuse or overrides some methods.

I've pushed the Hibernate Reactive branch upstream: https://github.com/hibernate/hibernate-reactive/tree/wip/ORM-6

I haven't created a JIRA yet, but if we agree with these changes I'm going to create one.

Thanks

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/action/internal/ComparableEntityAction.java
+++ b/hibernate-core/src/main/java/org/hibernate/action/internal/ComparableEntityAction.java
@@ -1,0 +1,17 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.action.internal;
+
+/**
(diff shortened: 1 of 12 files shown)


DavideD:
> I haven't created a JIRA yet, but if we agree with these changes I'm going to create one.

Nevermind, I've created a JIRA so that CI is happy.

beikov (commented):
I think that the changes in `AbstractEntityPersister` might cause you some trouble as @sebersole is changing how these things work in 6.2, so I guess he'll have to look over this or we wait until his work is merged before progressing with this PR. The other changes are mostly to internal types and are fine AFAICS.

PR #5382, opened 2022-10-05
mention window functions in HQL doc

gavinking:


gavinking:
@beikov does this look vaguely sensible?

Is there something else we should mention about window functions?

gavinking:
Thanks @beikov, I've incorporated that information.

PR #5384, opened 2022-10-06
HHH-15573 SpannerDialect, schema creation generates SQL containing 'null' instead of the column type

dreab8:
https://hibernate.atlassian.net/browse/HHH-15573


beikov (approved):
LGTM

PR #5389, opened 2022-10-06
HHH-15518 bless degrees() and radians() as standard functions

gavinking:
These aren't "standard" functions, but we should test them anyway.


gavinking:
Well, hell, it's easy enough to just bless these as "standard" so let's just do it.

PR #5247, opened 2022-08-31
Avoid creating a RollbackException when converting an exception

dreab8:
https://hibernate.atlassian.net/browse/HHH-15472

This PR should be applied to 6.2.0

PR #5394, opened 2022-10-07
get rid of ugly message

gavinking:
and refresh code in ExceptionConverterImpl

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/ExceptionConverterImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/ExceptionConverterImpl.java
@@ -25,8 +25,6 @@ import org.hibernate.engine.spi.ExceptionConverter;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.exception.LockAcquisitionException;
 import org.hibernate.loader.MultipleBagFetchException;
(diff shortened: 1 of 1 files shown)


dreab8:
Hi @gavinking, 

I'm not sure why we are returning a `RollbackException` but I think it is not correct so I created https://github.com/hibernate/hibernate-orm/pull/5247
what do you think?

gavinking:
Um where precisely? Did I mess up?

dreab8:
no you don't I think it was messed up, it believe but I may be wrong that it is not quite correct to always throw a `RollbackException` when converting a commit exception.

gavinking:
Ah Ok i will take a closer look when I get back home.

gavinking:
@dreadb8 I might be wrong, but the existing behavior looks correct-ish to me. 

The spec says:

>The `RollbackException` is thrown by the persistence provider when `EntityTransaction.commit` fails.

 and the Javadoc for `Transaction.commit()` saysL

| @throws RollbackException if the commit fails

which looks like what the code does currently.

Why do you think it's wrong?

dreab8:
ok, so just my bad interpretation of the `RollbackException`, sorry for the noise :( and thanks for the explanation :)

gavinking:
OK, cool 👍

PR #5396, opened 2022-10-07
HHH-15547 Multi-fetch with a single bag returns duplicates

dreab8:
https://hibernate.atlassian.net/browse/HHH-15547

Not sure this is the correct or best solution for the problem.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -52,6 +52,7 @@ import org.hibernate.internal.FilterHelper;
 import org.hibernate.internal.util.collections.CollectionHelper;
 import org.hibernate.internal.util.collections.Stack;
 import org.hibernate.internal.util.collections.StandardStack;
+import org.hibernate.loader.BagFetchException;
 import org.hibernate.loader.MultipleBagFetchException;
 import org.hibernate.metamodel.CollectionClassification;
 import org.hibernate.metamodel.MappingMetamodel;
@@ -98,6 +99,7 @@ import org.hibernate.metamodel.model.domain.PluralPersistentAttribute;
 import org.hibernate.metamodel.model.domain.SingularPersistentAttribute;
 import org.hibernate.metamodel.model.domain.internal.AnyDiscriminatorSqmPath;
 import org.hibernate.metamodel.model.domain.internal.AnyDiscriminatorSqmPathSource;
+import org.hibernate.persister.collection.OneToManyPersister;
 import org.hibernate.query.derived.AnonymousTupleTableGroupProducer;
 import org.hibernate.query.derived.AnonymousTupleType;
 import org.hibernate.metamodel.model.domain.internal.BasicSqmPathSource;
@@ -436,6 +438,7 @@ public abstract class BaseSqmToSqlAstConverter<T extends Statement> extends Base
 
 	private int fetchDepth;
 	private String currentBagRole;
+	private boolean containsCardinalityAlteringJoin;
 	private boolean resolvingCircularFetch;
 	private boolean deduplicateSelectionItems;
 	private ForeignKeyDescriptor.Nature currentlyResolvingForeignKeySide;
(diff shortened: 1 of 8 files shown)


beikov (approved):
LGTM

beikov:
Seems 
`BagFetchExceptionTest.fetchingABagAndJoiningAnotherCollectionShouldNotThrowAnException` is failing.

dreab8:
@beikov it seems we have the same issue even when the query contains a fetched bag and a join without fetching of another collection

beikov:
>@beikov it seems we have the same issue even when the query contains a fetched bag and a join without fetching of another collection

Yeah, since any *-to-many join will alter the cardinality of the fetch join for the bag. I guess we could throw an exception there too.

dreab8:
yeap, I noticed also that if we have an entity with a eager bag and an eager collection a `Sesison#get` returns a wrong result

hibernate-core/src/test/java/org/hibernate/orm/test/query/criteria/internal/hhh14916/Author.java
```diff
@@ -21,5 +22,6 @@ public class Author {
 	public String name;
 
 	@OneToMany(fetch = FetchType.LAZY, mappedBy = "author", orphanRemoval = true, cascade = CascadeType.ALL)
+	@OrderColumn
```
beikov, line 25: Instead of requiring this change, I'd much rather distinguish this `LIST` association from a "real bag". Inverse collections (using `mappedBy`) can never allow duplicates, so we should actually always do deduplication and not treat these associations like bags.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -2790,7 +2795,8 @@ private TableGroup consumeAttributeJoin(
 			SqmAttributeJoin<?, ?> sqmJoin,
 			TableGroup lhsTableGroup,
 			TableGroup ownerTableGroup,
-			boolean transitive) {
+			boolean transitive,
+			boolean explicit) {
```
beikov, line 2799: This is always true, so it seems unnecessary to introduce.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -2918,7 +2938,23 @@ private TableGroup consumeAttributeJoin(
 		return joinedTableGroup;
 	}
 
-	private TableGroup consumeCrossJoin(SqmCrossJoin<?> sqmJoin, TableGroup lhsTableGroup, boolean transitive) {
+	private TableGroup consumeCrossJoin(SqmCrossJoin<?> sqmJoin, TableGroup lhsTableGroup, boolean transitive, boolean explicit) {
+
+		final SqmPathSource<?> pathSource = sqmJoin.getReferencedPathSource();
+
+		if ( pathSource instanceof PluralPersistentAttribute ) {
```
beikov, line 2945: Is it really possible to have a cross join for an attribute?

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -7090,6 +7124,14 @@ else if ( getLoadQueryInfluencers().hasEnabledFetchProfiles() ) {
 		}
 	}
 
+//	private static boolean isABagThatCanContainsDuplicates(PluralAttributeMapping pluralAttributeMapping) {
+//		final CollectionClassification collectionClassification = pluralAttributeMapping.getMappedType()
+//				.getCollectionSemantics()
+//				.getCollectionClassification();
+//		return collectionClassification == CollectionClassification.BAG
+//				&& !( pluralAttributeMapping.getCollectionDescriptor() instanceof OneToManyPersister );
+//	}
```
beikov, line 7133:
```suggestion
	private static boolean isRealBag(PluralAttributeMapping pluralAttributeMapping) {
		final CollectionClassification collectionClassification = pluralAttributeMapping.getMappedType()
				.getCollectionSemantics()
				.getCollectionClassification();
		return collectionClassification == CollectionClassification.BAG
				&& pluralAttributeMapping.getCollectionDescriptor().getMappedByProperty() == null;
	}
```

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -7066,6 +7095,11 @@ else if ( getLoadQueryInfluencers().hasEnabledFetchProfiles() ) {
 							.getCollectionSemantics()
 							.getCollectionClassification();
 					if ( collectionClassification == CollectionClassification.BAG ) {
+						if ( containsCardinalityAlteringJoin && !(pluralAttributeMapping.getCollectionDescriptor() instanceof OneToManyPersister)) {
```
beikov, line 7098: We should only consider real bags in the `currentBagRole`

```suggestion
					if ( isRealBag( pluralAttributeMapping ) ) {
						if ( containsCardinalityAlteringJoin ) {
```

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -2806,14 +2809,40 @@ private TableGroup consumeAttributeJoin(
...
 			assert modelPart instanceof PluralAttributeMapping;
 
 			final PluralAttributeMapping pluralAttributeMapping = (PluralAttributeMapping) modelPart;
 
+			if ( pluralAttributeMapping.getMappedType()
+					.getCollectionSemantics()
+					.getCollectionClassification() != CollectionClassification.BAG ) {
+				containsCardinalityAlteringJoin = true;
+			}
+
```
beikov, line 2822: I think it would be easiest if you just call the `alterResultCardinality` method and encapsulate some logic there.

```suggestion
			alterResultCardinality( sqmJoin );
```

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -2806,14 +2809,40 @@ private TableGroup consumeAttributeJoin(
...
+					.getCollectionClassification() == CollectionClassification.BAG ) {
+				// if a bag collection descriptor is a OneToManyPersister then it cannot contain duplicates
+				if ( !( pluralAttributeMapping.getCollectionDescriptor() instanceof OneToManyPersister) ) {
+					// cannot join a collection belonging to a fetched BAG, it would produce a wrong cardinality for the BAG elements
+					throw new BagFetchException( sqmJoinNavigablePath.getIdentifierForTableGroup() );
+				}
+				else {
+					containsCardinalityAlteringJoin = true;
+				}
+			}
```
beikov, line 2844: I think this shouldn't be necessary after the other changes I proposed.

beikov (commented):
What do you think about moving the code to `org.hibernate.sql.results.graph.collection.internal.BagInitializer#readCollectionRow` instead? The deduplication is only relevant for join fetching, so it feels a bit misplaced for my taste to add that to `PersistentBag`.

PR #5397, opened 2022-10-07
HHH-15585: Schema validation for DB2 alias

dLoewy:


dreab8:
Hi @dLoewy , I had already created a PR for this issue https://github.com/hibernate/hibernate-orm/pull/5393 I do not think we need a new setting

PR #5402, opened 2022-10-11
CI: use ci/build.sh in Jenkinsfile and only run checkstyle in the H2 build

yrodiere:
cc @gavinking


gavinking:
Thanks man!!

yrodiere:
The TIDB is giving me trouble, I added a bit of logs, let's try again...

yrodiere:
Looks like this was a transient failure.

Everything seems to be working now, and I checked that `checkstyleMain` is only being executed when testing against H2.

This only affects CI, local builds will be unaffected.

So, ready to merge as far as I'm concerned.

beikov:
Great work, thanks!

PR #5406, opened 2022-10-12
HHH-15598 Removal of enhancer helpers

Sanne:
As proposed in https://hibernate.atlassian.net/browse/HHH-15598.

The org.hibernate.bytecode.spi.InstrumentedClassLoader is a little utility we don't use ourselves anymore. It's simple enough that anyone needing such a thing could re-implement it themselves so I suggest we remove it.

Same for org.hibernate.bytecode.spi.ByteCodeHelper: it has no use within our own runtime code, but it's used by tests - so I'd move it to the testing utilities.

Since these are SPIs I’m proposing the change for 6.2, without need for a deprecation cycle.

I’ll send a draft to check if this is agreeable.

Would this be OK @sebersole , @scottmarlow , @yrodiere  ? I'd like to clean this up soon as it limits the use of ByteBuddy to a specific couple of packages, makes some further cleanup easier.


yrodiere (approved):
No objection here.

Note that `InstrumentedClassLoader` kind of duplicates `org.hibernate.testing.bytecode.enhancement.BytecodeEnhancerRunner.EnhancingClassLoader`, unless I'm mistaken.

Sanne:
Good point about `EnhancingClassLoader`, but it's in testing already so it doesn't bother me - and it's actually useful :)

sebersole (approved):

scottmarlow:
Seems fine to me.  Should https://docs.jboss.org/hibernate/orm/6.0/internals/internal.txt be updated to include org.hibernate.bytecode.spi as well?

scottmarlow (approved):

Sanne:
> Seems fine to me. Should https://docs.jboss.org/hibernate/orm/6.0/internals/internal.txt be updated to include org.hibernate.bytecode.spi as well?

No objections but I'm not sure - it's literally an SPI, do we consider those internal?

PR #5411, opened 2022-10-14
Reproducer for HHH-15604 Identically-named association in entity root and elementcollection of embeddables leads to assertion error

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15604

Just a reproducer, no fix at the moment.


yrodiere:
Superseded by https://github.com/hibernate/hibernate-orm/pull/5863

PR #5421, opened 2022-10-18
HHH-15566 Improve efficiency of CallbackRegistryImpl

Sanne:
- https://hibernate.atlassian.net/browse/HHH-15566


dreab8 (approved):

hibernate-core/src/main/java/org/hibernate/jpa/event/internal/CallbackRegistryImpl.java
```diff
@@ -6,51 +6,57 @@
...
+	private final MapBackedClassValue<Callback[]> postLoads;
+
+	public CallbackRegistryImpl(
+			ConcurrentMap<Class<?>, Callback[]> preCreates,
+			ConcurrentMap<Class<?>, Callback[]> postCreates,
+			ConcurrentMap<Class<?>, Callback[]> preRemoves,
+			ConcurrentMap<Class<?>, Callback[]> postRemoves,
+			ConcurrentMap<Class<?>, Callback[]> preUpdates,
+			ConcurrentMap<Class<?>, Callback[]> postUpdates,
+			ConcurrentMap<Class<?>, Callback[]> postLoads) {
```
beikov, line 44: Maybe I'm missing something, but it seems the `Builder` is not used concurrently and the only other method that mutates maps is `CallbackRegistryImpl#release` which is called by `EventEngine#stop`, though that method has no caller. So we could use a plain `HashMap` as well, right? Apart from that, is there any reason for clearing the maps? The `EventEngine` and hence the `CallbackRegistryImpl` will become unreachable as soon as the `SessionFactoryImpl` becomes unreachable, so clearing the maps seems a bit moot to me, especially since it seems to require the use of a concurrent map.

hibernate-core/src/main/java/org/hibernate/internal/util/collections/MapBackedClassValue.java
```diff
@@ -0,0 +1,52 @@
...
+	 * we won't be clearing the ClassValue: this is useful
+	 * only to avoid classloader leaks since the Map
+	 * may hold references to user classes.
+	 * Since ClassValue is also possibly caching state,
+	 * it might be possible to retrieve some values after this
+	 * but shouldn't be relied on.
+	 * ClassValue doesn't leak references to classes.
+	 */
+	public void dispose() {
+		this.map.clear();
```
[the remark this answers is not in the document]
Sanne: ClassValue internally uses weak references to Class types - I mentioned that in the above comments

hibernate-core/src/main/java/org/hibernate/jpa/event/internal/CallbackRegistryImpl.java
Sanne: >  So we could use a plain HashMap as well, right?

No, while this Map isn't going to mutated it's still going to be *read* concurrently and that requires a ConcurrentMap. BTW a `ConcurrentHashMap` outperforms a plain `HashMap` in such use case as well.

> Apart from that, is there any reason for clearing the maps? 

I wondered about that myself and wasn't entirely sure. It seems previously it was very careful with these references so I tried to maintain that. I suppose it's possible that we get in a reference circularity, which gets quite bad when there's Class types and classloaders involved.

Sanne:
I've revised this slightly, preferring to opt for a defensive copy of the Map rather than forcing to be provided with a ConcurrentMap; this mostly to consume less memory.

Also, we're now able to throw an exception for unintended usage after dispose.

PR #5422, opened 2022-10-18
HHH-15610 Try getting rid of CacheKeyValueDescriptor

beikov:
https://hibernate.atlassian.net/browse/HHH-15610


sebersole:
`CacheKeyValueDescriptor` is internal, but exposed via API/SPI.  Yes, that was a mistake but it-is-what-it-is now.  

I'm more inclined to deprecate it for the time being

gavinking:
> I'm more inclined to deprecate it for the time being

In which version did we introduce it?

beikov:
Since the API/SPI was leaking an internal type which people couldn't have relied on anyway, I'll continue with the removal after some internal discussion about it.

PR #5423, opened 2022-10-18
explain wtf 'join lateral' means

gavinking:
...and that it's the same thing as `cross apply`.


gavinking:
@beikov does all this look correct to you?

beikov (approved):
LGTM

PR #5424, opened 2022-10-19
HHH-15613 remove 'lateral' from fromRoot rule

gavinking:
This is currently ignored by HQL, is undocumented, and untested, and anyway would be a synonym for `join lateral`.

Also squash some warnings in `SemanticQueryBuilder`.

### DIFF
--- a/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
+++ b/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4
@@ -176,8 +176,8 @@ entityWithJoins
  * A root entity declaration in the 'from' clause, with optional identification variable
  */
 fromRoot
-	: entityName variable?									# RootEntity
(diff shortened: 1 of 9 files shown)


beikov (approved):
LGTM. Might be worth also getting rid of `org.hibernate.query.criteria.JpaSelectCriteria#fromLateral`

gavinking:
Ah, sure, OK, will do!

gavinking:
Oh right, now I see why this seemed to be unused, because of this untypesafe code:

```java
final boolean lateral = ( (TerminalNode) firstChild ).getSymbol().getType() == HqlParser.LATERAL;
```

which mean that nothing broke when I changed the grammar.

We *really* need to move away from doing this sort of stuff in `SemanticQueryBuilder`, it's super-fragile!

gavinking:
This is now https://hibernate.atlassian.net/browse/HHH-15613.

PR #5432, opened 2022-10-19
HHH-15616 Mitigate performance impact of entity enhancement on Klass's _secondary_super_cache

Sanne:
Proposal for
 - https://hibernate.atlassian.net/browse/HHH-15616

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/EnhancedEntity.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/EnhancedEntity.java
@@ -1,0 +1,28 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.engine.spi;
+
+/**
+ * This is a special marker interface designed to represent the union of several traits:
+ * - ManagedEntity
+ * - Managed
+ * - PersistentAttributeInterceptable
+ * - SelfDirtinessTracker
+ * The need for such a "union" isn't natural in the Java language, but represents a technicality
+ * we need to bypass performance issues caused by https://bugs.openjdk.org/browse/JDK-8180450
+ * @see org.hibernate.engine.internal.ManagedTypeHelper
+ * @see org.hibernate.engine.spi.Managed
+ * @see org.hibernate.engine.spi.ManagedEntity
+ * @see PersistentAttributeInterceptable
+ * @see SelfDirtinessTracker
+ */
+public interface EnhancedEntity extends ManagedEntity, PersistentAttributeInterceptable, SelfDirtinessTracker {
+
+	//TODO what about ExtendedSelfDirtinessTracker ?
+	//TODO CompositeTracker, ManagedMappedSuperclass
+
+}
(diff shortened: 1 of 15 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/EnhancedEntity.java
```diff
@@ -0,0 +1,28 @@
...
+ * we need to bypass performance issues caused by https://bugs.openjdk.org/browse/JDK-8180450
+ * @see org.hibernate.engine.internal.ManagedTypeHelper
+ * @see org.hibernate.engine.spi.Managed
+ * @see org.hibernate.engine.spi.ManagedEntity
+ * @see PersistentAttributeInterceptable
+ * @see SelfDirtinessTracker
+ */
+public interface EnhancedEntity extends ManagedEntity, PersistentAttributeInterceptable, SelfDirtinessTracker {
+
+	//TODO what about ExtendedSelfDirtinessTracker ?
```
beikov, line 25: IMO `ExtendedSelfDirtinessTracker` should be merged into `SelfDirtinessTracker`. I don't see a value in having a separate "trait" for this. Also, I think we should deprecate the `inlineDirtyChecking` property and enable this by default to get rid of `SelfDirtinessTracker` in general. It was nice to have the option to disable this while it was new, but let's get over it and prepare to remove the config setting.

hibernate-core/src/main/java/org/hibernate/engine/spi/EnhancedEntity.java
```diff
@@ -0,0 +1,28 @@
...
+	//TODO CompositeTracker, ManagedMappedSuperclass
```
beikov, line 26: When getting rid off `inlineDirtyChecking`, we can also merge `CompositeTracker` with e.g. `ManagedComposite` to a `EnhancedComposite`. IMO we could introduce that already.

hibernate-core/src/main/java/org/hibernate/engine/spi/EnhancedEntity.java
```diff
@@ -0,0 +1,28 @@
...
+ * - SelfDirtinessTracker
+ * The need for such a "union" isn't natural in the Java language, but represents a technicality
+ * we need to bypass performance issues caused by https://bugs.openjdk.org/browse/JDK-8180450
+ * @see org.hibernate.engine.internal.ManagedTypeHelper
+ * @see org.hibernate.engine.spi.Managed
+ * @see org.hibernate.engine.spi.ManagedEntity
+ * @see PersistentAttributeInterceptable
+ * @see SelfDirtinessTracker
+ */
+public interface EnhancedEntity extends ManagedEntity, PersistentAttributeInterceptable, SelfDirtinessTracker {
```
beikov, line 23: I also think we should deprecate the `lazyLoading` option and enable it by default, so that we can get rid off `PersistentAttributeInterceptable` as a separate trait in the long run.

hibernate-core/src/main/java/org/hibernate/engine/spi/EnhancedEntity.java
Sanne: Agreed - but you're looking at the patch version for 5.6. IMO for more substantial refactoring that belongs in `main`

beikov: Of course, I just wanted to add these comments somewhere :)

beikov:
FYI, I create a JIRA issue for the deprecation: https://hibernate.atlassian.net/browse/HHH-15641

teopapath:
In version 6.2.9 of hibernate-enhance-maven-plugin it seems that enableLazyInitialization is set to false by default although in https://docs.jboss.org/hibernate/orm/6.2/migration-guide/migration-guide.html it is stated that it will be enabled by default. 
Do i miss something or is it correct that LazyInitialization is not working in 6.2?

teopapath:
It seems that the removal of enableLazyInitialization from hibernate-enhance-maven-plugin does not work as expected. @Basic(fetch = FetchType.LAZY) annoteded prperties are no longer fetched lazily but eagerly. Tested in version 6.2.9

PR #5434, opened 2022-10-20
HHH-15608 Error In Native Query where column is mapped twice

dreab8:
https://hibernate.atlassian.net/browse/HHH-15608


hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -996,9 +994,11 @@ else if ( batchSize > 1 ) {
 //					formnos[l] = -1;
 //					columns.add( colName );
 					cols[i] = colName;
-					aliases.add( selectable.getAlias( dialect, prop.getValue().getTable() ) );
+					final String columnAlias = selectable.getAlias( dialect, prop.getValue().getTable() );
+					if ( prop.isSelectable() && !aliases.contains( columnAlias ) ) {
```
beikov, line 998: I guess you should add something similar for `formulaAliases` in the if-block above

beikov (commented):
LGTM, but the new test seems to fail on MySQL

PR #5438, opened 2022-10-21
HHH-15618 Accept TypedParameterValue for procedure

quaff:
Fix https://github.com/spring-projects/spring-data-jpa/issues/2544

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/procedure/internal/ProcedureCallImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/procedure/internal/ProcedureCallImpl.java
@@ -35,6 +35,7 @@ import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.internal.util.StringHelper;
+import org.hibernate.jpa.TypedParameterValue;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.procedure.NoSuchParameterException;
 import org.hibernate.procedure.ParameterRegistration;
@@ -66,6 +67,7 @@ import org.jboss.logging.Logger;
  * Standard implementation of {@link org.hibernate.procedure.ProcedureCall}
  *
  * @author Steve Ebersole
+ * @author Yanming Zhou
  */
 public class ProcedureCallImpl<R>
 		extends AbstractProducedQuery<R>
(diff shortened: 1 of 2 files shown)


quaff:
Could anyone review this?

beikov:
I don't see a reason for this PR. There is an explicit API that can be used to bind parameters in a typed fashion. Also, `TypedParameterValue` is usually handled on the `Query` level, so if we do this at all, it should be handled in `ProcedureCallImpl`. Also, it's nice that you provide an implementation for 5.6, but we also need a test and possible fix for the `main` branch.

quaff:
> I don't see a reason for this PR. There is an explicit API that can be used to bind parameters in a typed fashion. Also, `TypedParameterValue` is usually handled on the `Query` level, so if we do this at all, it should be handled in `ProcedureCallImpl`. Also, it's nice that you provide an implementation for 5.6, but we also need a test and possible fix for the `main` branch.

@beikov Updated as you suggested.

gregturn:
Adopting Hibernate's `TypedParameterValue` wrapper is causing things to break in other places as that wrapper type percolates through our code. A big one our community runs into is stored procedures. 

With @quaff having offered a patch in the location your cited, what are the odds you could merge this patch?

gregturn:
Regarding Hibernate 6.1, it appears to already be handled way up the hierarchy in `AbstractCommonQueryContract`.

```java
	@Override
	public CommonQueryContract setParameter(int position, Object value) {
		if ( value instanceof TypedParameterValue ) {
			final TypedParameterValue<Object> typedValue = (TypedParameterValue<Object>) value;
			final BindableType<Object> type = typedValue.getType();
			if ( type != null ) {
				return setParameter( position, typedValue.getValue(), type );
			}
			else {
				return setParameter( position, typedValue.getValue(), typedValue.getTypeReference() );
			}
		}

		final QueryParameterImplementor<?> param = getParameterMetadata().getQueryParameter( position );

		if ( param == null ) {
			throw new IllegalArgumentException( "Positional parameter [" + position + "] is not registered with this procedure call" );
		}

		if ( param.allowsMultiValuedBinding() ) {
			final BindableType<?> hibernateType = param.getHibernateType();
			if ( hibernateType == null || isInstance( hibernateType, value ) ) {
				if ( value instanceof Collection && !isRegisteredAsBasicType( value.getClass() ) ) {
					//noinspection rawtypes,unchecked
					return setParameterList( param, (Collection) value );
				}
			}
		}

		locateBinding( position ).setBindValue( value, resolveJdbcParameterTypeIfNecessary() );
		return this;
	}
```

beikov:
>Updated as you suggested.

@quaff the changes look good, but please also create a PR against the main branch. Chances are good that we will merge this also for 5.6 @gregturn

quaff:
Main branch(v6) doesn't need change if I understand correctly as @gregturn
said.

Christian Beikov ***@***.***> 于2022年12月21日周三 06:06写道：

beikov:
But it needs at least a test case to verify it works correctly.

quaff:
> But it needs at least a test case to verify it works correctly.

@beikov I've created https://github.com/hibernate/hibernate-orm/pull/5819.

beikov:
Thanks!

PR #5449, opened 2022-10-26
Reproducer for HHH-15634 Lazy basic property does not get updated on change

yrodiere:
https://hibernate.atlassian.net/browse/HHH-15634

Reproducer only, no fix.


Sanne:
Superseded by #5461 - thanks!

PR #5450, opened 2022-10-26
HHH-15632, HHH-15633, HHH-15635 Various perf improvements

beikov:
https://hibernate.atlassian.net/browse/HHH-15632
https://hibernate.atlassian.net/browse/HHH-15633
https://hibernate.atlassian.net/browse/HHH-15635

PR #5452, opened 2022-10-26
Forward-porting #5450 to main

Sanne:
https://hibernate.atlassian.net/browse/HHH-15632
https://hibernate.atlassian.net/browse/HHH-15633
https://hibernate.atlassian.net/browse/HHH-15635

The second patch didn't apply cleanly, hopefully I didn't screw it up as I edited from commandline. Letting CI figure it out.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/EntityKey.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/EntityKey.java
@@ -10,7 +10,6 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
-import java.util.Objects;
 
 import org.hibernate.AssertionFailure;
 import org.hibernate.persister.entity.EntityPersister;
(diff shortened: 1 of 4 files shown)


beikov:
Thanks for porting :)

Sanne:
@beikov I see you marked JIRA resolved, but shouldn't these be set as fixed for version 6.2 too?

beikov:
We agreed that there is no need for also setting 6.2.0 since 6.1.5 should imply that.

Sanne:
Really? That seems a big change to established processes and I don't think
it can work well.

With 6.2 already branched,  how does one distinguish a patch that has only
been pushed to 6.1 and forgot to be applied to main (like you had done) ?

Please ensure that the Jira changelogs are always consistent with the
content of the branch.

beikov:
It's unfortunate that we even maintain two branches for so long.

>With 6.2 already branched,  how does one distinguish a patch that has only been pushed to 6.1 and forgot to be applied to main (like you had done) ?

We only resolve JIRA issues when we are done with them. I usually assign 6.1.x if I want it backported and only resolve it once I merged to both branches. IMO it doesn't matter how we do this, we can always forget to do something.

>Please ensure that the Jira changelogs are always consistent with the content of the branch.

That's what we are trying :)

PR #5454, opened 2022-10-27
HHH-12930 much more flexible handling of associations on non-primary keys

gavinking:
So this was *really* tricky to get right, but now:

- we handle embeddables forming part of the unique key in the target entity
- error messages are *much* improved (the previous error was incomprehensible)
- the code is *much* more robust and easier to understand
- I have left *copious* comments in the code explaining WFT is going on, because it was previously really difficult to understand


gavinking:
Possibly-related issue reports include:

- https://hibernate.atlassian.net/browse/HHH-15355
- https://hibernate.atlassian.net/browse/HHH-13054
- https://hibernate.atlassian.net/browse/HHH-14940
- https://hibernate.atlassian.net/browse/HHH-15277
- https://hibernate.atlassian.net/browse/HHH-15627
- https://hibernate.atlassian.net/browse/HHH-14687

I need to go through and see which (if any) of these issues were fixed by this PR.

But first I need to merge it b/c I'm getting into a mess of conflicts with https://github.com/hibernate/hibernate-orm/pull/5447.

PR #5461, opened 2022-10-27
HHH-15634 Lazy basic property does not get updated on change

dreab8:
https://hibernate.atlassian.net/browse/HHH-15634

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
@@ -3841,10 +3841,28 @@ public abstract class AbstractEntityPersister
 		if ( entry == null && !isMutable() ) {
 			throw new IllegalStateException( "Updating immutable entity that is not in session yet" );
 		}
-		if ( entityMetamodel.isDynamicUpdate() && dirtyFields != null ) {
+		if ( dirtyFields != null && entityMetamodel.isDynamicUpdate() ) {
 			// We need to generate the UPDATE SQL when dynamic-update="true"
 			propsToUpdate = getPropertiesToUpdate( dirtyFields, hasDirtyCollection );
 			// don't need to check laziness (dirty checking algorithm handles that)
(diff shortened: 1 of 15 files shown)


hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -3852,6 +3852,24 @@ public void update(
...
 		}
+		else if ( dirtyFields != null && hasUninitializedLazyProperties( object ) && hasLazyDirtyFields( dirtyFields ) ) {
+			// We need to generate the UPDATE SQL when there are dirty lazy fields
+			propsToUpdate = getPropertiesToUpdate( dirtyFields, hasDirtyCollection );
+			// don't need to check laziness (dirty checking algorithm handles that)
+			final boolean[] propertyLaziness = getPropertyLaziness();
+			// we add also all the non lazy properties because dynamic update is false
+			for ( int i = 0; i < propertyLaziness.length; i++ ) {
+				if ( propertyLaziness[i] == false ) {
+					propsToUpdate[i] = true;
```
beikov, line 3863: We should probably have a quick discussion about this next week with @sebersole to hear his thoughts on this topic.

beikov (commented):
Seems that you accidently also fixed https://hibernate.atlassian.net/browse/HHH-10747 since a failure expected test `org.hibernate.orm.test.bytecode.enhancement.lazy.LazyLoadingByEnhancerSetterTest` is now running through 😄

yrodiere (approved):
I ran the reproducer in Quarkus (https://github.com/quarkusio/quarkus/pull/28881) against Hibernate ORM 5.6 compiled with a backport of this patch (#5465), and I confirm it fixes all issues.

Thanks a lot @dreab8!

dreab8:
@beikov gave me a huge help :) by the way we have to check that the call of the `InlineDirtyCheckerEqualsHelper` does not create any issue

yrodiere:
Then thanks to @beikov too :)

PR #5463, opened 2022-10-27
Fix incorrectly reported deprecated setting

biergit:
The `javax.persistence.sharedCache.mode` setting reported here is not set by the user but [by Hibernate itself](https://github.com/hibernate/hibernate-orm/blob/ba48130c3f58acf64fa7069205f67a68eeb747ba/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java#L610) . User settings would already be reported [here](https://github.com/hibernate/hibernate-orm/blob/ba48130c3f58acf64fa7069205f67a68eeb747ba/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/EntityManagerFactoryBuilderImpl.java#L606).


beikov:
This was fixed via https://github.com/hibernate/hibernate-orm/pull/5752

PR #5474, opened 2022-10-31
HHH-15644 Add test-case for issue

theigl:
Reproducer for https://hibernate.atlassian.net/browse/HHH-15644


dreab8:
Thanks @theigl for the test, I have included it in https://github.com/hibernate/hibernate-orm/pull/5537

PR #5481, opened 2022-10-31
HHH-1661 throw when merge() applied to a definitely-removed instance 

jrenaat:
https://hibernate.atlassian.net/browse/HHH-1661

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultMergeEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultMergeEventListener.java
@@ -405,17 +405,26 @@ public class DefaultMergeEventListener
 		final Object result = source.getLoadQueryInfluencers().fromInternalFetchProfile(
 				CascadingFetchProfile.MERGE,
 				() -> source.get( entityName, clonedIdentifier )
-
 		);
-		if ( result == null ) {
-			//TODO: we should throw an exception if we really *know* for sure
-			//      that this is a detached instance, rather than just assuming
-			//throw new StaleObjectStateException(entityName, id);
 
+		if ( result == null ) {
+			LOG.trace( "Detached instance not found in database" );
 			// we got here because we assumed that an instance
-			// with an assigned id was detached, when it was
-			// really persistent
-			entityIsTransient( event, clonedIdentifier, copyCache );
+			// with an assigned id and no version was detached,
+			// when it was really transient (or deleted)
+			final Boolean knownTransient = persister.isTransient( entity, source );
+			if ( knownTransient == Boolean.FALSE ) {
+				// we know for sure it's detached (generated id
+				// or a version property), and so the instance
+				// must have been deleted by another transaction
+				throw new StaleObjectStateException( entityName, id );
+			}
+			else {
+				// we know for sure it's transient, or we just
+				// don't have information (assigned id and no
+				// version property) so keep assuming transient
+				entityIsTransient( event, clonedIdentifier, copyCache );
+			}
 		}
 		else {
 			// before cascade!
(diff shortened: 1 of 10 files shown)


migration-guide.adoc
```diff
@@ -37,4 +37,7 @@ To retain backwards compatibility, configure the setting `hibernate.type.preferr
 Previously, since Hibernate 6.0, columns of type `number` with scale 0 on Oracle were interpreted as `boolean`, `tinyint`, `smallint`, `int`, or `bigint`,
 depending on the precision.
 
-Now, columns of type `number` with scale 0 are interpreted as `int` or `bigint` depending on the precision.
\ No newline at end of file
+Now, columns of type `number` with scale 0 are interpreted as `int` or `bigint` depending on the precision.
+
+=== Merging stale objects
+An OptimisticLockException will now be thrown when a previously deleted instance of an entity that has optimistic locking enabled, is merged by another session. Note however that, at the moment, for this to work correctly with primitive version types (int, short and long) these have to be initialized to 0. For the primitive wrapper types this will work as-is.
```
beikov, line 43:
```suggestion
=== Merge versioned entity when row is deleted

So far, merging detached entities with a version assigned, resulted in an insert when the row was not found. Since this is intuitive and goes against the principle of optimistic locking, this was fixed to now throw an `OptimisticLockException` if the detached entity has a version assigned that is non-default value. For object type versions, any non-null value is considered non-default. For primitive type versions, any value other than the JVM default value (`0`) is considered non-default.
```

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultMergeEventListener.java
```diff
@@ -318,12 +318,26 @@ protected void entityIsDetached(MergeEvent event, MergeContext copyCache) {
...
+				Object versionValue = persister.getValue( entity, persister.getVersionProperty() );
+				if ( versionValue != null ) {
+					boolean isPrimitiveVersion = persister.getVersionMapping()
+							.getVersionAttribute()
+							.getPropertyAccess()
+							.getGetter()
+							.getReturnTypeClass()
+							.isPrimitive();
+					// IF primitive AND defaultvalue == versionvalue 0 => DON'T throw
+					// TODO evaluate if the primitive version types can be seeded to 1 (see also StaleObjectMergeTest.testStalePrimitiveAndWrapperVersionEntityMerged()
```
beikov, line 335: Please remove the todo, as it is not actionable here. We need to discuss changing the seed value separately.

migration-guide.adoc
gavinking: Um, there's are some incorrect comma usage in both the original and the suggested change. You can't put a comma between the subject and its verb, I suppose. Also, did you mean "unintuitive"? Also, did you mean "**a** non-default value".

beikov: I hate commas, so yeah, probably I misplaced one somewhere. Please suggest the correct version.

gavinking: Previously, merging a detached entity with an assigned version resulted in a SQL insert whenever there was no matching row in the database (for example, if the object had been deleted in another transaction). This behavior was unexpected and violated the rules of optimistic locking. 

An `OptimisticLockException` is now thrown in this situation if the version of the detached entity has any value other than its Java default value:

- For object types, `null` is the default value. 
- For primitive types, zero is the default value.

Sanne:
Some conflicts materialized - sorry :) @jrenaat could you rebase it?

beikov:
I have to adapt the PR a bit after the discussion today with Steve. Do not merge yet please, I'll take care of the rest.

gavinking (changes requested):
I think this is on the right track, but as per my comment, I don't think it's quite perfect.

Also please see what I wrote here: https://hibernate.atlassian.net/browse/HHH-1661?focusedCommentId=115352

gavinking:
This is now ready to go.

However, this is a breaking change, and I wonder if we should be targeting H7?

gavinking (approved):

gavinking:
This is good with the latest changes from @beikov. I've squashed everything down to one commit.

PR #5482, opened 2022-10-31
HHH-15175 - Remove support for database versions

jrenaat:
Note on minimum supported versions in the migration doc.


migration-guide.adoc
```diff
@@ -37,4 +37,54 @@ To retain backwards compatibility, configure the setting `hibernate.type.preferr
...
+The minimum supported dialect versions are as follows:
+
+|===
+|Dialect |Minimum supported version
+
+|MySQL
+|5.7
+
+|SQL Server
+|10.0
```
beikov, line 55: Let's name it like people know it, which is the 2008 version.

PR #5489, opened 2022-11-01
HHH-15652 add missing Session.remove(entityName, object) method

gavinking:
Since we deprecated `delete()` there's been no blessed way to do this.


gavinking:
cc @sebersole.

PR #5494, opened 2022-11-02
HHH-1716 - Added test for issue, and for identifier rollback setting

jrenaat:


gavinking:
I have closed the original issue, and also, FTR, "identifier rollback" has nothing to do with tx rollback. We make *no* guarantees about the state of entities after a tx rolls back, and the rule is that you must discard them and the session.

PR #5500, opened 2022-11-03
Correction to the native sql queries section

jrenaat:
The result transformer-part that is mentioned in the doc has disappeared from the test, rendering it effectively identical to the `sql-hibernate-entity-associations-query-many-to-one-join-example` test just above it, so I suggest removing the duplicate.

### DIFF
--- a/documentation/src/test/java/org/hibernate/userguide/sql/SQLTest.java
+++ b/documentation/src/test/java/org/hibernate/userguide/sql/SQLTest.java
@@ -18,6 +18,7 @@ import org.hibernate.dialect.PostgreSQLDialect;
 import org.hibernate.orm.test.jpa.BaseEntityManagerFunctionalTestCase;
 import org.hibernate.loader.NonUniqueDiscoveredSqlAliasException;
 import org.hibernate.proxy.HibernateProxy;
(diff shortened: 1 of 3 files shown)


documentation/src/test/java/org/hibernate/userguide/sql/SQLTest.java
```diff
@@ -369,27 +369,6 @@ public void test_sql_hibernate_entity_associations_query_many_to_one_join_exampl
...
-		doInJPA(this::entityManagerFactory, entityManager -> {
-			Session session = entityManager.unwrap(Session.class);
-			//tag::sql-hibernate-entity-associations-query-many-to-one-join-result-transformer-example[]
-			List<Phone> phones = session.createNativeQuery(
-				"SELECT {ph.*}, {pr.*} " +
-				"FROM Phone ph " +
-				"JOIN Person pr ON ph.person_id = pr.id")
-			.addEntity("ph", Phone.class)
-			.addJoin("pr", "ph.person")
-			.list();
```
beikov, line 383: I think I'd prefer if you rather fix the example.

```
			q.setTupleTransformer( (tuple, alias) -> (Phone) tuple[0] );
```

PR #5505, opened 2022-11-04
HHH-15666 When a reference entry is found in the Second Level Cache there is not need to set its values 

dreab8:
https://hibernate.atlassian.net/browse/HHH-15666


beikov (approved):
Awesome! Please also backport :)

PR #5506, opened 2022-11-04
HHH-15665 - Mariadb is missing identifier quote on SEQUENCE QUERY

rusher:
MariaDB is missing identifier quote on SEQUENCE QUERY:
https://hibernate.atlassian.net/browse/HHH-15665


beikov:
Closing as this was fixed already

PR #5508, opened 2022-11-04
HHH-15670 Implement RequiresDialectFeature.reverse()

marschall:
- implement RequiresDialectFeature.reverse()
- remove not-feature checks to reverse = true


beikov:
Thanks!

PR #5509, opened 2022-11-04
HHH-15672 introduce Generated(UPDATE) for properties only generated on update

gavinking:
This "just" adds a new value to the enum `GenerationTiming`.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/annotations/DatabaseTimestampsColumnTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/annotations/DatabaseTimestampsColumnTest.java
@@ -1,0 +1,138 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.orm.test.annotations;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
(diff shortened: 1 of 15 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/annotations/DatabaseTimestampsColumnTest.java
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        public Date getTimestamp() {
+            return timestamp;
+        }
+    }
+
+    @ValueGenerationType(generatedBy = TimestampValueGeneration.class)
+    @Retention(RetentionPolicy.RUNTIME)
+    public @interface Timestamp { GenerationTime value(); }
```
beikov, line 79: I guess you could use `@CurrentTimestamp` instead of defining a custom annotation?

beikov (commented):
LGTM except for minor things. I guess that you will have to implement `supportsStandardCurrentTimestampFunction()` to return false on some `Dialect` classes though, not sure anymore which one didn't support this.

hibernate-core/src/test/java/org/hibernate/orm/test/annotations/DatabaseTimestampsColumnTest.java
gavinking: Well of course. But it's a test.

gavinking:
> I guess that you will have to implement `supportsStandardCurrentTimestampFunction()` to return false on some Dialect classes though, not sure anymore which one didn't support this.

Probably, but I need to see some tests fail to know if there really are any.

gavinking:
I've been leaving this sitting here because I didn't want to step on @sebersole's work on insert stuff, but this commit really only touches 3 lines of `AbstractEntityPersister`, so perhaps I should just merge it.

PR #5515, opened 2022-11-07
HHH-15680, HHH-15681, HHH-15682, HHH-15683, HHH-15684 lock options and lock scopes

gavinking:
- deprecate `LockRequest` and `buildLockRequest()`, and have `lock()` accept `LockOptions`
- `LockOptions.scope` should be a `PessimisticLockScope`, not `boolean`
- static final instances of `LockOptions` are mutable (!)
- also a bunch of Javadoc improvement for `LockOptions` and friends
- also add some convenience constructors to `LockOptions`
- deprecate `SelectionQuery.setAliasSpecificLockMode()` in favor of `setLockMode()` overload
- clean up the handling and documentation of `setLockOptions()` for queries, with a change to semantics of `LockOptions.overlay()`

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/LockMode.java
+++ b/hibernate-core/src/main/java/org/hibernate/LockMode.java
@@ -24,14 +24,15 @@ import jakarta.persistence.LockModeType;
  */
 public enum LockMode {
 	/**
-	 * No lock required. If an object is requested with this lock
-	 * mode, a {@link #READ} lock will be obtained if it is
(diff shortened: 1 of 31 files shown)


gavinking:
So there's a really important fix in this work, and lots of improvements.

If anyone wants to step up to defend `LockRequest` or `SelectionQuery.setAliasSpecificLockMode()` then I can easily just remove the deprecation tags.

I think it's safe to merge it.

PR #5525, opened 2022-11-09
HHH-14397 : update SQLServerDialect param list size

nivruth-naka:
This change updates the param list size which is currently set to 2100 and the allowed limit seems to be much higher than that as mentioned in the ticket and also [Microsoft documentation on error 8632](https://learn.microsoft.com/en-US/sql/relational-databases/errors-events/mssqlserver-8632-database-engine-error?view=sql-server-ver16)


beikov:
Seems like `MaxInExpressionParameterPaddingTest` is failing on SQL Server:

```
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: The incoming request has too many parameters. The server supports a maximum of 2100 parameters. Reduce the number of parameters and resend the request.
```

Seems that reference is not correct?

nivruth-naka:
My guess is the driver has the wrong limit on the parameter list size. Opened a [discussion](https://github.com/microsoft/mssql-jdbc/issues/1980) to check with mssql-jdbc contributors

beikov:
I'll close this PR now like the JIRA issue since the limit of 2100 parameters seems to be coming from the server itself.

PR #5527, opened 2022-11-09
HHH-15693 Introduce a fast-path access for ClassLoaderService being retrieved from ServiceRegistry

Sanne:
https://hibernate.atlassian.net/browse/HHH-15693

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/service/internal/AbstractServiceRegistryImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/service/internal/AbstractServiceRegistryImpl.java
@@ -17,6 +17,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 
 import org.hibernate.boot.registry.BootstrapServiceRegistry;
+import org.hibernate.boot.registry.classloading.spi.ClassLoaderService;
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/service/internal/AbstractServiceRegistryImpl.java
```diff
@@ -184,6 +185,14 @@ private void registerAlternate(Class<?> alternate, Class<?> target) {
 
 	@Override
 	public <R extends Service> R getService(Class<R> serviceRole) {
+		//Fast-path for ClassLoaderService as it's extremely hot during bootstrap
+		//(and after bootstrap service loading performance is less interesting as it's
+		//ideally being cached by long term consumers)
+		if ( ClassLoaderService.class.equals( serviceRole ) ) {
```
andrei-ivanov, line 191: I think you can even use `==` instead of `equals` here

Sanne: Good point, thanks!

Sanne: https://github.com/hibernate/hibernate-orm/pull/5577

PR #5540, opened 2022-11-10
HHH-15703 Append remainder of SQL fragment after last parameter

papegaaij:


beikov:
Very much appreciated, thanks!

PR #5543, opened 2022-11-10
HHH-15679 Use OffsetTime on JDBC level for TIME WITH TIME ZONE type

beikov:
https://hibernate.atlassian.net/browse/HHH-15679

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/SqlTypes.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/SqlTypes.java
@@ -593,6 +593,7 @@ public class SqlTypes {
 	public static boolean isTemporalType(int typeCode) {
 		return typeCode == DATE
 			|| typeCode == TIME
+			|| typeCode == TIME_WITH_TIMEZONE
 			|| typeCode == TIMESTAMP
 			|| typeCode == TIMESTAMP_WITH_TIMEZONE
 			|| typeCode == TIMESTAMP_UTC;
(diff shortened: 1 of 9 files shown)


gavinking:
-1. The SQL type `time with time zone` is just conceptually broken and nonsense.

gavinking:
See the discussion here:

https://www.postgresql.org/docs/current/datatype-datetime.html#id-1.5.7.13.20.2

I used to have a link to this in the code, but looks like someone deleted it.

gavinking:
I mean, do we even know of a database which supports this type? Oracle doesn't, Postgres doesn't (it just converts it to UTC), MySQL doesn't, and doesn't look like DB2 does either.

gavinking:
> See the discussion here:
> 
> https://www.postgresql.org/docs/current/datatype-datetime.html#id-1.5.7.13.20.2
> 
> I used to have a link to this in the code, but looks like someone deleted it.

Another link: https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_timetz

beikov:
You are right, the mapping as implemented here is broken. I don't know what got into me 😆 

>But even if it's just an offset, what is the point of that? Just convert it to UTC! What possible value could there be to storing a time with an offset?

I think that the use case of the user is a recurring event at a specific zone offset, but even then I don't think that retaining the offset is very helpful. I guess the only possibly interesting type (which doesn't exist) would be a `ZonedTime`, but let's not waste our time. I'll close this PR.

PR #5544, opened 2022-11-10
HHH-15679 proposed fix to OffsetTime handling

gavinking:
The idea is: convert all `OffsetTimes` to the system offset before sending them on.

The justification for that move is that when we _read them back_, we give them the system offset.

@beikov @yrodiere WDYT?

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/OffsetTimeJavaType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/OffsetTimeJavaType.java
@@ -11,10 +11,10 @@ import java.sql.Timestamp;
 import java.sql.Types;
 import java.time.Instant;
 import java.time.LocalDate;
+import java.time.LocalTime;
 import java.time.OffsetDateTime;
 import java.time.OffsetTime;
 import java.time.ZoneOffset;
-import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.Calendar;
 import java.util.Date;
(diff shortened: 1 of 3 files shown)


gavinking:
@yrodiere I'm having a hard time knowing what to make of this code in the test:

```java
	@Override
	protected OffsetTime getExpectedPropertyValueAfterHibernateRead() {
		// For some reason, the offset is not stored, so the restored values use the offset from the default JVM timezone.
		if ( TimeAsTimestampRemappingH2Dialect.class.equals( getRemappingDialectClass() ) ) {
			return getOriginalPropertyValue().withOffsetSameLocal( OffsetDateTime.now().getOffset() );
		}
		else {
			// When storing time as java.sql.Time, we only get second precision (not nanosecond)
			return getOriginalPropertyValue().withNano( 0 ).withOffsetSameLocal( OffsetDateTime.now().getOffset() );
		}
	}
```

`withOffsetSameLocal()` gives you back a different time. The test seems to be asserting the the round-trip to the database *does* change the time. Which I would have thought is not what we want.

gavinking:
UPDATE: I made two refinements of the initial proposal:

1. Account for the "jdbc timezone", i.e. `hibernate.jdbc.time_zone`, a thing I really hate but was easy enough to accommodate. All I do is:
    - when writing, convert to the jdbc timezone, if it is set, and otherwise convert to the system timezone
    - when reading, interpret the time in the jdbc timezone, if it is set, or in the system timezone if it is not set, and then finally convert to the system timezone (which is a noop if there is no jdbc timezone set).
2. Changed the `OffsetTimeTest` so that it matches the semantics I'm proposing. Apparently it was written to explicitly assert that we throw away the offset and treat `OffsetTime`s as `LocalTime`s, which is the previous behavior I've been calling a "bug", but was apparently by design.

Thoughts?

gavinking:
(Note: the failure on Sybase doesn't look like it's related to my change. I'll rerun that job in a bit.)

gavinking:
We need to decide what to do with this. It works, but it's a change to behavior.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/OffsetTimeJavaType.java
```diff
@@ -76,31 +76,44 @@ public <X> X unwrap(OffsetTime offsetTime, Class<X> type, WrapperOptions options
...
 		}
 
+		// for java.time types, we assume that the JDBC timezone, if any, is ignored
+
 		if ( OffsetTime.class.isAssignableFrom( type ) ) {
 			return (X) offsetTime;
 		}
 
+		if ( LocalTime.class.isAssignableFrom( type ) ) {
+			return (X) offsetTime.withOffsetSameInstant( getCurrentSystemOffset() ).toLocalTime();
```
beikov, line 86: Can you add a comment on why you are not using `getCurrentJdbcOffset` here? Same for the `wrap` method.

beikov (commented):
Overall, I like the change and I think it was just buggy before. I don't think anyone could have reasonably made use of `OffsetTime` before, which is why I think this breaking change is fine, but lets hear what others think.

hibernate-core/src/main/java/org/hibernate/type/descriptor/java/OffsetTimeJavaType.java
gavinking: The comment is above:

    // for java.time types, we assume that the JDBC timezone, if any, is ignored

this sort-of doesn't make much sense, unless you know what I have planned:

https://github.com/hibernate/hibernate-orm/pull/5546

beikov: Aha, yeah that makes sense then, thanks!

yrodiere:
> @yrodiere I'm having a hard time knowing what to make of this code in the test:
> 
> ```java
> 	@Override
> 	protected OffsetTime getExpectedPropertyValueAfterHibernateRead() {
> 		// For some reason, the offset is not stored, so the restored values use the offset from the default JVM timezone.
> 		if ( TimeAsTimestampRemappingH2Dialect.class.equals( getRemappingDialectClass() ) ) {
> 			return getOriginalPropertyValue().withOffsetSameLocal( OffsetDateTime.now().getOffset() );
> 		}
> 		else {
> 			// When storing time as java.sql.Time, we only get second precision (not nanosecond)
> 			return getOriginalPropertyValue().withNano( 0 ).withOffsetSameLocal( OffsetDateTime.now().getOffset() );
> 		}
> 	}
> ```
> 
> `withOffsetSameLocal()` gives you back a different time. The test seems to be asserting the the round-trip to the database _does_ change the time. Which I would have thought is not what we want.

Yes, those were non-regression tests back when the code handling date/time types was in flux and I wanted to make sure we weren't breaking more than we were fixing. It's totally possible that the previous behavior was buggy.

There are some cases where ORM was storing data incorrectly (wrong offset, ...), but restoring the correct data upon reading (same error, but in the opposite direction). Changing such a behavior, while correct, would effectively break applications with pre-existing data. So, not something easy to fix.

If you think your patch won't lead to data loss/corruption, then that's fine. If not, well... up to y'all: you'll be the ones to deal with the fallout :)

gavinking:
@yrodiere OK, thanks. Looks like I ended up just changing the test to use `withOffsetSameInstant()` instead of `withOffsetSameLocal()`, which seems like it's the more correct condition.

PR #5546, opened 2022-11-11
[on hold] experimental new time/datetime handling

gavinking:
The idea here is to get away from `java.sql.Time`, `java.sql.Timestamp` and use `java.time` types all the way to JDBC.

Let's see how many JDBC drivers this breaks.


gavinking:
Well, so this was a fun and enlightening (though time-wasting) exercise, and this approach works pretty well on MySQL and Postgres.

But across other DBs the support for the `java.time` types is still much too uneven for it to be worth the effort. 

So this is a no go.

gavinking:
This was implemented a while ago by Steve.

PR #5547, opened 2022-11-11
HHH-15704 - Fix regression in StandardForeignKeyExporter 

jrenaat:
and add test case for this issue


beikov (approved):
Great work!

PR #5561, opened 2022-11-14
proposed new API for cleaning catalogs and schemas

gavinking:
Here's the bare bones of an API that would make it easy to clean up after tests.

Obviously these APIs could grow a lot more interesting operations over time.

WDYT?


hibernate-core/src/main/java/org/hibernate/dialect/cleaner/JdbcConnectionContext.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.io.InputStream;
+import java.sql.Connection;
+import java.sql.Driver;
+import java.util.Properties;
+
+/**
+ * @author Christian Beikov
+ */
+public final class JdbcConnectionContext {
```
beikov, line 19: IMO this should stay in the `hibernate-testing` module.

hibernate-core/src/main/java/org/hibernate/dialect/cleaner/DatabaseCleanerContext.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.dialect.cleaner;
+
+/**
+ * @author Christian Beikov
+ */
+public final class DatabaseCleanerContext {
```
beikov, line 12: IMO this should stay in the `hibernate-testing` module.

hibernate-core/src/main/java/org/hibernate/dialect/cleaner/DatabaseCleaner.java
```diff
@@ -0,0 +1,42 @@
...
+			JdbcConnectionContext.work( cleaner::clearAllSchemas );
+		}
+	}
+
+	static void clearData() {
+		final DatabaseCleaner cleaner = DatabaseCleanerContext.CLEANER;
+		if ( cleaner != null ) {
+			JdbcConnectionContext.work( cleaner::clearAllData );
+		}
+	}
```
beikov, line 28: Dunno where these two methods are used, but they shouldn't be part of the API since they are using `JdbcConnectionContext`

beikov (commented):
Looks good, though please don't move the `JdbcConnectionContext` and `DatabaseCleanerContext` to the API.

gavinking:
> Looks good, though please don't move the `JdbcConnectionContext` and `DatabaseCleanerContext` to the API.

Ah, right, yes, thanks. Fixed.

beikov (approved):
LGTM

gavinking:
Can this SQL possibly be correct:

```java
								"SELECT tbl.owner || '.\"' || tbl.table_name || '\"', c.constraint_name FROM (" +
										"SELECT owner, table_name " +
										"FROM all_tables " +
```

That `tbl` alias looks to be undefined.

Has this ever been tested?

beikov:
The alias is defined at the last line of the SQL statement right after the closing parenthesis. Yes, this code is tested as I mostly copied parts from Blaze-Persistence: https://github.com/Blazebit/blaze-persistence/tree/main/testsuite-base/jpa/src/main/java/com/blazebit/persistence/testsuite/base/jpa/cleaner

gavinking:
> The alias is defined at the last line of the SQL statement right after the closing parenthesis. Yes, this code is tested as I mostly copied parts from Blaze-Persistence: https://github.com/Blazebit/blaze-persistence/tree/main/testsuite-base/jpa/src/main/java/com/blazebit/persistence/testsuite/base/jpa/cleaner

Yeah, yeah, sorry, I figured it out.

gavinking:
So now I've fleshed out this API and moved it to its own package.

I'm a little bit conflicted about it, since one could reasonably ask what it even has to do with Hibernate? Why not just use JDBC for this.

However, Hibernate already has a connection provider, and knows the SQL dialect, so it's just really quite convenient to have it there.

beikov:
I agree that it is a nice convenience, just like DDL support.

PR #5585, opened 2022-11-18
HHH-15735 Fix JDBC Connection leak in ExtractionContext #getQueryResults

kennymacleod:
Fixed JDBC `Connection` leak. Simple fix, just add the `Connection` to the `try-with-resources` block.


gavinking:
In this case, the client is not responsible for closing that connection. It is closed deeper in the works.

PR #5590, opened 2022-11-21
HHH-15669 Fix test failures when using Oracle 21

maesenka:
Rings in Oracle polygons may be shifted depending on how it is processed. The equality test now takes this into account.

Add test to investigate st_within test failure.


beikov (approved):
LGTM. IMO this can be merged when tests succeed!

PR #5592, opened 2022-11-21
HHH-15711 ArgumentTypesValidator doesn't support enums in string functions

dreab8:
https://hibernate.atlassian.net/browse/HHH-15711


beikov (approved):
LGTM

PR #5597, opened 2022-11-22
Add back the MutableIdentifierGeneratorFactory

DavideD:
Fix [HHH-15746](https://hibernate.atlassian.net/browse/HHH-15746)

Hibernate Reactive needs to be able to register its own generators.
This interface was in ORM 5 and I'm just adding it back.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataBuilderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataBuilderImpl.java
@@ -64,6 +64,7 @@ import org.hibernate.engine.jdbc.connections.spi.MultiTenantConnectionProvider;
 import org.hibernate.engine.jdbc.spi.JdbcServices;
 import org.hibernate.id.factory.IdentifierGeneratorFactory;
 import org.hibernate.id.factory.internal.StandardIdentifierGeneratorFactory;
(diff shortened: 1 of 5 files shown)


Sanne:
I remember that the *mutable* aspect of it was problematic.

In ORM5 we had two possible layers: `MutableIdentifierGeneratorFactory` and `IdentifierGeneratorFactory` - both of them were a `Service`, so you could customize the implementation by changing the service initiator.

Are you sure you don't just need the plain version `IdentifierGeneratorFactory` (not the mutable one) ?

Sanne:
@DavideD funny coincidence :)
 https://github.com/Sanne/quarkus/blob/cce8938411814e53b6f9eb6fd6ba1509843e0db7/extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/service/QuarkusMutableIdentifierGeneratorFactory.java#L32-L33

DavideD:
> Are you sure you don't just need the plain version `IdentifierGeneratorFactory` (not the mutable one) ?

I need to check. The reality is that I've just copied what was in ORM 5.
I was wondering why `IdentiierGeneratorFactory` is not a service.

Sanne:
I'll try to promote the `IdentifierGeneratorFactory` to a `Service` so that it's overridable, let's see if we can live without the mutation capability.

Sanne:
replaced by #5600 5600

PR #5599, opened 2022-11-22
Small fixes and cleanups

beikov:
Here are a few fixes and cleanups, as well as preparation work for struct type support: https://hibernate.atlassian.net/browse/HHH-15327


gavinking (commented):
Why did you revert this?

beikov:
>Why did you revert this?

Because the requested `JavaType` from a selection might be different from the expression `JavaType`. I'll check again if I can drop this requirement, but I think that this is necessary for support struct types.

gavinking:
> > Why did you revert this?
> 
> Because the requested `JavaType` from a selection might be different from the expression `JavaType`. I'll check again if I can drop this requirement, but I think that this is necessary for support struct types.

IIRC that parameter was simply never used by any implementation of the method.

beikov:
>IIRC that parameter was simply never used by any implementation of the method.

Yes, but I started using it on my branch 😆

gavinking:
> > IIRC that parameter was simply never used by any implementation of the method.
> 
> Yes, but I started using it on my branch 😆

OK I see. 

The other reason I hated it, if I recall correctly, was it was causing warnings due to raw types or something. Since you added it back in, it would be nice if you could figure out how to make them go away.

beikov:
Seems I didn't need this anymore after all, so I remove the JavaType stuff again from this PR.

gavinking:
Nice.

PR #5600, opened 2022-11-23
HHH-15746 Allow replacing the StandardIdentifierGeneratorFactory by making it a Service again

Sanne:
https://hibernate.atlassian.net/browse/HHH-15746

Replaces #5597


DavideD (approved):
Seems to work for Hibernate Reactive.
I might find something later after fixing some tests, but I think we can merge it.

Sanne:
thanks!

PR #5605, opened 2022-11-24
Try switching to regular oracle images instead of full

beikov:


beikov:
Can't do that because certain spatial features in Oracle require Java to be installed in the DB container, which only the full flavor has.

beikov:
PR is running here: https://github.com/hibernate/hibernate-orm/pull/7930

PR #5612, opened 2022-11-25
HHH-15759 Entity with @CollectionTable and entity with relation to element collection table, cause generate wrong FK, if first entity key field lexicographically after collection element.

AntonBarkan:
Entity with @CollectionTable and entity with relation to element collection table, cause generate wrong FK, if first entity key field lexicographically after collection element.


beikov:
Can you please try your test case with the latest state on `main` but without your fix? We recently fixed something in this area: 70873f5d2fdaeefb5462c69bed595247d299aa40

dreab8:
Superseded by https://github.com/hibernate/hibernate-orm/pull/5739

PR #5613, opened 2022-11-25
HHH-15760 Fix all javadoc errors of type 'self-closing element not allowed' and other small html/syntax errors

mbladel:
https://hibernate.atlassian.net/browse/HHH-15760


gavinking:
nonono this is literally according to Oracle's Javadoc style guide!

(And it's what all the JDK Javadocs do.)

hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/IngresDialect.java
```diff
@@ -413,7 +413,7 @@ else if ( getVersion().isSameOrAfter( 9, 3 ) ) {
 	// lock acquisition support ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 	/**
-	 * <TT>FOR UPDATE</TT> only supported for cursors
+	 * {@code FOR UPDATE} only supported for cursors
```
gavinking, line 416: These sort of changes are good though :-)

mbladel:
> For reference: https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html

Most of the changes are `<p/>` tags that cause errors when built without the `Xdoclint:none` option being replace with `<p>` tags. This should be in line with the style guide you mentioned I beliebe and would solve those errors.

gavinking:
> > For reference: https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html
> 
> Most of the changes are `<p/>` tags that cause errors when built without the `Xdoclint:none` option being replaced with `<p>` tags. This should be in line with the style guide you mentioned I believe and would solve those errors.

Ooooh apologies, I misread the diff and thought you were changing `<p>`s to `<p/>`s, but actually you're doing the opposite, which is a change I approve of.

My bad, carry on...

beikov:
Thanks!

PR #5621, opened 2022-11-26
Revert "Disable dynamic sequence caching for Oracle"

beikov:
Reverts hibernate/hibernate-orm#5618


beikov:
Reverting because it doesn't work and also makes no sense for 6.1 anyway since we aren't testing against 21 on this branch

PR #5622, opened 2022-11-26
HHH-15759 test case

AntonBarkan:


beikov:
Ok, so it seems the issue you are observing is not yet fixed. Thanks for trying this out!

PR #5625, opened 2022-11-26
HHH-15765 Fix build failure by upgrading nu.studer.credentials to 3.0

mrotteveel:
Fixes [HHH-15765](https://hibernate.atlassian.net/browse/HHH-15765)


mrotteveel:
The build failure is caused by nu.studer:java-ordered-properties:1.0.1, which is used by nu.studer.credentials:2.1, missing from Maven Central (maybe it was previously source from jcenter/bintray?)

beikov:
Thanks!

PR #5626, opened 2022-11-26
gradle build is too noisy

gavinking:
Make it quieter. Also delete some obsolete stuff from before move to Java 11.


beikov:
Thanks

PR #5629, opened 2022-11-27
deprecate @Target which dupes functionality of JPA annotations

gavinking:
...and refresh some misc annotation javadoc


beikov:
Just a thought, but maybe it's useful for embeddables since `@Embedded` has no `target` member?

gavinking:
> Just a thought, but maybe it's useful for embeddables since `@Embedded` has no `target` member?

Yeah, I mentioned that case in the issue https://hibernate.atlassian.net/browse/HHH-15772

My conclusion was: it's still not useful. There's no way in which `@Target(FooImpl.class) Foo foo;` is better code than `FooImpl foo;`

PR #5630, opened 2022-11-28
HHH-15774 @CurrentTimestamp takes over from @Source

gavinking:
... and make value generation work better with version properties.


gavinking:
Screw it, I'm going to merge this: I've just spent hours wrestling with test failures, none of which had really anything to do with this work, and every time I rebased new and curious phenomena arose. Now, things are worse than when I started, but I know for sure the failures aren't due to my changes.

PR #5633, opened 2022-11-28
HHH-15725 Criteria API Expression.as adds cast even when the cast type is equal to the expression type

dreab8:
https://hibernate.atlassian.net/browse/HHH-15725


dreab8:
Superseded by https://github.com/hibernate/hibernate-orm/pull/5634

PR #5634, opened 2022-11-28
HHH-15725 Criteria API Expression.as adds cast even when the cast type is equal to the expression type

dreab8:
https://hibernate.atlassian.net/browse/HHH-15725

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AbstractSqmExpression.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AbstractSqmExpression.java
@@ -94,7 +94,7 @@ public abstract class AbstractSqmExpression<T> extends AbstractJpaSelection<T> i
 
 	@Override
 	public <X> SqmExpression<X> as(Class<X> type) {
-		return nodeBuilder().cast( this, type );
+		return new AsWrapperSqmExpression( this, type );
 	}
 
 	@Override
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
@@ -1,0 +1,51 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.query.sqm.tree.expression;
+
+import org.hibernate.query.sqm.SemanticQueryWalker;
+import org.hibernate.query.sqm.SqmExpressible;
+import org.hibernate.query.sqm.tree.SqmCopyContext;
+
+public class AsWrapperSqmExpression<T> extends AbstractSqmExpression<T> {
+	private final SqmExpression<T> expression;
+
+	public AsWrapperSqmExpression(SqmExpression<T> expression, Class<T> type) {
+		super(
+				expression.nodeBuilder().getTypeConfiguration().getBasicTypeForJavaType( type ),
+				expression.nodeBuilder()
+		);
+		this.expression = expression;
+	}
+
+	AsWrapperSqmExpression(SqmExpressible<T> type, SqmExpression<T> expression) {
+		super( type, expression.nodeBuilder() );
+		this.expression = expression;
+	}
+
+	@Override
+	public <X> X accept(SemanticQueryWalker<X> walker) {
+		return walker.visitAsWrapperExpression( this );
+	}
+
+	@Override
+	public void appendHqlString(StringBuilder sb) {
+		sb.append( "cast(" );
+		expression.appendHqlString( sb );
+		sb.append( " as " );
+		sb.append( getNodeType().getTypeName() );
+		sb.append( ')' );
+	}
+
+	@Override
+	public SqmExpression<T> copy(SqmCopyContext context) {
+		return new AsWrapperSqmExpression( getExpressible(), expression.copy( context ) );
+	}
+
+	public SqmExpression<?> getExpression() {
+		return expression;
+	}
+}
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/SemanticQueryWalker.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/SemanticQueryWalker.java
@@ -45,6 +45,7 @@ import org.hibernate.query.sqm.tree.domain.SqmPluralValuedSimplePath;
 import org.hibernate.query.sqm.tree.domain.SqmSetJoin;
 import org.hibernate.query.sqm.tree.domain.SqmSingularJoin;
 import org.hibernate.query.sqm.tree.domain.SqmTreatedPath;
+import org.hibernate.query.sqm.tree.expression.AsWrapperSqmExpression;
 import org.hibernate.query.sqm.tree.expression.JpaCriteriaParameter;
 import org.hibernate.query.sqm.tree.expression.SqmAny;
 import org.hibernate.query.sqm.tree.expression.SqmAnyDiscriminatorValue;
@@ -417,4 +418,6 @@ public interface SemanticQueryWalker<T> {
 	T visitMapEntryFunction(SqmMapEntryReference<?, ?> function);
 
 	T visitFullyQualifiedClass(Class<?> namedClass);
+
+	T visitAsWrapperExpression(AsWrapperSqmExpression expression);
 }
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmTreePrinter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmTreePrinter.java
@@ -34,6 +34,7 @@ import org.hibernate.query.sqm.tree.domain.SqmIndexAggregateFunction;
 import org.hibernate.query.sqm.tree.domain.SqmPluralPartJoin;
 import org.hibernate.query.sqm.tree.domain.SqmPluralValuedSimplePath;
 import org.hibernate.query.sqm.tree.domain.SqmTreatedPath;
+import org.hibernate.query.sqm.tree.expression.AsWrapperSqmExpression;
 import org.hibernate.query.sqm.tree.expression.JpaCriteriaParameter;
 import org.hibernate.query.sqm.tree.expression.SqmAny;
 import org.hibernate.query.sqm.tree.expression.SqmAnyDiscriminatorValue;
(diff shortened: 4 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
```diff
@@ -0,0 +1,47 @@
...
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.query.sqm.tree.expression;
+
+import org.hibernate.query.sqm.SemanticQueryWalker;
+import org.hibernate.query.sqm.SqmExpressible;
+import org.hibernate.query.sqm.tree.SqmCopyContext;
+
+public class AsWrapperSqmExpression<T> extends AbstractSqmExpression<T> {
```
beikov, line 13: I'd personally prefer naming this `TreatExpression` and use the same `treat(.. as type)` syntax that we use for `from` nodes. Wdyt @sebersole ?

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AbstractSqmExpression.java
```diff
@@ -101,7 +102,12 @@ public SqmExpression<String> asString() {
 
 	@Override
 	public <X> SqmExpression<X> as(Class<X> type) {
-		return nodeBuilder().cast(this, type);
+		if ( this instanceof SqlTypedMapping ) {
+			if ( ( (SqlTypedMapping) this ).getColumnDefinition() == null && getJavaType() == type ) {
+				return new AsWrapperSqmExpression(this, type);
+			}
+		}
+		return nodeBuilder().cast( this, type );
```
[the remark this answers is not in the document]
dreab8: yeap just noticed, changed to `	if ( this.getJavaTypeDescriptor().getJavaType().equals( type ) ) {
`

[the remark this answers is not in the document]
dreab8: but not sure it is 100% correct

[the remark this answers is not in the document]
beikov: I think it should be just `return new AsWrapperSqmExpression(this, type);`, unconditionally. In the `BaseSqmToSqlAstConverter` we can then decide to construct a cast function node or not, based on the actual `SqlTypedMapping`.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
sebersole: I think I'd prefer to avoid the confusion with TREAT

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AbstractSqmExpression.java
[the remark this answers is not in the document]
sebersole: I think this is reasonable suggestion

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
beikov: So do we want to support a HQL syntax for this as well? Or would you say this is a JPA Criteria only concept?

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AbstractSqmExpression.java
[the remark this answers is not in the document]
dreab8: done :)

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AbstractSqmExpression.java
```diff
@@ -101,7 +102,10 @@ public SqmExpression<String> asString() {
 
 	@Override
 	public <X> SqmExpression<X> as(Class<X> type) {
-		return nodeBuilder().cast(this, type);
+		if ( this.getJavaTypeDescriptor().getJavaType().equals( type ) ) {
+			return new AsWrapperSqmExpression( this, type );
+		}
+		return nodeBuilder().cast( this, type );
```
beikov, line 108:
```suggestion
		return new AsWrapperSqmExpression( this, type );
```

dreab8: yeap I forgot to change that part :P

beikov (commented):
LGTM. IMO we only need to discuss HQL support now, but after that, this can go in

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
sebersole: I'm not seeing any benefit to this in HQL.  Is there something I'm missing?

beikov: Here comes a use case that just came to mind. Projecting something that is of type `java.sql.Timestamp` in a constructor to `Instant` via e.g.:

```
select new MyDto(e.id, treat(e.created as Instant))
from MyEntity e
```

gavinking: Yeah, right, you might want to cast the "HQL" (i.e. Java) type to make HQL's type system happy, even though at the database level there is no work to do because the two Java types have already been erased down to the same SQL type.

It actually does make sense.

dreab8:
@beikov what about merging this an then you can open a github discussion for the HQL support?

beikov:
>what about merging this an then you can open a github discussion for the HQL support?

I think we should at least name the expression class like the syntax construct that we would potentially use for HQL, to avoid confusion later. I think we all agree that such a function is useful for some cases in HQL (e.g. treating Timestamp as Instant), the question is, how do we want to name it. Any ideas @sebersole?

I acknowledge that naming it `treat` might be a bad idea, so how about `treat_as`? Then we could name the expression class `TreatAsSqmExpression`

gavinking:
> I acknowledge that naming it treat might be a bad idea, so how about `treat_as`? Then we could name the expression class `TreatAsSqmExpression`

Please, please, please no more snake case in HQL. It’s the ugliest thing ever to see a mix of camel case and snake case in the same language. 

(We already have a problem there with some of the aggregate functions we added recently, but I’m accepting that because that’s the standard naming on SQL. But this isn’t a standard SQL construct so we should stick to our convention which is Java naming conventions.)

But anyway I don’t understand: how is this not just a flavor of `cast( as )`?

gavinking:
> But anyway I don’t understand: how is this not just a flavor of `cast( as )`?

So look, AFAICS, there's _no need_ for this operation in HQL. Why not? Because in HQL we already have a `cast()` function, which JPQL does not have. 

Yes, it's a bit weird that the criteria API has its own essentially-unspecified typecasting operation, and honestly I'm not clear on quite how Mike managed to sneak that one in there but _we_ should consider it to be a synonym for `cast()`, not a separate new thing.

If we have a typecast that only exists at the level of the HQL type system, and not in the database, then we can still express that using `cast( ... as .. )`, for example, in:

    select new MyDto(e.id, cast(e.created as Instant)) from MyEntity e

we should just be smart enough to detect that no typecast is needed, and optimize it away.

beikov:
>we should just be smart enough to detect that no typecast is needed, and optimize it away.

I think we can do this as well and that we should model this through the cast optimization you mentioned, but this is where @sebersole said, that it might be surprising for users when they write a cast explicitly in HQL, but it doesn't show up as cast in SQL.

gavinking:
> it might be surprising for users when they write a cast explicitly in HQL, but it doesn't show up as cast in SQL.

I mean, I dunno, it might be equally-surprising to see a timestamp getting unnecessarily cast to timestamp...

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	@Override
+	public void appendHqlString(StringBuilder sb) {
+		expression.appendHqlString( sb );
+	}
+
+	@Override
+	public SqmExpression<T> copy(SqmCopyContext context) {
+		return new AsWrapperSqmExpression<T>( getExpressible(), expression );
```
beikov, line 41:
```suggestion
		return new AsWrapperSqmExpression<T>( getExpressible(), expression.copy( context ) );
```

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	@Override
+	public <X> X accept(SemanticQueryWalker<X> walker) {
+		return walker.visitAsWrapperExpression( this );
+	}
+
+	@Override
+	public void appendHqlString(StringBuilder sb) {
+		expression.appendHqlString( sb );
```
beikov, line 36: Until we come up with a proper name for this in HQL I think this should render to a `cast`.

```suggestion
		sb.append( "cast(" );
		expression.appendHqlString( sb );
		sb.append( " as " );
		sb.append( type.getJavaType().getName() );
		sb.append( ')' );
```

beikov (approved):

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
dreab8: why? may be I'm missing something but I don't understand this

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
dreab8: applied your suggestion

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
beikov: Because this operation is like a cast that can be optimized away.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/expression/AsWrapperSqmExpression.java
beikov: Here a few suggestions for function names:

* `morph(timestampProperty as Instant)`
* `transform(timestampProperty as Instant)`
* `transmute(timestampProperty as Instant)`
* `cast_optional(timestampProperty as Instant)`

Here a few names I dismissed because SQL functions with the same name exist:

* `convert` 
* `translate`

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/8681

PR #5641, opened 2022-11-29
unify new ValueGenerationStrategy interfaces with old ValueGeneration

gavinking:
- introduce `AbstractValueGeneration` to abstract over IMVGS and IDVGS
- make `ValueGeneration` a mixin of IMVGS with IDVGS
- make IDVGS accept a Dialect for SQL fragment generation (it was based on an obsolete version of `ValueGeneration`)
- adapt all the code which only handled single-column value generation to handle multiple columns, as introduced by IDVGS
- make internal code pass around `AbstractValueGeneration` instead of `ValueGeneration`

Still to do: the whole `AnnotationValueGeneration` stuff still requires the use of the mixin interface, and you can't use IMVGS or IDVGS directly. That bit is going to require a bit more thinking about backward compatibility.

This is funny. Last night I started work on this (which I've been wanting to do for a while) and added my own IMVG and IDVG interfaces. And I was happily working away, starting from the front end of all this, until I got almost right to the back end, and discovered that @sebersole had already just a few hours earlier introduced essentially the same interfaces in the same package just with `Strategy` appended to the interface names! So I just merged his interfaces with mine and it all worked straight away.


gavinking:
@sebersole take a look please.

beikov:
I'd prefer naming the interface `ValueGenerationStrategy` instead of `AbstractValueGeneration` fwiw

gavinking:
> I'd prefer naming the interface `ValueGenerationStrategy` instead of `AbstractValueGeneration` fwiw

Yeah that's not a bad suggestion.

On the other hand what I was going to propose was to move the new interfaces out into a new package, and drop the `Strategy` off all of them. So it would just be `ValueGenerator`, `InMemoryValueGenerator`, and `InDatabaseValueGenerator`. Hell, I would even consider dropping the `Value` bit since it just seems like a noiseword.

gavinking:
But let's go with your suggestion for now. I renamed it.

sebersole:
> > I'd prefer naming the interface `ValueGenerationStrategy` instead of `AbstractValueGeneration` fwiw
> 
> Yeah that's not a bad suggestion.
> 
> On the other hand what I was going to propose was to move the new interfaces out into a new package, and drop the `Strategy` off all of them. So it would just be `ValueGenerator`, `InMemoryValueGenerator`, and `InDatabaseValueGenerator`. Hell, I would even consider dropping the `Value` bit since it just seems like a noiseword.

Dropping Value here, imo, makes this confusing with id generation.  This is about non-id generation specifically

gavinking:
How about I merge this, a bit later today, to give you guys a bit more time to object, as a phase 1?

Then we can think about whatever phase 2 is. (Deal with `AnnotationValueGeneration`, argue over names and packages, etc.)

Because as it is this is already a good atomic step forward.

sebersole:
I'm fine with that.  The overall work makes sense.  The rest is details.  Though ofc those details need to be sorted prior to 6.2 but we have some time for that

PR #5643, opened 2022-11-29
HHH-15731 Multi-level Derived Tables fail with Unsupported path source: org.hibernate.query.derived.AnonymousTupleSqmPathSource

dreab8:
https://hibernate.atlassian.net/browse/HHH-15731


beikov (approved):
LGTM

PR #5646, opened 2022-11-29
HHH-15778 Fetching an Entity with a lazily loaded Embeddable with more fields than the parent results in an ArrayIndexOutOfBoundsException

GeorgEchterling:
https://hibernate.atlassian.net/browse/HHH-15778

I made an attempt at fixing the bug, but I don't know enough about Hibernate internals to be sure about these changes.


dreab8:
Thaks @GeorgEchterling your fix LGTM, i have rebased it on main and created https://github.com/hibernate/hibernate-orm/pull/5809, if it passes all the test I'm going to merge it.

PR #5647, opened 2022-11-29
HHH-15733 @Convert on Map should not require "attributeName=value"

mbladel:
https://hibernate.atlassian.net/browse/HHH-15733

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/internal/CollectionPropertyHolder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/internal/CollectionPropertyHolder.java
@@ -78,6 +78,7 @@ public class CollectionPropertyHolder extends AbstractPropertyHolder {
 
 	private void buildAttributeConversionInfoMaps(
 			XProperty collectionProperty,
+			boolean isComposite,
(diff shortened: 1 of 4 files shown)


mbladel:
I'm open to discuss if there's a better way to check wether the key and value of the Map collection is of a basic type allowing to default to `attributeName="value"` for the `@Convert`.

mbladel:
Rebased PR with main to solve conflicts and cleared up the test.

hibernate-core/src/main/java/org/hibernate/boot/model/internal/CollectionPropertyHolder.java
```diff
@@ -326,6 +340,10 @@ public String toString() {
 	boolean prepared;
 
 	public void prepare(XProperty collectionProperty) {
+		prepare(collectionProperty, true);
```
mbladel, line 343: This default signature for `prepare` might not be necessart since `CollectionPropertyHolder` is in an internal package.

beikov: Correct, you can just change the method. No need to care about backwards compatibility here.

beikov (approved):
LGTM

PR #5649, opened 2022-11-29
HHH-15741 Explicit declaration assigned identifier generator does not work

dreab8:
https://hibernate.atlassian.net/browse/HHH-15741


beikov (approved):
From what I can tell, the change looks fine to me

PR #5652, opened 2022-11-30
POC - New dispatch approach to manage type checks of enhanced entities

Sanne:
This is a prototype I need to share and get tested by CI.

Explanations to follow...


hibernate-core/src/main/java/org/hibernate/engine/internal/ManagedTypeHelper.java
```diff
@@ -6,95 +6,128 @@
...
+		}
+	};
+
 	/**
 	 * @param type
 	 * @return true if and only if the type is assignable to a {@see Managed} type.
 	 */
 	public static boolean isManagedType(final Class type) {
-		return EnhancedEntity.class.isAssignableFrom( type ) || Managed.class.isAssignableFrom( type );
+		return typeMetaCache.get( type ).isManagedType;
```
franz1981, line 76: less vcalls, nice :)

Sanne:
This was since implemented in https://hibernate.atlassian.net/browse/HHH-15790

PR #5655, opened 2022-11-30
HHH-15788, HHH-15789  Unify value generators and id generators

gavinking:
I have managed to turn `IdentifierGenerator` into just a special type of `InMemoryGenerator`, and given id generators and value generators the same sort of options for initialization. This is a *really* nice cleanup.

It would now be really straightforward to accept *any* `InMemoryGenerator` as an `IdentifierGenerator`, making the `IdentifierGenerator` interface optional. I'm not doing that in this commit, because I don't think it can be done cleanly without breakage.


gavinking:
Hrm, so `VersionSourceTest` just failed on every database, but ... all tests pass for me locally :-/

gavinking:
@sebersole check out what I did here.

PR #5661, opened 2022-12-01
make PostInsertIdentifierGenerator extend InDatabaseGenerator

gavinking:
Initial stab at this. Seems to have worked out reasonably smoothly, except for one thing.

This method:

```java
InsertGeneratedIdentifierDelegate getInsertGeneratedIdentifierDelegate(
	PostInsertIdentityPersister persister,
	Dialect dialect,
	boolean isGetGeneratedKeysEnabled);
```

Looks too technical to fit comfortably on `InDatabaseGenerator`.

And anyway `InDatabaseGenerator` has its own thing: `GeneratedValuesProcessor` which is pretty different.

Not sure if it's possible to unify these two mechanisms, so I might just introduce a separate interface for that one method.

cc @sebersole


gavinking:
Going to merge this because it's a really great step forward. We can of course still debate the details.

PR #5663, opened 2022-12-02
HHH-15791 Avoid excessive Session casts to EventSource

Sanne:
https://hibernate.atlassian.net/browse/HHH-15791


Sanne:
Updated the migration guide:
https://github.com/hibernate/hibernate-orm/blob/main/migration-guide.adoc#changes-in-integration-contracts-spis

PR #5032, opened 2022-05-09
[HHH-15259] Explicitly add JavaDoc to make @deprecated hint visible in Eclipse

csware:
The newly added method overrides (i.e., https://github.com/hibernate/hibernate-orm/commit/aa08c90b52c4de2c86f33f31b0d5d27b461d1afc) "hide" the JavaDoc from QueryProducer (at least in Eclipse 2022-03). Explicitly add the javadoc here to provide users the @deprecated JavaDoc hint on how to resolve the deprecation warnings in Eclipse.

PR #5667, opened 2022-12-02
HHH-15792: Explicitly add JavaDoc to make @deprecated hint for createSQLQuery visible in Eclipse

csware:
Explicitly add the JavaDoc to Session.java to provide users the @deprecated JavaDoc hint on how to resolve the deprecation warnings.

Without this patch, e.g., in Eclipse, there are no warning generated when the method createSQLQuery is used which makes upgrading to 6.x harder (cf. #5032).


beikov:
Thanks!

PR #5671, opened 2022-12-02
HHH-15790 Type Pollution Mitigation on generic BiConsumer

franz1981:
As explained in the code itself, this is going to cover a very stealth case of type pollution due to how type erasure works in Java.

Sadly, despite https://github.com/RedHatPerf/type-pollution-agent has introduced support for lambda redefinition, given that the underlying ByteBuddy feature is "experimental" (see https://github.com/raphw/byte-buddy/issues/558#issuecomment-440443548 for more info ) isn't always possible to verify it in the usual pollution reports.

Luckly the flamegraphs are good enough to spot it
image


Sanne:
Expanded & added JIRA references. Replaced by #5684

PR #5679, opened 2022-12-03
HHH-15801 IdentifierGeneratorHelper should compare column names case insensitively

mrotteveel:
Fixes [HHH-15801](https://hibernate.atlassian.net/browse/HHH-15801)


gavinking:
Thank you!

PR #5680, opened 2022-12-03
HHH-15797 proposed solution 

gavinking:
See https://github.com/hibernate/hibernate-orm/discussions/5349#discussioncomment-4298078


hibernate-core/src/main/java/org/hibernate/annotations/GenerationTime.java
```diff
@@ -6,46 +6,80 @@
...
-	UPDATE( GenerationTiming.UPDATE ),
+	UPDATE,
 	/**
-	 * Indicates the value is generated on insert and on update.
+	 * Indicates that a new value is generated on insert and on update.
+	 *
+	 * @since 6.2
 	 */
-	ALWAYS( GenerationTiming.ALWAYS );
+	INSERT_OR_UPDATE,
```
sebersole, line 43: I prefer ALWAYS (or BOTH) over a needlessly longer name with underscores.  "Reads" better imo.

It should be AND at the very least, not OR

hibernate-core/src/main/java/org/hibernate/annotations/GeneratorType.java
```diff
@@ -83,8 +67,6 @@
 	 * <li>If {@link GenerationTime#ALWAYS}, the value will be generated before
 	 *     each SQL {@code insert} or {@code update} statement is executed.
 	 * </ul>
-	 * @deprecated use {@link #timing()}
 	 */
-	@Deprecated(since = "6.2")
 	GenerationTime when() default GenerationTime.ALWAYS;
```
sebersole, line 71: You deprecated ALWAYS in favor of INSERT_OR_UPDATE.  The default should be INSERT_OR_UPDATE here.

* I disagree with INSERT_OR_UPDATE overall, so :shrug:

hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
```diff
@@ -2547,22 +2534,22 @@ private void bindProperty(
 							mappingDocument.getOrigin()
 					);
 				}
-				if ( generationTiming.isNotNever() ) {
-					property.setValueGeneratorCreator(context -> new GeneratedValueGeneration( generationTiming ) );
+				if ( timing != GenerationTiming.NEVER ) {
```
sebersole, line 2537: I find it odd that you go out of the way to define simplified checks for all other cases (`#includesInsert`, etc) but here directly check equality on the enum.  Seems like this should be consistent

hibernate-core/src/main/java/org/hibernate/generator/Generator.java
```diff
@@ -48,24 +49,19 @@
...
-	GenerationTiming getGenerationTiming();
-
 	/**
 	 * Determines if the property value is generated in Java, or by the database.
 	 *
 	 * @return {@code true} if the value is generated by the database, or false if it is
 	 *		   generated in Java using a {@link ValueGenerator}.
 	 */
 	boolean generatedByDatabase();
+
```
sebersole, line 59: VERY much dislike decomposing the enum into 2 booleans

sebersole: Well, I guess this depends whether we really, actually support UPDATE.  I believe the truly valid options are (1) never, (2) on insert and (3) on both insert and update.

If we do in fact support just UPDATE, then my main concern here is gone.

gavinking: Yeah, it was introduced in https://hibernate.atlassian.net/browse/HHH-15672.

hibernate-core/src/main/java/org/hibernate/boot/model/source/internal/hbm/ModelBinder.java
gavinking: aaaaaaah good point. I added `isNotNever()` in an earlier PR, because I was using it in lots of places. Not anymore.

So I was planning on removing both `isNotNever()` and `isAlways()` because after this change they were not needed: 

- `isNotNever()` was now only used in that one place, and
- `isAlways()` is now unused (but then I actually forgot to remove it!)

I should either:

- remove `isAlways()`, or
- put back `isNotNever()`.

Your call, I don't care.

hibernate-core/src/main/java/org/hibernate/annotations/GeneratorType.java
gavinking: Yeah but the member is also deprecated :)

hibernate-core/src/main/java/org/hibernate/annotations/GenerationTime.java
gavinking: So I struggled quite a lot with this one. The issue is that the name I originally wanted to use, `timing`, was taken, and I had to look for something else. I tried a few different things like `on`, which I quite liked, but none of them read well with `ALWAYS`. And then the name I *really* liked, `event`, reads *really* badly with `ALWAYS`.

And then it sort of hit me that `ALWAYS` is a bit too strong anyway. It's not *always* regenerated, only when the entity is inserted or updated.

So:

-  "generated when `event=INSERT_OR_UPDATE`" or 
- "`@CurrentTimestamp` of the last `INSERT_OR_UPDATE` `event`"

sounded reasonable to me. (But I'm very open to other ideas.)

gavinking: > It should be AND at the very least, not OR

Depends how you read it. If we think its describing a union of sets, then yes, "and" is what most people use to describe a union in everyday speech.

But if we think of it as a predicate applied to a single event (a "filter") then it's "or". (Mathematically, set union actually corresponds to logical disjunction.)

So it it were plural, I would go with "and", as in `INSERTS_AND_UPDATES`. But since it's singular, `INSERT_OR_UPDATE` sounds more correct to me. `INSERT_AND_UPDATE` sounds weird because no event is *both* an insert *and* an update at the same time.

hibernate-core/src/main/java/org/hibernate/generator/Generator.java
gavinking: @sebersole your comments have got me thinking. I wonder if, for the sake of future extensibility, and elegance, it would be better if we introduced something like:

```
enum GenerationEventType {
     INSERT, UPDATE
}
```

and then used `EntityEventType[]` to express these things.

This would follow the pattern established by `CascadeType`. So "always" would be `{INSERT,UPDATE}` and "never" would be `{}`.

gavinking: I have just implemented that, so you can see what I mean. I think it worked out well, much better than the "two booleans" thing. And I always wanted to use `EnumSet` for something ;-)

gavinking: Introducing `EventType` also made the mapping to version generation cleaner. Nice.

gavinking:
Closing, superseded by #5685 after rebase.

PR #5697, opened 2022-12-05
HHH-15812 Firebird dialect improvements

mrotteveel:
Fixes [HHH-15812](https://hibernate.atlassian.net/browse/HHH-15812)


beikov (approved):
Nice work. Looks good to me

gavinking:
Excellent, looks very comprehensive!

mrotteveel:
I still have some tests failing with the dialect, but so far I have been able to attribute most of them to bugs in Jaybird (or bugs in Jaybird obscure the actual problem). I'll investigate those further after I have been able to fix them in Jaybird. There also some failures when all tests are run, but not when the test is run individually: a table is still in use by some previous test, so it can't be dropped, resulting in the wrong table columns during the test.

mrotteveel:
@beikov This can be merged

PR #5699, opened 2022-12-05
HHH-15819, HHH-15820, HHH-15828 fix extract(offset hour from datetime), etc

gavinking:
bug noticed by @mrotteveel


beikov (approved):

gavinking:
OK so *this* is frustrating. One of my new tests is failing on Sybase here, but it runs just fine on my laptop. :-/

gavinking:
That worked, finally!

PR #5704, opened 2022-12-05
HHH-15653 Named Native Query cannot be registered/used with named parameters

dreab8:
https://hibernate.atlassian.net/browse/HHH-15653

The issue is that the sql string of the memento used by `addNamedQuery` has been adjusted, so instead of the named parameter it contains the `?` , this causes the creation of a ordinal instead of using the registered name parameter.


beikov (approved):
LGTM

PR #5711, opened 2022-12-06
[draft] Exploring options to fix type pollution on Query executions

Sanne:
Another follow up to HHH-15791 based on `6.1` for the performance team to test.


Sanne:
ouch - you're right in closing this, but I had forgotten about this one and it was a lot of good work. Would be nice to be able to find some time to see if there's some good ideas in here which are still relevant.

In general I don't consider "workarounds" for type pollution worth it anymore; they were primarily intended to see how far we could get the system in an hypotethical world in which the JVM was not affected. But it has been fixed properly, so now this is only of interest to those people who aren't upgrading yet - less so of a priority for us.

However there have been cases during this work in which *thinking* carefully about the types we actually improved our code. So when it's not a dirty-hack-workaround but a cleanup I'd like to still apply them.

This one in particular was a good one, as far as I can see.

gavinking:
> However there have been cases during this work in which thinking carefully about the types we actually improved our code.

I'm not sure that changing `SharedSessionContractImplementor` or `SessionImplementor` to `EventSource` is an improvement though. Generally we prefer to pass the more abstract type, if that's all that needed. 

(On the other hand, if it eliminates typecasts further down the line, it *is* an improvement.)

Sanne:
> On the other hand, if it eliminates typecasts further down the line, it is an improvement.

Right - it's my recollection that several such improvements were triggered. But I'd need to double check, certainly not re-applying this blindly.

PR #5712, opened 2022-12-06
HHH-15745 IllegalArgumentException while executing Like query with Escapes

mbladel:
[HHH-15745](https://hibernate.atlassian.net/browse/HHH-15745)

The issue was related to unnecessary handling of escape sequences in string literals that should be removed.


hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlLexer.g4
```diff
@@ -75,10 +75,13 @@ HEX_LITERAL : '0' [xX] HEX_DIGIT+ LONG_SUFFIX?;
 fragment SINGLE_QUOTE : '\'';
 fragment DOUBLE_QUOTE : '"';
 
-STRING_LITERAL
+STRING_LITERAL : SINGLE_QUOTE ( SINGLE_QUOTE SINGLE_QUOTE | ~('\'') )* SINGLE_QUOTE;
+
+JAVA_STRING_LITERAL
 	: DOUBLE_QUOTE ( ESCAPE_SEQUENCE | DOUBLE_QUOTE DOUBLE_QUOTE | ~('"') )* DOUBLE_QUOTE
```
beikov, line 81:
```suggestion
	: DOUBLE_QUOTE ( ESCAPE_SEQUENCE | ~('"') )* DOUBLE_QUOTE
```

hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlLexer.g4
```diff
@@ -75,10 +75,13 @@ HEX_LITERAL : '0' [xX] HEX_DIGIT+ LONG_SUFFIX?;
...
-	| SINGLE_QUOTE ( ESCAPE_SEQUENCE | SINGLE_QUOTE SINGLE_QUOTE | ~('\'') )* SINGLE_QUOTE
-	;
+ 	| [jJ] SINGLE_QUOTE ( ESCAPE_SEQUENCE | SINGLE_QUOTE SINGLE_QUOTE | ~('\'') )* SINGLE_QUOTE
+ 	| [jJ] DOUBLE_QUOTE ( ESCAPE_SEQUENCE | DOUBLE_QUOTE DOUBLE_QUOTE | ~('\'') )* DOUBLE_QUOTE
```
beikov, line 83:
```suggestion
 	| [jJ] SINGLE_QUOTE ( ESCAPE_SEQUENCE | ~('\'') )* SINGLE_QUOTE
 	| [jJ] DOUBLE_QUOTE ( ESCAPE_SEQUENCE | ~('\'') )* DOUBLE_QUOTE
```

hibernate-core/src/main/java/org/hibernate/internal/util/QuotingHelper.java
```diff
@@ -70,13 +70,24 @@ public static String unquoteIdentifier(String text) {
...
+		return unquoteString( text, false );
+	}
+
+	public static String unquoteJavaStringLiteral(String text) {
+		return unquoteString( text, true );
+	}
+
+	public static String unquoteString(String text, boolean unescape) {
 		assert text.length() > 1;
+		int start = 0;
```
beikov, line 82: I'd prefer passing the start index explicitly.

```suggestion
		return unquoteString( text, false, 0 );
	}

	public static String unquoteJavaStringLiteral(String text) {
		final char firstChar = text.charAt( 0 );
		return unquoteString( text, true, firstChar == 'j' || firstChar == 'J' ? 1 : 0 );
	}

	public static String unquoteString(String text, boolean unescape, int start) {
		assert text.length() > 1;
```

hibernate-core/src/main/java/org/hibernate/internal/util/QuotingHelper.java
```diff
@@ -70,13 +70,24 @@ public static String unquoteIdentifier(String text) {
...
 		final int end = text.length() - 1;
-		final char delimiter = text.charAt( 0 );
+		final char delimiter = unescape && Character.toLowerCase( text.charAt( start ) ) == 'j' ?
+				text.charAt( ++start ) :
+				text.charAt( start );
```
beikov, line 86: The suggestion to pass the start index explicitly is mostly because this conditional expression increment feels like voodoo 😄 

```suggestion
		final char delimiter = text.charAt( start );
```

beikov (commented):
Great work!

hibernate-core/src/main/java/org/hibernate/internal/util/QuotingHelper.java
mbladel: Yep, I guess one shouldn't code late at night 😁 anyways this was a quick WIP, mainly wanted to update the code with the new ideas but still waiting on any news in the discussion.

For now I applied the proposed changes, thanks.

hibernate-core/src/main/java/org/hibernate/internal/util/QuotingHelper.java
```diff
@@ -70,13 +70,22 @@ public static String unquoteIdentifier(String text) {
...
+		final char delimiter = text.charAt( start );
 		assert delimiter == text.charAt( end );
 		// Unescape the parsed literal and handle escape sequences
-		final StringBuilder sb = new StringBuilder( text.length() - 2 );
-		for ( int i = 1; i < end; i++ ) {
+		final StringBuilder sb = new StringBuilder( text.length() - ( start + 2 ) );
+		for ( int i = start + 1; i < end; i++ ) {
 			char c = text.charAt( i );
 			switch ( c ) {
 				case '\'':
```
beikov, line 91: The double unquoting should actually only happend when `!escape`, so I think it might be worth splitting the method into two. One that does just the double quote unescaping i.e. move logic to `unquoteStringLiteral`, and one that does just the escape sequence unescaping i.e. move logic to `unquoteJavaStringLiteral`

hibernate-core/src/main/java/org/hibernate/internal/util/QuotingHelper.java
beikov: Don't worry 😁

hibernate-core/src/main/java/org/hibernate/internal/util/QuotingHelper.java
mbladel: Done!

beikov (approved):
Awesome, thanks a lot!

PR #5713, opened 2022-12-06
HHH-15674 @NamedEntityGraph not working on entity that have composite PK

dreab8:
https://hibernate.atlassian.net/browse/HHH-15674

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/EntityResultGraphNode.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/EntityResultGraphNode.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.sql.results.graph.entity;
 
+import org.hibernate.graph.spi.GraphImplementor;
 import org.hibernate.metamodel.mapping.EntityMappingType;
 import org.hibernate.sql.results.graph.DomainResultGraphNode;
 import org.hibernate.sql.results.graph.FetchParent;
@@ -37,5 +38,12 @@ public interface EntityResultGraphNode extends DomainResultGraphNode, FetchParen
 	@Override
 	default EntityMappingType getReferencedMappingContainer() {
 		return getEntityValuedModelPart().getEntityMappingType();
+	}
+
+	@Override
+	default boolean appliesTo(GraphImplementor graphImplementor) {
+		return graphImplementor.appliesTo(
+				getEntityValuedModelPart().getEntityMappingType().getJavaType().getJavaTypeClass()
+		);
 	}
 }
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/internal/StandardEntityGraphTraversalStateImpl.java
```diff
@@ -117,19 +118,51 @@ private Class<?> getEntityCollectionPartJavaClass(CollectionPart collectionPart)
...
+		else if ( !( fetchParent instanceof EntityResultGraphNode ) ) {
+			if ( fetchParent instanceof EmbeddableFetchImpl ) {
+				return appliesTo( (EmbeddableFetchImpl) fetchParent );
+			}
+			else {
+				return false;
+			}
+		}
+
+		return appliesTo( (EntityResultGraphNode) fetchParent );
```
beikov, line 133:
```suggestion
		if ( fetchParent instanceof EmbeddableFetchImpl ) {
			return appliesTo( (EmbeddableFetchImpl) fetchParent );
		}
		else if ( fetchParent instanceof EntityResultGraphNode ) {
			return appliesTo( (EntityResultGraphNode) fetchParent );
		}
		else {
			return false;
		}

```

beikov (commented):
Looks ok to me, but I think the `appliesTo` methods might be better moved to `DomainResultGraphNode` i.e. `boolean appliesTo(ManagedDomainType<?> domainType)`

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/EntityResultGraphNode.java
```diff
@@ -38,4 +40,14 @@ default JavaType<?> getResultJavaType() {
 	default EntityMappingType getReferencedMappingContainer() {
 		return getEntityValuedModelPart().getEntityMappingType();
 	}
+
+	@Override
+	default boolean appliesTo(GraphImplementor graphImplementor) {
+		assert graphImplementor.getGraphedType() instanceof EntityDomainType;
```
beikov, line 46: This looks way better, but don't you think that this assert statement should rather be this?

```java
if ( !( graphImplementor.getGraphedType() instanceof EntityDomainType ) ) {
    return false;
}
```

I have no idea how this is called, so this could be totally wrong, just wanted to mention it so you give it a second thought.

dreab8: changed this method implementation, what do you think?

beikov (approved):

dreab8:
Thanks @beikov

PR #5727, opened 2022-12-08
Fix temporal round in tests and render calendar with milliseconds precision

beikov:


hibernate-core/src/test/java/org/hibernate/orm/test/timezones/JDBCTimeZoneZonedTest.java
```diff
@@ -39,19 +42,24 @@ public class JDBCTimeZoneZonedTest {
...
+				);
+				assertEquals(
+						DateTimeUtils.roundToDefaultPrecision( nowOffset.toInstant(), dialect ),
+						DateTimeUtils.roundToDefaultPrecision( z.offsetDateTime.toInstant(), dialect )
+				);
 			}
-			assertEquals( systemZone, z.zonedDateTime.getZone() );
-			assertEquals( systemOffset, z.offsetDateTime.getOffset() );
+			assertEquals( ZoneId.systemDefault(), z.zonedDateTime.getZone() );
+			assertEquals( now.getOffset().normalized(), z.offsetDateTime.getOffset().normalized() );
```
gavinking, line 62: But this isn't testing quite the same thing. 

    getOffset().normalized()

is not quite the same things as:

    getOffset()

PR #5729, opened 2022-12-08
HHH-15327 Support mapping aggregate embeddables as struct/json

beikov:
After many reviews and discussions elsewhere, this is done now. I won't split this PR up further as it was taking too much time already. Further improvements go into separate PRs. I acknowledge that user-guide docs are missing. Unless someone beats me to it, I'll work on that in the next few days and make sure it's there for the final 6.2 release.

https://hibernate.atlassian.net/browse/HHH-15327


hibernate-core/src/main/java/org/hibernate/annotations/Struct.java
```diff
@@ -0,0 +1,46 @@
...
+ * <pre>
+ *    Example:
+ *
+ *    public class Order {
+ *        &#064;Embedded
+ *        &#064;Struct(name = "CUST")
+ *        private Customer customer;
+ *    }
+ * </pre>
+ *
```
gavinking, line 37: Missing @author and @since tags?

hibernate-core/src/main/java/org/hibernate/annotations/Struct.java
```diff
@@ -0,0 +1,46 @@
...
+import static java.lang.annotation.ElementType.TYPE;
+
+/**
+ * Specifies the UDT (user defined type) name for the annotated embeddable or embedded.
+ *
+ * <pre>
+ *    Example:
+ *
+ *    &#064;Embeddable
+ *    &#064;Struct(name = "CUST")
```
gavinking, line 24: Better to use `{@code ..... }` here.

gavinking:
This is looking really good, and it's exciting to have it in 6.2!

PR #5737, opened 2022-12-11
HHH-15848 session.isDirty() shouldn't throw exception for transient many-to-one object in a session

chenhh021:
https://hibernate.atlassian.net/browse/HHH-15848


gavinking (changes requested):
I agree that this is a bug, but it's a much deeper bug that what is contemplated by the proposed fix: essentially, `isDirty()` has all sorts of side effects on the state of the session. This makes me think that we should perhaps even just deprecate this method of `Session`.

The proposed fix is definitely not OK, because it can in principle result in database access from `ManyToOneType.isDirty()`.

jrenaat:
Closing this as per @gavinking's last comment. More discussion needed.

gavinking:
It's not that more discussion is needed, it's that I already fixed it in https://github.com/hibernate/hibernate-orm/pull/9476 by completely reimplementing `isDirty()` from scratch.

PR #5739, opened 2022-12-12
HHH-15759 Entity with @CollectionTable and entity with relation to element collection table, cause generate wrong FK, if first entity key field lexicographically after collection element.

dreab8:
https://hibernate.atlassian.net/browse/HHH-15759

alternative solution for https://github.com/hibernate/hibernate-orm/pull/5612


beikov (approved):
LGTM

PR #5744, opened 2022-12-12
HHH-15853 Avoid overhead from runtime allocations of AttributeMetadataAccess

Sanne:
https://hibernate.atlassian.net/browse/HHH-15853


beikov (approved):
Great, that's exactly what I had in mind 😃

PR #5746, opened 2022-12-12
HHH-15736 Make sure dialects have default like-predicate escape char configured

mbladel:


hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
```diff
@@ -137,22 +138,29 @@ public Size resolveSize(
...
 		this( MINIMUM_VERSION );
 	}
 
 	public MySQLDialect(DatabaseVersion version) {
-		this( version, 4 );
+		this( version, 4, false );
 	}
 
-	public MySQLDialect(DatabaseVersion version, int bytesPerCharacter) {
+	public MySQLDialect(DatabaseVersion version, int bytesPerCharacter, boolean noBackslashEscapes) {
```
beikov, line 151: First, we need a new constructor and leave the signature of the old one untouched for backwards compatibility. Apart from that, I'd like to ask you to introduce a `MySQLServerConfiguration` class, which encapsulates the `bytesPerCharacter` and `noBackslashEscapes`, which we determine in a single query through a factory method `MySQLServerConfiguration.fromDatabaseMetadata(DatabaseMetadata)`. The legacy dialects can then also use that.

hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/MariaDBLegacySqlAstTranslator.java
```diff
@@ -157,6 +160,75 @@ else if ( expression instanceof Summarization ) {
...
+			else {
+				// replace(<value>, '\\', '\\\\')
+				appendSql( "replace" );
+				appendSql( OPEN_PARENTHESIS );
+				pattern.accept( this );
+				String replaceParams = "'\\\\','\\\\\\\\'";
+				if ( getDialect().isNoBackslashEscapesEnabled() ) {
+					replaceParams = replaceParams.replace( "\\\\", "\\" );
+				}
+				appendSql( "," + replaceParams );
```
beikov, line 226:
```suggestion
				if ( getDialect().isNoBackslashEscapesEnabled() ) {
					appendSql( ",'\\','\\\\'" );
				}
				else {
					appendSql( ",'\\\\','\\\\\\\\'" );
				}
```

hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/MariaDBLegacySqlAstTranslator.java
```diff
@@ -157,6 +160,75 @@ else if ( expression instanceof Summarization ) {
...
+			escapeLikePattern( likePredicate.getPattern(), likePredicate.getEscapeCharacter() );
+			appendSql( CLOSE_PARENTHESIS );
+		}
+		if ( likePredicate.getEscapeCharacter() != null ) {
+			appendSql( " escape " );
+			likePredicate.getEscapeCharacter().accept( this );
+		}
+	}
+
+	protected void escapeLikePattern(Expression pattern, Expression escapeCharacter) {
```
beikov, line 193: I think it would be good to move this method to the `AbstractSqlAstTranslator` to avoid duplication of the logic and name it `renderBackslashEscapedLikePattern` as well as pass a `noBackslashEscapesEnabled` parameter.

beikov (commented):
Good work! I left a few comments which you can hopefully incorporate to reduce code duplication a bit.

hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java
mbladel: Should I leave the protected method `getCharacterSetBytesPerCharacter` in `MySQLDialect` for compatibility?

beikov: Yeah, but let's deprecate them.

hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/MariaDBLegacySqlAstTranslator.java
beikov: Sounds good to me

PR #5748, opened 2022-12-13
HHH-15822 Unexpected org.hibernate.UnknownEntityTypeException: Unable to locate persister

mbladel:
[HHH-15822](https://hibernate.atlassian.net/browse/HHH-15822)

[HHH-15822]: https://hibernate.atlassian.net/browse/HHH-15822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/SqmEntityValuedSimplePath.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/SqmEntityValuedSimplePath.java
@@ -63,10 +63,18 @@ public class SqmEntityValuedSimplePath<T> extends AbstractSqmSimplePath<T> {
 	}
 
 	@Override
-	public EntityDomainType<T> getNodeType() {
+	public SqmPathSource<T> getNodeType() {
 		//noinspection unchecked
-		return (EntityDomainType<T>) getReferencedPathSource().getSqmPathType();
+		return (SqmPathSource<T>) getReferencedPathSource().getSqmPathType();
 	}
+// We can't expose that the type is a EntityDomainType because it could also be a MappedSuperclass
+// Ideally, we would specify the return type to be IdentifiableDomainType, but that does not implement SqmPathSource yet
+// and is hence incompatible with the return type of the super class
+//	@Override
+//	public EntityDomainType<T> getNodeType() {
+//		//noinspection unchecked
+//		return (EntityDomainType<T>) getReferencedPathSource().getSqmPathType();
+//	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/internal/AttributeFactory.java
```diff
@@ -228,9 +228,9 @@ public static <Y> DomainType<Y> determineSimpleType(ValueContext typeContext, Me
 					if ( domainType == null ) {
 						// Due to the use of generics, it can happen that a mapped super class uses a type
 						// for an attribute that is not a managed type. Since this case is not specifically mentioned
-						// in the Jakarta Persistence spec, we handle this by returning a "dummy" entity type
+						// in the Jakarta Persistence spec, we handle this by returning an Object entity type
 						final JavaType<Y> domainJavaType = context.getJavaTypeRegistry().resolveDescriptor(
-								typeContext.getJpaBindableType()
+								Object.class
 						);
```
beikov, line 234: Doesn't this lead to reporting that the java type of the `parent` attribute is `Object` rather than the IMO more correct `AbstractTreeableEntity`?

mbladel: It does, but when the type is `AbstractTreeableEntity` queries on the `parent` attribute fail with the error as reported in the issue. This is because when executing queries the attribute `parent` is resolved to the mapper superclass, which is not a persistable Entity.

quaff: @mbladel It introduce new problem: `root.get("parent").get("id")` will produce `org.hibernate.query.SemanticException: Could not resolve attribute 'id' of 'java.lang.Object'"`

mbladel: @quaff thank you for the feedback. This error is not caused by the edit, even if before it was more readable reporting `AbstractTreeableEntity` instead of `Object`.

quaff: 🤝 Just remind you this PR is not a complete fix.

quaff: I think generic type should be resolved like this:
```java
Type resolvedType = org.springframework.core.GenericTypeResolver.resolveType(
		AbstractTreeableEntity.class.getDeclaredField("parent").getGenericType(), TestEntity.class);
assert (resolvedType == TestEntity.class);
```

PR #5753, opened 2022-12-13
HHH-15854 Performance regressions in Hibernate 6

dreab8:
https://hibernate.atlassian.net/browse/HHH-15854

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/AbstractImmediateCollectionInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/AbstractImmediateCollectionInitializer.java
@@ -71,6 +71,9 @@ public abstract class AbstractImmediateCollectionInitializer extends AbstractCol
 		if ( collectionInstance != null || collectionKey == null) {
 			return;
 		}
+		if ( parentAccess != null && !rowProcessingState.isQueryExecution() && parentAccess.findFirstEntityInitializer().isInitialized() ) {
+			return;
+		}
 
 		if ( CollectionLoadingLogger.TRACE_ENABLED ) {
 			COLL_LOAD_LOGGER.tracef(
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/SelectEagerCollectionInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/SelectEagerCollectionInitializer.java
@@ -28,6 +28,9 @@ public class SelectEagerCollectionInitializer extends AbstractCollectionInitiali
 
 	@Override
 	public void resolveInstance(RowProcessingState rowProcessingState) {
+		if ( parentAccess != null && !rowProcessingState.isQueryExecution() && parentAccess.findFirstEntityInitializer().isInitialized() ) {
+			return;
+		}
 		resolveInstance( rowProcessingState, true );
 	}
 
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
@@ -53,7 +53,6 @@ import org.hibernate.sql.results.graph.DomainResultAssembler;
 import org.hibernate.sql.results.graph.Fetch;
 import org.hibernate.sql.results.graph.Fetchable;
 import org.hibernate.sql.results.graph.basic.BasicResultAssembler;
-import org.hibernate.sql.results.graph.entity.internal.EntityResultInitializer;
 import org.hibernate.sql.results.internal.NullValueAssembler;
 import org.hibernate.sql.results.jdbc.spi.JdbcValuesSourceProcessingState;
 import org.hibernate.sql.results.jdbc.spi.RowProcessingState;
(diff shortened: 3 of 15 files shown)


Sanne:
let's not call an issue "Performance regressions in Hibernate 6 " - that's not specific enough

Sanne:
especially let's not call it "regressions" in plural - otherwise to fix the issue in JIRA you first have to ensure you fixed them all, or it would be misleading

theigl:
Hi @Sanne. I couldn't think of a better description when I created the JIRA issue, because I wasn't sure if my initial analysis was correct. Please feel free to change the issue title to something more concrete.

Sanne:
@theigl sure, no worries :) just reminding @dreab8 to rename it eventually before finishing. It's definitely easier to describe an issue after analysis, so you did the right thing.

dreab8:
I think we can split the Jira in two parts, one Jira for the problems with collections/associations initialization and another for query plan caching criteria queries

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
```diff
@@ -475,6 +473,9 @@ public void resolveInstance(RowProcessingState rowProcessingState) {
 			final Object existingEntity = persistenceContext.getEntity( entityKey );
 			if ( existingEntity != null ) {
 				this.entityInstance = existingEntity;
+				if ( existingLoadingEntry == null && isExistingEntityInitialized( existingEntity ) ) {
+					this.isInitialized = true;
```
beikov, line 477: Should we maybe still call `notifyResolutionListeners` for the instance?

hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/AbstractImmediateCollectionInitializer.java
```diff
@@ -71,6 +71,9 @@ public void resolveInstance(RowProcessingState rowProcessingState) {
 		if ( collectionInstance != null || collectionKey == null) {
 			return;
 		}
+		if ( parentAccess != null && !rowProcessingState.isQueryExecution() && parentAccess.findFirstEntityInitializer().isInitialized() ) {
```
beikov, line 74: What happens when the entity in the persistence context has an uninitialized collection? Aren't we "wrongly" skipping initialization then?

hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/SelectEagerCollectionInitializer.java
```diff
@@ -28,6 +28,9 @@ public SelectEagerCollectionInitializer(
 
 	@Override
 	public void resolveInstance(RowProcessingState rowProcessingState) {
+		if ( parentAccess != null && !rowProcessingState.isQueryExecution() && parentAccess.findFirstEntityInitializer().isInitialized() ) {
+			return;
```
beikov, line 32: Shouldn't we at least call `persistenceContext.addNonLazyCollection( collectionInstance );` here?

hibernate-core/src/main/java/org/hibernate/sql/results/internal/StandardRowReader.java
```diff
@@ -134,14 +138,22 @@ private void coordinateInitializers(RowProcessingState rowProcessingState) {
 			}
 		}
 
-		for ( int i = 0; i < numberOfInitializers; i++ ) {
+		final Initializer rootInitializer = initializers.get( numberOfInitializers - 1 );
```
beikov, line 141:
```suggestion
		final int lastIndex = numberOfInitializers - 1;
		final Initializer rootInitializer = initializers.get( lastIndex );
```

and use `lastIndex` in the loop conditions below

hibernate-core/src/main/java/org/hibernate/sql/results/internal/StandardRowReader.java
```diff
@@ -134,14 +138,22 @@ private void coordinateInitializers(RowProcessingState rowProcessingState) {
...
+		rootInitializer.resolveInstance( rowProcessingState );
+
+		if ( !rowProcessingState.isQueryExecution()
```
beikov, line 144: How about we generalize this optimization instead of limiting it just to non-query executions? I'd suggest we add a `default int numberOfNestedInitializers(){ return 0; }` to `Initializer` and use that information to "skip" sub-initializers while traversing the initializers list backwards through something like this:

```
		for ( int i = lastIndex; i >= 0; i-- ) {
			Initializer initializer = initializers.get( i );
			if ( !( initializer instanceof EntityDelayedFetchInitializer ) ) {
				initializer.resolveInstance( rowProcessingState );
				if ( initializer.isInitialized() ) {
					i -= initializer.numberOfNestedInitializers();
				}
			}
		}
```

hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/SelectEagerCollectionInitializer.java
dreab8: I think it should already be in the PC

beikov: Can you add a test that uses `maxFetchDepth = 0`, which should trigger this particular scenario? First load the owner entity, then do a query with an entity graph for that collection. The query should make sure the collection becomes initialized.

hibernate-core/src/main/java/org/hibernate/sql/results/internal/StandardRowReader.java
dreab8: we cannot do it for query 
`select Entity e....`
and then
`select Entity e fetch join e.collection`
and I think we can have the same issue with EntityGraphs

dreab8: for `EntityGraph` my check `!rowProcessingState.isQueryExecution()` is not enough :(

beikov: Why do you think this is not possible? My intuition is that this can be done, but maybe I'm missing something.

dreab8:
Superseded by https://github.com/hibernate/hibernate-orm/pull/5766

PR #5765, opened 2022-12-14
HHH-15869 Use explicit transactions for all SQL operations

maesenka:
The test fails on e.g. CockroachDB 21.2 because first DDL is in a Transaction but not the other DML/DDL statements.


beikov (approved):
Thanks a lot!

PR #5771, opened 2022-12-15
HHH-15768 incorrect deprecation warning being logged: HHH90000021: Encountered deprecated setting [javax.persistence.sharedCache.mode], use [jakarta.persistence.sharedCache.mode] instead

mbladel:
Backport of https://github.com/hibernate/hibernate-orm/pull/5752

### DIFF
--- a/hibernate-core/src/test/resources/hibernate.properties
+++ b/hibernate-core/src/test/resources/hibernate.properties
@@ -22,7 +22,7 @@ hibernate.max_fetch_depth 5
 hibernate.cache.region_prefix hibernate.test
 hibernate.cache.region.factory_class org.hibernate.testing.cache.CachingRegionFactory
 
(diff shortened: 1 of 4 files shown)


hibernate-core/src/test/resources/hibernate.properties
```diff
@@ -22,7 +22,7 @@ hibernate.max_fetch_depth 5
 hibernate.cache.region_prefix hibernate.test
 hibernate.cache.region.factory_class org.hibernate.testing.cache.CachingRegionFactory
 
-javax.persistence.validation.mode=NONE
+jakarta.persistence.validation.mode=NONE
 hibernate.service.allow_crawling=false
```
mbladel, line 26: This was causing test failures because it set a default for the deprecated setting. Now the property is aligned with `main` branch.

PR #5772, opened 2022-12-15
HHH-15736 Make sure dialects have default like-predicate escape char configured

mbladel:
[HHH-15736](https://hibernate.atlassian.net/browse/HHH-15736) - Solve errors in CockroachDB and MySQL 5.7 highlighted by new tests introduced in the previous PR https://github.com/hibernate/hibernate-orm/pull/5746.


beikov:
Thanks!

PR #5779, opened 2022-12-15
HHH-15550 Allow controlling whether Global- and PersistentTemporaryTables should be created

mbladel:
[HHH-15550](https://hibernate.atlassian.net/browse/HHH-15550)

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/temptable/GlobalTemporaryTableStrategy.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/temptable/GlobalTemporaryTableStrategy.java
@@ -28,6 +28,9 @@ public class GlobalTemporaryTableStrategy {
 	private static final Logger log = Logger.getLogger( GlobalTemporaryTableStrategy.class );
 
 	public static final String SHORT_NAME = "global_temporary";
+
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/temptable/PersistentTableStrategy.java
```diff
@@ -60,6 +65,19 @@ public void prepare(
...
+				CREATE_ID_TABLES,
+				StandardConverters.BOOLEAN,
+				true
+		);
+
+		if (!createIdTables ) {
+			return;
+		}
+
 		prepared = true;
```
beikov, line 81: I think you should move this before the if block

hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/temptable/GlobalTemporaryTableStrategy.java
```diff
@@ -55,6 +59,19 @@ public void prepare(
...
+				CREATE_ID_TABLES,
+				StandardConverters.BOOLEAN,
+				true
+		);
+
+		if ( !createIdTables ) {
+			return;
+		}
+
 		prepared = true;
```
beikov, line 75: I think you should move this before the if block

beikov (approved):
Minor comment, other than that, this looks great!

PR #5782, opened 2022-12-16
HHH-15398 Move to Jakarta Persistence 3.1 as baseline

mbladel:
https://hibernate.atlassian.net/browse/HHH-15398

Also cleaned up some compatibility logic for Jakarta 3.0 introduced with https://github.com/hibernate/hibernate-orm/pull/4832 since it's no longer needed.


hibernate-core/src/main/java/org/hibernate/id/factory/internal/StandardIdentifierGeneratorFactory.java
```diff
@@ -108,11 +108,7 @@ private void registerJpaGenerators() {
 		generatorTypeStrategyMap.put( GenerationType.SEQUENCE, SequenceGenerationTypeStrategy.INSTANCE );
 		generatorTypeStrategyMap.put( GenerationType.TABLE, TableGenerationTypeStrategy.INSTANCE );
 		generatorTypeStrategyMap.put( GenerationType.IDENTITY, IdentityGenerationTypeStrategy.INSTANCE );
-		try {
-			generatorTypeStrategyMap.put( GenerationType.valueOf( "UUID" ), UUIDGenerationTypeStrategy.INSTANCE );
-		}
-		catch (IllegalArgumentException ignore) {
-		}
+		generatorTypeStrategyMap.put( GenerationType.valueOf( "UUID" ), UUIDGenerationTypeStrategy.INSTANCE );
```
beikov, line 111: If you change the dependency version you should be able to use this:

```suggestion
		generatorTypeStrategyMap.put( GenerationType.UUID, UUIDGenerationTypeStrategy.INSTANCE );
```

hibernate-core/src/main/java/org/hibernate/bytecode/spi/ClassTransformer.java
```diff
@@ -35,5 +37,5 @@ byte[] transform(
 			String className,
 			Class<?> classBeingRedefined,
 			ProtectionDomain protectionDomain,
-			byte[] classfileBuffer);
+			byte[] classfileBuffer)  throws TransformerException;
```
mbladel, line 40: This is a change to an SPI method but should be inline with the [Jakarta 3.1 spec](https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/spi/classtransformer) of this interface.

beikov: Yeah, this is fine

hibernate-core/src/main/java/org/hibernate/id/factory/internal/StandardIdentifierGeneratorFactory.java
mbladel: Correct, I missed it - now it's fixed thanks!

beikov (approved):

PR #5789, opened 2022-12-16
HHH-15900 ID attributes  with "select" generator are mapped to identity columns

mbladel:
https://hibernate.atlassian.net/browse/HHH-15900

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/id/SelectGenerator.java
+++ b/hibernate-core/src/main/java/org/hibernate/id/SelectGenerator.java
@@ -8,14 +8,16 @@ package org.hibernate.id;
 
 import java.util.Properties;
 
+import org.hibernate.dialect.Dialect;
 import org.hibernate.generator.InDatabaseGenerator;
+import org.hibernate.id.factory.spi.StandardGenerator;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.service.ServiceRegistry;
 import org.hibernate.type.Type;
 
 /**
  * A generator that {@code select}s the just-{@code insert}ed row to determine the
- * {@code IDENTITY} column value assigned by the database. The correct row is located
+ * column value assigned by the database. The correct row is located
  * using a unique key of the entity, either:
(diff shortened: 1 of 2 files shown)


gavinking:
Hrm. I wonder if this is the correct fix for this. Perhaps a better fix would be to stop using `instanceof IdentityGenerator` to drive DDL generation.

gavinking:
What I mean is: this points to us needing a more robust/flexible way of deciding when to use an identity column.

mbladel:
@gavinking I agree, considering a more solid way of distinguishing identity columns would definitely not hurt. Still, separating `SelectGenerator` from `IdentityGenerator` made sense IMO because sometimes it could be applied to something you just want to select (e.g. a column populated by a simple trigger) but that is not an identity.

gavinking:
Ah, OK, but I’m not sure it’s a good way to do that. The whole `InDatabaseGenerator` branch does that, and in a more efficient way, since multiple fields can be fetched in a single select.

gavinking:
Well, actually the real difference between the `select` generator and the other `InDatabaseGenerator`s is that it uses a `@NaturalId` to locate the inserted row instead of using the PK. And this makes it a bit of an odd thing that doesn't fit completely comformtably within either the older or new abstractions. And as a result it actually *doesn't* work for anything except the PK column.

Like I wrote:

```java
 * Arguably, this class breaks the natural separation of responsibility between the
 * {@linkplain InDatabaseGenerator generator} and the coordinating
 * code, since it's role is to specify how the generated value is <em>retrieved</em>.
```

gavinking:
Now, on the other hand, what you're saying is that the original purpose of the `select` generator was not to be a worse `identity` generator but to handle other cases like triggers, and yeah, OK, looks like you're right about that and I simply messed up when I refactored all this stuff.

So I guess this proposed change is correct.

gavinking:
Yep, you're right, nice catch, thanks.

mbladel:
Thank you for taking the time to clarify the issue!

PR #5792, opened 2022-12-18
move capabilities of 'select' generator to @Generated, and massively improve them

gavinking:


gavinking:
This accidentally turned into a pretty big chunk of work. But I think I'm happy with the end result.

PR #5797, opened 2022-12-19
HHH-15794 NullPointerException when constructing mapping model for nested embeddables with not optional ManyToOne

dreab8:
https://hibernate.atlassian.net/browse/HHH-15794

[HHH-15794]: https://hibernate.atlassian.net/browse/HHH-15794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
@@ -281,9 +281,10 @@ public class ToOneAttributeMapping
 				if ( CollectionPart.Nature.fromNameExact( navigableRole.getParent().getLocalName() ) != null ) {
 					// * the to-one's parent is directly a collection element or index
 					// * therefore, its parent-parent should be the collection itself
-					final String pluralAttributeName = StringHelper.unqualify( navigableRole.getParent().getParent().getLocalName() );
-					final PluralAttributeMapping pluralAttribute = (PluralAttributeMapping) declaringEntityPersister.findAttributeMapping( pluralAttributeName );
+					final PluralAttributeMapping pluralAttribute = (PluralAttributeMapping) declaringEntityPersister.findByPath(
+							navigableRole.getParent().getParent().getFullPath().substring( declaringEntityPersister.getNavigableRole().getFullPath().length() + 1 )  );
 					assert pluralAttribute != null;
+
 					final QueryableCollection persister = (QueryableCollection) pluralAttribute.getCollectionDescriptor();
 					isKeyTableNullable = !persister.getTableName().equals( targetTableName );
 				}
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/embeddable/NestedEmbeddedWitnNotOptionalManyToOneTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/embeddable/NestedEmbeddedWitnNotOptionalManyToOneTest.java
@@ -1,0 +1,150 @@
+package org.hibernate.orm.test.embeddable;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.hibernate.testing.orm.junit.DomainModel;
+import org.hibernate.testing.orm.junit.SessionFactory;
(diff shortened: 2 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -5591,7 +5591,7 @@ public ModelPart findSubPart(String name, EntityMappingType treatTargetType) {
 		}
 
 		for ( AttributeMapping attribute : declaredAttributeMappings.valueIterator() ) {
-			if ( attribute instanceof EmbeddableValuedModelPart && attribute instanceof VirtualModelPart ) {
+			if ( attribute instanceof EmbeddableValuedModelPart  ) {
```
beikov, line 5594: I don't think this change is correct. We should only look for model parts within the embeddable type if the embeddable type is virtual/embedded, so I guess that the caller of this function should rather use `findByPath()` instead.

dreab8: @beikov reverted this change and used `findByPath()` after fixing it!

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ModelPartContainer.java
```diff
@@ -30,13 +30,17 @@ default ModelPart findByPath(String path) {
...
 		int dotIndex;
 		ModelPartContainer modelPartContainer = this;
-		while ( ( dotIndex = path.indexOf( '.', nextStart ) ) != -1 ) {
+		while ( ( dotIndex = path.indexOf( '.', nextStart ) ) != -1   ) {
 			modelPartContainer = (ModelPartContainer) modelPartContainer.findSubPart(
 					path.substring( nextStart, dotIndex ),
 					null
 			);
+			if ( modelPartContainer == null ) {
+				modelPartContainer = this;
```
beikov, line 39: At this point, the method should return null as no model part could be found for the path segment.

dreab8: with nested embedded this was not working

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -281,9 +281,10 @@ && equal( join.getKey(), manyToOne ) ) {
 				if ( CollectionPart.Nature.fromNameExact( navigableRole.getParent().getLocalName() ) != null ) {
 					// * the to-one's parent is directly a collection element or index
 					// * therefore, its parent-parent should be the collection itself
-					final String pluralAttributeName = StringHelper.unqualify( navigableRole.getParent().getParent().getLocalName() );
-					final PluralAttributeMapping pluralAttribute = (PluralAttributeMapping) declaringEntityPersister.findAttributeMapping( pluralAttributeName );
+					final PluralAttributeMapping pluralAttribute = (PluralAttributeMapping) declaringEntityPersister.findByPath(
+							navigableRole.getParent().getParent().getLocalName() );
```
beikov, line 285: I think the change in `findByPath` becomes unnecessary when you use this instead:

```suggestion
							navigableRole.getParent().getParent().getFullPath().substring( declaringEntityPersister.getNavigableRole().getFullPath().length() + 1 ) );
```

I guess that the problem you are facing is that the navigable role of a collection persister contains the fully qualified class name of the entity? This should fix it.

dreab8: it was one of my possible solutions but I did not like it :P
Do you think it is better than the change to `findByPath()`?

beikov: The change in `findByPath` is not correct, sorry :|
If you think about it, it's easily possible to construct an example that will produce wrong results due to that change in `findByPath`

beikov (approved):

PR #5800, opened 2022-12-19
HHH-15802 ubQuery with "in" results in java.lang.ClassCastException: class org.hibernate.metamodel.mapping.internal.BasicEntityIdentifierMappingImpl cannot be cast to class org.hibernate.metamodel.mapping.EntityValuedModelPart

dreab8:

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -3743,7 +3743,15 @@ public abstract class BaseSqmToSqlAstConverter<T extends Statement> extends Base
 		final Expression result;
 		if ( actualModelPart instanceof EntityValuedModelPart ) {
 			final EntityValuedModelPart entityValuedModelPart = (EntityValuedModelPart) actualModelPart;
-			final EntityValuedModelPart inferredEntityMapping = (EntityValuedModelPart) getInferredValueMapping();
+			final EntityValuedModelPart inferredEntityMapping;
+			final MappingModelExpressible<?> inferredValueMapping = getInferredValueMapping();
+			if ( inferredValueMapping instanceof EntityIdentifierMapping ) {
+				inferredEntityMapping = ( (EntityIdentifierMapping) inferredValueMapping ).findContainingEntityMapping();
+			}
+			else {
+				inferredEntityMapping = (EntityValuedModelPart) inferredValueMapping;
+			}
+
 			final ModelPart resultModelPart;
 			final EntityValuedModelPart interpretationModelPart;
 			final TableGroup tableGroupToUse;
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/InCriteriaTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/InCriteriaTest.java
@@ -1,0 +1,128 @@
+package org.hibernate.orm.test.jpa.criteria;
+
+import java.util.HashSet;
+import java.util.Set;
+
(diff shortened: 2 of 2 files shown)


beikov:
I don't think that we should support this. IMO it was a bug that this worked before.

csware:
Why shouldn't this be supported? If not, then you ahould also adjust the API (data types) that this is not possible anymore.

beikov:
Try using the static metamodel and then you will see a compile error:

```java
criteria.where(root.get(Participation_.id).in(subQuery));
```

csware:
> Try using the static metamodel and then you will see a compile error:
> 
> ```java
> criteria.where(root.get(Participation_.id).in(subQuery));
> ```

That is exactly what I reported in the original issue (https://hibernate.atlassian.net/jira/software/c/projects/HHH/issues/HHH-15802) and that works with 5.6.14 AND 6.1.0 (and starts to fail with 6.1.1).

Also, ```criteria.where(root.get("id").in(subQuery));``` works with 5.6.14...

beikov:
Hmm, the `jakarta.persistence.criteria.Path#in()` methods don't specify the type variable, which seems fishy to me. It IMO just doesn't make sense to compare an expression of type `Integer` with an expression of type `Participation`. The easy fix for you is to just remove the `.get(Participation_.id)` part. We have to discuss this with the team, but -1 from me for continuing to support this IMO ill-typed way of doing things. These sort of special cases complicate things in the translation code all over the place for IMO no benefit. The proper way is even shorter for users to write.

csware:
Removing the .get part works for me and would be a solution (added that workaround to the issue).

beikov:
Let's please not call it a workaround, as it's the proper way to do the comparison you are aiming for. You surely understand that comparing apples with oranges simply isn't the right thing to do. Most statically typed programming languages will not allow to comparing objects with values and I think nobody should rely on something like that.

csware:
There are still other places in current Hibernate (>= 6.1.1) where this still works:

```
CriteriaBuilder builder = session.getCriteriaBuilder();
CriteriaQuery<Lecture> criteria = builder.createQuery(Lecture.class);
Root<Lecture> root = criteria.from(Lecture.class);
criteria.select(root);
criteria.orderBy(builder.asc(root.get(Lecture_.name)), builder.desc(root.get(Lecture_.semester)));

Subquery<Lecture> subQuery = criteria.subquery(Lecture.class);
Root<Participation> lecturesTakingPartIn = subQuery.from(Participation.class);
subQuery.select(lecturesTakingPartIn.get(Participation_.lecture));
subQuery.where(builder.equal(lecturesTakingPartIn.get(Participation_.user), user));

criteria.where(builder.not(root.get(Lecture_.id).in(subQuery)));

return session.createQuery(criteria).list();
```

gavinking:
> Hmm, the `jakarta.persistence.criteria.Path#in()` methods don't specify the type variable, which seems fishy to me. It IMO just doesn't make sense to compare an expression of type `Integer` with an expression of type `Participation`. The easy fix for you is to just remove the `.get(Participation_.id)` part. We have to discuss this with the team, but -1 from me for continuing to support this IMO ill-typed way of doing things. These sort of special cases complicate things in the translation code all over the place for IMO no benefit. The proper way is even shorter for users to write.

I have no strong opinion on this, but it's *definitely* wrong that the user experiences a `ClassCastException` instead of a meaningful error, so *something* needs to be fixed here.

beikov:
Closing this in favor of https://github.com/hibernate/hibernate-orm/pull/6189 since we will rather make sure we drop support for this unsafe error prone comparison of entity valued paths against basic valued paths/literals

gavinking:
> drop support for this unsafe error prone comparison of entity valued paths against basic valued paths/literals

is it documented in the migration guide?

beikov:
@dreab8 will document it as part of https://github.com/hibernate/hibernate-orm/pull/6189

PR #5801, opened 2022-12-19
HHH-15802 SubQuery with "in" results in java.lang.ClassCastException: class org.hibernate.metamodel.mapping.internal.BasicEntityIdentifierMappingImpl cannot be cast to class org.hibernate.metamodel.mapping.EntityValuedModelPart

dreab8:


beikov:
Closing this in favor of https://github.com/hibernate/hibernate-orm/pull/6189 since we will rather make sure we drop support for this unsafe error prone comparison of entity valued paths against basic valued paths/literals

PR #5802, opened 2022-12-19
 HHH-15842 Convert simple foreign key values before binding

kkoniuszy:
In [HHH-15842](https://hibernate.atlassian.net/browse/HHH-15842), selecting and persisting entities which have a many-to-one relationship to another entity whose primary key is an enum fails because the foreign key values are not converted to the appropriate JDBC type, thus at some point parameter binding fails with a ClassCastException or an AssertionError in JdbcParameterBindingImpl:20.

This commit modifies SimpleForeignKeyDescriptor so that disassemble, forEachDisassembledJdbcValue, breakDownJdbcValues and forEachJdbcValue will attempt to convert the value to a relational one using the appropriate JdbcMapping

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/SimpleForeignKeyDescriptor.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/SimpleForeignKeyDescriptor.java
@@ -424,7 +424,7 @@ public class SimpleForeignKeyDescriptor implements ForeignKeyDescriptor, BasicVa
 
 	@Override
 	public Object disassemble(Object value, SharedSessionContractImplementor session) {
-		return value;
+		return getJdbcMapping().convertToRelationalValue( value );
 	}
 
 	@Override
@@ -460,7 +460,7 @@ public class SimpleForeignKeyDescriptor implements ForeignKeyDescriptor, BasicVa
 
 	@Override
 	public void breakDownJdbcValues(Object domainValue, JdbcValueConsumer valueConsumer, SharedSessionContractImplementor session) {
-		valueConsumer.consume( domainValue, keySide.getModelPart() );
+		valueConsumer.consume( disassemble( domainValue, session ), keySide.getModelPart() );
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/SimpleForeignKeyDescriptor.java
```diff
@@ -454,13 +454,13 @@ public int forEachDisassembledJdbcValue(
 			int offset,
 			JdbcValuesConsumer valuesConsumer,
 			SharedSessionContractImplementor session) {
-		valuesConsumer.consume( offset, value, getJdbcMapping() );
+		valuesConsumer.consume( offset, disassemble( value, session ), getJdbcMapping() );
```
beikov, line 457: Please revert this. The passed in value already has to be in disassembled form.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/SimpleForeignKeyDescriptor.java
```diff
@@ -506,7 +506,8 @@ public int forEachJdbcValue(
 			int offset,
 			JdbcValuesConsumer valuesConsumer,
 			SharedSessionContractImplementor session) {
-		valuesConsumer.consume( offset, value, targetSide.getModelPart().getJdbcMapping() );
+		JdbcMapping jdbcMapping = targetSide.getModelPart().getJdbcMapping();
+		valuesConsumer.consume( offset, jdbcMapping.convertToRelationalValue( value ), jdbcMapping );
```
beikov, line 510:
```suggestion
		valuesConsumer.consume( offset, disassemble( value, session ), jdbcMapping );
```

beikov (changes requested):
After fixing the things I mentioned, this can go in!

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/hhh15842/EnumForeignKeyParameterMappingTest.java
```diff
@@ -0,0 +1,195 @@
...
+		A, B
+	}
+
+	@Entity(name = "EntityWithEnumForeignKey")
+	@Table(name = "test_entity_enum_fk")
+	public static class EntityWithEnumForeignKey {
+
+		@Id
+		@Column(name = "id")
+		private UUID id;
```
beikov, line 120: Please replace uses of `UUID` with `String` or `Long`. Sybase has some issues with trailing zeros and random UUIDs in other tests already cause intermittent failures. We don't need another test that does the same thing ;)

beikov (approved):

beikov:
Thanks a lot!

PR #5807, opened 2022-12-20
rename Generator subtypes

gavinking:
and also handle multi-property `@NaturalId`s in generator stuff


beikov:
How about the name `PreFlushGenerator`?

gavinking:
> How about the name `PreFlushGenerator`?

Hrm. I don't hate it. 

OK, thanks, let me let that suggestion stew for a bit.

PR #5815, opened 2022-12-20
HHH-15868 Expose more standard HQL functions through HibernateCriteriaBuilder

mbladel:
https://hibernate.atlassian.net/browse/HHH-15868

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
@@ -356,6 +356,14 @@ public class DB2Dialect extends Dialect {
 				new DB2FormatEmulation( queryEngine.getTypeConfiguration() )
 		);
 
+		queryEngine.getSqmFunctionRegistry().patternDescriptorBuilder( "atan2", "atan2(?2,?1)" )
+				.setInvariantType(
+						queryEngine.getTypeConfiguration().getBasicTypeRegistry().resolve( StandardBasicTypes.DOUBLE )
+				)
+				.setExactArgumentCount( 2 )
+				.setParameterTypes( FunctionParameterType.NUMERIC, FunctionParameterType.NUMERIC )
+				.register();
+
 		queryEngine.getSqmFunctionRegistry().namedDescriptorBuilder( "posstr" )
 				.setInvariantType(
 						queryEngine.getTypeConfiguration().getBasicTypeRegistry().resolve( StandardBasicTypes.INTEGER )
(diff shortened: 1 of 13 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmCriteriaNodeBuilder.java
```diff
@@ -2569,6 +2571,498 @@ private static SessionFactory locateSessionFactoryOnDeserialization(String uuid,
...
+				sqmArguments,
+				getTypeConfiguration().getBasicTypeForJavaType( type ),
+				queryEngine,
+				getJpaMetamodel().getTypeConfiguration()
+		);
+	}
+
+	@Override
+	public SqmFunction<String> format(Expression<? extends TemporalAccessor> datetime, String pattern) {
+		SqmExpression<String> patternLiteral = value( pattern );
```
beikov, line 2591: I actually believe that this must be a literal:

```suggestion
		SqmExpression<String> patternLiteral = literal( pattern );
```

beikov (commented):
Great work!

beikov:
Look into `FunctionTests` which shows how some tests need to be skipped on certain databases. You need to add some `@RequiresDialectFeature` annotations.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmCriteriaNodeBuilder.java
mbladel: Actually it had to be a `SqmFormat` to allow for dialect translation of the pattern, now it's fixed.

mbladel:
Thank you @beikov, I added dialect checks to tests and made some other small adjustments.

hibernate-core/src/main/java/org/hibernate/dialect/DB2Dialect.java
```diff
@@ -356,6 +357,11 @@ public void initializeFunctionRegistry(QueryEngine queryEngine) {
 				new DB2FormatEmulation( queryEngine.getTypeConfiguration() )
 		);
 
+		queryEngine.getSqmFunctionRegistry().register(
+				"atan2",
+				new DB2Atan2Function( queryEngine.getTypeConfiguration() )
+		);
```
mbladel, line 363: While testing I noticed that DB2 requires the parameters to the `atan2` function in reverse order. This change fixes it, don't know if there's a better/more standard way of doing this.

beikov: Usually we use a pattern for this purpose:

```
queryEngine.getSqmFunctionRegistry().patternDescriptorBuilder("atan2", "atan2(?2,?1)")
	.setInvariantType( queryEngine.getTypeConfiguration().getBasicTypeRegistry().resolve( StandardBasicTypes.DOUBLE ) )
	.setExactArgumentCount( 2 )
	.setParameterTypes(NUMERIC, NUMERIC )
	.register();
```

mbladel: Thanks for the tip, I thought there was a better way to do it. Updated it now.

beikov (approved):

PR #5818, opened 2022-12-21
HHH-15919 make trunc() / truncate() portable in HQL

gavinking:
Still non-"standard" because tricky to implement on Sybase and Derby.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
@@ -263,11 +263,82 @@ public class CommonFunctionFactory {
 
 	public void trunc() {
 		functionRegistry.namedDescriptorBuilder( "trunc" )
+				.setReturnTypeResolver( useArgType( 1 ) )
 				.setArgumentCountBetween( 1, 2 )
 				.setParameterTypes(NUMERIC, INTEGER)
-				.setInvariantType(doubleType)
 				.setArgumentListSignature( "(NUMERIC number[, INTEGER places])" )
 				.register();
+		functionRegistry.registerAlternateKey( "truncate", "trunc" );
+	}
+
+	/**
+	 * MySQL
+	 */
+	public void trunc_truncate() {
+		functionRegistry.registerUnaryBinaryPattern(
+				"trunc",
+				"truncate(?1,0)",
+				"truncate(?1,?2)",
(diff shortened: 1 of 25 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
```diff
@@ -268,6 +268,82 @@ public void trunc() {
...
+				"sign(?1)*floor(abs(?1))",
+				"sign(?1)*floor(abs(?1)*power(10,?2))/power(10,?2)",
+				NUMERIC, INTEGER,
+				typeConfiguration
+		).setArgumentListSignature( "(NUMERIC number[, INTEGER places])" );
+		functionRegistry.registerAlternateKey( "truncate", "trunc" );
+	}
+
+	/**
+	 * PostgreSQL (only works if the second arg is constant, as it almost always is)
```
beikov, line 320: PostgreSQL has a two arg `trunc` method. Am I missing the reason why you added this?
See https://www.postgresql.org/docs/8.1/functions-math.html

gavinking: The two-arg form only accepts `numeric`. It's right there in that link you posted.

beikov: Did you try it? The docs are sometimes misleading with the types they name.

gavinking: > ERROR:  function round(double precision, integer) does not exist

beikov: Hm you are right, but should we maybe render something different depending on the type? Seems "wasteful" to me to do all that `floor` and division stuff also for `numeric` typed expressions especially since round/trunc will mostly be done just on `numeric` typed columns anyway.

gavinking: I'm not motivated to spend time on that.

gavinking: > I'm not motivated to spend time on that.

(Since I care more about scientific computing than about financial computing, I care more about `double` than I do about `numeric` haha.)

PR #5827, opened 2022-12-21
HHH-15914 - fix native query when the result class is Object or Object[]

jrenaat:


csware:
Shouldn't there be a test to prevent regressions in the future?

PR #5830, opened 2022-12-22
HHH-15926 allow @Type and @CompositeType to be applied as meta-annotations

gavinking:
this was almost zero work, the implementation was already there


beikov (approved):
LGTM

gavinking:
> LGTM

Yeah, I think I may as well merge it. It's useful, and completely harmless. It doesn't support everything described in https://github.com/hibernate/hibernate-orm/discussions/5716, but that stuff can *easily* be added later with no breakage.

PR #5831, opened 2022-12-22
HHH-15901 Criteria Query throws a NullPointerException when I use an empty predicates list

mbladel:
https://hibernate.atlassian.net/browse/HHH-15901

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/select/SqmQuerySpec.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/select/SqmQuerySpec.java
@@ -354,15 +354,23 @@ public class SqmQuerySpec<T> extends SqmQueryPart<T>
 
 	@Override
 	public SqmQuerySpec<T> setRestriction(Predicate... restrictions) {
-		SqmWhereClause whereClause = getWhereClause();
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/spi/BaseSemanticQueryWalker.java
```diff
@@ -436,7 +436,7 @@ public Object visitValues(SqmValues values) {
 
 	@Override
 	public Object visitWhereClause(SqmWhereClause whereClause) {
-		if ( whereClause == null ) {
+		if ( whereClause == null || whereClause.getPredicate() == null ) {
```
beikov, line 439: IMO instead of doing a null check here again, we should rather make sure that the predicate in the `SqmWhereClause` can never become null by setting the `whereClause` in the `SqmQuerySpec` to `null` instead of setting a `null` `SqmPredicate`. Can you please adapt the set methods instead?

mbladel: Updated the logic to check for `null` and empty arrays when setting multiple predicates.

PR #5851, opened 2022-12-28
HHH-15942 introduce ForcedFlushMode 

gavinking:
for specifying whether a query flushes or not:

- replaces `FlushModeType` in the annotation package
- much less confusing when applied to a `Query`
  * what do `MANUAL` and `COMMIT` mean for a `Query`?
  * how is `AUTO` useful for a` Query`?

- also make `Query.getHibernateFlushMode()` obey its documented semantics by returning the session flush mode instead of null when unset

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/ForcedFlushMode.java
+++ b/hibernate-core/src/main/java/org/hibernate/ForcedFlushMode.java
@@ -1,0 +1,33 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later.
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate;
+
+/**
+ * Enumerates the possible flush modes for execution of a
+ * {@link org.hibernate.query.Query}. A "forced" flush mode
+ * overrides the {@linkplain Session#getHibernateFlushMode()
+ * flush mode of the session}.
+ *
+ * @author Gavin King
+ *
+ * @since 6.2
+ */
+public enum ForcedFlushMode {
+	/**
+	 * Flush before executing the query.
+	 */
+	FORCE_FLUSH,
+	/**
+	 * Do not flush before executing the query.
+	 */
+	FORCE_NO_FLUSH,
+	/**
+	 * Let the owning {@link Session session} decide whether
+	 * to flush, depending on its {@link FlushMode}.
+	 */
+	NO_FORCING
+}
--- a/hibernate-core/src/main/java/org/hibernate/query/Query.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/Query.java
@@ -17,6 +17,7 @@ import java.util.stream.Stream;
 
 import org.hibernate.CacheMode;
 import org.hibernate.FlushMode;
+import org.hibernate.ForcedFlushMode;
 import org.hibernate.Incubating;
 import org.hibernate.LockMode;
 import org.hibernate.LockOptions;
@@ -854,6 +855,9 @@ public interface Query<R> extends SelectionQuery<R>, MutationQuery, TypedQuery<R
 
 	@Override
 	Query<R> setHibernateFlushMode(FlushMode flushMode);
+
+	@Override
+	Query<R> setForcedFlushMode(ForcedFlushMode forcedFlushMode);
 
 	@Override
 	Query<R> setCacheable(boolean cacheable);
--- a/hibernate-core/src/main/java/org/hibernate/Session.java
+++ b/hibernate-core/src/main/java/org/hibernate/Session.java
@@ -148,7 +148,7 @@ public interface Session extends SharedSessionContract, EntityManager {
 	void flush();
 
 	/**
-	 * Set the current {@link FlushModeType JPA flush mode} for this session.
+	 * Set the current {@linkplain FlushModeType JPA flush mode} for this session.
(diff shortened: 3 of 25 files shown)


hibernate-core/src/main/java/org/hibernate/ForcedFlushMode.java
```diff
@@ -0,0 +1,33 @@
...
+	/**
+	 * Do not flush before executing the query.
+	 */
+	FORCE_NO_FLUSH,
+	/**
+	 * Let the owning {@link Session session} decide whether
+	 * to flush, depending on its {@link FlushMode}.
+	 */
+	NO_FORCING
+}
```
gavinking, line 33: Everyone, please take a look at this new thing!

sebersole:
> * what do `MANUAL` and `COMMIT` mean for a `Query`?

Both of those are pseudonyms for never wrt a Query

> * how is `AUTO` useful for a` Query`?

AUTO is actually the default for a Query.  Not sure what you are asking there tbh

sebersole:
So iiuc this basically adds a whole new enum that just limits the set of values from another enum for use in a specific context.

gavinking:
> Both of those are pseudonyms for never wrt a Query

Yes of course, *I* know that, but it's nonobvious to a casual onlooker.

> `AUTO` is actually the default for a Query. Not sure what you are asking there tbh

What I'm saying is there's too many things there and it's hard for the user to pick the right one. If I'm going to set this at the level of a query (rather than at the level of the session, where it all of the values are useful) then I'm setting it because I either want to:

- force a flush, or
- force it _not_ to flush.

hibernate-core/src/main/java/org/hibernate/query/Query.java
```diff
@@ -855,6 +856,9 @@ default Query<R> applyLoadGraph(@SuppressWarnings("rawtypes") RootGraph graph) {
 	@Override
 	Query<R> setHibernateFlushMode(FlushMode flushMode);
```
sebersole, line 857: Is there a reason this is not deprecated?  It seems to me the intent here is to replace `FlushMode` when used in the context of a Query

gavinking: I could deprecate it if you prefer.

gavinking:
> So iiuc this basically adds a whole new enum that just limits the set of values from another enum for use in a specific context.

It _replaces_ `org.hibernate.annotations.FlushModeType`, which takes all the values of `org.hibernate.FlushMode`, _and adds an additional value_.

hibernate-core/src/main/java/org/hibernate/ForcedFlushMode.java
sebersole: Given that this is solely about queries, I'd be more inclined to name this `QueryFlushMode` or move it into `o.h.query`.  For `QueryFlushMode`, I'd go with:

```
enum QueryFlushMode {
    FLUSH,
    SKIP,
    /**
     * Defer to {@link Session#getHibernateFlushMode}
     */
    AUTO
}
```

I get why you named it as you did.  I just find it confusing

gavinking:
I mean, to clarify, my first draft of this was to just deprecate `FlushModeType` _with no replacement_, but then I got to thinking about it and decided that this was nicer.

hibernate-core/src/main/java/org/hibernate/ForcedFlushMode.java
gavinking: 1. `QueryFlushMode` is fine by me, it was also my first thought.
2. You're right, I agree it belongs in `o.h.query`.
3. `FLUSH` is fine.
4. `SKIP` I don't find _extremely_ clear though I could live with it.
5. I don't like `AUTO` here because it doesn't have the same semantics as `FlushMode.AUTO`.

Sanne:
Should we clarify how and if such new options are affected by the configuration properties?

For example we have property `org.hibernate.flushMode`, I think at least its documentation should say something about how it interacts with the "query flush modes".

gavinking:
@Sanne you mean the query hint? That is unaffected by this proposal because it's mostly a JPA thing. When using Hibernate-native APIs we don't need hints.

I think there's also a config property but IIRC it's just a default value for that hint. (And I don't consider it a useful thing.)

This proposal is very much only affecting the native APIs.

Sanne:
I'm referring to the configuration property `AvailableSettings#FLUSH_MODE`, AFAIK it changes the `FlushMode`, impacting all Session(s) - I'm not sure what you mean by "mostly a JPA thing"?

If they have no relation that's totally fine, but I think that should be made clear.

gavinking:
> I'm referring to the configuration property `AvailableSettings#FLUSH_MODE`, AFAIK it changes the `FlushMode`, impacting all Session(s) - I'm not sure what you mean by "mostly a JPA thing"?

Yeah, OK, you're right, it is actually a session-level setting, not a query hint, even though:

- it _looks like a query hint_, since it doesn't satisfy our naming convention for configuration properties, and
- it's only _documented as a query hint_.

Actually I think we should deprecate this thing, I don't see how it's useful, and honestly it almost looks like it got there by accident.

> If they have no relation that's totally fine, but I think that should be made clear.

This proposal has no relation _at all_ to the session-level flush mode, that's correct.

gavinking:
> Actually I think we should deprecate this thing

https://github.com/hibernate/hibernate-orm/pull/5954

Actually it's fine, as a `EntityManager`-level JPA hint, which is what it started out as. It just should exist as a Hibernate config property in `AvailableSettings`.

gavinking:
superseded by #8952

PR #5853, opened 2022-12-28
HHH-15945 delete some obsolete code + move 2 interfaces b/w SPI packages

gavinking:
The sensitive bit here is the move of `TransactionObserver` and `IsolationDelegate` to the same package as `TransactionCoordinator`.

The deleted code is stuff that is already unused since H5, and so nobody can possibly be depending on it.


gavinking:
I think I will merge this soon, but I need you guys to let me know if it's a problem to break custom implementations of `TransactionCoordinator`. I assume that no such custom implementations exist, but I might be wrong!

gavinking:
It's easy enough to move them back to `org.hibernate.engine.transaction.spi` if that part of this change is determined to be a mistake.

PR #5862, opened 2022-12-30
add a SQLFunctionTemplate back in to ease migration

gavinking:
People have run into the lack of this.


gavinking:
WDYT folks? Should we do this DYT?

beikov:
Sounds ok to me, but ultimately, I think many custom registrations might have become obsolete since we have "standard" equivalents already registered. So maybe we should let people run into this to let them think about this a bit more?

gavinking:
> Sounds ok to me, but ultimately, I think many custom registrations might have become obsolete since we have "standard" equivalents already registered. 

Yeah that's what I thought too, until stackoverflow disabused me of this notion.

gavinking:
Looks like we're never going to do this. Or if we do one day define a simpler way to do it, it won't be the old `SQLFunctionTemplate`.

PR #5863, opened 2022-12-30
HHH-15604 HHH-15966 Identically-named association in entity root and elementcollection of embeddables leads to assertion error

dreab8:
https://hibernate.atlassian.net/browse/HHH-15604
https://hibernate.atlassian.net/browse/HHH-15966

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
@@ -1009,7 +1009,7 @@ public class ToOneAttributeMapping
 			}
 			return false;
 		}
-		else if ( parentNavigablePath.getParent() != null && creationState.resolveModelPart( parentNavigablePath.getParent() ) instanceof EmbeddedCollectionPart ) {//todo: handle recursively?
+		else if ( isParentEmbeddedCollectionPart( creationState, parentNavigablePath.getParent() ) ) {
 			/*
 				class EntityA{
 					@OneToOne(mappedBy = "identicallyNamedAssociation", fetch = FetchType.EAGER)
@@ -1079,6 +1079,22 @@ public class ToOneAttributeMapping
 			return false;
 		}
 		return parentNavigablePath.isSuffix( bidirectionalAttributePath );
+	}
+
+	private boolean isParentEmbeddedCollectionPart(DomainResultCreationState creationState, NavigablePath parentNavigablePath) {
+		while ( parentNavigablePath != null ) {
+			final ModelPart parentModelPart = creationState.resolveModelPart( parentNavigablePath );
+			if ( parentModelPart instanceof EmbeddedCollectionPart ) {
+				return true;
+			}
+			else if ( parentModelPart instanceof EmbeddableValuedModelPart ) {
+				parentNavigablePath = parentNavigablePath.getParent();
+			}
+			else {
+				return false;
+			}
+		}
+		return false;
 	}
 
 	private Fetch createCircularBiDirectionalFetch(
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
@@ -24,6 +24,7 @@ import org.hibernate.internal.util.IndexedConsumer;
 import org.hibernate.internal.util.StringHelper;
 import org.hibernate.internal.util.collections.ArrayHelper;
 import org.hibernate.mapping.Collection;
+import org.hibernate.mapping.Component;
 import org.hibernate.mapping.Join;
 import org.hibernate.mapping.ManyToOne;
 import org.hibernate.mapping.OneToOne;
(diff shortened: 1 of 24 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -945,7 +949,31 @@ else if ( parentModelPart instanceof EntityCollectionPart ) {
 			}
 			return false;
 		}
-		if ( cardinality == Cardinality.MANY_TO_ONE ) {
+		else if ( parentNavigablePath.getParent() != null && creationState.resolveModelPart( parentNavigablePath.getParent() ) instanceof EmbeddedCollectionPart ) {
+			/*
+				class EntityA{
+					@OneToOne(mappedBy = "identicallyNamedAssociation", fetch = FetchType.EAGER)
```
beikov, line 955: Shouldn't this example rather use `mappedBy = "embeddable.identicallyNamedAssociation"`? Also, is there a test for this example? I kind of have the feeling that your parent check `declaringTableGroupProducer.getNavigableRole().getLocalName().equals( oneToOne.getReferencedEntityName() )` should be already enough and that we don't need this extra branch, or am I missing something?

dreab8: i don't think we support `mappedBy = "embeddable.identicallyNamedAssociation"`

dreab8: the test has a `mappedby` referring to the  [association](https://github.com/hibernate/hibernate-orm/pull/5863/files#diff-fd953c2b59c78a090f7baad15ed677fcc41d15fbd3bb169f9592b90c6b7c8d65R151) not the of the embeddable.
Without the extra checking the test fails at [assertThat( a2.getB() ).isEqualTo( b2 )](https://github.com/hibernate/hibernate-orm/pull/5863/files#diff-fd953c2b59c78a090f7baad15ed677fcc41d15fbd3bb169f9592b90c6b7c8d65R68)

beikov: >i don't think we support mappedBy = "embeddable.identicallyNamedAssociation"

Hmm, we should though IMO 😄 

>the test has a mappedby referring to the association not the of the embeddable.

Ok, I didn't realize that this is a separate method. What happens when the association is part of an embeddable within an element collection though? What I'm trying to get to is, I think this needs a "recursive" implementation that checks parents, continuing the search when an embeddable is encountered.

yrodiere: > i don't think we support mappedBy = "embeddable.identicallyNamedAssociation"

We actually do, I've used that in ORM 5.

ORM 5 even supports this with an element collection, e.g. `mappedBy = "elementCollection.identicallyNamedAssociation"`...

yrodiere: Well there's this example in the javadoc of `@OneToOne`:

```

 
     Example 3: One-to-one association from an embeddable class to another entity.
 
     @Entity
     public class Employee {
        @Id int id;
        @Embedded LocationDetails location;
        ...
     }
 
     @Embeddable
     public class LocationDetails {
        int officeNumber;
        @OneToOne ParkingSpot parkingSpot;
        ...
     }
 
     @Entity
     public class ParkingSpot {
        @Id int id;
        String garage;
        @OneToOne(mappedBy="location.parkingSpot") Employee assignedTo;
         ... 
     } 
```

So I'd say JPA requires it...

yrodiere: Also, in the same javadoc:

> The value of each identifier used with the dot notation is the name of the respective embedded field or property.

gavinking: Ugh, sorry, my bad, there are indeed multiple examples of this in the spec. I didn't remember this and I didn't check.

OK, good, so it *is* documented.

Sorry for the noise.

dreab8: added also the test proposed by Yoann

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -244,7 +247,8 @@ && equal( join.getKey(), manyToOne ) ) {
 							final Value value = property.getValue();
 							if ( value instanceof OneToOne ) {
 								final OneToOne oneToOne = (OneToOne) value;
-								if ( name.equals( oneToOne.getMappedByProperty() )
+								if (declaringTableGroupProducer.getNavigableRole().getLocalName().equals( oneToOne.getReferencedEntityName() )
+								&& name.equals( oneToOne.getMappedByProperty() )
```
beikov, line 251: I have the feeling that this will miss out on properties that are part of embeddables i.e. we will never detect the `bidirectionalAttributeName` for a `@OneToOne` that uses e.g. `mappedBy = "embeddable.abc"` because `name` will never contain the `.`. IMO this should rather use the local name of the navigable role to determine the bidirectional "path". That thing possibly being a path is the second thing I'm concerned with. Our current uses of `bidirectionalAttributeName` do not take into account that this could be a path i.e. we should use `findByPath` instead of `findSubPart`. Also, some of the current `isBidirectionalAttributeName()` logic assumes that it can compare `bidirectionalAttributeName` against a `NavigablePath#getLocalName()`, which isn't the case if we admit that it is a path.
I suggest we rather model this as `SelectablePath bidirecationalAttributePath` and adapt implementations accordingly.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -954,7 +1009,31 @@ else if ( parentModelPart instanceof EntityCollectionPart ) {
 			}
 			return false;
 		}
-		if ( cardinality == Cardinality.MANY_TO_ONE ) {
+		else if ( parentNavigablePath.getParent() != null && creationState.resolveModelPart( parentNavigablePath.getParent() ) instanceof EmbeddedCollectionPart ) {//todo: handle recursively?
```
beikov, line 1012: We are still missing a test for this todo comment, but after that, we should be good to go. The test case would involve a `*ToOne` association mapping within an embeddable within an element collection. Also, the code comment in this if branch does not seem to fit the condition `instanceof EmbeddedCollectionPart` which makes it slightly confusing.

PR #5864, opened 2022-12-30
make Dialect implement TypeContributor and FunctionContributor

gavinking:
There is little *concrete* benefit to this change, and it's slightly break-y but:

- it removes a dependence on the `@Incubating` type `QueryEngine` from `Dialect`
- we eat our own dogfood, that is, use our own abstractions
- it appears that this was how it was *originally* intended to be

... and, I guess I just *like* it.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -46,7 +46,10 @@ import org.hibernate.LockMode;
 import org.hibernate.LockOptions;
 import org.hibernate.ScrollMode;
 import org.hibernate.boot.TempTableDdlTransactionHandling;
+import org.hibernate.boot.model.FunctionContributions;
+import org.hibernate.boot.model.FunctionContributor;
 import org.hibernate.boot.model.TypeContributions;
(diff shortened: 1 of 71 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -1420,8 +1435,14 @@ public String toString() {
 
 	// database type mapping support ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+
+	@Override
+	public void contribute(TypeContributions typeContributions, ServiceRegistry serviceRegistry) {
```
beikov, line 1440: I would prefer marking this as `final` so that people only have to override a the `contributeTypes` method in a custom dialect if necessary.

beikov:
It looks cleaner indeed, but let's ask @sebersole about how he feels about the breakage.

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
gavinking: Yeah sure, that's a good idea.

gavinking:
Closing in favor of #5893

PR #5865, opened 2022-12-31
HHH-15954 correct implementation of JPA SharedCacheMode for .hbm.xml

gavinking:
- even though I hate most of its members, I think `NONE` is pretty useful
- anyway we may as well make it work, since it's trivial

Let's see what breaks.


gavinking:
The tests all passed, so let's assume that this didn't break anything.

PR #5866, opened 2022-12-31
HHH-15839 CriteriaBuilder treat method on Path causes ClassCastException

dreab8:
https://hibernate.atlassian.net/browse/HHH-15839


beikov (approved):
LGTM

PR #5870, opened 2023-01-01
HHH-15958 much better impl of @RowId annotation

gavinking:
- the rowid pseudo-column and type are now determined automatically from Dialect
- works (after all these years) in Postgres (and also on h2)
- introduce RowIdJdbcType (not strictly necessary, but a nicety)


gavinking:
Well, this was my first drunk-coding in quite a while and it turned out pretty well, I think...

PR #5872, opened 2023-01-01
HHH-15958 support @RowId on DB2

gavinking:


hibernate-core/src/main/java/org/hibernate/dialect/DB2iDialect.java
```diff
@@ -149,4 +151,23 @@ protected <T extends JdbcOperation> SqlAstTranslator<T> buildTranslator(
...
 		};
 	}
+
+	// I speculate that this is a correct implementation of rowids for DB2 for i,
+	// just on the basis of the DB2 docs, but I currently have no way to test it
+	// Note that the implementation inherited from DB2Dialect for LUW will not work!
+
+	@Override
+	public String rowId(String rowId) {
+		return rowId.isEmpty() ? "rowid_" : rowId;
```
hosamaly, line 161: This doesn't work for merge statements on v7.5...

```
[SQL0205] Column ROWID_ not in table my_table
A column with the name ROWID_ does not exist in table or view my_table
```

beikov: Then please create a Jira ticket that describes what goes wrong.

hosamaly: Created [HHH-20069](https://hibernate.atlassian.net/browse/HHH-20069).
Thanks.

PR #5875, opened 2023-01-02
HHH-15962 Provide simple default impl for new members of `SchemaManagementTool` interface

marko-bekhta:
https://hibernate.atlassian.net/browse/HHH-15962

- make both `getSchemaTruncator/buildGenerationTargets` as default methods throwing `UnsupportedOperationException`

Also, a quick question on `buildGenerationTargets(..)` - do we want to expose it in the interface since the `SchemaCreatorImpl`/`SchemaDropperImpl`/`SchemaTruncatorImpl` that are using it all require the `HibernateSchemaManagementTool` internal implementation rather than the SPI interface?


yrodiere:
To clarify, we would need this to to test Hibernate Search against both ORM 6.2 and ORM 6.1.

Pretty please? :)

gavinking:
Fine by me.

PR #5882, opened 2023-01-03
almost eliminate the use of TypeConfiguration as a source of SessionFactory

gavinking:
This change helps clean up the initialization lifeycle, and is more typesafe.


gavinking:
Going to integrate this even though I have a strong suspicion that it will break Quarkus integration.

@Sanne can we talk about this change sometime tomorrow please?

PR #5884, opened 2023-01-03
HHH-15856 pollution mitigations

whitingjr:
Mitgations for Loadable, EntityPersister and EntityInitializer.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntityDelayedFetchInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntityDelayedFetchInitializer.java
@@ -24,8 +24,6 @@ import org.hibernate.spi.NavigablePath;
 import org.hibernate.sql.results.graph.AbstractFetchParentAccess;
 import org.hibernate.sql.results.graph.DomainResultAssembler;
 import org.hibernate.sql.results.graph.FetchParentAccess;
-import org.hibernate.sql.results.graph.embeddable.EmbeddableInitializer;
-import org.hibernate.sql.results.graph.entity.AbstractEntityInitializer;
 import org.hibernate.sql.results.graph.entity.EntityInitializer;
 import org.hibernate.sql.results.graph.entity.LoadingEntityEntry;
 import org.hibernate.sql.results.jdbc.spi.RowProcessingState;
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/jdbc/internal/JdbcValuesSourceProcessingStateStandardImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/jdbc/internal/JdbcValuesSourceProcessingStateStandardImpl.java
@@ -199,7 +199,7 @@ public class JdbcValuesSourceProcessingStateStandardImpl implements JdbcValuesSo
 						callback.invokeAfterLoadActions(
 								getSession(),
 								loadingEntityEntry.getEntityInstance(),
-								(Loadable) loadingEntityEntry.getDescriptor()
+								loadingEntityEntry.getDescriptor().asLoadable()
 						);
 					}
 				}
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ManagedMappingType.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/ManagedMappingType.java
@@ -10,6 +10,7 @@ import java.util.function.Consumer;
 
 import org.hibernate.internal.util.IndexedConsumer;
 import org.hibernate.persister.entity.AttributeMappingsList;
+import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.sql.results.graph.FetchableContainer;
 import org.hibernate.type.descriptor.java.JavaType;
 
@@ -110,5 +111,9 @@ public interface ManagedMappingType extends MappingType, FetchableContainer {
 			}
 		}
 		return false;
+	}
+
+	default EntityPersister asEntityPersister() {
+		return null;
 	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/EntityPersister.java
@@ -1022,4 +1022,13 @@ public interface EntityPersister extends EntityMappingType, RootTableGroupProduc
 	@Deprecated(since = "6.2")
 	String ENTITY_ID = "id";
 
+	@Deprecated
+	default org.hibernate.persister.entity.Loadable asLoadable() {
+		return null;
+	}
+
+	@Override
+	default EntityPersister asEntityPersister() {
+		return this;
+	}
 }
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/Loadable.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/Loadable.java
@@ -114,4 +114,10 @@ public interface Loadable extends EntityPersister {
 	 * @return The proper table alias for qualifying the given column.
 	 */
 	String getTableAliasForColumn(String columnName, String rootAlias);
+
+	@Override
+	default Loadable asLoadable() {
+		return this;
+	}
+
 }
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/Initializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/Initializer.java
@@ -81,7 +81,7 @@ public interface Initializer {
 			final EntityPersister concreteDescriptor = parentAccess.findFirstEntityInitializer()
 					.getConcreteDescriptor();
 			if ( concreteDescriptor.getEntityMetamodel().isPolymorphic() ) {
-				final EntityPersister declaringType = (EntityPersister) referencedModelPart.getDeclaringType();
+				final EntityPersister declaringType = referencedModelPart.getDeclaringType().asEntityPersister();
 				if ( concreteDescriptor != declaringType ) {
 					if ( !declaringType.getSubclassEntityNames().contains( concreteDescriptor.getEntityName() ) ) {
 						return false;
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
@@ -115,8 +115,8 @@ public abstract class AbstractEntityInitializer extends AbstractFetchParentAcces
 			AssemblerCreationState creationState) {
 		super();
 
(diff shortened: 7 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntityDelayedFetchInitializer.java
```diff
@@ -256,4 +256,7 @@
 		return true;
 	}
 
+	public EntityInitializer asEntityInitializer() {
```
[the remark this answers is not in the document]
whitingjr: corrected.

Sanne:
nitpick: finishing each commit description with a "." looks a bit odd :)

hibernate-core/src/test/java/org/hibernate/orm/test/cfg/persister/GoofyPersisterClassProvider.java
```diff
@@ -775,6 +775,16 @@ public void forEachAttributeMapping(Consumer<? super AttributeMapping> action) {
 		public JavaType getMappedJavaType() {
 			return null;
 		}
+
+		@Override
+		public org.hibernate.persister.entity.Loadable asLoadable() {
+			return ( org.hibernate.persister.entity.Loadable ) this;
```
Sanne, line 781: the cast should be unnecessary? That's the point of using this pattern.

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/ejb3configuration/PersisterClassProviderTest.java
```diff
@@ -794,6 +795,16 @@ public void forEachAttributeMapping(Consumer<? super AttributeMapping> action) {
 		public JavaType getMappedJavaType() {
 			return null;
 		}
+
+		@Override
+		public Loadable asLoadable() {
+			return ( Loadable ) this;
```
Sanne, line 801: cast is unnecessary?

hibernate-core/src/test/java/org/hibernate/orm/test/legacy/CustomPersister.java
```diff
@@ -866,4 +867,14 @@ public void forEachAttributeMapping(Consumer<? super AttributeMapping> action) {
 	public JavaType getMappedJavaType() {
 		return null;
 	}
+
+	@Override
+	public Loadable asLoadable() {
+		return ( Loadable ) this;
```
Sanne, line 873: cast should be unnecessary?

hibernate-core/src/main/java/org/hibernate/sql/results/graph/embeddable/AbstractEmbeddableInitializer.java
```diff
@@ -464,4 +464,9 @@ public void finishUpRow(RowProcessingState rowProcessingState) {
 	public String toString() {
 		return getClass().getSimpleName() + "(" + navigablePath + ") : `" + getInitializedPart().getJavaType().getJavaTypeClass() + "`";
 	}
+
+	@Override
+	public EntityInitializer asEntityInitializer() {
+		return ( EntityInitializer ) this;
```
Sanne, line 470: doesn't need a cast?

hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/AbstractCollectionInitializer.java
```diff
@@ -183,4 +184,9 @@ public CollectionKey resolveCollectionKey(RowProcessingState rowProcessingState)
 	public void finishUpRow(RowProcessingState rowProcessingState) {
 		collectionKey = null;
 	}
+
+	@Override
+	public EntityInitializer asEntityInitializer() {
+		return ( EntityInitializer ) this;
```
Sanne, line 190: doesn't need a cast?

hibernate-core/src/main/java/org/hibernate/sql/results/graph/AbstractFetchParentAccess.java
```diff
@@ -43,4 +45,9 @@ protected void notifyResolutionListeners(Object resolvedInstance) {
 
 		listeners.clear();
 	}
+
+	@Override
+	public EntityInitializer asEntityInitializer() {
+		return ( EntityInitializer ) this;
```
Sanne, line 51: doesn't need a cast?

hibernate-core/src/main/java/org/hibernate/query/derived/AnonymousTupleEntityValuedModelPart.java
```diff
@@ -693,4 +693,9 @@ public boolean isSimpleJoinPredicate(Predicate predicate) {
 				? ( (TableGroupJoinProducer) delegate ).isSimpleJoinPredicate( predicate )
 				: false;
 	}
+
+	@Override
+	public EntityPersister asEntityPersister() {
+		return ( EntityPersister ) this;
```
Sanne, line 699: doesn't need a cast?

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -6151,4 +6151,14 @@ public final boolean isAllNull(Object[] array, int tableNumber) {
 		}
 		return true;
 	}
+
+	@Override
+	public Loadable asLoadable() {
```
Sanne, line 6156: this might work, but I'd have preferred to add such overrides as default methods in the interface? Was there a reason to not do so?

That would also have avoided the need to implement them in the various test implementations.

Sanne (changes requested):

hibernate-core/src/test/java/org/hibernate/orm/test/cfg/persister/GoofyPersisterClassProvider.java
Sanne: Also this method is unnecessary if you added the `asLoadable` method on the `Loadable` interface.

Sanne:
A suggestion, this can be done in a simpler way:
 - https://github.com/hibernate/hibernate-orm/commit/fc6c2cbd05aabe47fd482cc8a2d8ea5089b00734

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/ejb3configuration/PersisterClassProviderTest.java
whitingjr: Intellij IDEA warning me again.

hibernate-core/src/test/java/org/hibernate/orm/test/legacy/CustomPersister.java
whitingjr: Intellij IDEA warning me again.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/embeddable/AbstractEmbeddableInitializer.java
whitingjr: Intellij IDEA warning me again.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/AbstractCollectionInitializer.java
whitingjr: Intellij IDEA warning me again.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/AbstractFetchParentAccess.java
whitingjr: Intellij IDEA warning me again.

hibernate-core/src/main/java/org/hibernate/query/derived/AnonymousTupleEntityValuedModelPart.java
whitingjr: Intellij IDEA warning me again.

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
whitingjr: Yes the reason is the Loadable instance type is not available at the EntityPersister interface implementation. At extended interfaces it is. Attempting this causes the compiler to report this as an error.

> error: incompatible types: EntityPersister cannot be converted to Loadable

hibernate-core/src/main/java/org/hibernate/sql/results/graph/collection/internal/AbstractCollectionInitializer.java
Sanne: I quickly checked this case: the warnings IDEA is showing you are correct because this type is NOT an `EntityInitializer`.
Seems something is wrong here, and that's why I'd prefer to not have the casts as they will point to real issues.

hibernate-core/src/main/java/org/hibernate/sql/results/jdbc/internal/JdbcValuesSourceProcessingStateStandardImpl.java
```diff
@@ -199,7 +199,7 @@
 						callback.invokeAfterLoadActions(
 								getSession(),
 								loadingEntityEntry.getEntityInstance(),
-								(Loadable) loadingEntityEntry.getDescriptor()
+								loadingEntityEntry.getDescriptor().asLoadable()
```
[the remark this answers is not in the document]
whitingjr: setting the new method EntityPersister.asLoadable() as deprecated is tripping up the CodeQL. Remove the annotation ?

Sanne:
I've rebased it, and then removed all the overrides you don't actually need to make it simpler.

And now that it's simpler and I can understand what you're doing, I see it's adding a couple of public methods directly on the public API :(  not sure if that's a good idea. 

@sebersole what do you think of adding the following two methods on the `EntityPersister` API ?

```
       default org.hibernate.persister.entity.Loadable asLoadable() {
		return null;
	}

	@Override
	default EntityPersister asEntityPersister() {
		return this;
	}
```

these are methods which are only needed for technical reasons as a workaround for a JDK performance issue. They aren't of interest to end users, so it bothers me a bit to have them there.

hibernate-core/src/main/java/org/hibernate/query/derived/AnonymousTupleEntityValuedModelPart.java
Sanne: well IDEA was right, this cast can't possibly work. I've removed the method, and as expected all tests are still passing: that's because this isn't ever invoked, otherwise you'd have seen a classcast exception.

Same for all other IDEA warnings.

beikov:
Closing since the Jira issue was closed.

PR #5886, opened 2023-01-04
HHH-15971 Fix Oracle CI build for 11g version

loiclefevre:
This PR fixes nightly 11g build:
- by forcing the oracle_legacy_ci profile hence targeting XE database service name (instead of xepdb1 when using oracle_ci profile)
- it also optimizes this build by disabling the recycle bin and increasing the session_cached_cursors parameter (from 50 to 500)


beikov:
Thanks!

PR #5887, opened 2023-01-04
HHH-15866 Hibernate validation fails when OneToMany refers to fereign key in embeddable object and is marked as nullable false

mbladel:
Backport for https://github.com/hibernate/hibernate-orm/pull/5879


beikov (approved):
LGTM

PR #5889, opened 2023-01-04
Fix identity select statement caching

beikov:


hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -833,15 +833,15 @@ else if ( batchSize > 1 ) {
 	}
 
 	private String getIdentitySelectString(Dialect dialect) {
-		if ( dialect.getIdentityColumnSupport().supportsInsertSelectIdentity() ) {
```
gavinking, line 836: Really, what was wrong with that?

gavinking: Oh I see. Yeah, it's wrong.

beikov: See https://ci.hibernate.org/job/hibernate-orm-nightly/job/main/78/testReport/junit/org.hibernate.orm.test.id/IdentityIdEntityTest/Build___h2_1_4___Test___testIdentityEntityWithDisabledGetGeneratedKeys_SessionFactoryScope_/

gavinking: Oops. Nice catch.

PR #5890, opened 2023-01-04
sort out an issue with function argument validation

gavinking:
This is something I just didn't want to deal with last night.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/produce/function/ArgumentTypesValidator.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/produce/function/ArgumentTypesValidator.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.query.sqm.produce.function;
 
+import org.hibernate.HibernateException;
 import org.hibernate.QueryException;
 import org.hibernate.metamodel.MappingMetamodel;
 import org.hibernate.metamodel.mapping.JdbcMapping;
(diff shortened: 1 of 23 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/produce/function/ArgumentTypesValidator.java
```diff
@@ -128,43 +141,63 @@ public void validate(
...
+		return jdbcType.getDefaultSqlTypeCode();
 	}
 
-	private JdbcMapping getJdbcMapping(SqmTypedNode<?> argument, MappingMetamodel metamodel) {
+	private JdbcMapping getJdbcMappingForEnum(SqmTypedNode<?> argument, TypeConfiguration typeConfiguration) {
 		if ( argument instanceof SqmPath<?> ) {
 			final SqmPath<?> path = (SqmPath<?>) argument;
-			final ModelPartContainer modelPartContainer = getModelPartContainer( path.getLhs(), metamodel );
+			final ModelPartContainer modelPartContainer = getModelPartContainer( path.getLhs(), typeConfiguration );
 			final ModelPart part = modelPartContainer.findSubPart( path.getReferencedPathSource().getPathName(), null );
```
dreab8: I may be wrong but probably we have have to check that `modelPartContainer` is not null before calling `findSubPart`

gavinking: @dreab8 in the end that code got removed in a later commit.

dreab8: sorry for the noise

PR #5893, opened 2023-01-04
make Dialect implement TypeContributor and FunctionContributor

gavinking:
Rebase of https://github.com/hibernate/hibernate-orm/pull/5864.

One benefit of this change is that it allows the lifecycle of the `SqmFunctionRegistry` to be decoupled from instantiation of the `SessionFactoryImpl` and `QueryEngine`.

This going to be useful if we want to fully-support `FunctionContributor`s the way we support `TypeContributor`s.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
@@ -47,6 +47,10 @@ public interface FunctionContributor {
 	 * <p>
 	 * The range 0-500 is reserved for Hibernate, range 500-1000 for libraries and
 	 * 1000-Integer.MAX_VALUE for user-defined FunctionContributors.
+	 * <p>
+	 * Contributions from higher precedence contributors (higher numbers) effectively override 
+	 * contributions from lower precedence.  E.g. if a contributor with precedence 1000 contributes a 
+	 * function named {@code "max"}, that will override Hibernate's standard function of that name.
 	 *
 	 * @return the ordinal for this FunctionContributor
 	 */
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
@@ -35,17 +35,18 @@ package org.hibernate.boot.model;
 public interface FunctionContributor {
 
 	/**
-	 *  Contribute functions
+	 * Contribute functions
 	 *
 	 * @param functionContributions The target for the contributions
 	 */
 	void contributeFunctions(FunctionContributions functionContributions);
 
 	/**
-	 * Determines order in which the contributions will be applied (lowest ordinal first).
-	 *
-	 * The range 0-500 is reserved for Hibernate, range 500-1000 for libraries and 1000-Integer.MAX_VALUE for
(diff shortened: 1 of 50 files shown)


sebersole:
> One benefit of this change is that it allows the lifecycle of the `SqmFunctionRegistry` to be decoupled from instantiation of the `SessionFactoryImpl` and `QueryEngine`.

Awesome!  This was what I actually wanted originally but ran into temporary problems and never got back to it - so huge +1 to the principle.

gavinking:
@sebersole  Alright then could you take a quick look at this (the real diff is tiny, if you ignore the impact on `Dialect` classes themselves, and let me know if I can merge it?

hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
```diff
@@ -35,17 +35,18 @@
...
 	/**
-	 * Determines order in which the contributions will be applied (lowest ordinal first).
-	 *
-	 * The range 0-500 is reserved for Hibernate, range 500-1000 for libraries and 1000-Integer.MAX_VALUE for
-	 * user-defined FunctionContributors.
+	 * Determines order in which the contributions will be applied
+	 * (lowest ordinal first).
+	 * <p>
+	 * The range 0-500 is reserved for Hibernate, range 500-1000 for libraries and
+	 * 1000-Integer.MAX_VALUE for user-defined FunctionContributors.
```
sebersole, line 49: Since we are touching this, I think we should improve this even more to highlight the fact that higher numbers take precedence (aka will override previous registrations).  Its not immediately evident imo

hibernate-core/src/main/java/org/hibernate/query/spi/QueryEngine.java
```diff
@@ -103,8 +103,7 @@ private static SqmFunctionRegistry createFunctionRegistry(
 			contributor.contributeFunctions( functionContributions );
 		}
 
-		// can't move this here just yet!
-		//dialect.initializeFunctionRegistry( this );
+		dialect.initializeFunctionRegistry( functionContributions );
```
sebersole, line 106: Yaay!

sebersole (commented):
lgtm

Added one comment not strictly about this PR, but just an observation

hibernate-core/src/main/java/org/hibernate/boot/model/FunctionContributor.java
sebersole, line 49:
```suggestion
	 * Determines order in which the contributions will be applied
	 * (lowest ordinal first).
	 * <p>
	 * The range 0-500 is reserved for Hibernate, range 500-1000 for libraries and
	 * 1000-Integer.MAX_VALUE for user-defined FunctionContributors.
	 * <p>
	 * Contributions from higher precedence contributors (higher numbers) effectively override 
	 * contributions from lower precedence.  E.g. if a contributor with precedence 1000 contributes a 
	 * function named {@code "max"}, that will override Hibernate's standard function of that name.
```
Something like that?

gavinking: applied

gavinking:
OK then, tests pass, and everyone approves, so I will merge it!

PR #5910, opened 2023-01-05
HHH-15950 HHH-15982 Bidirectional one-to-one associations handling changes

mbladel:
[HHH-15950](https://hibernate.atlassian.net/browse/HHH-15950) The `AssertionError` reported by the user was already resolved by https://github.com/hibernate/hibernate-orm/pull/5802 but the examples highlighted some unexpected behavior for bi-directional one-to-one associations.

[HHH-15982](https://hibernate.atlassian.net/browse/HHH-15982) This issue briefly explains the problems that we encountered and that we're trying to solve.

cc @dreab8 

[HHH-15950]: https://hibernate.atlassian.net/browse/HHH-15950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HHH-15982]: https://hibernate.atlassian.net/browse/HHH-15982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
@@ -991,7 +991,8 @@ public class ToOneAttributeMapping
 			}
 			return false;
 		}
-		return parentNavigablePath.getLocalName().equals( bidirectionalAttributeName );
+		return parentNavigablePath.getLocalName().equals( bidirectionalAttributeName ) ||
+				parentNavigablePath.getLocalName().equals( identifyingColumnsTableExpression );
 	}
 
 	public String getBidirectionalAttributeName(){
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
@@ -378,7 +378,7 @@ public abstract class AbstractEntityInitializer extends AbstractFetchParentAcces
 
 	@Override
 	public void resolveInstance(RowProcessingState rowProcessingState) {
-		if ( !missing ) {
+		if ( !missing && !isInitialized ) {
 			// Special case map proxy to avoid stack overflows
 			// We know that a map proxy will always be of "the right type" so just use that object
 			final LoadingEntityEntry existingLoadingEntry =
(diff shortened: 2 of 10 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/ToOneAttributeMapping.java
```diff
@@ -270,9 +270,12 @@ && equal( join.getKey(), manyToOne ) ) {
 				}
 				this.bidirectionalAttributeName = bidirectionalAttributeName;
 			}
-			else {
+			else if ( cardinality == Cardinality.MANY_TO_ONE ) {
 				this.bidirectionalAttributeName = referencedPropertyName;
 			}
+			else {
+				this.bidirectionalAttributeName = null;
+			}
```
mbladel, line 278: This change aims to correctly detect bi-directional circular associations when the attribute mapping is the `mappedBy` side of a `OneToOne` association with `FetchType.EAGER`. Without this, the circular fetch would not be recognized as bi-directional thus triggering an additional query identical to the first one which was not needed.

Setting `bidirectionalAttributeName` to null allows the method `isBidirectionalAttributeName` to check the parent mapping for bi-directionality obtaining a correct result.

beikov: It feels a bit fishy to set `bidirectionalAttributeName` to null even though we know the value. Is it not possible to fix `isBidirectionalAttributeName` without this change?

mbladel: I did try that at first but I was only able to make the check work by setting the property to null. It looked like `bidirectionalAttributeName` *expected* a null value when checking the `mappedBy` side of a one-to-one association, I couldn't find any other condition that wouldn't break other test cases.

mbladel: Found another way to check bidirectionality without null, using `identifyingColumnsTableExpression`.

beikov (approved):
Looks a lot cleaner to me now :)

PR #5911, opened 2023-01-05
HHH-15986 Eager Bidirectional association, initializing an HibernateProxy should associate to the circular association the HibernateProxy  itself

dreab8:
https://hibernate.atlassian.net/browse/HHH-15986

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/internal/domain/CircularBiDirectionalFetchImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/internal/domain/CircularBiDirectionalFetchImpl.java
@@ -155,8 +155,7 @@ public class CircularBiDirectionalFetchImpl implements BiDirectionalFetch {
 					final EntityPersister entityPersister = (EntityPersister) fetchable.asAttributeMapping().getMappedType();
 					final CollectionKey collectionKey = circ.resolveCollectionKey( rowProcessingState );
 					final Object key = collectionKey.getKey();
-					final SharedSessionContractImplementor session = rowProcessingState.getJdbcValuesSourceProcessingState()
-							.getSession();
+					final SharedSessionContractImplementor session = rowProcessingState.getSession();
 					final PersistenceContext persistenceContext = session.getPersistenceContext();
 					if ( fetchable.getReferencedPropertyName() != null ) {
 						return loadByUniqueKey( entityPersister, key, session, persistenceContext );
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/internal/domain/CircularBiDirectionalFetchImpl.java
```diff
@@ -182,11 +181,13 @@ public Object assemble(RowProcessingState rowProcessingState, JdbcValuesSourcePr
 			final Object initializedInstance = initializer.getInitializedInstance();
 			final LazyInitializer lazyInitializer = HibernateProxy.extractLazyInitializer( initializedInstance );
 			if ( lazyInitializer != null ) {
-				if ( initializedInstance.getClass().isAssignableFrom( javaType.getJavaTypeClass() ) ) {
-					return initializedInstance;
-				}
```
beikov, line 187: Shouldn't we still do this if the circular association is marked as LAZY?

dreab8: not sure I get your comment, by the way this issue seems happening only when we have an EAGER association

beikov: With your change, you are now always initializing and narrowing the proxy, but shouldn't we only do this when the association should be fetched eagerly?

I think you should re-add the following:

```java
				if ( timing == LAZY && initializedInstance.getClass().isAssignableFrom( javaType.getJavaTypeClass() ) ) {
					return initializedInstance;
				}
```

hibernate-core/src/main/java/org/hibernate/sql/results/internal/domain/CircularBiDirectionalFetchImpl.java
```diff
@@ -182,11 +181,14 @@ public Object assemble(RowProcessingState rowProcessingState, JdbcValuesSourcePr
 			final Object initializedInstance = initializer.getInitializedInstance();
 			final LazyInitializer lazyInitializer = HibernateProxy.extractLazyInitializer( initializedInstance );
 			if ( lazyInitializer != null ) {
-				if ( initializedInstance.getClass().isAssignableFrom( javaType.getJavaTypeClass() ) ) {
+				initializer.initializeInstance( rowProcessingState );
```
beikov, line 184: I think we should only do that if `fetchTiming != EAGER`

dreab8: fixed

hibernate-core/src/main/java/org/hibernate/sql/results/internal/domain/CircularBiDirectionalFetchImpl.java
```diff
@@ -182,11 +181,14 @@ public Object assemble(RowProcessingState rowProcessingState, JdbcValuesSourcePr
 			final Object initializedInstance = initializer.getInitializedInstance();
 			final LazyInitializer lazyInitializer = HibernateProxy.extractLazyInitializer( initializedInstance );
 			if ( lazyInitializer != null ) {
-				if ( initializedInstance.getClass().isAssignableFrom( javaType.getJavaTypeClass() ) ) {
+				final Class<?> concreteProxyClass = initializer.getConcreteDescriptor().getConcreteProxyClass();
+				if ( concreteProxyClass.isInstance( initializedInstance ) ) {
 					return initializedInstance;
```
beikov, line 186: Not sure if that can happen, but I believe we should add the following logic:

```suggestion
					if ( timing == FetchTiming.IMMEDIATE ) {
						initializer.initializeInstance( rowProcessingState );
					}
					return initializedInstance;
```

Alternatively, we should at least add an assert:

```suggestion
					assert timing != FetchTiming.IMMEDIATE;
					return initializedInstance;
```

dreab8: if the proxy has not already been initialized it will done later, we only need to call `initializer.initializeInstance( rowProcessingState );` when we need to return the proxy initialized instance and not the proxy.

PR #5912, opened 2023-01-05
HHH-15985 Use builtin PostgreSQL trunc and round two-argument functions when possible

mbladel:
https://hibernate.atlassian.net/browse/HHH-15985

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
@@ -27,6 +27,7 @@ import org.hibernate.boot.model.FunctionContributions;
 import org.hibernate.boot.model.TypeContributions;
 import org.hibernate.dialect.function.CommonFunctionFactory;
 import org.hibernate.dialect.function.FormatFunction;
+import org.hibernate.dialect.function.PostgreSQLTruncRoundFunction;
 import org.hibernate.dialect.identity.CockroachDBIdentityColumnSupport;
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
```diff
@@ -413,6 +414,11 @@ public void initializeFunctionRegistry(FunctionContributions functionContributio
 		functionFactory.listagg_stringAgg( "string" );
 		functionFactory.inverseDistributionOrderedSetAggregates();
 		functionFactory.hypotheticalOrderedSetAggregates_windowEmulation();
+
+		functionContributions.getFunctionRegistry().register(
+				"trunc", new PostgreSQLTruncRoundFunction( "trunc", true )
```
beikov, line 419:
```suggestion
				"trunc", 
				// CockroachDB didn't support the two-argument trunc before version 22.2
				new PostgreSQLTruncRoundFunction( "trunc", getVersion().isSameOrAfter( 22, 2 ) )
```

beikov (commented):
Looks good. Please also add the function registrations to the legacy dialects.

hibernate-core/src/main/java/org/hibernate/dialect/CockroachDialect.java
mbladel: Oops, must have accidentally removed the version check while merging. Thanks!

beikov (approved):

PR #5919, opened 2023-01-08
promote joinTransaction() and isJoinedToTransaction() to SharedSessionContract

gavinking:
This makes these operations available on `StatelessSession` instead of just on `Session`.

See https://twitter.com/omidboot/status/1612019552904253441

@sebersole what do you think of this proposal?


sebersole (commented):
lgtm

PR #5921, opened 2023-01-09
HHH-15950 HHH-15982 Bidirectional one-to-one associations handling changes

mbladel:
Backport of https://github.com/hibernate/hibernate-orm/pull/5910 and https://github.com/hibernate/hibernate-orm/pull/5922

PR #5922, opened 2023-01-09
HHH-15982 Refactor tests to use StatementInspector and correct bidirectionalAttributeName handling

mbladel:
Refactor tests as suggested by https://github.com/hibernate/hibernate-orm/pull/5921#discussion_r1064505664.

Also, only set `bidirectionalAttributeName` when needed and rollback the check for bidirectional associations that was wrongly using the table expression.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/mapping/converted/converter/onetoone/BidirectionalOneToOneWithConverterEagerTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/mapping/converted/converter/onetoone/BidirectionalOneToOneWithConverterEagerTest.java
@@ -142,7 +142,7 @@ public class BidirectionalOneToOneWithConverterEagerTest {
 	}
 
 	@Entity(name = "FooEntity")
-	@Table(name = "foo")
+	@Table(name = "foo_table")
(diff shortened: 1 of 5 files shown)


beikov (approved):

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/converted/converter/onetoone/BidirectionalOneToOneWithConverterEagerTest.java
```diff
@@ -99,24 +91,18 @@ public void jdbcExecuteStatementStart() {
 
 	@Test
 	public void testBidirectionalFetchInverse(SessionFactoryScope scope) {
+		final SQLStatementInspector sqlStatementInterceptor = (SQLStatementInspector) scope.getStatementInspector();
```
jrenaat, line 94: FYI, you can call the `getCollectingStatementInspector()` method, that way you can get rid of the cast ...

mbladel: thanks, gonna use that!

PR #5928, opened 2023-01-10
HHH-15957 Hibernate 6.1.6 Infinite Recursion/StackOverflow Using DB2 Custom Dialect

mbladel:
https://hibernate.atlassian.net/browse/HHH-15957


beikov (approved):

matthenry87:
The 6.1.7-SNAPSHOT allowed me to extend DB2zDialect and override the getVersion() as a temporary fix for our DB2 z/OS Application Compatibility setting. Thanks!!

According to your releases page, it seems your cadence is just over once a month. Do we have a date for 6.1.7.Final?

beikov:
We will release it in the next few days.

PR #5929, opened 2023-01-10
HHH-16010 fix two bugs in natural id handling 

gavinking:
These bugs are visible just by inspection of the code.

- fix place where id and entity had always been passed in reversed order
- fix place where whole state array was passes instead of natural id array
- change the API of NaturalIdResolutions because it never needs the session, and the fix involved calling it from a place we did not have one
- and also clean up a bunch of warnings

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/AbstractEntityEntry.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/AbstractEntityEntry.java
@@ -14,31 +14,43 @@ import org.hibernate.AssertionFailure;
 import org.hibernate.CustomEntityDirtinessStrategy;
 import org.hibernate.HibernateException;
 import org.hibernate.LockMode;
-import org.hibernate.Session;
 import org.hibernate.bytecode.enhance.spi.interceptor.EnhancementAsProxyLazinessInterceptor;
 import org.hibernate.collection.spi.PersistentCollection;
-import org.hibernate.engine.spi.CachedNaturalIdValueSource;
 import org.hibernate.engine.spi.EntityEntry;
 import org.hibernate.engine.spi.EntityEntryExtraState;
 import org.hibernate.engine.spi.EntityKey;
 import org.hibernate.engine.spi.PersistenceContext;
-import org.hibernate.engine.spi.PersistentAttributeInterceptable;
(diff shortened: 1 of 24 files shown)


hibernate-core/src/main/java/org/hibernate/engine/internal/AbstractEntityEntry.java
```diff
@@ -426,7 +422,9 @@ public void setReadOnly(boolean readOnly, Object entity) {
 			setStatus( Status.MANAGED );
 			loadedState = getPersister().getValues( entity );
 			getPersistenceContext().getNaturalIdResolutions().manageLocalResolution(
-					id, loadedState, persister,
+					id,
+					getPersister().getNaturalIdMapping().extractNaturalIdFromEntityState( loadedState ),
```
gavinking, line 426: This was the bug I noticed just by inspection.

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultEvictEventListener.java
```diff
@@ -111,7 +111,7 @@ protected void doEvict(
 
 		final PersistenceContext persistenceContext = session.getPersistenceContextInternal();
 		if ( persister.hasNaturalIdentifier() ) {
-			persistenceContext.getNaturalIdResolutions().handleEviction( object, key.getIdentifier(), persister );
+			persistenceContext.getNaturalIdResolutions().handleEviction( key.getIdentifier(), object, persister );
```
gavinking, line 114: This was the bug reported in issue https://hibernate.atlassian.net/browse/HHH-16010

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/NaturalIdMapping.java
```diff
@@ -39,6 +40,7 @@
  *
  * @author Steve Ebersole
  */
+@Incubating
 public interface NaturalIdMapping extends VirtualModelPart {
```
gavinking, line 44: I have removed the session from the methods of this SPI, since it is not needed, and because I needed to call `extractNaturalIdFromEntityState()` from a place where I did not have one.

gavinking:
Merged.

@sebersole please let me know if you have any reason to object to the changes to `NaturalIdResolutions`.

PR #5931, opened 2023-01-11
HHH-16014 Querying property from treated path causes IllegalArgumentException: Already registered a copy

mbladel:
https://hibernate.atlassian.net/browse/HHH-16004 was already solved by https://github.com/hibernate/hibernate-orm/pull/5866 but the test highlighted new issues, see: https://hibernate.atlassian.net/browse/HHH-16014.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/SqmCopyContext.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/SqmCopyContext.java
@@ -8,6 +8,8 @@ package org.hibernate.query.sqm.tree;
 
 import java.util.IdentityHashMap;
 
+import org.hibernate.query.sqm.tree.domain.SqmPath;
+
 /**
  *
  */
@@ -21,9 +23,14 @@ public interface SqmCopyContext {
 		final IdentityHashMap<Object, Object> map = new IdentityHashMap<>();
 		return new SqmCopyContext() {
 			@Override
+			@SuppressWarnings("unchecked")
 			public <T> T getCopy(T original) {
-				//noinspection unchecked
-				return (T) map.get( original );
+				if (original instanceof SqmPath) {
+					return (T) getPathCopy( (SqmPath<?>) original );
+				}
+				else {
+					return (T) map.get( original );
+				}
 			}
 
 			@Override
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/SqmAnyValuedSimplePath.java
```diff
@@ -44,6 +44,7 @@ public SqmAnyValuedSimplePath(
 
 	@Override
 	public SqmAnyValuedSimplePath<T> copy(SqmCopyContext context) {
+		final SqmPath<?> lhsCopy = getLhs().copy( context );
```
beikov, line 47: If `this` path was already copied before, copying the `lhs` here should have no effect, hence why it was only done when the `this` path was not yet copied. Can you please explain a bit what is happening here and why this makes a difference?

mbladel: OFC: when `getLhs()` was a treated path (e.g. `SqmTreatedSimplePath`), the `copy()` method of that triggered another copy of `this` while the first one was still happening - this is because `this` would be in the `reusablePaths` of the treated path object. The second copy of `this` triggered by `getLhs().copy()` (the treated path) would then be registered but when that finished we would still be inside the original `copy` method of `this`. This is what caused the exception *Already registered a copy*.

Moving the `getLhs().copy()` method up, before the check for existing copies, solves this issue: when the treated path copies `this` now the copy is registered early and we don't try to create another because we will find the existing one.

beikov: Sounds like we should rather defer the copying of the `reusablePaths` or maybe even skip that altogether, since the reusable paths will be visited eventually anyway throughout the copying.

mbladel: I did try skipping that and it looked like it would solve the issue, but wouldn't the `getLhs()` copy then be missing the `reusablePaths`? Also, deferring would be hard because we would need to defer the `getLhs().copyTo()` after `this` copy has finished but we have no way of knowing that from the treated path.

PR #5938, opened 2023-01-12
allow cast(thing as Instant) and cast(thing as Duration) in HQL

gavinking:


gavinking:
Closing in favor of #5938.

PR #5939, opened 2023-01-12
HHH-15928 Distinct with maxResults fails under SQLServer

mbladel:
https://hibernate.atlassian.net/browse/HHH-15928

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/SQLServerSqlAstTranslator.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/SQLServerSqlAstTranslator.java
@@ -230,6 +230,10 @@ public class SQLServerSqlAstTranslator<T extends JdbcOperation> extends Abstract
 			else if ( version.isBefore( 11 ) || !isRowsOnlyFetchClauseType( queryPart ) ) {
 				return OffsetFetchClauseMode.EMULATED;
 			}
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/SQLServerSqlAstTranslator.java
```diff
@@ -293,7 +297,12 @@ protected void renderFetchPlusOffsetExpression(
 			Expression fetchClauseExpression,
 			Expression offsetClauseExpression,
 			int offset) {
-		renderFetchPlusOffsetExpressionAsSingleParameter( fetchClauseExpression, offsetClauseExpression, offset );
+		if ( getCurrentQueryPart().isRoot() ) {
+			renderFetchPlusOffsetExpressionAsSingleParameter( fetchClauseExpression, offsetClauseExpression, offset );
+		}
+		else {
+			super.renderFetchPlusOffsetExpression( fetchClauseExpression, offsetClauseExpression, offset );
+		}
```
beikov, line 305: I'm not quite sure what the purpose of this change is. Can you explain it? I believe SQL Server is picky about requiring a simple expression literal/parameter to appear in the top or fetch clause, so this code creates a single parameter instead of an addition expression `offset + limit`, but why would we only need that for the root?

beikov (approved):
Other than what I just commented, this PR looks fine to me

beikov (approved):

PR #5941, opened 2023-01-12
HHH-16031 Reproducer for "@ManyToMany with @JoinTable(inverseColumn = ...) and SortedSet may results in data loss"

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16031

Reproducer only, no fix.


yrodiere:
You can find the same test backported to 5.6 here: https://github.com/yrodiere/hibernate-orm/tree/HHH-16031-5.6
It does not fail on 5.6.

yrodiere:
Superseded by #5965

PR #5943, opened 2023-01-12
HHH-16035 add tests demoing bugs in HQL Duration support

gavinking:
Some of these issues probably arose with the change to use SQL `interval` types to store `Duration`, but some of them might have pre-existed that, since we just did not have very good tests here.


gavinking:
@beikov I'm trying to kill `PostgreSQLIntervalSecondJdbcType`. Please speak up if you really want to save its life.

gavinking:
> Yeah, please don't remove it. I want to allow people to use that DDL type.

But it's untested, unused, and is causing bugs.

How about I remove it, and when you've fixed the bugs, you can put it back in?

gavinking:
Oh also: I'm not sure, but I think you need to reimplement it without the use of reflection, because I bet that reflecting on JDBC driver classes is going to cause problems in Quarkus. Is that right @Sanne?

gavinking:
> Oh also: I'm not sure, but I think you need to reimplement it without the use of reflection, because I bet that reflecting on JDBC driver classes is going to cause problems in Quarkus. Is that right @Sanne?

Well, perhaps it simply gets disabled in Quarkus. (Its use is protected by a call to `PostgreSQLPGObjectJdbcType.isUsable()`.) But that's all the more reason to dislike this thing.

gavinking:
@beikov _alternatively_, I could change the default value of `hibernate.type.preferred_duration_jdbc_type` to `NUMERIC`.

This would fix the bug I'm most worried about on both h2 and Postgres. The cost would be that `Duration` would not be stored as `interval second` on h2. But that seems completely fine to me. People usually use h2 for testing, and so it's not terribly useful that we store `Duration`s as `numeric` on every other database, but then as `interval second` on h2. Make it _less_ useful as a platform for tests.

(I'm fine with either solution.)

Sanne:
Right we should generally avoid reflection, however if there's compelling need for it it can be handled, just needs some extra steps. The most important being to keep track of it, as otherwise the missing metadata causes really puzzling hidden problems: at least being aware that it might be related to a new use of reflection helps a lot.

The Hibernate ORM module `hibernate-graalvm` can be used to define such metadata.

Bear in mind also: any use of reflection which is happening in a static initializer, or generally happens before the `SessionFactory` is started (e.g. during the constructor of the Dialect) is not a problem and should work w/o any further concern.

Sanne:
> But it's untested, unused, and is causing bugs.
> How about I remove it, and when you've fixed the bugs, and written tests for it, you can put it back in?

I like that approach - if a new features has no tests or documentation it's not ready.

gavinking:
>  if there's compelling need for it it can be handled, just needs some extra steps

Yeah, I don't see this one as compelling at all.

> any use of reflection which is happening in a static initializer, or generally happens before the `SessionFactory` is started (e.g. during the constructor of the `Dialect`) is not a problem and should work w/o any further concern.

Here the reflection lookups happen at startup, but the *use* of reflection happens at runtime, so I think it counts as a concern.

Actually, now that you mention this, it looks to me like what would happen is that:

- the reflection lookups would succeed at build-time, and so the Postgres `Dialect` would decide to use all this reflective stuff, but then
- it would fail to work at runtime, where reflection is unavailable.

But I have not actually tested this to see if it's true.

gavinking:
> @beikov _alternatively_, I could change the default value of `hibernate.type.preferred_duration_jdbc_type` to `NUMERIC`.

Here's that alternative: https://github.com/hibernate/hibernate-orm/pull/5951

beikov:
>it would fail to work at runtime, where reflection is unavailable.

Which reflection at runtime are you referring to? `PostgreSQLIntervalSecondJdbcType` sets up all the reflective parts in a static block, so this should be fine also for Quarkus.

Also see

>Bear in mind also: any use of reflection which is happening in a static initializer, or generally happens before the SessionFactory is started (e.g. during the constructor of the Dialect) is not a problem and should work w/o any further concern.

There is some documentation that talks about this:

* https://docs.jboss.org/hibernate/orm/6.2/userguide/html_single/Hibernate_User_Guide.html#_duration
* https://docs.jboss.org/hibernate/orm/6.2/userguide/html_single/Hibernate_User_Guide.html#hql-duration-literals
* https://docs.jboss.org/hibernate/orm/6.2/userguide/html_single/Hibernate_User_Guide.html#basic-legacy-provided

and also the migration guide: https://github.com/hibernate/hibernate-orm/blob/6.0/migration-guide.adoc#instant-mapping-changes

If two of the new features don't work together well in some scenarios, then how about we fix that instead of removing one feature? I'm on PTO now for a week, but this sounds like a bug that we can fix. No reason to remove features that were introduced in 6.0 now, literally hours before doing the 6.2 final release.

gavinking:
> sets up all the reflective parts in a static block, so this should be fine also for Quarkus.

It sets them up in a static, but it calls them at runtime. What are the rules here: am I allowed to call `Method` objects at runtime if I've retrieved them at build time. I'm not sure of the answer to that. I thought I was _not_ allowed to do that but I may simply be wrong and also I may just be *way* out of date. @Sanne?

> There is some documentation that talks about this:

Not sure what your point is. The only thing those docs say are:

> By default, Hibernate will map `Duration` to the `INTERVAL_SECOND` SQL type and fallback to `NUMERIC` if necessary.

In fact, the _only_ database where Hibernate maps `Duration` to the SQL type `INTERVAL SECOND` is h2, so this is almost verging on misleading.

> If two of the new features don't work together well in some scenarios, then how about we fix that instead of removing one feature?

Fine, if you prefer, I can go with the proposed "alternative" solution instead of removing this class. Like I said: I'm fine either way.

gavinking:
> It sets them up in a static, but it calls them at runtime. What are the rules here: am I allowed to call `Method` objects at runtime if I've retrieved them at build time. I'm not sure of the answer to that. I thought I was not allowed to do that but I may simply be wrong and also I may just be way out of date. @Sanne?

FTR I asked Sanne via backchannel and he's not sure either. He thinks it's more likely to work via `VarHandle` but we would have to test it.

gavinking:
Closing in favor of #5951.

PR #5945, opened 2023-01-12
HHH-160333 Many-to-Many inverse mapping referencing the same class uses pk instead of fk field for removal

wburns:
…es pk instead of fk field for removal

* Just reproducer for issue

https://hibernate.atlassian.net/browse/HHH-16033


dreab8:
included into https://github.com/hibernate/hibernate-orm/pull/5965

PR #5949, opened 2023-01-12
HHH-16020 - Fix for incorrect offset parameter index

jrenaat:

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/pagination/AbstractLimitHandler.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/pagination/AbstractLimitHandler.java
@@ -194,7 +194,7 @@ public abstract class AbstractLimitHandler implements LimitHandler {
 
 		if ( bindOffset ) {
 			statement.setInt(
-					index + ( reverse || !bindLimit ? 1 : 0 ),
+					index + ( reverse && bindLimit ? 1 : 0 ),
 					getFirstRow( limit )
 			);
 		}
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/query/NativeQueryLimitOffsetTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/query/NativeQueryLimitOffsetTest.java
@@ -1,0 +1,132 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.orm.test.query;
+
+import java.util.List;
+
+import org.hibernate.dialect.AbstractHANADialect;
+import org.hibernate.dialect.MySQLDialect;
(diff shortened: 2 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/pagination/AbstractLimitHandler.java
```diff
@@ -194,7 +194,7 @@ protected final int bindLimitParameters(Limit limit, PreparedStatement statement
 
 		if ( bindOffset ) {
 			statement.setInt(
-					index + ( reverse || !bindLimit ? 1 : 0 ),
+					index + ( reverse && bindLimit ? 1 : 0 ),
```
beikov, line 197: Comparing that to the original code, it seems like the `bindLimit` part should maybe not be considered at all: https://github.com/hibernate/hibernate-orm/blob/5.6/hibernate-core/src/main/java/org/hibernate/dialect/pagination/AbstractLimitHandler.java#L154

jrenaat: If it's reverse, i.e. limit has to come before the offset, AND there is a limit to be parametered, then IIUC the index should be +1, if on the other hand it's reverse but no limit was specified, it should be left as is, so unless i have this completely wrong, the bindLimit part has to be there ?

dreab8: Hi @jrenaat 
I have just changed line 197 to `index + ( reverse ? 1 : 0 ),` and then run your test, it passed.
So probably @beikov is right and we don't need the `bindList` part

jrenaat: That's no guarantee at all, since the only dialects that actually have reverse to true are the HANA and Spanner dialects, which are not being tested in CI.
If reverse is true, but no bindLimit is defined, then the index should NOT be bumped +1, which will not be the case with `index + ( reverse ? 1 : 0 )`where it would indeed be bumped in the absence of the bindlimit.
Correct me if I'm missing something

dreab8: testing on HANA and it seems this part is broken despite the `bindLimit` part

dreab8: I think HANA does not support offset without a limit, so when `bindOffset` is true then `bindLimit` is true too.

dreab8: so I think you have to skip `testPartialLimitOffsetOnNativeQuery` also for HANA

dreab8: it seems the same for Spanner

dreab8: by the way I think logically your change is correct, in case of `reverse == true` if there is not `bindLimit` we have to add `0` and not `1` to the offset parameter index

PR #5951, opened 2023-01-13
HHH-16035 change the default for hibernate.type.preferred_duration_jdbc_type to NUMERIC

gavinking:
There's really no value at all in having h2 as the only platform where `Duration` is persisted as `interval second` by default. People usually use h2 for testing, and probably actually prefer if the schema is more similar to the schema of their "real" database.

Also, this tiny change fixes issues 1. and 2. of HHH-16035 as a side-effect.


gavinking:
@beikov doesn't want to remove `PostgreSQLIntervalSecondJdbcType` (though I think it's value is highly debatable) and so I'm inclined to go with this as a fix to the two problems found in HHH-16035 and the potential issues with the use of reflection on Quarkus.

gavinking:
Comments?

gavinking:
OK, so apart from fixing most of the bugs, this patch has exactly one side-effect: it changes the default DDL type for `Duration` only on h2. And the change makes h2 consistent with the DDL type _on every other database_.

So I'm going to merge this.

OrangeDog:
>only on h2

No, this affects PostgreSQL too.

The migration guides all still say that `INTERVAL_SECOND` is the default. That should be updated to avoid users (like me) having to spend time investigating why it's not working.

OrangeDog:
https://docs.jboss.org/hibernate/orm/6.0/migration-guide/migration-guide.html

> Duration now maps to the type code `SqlType.INTERVAL_SECOND` by default.

None of the later migration guides mention that the default was changed.

PR #5954, opened 2023-01-13
deprecate config property org.hibernate.flushMode on AvailableSettings

gavinking:
It's actually a JPA hint, and rightfully belongs on `HibernateHints`.

cc @Sanne

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
@@ -3068,8 +3068,13 @@ public interface AvailableSettings {
 	/**
 	 * Used to determine flush mode.
 	 *
+	 * @deprecated There are much better ways to control the flush mode of a session,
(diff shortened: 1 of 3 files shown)


Sanne:
Personally I completely agree... it's a dangerous property, even though I can see the appeal in certain scenarious, e.g. fully read-only systems - although it would be perfectly viable to set per-session on whatever general utility people will use as Session factory so I don't think I'd miss the property, even in the far fetched scenarios.

However there seem to be people asking for it even in Quarkus, just happened recently again:
 - https://github.com/quarkusio/quarkus/issues/29604#issuecomment-1380714695

We might want to proceed with this anyway, but perhaps let's give them a chance to explain themselves first?
Or at least wait for some other team member's opinions - at least @sebersole ?

cc/ @yrodiere

gavinking:
I mean, all that is fine, I don't care if we have a config property for this. But that config property has to be called something like:

    hibernate.default_flush_mode

or even 

    hibernate.session.default_flush_mode

or whatever.

It can't be called `org.hibernate.flushMode` because that's not the name of a config property :-)

PR #5958, opened 2023-01-15
Extract unique constraint names from error messages by H2 Database

tshibata:
H2Dialect should extract constraint names from the error messages, such as:
Unique index or primary key violation: "PUBLIC.ACCOUNTS_BY_USERNAME ON ...
(In this case, it extracts "ACCOUNTS_BY_USERNAME")


beikov:
I don't mind cutting off the ` ON ...` part since that really isn't part of the constraint name, but the schema name definitely is.
If you want this feature, please [create a Jira issue](https://hibernate.atlassian.net/) first and follow our [contribution guidelines](https://github.com/hibernate/hibernate-orm/blob/main/CONTRIBUTING.md).

gavinking:
Related: https://github.com/h2database/h2database/issues/4021

PR #5960, opened 2023-01-16
Reproducer for HHH-16049 Setting a property to its current value with bytecode enhancement enabled results in unnecessary SQL Update in some (many) cases

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16049

Draft: this is just a reproducer, not a fix.


yrodiere:
Superseded by  #5967

PR #5961, opened 2023-01-16
Reproducer for HHH-16049 on branch 5.6 - Setting a property to its current value with bytecode enhancement enabled results in unnecessary SQL Update in some (many) cases

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16049

Draft: this is just a reproducer, not a fix.

Backport of #5960


yrodiere:
Actually not yet, we'd need a backport of the fix :)

yrodiere:
Superseded by #5968

PR #5962, opened 2023-01-16
HHH-16040 AssertionError for @JoinColumn with char[] property referencing a String property

mbladel:
https://hibernate.atlassian.net/browse/HHH-16040
https://hibernate.atlassian.net/browse/HHH-15916

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/TreatPathTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/TreatPathTest.java
@@ -42,6 +42,7 @@ import jakarta.persistence.criteria.Predicate;
 import jakarta.persistence.criteria.Root;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
 
 /**
  * @author Marco Belladelli
@@ -65,7 +66,7 @@ public class TreatPathTest {
 			term.setLength( 4 );
 			term.setLanguage( language );
 			term.setAnyProperty( stringProperty );
-			term.setSynonyms( new ArrayList<>( List.of( "ciao" ) ) );
+			term.setSynonyms( new ArrayList<>() );
 			term.setEmbeddableProperty( new EmbeddableType( "ciao" ) );
 			Linkage linkage = new Linkage();
 			linkage.setTerm( term );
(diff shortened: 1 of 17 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/TreatPathTest.java
```diff
@@ -129,13 +131,17 @@ public void testTreatAnyValue(EntityManagerFactoryScope scope) {
...
+	}
+
+	private void testCriteriaTreat(EntityManager entityManager, String property, Object value, boolean useIn) {
+		final CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+		final CriteriaQuery<Linkage> criteria = cb.createQuery( Linkage.class );
+		final Root<Linkage> root = criteria.from( Linkage.class );
+		final Path<LocalTerm> asLocalTerm = cb.treat( root.get( "term" ), LocalTerm.class );
+		final Predicate predicate;
+		if ( useIn ) {
+			predicate = cb.literal( value ).in( asLocalTerm.get( property ) );
```
beikov, line 144: This refactoring kind of seems wrong. You switched from `asLocalTerm.get( property ).in( value )` to `cb.literal( value ).in( asLocalTerm.get( property ) )`, which is reverse AFAIU.

beikov (commented):
Other than what I commented, looks good to me!

hibernate-core/src/test/java/org/hibernate/orm/test/jpa/criteria/TreatPathTest.java
mbladel: That was on purpose, I think the way I wrote the test before was wrong. This case with plural values is a bit flaky anyway - I thought about disabling it but for the purpose of checking treat itself it was enough to expect an `UsupportedOperationException`.

mbladel: Fixed the test now.

beikov (commented):
It seems `CharArrayToStringInEmbeddedJoinColumnOrFormulaTest` fails on Sybase for some reason.

mbladel:
> It seems `CharArrayToStringInEmbeddedJoinColumnOrFormulaTest` fails on Sybase for some reason.

Yep, it was caused by the `trim` function in the formula. I just removed it since it's not relevant for the purpose of the test.

beikov (approved):
LGTM

PR #5948, opened 2023-01-12
HHH 16031 @ManyToMany with @JoinTable(inverseColumn = ...) and SortedSet may results in data loss

dreab8:
https://hibernate.atlassian.net/browse/HHH-16031
https://hibernate.atlassian.net/browse/HHH-16033

PR #5965, opened 2023-01-17
HHH-16031 @ManyToMany with @JoinTable(inverseColumn = ...) and SortedSet may results in data loss

dreab8:
https://hibernate.atlassian.net/browse/HHH-16031
https://hibernate.atlassian.net/browse/HHH-16033

alternative to https://github.com/hibernate/hibernate-orm/pull/5948

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultDeleteEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultDeleteEventListener.java
@@ -263,6 +263,7 @@ public class DefaultDeleteEventListener implements DeleteEventListener,	Callback
 			&& !persister.hasSubclasses() //TODO: should be unnecessary, using EntityPersister.getSubclassPropertyTypeClosure(), etc
 			&& !persister.hasCascadeDelete()
 			&& !persister.hasNaturalIdentifier()
+			&& !persister.hasCollectionNotReferencingPK()
 			&& !hasRegisteredRemoveCallbacks( persister )
 			&& !hasCustomEventListeners( source );
 	}
--- a/hibernate-core/src/main/java/org/hibernate/persister/collection/BasicCollectionPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/collection/BasicCollectionPersister.java
@@ -454,17 +454,17 @@ public class BasicCollectionPersister extends AbstractCollectionPersister {
 		// WHERE
 
 		if ( attribute.getIdentifierDescriptor() != null ) {
-			attribute.getIdentifierDescriptor().forEachSelectable( updateBuilder::addKeyRestriction );
+			attribute.getIdentifierDescriptor().forEachSelectable( updateBuilder::addKeyRestrictionLeniently );
 		}
 		else {
-			attribute.getKeyDescriptor().getKeyPart().forEachSelectable( updateBuilder::addKeyRestriction );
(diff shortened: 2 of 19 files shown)


hibernate-core/src/main/java/org/hibernate/persister/collection/BasicCollectionPersister.java
```diff
@@ -597,19 +597,19 @@ private RestrictedTableMutation<JdbcMutationOperation> generateDeleteRowAst(Muta
...
 		if ( pluralAttribute.getIdentifierDescriptor() != null ) {
-			deleteBuilder.addKeyRestriction( pluralAttribute.getIdentifierDescriptor() );
+			deleteBuilder.addKeyRestrictionLeniently( pluralAttribute.getIdentifierDescriptor() );
 		}
 		else {
 			pluralAttribute
 					.getKeyDescriptor()
 					.getKeyPart()
-					.forEachSelectable( (index, selectable) -> deleteBuilder.addKeyRestriction( selectable ) );
+					.forEachSelectable( (index, selectable) -> deleteBuilder.addKeyRestrictionLeniently( selectable ) );
```
sebersole, line 606: A general observation as this happens a bunch - 

We should add a form of `#addKeyRestrictionLeniently` which accepts the index as well.  Even though it will ignore the index, that will allow the use of a method reference as opposed to lambdas

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultDeleteEventListener.java
```diff
@@ -263,6 +263,7 @@ private boolean canBeDeletedWithoutLoading(EventSource source, EntityPersister p
 			&& !persister.hasSubclasses() //TODO: should be unnecessary, using EntityPersister.getSubclassPropertyTypeClosure(), etc
 			&& !persister.hasCascadeDelete()
 			&& !persister.hasNaturalIdentifier()
+			&& !persister.hasNaturalIdCollections()
```
sebersole, line 266: Not sure about the name here.  I looked through how this comes to be and I think its not really related to what the feature of mapping a natural-id.  If I understand correctly, this is merely indicating whether the entity contains any collection where the collection-key does not refer to the entity's identifier.  

Assuming that understanding is correct, I think the term is a bit confusing.

sebersole (commented):
I left some comments, but there are ancillary.  Overall it looks good

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultDeleteEventListener.java
dreab8: yes  you are right, `hasNonPKCollection` may be a little better but still horrible and probably also confusing
or `hasCollectionNotReferencingPK`. Any suggestion is more than welcome :P

hibernate-core/src/main/java/org/hibernate/persister/collection/BasicCollectionPersister.java
dreab8: it was already there but not used in this case :( fixed this
I added instead one for `addKeyRestriction`

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultDeleteEventListener.java
sebersole: `hasCollectionNotReferencingPK` sounds reasonable

dreab8: method name changed, thanks

PR #5971, opened 2023-01-17
HHH-15916 - Regressions when using JoinColumnOrFormula on char[] and String

chrisaige:
The problem is that its not possible to join a column of type String with a column of type char[] even tho they both are of the same JDBCType my proposal would be to support additional setters that allow to set the value correctly as it would be to cumbersome to try to use the JDBCType coverter


beikov:
Pretty sure this is covered by https://github.com/hibernate/hibernate-orm/pull/5962 now, so I'll close this PR.

PR #5974, opened 2023-01-18
HHH-16050 Optimize iteration efficiency in StandardStack 

mbladel:
https://hibernate.atlassian.net/browse/HHH-16050

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/util/collections/StandardStack.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/util/collections/StandardStack.java
@@ -6,9 +6,9 @@
  */
 package org.hibernate.internal.util.collections;
 
-import java.util.ArrayDeque;
-import java.util.Deque;
-import java.util.Iterator;
+import java.lang.reflect.Array;
+import java.util.Arrays;
+import java.util.NoSuchElementException;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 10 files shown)


hibernate-core/src/main/java/org/hibernate/internal/util/collections/StandardStack.java
```diff
@@ -6,139 +6,159 @@
...
+	private int head = 0;
+	private int tail = 0;
 
-	private ArrayDeque<T> internalStack;
-	private static final Object NULL_TOKEN = new Object();
-
-	public StandardStack() {
+	@SuppressWarnings("unchecked")
+	public StandardStack(Class<T> type) {
+		elements = (T[]) Array.newInstance( type, 8 );
```
Sanne, line 36: There's a non-negligible change that the `StandardStack` might be constructed, but then nothing gets added.

So I think it would be best to allocate the Array lazily?

hibernate-core/src/main/java/org/hibernate/internal/util/collections/StandardStack.java
```diff
@@ -6,139 +6,159 @@
...
-
-	private Deque<T> stackInstanceExpected() {
-		if ( internalStack == null ) {
-			//"7" picked to use 8, but skipping the odd initialCapacity method
-			internalStack = new ArrayDeque<>( 7 );
+	public void push(T e) {
+		head = dec( head );
+		elements[head] = e;
+		if ( head == tail ) {
+			grow();
```
Sanne, line 49: It's a bit surprising that we might end up growing the array right *after* having stored the last element.

Would be fantastic to avoid growing after a push, but rather make space if there is no space before a push?

I guess this also relates with the reason about why you needed to pre-allocate the array even on construction. Neither should be necessary.

hibernate-core/src/main/java/org/hibernate/internal/util/collections/StandardStack.java
```diff
@@ -6,139 +6,159 @@
...
 	@Override
 	public void clear() {
-		if ( internalStack != null ) {
-			internalStack.clear();
+		circularClear( elements, head, tail );
+		head = tail = 0;
+	}
+
+	private static void circularClear(Object[] es, int i, int end) {
+		for ( int to = ( i <= end ) ? end : es.length; ; i = 0, to = end ) {
```
Sanne, line 95: I realize now that I gave you a bad suggestion, sorry.
In our particular case we don't really need a circular array; I had forgotten that `ArrayDeque` is based on a circular array, I guess because it allows adding elements either at the end or at the beginning of the queue (possibly needing to shift all entries).

but our use case is much simpler: we only have pop/push modification operations; this implies you should be able to simplify this all a little bit, as you don't need to keep track of both the head and the tail, only needing to ensure the array grows if there is not enough space: we always only insert new elements at the end.

The good news: you can indeed make this much more efficient than the existing implementation :)

hibernate-core/src/test/java/org/hibernate/orm/test/util/StandardStackTest.java
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	@Test
+	public void testClear() {
+		final Stack<Integer> stack = allocateStack( 42 );
+		assertEquals( 42, stack.depth() );
+		assertFalse( stack.isEmpty() );
+		stack.clear();
+		assertEquals( 0, stack.depth() );
+		assertTrue( stack.isEmpty() );
```
Sanne, line 43: let's also test for that the various "find" methods work correctly on both an empty and a cleared stack?

Sanne (changes requested):
Excellent work!

But I'm suggesting a small change in the algorythm as I just realized we can do even better than the original ArrayDeque in our context :)

No rush - and great job!

hibernate-core/src/main/java/org/hibernate/internal/util/collections/StandardStack.java
Sanne: N.B. when I wrote this comment I hadn't realized the implementation was based on a circular array. This one might get clearer if you read the other comments first.

hibernate-core/src/main/java/org/hibernate/internal/util/collections/StandardStack.java
mbladel: You're right, I was too focused on reproducing the previous behavior that I didn't stop and think if there was a better way of achieving this. The implementation will also be a lot easier tracking just the top of the stack. Thanks for all the suggestions!

Sanne (approved):
Thanks, looks great!

PR #5977, opened 2023-01-18
HHH-16058 Removing Environment#getBytecodeProvider

Sanne:
https://hibernate.atlassian.net/browse/HHH-16058

cc/ @sebersole and @DavideD for awareness

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/Environment.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/Environment.java
@@ -13,6 +13,7 @@ import java.util.Properties;
 import org.hibernate.HibernateException;
 import org.hibernate.Internal;
 import org.hibernate.Version;
+import org.hibernate.bytecode.internal.BytecodeProviderInitiator;
 import org.hibernate.bytecode.spi.BytecodeProvider;
 import org.hibernate.internal.CoreMessageLogger;
 import org.hibernate.internal.util.ConfigHelper;
@@ -35,8 +36,8 @@ import static org.hibernate.internal.log.DeprecationLogger.DEPRECATION_LOGGER;
  * always determined by the {@code Environment} properties in {@link #getProperties()}.
  * </ul>
  * <p>
- * The only system-level properties are {@value #USE_REFLECTION_OPTIMIZER} and
- * {@value #BYTECODE_PROVIDER}.
+ * The only system-level property is {@value #USE_REFLECTION_OPTIMIZER},
+ * and it's deprecated.
  * <p>
  * {@code Environment} properties are populated by calling {@link System#getProperties()}
  * and then from a resource named {@code /hibernate.properties}, if it exists. System
@@ -141,7 +142,6 @@ import static org.hibernate.internal.log.DeprecationLogger.DEPRECATION_LOGGER;
 public final class Environment implements AvailableSettings {
 	private static final CoreMessageLogger LOG = Logger.getMessageLogger( CoreMessageLogger.class, Environment.class.getName());
 
-	private static final BytecodeProvider BYTECODE_PROVIDER_INSTANCE;
 	private static final boolean ENABLE_REFLECTION_OPTIMIZER;
 
 	private static final Properties GLOBAL_PROPERTIES;
@@ -193,8 +193,6 @@ public final class Environment implements AvailableSettings {
 		else {
 			DEPRECATION_LOGGER.deprecatedSettingForRemoval( USE_REFLECTION_OPTIMIZER, "true" );
 		}
-
-		BYTECODE_PROVIDER_INSTANCE = buildBytecodeProvider( GLOBAL_PROPERTIES );
 	}
 
 	/**
@@ -203,12 +201,11 @@ public final class Environment implements AvailableSettings {
 	 * @return True if reflection optimization should be used; false otherwise.
 	 *
 	 * @see #USE_REFLECTION_OPTIMIZER
-	 * @see #getBytecodeProvider()
 	 * @see BytecodeProvider#getReflectionOptimizer
 	 *
 	 * @deprecated Deprecated to indicate that the method will be moved to
 	 * {@link org.hibernate.boot.spi.SessionFactoryOptions} /
-	 * {@link org.hibernate.boot.SessionFactoryBuilder} - probably in 6.0.
+	 * {@link org.hibernate.boot.SessionFactoryBuilder}.
 	 * See <a href="https://hibernate.atlassian.net/browse/HHH-12194">HHH-12194</a> and
 	 * <a href="https://hibernate.atlassian.net/browse/HHH-12193">HHH-12193</a> for details
 	 */
@@ -218,22 +215,10 @@ public final class Environment implements AvailableSettings {
 	}
 
 	/**
-	 * @deprecated Deprecated to indicate that the method will be moved to
-	 * {@link org.hibernate.boot.spi.SessionFactoryOptions} /
-	 * {@link org.hibernate.boot.SessionFactoryBuilder} - probably in 6.0.
-	 * See <a href="https://hibernate.atlassian.net/browse/HHH-12194">HHH-12194</a> and
-	 * <a href="https://hibernate.atlassian.net/browse/HHH-12193">HHH-12193</a> for details
-	 */
-	@Deprecated
-	public static BytecodeProvider getBytecodeProvider() {
-		return BYTECODE_PROVIDER_INSTANCE;
-	}
-
-	/**
 	 * Disallow instantiation
 	 */
 	private Environment() {
-		throw new UnsupportedOperationException();
+		//not to be constructed
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


Sanne:
@yrodiere would you like to review this one?

hibernate-core/src/main/java/org/hibernate/jpa/internal/enhance/EnhancingClassTransformerImpl.java
```diff
@@ -23,9 +26,12 @@
 public class EnhancingClassTransformerImpl implements ClassTransformer {
 
 	private final EnhancementContext enhancementContext;
+	private final BytecodeProvider bytecodeProvider;
 
 	public EnhancingClassTransformerImpl(EnhancementContext enhancementContext) {
+		Objects.requireNonNull( enhancementContext );
 		this.enhancementContext = enhancementContext;
+		this.bytecodeProvider = BytecodeProviderInitiator.buildDefaultBytecodeProvider();
```
yrodiere, line 34: Isn't that possibly introducing another problem? You won't be using the same bytecode provider as Hibernate ORM...

But I must admit I don't know what this class is used for anyway :/

hibernate-core/src/main/java/org/hibernate/jpa/internal/enhance/EnhancingClassTransformerImpl.java
```diff
@@ -23,9 +26,12 @@
 public class EnhancingClassTransformerImpl implements ClassTransformer {
 
 	private final EnhancementContext enhancementContext;
+	private final BytecodeProvider bytecodeProvider;
 
 	public EnhancingClassTransformerImpl(EnhancementContext enhancementContext) {
```
yrodiere, line 31: There is something weird going on.

This constructor is only ever called from here:

https://github.com/hibernate/hibernate-orm/blob/a986a3806a2756b73845a2e9942b5ea86b0893fd/hibernate-core/src/main/java/org/hibernate/jpa/boot/internal/PersistenceUnitInfoDescriptor.java#L111-L114

But then all implementations of `PersistenceUnitInfo#addTransformer` are no-ops...

This component is probably not being tested. I'd suggest tring your patch in WildFly before merging?

hibernate-core/src/main/java/org/hibernate/proxy/AbstractLazyInitializer.java
```diff
@@ -273,7 +273,11 @@ protected void prepareForPossibleLoadingOutsideTransaction() {
 		if ( session != null ) {
 			allowLoadOutsideTransaction = session.getFactory().getSessionFactoryOptions().isInitializeLazyStateOutsideTransactionsEnabled();
 
-			if ( allowLoadOutsideTransaction && sessionFactoryUuid == null ) {
+			if ( sessionFactoryUuid == null ) {
+				//we're going to need the UUID even if we the SessionFactory configuration doesn't
+				// allow any operations on it, as we need it to match deserialized objects with
+				// the originating SessionFactory: at very least it's useful to actually get
+				// such configuration, so to know if such operation isn't allowed or configured otherwise.
```
yrodiere, line 280: Looks like you've changed the purpose of this method; maybe it should be renamed?

I don't know if this class is API or "looks like API but actually is not", though...

hibernate-core/src/main/java/org/hibernate/proxy/pojo/bytebuddy/SerializableProxy.java
```diff
@@ -109,13 +110,32 @@ CompositeType getComponentIdType() {
...
-		HibernateProxy proxy = ( (BytecodeProviderImpl) bytecodeProvider ).getByteBuddyProxyHelper().deserializeProxy( this );
+		final SessionFactoryImplementor sessionFactory = retrieveMatchingSessionFactory( this.sessionFactoryUuid );
+		BytecodeProviderImpl byteBuddyBytecodeProvider = retrieveByteBuddyBytecodeProvider( sessionFactory );
+		HibernateProxy proxy = byteBuddyBytecodeProvider.getByteBuddyProxyHelper().deserializeProxy( this );
 		afterDeserialization( (ByteBuddyInterceptor) proxy.getHibernateLazyInitializer() );
 		return proxy;
 	}
+
+	private static SessionFactoryImplementor retrieveMatchingSessionFactory(final String sessionFactoryUuid) {
+		Objects.requireNonNull( sessionFactoryUuid );
```
yrodiere, line 121: We won't be able to deserialize entities serialized before this patch, since the uuid will be missing.

Not sure what the backwards compatibility expectations of serialization are, though.

hibernate-core/src/main/java/org/hibernate/proxy/pojo/bytebuddy/SerializableProxy.java
```diff
@@ -109,13 +110,32 @@ CompositeType getComponentIdType() {
...
+		}
+	}
+
+	private static BytecodeProviderImpl retrieveByteBuddyBytecodeProvider(final SessionFactoryImplementor sessionFactory) {
+		final BytecodeProvider bytecodeProvider = sessionFactory.getServiceRegistry().getService( BytecodeProvider.class );
+		if ( bytecodeProvider instanceof BytecodeProviderImpl ) {
+			return (BytecodeProviderImpl) bytecodeProvider;
+		}
+		else {
+			throw new IllegalStateException( "The bytecode provider is not ByteBuddy: unable to deserialize a SerializableProxy proxy." );
```
yrodiere, line 137: I know that's a pre-existing message, but don't we generally use the format "<Problem>: <reason>"?

```suggestion
			throw new IllegalStateException( "Unable to deserialize a SerializableProxy proxy: the bytecode provider is not ByteBuddy." );
```

hibernate-testing/src/main/java/org/hibernate/testing/bytecode/enhancement/BytecodeEnhancerRunner.java
```diff
@@ -182,7 +184,7 @@ private static ClassLoader buildEnhancerClassLoader(
 			EnhancementContext enhancerContext,
 			List<EnhancementSelector> selectors) {
 		return new EnhancingClassLoader(
-				Environment.getBytecodeProvider().getEnhancer( enhancerContext ),
+				buildDefaultBytecodeProvider().getEnhancer( enhancerContext ),
```
yrodiere, line 187: You usually reset caches explicitly at the end, even if the bytecode provider is going to get garbage collected, so I suppose there's a risk of memory leak of classloader pollution or something...

But here you're not doing it? Shouldn't you keep the bytecode provider around to reset caches after the runner finishes "running"?

yrodiere (commented):
Well that does looks dangerous, but I guess we like to live dangerously :) A few comments below.

By the way, does this also solve https://hibernate.atlassian.net/browse/HHH-12193 ?

Sanne:
> By the way, does this also solve https://hibernate.atlassian.net/browse/HHH-12193 ?

Right, thanks for finding that one.

hibernate-testing/src/main/java/org/hibernate/testing/bytecode/enhancement/BytecodeEnhancerRunner.java
Sanne: A missing resetCaches() method shouldn't leak the classloader if it's not held in a static field (like it originally was), as it's now held only from the ServiceRegistry instance; AFAIK it's just about freeing up some memory.
I've added the `resetCaches` in other places because it's a saner approach, but it shouldn't be strictly necessary - in fact it wasn't there before.

N.B. the `resetCaches` method didn't exist in the past, I've added it recently to reduce runtime consumption of a bootstrapped Quarkus application.

hibernate-core/src/main/java/org/hibernate/jpa/internal/enhance/EnhancingClassTransformerImpl.java
Sanne: This is used by JPA's "enhance at deployment time" capability, used on EE servers.

And yes I'd like to make the BytecodeProvider implementation pluggable, but it really isn't today: choices are exclusively `bytebuddy` or `none`. Clearly this use case can't use `none`.

(there are other areas too in which it's implied that it needs to be the ByteBuddy one - I'd certainly like to improve that but I'm considering that a different issue)

hibernate-core/src/main/java/org/hibernate/cfg/Environment.java
```diff
@@ -246,36 +231,27 @@ public static Properties getProperties() {
...
+	 */
+	@Deprecated
 	public static final String BYTECODE_PROVIDER_NAME_BYTEBUDDY = "bytebuddy";
-	public static final String BYTECODE_PROVIDER_NAME_NONE = "none";
-	public static final String BYTECODE_PROVIDER_NAME_DEFAULT = BYTECODE_PROVIDER_NAME_BYTEBUDDY;
 
-	public static BytecodeProvider buildBytecodeProvider(Properties properties) {
-		String provider = ConfigurationHelper.getString( BYTECODE_PROVIDER, properties, BYTECODE_PROVIDER_NAME_DEFAULT );
-		return buildBytecodeProvider( provider );
-	}
```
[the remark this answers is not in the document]
Sanne: ho dear, didn't notice that was public. Thanks, restoring this - and deprecating.

hibernate-core/src/main/java/org/hibernate/jpa/internal/enhance/EnhancingClassTransformerImpl.java
Sanne: Good idea, yes WildFly has a similar implementation but it's a different one.

hibernate-core/src/main/java/org/hibernate/proxy/pojo/bytebuddy/SerializableProxy.java
Sanne: The UUID was always supposed to be there in practice - I only noticed this wasn't happening consistently because we have some proxy serialization tests which were performing this while in the odd configuration of not having set the `sessionFactoryUuid` first.
However in a "real" scenario this would have failed as the proxy can't be initialized.

Sanne:
pushed a reviewed version

hibernate-core/src/main/java/org/hibernate/jpa/internal/enhance/EnhancingClassTransformerImpl.java
yrodiere: > And yes I'd like to make the BytecodeProvider implementation pluggable

I was thinking more that you're not using the same *instance* of `BytecodeProvider`, whereas before this patch, you were. I have no idea what kind of impact this can have, but since the bytecode provider is stateful, it might be a problem?

yrodiere (approved):
Thanks. LGTM as long as you're confident about the impact on `EnhancingClassTransformerImpl.java` and WildFly.
EDIT: And serialization :)

hibernate-core/src/main/java/org/hibernate/proxy/AbstractLazyInitializer.java
Sanne: Good catch, I've amended the javadoc to clarify.

Sanne:
:+1:  still checking WildFly, will merge if I spot no issues there.

Sanne:
All good on the WildFly side - if we bypass the lack of an upodated Infinispan 2LC module.

Merging

PR #5979, opened 2023-01-19
HHH-16059 Add getters and relax scopes

DavideD:
So that Hibernate Reactive can avoid copy and paste when extending CteInsertStrategy and CteMutationStrategy

Fix https://hibernate.atlassian.net/browse/HHH-16059


dreab8:
LGTM

PR #5983, opened 2023-01-19
HHH-16061 SqmDynamicInstantiation warns about dynamic Map instantiation when using an entity

dreab8:
https://hibernate.atlassian.net/browse/HHH-16061


sebersole (commented):
lgtm

PR #5987, opened 2023-01-20
HHH-15921@BatchSize and @IdClass on join column throws exception

dreab8:
https://hibernate.atlassian.net/browse/HHH-15921
https://hibernate.atlassian.net/browse/HHH-16039
https://hibernate.atlassian.net/browse/HHH-16025

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializerBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializerBuilder.java
@@ -1,0 +1,111 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.sql.results.graph.entity.internal;
+
+import org.hibernate.metamodel.internal.StandardEmbeddableInstantiator;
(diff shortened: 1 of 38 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializerBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+					parentAccess,
+					fetchedAttribute,
+					navigablePath,
+					entityPersister,
+					keyResult.createResultAssembler( parentAccess, creationState )
+			);
+		}
+	}
+
+	private static boolean isEmbeddedIdPart(FetchParentAccess parentAccess) {
```
beikov, line 77: Let's rename that to e.g. `shouldDisableBatching` and also add checks for the `EmbeddableValuedModelPart` to see if it has a custom instantiator, since when that is present, we can't inject batch loaded values.

beikov: Not sure if we want to handle this as part of this PR as well, but there are other issues related to batch fetching where we will have to disable batching:

* https://hibernate.atlassian.net/browse/HHH-16025
* https://hibernate.atlassian.net/browse/HHH-16039

beikov: Also, it would be nice to have a sort of batching implementation where we return a proxy for the initialized instance (if possible) and only batch initialize a these objects in the finishUp method.

PR #5988, opened 2023-01-20
HhHH-15921 @BatchSize and @IdClass on join column throws exception

dreab8:
https://hibernate.atlassian.net/browse/HHH-15921


dreab8:
rejected

PR #5991, opened 2023-01-20
HHH-16043 Hibernate 6.x breaks collection batch fetching

mbladel:
https://hibernate.atlassian.net/browse/HHH-16043


hibernate-core/src/main/java/org/hibernate/loader/ast/internal/CollectionLoaderBatchKey.java
```diff
@@ -97,7 +100,7 @@ public PersistentCollection<?> load(
 					null,
 					attributeMapping.getKeyDescriptor(),
 					null,
-					batchSize,
```
sebersole, line 100: Grr, good catch

beikov (approved):

PR #5992, opened 2023-01-20
HHH-16069 - Defer handling of custom types until after deployment time

sebersole:
This form simply falls back to non-CDI creation if the BeanManager is not yet ready to use.

Not sure how kosher this is as it happens inconsistently such that some UserType implementations get resolved via CDI and others not.  The inconsistency concerns me.

Going to create an alternative that simply never uses CDI for UserType when using the extended or delayed strategies


vpriscan:
Answer to apparent inconsistencies:
No user type is resolved by CDI, but the difference is that user types that **do not** implement DynamicParameterizedType end up in JpaCompliantLifecycleStrategy.BeanImpl.initialize(), which has a null check for beanManager and falls back to fallbackProducer. On the other hand user types that **do** implement DynamicParameterizedType end up in JpaCompliantLifecycleStrategy.NamedBeanImpl.initialize() that does not have a null check for beanManager and throws NotYetReadyException unconditionally.
Fixed by [6018](https://github.com/hibernate/hibernate-orm/pull/6018)

sebersole:
The report was specifically for custom type resolution.  But the solution fixes this for all non-JPA-required beans.

PR #5994, opened 2023-01-20
HHH-16069 - Defer handling of custom types until after deployment time

sebersole:
This form circumvents CDI hosting of UserType impls when extended or delayed strategies are used.

Alternative to https://github.com/hibernate/hibernate-orm/pull/5992


hibernate-core/src/main/java/org/hibernate/resource/beans/spi/ManagedBeanRegistry.java
```diff
@@ -24,15 +24,48 @@ public interface ManagedBeanRegistry extends Service {
...
 
 	/**
-	 * Get a bean reference by name and contract.
+	 * Form of {@link #getBean(Class)} allowing to indicate whether
+	 * it is required to use CDI if it is available
+	 *
+	 * @implNote Defaulted for backwards compatibility
 	 */
-	<T> ManagedBean<T> getBean(String beanName, Class<T> beanContract);
+	default <T> ManagedBean<T> getBean(Class<T> beanClass, boolean cdiRequiredIfAvailable) {
```
yrodiere, line 32: You might want to use an enum instead of a boolean, both for more explicit code on call sites and to be free to add other behavior later if need be?

scottmarlow (approved):

hibernate-core/src/main/java/org/hibernate/resource/beans/container/internal/CdiBeanContainerExtendedAccessImpl.java
```diff
@@ -41,8 +41,12 @@ private CdiBeanContainerExtendedAccessImpl(ExtendedBeanManager beanManager) {
 	protected <B> ContainedBeanImplementor<B> createBean(
 			Class<B> beanType,
 			BeanLifecycleStrategy lifecycleStrategy,
-			BeanInstanceProducer fallbackProducer) {
-		if ( usableBeanManager == null ) {
+			BeanInstanceProducer fallbackProducer,
+			boolean cdiRequiredIfAvailable) {
+		if ( !cdiRequiredIfAvailable ) {
```
yrodiere, line 46: Shouldn't the `cdiRequiredIfAvailable` flag be ignored if `usableBeanManager` is non-null (and thus CDI is already available)?

E.g.:

```suggestion
		if ( !cdiRequiredIfAvailable && usableBeanManager == null ) {
```

That way if the caller is lucky and CDI was already initialized, they can get a CDI bean.

hibernate-core/src/main/java/org/hibernate/resource/beans/container/spi/AbstractCdiBeanContainer.java
```diff
@@ -23,35 +23,50 @@
...
+	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	// un-named bean support
 
 	@Override
-	public <B> ContainedBean<B> getBean(
+	public final <B> ContainedBean<B> getBean(
 			Class<B> beanType,
 			LifecycleOptions lifecycleOptions,
 			BeanInstanceProducer fallbackProducer) {
+		return getBean( beanType, lifecycleOptions, fallbackProducer, false );
```
yrodiere, line 37: This changes the default behavior, though... Including for Hibernate Search in the "delayed" case. I'm not sure of the implications.

Since you're (likely) always going to call `getBean()` with the explicit `cdiRequiredIfAvailable` flag yourself, can you please preserve the legacy behavior for this method? Just for backwards compatibility of integrations?

hibernate-core/src/main/java/org/hibernate/resource/beans/container/spi/BeanContainer.java
```diff
@@ -27,9 +27,38 @@ <B> ContainedBean<B> getBean(
...
+	 * Form of {@link #getBean(Class, LifecycleOptions, BeanInstanceProducer)}
+	 * allowing to indicate whether it is required to use CDI if it is available
+	 *
+	 * @implNote Defaulted for backwards compatibility
+	 */
+	default <B> ContainedBean<B> getBean(
+			Class<B> beanType,
+			LifecycleOptions lifecycleOptions,
+			BeanInstanceProducer fallbackProducer,
+			boolean cdiRequiredIfAvailable) {
```
yrodiere, line 40: Maybe another name would be better, since `BeanContainer` can be used for non-CDI implementations? E.g. `beanContainerRequiredIfAvailable` or something like that.

hibernate-core/src/test/java/org/hibernate/orm/test/cdi/general/nonregistrymanaged/NonRegistryManagedBeanConsumingIntegrator.java
```diff
@@ -69,68 +69,80 @@ public void integrate(Metadata metadata, SessionFactoryImplementor sessionFactor
 		applicationScopedBean1 = (ContainedBeanImplementor) beanContainer.getBean(
 				TheApplicationScopedBean.class,
 				this,
-				fallbackBeanInstanceProducer
+				fallbackBeanInstanceProducer,
+				true
```
yrodiere, line 73: I can say for sure that existing integrators using this API (i.e. Hibernate Search) do not, at the moment, use `true`.

Since these tests are specifically about checking compatibility with Hibernate Search, it's probably better to keep calling the old method without the `cdiRequiredIfAvailable` flag?

yrodiere (changes requested):
Thanks. A few comments below.

I'm mainly concerned about changing the behavior of `getBean` in `BeanContainer`, since Hibernate Search relies on that.

PR #5999, opened 2023-01-23
HHH-15944 Joins on non-managed-type polymorphisms don't work

dreab8:
https://hibernate.atlassian.net/browse/HHH-15944


hibernate-core/src/main/java/org/hibernate/query/hql/internal/QuerySplitter.java
```diff
@@ -183,6 +188,10 @@ private static SqmRoot<?> findUnmappedPolymorphicReference(SqmDeleteOrUpdateStat
 		return null;
 	}
 
+	public interface FromCopyProvider {
```
sebersole, line 191: This reference is only ever used internally right?  Assuming so, an interface is probably overkill

sebersole: Oops, no I see now you expose use it in other classes

beikov: Andrea and I talked about this today, and we came to the conclusion that this can be done differently in a non layer breaking manner.

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/6043

PR #6000, opened 2023-01-23
clean up impl of query hint interpretation

gavinking:
use 'switch' statements


gavinking:
I'm going to go right ahead and just assume that the failure on Sybase is nothing to do with this change.

PR #6007, opened 2023-01-24
List of all configuration properties in the user guide

marko-bekhta:
I've drafted something together based on the annotation processor we've used in Hibernate Search.

Original AP also used an annotation that allowed more granular configuration if some properties should be collected/ignored or if there are multiple prefixes for a property, and so on. I haven't removed that one yet, as it might still be helpful in ORM as well. If we decide that we want to use the annotation to group settings into sections using this annotation - the trouble with it would be that most of the properties live in a single interface - `AvailableSettings` so that would mean we'd need to annotate each property. 

Then there are Gradle tasks - I've made it, so the things work, but I bet there's a nicer/cleaner way of doing this - so any feedback on that is more than welcome 😃 

Here's how it looks in the user guide docs:
image

Also attached zip has a built user doc if someone wants to take a look without building this on their own.
[docs.zip](https://github.com/hibernate/hibernate-orm/files/10489896/docs.zip)


sebersole:
I applied this upstream.  Thanks Marko!

PR #6009, opened 2023-01-24
Update memory config for CockroachDB configuration

maesenka:
With the previous setting the cockroach process is killed by the OOM-killer.


Sanne:
thanks!

PR #6010, opened 2023-01-24
Backport of "HHH-15709  HQL local datetime literals on Postgres, Cockroach H2"

reda-alaoui:


beikov:
We don't do backports to 6.1 anymore. Switch to 6.2 please.

PR #6015, opened 2023-01-25
HHH-16070 Exception when find by association id that is a generic @EmbeddedId with @MappedSuperclass

mbladel:
https://hibernate.atlassian.net/browse/HHH-16070

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/internal/PropertyBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/internal/PropertyBinder.java
@@ -8,6 +8,7 @@ package org.hibernate.boot.model.internal;
 
 import java.lang.annotation.Annotation;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/internal/PropertyBinder.java
```diff
@@ -342,6 +336,65 @@ private Property bind(Property property) {
...
+	private void setDeclaredIdentifier(RootClass rootClass, MappedSuperclass superclass, Property prop) {
+		if ( superclass == null ) {
+			rootClass.setDeclaredIdentifierProperty( prop );
+			return;
+		}
+		// If the type has type parameters, we have to set the declared identifier property on the rootClass
+		// to be able to retrieve it with the correct type based on type variable assignment in the subclass
+		final Class<?> type = buildingContext.getBootstrapContext().getReflectionManager().toClass( declaringClass );
+		if ( type.getTypeParameters().length == 0 ) {
+			rootClass.setDeclaredIdentifierProperty( prop );
```
mbladel, line 348: I think this should be
```suggestion
			superclass.setDeclaredIdentifierProperty( prop );
```

PR #6016, opened 2023-01-25
6.0 migration guide: heading fix + mention dialects

yrodiere:
I noticed the deprecation of spatial or version-specific dialects was not mentioned in the migration guide, so here's an attempt to fix that.


gavinking:
Thank you @yrodiere, I noticed this myself a few days ago, and didn't find time to fix it.

beikov:
Thanks!

PR #6018, opened 2023-01-25
HHH-16096 Passing an ExtendedBeanManager which is notified too late leads to initialization error

vpriscan:
Add a null check for BeanManager in org.hibernate.resource.beans.container.internal.JpaCompliantLifecycleStrategy.NamedBeanImpl.initialize and use fallback producer if null.

[HHH-16096](https://hibernate.atlassian.net/browse/HHH-16096)

[HHH-16096]: https://hibernate.atlassian.net/browse/HHH-16096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/resource/beans/container/internal/JpaCompliantLifecycleStrategy.java
+++ b/hibernate-core/src/main/java/org/hibernate/resource/beans/container/internal/JpaCompliantLifecycleStrategy.java
@@ -223,6 +223,20 @@ public class JpaCompliantLifecycleStrategy implements BeanLifecycleStrategy {
 				return;
 			}
 
+			if ( beanManager == null ) {
+				try {
+					beanInstance = fallbackProducer.produceBeanInstance( beanType );
+					return;
+				}
+				catch (Exception e) {
+					// the CDI BeanManager is not know to be ready for use and the
+					// fallback-producer was unable to create the bean...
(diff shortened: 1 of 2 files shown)


yrodiere:
@vpriscan I think what you're trying to do was solved by https://github.com/hibernate/hibernate-orm/pull/6027?

Can you please check and report back here?

yrodiere (changes requested):
Waiting for feedback

vpriscan:
@yrodiere Yes, the root issue (initialization error) has been avoided by that PR, 
but this PR still adds a null check and a fallback in JpaCompliantLifecycleStrategy.NamedBeanImpl.initialize for completeness and consistency with JpaCompliantLifecycleStrategy.BeanImpl.initialize, where the exact same null check and fallback exists.

yrodiere (commented):
Ah, I didn't see that. Makes sense.

I'm a bit surprised https://hibernate.atlassian.net/browse/HHH-14914 introduced that code to begin with, though. I'll let @sebersole confirm it's still relevant.

sebersole:
The build failure on Jenkins is unrelated.  Going to integrate this.  Thanks @vpriscan !

PR #6022, opened 2023-01-25
HHH-16090 BasicResultAssembler logging optimisations

Sanne:
Very minor optimisation
 - https://hibernate.atlassian.net/browse/HHH-16090


sebersole:
Changes like this make me think we should just use protected SYSOUT statements ;)

lgtm

PR #6024, opened 2023-01-25
HHH-16020 - Fix for error that surfaced in the LegacyOracleLimitHandler

jrenaat:
... due to the fix for HHH-16020

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/pagination/LegacyOracleLimitHandler.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/pagination/LegacyOracleLimitHandler.java
@@ -62,6 +62,16 @@ public class LegacyOracleLimitHandler extends AbstractLimitHandler {
 	}
 
 	@Override
+	public boolean supportsOffset() {
+		return true;
+	}
+
+	@Override
+	public boolean forceLimitUsage() {
+		return true;
+	}
+
+	@Override
 	public boolean bindLimitParametersInReverseOrder() {
 		return true;
 	}


jrenaat:
@beikov This should fix the failing `NativeQueryLimitOffsetTest.testPartialLimitOffsetOnNativeQuery()` for Oracle 11, could you check what CI has to say about it?
I tested it locally against Oracle 11, and found that 2 tests of the LockTest were failing. After some debugging it seemed not related to this change, and when running the tests on main, these same 2 tests also fail, so I assume that these failures are unrelated.

hibernate-core/src/main/java/org/hibernate/dialect/pagination/LegacyOracleLimitHandler.java
```diff
@@ -61,6 +61,16 @@ public boolean supportsLimit() {
...
 
+	@Override
+	public boolean supportsOffset() {
+		return true;
+	}
+
+	@Override
+	public boolean forceLimitUsage() {
+		return true;
+	}
```
beikov, line 72: In 5.6 we never had an implementation that returned anything other than `false` for `forceLimitUsage`, so can you explain to me why we need this?

jrenaat: For Oracle 11, the query ends up being:
`select *
from
    (
        select row_.*,rownum rownum_
        from (
                select id
                from Person
                where name like ?
            ) row_
        where rownum<=?
    )
where rownum_>?`
Hence, a limit parameter is needed, and it is set in `AbstractLimitHandler.bindLimitParameters()`. However for the limit parameter to be set, it needs `hasMaxRows && supportsLimit() || forceLimitUsage();`. Hence, since there is no maxrows indicated, I solved it by overriding forceLimitUsage.
It's either that, or the Oracle 11 dialect should not allow an offset without also indicating a max rows, I think?

sebersole:
> I tested it locally against Oracle 11, and found that 2 tests of the LockTest were failing. After some debugging it seemed not related to this change

Yes, I've seen this as well.

PR #6025, opened 2023-01-25
Draft solution for excessive logging via ignoreImmutablePropertyModification

Sanne:
**A very unpolished draft** for the problem discussed on https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/topic/Exteme.20amounts.20of.20warnings.20logged

Essentially it seems we flag properties as Dirty also when not having loaded the original property : in such cases it needs to be treated like dirty, but also in this case we shouldn't be blaming the user for having done something wrong.

So I'm experimenting with changing `boolean isDirty();` into returning a three-state boolean: DIRTY,  NOT_DIRTY and LIKELY_DIRTY.

LIKELY_DIRTY and DIRTY are to be treated the same in terms of ORM semantics, but when there's no full certainty we can avoid logging the warning.


Sanne:
Closing, replaced by https://github.com/hibernate/hibernate-orm/pull/6028

PR #6026, opened 2023-01-25
HHH-4299 - Added test for issue

jrenaat:


jrenaat:
Test case to prove that HHH-4299 is no longer an issue

beikov:
But `EmbeddableWithGenericAndMappedSuperClassTest` seems to fail for some reason, so maybe it's still an issue?

jrenaat:
> But `EmbeddableWithGenericAndMappedSuperClassTest` seems to fail for some reason, so maybe it's still an issue?

Yeah, it passed on H2 (locally), so I assumed it had been fixed, but I just rebased main onto it, and now it also fails locally for me with the same exception, which makes me think that some recent commit broke something ... have to investigate a bit more.

PR #6027, opened 2023-01-25
HHH-16069 - Skip CDI for Hibernate extensions by default

sebersole:
This is a fourth option for dealing with "deferred" CDI access.

In this one we only look for converters and listeners in CDI by default.  To enable resolving Hibernate extensions in CDI, a setting (`hibernate.cdi.extensions`, false by default) has been introduced

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataBuilderImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataBuilderImpl.java
@@ -607,6 +607,7 @@ public class MetadataBuilderImpl implements MetadataBuilderImplementor, TypeCont
 
 		private final String schemaCharset;
 		private final boolean xmlMappingEnabled;
+		private final boolean allowExtensionsInCdi;
 
 		public MetadataBuildingOptionsImpl(StandardServiceRegistry serviceRegistry) {
 			this.serviceRegistry = serviceRegistry;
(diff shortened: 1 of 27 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/TypeDefinition.java
```diff
@@ -137,6 +137,9 @@ private static BasicValue.Resolution<?> createResolution(
 			Map<?,?> usageSiteProperties,
 			JdbcTypeIndicators indicators,
 			MetadataBuildingContext context) {
+		if ( context.getBuildingOptions().disallowExtensionsInCdi() ) {
+
```
andrei-ivanov, line 141: leftover?

sebersole: Yep.  Thanks!

hibernate-core/src/main/java/org/hibernate/boot/internal/MetadataBuilderImpl.java
```diff
@@ -951,6 +958,11 @@ public boolean isXmlMappingEnabled() {
 			return xmlMappingEnabled;
 		}
 
+		@Override
+		public boolean disallowExtensionsInCdi() {
```
cloudyster, line 962: Hmm, this looks strange to me
`disallowExtensionsInCdi()` returns `allowExtensionsInCdi`. Shouldn't it return `!allowExtensionsInCdi`?

PR #6030, opened 2023-01-26
HHH-16103 - MERGE for optional table updates on SQL Server

sebersole:
https://hibernate.atlassian.net/browse/HHH-16103


daangroot2:
@sebersole Hi, this has introduced a rounding issue for DECIMAL columns with a scale greater than 2, because values is being cast to numeric(38,2). Setting a precision and scale in @Column does not have any effect.

PR #6033, opened 2023-01-26
HHH-16108 NullPointerException when flushing a (simple) entity update for models with bytecode enhancement and multiple one-to-one associations (some lazy)

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16108

Reproducer only, no fix.


yrodiere:
Superseded by #6034

PR #6038, opened 2023-01-26
HHH-15964 Incorrect results for pageable EntityGraph with Hibernate 6.1.6

reda-alaoui:


hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -6,6 +6,12 @@
  */
 package org.hibernate.query.sqm.internal;
 
+import static org.hibernate.jpa.HibernateHints.*;
```
sebersole, line 9: We prefer no `.*` imports.

sebersole (commented):
Aside from the code style comment, this looks good to me.  Going to let the CI runs finish to verify.

Thanks!

reda-alaoui:
@sebersole , https://github.com/Cosium/hibernate-orm/blob/e950cd882cb5cc06f80c1834e2b7c084d932004c/hibernate-core/src/test/java/org/hibernate/orm/test/entitygraph/EntityGraphWithFetchAnnotationTest.java#L75-L75 is failing. 
I think I need to be able to retrieve the FetchMode from https://github.com/Cosium/hibernate-orm/blob/e950cd882cb5cc06f80c1834e2b7c084d932004c/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/AppliedGraphs.java#L26-L26 to eventually return false. Could you tell me how to retrieve that ?

reda-alaoui:
> to be able to retrieve the FetchMode

Or maybe the FetchTiming?

reda-alaoui:
I had to expose `org.hibernate.FetchMode` on `org.hibernate.metamodel.model.domain.PersistentAttribute` to consider EntityGraph collection nodes flagged with `FetchMode.Select` as non collection fetches.
Hope it makes sense to you :)

reda-alaoui:
Ok, we can forget everything I said before.
Waiting for your review :)

sebersole:
We are unfortunately dealing with 2 separate metamodels -

1. First is the JPA-based metamodel (`jakarta.persistence.metamodel`) which we call the "domain metamodel" because it does not contain "mapping" information.  
2. Second is the runtime "mapping model", is the one that understands aspects of fetching (timing and style).

It should be possible to expose the fetch timing (really only the timing is important) from the "domain metamodel".

> consider EntityGraph collection nodes flagged with FetchMode.Select as non collection fetches

This perfectly illustrates the distinction between fetch timing and style.  An attribute with `FetchMode.Select` might be either eager or lazy.  The SELECT aspect simply describes how we will load it when we need to load it.  It is that "when" part that is important here.

Does that all make sense?

reda-alaoui:
> We are unfortunately dealing with 2 separate metamodels -
> 
> 1. First is the JPA-based metamodel (`jakarta.persistence.metamodel`) which we call the "domain metamodel" because it does not contain "mapping" information.
> 2. Second is the runtime "mapping model", is the one that understands aspects of fetching (timing and style).
> 
> It should be possible to expose the fetch timing (really only the timing is important) from the "domain metamodel".
> 
> > consider EntityGraph collection nodes flagged with FetchMode.Select as non collection fetches
> 
> This perfectly illustrates the distinction between fetch timing and style. An attribute with `FetchMode.SELECT` might be either eager or lazy. The SELECT aspect simply describes how we will load it when we need to load it. It is that "when" part that is important here.
> 
> Does that all make sense?

Yes thank you. 

Although, in the end, I didn't need to expose such information :)

reda-alaoui:
My changes caused damages on interaction between EntityGraph and the 2nd level cache. I found this on my own project tests. I added a test ( `CacheableEntityGraphTest` ) to cover this case.

There are multiple issues linked to HHH-15964 :
1. The one described by HHH-15964 . It is caused by the fact that `QuerySqmImpl#doList` wasn't looking at the EntityGraph for collection fetches, leading to cartesian product result. Covered by `EntityGraphTest#paginationOverCollectionFetch`
2. A collection annotated with `EAGER` exhibited 2 behaviours depending on the presence or not of an *empty loadgraph*. No issue without  *empty loadgraph*. With  *empty loadgraph*, this leads to cartesian product result. This was caused by the fact that `StandardEntityGraphTraversalStateImpl.traverse`, can return `joined == true` for an *empty loadgraph*. For the same case without EntityGraph, joined is false. Covered by `EntityGraphTest#paginationOverEagerCollectionWithEmptyEG`.
3. My fix lead to `CacheableEntityGraphTest` failing with `Unable to resolve owner of loading collection [[com.cos.hibernate.cache.SecondLevelCacheTest$Tag.types#1]] for second level caching`

A wild guess would be that all could be resolved if `QuerySqmImpl` had the exact same algorithm as `StandardEntityGraphTraversalStateImpl#traverse` to determine the presence of a collection fetch. 
Currently, in my change, `QuerySqmImpl` relies on https://github.com/hibernate/hibernate-orm/blob/2c78d83b5a47bbaee0a5522d0167c944634bcedd/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/ConcreteSqmSelectQueryPlan.java#L172 which is only interested by the explicit passed EntityGraph. I think it should also be aware of the implicit graph (e.g. the one formed by annotations hints? ).

reda-alaoui:
@sebersole , many tests seem to expect a different behaviour between:
* an "implicit" fetch ( i.e. `ManyToOne(fetch = EAGER)` ) 
and 
* an "explicit" fetch (An attribute part of an EntityGraph)

I think it is abnormal. But given the number of tests with this expectation, could you give me the rationale behind this? Many of those tests now fail because they were expecting multiple select but now face a single one.

gavinking:
> I think it is abnormal. 

It's totally normal because `ManyToOne(fetch = EAGER)` is always, and _should_ always be ignored in queries.

The `fetch = EAGER` setting is _only_ for when an entity is fetched by its id.

reda-alaoui:
Hello @gavinking,

Thank you for your answer.

I didn't know that. Is that specified by the JPA specification or a choice of implementation? Because I can't find this in the [jakarta specification](https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html).

reda-alaoui:
Also, executing a query with an `empty loadgraph` or `no graph at all` can lead to 2 different behaviours. Is that normal?

gavinking:
Seems obvious that if there is no `without join fetch` syntax, then fetching must be disabled by default.

Also I designed this shit so I know that’s how it’s supposed to work :)

gavinking:
But, sure, the JPA spec is intentionally silent on exactly how and when associations are fetched, so, sure, it’s an “implementation decision” from that point of view.

reda-alaoui:
> Seems obvious that if there is no `without join fetch` syntax, then fetching must be disabled by default.
> 
> Also I designed this shit so I know that’s how it’s supposed to work :)

https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/sql/results/internal/StandardEntityGraphTraversalStateImpl.java#L93-L96 overrides https://github.com/hibernate/hibernate-orm/blob/cd50a98c665d61659a46b1aefa957418df94f5f3/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java#L7150-L7150 when an *empty* loadgraph is passed. On the other hand, if no graph is passed, `joined` is kept untouched. I guess we can agree it's a bug?

Sorry for all those naive questions, I am just trying to fix the original issue within the design contraints that I am not aware of :)

sebersole:
No graph and an empty graph are different though depending on the fetch-graph v. load-graph semantic.

This stuff is quite difficult to grok, so no worries about questions

reda-alaoui:
Thank you Steve.

> No graph and an empty graph are different though depending on the fetch-graph v. load-graph semantic.

So I understand that `no graph` should be equivalent to `empty loadgraph`.

> It's totally normal because ManyToOne(fetch = EAGER) is always, and should always be ignored in queries.
> The fetch = EAGER setting is only for when an entity is fetched by its id.

Does this rule only apply to `ManyToOne` ? i.e. Does it also apply to `OneToOne` and `OneToMany` ?

sebersole:
> > No graph and an empty graph are different though depending on the fetch-graph v. load-graph semantic.
> 
> So I understand that `no graph` should be equivalent to `empty loadgraph`.

Honestly I have to go back and read the spec every time wrt fetch v. load.  This stuff is extremely convoluted imho.


> > It's totally normal because ManyToOne(fetch = EAGER) is always, and should always be ignored in queries.
> > The fetch = EAGER setting is only for when an entity is fetched by its id.
> 
> Does this rule only apply to `ManyToOne` ? i.e. Does it apply to `OneToOne` and `OneToMany` ?

This comes back to the notion of fetch timing versus fetch style.  The easiest way to remember this is that *all* queries ignore all "mapped fetching" in terms of style.  So for a query we will never *join* fetch (the style) an association regardless of cardinality (one-to-one, etc).  Now timing does dictate we may have to still immediately fetch the association, but that will always be a subsequent (n+1) select.

This is where entity-graph and query join fetches come in to play.

sebersole:
> think it is abnormal. But given the number of tests with this expectation, could you give me the eventual rationale behind this?

The expectation about entity-graphs simply apply to what gets loaded, not how.  I think that is probably your confusion

gavinking:
Yeah the load graph vs fetch graph stuff isn’t sensible or intuitive at all. IIRC one is an override, the other one starts from zero like a query. But it’s impossible to remember which is which because the  names don’t reflect the difference at all.

reda-alaoui:
Ready for review.

This change:
1. fixes the original issue by checking the EntityGraph for collection fetches using `AppliedGraphs#containsCollectionFetches`
2. aligns `empty loadgraph`  behaviour on `absent graph` behaviour. According to our discussion, those behaviours should be aligned. Also this is the easiest way to allow `AppliedGraphs#containsCollectionFetches` to make correct predictions about what is going to happen in `StandardEntityGraphTraversalStateImpl#traverse` .
3. adapt failing tests to the change brought by the previous point (2.).

sebersole:
> When the javax.persistence.fetchgraph property is used to specify an entity graph,
attributes that are specified by attribute nodes of the entity graph are treated as FetchType.EAGER
and attributes that are not specified are treated as FetchType.LAZY.
> ...
> When the javax.persistence.loadgraph property is used to specify an entity graph, attributes
that are specified by attribute nodes of the entity graph are treated as FetchType.EAGER and
attributes that are not specified are treated according to their specified or default FetchType.

The main distinction being how to handle "attributes that are not specified".  However, as I said, there is also an "operational" component to this - between loads and queries.  

You keep coming back to "loadgraphs", so let's focus there for now.  There, in fact, you are correct that an empty and an absent graph should have the same outcome.  Since you have effectively not "specified attributes" the attributes are eager/lazy (timing) relative to their mapping.  For associations mapped as eager, the style will be some form of subsequent select (select, batch, subselct, ...).

Always bear in mind also that the spec simply declares the minimum that must be fetched; Hibernate is free to fetch more.

> Adapts failing tests to the change brought by the previous point (2.).

We'll have to look at those a little more closely.  It is generally a no go to change tests to match outcomes with code changes.  The tests are generally there for regression protection.

reda-alaoui:
> We'll have to look at those a little more closely. It is generally a no go to change tests to match outcomes with code changes. The tests are generally there for regression protection.

Yes I was expecting this point to be the most controversial. But could not find a better way.

gavinking:
> > > It's totally normal because `@ManyToOne(fetch = EAGER)` is always, and should always be ignored in queries.
> > > The `fetch = EAGER` setting is only for when an entity is fetched by its id.
> > 
> > 
> > Does this rule only apply to `ManyToOne` ? i.e. Does it apply to `OneToOne` and `OneToMany` ?
> 
> This comes back to the notion of fetch timing versus fetch style. The easiest way to remember this is that _all_ queries ignore all "mapped fetching" in terms of style. So for a query we will never _join_ fetch (the style) an association regardless of cardinality (one-to-one, etc). Now timing does dictate we may have to still immediately fetch the association, but that will always be a subsequent (n+1) select.

Sorry, I just noticed that what I wrote (while perhaps _literally true_) was misleading. @sebersole's description is _much_ better.

Queries *do* ignore `fetch = EAGER`, but that doesn't mean the association won't be eagerly fetched using a subsequent load-by-id. As Steve says, this distinction is a bit subtle.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -510,7 +511,7 @@ protected List<R> doList() {
 		getSession().prepareForQueryExecution( requiresTxn( getQueryOptions().getLockOptions().findGreatestLockMode() ) );
 
 		final SqmSelectStatement<?> sqmStatement = (SqmSelectStatement<?>) getSqmStatement();
-		final boolean containsCollectionFetches = sqmStatement.containsCollectionFetches();
+		final boolean containsCollectionFetches = sqmStatement.containsCollectionFetches() || AppliedGraphs.containsCollectionFetches(getQueryOptions());
```
sebersole, line 514:
```suggestion
		final boolean containsCollectionFetches = sqmStatement.containsCollectionFetches() || AppliedGraphs.containsCollectionFetches( getQueryOptions() );
```

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -534,7 +535,10 @@ protected List<R> doList() {
 				else {
 					toIndex = resultSize;
 				}
-				return list.subList( first, toIndex > resultSize ? resultSize : toIndex );
+				if (first > resultSize) {
```
sebersole, line 538:
```suggestion
				if ( first > resultSize ) {
```

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -534,7 +535,10 @@ protected List<R> doList() {
...
+					return new ArrayList<>(0);
+				}
+				return list.subList( first, toIndex > resultSize ? resultSize : toIndex);
```
sebersole, line 541:
```suggestion
				return list.subList( first, toIndex > resultSize ? resultSize : toIndex );
```

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -7214,10 +7214,14 @@ private Fetch createFetch(FetchParent fetchParent, Fetchable fetchable, Boolean
 							isKeyFetchable
 					);
 
-					fetchTiming = traversalResult.getFetchTiming();
-					joined = traversalResult.isJoined();
-					if ( shouldExplicitFetch( maxDepth, fetchable ) ) {
-						explicitFetch = true;
+					EntityGraphTraversalState.FetchStrategy fetchStrategy = traversalResult.getFetchStrategy();
+					if (fetchStrategy != null) {
```
sebersole, line 7218:
```suggestion
					if ( fetchStrategy != null ) {
```

hibernate-core/src/main/java/org/hibernate/sql/results/internal/StandardEntityGraphTraversalStateImpl.java
```diff
@@ -48,7 +48,7 @@ public void backtrack(TraversalResult previousContext) {
 	public TraversalResult traverse(FetchParent fetchParent, Fetchable fetchable, boolean exploreKeySubgraph) {
 		assert !(fetchable instanceof CollectionPart);
 		if ( fetchable instanceof NonAggregatedIdentifierMapping ) {
-			return new TraversalResult( currentGraphContext, FetchTiming.IMMEDIATE, true );
+			return new TraversalResult( currentGraphContext, new FetchStrategy(FetchTiming.IMMEDIATE, true) );
```
sebersole, line 51:
```suggestion
			return new TraversalResult( currentGraphContext, new FetchStrategy( FetchTiming.IMMEDIATE, true ) );
```

hibernate-core/src/main/java/org/hibernate/sql/results/internal/StandardEntityGraphTraversalStateImpl.java
```diff
@@ -58,12 +58,11 @@ public TraversalResult traverse(FetchParent fetchParent, Fetchable fetchable, bo
...
 		currentGraphContext = null;
-		FetchTiming fetchTiming = null;
-		boolean joined = false;
+		FetchStrategy fetchStrategy = null;
 
 		if ( attributeNode != null ) {
-			fetchTiming = FetchTiming.IMMEDIATE;
-			joined = true;
+
+			fetchStrategy = new FetchStrategy(FetchTiming.IMMEDIATE, true);
```
sebersole, line 65:
```suggestion
			fetchStrategy = new FetchStrategy( FetchTiming.IMMEDIATE, true );
```

sebersole (changes requested):
Lot of code style problems.  I applied suggestions.  Would be great if you could follow style guidelines in the future, makes reviewing much simpler.

As it is, it is a bit hard to follow the logic completely among all of the codestyle suggestion comments.  Apply those and I'll come back and take a look.

I believe your fix centers around the change to `EntityGraphTraversalState` but want to look at it again with fresh diffs.

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/AppliedGraphs.java
```diff
@@ -0,0 +1,43 @@
...
+
+	private AppliedGraphs() {
+	}
+
+	public static boolean containsCollectionFetches(QueryOptions queryOptions) {
+		final AppliedGraph appliedGraph = queryOptions.getAppliedGraph();
+		return appliedGraph != null && appliedGraph.getGraph() != null && containsCollectionFetches(appliedGraph.getGraph());
+	}
+
+	private static boolean containsCollectionFetches(GraphImplementor<?> graph) {
```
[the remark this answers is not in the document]
reda-alaoui: @sebersole , `AttributeNode` and `GraphImplementor` can be mutable. `AttributeNode` also holds a map of mutable `SubGraphImplementor` . 

IMO, tracking this (having a boolean attribute `containsCollectionAttribute` updated on each mutation and accessible from RootGraph) can't be done without some refactoring of those classes. For example:
* making https://github.com/Cosium/hibernate-orm/blob/b0867fb96ed7e1f2c903903c54bc6c6f7ecabe10/hibernate-core/src/main/java/org/hibernate/graph/spi/AttributeNodeImplementor.java#L24-L24 return unmodifiable maps
* wrapping https://github.com/Cosium/hibernate-orm/blob/b0867fb96ed7e1f2c903903c54bc6c6f7ecabe10/hibernate-core/src/main/java/org/hibernate/graph/internal/AttributeNodeImpl.java#L36-L36 SubGraphImplementor allowing to track mutations from the parent node

I can do that but maybe you had something simpler in mind?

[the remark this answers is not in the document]
sebersole: > IMO, tracking this (having a boolean attribute `containsCollectionAttribute` updated on each mutation and accessible from RootGraph) can't be done without some refactoring of those classes.

Fair.  I forgot these are acyclic trees - thought they had references to their container which we could use to pass these details "up".  But I misremembered.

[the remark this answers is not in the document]
reda-alaoui: So, do you want me to proceed with the modifications or should I keep `AppliedGraphs` in its current state ? :)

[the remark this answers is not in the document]
sebersole: Is it fair to say I'd never use entity-graphs, so I would personally not care about the performance? ;)

If so, I'm fine with this as-is haha

sebersole:
This looks good to me.  You put it back in Draft status - was there something else you planned on doing here?

reda-alaoui:
> This looks good to me. You put it back in Draft status - was there something else you planned on doing here?

I was planning to fix https://github.com/hibernate/hibernate-orm/pull/6038#discussion_r1112058047 . But I am ok to merge it as is. I will come back with a follow-up pull request to improve `AppliedGraphs` performance.

sebersole:
Depends how you plan to address it.  Especially, I would not make the tree cyclic to handle this.

reda-alaoui:
> Especially, I would not make the tree cyclic to handle this.
Ok noted. IMHO, I think it is better to fix this in another PR anyway.

arvgord:
With pagination and fetch I get 
`WARN: HHH90003004: firstResult/maxResults specified with collection fetch; applying in memory`
like in this [test](https://github.com/arvgord/hibernate-fix-in-memory-pagination/blob/5a1afb4999399530a54fdd9a48846ee786a20567/src/test/kotlin/IssueTest.kt#L22)
Can you split a query like in this [test](https://github.com/arvgord/hibernate-fix-in-memory-pagination/blob/5a1afb4999399530a54fdd9a48846ee786a20567/src/test/kotlin/IssueTest.kt#L47) if using fetch(entity graph) to improve performance?

reda-alaoui:
> With pagination and fetch I get `WARN: HHH90003004: firstResult/maxResults specified with collection fetch; applying in memory` like in this [test](https://github.com/arvgord/hibernate-fix-in-memory-pagination/blob/5a1afb4999399530a54fdd9a48846ee786a20567/src/test/kotlin/IssueTest.kt#L22) Can you split a query like in this [test](https://github.com/arvgord/hibernate-fix-in-memory-pagination/blob/5a1afb4999399530a54fdd9a48846ee786a20567/src/test/kotlin/IssueTest.kt#L47) if using fetch(entity graph) to improve performance?

This matches the behaviour preceding the regression. Your (feature?) request is out of scope IMO.

sebersole (approved):

PR #6043, opened 2023-01-27
HHH-15944 Joins on non-managed-type polymorphisms don't work

dreab8:
https://hibernate.atlassian.net/browse/HHH-15944

alternative solution to https://github.com/hibernate/hibernate-orm/pull/5999

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/hql/internal/QuerySplitter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/hql/internal/QuerySplitter.java
@@ -22,10 +22,31 @@ import org.hibernate.query.criteria.JpaSearchOrder;
 import org.hibernate.query.sqm.tree.SqmQuery;
 import org.hibernate.query.sqm.tree.cte.SqmCteTableColumn;
 import org.hibernate.query.sqm.tree.cte.SqmSearchClauseSpecification;
+import org.hibernate.query.sqm.tree.domain.SqmBagJoin;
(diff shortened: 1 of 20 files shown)


hibernate-core/src/main/java/org/hibernate/query/hql/internal/QuerySplitter.java
```diff
@@ -1037,6 +1381,27 @@ public SqmCreationOptions getCreationOptions() {
...
+			SqmFrom copy = sqmFromCopyMap.get( sqmFrom );
+			if ( copy != null ) {
+				return (X) copy;
+			}
+			else {
+				return (X) sqmFrom.accept( this );
+			}
+		}
+
+		private void visitJoins(SqmFrom<?, ?> join, SqmAttributeJoin copy) {
```
beikov, line 1394: Nice refactoring. I think you should be able to use that for all joins though i.e. `visitCrossJoin` etc.

Also, what about fetches and treats? Do we not support that in polymorphic queries? Might be worth adding an assert statement here then at least to make this clear.

beikov (commented):
Looks like a nice first step!

hibernate-core/src/main/java/org/hibernate/query/hql/internal/QuerySplitter.java
dreab8: done, fI added a test for fetches and they seems working, for treat I added an exception

beikov (approved):
LGTM

PR #6045, opened 2023-01-27
HHH-15822 Unexpected org.hibernate.UnknownEntityTypeException: Unable to locate persister

mbladel:
Backport of https://github.com/hibernate/hibernate-orm/pull/5748.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -58,7 +58,7 @@ dependencyResolutionManagement {
     versionCatalogs {
         libs {
             version( "antlr", "4.10.1" )
-            version( "hcann", "6.0.2.Final" )
+            version( "hcann", "6.0.6.Final" )
             version( "geolatte", "1.8.2" )
             version( "byteBuddy", "1.12.18" )
             version( "agroal", "2.0" )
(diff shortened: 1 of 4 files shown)


settings.gradle
```diff
@@ -58,7 +58,7 @@ dependencyResolutionManagement {
     versionCatalogs {
         libs {
             version( "antlr", "4.10.1" )
-            version( "hcann", "6.0.2.Final" )
+            version( "hcann", "6.0.6.Final" )
```
beikov, line 61: Hmm, I think @yrodiere won't like that, which is why we were still on 6.0.2 on the 6.1 branch. Thoughts @yrodiere?

yrodiere: IIRC I did not like the last attempt to upgrade because you switched to a version of HCANN that had backwards-incompatible changes, which led to problems with Hibernate Search: we'd lose compatibility of Hibernate Search with whatever the next version of ORM is.

I don't remember if that version was a 6.0.something or a 6.1.something, though. If HCANN 6.0.x is free of backwards-incompatible changes, this *might* work. I'd have to test it, though... I'll have a look next week.

yrodiere: In particular if ORM does not rely on any new API/behavior of HCANN, it might just work...

yrodiere: @mbladel As far as I can tell you changes don't introduce a dependency on a new class/method in HCANN, so ORM will still work with HCANN 6.0.2 even after your change, it just will potentially be affected by HHH-15822.

I've just run the Hibernate Search test suite against ORM 6.1 with your patch and everything passes.

So, +1 from me for this upgrade. As long as we don't introduce reliance on new HCANN classes/methods in future commits :) But since 6.1 is reaching EOL, I'm guessing that's unlikely.

mbladel: Thanks for the feedback @yrodiere!

I'll let @beikov decide how to move forward since the solution which required the new HCANN version was actually implemented by him. Let me know if I can help in any way.

PR #6047, opened 2023-01-27
HHH-15443 Allow JdbcType to wrap read and write expressions

sebersole:
Rebasing of #5227


sebersole:
Applied upstream manually

PR #6052, opened 2023-01-31
HHH-15998 Incomplete group by expression when with "NEW" operator

mbladel:
https://hibernate.atlassian.net/browse/HHH-15991
https://hibernate.atlassian.net/browse/HHH-15998

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -2427,7 +2427,31 @@ public abstract class BaseSqmToSqlAstConverter<T extends Statement> extends Base
 			return true;
 		}
 		for ( SqmSelection<?> selection : selections ) {
-			if ( selection.getSelectableNode() == from ) {
+			if ( selectableNodeContains( selection.getSelectableNode(), from ) ) {
+				return true;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -2427,9 +2427,17 @@ private boolean selectClauseContains(SqmFrom<?, ?> from) {
...
+			if ( selectableNode == from ) {
 				return true;
 			}
+			else if ( selectableNode instanceof SqmDynamicInstantiation ) {
+				for ( SqmDynamicInstantiationArgument<?> argument : ( (SqmDynamicInstantiation<?>) selectableNode ).getArguments() ) {
+					if ( argument.getSelectableNode() == from ) {
+						return true;
+					}
+				}
+			}
```
beikov, line 2440: Dynamic instantiations can be nested, so this should be a recursive check.

mbladel: Updated the PR, now `selectClauseContains` checks dynamic instantiations recursively.

I also added similar logic for `order by` expressions, which have to be expanded if they were also expanded in the `group by`, and a test for that.

beikov (approved):

PR #6054, opened 2023-01-31
HHH-16117 Querying entity with collection in Embeddable causes 'A collection with cascade="all-delete-orphan" was no longer referenced by the owning entity instance'

dreab8:
https://hibernate.atlassian.net/browse/HHH-16117


beikov (approved):
LGTM

PR #6056, opened 2023-01-31
HHH-15665 - Fix and added test for issue

jrenaat:

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/SequenceInformationExtractorMariaDBDatabaseImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/SequenceInformationExtractorMariaDBDatabaseImpl.java
@@ -12,6 +12,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
+import org.hibernate.boot.model.naming.Identifier;
 import org.hibernate.boot.model.relational.QualifiedSequenceName;
 import org.hibernate.engine.jdbc.env.spi.IdentifierHelper;
 import org.hibernate.tool.schema.extract.spi.ExtractionContext;
@@ -28,7 +29,7 @@ public class SequenceInformationExtractorMariaDBDatabaseImpl extends SequenceInf
 
 	// SQL to get metadata from individual sequence
 	private static final String SQL_SEQUENCE_QUERY =
-			"SELECT '%1$s' as sequence_name, minimum_value, maximum_value, start_value, increment, cache_size FROM %1$s ";
+			"SELECT '%1$s' as sequence_name, minimum_value, maximum_value, start_value, increment, cache_size FROM %2$s ";
 
 	private static final String UNION_ALL =
 			"UNION ALL ";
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/SequenceInformationExtractorMariaDBDatabaseImpl.java
```diff
@@ -56,7 +57,7 @@ public Iterable<SequenceInformation> extractMetadata(ExtractionContext extractio
 				if ( sequenceInfoQueryBuilder.length() > 0 ) {
 					sequenceInfoQueryBuilder.append( UNION_ALL );
 				}
-				sequenceInfoQueryBuilder.append( String.format( SQL_SEQUENCE_QUERY, sequenceName ) );
+				sequenceInfoQueryBuilder.append( String.format( SQL_SEQUENCE_QUERY, sequenceName, Identifier.toIdentifier( sequenceName ) ) );
 			}
```
dreab8, line 61: what about keeping `%1Ss` in `SQL_SEQUENCE_QUERY ` and do `String.format( SQL_SEQUENCE_QUERY,Identifier.toIdentifier( sequenceName ) `;?

beikov: If we did that, then we would potentially read a sequence name `` `seq-1` `` i.e. with the quoting part of the string, which is not correct.

dreab8: so when the sequence is quoted we extract a not quoted name?

beikov: Look at the SQL statement `SELECT '%1$s' as sequence_name, minimum_value, maximum_value, start_value, increment, cache_size FROM %2$s`, the first message format parameter is within a SQL string literal, within which we don't need quoting. We only need quoting for when we render the name as identifier.

beikov (approved):

hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/SequenceInformationExtractorMariaDBDatabaseImpl.java
dreab8: sorry I missed the `'%1$s` so rephrase my question cannot be `SELECT %1$s as sequence_name, minimum_value, maximum_value, start_value, increment, cache_size FROM %1$s`
and then do `String.format( SQL_SEQUENCE_QUERY,Identifier.toIdentifier( sequenceName ) ;` ?

dreab8: ok  it seems that my proposal is not working :(

beikov: No, because quoting with `` ` `` implies that the quoted thing is an identifier. We want to select a string though.

jrenaat: TBH I also don't see what the advantage of the alternative approach is/would be ?

PR #6061, opened 2023-02-02
HHH-15970 Issue when combining Postgres driver, @MappedSuperclass, generics and enums

dreab8:
https://hibernate.atlassian.net/browse/HHH-15970
https://hibernate.atlassian.net/browse/HHH-16075 seems that has been resolved

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/mapping/Column.java
+++ b/hibernate-core/src/main/java/org/hibernate/mapping/Column.java
@@ -9,10 +9,13 @@ package org.hibernate.mapping;
 import java.io.Serializable;
 import java.sql.Types;
 import java.util.Locale;
+import java.util.Objects;
 
-import org.hibernate.HibernateException;
+import org.hibernate.AssertionFailure;
 import org.hibernate.MappingException;
+import org.hibernate.boot.Metadata;
 import org.hibernate.boot.model.TruthValue;
+import org.hibernate.boot.model.relational.Database;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/mapping/Column.java
```diff
@@ -60,6 +66,7 @@ public class Column implements Selectable, Serializable, Cloneable, ColumnTypeIn
 	private String customWrite;
 	private String customRead;
 	private Size columnSize;
+	private String specializedTypeDeclaration;
```
beikov, line 69: I guess this can be removed?

hibernate-core/src/main/java/org/hibernate/type/descriptor/jdbc/JdbcType.java
```diff
@@ -64,9 +76,10 @@ default String getFriendlyName() {
 	 * to obtain the SQL column type.
 	 *
 	 * @return a JDBC type code
+	 * @since 6.2
 	 */
-	default int getDefaultSqlTypeCode() {
-		return getJdbcTypeCode();
+	default int getDdlTypeCode() {
```
beikov, line 81: I think the changes in this class should be reverted. The use of `getDdlTypeCode()` in `Column` should be substituted with `getDefaultSqlTypeCode()` for the 6.1 branch

beikov (approved):

PR #6063, opened 2023-02-02
HHH-16003 Wrong alias used in subquery generated from JPQL

mbladel:
https://hibernate.atlassian.net/browse/HHH-16003

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -3351,7 +3351,7 @@ public abstract class BaseSqmToSqlAstConverter<T extends Statement> extends Base
 		if ( sqmPath instanceof SqmEntityValuedSimplePath<?>
 				|| sqmPath instanceof SqmEmbeddedValuedSimplePath<?>
 				|| sqmPath instanceof SqmAnyValuedSimplePath<?> ) {
-			final TableGroup existingTableGroup = fromClauseIndex.findTableGroup( sqmPath.getNavigablePath() );
+			final TableGroup existingTableGroup = fromClauseIndex.findTableGroupForGetOrCreate( sqmPath.getNavigablePath() );
 			if ( existingTableGroup == null ) {
 				final TableGroup createdTableGroup = createTableGroup(
 						fromClauseIndex.getTableGroup( sqmPath.getLhs().getNavigablePath() ),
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3443,6 +3438,28 @@ else if ( parentPath instanceof SqmFrom<?, ?> ) {
...
+		if ( existingTableGroup != null && path instanceof SqmEmbeddedValuedSimplePath ) {
+			// When an embedded property is referenced by the same navigable paths but different roots (e.g. same table and same alias in a sub-query)
+			// we need to check if the parent's primary table reference is the correct one, and not one taken from the parent from clause access
+			final TableGroup parentTableGroup = fromClauseIndex.findTableGroup( getParentPath( path ).getNavigablePath() );
+			if ( parentTableGroup != null && parentTableGroup.getPrimaryTableReference() != existingTableGroup.getPrimaryTableReference() ) {
+				return null;
+			}
+		}
+		return existingTableGroup;
+	}
```
beikov, line 3452: Accessing the primary table reference potentially initializes a `LazyTableGroup`, so I'm not sure this check is such a good idea. Overall, I think that we should introduce a new method in `FromClauseAccess` and with a special implementation for `SimpleFromClauseAccessImpl`, because potentially every use which wants to create table groups if none can be found might run into this issue. I found that `FromClauseAccess#resolveTableGroup`, `BaseSqmToSqlAstConverter#prepareForSelection` and both `BaseSqmToSqlAstConverter#prepareReusablePath` should use this new method `findTableGroupForGetOrCreate` (I'm open for a different name). The `findTableGroupForGetOrCreate` method can then nicely implement the check that the parent table group must match the one which can be found by looking up the path. Something like this:

```java
	@Override
	public TableGroup findTableGroupForGetOrCreate(NavigablePath navigablePath) {
		final TableGroup tableGroup = findTableGroup( navigablePath );
		if ( tableGroup == null || !( tableGroup.getModelPart() instanceof EmbeddableValuedModelPart ) ) {
			return tableGroup;
		}
		// Traverse up the navigable path to the point where resolving the path leads us to a regular TableGroup
		NavigableRole parentRole = tableGroup.getModelPart().getNavigableRole().getParent();
		NavigablePath parentPath = navigablePath.getParent();
		while ( parentRole.getParent() != null ) {
			parentRole = parentRole.getParent();
			parentPath = parentPath.getParent();
		}
		return findTableGroup( parentPath ) == ((VirtualTableGroup) tableGroup).getUnderlyingTableGroup();
	}
```

while also adding `getUnderlyingTableGroup` to `VirtualTableGroup`

mbladel: Thanks for the suggestion, sorry I missed the lazy initialization problem with my solution.

Exposing `underlyingTableGroup` should do the trick. I'll do some tests and update the PR soon.

beikov (approved):

PR #6071, opened 2023-02-03
HHH-16112 Using hibernate.jdbc.batch_size configuration causes BatchedTooManyRowsAffectedException for OneToMany relationship with orphan removal

mbladel:
https://hibernate.atlassian.net/browse/HHH-16112

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/collection/OneToManyPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/collection/OneToManyPersister.java
@@ -25,6 +25,7 @@ import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.internal.FilterAliasGenerator;
 import org.hibernate.internal.util.NullnessHelper;
 import org.hibernate.internal.util.collections.ArrayHelper;
+import org.hibernate.jdbc.Expectations;
 import org.hibernate.mapping.Collection;
 import org.hibernate.metamodel.mapping.CollectionPart;
 import org.hibernate.metamodel.mapping.EntityIdentifierMapping;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/persister/collection/OneToManyPersister.java
```diff
@@ -390,7 +391,8 @@ public RestrictedTableMutation<JdbcMutationOperation> generateDeleteAllAst(Mutat
 				keyRestrictionBindings,
 				null,
 				parameterBinders,
-				sqlWhereString
+				sqlWhereString,
+				Expectations.NONE
```
beikov, line 395: Shouldn't the fix rather be in `org.hibernate.persister.collection.AbstractCollectionPersister#buildCollectionTableMapping` where we create the `MutationDetails` with a certain `Expectation` for the delete all part?

mbladel: We would have to change not the *delete all* part `MutationDetails` but the update one, since the `org.hibernate.persister.collection.OneToManyPersister#generateDeleteAllAst` method creates an update operation for the one-shot delete.

Since we have a `@JoinColumn`, the related table is the child entity one and not a dedicated relationship table. It would work, but then all updates on the related entity's table would have no Expectation, so I though creating a dedicated case just for the one-shot update would be better.

beikov: Hmm, maybe @sebersole has some thoughts on that.

sebersole: `MutationDetails` and command creation get a bit odd when it comes to collections because of this "some deletes are updates" situation.  To be honest, I'm surprised this has not caused more problems.

I think the solution here is fine.

sebersole (commented):
lgtm

PR #6074, opened 2023-02-04
HHH-15707 - Fix Gradle plugin with Kotlin 1.7.0 or higher

tomas-c:
Since Kotlin version 1.7.0 the KotlinCompile task no longer extends Gradle's AbstractCompile.

This commit updates Hibernate Gradle enhancement plugin to not cast to AbstractCompile and instead use reflection to invoke the "getDestinationDirectory" method.

It also updates the Kotlin version on used to test the Gradle enhancement (but remains backwards compatible with previous Kotlin versions).

### DIFF
--- a/tooling/hibernate-gradle-plugin/src/main/java/org/hibernate/orm/tooling/gradle/HibernateOrmPlugin.java
+++ b/tooling/hibernate-gradle-plugin/src/main/java/org/hibernate/orm/tooling/gradle/HibernateOrmPlugin.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.orm.tooling.gradle;
 
+import java.lang.reflect.Method;
 import java.util.Set;
 
 import org.gradle.api.Action;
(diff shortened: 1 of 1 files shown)


beikov:
Thanks for the fix!

sebersole:
Hello Tomas Cerskus (@tomas-c),
I hope you’re well.

I'm going to need your help with a copyright matter: we’re exploring the possibility of changing the license of all Hibernate projects to Apache Software License v2 (ASL2).

You authored various contributions to our projects, some of which are currently licensed as LGPL2.1; I would love to have your explicit permission so that we might proceed with this change without further delays.

Could you reply and state that you agree to make all your contributions available under the ASL2?

The terms of the ASL2 can be found here: https://apache.org/licenses/LICENSE-2.0.

Also, if you have objections, please let me know. The Hibernate team believes that the ASL2 license would be more suitable for the future of our projects as it’s become more common and widely accepted; we think that in spirit and intent, it’s not a significant change, but would resolve some concerns that users have occasionally raised.

Please note that we’re not sure yet if we’ll be able to do this; I merely hope we can make it happen.

Regards,
Steve, on behalf of the Hibernate team

tomas-c:
Hi Steve,

I agree to make all my past contributions to this project available under
the ASL2.

Thank you for asking!

-Tomas

PR #6088, opened 2023-02-07
HHH-16123 Invalid generated SQL query when accessing join from a treated abstract superclass path

mbladel:
https://hibernate.atlassian.net/browse/HHH-16123


beikov:
Thanks for the test and the inspiration. I added another test which uncovered a problem with the fix, so I took a deeper look and came up with something different. Thanks for your pre-work :)
Superseded by https://github.com/hibernate/hibernate-orm/pull/6097

PR #6111, opened 2023-02-10
HHH-16169 - Fix potential NullPointerException in CollectionEntry

jrenaat:

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/spi/CollectionEntry.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/spi/CollectionEntry.java
@@ -375,7 +375,7 @@ public final class CollectionEntry implements Serializable {
 	@Override
 	public String toString() {
 		String result = "CollectionEntry" +
-				MessageHelper.collectionInfoString( loadedPersister.getRole(), loadedKey );
+				MessageHelper.collectionInfoString( role, loadedKey );
 		if ( currentPersister != null ) {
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/engine/spi/CollectionEntry.java
```diff
@@ -375,7 +375,7 @@ public void setRole(String role) {
 	@Override
 	public String toString() {
 		String result = "CollectionEntry" +
-				MessageHelper.collectionInfoString( loadedPersister.getRole(), loadedKey );
+				MessageHelper.collectionInfoString( loadedPersister != null ? loadedPersister.getRole() : null, loadedKey );
```
beikov, line 378:
```suggestion
				MessageHelper.collectionInfoString( loadedPersister != null ? loadedPersister.getRole() : role, loadedKey );
```

jrenaat: Does that make sense? From what I can tell, if the persister is null, the role is going to be null as well ...

beikov: It's actually safer to refer to `role` directly instead of `loadedPersister.getRole()`

```suggestion
				MessageHelper.collectionInfoString( role, loadedKey );
```

PR #6112, opened 2023-02-10
experimental support for 'extract(epoch from ...)' in HQL

gavinking:
Let's just see how many dbs this works on.


beikov:
Seems to work. By the way, there is a bug report open for this. Could you please change the commit message to include the issue number? https://hibernate.atlassian.net/browse/HHH-16082

beikov (approved):

PR #6115, opened 2023-02-13
HHH-16170 Literal enum values do not work in HQL 'set' clause.

mbladel:
https://hibernate.atlassian.net/browse/HHH-16170


gavinking:
Awesome, thanks!

PR #6118, opened 2023-02-13
HHH-16177 Catalog/schema creation/dropping sometimes ignores default catalog/schema

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16177

Backported to branch 6.1: see #6119.
Backported to branch 5.6: see #6121.


beikov (approved):

beikov:
The failure on Sybase seems real though.

yrodiere:
> The failure on Sybase seems real though.

Right, same problem as SQL Server apparently: `DatabaseMetaData#getSchema`/`DatabaseMetaData#getTables` throws an exception when the catalog does not exist. I wonder if I shouldn't open another ticket about that, because it seems to me we shouldn't even try to check whether a schema/table exists if the catalog does not exist...

PR #6129, opened 2023-02-14
HHH-16184 Two queries are execute to initialize empty collections   

dreab8:
https://hibernate.atlassian.net/browse/HHH-16184

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/internal/ScrollableResultsImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/internal/ScrollableResultsImpl.java
@@ -7,6 +7,7 @@
 package org.hibernate.internal;
 
 import org.hibernate.HibernateException;
+import org.hibernate.engine.spi.PersistenceContext;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.sql.results.internal.RowProcessingStateStandardImpl;
 import org.hibernate.sql.results.jdbc.internal.JdbcValuesSourceProcessingStateStandardImpl;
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/spi/ListResultsConsumer.java
```diff
@@ -153,6 +153,7 @@ public List<R> consume(
 		final QueryOptions queryOptions = rowProcessingState.getQueryOptions();
 		RuntimeException ex = null;
 		try {
+			persistenceContext.beforeLoad();
```
beikov, line 156: In 5.6 we also did that for scrollable results it seems. IMO we should add that there as well.
It might be best to do it in the same place as 5.6, which would roughly be in `org.hibernate.sql.exec.internal.JdbcSelectExecutorStandardImpl#executeQuery` and `org.hibernate.sql.exec.internal.JdbcSelectExecutorStandardImpl#executeQueryScroll` around the `doExecuteQuery` calls.

dreab8: not sure about putting the `beforeLoad()` in `org.hibernate.sql.exec.internal.JdbcSelectExecutorStandardImpl#executeQuery`. I put inside the ListResultConsumer so we are sure we will always call the `persistenceContext.afterLoad();`

Working on the scroll

hibernate-core/src/main/java/org/hibernate/internal/ScrollableResultsImpl.java
```diff
@@ -122,15 +123,23 @@ private void prepareCurrentRow(boolean underlyingScrollSuccessful) {
...
-		getRowProcessingState().finishRowProcessing();
-		getJdbcValuesSourceProcessingState().finishUp();
+			getRowProcessingState().finishRowProcessing();
+			getJdbcValuesSourceProcessingState().finishUp();
 
-		getRowProcessingState().getSession().getPersistenceContext().initializeNonLazyCollections();
+			getRowProcessingState().getSession().getPersistenceContext().initializeNonLazyCollections();
+		}
+		finally {
+			persistenceContext.afterLoad();
```
beikov, line 141: I think this must happen before calling `initializeNonLazyCollections`, otherwise the loadCounter will always be `!= 0` and we will never load lazy collections.

dreab8: good catch :+1:

beikov (approved):

PR #6130, opened 2023-02-14
HHH-16130 Make date_trunc function accessible via CriteriaApi

mbladel:
https://hibernate.atlassian.net/browse/HHH-16130

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/criteria/HibernateCriteriaBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/criteria/HibernateCriteriaBuilder.java
@@ -998,6 +998,18 @@ public interface HibernateCriteriaBuilder extends CriteriaBuilder {
 	@Incubating
 	JpaFunction<Float> second(Expression<? extends TemporalAccessor> datetime);
 
+	/**
+	 * Truncates a date, time or datetime expression to the given {@link TemporalUnit}.
+	 * Supported units are: {@code YEAR}, {@code MONTH}, {@code DAY},  {@code HOUR}, {@code MINUTE}, {@code SECOND}.
+	 * <p>
+	 * Truncating translates to obtaining a value of the same type in which all temporal units smaller than {@code field} have been pruned.
+	 * For hours, minutes and second this means setting them to {@code 00}. For months and days, this means setting them to {@code 01}.
+	 *
+	 * @param datetime the date, time or datetime expression to be truncated
+	 * @param temporalUnit the temporal unit for truncation
+	 *
+	 * @return the truncated value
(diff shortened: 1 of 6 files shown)


gavinking:
Wait wait wait, I'm -1 on this one.

- `date_trunc` is _not_ on the list of portable HQL functions
- the name `date_trunc` doesn't follow our naming conventions, and
- AFAIK, it's only supported on four databases.

So I don't think we should add it to the criteria API.

mbladel:
I thought it would be nice giving the users the possibility of accessing it through Criteria, since there is no other way of accessing this function (not even the `function()` method can be used, which was a valid workaround in Hibernate 5).
If / when JPA standardizes the concept of a `TemporalUnit` we could consider removing this in favor of that.

If the policy is not to include non-portable HQL functions in Criteria I understand and we can close this.

gavinking:
> If the policy is not to include non-portable HQL functions in Criteria I understand and we can close this.

I think that's pretty much _got_ to be the policy, because without it, where does this end? There's all sorts of nice but unportable SQL functions we could add to `Criteria`.

sebersole:
> not even the function() method can be used, which was a valid workaround in Hibernate 5

Users can register that.  Or even the Dialects

beikov:
How about we emulate truncation (and hence make it portable) by doing something like pseudo-HQL:
```sql
a.ts - mod(a.ts by UNIT, UNIT_RANGE)
```

mbladel:
Updated this since https://github.com/hibernate/hibernate-orm/pull/6134 was merged.

hibernate-core/src/main/java/org/hibernate/query/criteria/HibernateCriteriaBuilder.java
```diff
@@ -997,6 +998,9 @@ <Y extends Comparable<? super Y>> JpaPredicate between(
 	@Incubating
 	JpaFunction<Float> second(Expression<? extends TemporalAccessor> datetime);
 
+	@Incubating
```
beikov, line 1001: Java doc would be nice

beikov (approved):

hibernate-core/src/main/java/org/hibernate/query/criteria/HibernateCriteriaBuilder.java
mbladel: Thanks, added it

beikov:
Any further reservations about adding this, now that we have a portable function implementation @gavinking?

gavinking:
Well currently it's still called `dateTrunc()` instead of  wing an overload of `truncate()`.

PR #6132, opened 2023-02-14
HHH-16175  AssertionError in StandardEntityGraphTraversalStateImpl.traverse when using entity graph

dreab8:
https://hibernate.atlassian.net/browse/HHH-16175


yrodiere (approved):
Well that was simpler than expected. Works for me :)

beikov (approved):

jrenaat:
The test seems to fail on sybase; when i run it locally i get the error "java.sql.SQLException: The "default data cache (id: 0)" is configured with 410 buffers.  The current query plan requires 648 buffers.  Please reconfigure the data cache and try the command again." I can only assume that this is similar to the failure on GH actions.
I tried adding `sp_cacheconfig default_data_cache, "16M"`to the docker_db script but for some reason i couldn't get that to transfer into my sybase instance, where i still get 8Mb
Another option is to just disable this test for sybase?

dreab8:
@jrenaat https://github.com/hibernate/hibernate-orm/pull/6147 should fix the failure

PR #6133, opened 2023-02-14
HHH-16082 - Correction to the epoch extraction for DB2

jrenaat:


gavinking:
If we’re going to go to the effort of having special code for older versions, why not just emulate `extract(epoch)` using the difference between timestamps, like we do with other databases?

PR #6134, opened 2023-02-15
HHH-16185 Standardize the date_trunc() function, emulating it for Dialects that don't support native versions

mbladel:
https://hibernate.atlassian.net/browse/HHH-16185


gavinking:
So the idea is format it as a string and then parse the string?

mbladel:
> So the idea is format it as a string and then parse the string?

Yep, when no native way of truncating exists.

gavinking:
Sounds good. (And useful.) But we need to figure out what we're going to call this function, since `date_trunc` doesn't conform to our conventions.

mbladel:
> Sounds good. (And useful.) But we need to figure out what we're going to call this function, since `date_trunc` doesn't conform to our conventions.

Sure, it should be easy enough to change the name. The convention is all lower case and no underscores, correct?

So maybe `datetrunc`?

gavinking:
> So maybe `datetrunc`?

Is it only for dates? Isn't it also for timestamps?

mbladel:
> Is it only for dates? Isn't it also for timestamps?

Yes it is. It should support all `TEMPORAL` types.

gavinking:
So the _natural_ thing to write here would be something like:

    truncate(datetime, second)

or even:

    truncate(datetime to second)

but of course this would mean overloading the existing `truncate` function. Not sure how I feel about that.

The reason I suggest the second option is that `to` is already a keyword, and we already support the ANSI syntax `extract(second from datetime)`.

sebersole:
I think the form `truncate(datetime, second)` is reasonable; it is consistent with the current 2-arg format.  The form using `to` changes the expected signature structure.  The overloading it natural imo.

mbladel:
While trying to rename the function to override the numeric `truncate` I noticed we might have a problem with the `truncate(expression, temporal_unit)` syntax.

If we have to disambiguate between the numeric and the datetime internal implementation, how can we know if the second parameter is being used as a `temporal_unit` (second, hour, minute) or if it's a path expression for the second parameter of the numeric `trunc` function?

The syntax `truncate(expression to temporal_unit)` could let us know for sure which type of function the user is calling. Otherwise a dedicated function name for datetime truncation would work as well.

gavinking:
Yeah we don’t currently have any infrastructure to support function overloading. We would have to build that.

sebersole:
So first, after looking at `TemporalUnit`, we actually have values there that are not valid for truncate (`WEEK_OF_MONTH`, `WEEK_OF_YEAR`,...).  That enum was developed for `extract`, so this is fine-ish.  In retrospect, I'd say these are more "temporal extraction units", but that's a minor naming concern.  But it is important to understand in relation to this discussion.  Here we want to allow truncation to actual temporal units, though honestly some "real" probably don't make a lot of practical sense - NANOSECOND, SECOND and even YEAR.  With that said...

> If we have to disambiguate between the numeric and the datetime internal implementation, how can we know if the second parameter is being used as a `temporal_unit` (second, hour, minute) or if it's a path expression for the second parameter of the numeric `trunc` function?

Given that we'd have a hard-coded list of values that we handle as a temporal-unit here, I'd just recognize them specifically - either 

- as keywords, if these are defined that way already, 
- as "soft" or "contextual" keywords

Also, FWIW, I'm not sure a path expression for the "scale" argument is very useful.  We could chose to limit the "scale" expression to just literals and parameters.  That would help with the "is it  a path" aspect - it would never be.


> The syntax `truncate(expression to temporal_unit)` could let us know for sure which type of function the user is calling. Otherwise a dedicated function name for datetime truncation would work as well.

If we end up having to go this route I am not a fan of simply reusing the name "truncate".  I think this suggestion goes well beyond "overloading" as we all understand it.

gavinking:
> Given that we'd have a hard-coded list of values that we handle as a temporal-unit here, I'd just recognize them specifically

Yes, that's the right way.

mbladel:
FYI updated the PR with:
* custom implementation for `trunc()` which handles both the numeric and datetime variant
* new parser rule for "contextual" temporal units
* updated UG with the new `trunc()` function for datetimes

beikov (changes requested):
Great work. I left a few comments, but overall, this looks nice!

hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
```diff
@@ -356,8 +350,8 @@ public void truncate() {
 	public void truncate_round() {
```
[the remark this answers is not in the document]
mbladel: This was not used, removed the method from function factory.

hibernate-core/src/main/java/org/hibernate/dialect/function/CommonFunctionFactory.java
```diff
@@ -259,93 +259,87 @@ public void moreHyperbolic() {
...
-				"sign(?1)*floor(abs(?1))",
-				"sign(?1)*floor(abs(?1)*1e?2)/1e?2",
-				NUMERIC, INTEGER,
-				typeConfiguration
-		).setArgumentListSignature( "(NUMERIC number[, INTEGER places])" );
-		functionRegistry.registerAlternateKey( "truncate", "trunc" );
+		trunc( "sign(?1)*floor(abs(?1))", "sign(?1)*floor(abs(?1)*1e?2)/1e?2", null, null );
 	}
 
 	public void truncate() {
```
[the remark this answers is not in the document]
mbladel: This was used in a couple community dialects, I now substituted it with `trunc_truncate()` and removed the unneeded method from the factory.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
```diff
@@ -512,13 +512,17 @@ public void testRoundTruncFunctions(SessionFactoryScope scope) {
...
 	public void testDateTruncFunction(SessionFactoryScope scope) {
 		scope.inTransaction(
-				session -> session.createQuery("select date_trunc(year,current_timestamp)", Timestamp.class).getSingleResult()
+				session -> {
+					session.createQuery( "select trunc(current_timestamp,year)", Timestamp.class ).getSingleResult();
+					session.createQuery( "select trunc(current_timestamp,month)", Timestamp.class ).getSingleResult();
+					session.createQuery( "select trunc(current_timestamp,day)", Timestamp.class ).getSingleResult();
+					session.createQuery( "select truncate(current_timestamp,hour)", Timestamp.class ).getSingleResult();
+					session.createQuery( "select truncate(current_timestamp,minute)", Timestamp.class ).getSingleResult();
+					session.createQuery( "select truncate(current_timestamp,second)", Timestamp.class ).getSingleResult();
```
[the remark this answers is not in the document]
mbladel: Yes, thanks to the `(TRUNC | TRUNCATE)` parser rule. It should also be tested by `org.hibernate.orm.test.query.hql.FunctionTests#testRoundTruncFunctions`.

hibernate-core/src/main/java/org/hibernate/dialect/function/TruncFunction.java
```diff
@@ -0,0 +1,232 @@
...
+			}
+			if ( datetimeTrunc.getPattern() == null ) {
+				final boolean useConvertToFormat = nodeBuilder.getSessionFactory()
+						.getJdbcServices()
+						.getDialect() instanceof SybaseDialect;
+				return new DateTruncEmulation(
+						toDateFunction,
+						useConvertToFormat,
+						typeConfiguration
+				).generateSqmFunctionExpression( arguments, impliedResultType, queryEngine, typeConfiguration );
```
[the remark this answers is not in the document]
mbladel: You're right, somehow I thought I saw some dialect-checking logic in custom functions but I was wrong.

I've now created dedicated functions to handle dialect-specific cases (Sybase, Oracle, PostgreSQL and SQLServer < 16).

hibernate-core/src/main/java/org/hibernate/dialect/function/DateTruncEmulation.java
```diff
@@ -0,0 +1,195 @@
...
+			// see: https://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.infocenter.dc36271.1600/doc/html/san1393050437990.html
+			sqlAppender.append( "datetime,substring(convert(varchar," );
+			sqlAstArguments.get( 0 ).accept( walker );
+			sqlAppender.append( ",21),1,17-len(" );
+			sqlAstArguments.get( 1 ).accept( walker );
+			sqlAppender.append( "))+" );
+			sqlAstArguments.get( 1  ).accept( walker );
+			sqlAppender.append( ",21" );
+		}
+		sqlAppender.append( ')' );
```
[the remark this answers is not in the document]
mbladel: Created additional `DateTruncConvertEmulation` and `SybaseDateTruncEmulation` that extend the base emulation function for specific cases. `DateTruncEmulation` is still used to handle "regular" emulation.

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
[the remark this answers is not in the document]
beikov: Ok, but using the `truncate` function in e.g. Criteria wouldn't work this way anymore. Can you please add a test for that and re-add the alternate keys?

hibernate-core/src/main/java/org/hibernate/dialect/function/TruncFunction.java
```diff
@@ -0,0 +1,211 @@
...
+ * @author Marco Belladelli
+ */
+public class TruncFunction extends AbstractSqmFunctionDescriptor implements FunctionRenderingSupport {
+	private final PatternRenderer truncPattern;
+	private final PatternRenderer twoArgTruncPattern;
+	protected final DatetimeTrunc datetimeTrunc;
+	private final PatternRenderer datetimeTruncPattern;
+	private final String toDateFunction;
+
+	protected boolean numericTrunc;
```
beikov, line 49: That's not going to work. The function descriptor is shared across everything, so having state here is not possible. You can instead create two implementations of `FunctionRenderingSupport` and use one or the other when instantiating `SelfRenderingSqmFunction`

hibernate-core/src/main/java/org/hibernate/dialect/function/SybaseTruncFunction.java
```diff
@@ -0,0 +1,65 @@
...
+	@Override
+	protected <T> SelfRenderingSqmFunction<T> generateSqmFunctionExpression(
+			List<? extends SqmTypedNode<?>> arguments,
+			ReturnableType<T> impliedResultType,
+			QueryEngine queryEngine,
+			TypeConfiguration typeConfiguration) {
+		final List<SqmTypedNode<?>> args = new ArrayList<>( arguments );
+		if ( arguments.size() == 2 && arguments.get( 1 ) instanceof SqmExtractUnit ) {
+			// datetime truncation
+			return new SybaseDateTruncEmulation( "convert", typeConfiguration ).generateSqmFunctionExpression(
```
beikov, line 43: Would be nice if you could keep the `SybaseDateTruncEmulation` in a private final variable initialized in the constructor.

hibernate-core/src/main/java/org/hibernate/dialect/function/SqlServerConvertTruncFunction.java
```diff
@@ -0,0 +1,66 @@
...
+	@Override
+	protected <T> SelfRenderingSqmFunction<T> generateSqmFunctionExpression(
+			List<? extends SqmTypedNode<?>> arguments,
+			ReturnableType<T> impliedResultType,
+			QueryEngine queryEngine,
+			TypeConfiguration typeConfiguration) {
+		final List<SqmTypedNode<?>> args = new ArrayList<>( arguments );
+		if ( arguments.size() == 2 && arguments.get( 1 ) instanceof SqmExtractUnit ) {
+			// datetime truncation
+			return new DateTruncConvertEmulation( typeConfiguration ).generateSqmExpression(
```
beikov, line 43: Would be nice if you could keep the `DateTruncConvertEmulation` in a private final variable initialized in the constructor.

hibernate-core/src/main/java/org/hibernate/dialect/function/PostgreSQLTruncFunction.java
```diff
@@ -0,0 +1,69 @@
...
+			QueryEngine queryEngine,
+			TypeConfiguration typeConfiguration) {
+		final List<SqmTypedNode<?>> args = new ArrayList<>( arguments );
+		if ( arguments.size() == 2 && arguments.get( 1 ) instanceof SqmExtractUnit ) {
+			// datetime truncation
+			numericTrunc = false;
+		}
+		else {
+			// numeric truncation
+			return new PostgreSQLTruncRoundFunction( "trunc", supportsTwoArguments ).generateSqmFunctionExpression(
```
beikov, line 50: Would be nice if you could keep the `PostgreSQLTruncRoundFunction` in a private final variable initialized in the constructor.

hibernate-core/src/main/java/org/hibernate/dialect/function/OracleTruncFunction.java
```diff
@@ -0,0 +1,104 @@
...
+					break;
+				case HOUR:
+					pattern = "HH";
+					break;
+				case MINUTE:
+					pattern = "MI";
+					break;
+				case SECOND:
+					// Oracle does not support truncating to seconds with the native function, use emulation
+					return new DateTruncEmulation( "to_date", typeConfiguration )
```
beikov, line 72: Would be nice if you could keep the `DateTruncEmulation` in a private final variable initialized in the constructor.

hibernate-core/src/main/java/org/hibernate/dialect/function/DateTruncEmulation.java
```diff
@@ -0,0 +1,160 @@
...
+		final SqmFormat sqmFormat = new SqmFormat(
+				"yyyy-MM-dd HH:mm:ss",
+				typeConfiguration.getBasicTypeForJavaType( String.class ),
+				nodeBuilder
+		);
+
+		return new SelfRenderingSqmFunction<>(
+				this,
+				this,
+				asList( datetime, formattedDatetime, sqmFormat ),
```
beikov, line 152: Seems the first argument is only used in `SybaseDateTruncEmulation#render` method. Can you please add a comment here about this?

hibernate-core/src/main/java/org/hibernate/dialect/function/DateTruncEmulation.java
```diff
@@ -0,0 +1,160 @@
...
+							),
+							null,
+							queryEngine,
+							typeConfiguration
+					);
+		}
+		else {
+			formattedDatetime = formatExpression;
+		}
+		final SqmFormat sqmFormat = new SqmFormat(
```
beikov, line 143: Would be nice if you could keep the `SqmFormat` in a private final variable initialized in the constructor.

hibernate-core/src/main/java/org/hibernate/dialect/function/TruncFunction.java
```diff
@@ -0,0 +1,211 @@
...
+			// numeric truncation
+			numericTrunc = true;
+		}
+
+		return new SelfRenderingSqmFunction<>(
+				this,
+				this,
+				args,
+				impliedResultType,
+				getArgumentsValidator(),
```
beikov, line 178:
```suggestion
				numericTrunc ? TruncArgumentsValidator.NUMERIC_VALIDATOR : TruncArgumentsValidator.DATETIME_VALIDATOR,
```

hibernate-core/src/main/java/org/hibernate/dialect/function/DateTruncEmulation.java
mbladel: If I'm initializing `DateTruncEmulation` as a private final variable in the constructor of `TruncFunction` I don't think I can access the `NodeBuilder` to pass to `SqmFormat` in the constructor.

beikov: The node builder is actually not important here. You could pass null for it.

hibernate-core/src/main/java/org/hibernate/dialect/function/SybaseTruncFunction.java
mbladel: Should I also make dialect-specific datetime trunc emulation functions inner classes of the already dialect-specific trunc functions (e.g. `SybaseTruncFunction$SybaseDateTruncEmulation`)?

beikov: Don't know if I understand exactly, but whatever you think works best ;)

hibernate-core/src/test/java/org/hibernate/orm/test/query/hql/FunctionTests.java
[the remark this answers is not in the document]
mbladel: Oh my bad, added the test in `CriteriaBuilderNonStandardFunctionsTest`. Also applied all latest suggestions.

beikov (approved):
LGTM

ManjunathM22:
> https://hibernate.atlassian.net/browse/HHH-16185

HI ,,

I have been badly stuck up with - date_trunc function can please help me -- my code is as below

Line - 1

private Date slotSystem;
timeFormat - yyyy-mm-dd

to_char(date_trunc('%1$s', c.slotSystem), :timeFormat) as period, " 

Line - 2
"     datetrunc('%1$s', c.slotSystem) "

Exception ; 
org.postgresql.util.PSQLException: ERROR: function datetrunc(unknown, bigint) does not exist
  Hint: No function matches the given name and argument types. You might need to add explicit type casts.

gavinking:
@ManjunathM22 The function is now called `truncate()`, as documented here:

<https://docs.jboss.org/hibernate/orm/6.2/userguide/html_single/Hibernate_User_Guide.html#hql-functions-datetime>

PR #6144, opened 2023-02-20
 @Where clause ignored after upgrading to Hibernate 6

sebersole:
HHH-16019 - @Where clause ignored after upgrading to Hibernate 6


beikov (approved):

sebersole:
Rebased on top of HHH-16198 work (which this ultimately requires)

sebersole:
Applied upstream

PR #6145, opened 2023-02-20
Reproducer for HHH-16199 org.hibernate.AssertionFailure: force initializing collection loading

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16199


yrodiere:
This was fixed by #6129

PR #6146, opened 2023-02-20
HHH-16193 LazyInitializationException when accessing loaded reference after session is closed using bytecode enhancement

dreab8:
https://hibernate.atlassian.net/browse/HHH-16193

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
+++ b/hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
@@ -12,11 +12,13 @@ import org.hibernate.NonUniqueObjectException;
 import org.hibernate.PersistentObjectException;
 import org.hibernate.TypeMismatchException;
 import org.hibernate.action.internal.DelayedPostInsertIdentifier;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
```diff
@@ -530,7 +535,16 @@ private Object doLoad(LoadEvent event, EntityPersister persister, EntityKey keyT
 					= CacheEntityLoaderHelper.INSTANCE.loadFromSessionCache( event, keyToLoad, options );
 			final Object entity = persistenceContextEntry.getEntity();
 			if ( entity != null ) {
-				return persistenceContextEntry.isManaged() ? entity : null;
```
beikov, line 533: Should we also backport this to 5? As far as I understand, the expectation of this method is that the result is an initialized entity, right? Not sure if not initializing it in 5 might cause the same issues as in 6, but just wanted to throw this thought in for consideration.

beikov (approved):
LGTM

hibernate-core/src/main/java/org/hibernate/event/internal/DefaultLoadEventListener.java
dreab8: I think in 5 it is not necessary

PR #6148, opened 2023-02-21
Reproducer for HHH-16209 Identically-named association in entity root and embeddable leads to mixup during association loading

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16209


beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/6171

PR #6152, opened 2023-02-22
HHH-16188 HHH-16238 Generic embeddable properties and identifiers

mbladel:
https://hibernate.atlassian.net/browse/HHH-16188
https://hibernate.atlassian.net/browse/HHH-16238

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/mapping/Property.java
+++ b/hibernate-core/src/main/java/org/hibernate/mapping/Property.java
@@ -57,6 +57,7 @@ public class Property implements Serializable, MetaAttributable {
 	private java.util.Map metaAttributes;
 	private PersistentClass persistentClass;
 	private boolean naturalIdentifier;
+	private boolean genericEmbeddable;
 	private boolean lob;
 	private java.util.List<CallbackDefinition> callbackDefinitions;
 	private String returnedClassName;
@@ -417,6 +418,14 @@ public class Property implements Serializable, MetaAttributable {
 		this.naturalIdentifier = naturalIdentifier;
 	}
 
+	public boolean isGenericEmbeddable() {
+		return genericEmbeddable;
+	}
+
+	public void setGenericEmbeddable(boolean genericEmbeddable) {
+		this.genericEmbeddable = genericEmbeddable;
+	}
+
 	public boolean isLob() {
 		return lob;
 	}
@@ -473,6 +482,7 @@ public class Property implements Serializable, MetaAttributable {
 		property.setMetaAttributes( getMetaAttributes() );
 		property.setPersistentClass( getPersistentClass() );
 		property.setNaturalIdentifier( isNaturalIdentifier() );
+		property.setGeneric( isGeneric() );
 		property.setLob( isLob() );
 		property.addCallbackDefinitions( getCallbackDefinitions() );
 		property.setReturnedClassName( getReturnedClassName() );
--- a/hibernate-core/src/main/java/org/hibernate/mapping/Property.java
+++ b/hibernate-core/src/main/java/org/hibernate/mapping/Property.java
@@ -57,7 +57,7 @@ public class Property implements Serializable, MetaAttributable {
 	private java.util.Map metaAttributes;
 	private PersistentClass persistentClass;
 	private boolean naturalIdentifier;
-	private boolean genericEmbeddable;
+	private boolean isGeneric;
 	private boolean lob;
 	private java.util.List<CallbackDefinition> callbackDefinitions;
 	private String returnedClassName;
@@ -418,12 +418,12 @@ public class Property implements Serializable, MetaAttributable {
 		this.naturalIdentifier = naturalIdentifier;
 	}
 
-	public boolean isGenericEmbeddable() {
-		return genericEmbeddable;
+	public boolean isGeneric() {
+		return isGeneric;
 	}
 
-	public void setGenericEmbeddable(boolean genericEmbeddable) {
-		this.genericEmbeddable = genericEmbeddable;
+	public void setGeneric(boolean generic) {
+		this.isGeneric = generic;
 	}
 
 	public boolean isLob() {
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetadataContext.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetadataContext.java
@@ -420,20 +420,26 @@ public class MetadataContext {
 	private void applyIdMetadata(PersistentClass persistentClass, IdentifiableDomainType<?> identifiableType) {
 		if ( persistentClass.hasIdentifierProperty() ) {
 			final Property declaredIdentifierProperty = persistentClass.getDeclaredIdentifierProperty();
+			//noinspection rawtypes
+			final AttributeContainer attributeContainer = (AttributeContainer) identifiableType;
 			if ( declaredIdentifierProperty != null ) {
 				final SingularPersistentAttribute<?, Object> idAttribute = attributeFactory.buildIdAttribute(
 						identifiableType,
 						declaredIdentifierProperty
 				);
-
-				//noinspection unchecked rawtypes
-				( ( AttributeContainer) identifiableType ).getInFlightAccess().applyIdAttribute( idAttribute );
+				//noinspection unchecked
+				attributeContainer.getInFlightAccess().applyIdAttribute( idAttribute );
 			}
 			else if ( persistentClass.getIdentifier() instanceof Component
 					&& persistentClass.getIdentifierProperty() != getSuperclassIdentifier( persistentClass ) ) {
 				// If the identifier is a generic component, we have to call buildIdAttribute anyway,
 				// as this will create and register the EmbeddableType for the subtype
-				attributeFactory.buildIdAttribute( identifiableType, persistentClass.getIdentifierProperty() );
+				final SingularPersistentAttribute<?, Object> concreteEmbeddable = attributeFactory.buildIdAttribute(
+						identifiableType,
+						persistentClass.getIdentifierProperty()
+				);
+				//noinspection unchecked
+				attributeContainer.getInFlightAccess().addConcreteEmbeddableAttribute( concreteEmbeddable );
 			}
 		}
 		else {
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetadataContext.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetadataContext.java
@@ -24,7 +24,6 @@ import org.hibernate.internal.HEMLogging;
 import org.hibernate.internal.util.ReflectHelper;
 import org.hibernate.internal.util.collections.CollectionHelper;
 import org.hibernate.mapping.Component;
-import org.hibernate.mapping.KeyValue;
 import org.hibernate.mapping.MappedSuperclass;
 import org.hibernate.mapping.PersistentClass;
 import org.hibernate.mapping.Property;
@@ -274,6 +273,7 @@ public class MetadataContext {
 
 					applyIdMetadata( safeMapping, jpaMapping );
 					applyVersionAttribute( safeMapping, jpaMapping );
+					applyGenericEmbeddableProperties( safeMapping, jpaMapping );
 
 					for ( Property property : safeMapping.getDeclaredProperties() ) {
 						if ( property.getValue() == safeMapping.getIdentifierMapper() ) {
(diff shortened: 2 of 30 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetadataContext.java
```diff
@@ -421,19 +421,13 @@ private void applyIdMetadata(PersistentClass persistentClass, IdentifiableDomain
...
+				applyIdAttribute( identifiableType, declaredIdentifierProperty );
 			}
 			else if ( persistentClass.getIdentifier() instanceof Component
 					&& persistentClass.getIdentifierProperty() != getSuperclassIdentifier( persistentClass ) ) {
-				// If the identifier is a generic component, we have to call buildIdAttribute anyway,
-				// as this will create and register the EmbeddableType for the subtype
-				attributeFactory.buildIdAttribute( identifiableType, persistentClass.getIdentifierProperty() );
+				// If the identifier is a different component than the superclass, we have to apply the id
+				// attribute anyway, as this will create and register the EmbeddableType for the subtype
+				applyIdAttribute( identifiableType, persistentClass.getIdentifierProperty() );
```
beikov, line 430: Unfortunately, this is not strictly correct. With this change, the metamodel types now have attribute objects for `id` that are different. This is a problem for the static metamodel and simply not how JPA "ticks".

I just had a recent discussion with @Sanne about a very similar example. IMO if we want to support this, then we have to implement this in `EmbeddedSqmPathSource` and `SqmEmbeddedValuedSimplePath`, by capturing the type variable assignments on the path and lookup the correct `EmbeddedSqmPathSource` based on that when finding a sub path source.

mbladel: > by capturing the type variable assignments on the path and lookup the correct EmbeddedSqmPathSource based on that when finding a sub path source

I'd like to try that, but I guess the correct `EmbeddedSqmPathSource`s with type parameters must be somehow *registered* to be able to retrieve them later. Any pointers / example on how it could be done?

mbladel: I updated the PR with a new solution that registers *concrete* generic embeddable attributes for identifier properties and retrieves them when creating `SqmPath`s.

I have opened another issue [HHH-16238](https://hibernate.atlassian.net/browse/HHH-16238) which expands on this for the handling of multiple non-identifier generic embeddable properties.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/SqmEmbeddedValuedSimplePath.java
```diff
@@ -77,6 +79,30 @@ public SqmPath<?> resolvePathPart(
 		return sqmPath;
 	}
 
+	@Override
+	public SqmPath<?> get(String attributeName) {
+		final DomainType<?> domainType = getRoot( getLhs() ).getReferencedPathSource().getSqmPathType();
```
beikov, line 84: It seems to me that a `getRoot` here is not correct. Can you explain why you did that? Also, it would be nice if we could somehow capture the fact that a path source is generic to only do the `findGenericEmbeddableAttribute` if necessary.

hibernate-core/src/test/java/org/hibernate/orm/test/annotations/generics/EmbeddedIdGenericsSuperclassTest.java
```diff
@@ -0,0 +1,272 @@
...
+		} );
+	}
+
+	@Test
+	public void testCustomerCriteria(SessionFactoryScope scope) {
+		scope.inTransaction( session -> {
+			final CriteriaBuilder cb = session.getCriteriaBuilder();
+			final CriteriaQuery<Customer> query = cb.createQuery( Customer.class );
+			final Root<Customer> root = query.from( Customer.class );
+			query.select( root ).where( cb.equal( root.get( "id" ).get( "someDomainField" ), 1 ) );
```
beikov, line 81: Please add some assertions on the type of the `root.get("id")` path to ensure we don't leak the generic type somehow.

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/SqmEmbeddedValuedSimplePath.java
mbladel: At this point `getRoot` was needed because we would be resolving not the "first level" embeddable but one of its sub-properties.

I've reworked the logic to be handled by `EmbeddedSqmPathSource` to be able to retrieve the correct type for the "first level" embeddable attribute. Also I added the `generic` flag like you suggested.

hibernate-core/src/test/java/org/hibernate/orm/test/annotations/generics/EmbeddedIdGenericsSuperclassTest.java
mbladel: Done (thanks to new logic, see other comment).

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/EmbeddedSqmPathSource.java
```diff
@@ -46,11 +49,27 @@ public SqmPath<J> createSqmPath(SqmPath<?> lhs, SqmPathSource<?> intermediatePat
...
+
+		final DomainType<?> lhsType = lhs.getReferencedPathSource().getSqmPathType();
+		final SqmPathSource<J> concretePathSource;
+		if ( isGeneric && lhsType instanceof ManagedDomainType ) {
+			//noinspection unchecked rawtypes
+			concretePathSource = ( (ManagedDomainType) lhsType ).findConcreteEmbeddableAttribute( getPathName() );
+		}
+		else {
+			concretePathSource = null;
+		}
```
beikov, line 61:
```suggestion
		final DomainType<?> lhsType;
		final SqmPathSource<J> concretePathSource;
		if ( isGeneric && ( lhsType = lhs.getReferencedPathSource().getSqmPathType() ) instanceof ManagedDomainType ) {
			//noinspection unchecked rawtypes
			concretePathSource = ( (ManagedDomainType) lhsType ).findConcreteEmbeddableAttribute( getPathName() );
		}
		else {
			concretePathSource = this;
		}
```

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/EmbeddedSqmPathSource.java
```diff
@@ -46,11 +49,27 @@ public SqmPath<J> createSqmPath(SqmPath<?> lhs, SqmPathSource<?> intermediatePat
...
+
 		return new SqmEmbeddedValuedSimplePath<>(
 				navigablePath,
-				this,
+				concretePathSource != null ? concretePathSource : this,
```
beikov, line 65:
```suggestion
				concretePathSource,
```

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/EmbeddedSqmPathSource.java
```diff
@@ -18,6 +20,7 @@
 public class EmbeddedSqmPathSource<J>
 		extends AbstractSqmPathSource<J>
 		implements CompositeSqmPathSource<J> {
+	private boolean isGeneric;
```
beikov, line 23: Can you make that final please and pass that through `buildIdAttribute` -> `SingularAttributeImpl` -> `SqmMappingModelHelper#resolveSqmPathSource` -> `EmbeddedSqmPathSource` instead of using a setter?

hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/SqmEmbeddedValuedSimplePath.java
beikov: Moving the logic from `SqmEmbeddedValuedSimplePath` to `EmbeddedSqmPathSource` might cause some trouble with the assertions I asked you to add, because the `EmbeddedSqmPathSource` is what `SqmEmbeddedValuedSimplePath` exposes as "type" information AFAICT, but only some additional assertions will be able to ensure we don't expose any generic info through JPA contracts.

hibernate-core/src/test/java/org/hibernate/orm/test/annotations/generics/EmbeddedIdGenericsSuperclassTest.java
```diff
@@ -0,0 +1,279 @@
...
+
+	@Test
+	public void testCustomerCriteria(SessionFactoryScope scope) {
+		scope.inTransaction( session -> {
+			final CriteriaBuilder cb = session.getCriteriaBuilder();
+			final CriteriaQuery<Customer> query = cb.createQuery( Customer.class );
+			final Root<Customer> root = query.from( Customer.class );
+			final Path<CustomerId> id = root.get( "id" );
+			assertThat( id ).isNotNull();
+			assertThat( id.getJavaType() ).isEqualTo( CustomerId.class );
```
beikov, line 84: Not sure if this is "correct". IMO the exposed type here should rather still be `DomainEntityId`, hence my comment about doing the generic lookup in `SqmEmbeddedValuedSimplePath#get` instead to avoid this exposure.

beikov: Can you also please add an assertion for `id.getModel()`?

mbladel: Oh sorry, I interpreted *don't expose any generic info* as we should return the actual concrete type instead of the generic one, but I see now what you meant.

Then this is indeed not correct, I'm gonna move back the logic to `SqmEmbeddedValuedSimplePath` and fix the assertions.

beikov: Wait a second. I'm checking the spec now to be sure.

mbladel: Updated with:
* handling of generic embedded non-identifier properties
* review of `getModel()` logic for `SingularAttributeImpl`

The new logic required some impl for `getNodeType()` and `getReferencedPathSource()` (see `AbstractSqmSimplePath`). I also added the new `JpaPath#getResolvedModel` as suggested.

hibernate-core/src/main/java/org/hibernate/mapping/Property.java
```diff
@@ -57,6 +57,7 @@ public class Property implements Serializable, MetaAttributable {
 	private java.util.Map metaAttributes;
 	private PersistentClass persistentClass;
 	private boolean naturalIdentifier;
+	private boolean isGeneric;
```
beikov, line 60: Don't forget to also copy over this information in the `copy()` method.

hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetadataContext.java
```diff
@@ -567,6 +573,25 @@ private <X> void applyVersionAttribute(MappedSuperclass mappingType, MappedSuper
 		}
 	}
 
+	private <X> void applyGenericEmbeddableProperties(
+			PersistentClass persistentClass,
+			EntityDomainType<X> entityType) {
+		final MappedSuperclass mappedSuperclass = persistentClass.getSuperMappedSuperclass();
+		if ( mappedSuperclass != null ) {
```
beikov, line 580: I think this should be a while loop, since a mapped super class can have another mapped super class. Also, I think you should think check if the `PersistentClass` as a superclass and then go on checking for mapped superclasses there.

I would suggest the following:

```suggestion
		final MappedSuperclass mappedSuperclass = getMappedSuperclass( persistentClass );
		while ( mappedSuperclass != null ) {
```

and at the bottom `mappedSuperclass = getMappedSuperclass( mappedSuperclass )` and the additional method:

```java
private MappedSuperclass getMappedSuperclass(PersistentClass persistentClass) {
    do {
        final MappedSuperclass mappedSuperclass = persistentClass.getSuperMappedSuperclass();
        if ( mappedSuperclass != null ) {
            return mappedSuperclass;
        }
        persistentClass = persistentClass.getSuperclass();
    while (persistentClass != null);
    return null;
}

private MappedSuperclass getMappedSuperclass(MappedSuperclass mappedSuperclass) {
    return mappedSuperclass.getSuperMappedSuperclass() != null
        ? mappedSuperclass.getSuperMappedSuperclass()
        : getMappedSuperclass( mappedSuperclass.getSuperPersistentClass() );
}
```

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/internal/CompositeSqmPathSource.java
```diff
@@ -12,4 +12,5 @@
  * @author Steve Ebersole
  */
 public interface CompositeSqmPathSource<J> extends SqmPathSource<J> {
+	boolean isGeneric();
```
beikov, line 15: IMO this should even go on `SqmPathSource`

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/AbstractIdentifiableType.java
```diff
@@ -423,8 +424,10 @@ else if ( id != null ) {
 				final EmbeddableDomainType<?> compositeType = (EmbeddableDomainType<?>) type;
 				return new EmbeddedSqmPathSource<>(
 						EntityIdentifierMapping.ROLE_LOCAL_NAME,
+						(SqmPathSource) id,
 						compositeType,
-						Bindable.BindableType.SINGULAR_ATTRIBUTE
+						Bindable.BindableType.SINGULAR_ATTRIBUTE,
+						false
```
beikov, line 430: Is this the `generic` flag? Is this correct?

hibernate-core/src/main/java/org/hibernate/metamodel/internal/MetadataContext.java
mbladel: Done (with while instead of do-while in `getMappedSuperclass(PersistentClass persistentClass)`  because `mappedSuperclass.getSuperPersistentClass()` could be null).

PS: it would be nice if there was a way to avoid iterating through all mapped superclasses and their properties at this point, but I couldn't think of any.

hibernate-core/src/main/java/org/hibernate/metamodel/model/domain/AbstractIdentifiableType.java
mbladel: Nope, with `isGeneric()` on `SqmPathSource' (which is great IMO) I could pass `id.isGeneric()` instead of false which should be correct.

PR #6161, opened 2023-02-23
HHH-16221Improve extensibility of CockroachDB and PostgreSQL dialects

Sanne:
https://hibernate.atlassian.net/browse/HHH-16221


Sanne:
Including https://hibernate.atlassian.net/browse/HHH-16222 as well

PR #6162, opened 2023-02-23
HHH-16166 ClassCastException, mixing mapping types for same field name on different entities

mbladel:
https://hibernate.atlassian.net/browse/HHH-16166

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/temptable/TemporaryTable.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/temptable/TemporaryTable.java
@@ -10,6 +10,7 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.UUID;
+import java.util.function.BiConsumer;
 import java.util.function.Function;
 
 import org.hibernate.boot.model.naming.Identifier;
@@ -18,6 +19,7 @@ import org.hibernate.boot.model.relational.QualifiedNameParser;
 import org.hibernate.boot.model.relational.QualifiedTableName;
 import org.hibernate.dialect.Dialect;
 import org.hibernate.engine.jdbc.Size;
+import org.hibernate.generator.Generator;
 import org.hibernate.id.OptimizableGenerator;
 import org.hibernate.id.enhanced.Optimizer;
 import org.hibernate.internal.CoreLogging;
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/temptable/TemporaryTable.java
```diff
@@ -216,8 +216,9 @@ public static TemporaryTable createIdTable(
 										}
 										final ModelPart fkTarget = keyDescriptor.getTargetPart();
 										if ( !fkTarget.isEntityIdentifierMapping() ) {
-											final Value value = entityBinding.getSubclassProperty( pluralAttribute.getAttributeName() )
-													.getValue();
+											final PersistentClass declaringClass = runtimeModelCreationContext.getBootModel()
+													.getEntityBinding( ( (EntityMappingType) pluralAttribute.getDeclaringType() ).getEntityName() );
```
beikov, line 220: `pluralAttribute.getDeclaringType()` might be an embeddable, which brings me to the next possible problem, we are not considering embeddables in this code block. We need to use something like `org.hibernate.query.sqm.mutation.internal.SqmMutationStrategyHelper#visitCollectionTables` here instead of `entityDescriptor.visitSubTypeAttributeMappings`

mbladel: Thanks for the tip, I've updated the solution with similar logic for plural attributes and embeddables - LMK what you think.

hibernate-core/src/main/java/org/hibernate/dialect/temptable/TemporaryTable.java
```diff
@@ -198,61 +202,98 @@ public static TemporaryTable createIdTable(
...
+								// PostInitCallbackEntry because the callbacks
+								// are not ordered. The exception is caught in
+								// MappingModelCreationProcess.executePostInitCallbacks()
+								// and the callback is re-queued.
+								throw new IllegalStateException( "Not yet ready: " + pluralAttribute );
+							}
+							final ModelPart fkTarget = keyDescriptor.getTargetPart();
+							if ( !fkTarget.isEntityIdentifierMapping() ) {
+								final PersistentClass declaringClass = runtimeModelCreationContext.getBootModel()
+										.getEntityBinding( ( (EntityMappingType) pluralAttribute.getDeclaringType() ).getEntityName() );
```
beikov, line 220: I think this will blow up with a class cast exception if you write a test for it. Can you check if this line is even hit for your test? Either way, you should be able to fix it by calling `findContainingEntityMapping()` instead of `getDeclaringType()`

mbladel: I did add a mapping with a (nested) embeddable which declares a plural attribute and I can confirm the line is indeed hit and its declaring type is still an EntityMappingType. Looking at the code looks like when creating `PluralAttributeMappingImpl`s we always pass `EntityPersister`s as declaring types, even if the `ManagedMappingType` would allow embeddables here.

Anyway `findContainingEntityMapping()` is safer, so I've updated the PR with that.

beikov: >Looking at the code looks like when creating PluralAttributeMappingImpls we always pass EntityPersisters as declaring types, even if the ManagedMappingType would allow embeddables here.

This sounds problematic to me 😄 
Would you mind trying to pass the right thing there and see if it works?

mbladel: I'm trying but I see some breakage, I'm afraid we assumed to have an `EntityPersister` as declaring type in some places throughout the code and I'll need to review those if we wanna change that.

mbladel: > I'm afraid we assumed to have an EntityPersister as declaring type in some places throughout the code

So it looks like apart from some issues when creating the plural attribute mapping itself, which should be easily solvable, the only other thing that breaks is `org.hibernate.sql.results.graph.Initializer#isAttributeAssignableToConcreteDescriptor`. 

I don't know if we have a nice way of getting the containing entity type from an `AttributeMapping`, but if we don't maybe I could add something like `getDeclaringEntity()` which is only different to `getDeclaringType()` for plural attributes? Let me know if you have any tips, thanks.

beikov: We already have `findContainingEntityMapping()` on `ManagedType`, so you could use `getDeclaringType().findContainingEntityMapping()` already.

mbladel: Sorry I keep forgetting about that method for some reason :disappointed: 

Looks like it should work, I'll push a separate commit with this change (also for `ToOneAttributeMapping`).

PR #6165, opened 2023-02-23
HHH-16198 - Fix splitting of SqlAstCreationState

sebersole:
Of the 7 total failures, 5 are related to hard-coded alias expectations - see `org.hibernate.orm.test.query.CompareEntityValuedPathsTest`


sebersole:
Integrated upstream

PR #6169, opened 2023-02-27
HHH-16233 Fix logs for mutable non-root entity

shin-mallang:
An incorrect error log message is given about @Immutable is declared to non-root entities that are not @Immutable declared.

@Immutable can only be declared on the Root Entity.
That is, it is not wrong to say that a non-root entity is mutable, so should not send the message `@Immutable used on a non root entity` to non-root entity that are not declared with @Immutable.

This should only happen for non-root entities that are immutable.

https://hibernate.atlassian.net/jira/software/c/projects/HHH/issues/HHH-16233

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/internal/EntityBinder.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/internal/EntityBinder.java
@@ -1188,7 +1188,7 @@ public class EntityBinder {
 		if ( persistentClass instanceof RootClass ) {
 			bindRootEntity();
 		}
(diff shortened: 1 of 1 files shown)


gavinking:
Oops, I suck!

shin-mallang:
@gavinking Did I do a good job??

gavinking:
> @gavinking Did I do a good job??

Yes, sure, we just need to make sure the tests pass.

gavinking:
Merged, thanks!

shin-mallang:
I'm honored😊

PR #6173, opened 2023-02-27
HHH-16211 Error with like predicate's escape literal and converted char array property

mbladel:
https://hibernate.atlassian.net/browse/HHH-16211

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -6925,9 +6925,9 @@ public abstract class BaseSqmToSqlAstConverter<T extends Statement> extends Base
 		return new LikePredicate(
 				visitWithInferredType( predicate.getMatchExpression(), predicate.getPattern() ),
 				visitWithInferredType( predicate.getPattern(), predicate.getMatchExpression() ),
-				predicate.getEscapeCharacter() == null
-						? null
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -6931,9 +6931,9 @@ public LikePredicate visitLikePredicate(SqmLikePredicate predicate) {
 		return new LikePredicate(
 				visitWithInferredType( predicate.getMatchExpression(), predicate.getPattern() ),
 				visitWithInferredType( predicate.getPattern(), predicate.getMatchExpression() ),
-				predicate.getEscapeCharacter() == null
-						? null
-						: visitWithInferredType( predicate.getEscapeCharacter(), predicate.getMatchExpression() ),
+				predicate.getEscapeCharacter() == null ?
+						null :
+						(Expression) predicate.getEscapeCharacter().accept( this ),
```
beikov, line 6936: I think we should push the `StandardBasicTypes.CHARACTER` type on the inference stack when visiting the escape character i.e. push `creationContext.getSessionFactory().getTypeConfiguration().getBasicTypeRegistry().resolve(StandardBasicTypes.CHARACTER)`, then `predicate.getEscapeCharacter().accept( this )` and finally pop again.

mbladel: Done this with `visitWithInferredType( predicate.getEscapeCharacter(), () -> basicType( Character.class ) )`.

PR #6178, opened 2023-02-28
HHH-16182 - Converted boolean values not always properly handled in predicates

sebersole:
HHH-16182 - Converted boolean values not always properly handled in predicates

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -88,16 +88,12 @@ import org.hibernate.metamodel.mapping.internal.OneToManyCollectionPart;
 import org.hibernate.metamodel.mapping.internal.SqlTypedMappingImpl;
 import org.hibernate.metamodel.mapping.internal.ToOneAttributeMapping;
 import org.hibernate.metamodel.mapping.ordering.OrderByFragment;
-import org.hibernate.type.descriptor.converter.internal.OrdinalEnumValueConverter;
-import org.hibernate.type.descriptor.converter.spi.BasicValueConverter;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -7105,20 +7100,21 @@ private JdbcMappingContainer getBooleanType() {
 	@Override
 	public Object visitBooleanExpressionPredicate(SqmBooleanExpressionPredicate predicate) {
 		final Expression booleanExpression = (Expression) predicate.getBooleanExpression().accept( this );
-		if ( booleanExpression instanceof SelfRenderingExpression ) {
-			final Predicate sqlPredicate = new SelfRenderingPredicate( (SelfRenderingExpression) booleanExpression );
-			if ( predicate.isNegated() ) {
-				return new NegatedPredicate( sqlPredicate );
-			}
-			return sqlPredicate;
-		}
```
beikov, line 7114: I think this code here was to support functions as predicates, though it seems we lack a test for this. Can you please re-add it or do you think the code path would never be hit?

beikov (approved):

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
sebersole: I mean, there should be a test...

sebersole: Did not hear a reply, so added 2 basic tests of boolean functions wrt predicates -

1. as a predicate - `... where boolean_function`
2. in a predicate - `... where boolean_function = true`

Both work with the code as-is

sebersole:
Applied upstream

PR #6181, opened 2023-02-28
HHH-15895 IllegalArgumentException :Cannot create binding for parameter referencen with criteria builder 

dreab8:
https://hibernate.atlassian.net/browse/HHH-15895

different instances of `SqmParameter` instances can have the same hashcode value, so when the number of `SqmJpaCriteriaParameterWrapper` is huge the probability of considering 2 instances equals is high if  
`return Integer.compare( o1.hashCode(), o2.hashCode() );` is used to compare them.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/DomainParameterXref.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/DomainParameterXref.java
@@ -42,38 +42,9 @@ public class DomainParameterXref {
 		// `.. where a.b = :param or a.c = :param`.  Here we have 2 SqmParameter
 		// references (one for each occurrence of `:param`) both of which map to
 		// the same QueryParameter.
-		final Map<SqmParameter,QueryParameterImplementor<?>> xrefMap = new TreeMap<>(
-				(o1, o2) -> {
-					if ( o1 instanceof SqmNamedParameter ) {
-						final SqmNamedParameter<?> one = (SqmNamedParameter<?>) o1;
-						return o2 instanceof SqmNamedParameter<?>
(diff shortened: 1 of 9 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/internal/DomainParameterXref.java
```diff
@@ -42,38 +42,9 @@ public static DomainParameterXref from(SqmStatement<?> sqmStatement) {
...
-//							return o1.getName().compareTo( o2.getName() );
-//						}
-//						else {
-							return Integer.compare( o1.hashCode(), o2.hashCode() );
-//						}
-					}
-
-					throw new HibernateException( "Unexpected SqmParameter type for comparison : " + o1 + " & " + o2 );
-				}
+		final Map<SqmParameter, QueryParameterImplementor<?>> xrefMap = new TreeMap<>(
```
beikov, line 45: I think we should question the use of `TreeMap` here in general. Why is this not a simple `HashMap`?

sebersole: Because the TreeMap does deduplication.  If we can still dedup with a HashMap, then sure

sebersole: Now, I'm not sure why that commented code got commented out and I think that may be part of the problem here.

However, the approach here is perfectly fine also.

sebersole: Though we should probably make `SqmParameter` as `Comparable` and drop the lambda in the `TreeMap` ctor

dreab8:
merged upstream

PR #6183, opened 2023-03-01
Change scopes in AbstractEntityInitializer and add getters/setters

DavideD:
Hibernate Reactive needs custom initializers

Fix https://hibernate.atlassian.net/browse/HHH-16242


DavideD:
Thanks

PR #6189, opened 2023-03-02
HHH-15802 SubQuery with 'in' results in java.lang.ClassCastExceptionclass org.hibernate.metamodel.mapping.internal.BasicEntityIdentifierMappingImpl cannot be cast to class org.hibernate.metamodel.mapping.EntityValuedModelPart

dreab8:

### DIFF
--- a/documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
+++ b/documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
@@ -2443,7 +2443,7 @@ public class HQLTest extends BaseEntityManagerFunctionalTestCase {
 			List<Person> persons = entityManager.createQuery(
 				"select p " +
 				"from Person p " +
-				"where p.createdOn > '1950-01-01' ",
+				"where p.createdOn > date 1950-01-01 ",
 				Person.class)
 			.getResultList();
 			//end::hql-relational-comparisons-example[]
--- a/documentation/src/test/java/org/hibernate/userguide/mapping/basic/InstantLiteralTest.java
+++ b/documentation/src/test/java/org/hibernate/userguide/mapping/basic/InstantLiteralTest.java
@@ -51,7 +51,7 @@ public class InstantLiteralTest extends BaseEntityManagerFunctionalTestCase {
 			.createQuery(
 				"select de " +
 				"from DateEvent de " +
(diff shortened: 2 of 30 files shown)


beikov (commented):
The change from `instanceof TemporalJavaType<?>` to `.isTemporalType()` and `instanceof UnknownBasicJavaType<?>` to `.isUnknownType()` has the potential to NPE, because not every expression needs to return a non-null type, so we should probably add a static helper methods that also do a null check.

beikov:
The following will be added to the 6.2 migration guide:

`````
[[query-path-comparison]]
=== Query Path comparison

As of 6.2, comparisons of paths are type checked early. This means that a comparison predicate in HQL or JPA Criteria
might fail to construct if the types of the left and right hand side are not compatible.

In general, two types T1 and T2 are considered compatible if

* T1 == T2
* T1 instanceof T2 or T2 instanceof T1
* T1 is temporal and T2 is temporal
* T1 or T2 is unknown
* T1 can be widened/coerced to T2, or the other way around

Widening/Coercion usually refers to e.g. widening an integer to a long, but can also mean
that a string constant can be interpreted as enum when comparing against an enum attribute.

Note that a comparison of a temporal attribute against a string literal worked before

```sql
from MyEntity e where e.temporalAttribute > '2020-01-01'
```

but has to be changed to the proper temporal literal now

```sql
from MyEntity e where e.temporalAttribute > date 2020-01-01
```
`````

adityadwadasi:
Hi Beikov,
I am upgrading the springboot from 2.7.18 to 3.3.3. 
Hibernate versions also upgraded as part of it to 6.5.2.Final.

After that I am facing the same ClassCastException as mentioned in the description.
I also tried in the latest version 6.6.1, I am still getting the same exception.

After debugging i found out that this exception is occurring when we are trying to get getSingleResult() from a criteria query which supposed to return count rows.

Could you please help me with this.

beikov:
@adityadwadasi Please try to create a reproducer with our [test case template](https://github.com/hibernate/hibernate-test-case-templates/blob/master/orm/hibernate-orm-6/src/test/java/org/hibernate/bugs/JPAUnitTestCase.java) and if you are able to reproduce the issue, create a bug ticket in our [issue tracker](https://hibernate.atlassian.net) and attach that reproducer.

PR #6198, opened 2023-03-03
HHH-16251 Hibernate Envers wrong modified flag on OneToMany relations

Sun8642:
Add test case to reproduce the issue


beikov:

PR #6199, opened 2023-03-03
HHH-16213 Right join wrongly added to subquery when using FetchMode.SUBSELECT and is null predicate

mbladel:
https://hibernate.atlassian.net/browse/HHH-16213

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstWalker.java
@@ -434,7 +434,12 @@ public class AbstractSqlAstWalker implements SqlAstWalker {
 
 	@Override
 	public void visitTableGroupJoin(TableGroupJoin tableGroupJoin) {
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AliasCollector.java
```diff
@@ -53,4 +53,17 @@ public void visitQueryPartTableReference(QueryPartTableReference tableReference)
...
 	}
+
+	@Override
+	public void visitTableGroupJoin(TableGroupJoin tableGroupJoin) {
+		final TableGroup joinedGroup = tableGroupJoin.getJoinedGroup();
+		if ( joinedGroup.isInitialized() ) {
+			// Only process already initialized table groups to avoid
+			// forced initialization of joined lazy table groups
+			joinedGroup.accept( this );
+		}
```
beikov, line 64: Can you please move this check up to the `AbstractSqlAstWalker`?

mbladel: Sure, updated.

PR #6203, opened 2023-03-04
Testing database configuration for MySQL has mismatched credentials

Sanne:
This is a bit odd - not sure why nobody noticed, perhaps you all have your custom scripts? I hope this won't be too inconvenient to fix.

Essentially the configuration expected by the integration tests e.g. 

```
./gradlew :hibernate-core:test -Pdb=mysql
```

Doesn't match the credentials of MySQL when started via

```
./docker_db.sh mysql
```

Wondering if I'm missing something but according to README this should work?

### DIFF
--- a/gradle/databases.gradle
+++ b/gradle/databases.gradle
@@ -84,8 +84,8 @@ ext {
                 mysql : [
                         'db.dialect' : 'org.hibernate.dialect.MySQLDialect',
                         'jdbc.driver': 'com.mysql.cj.jdbc.Driver',
-                        'jdbc.user'  : 'hibernateormtest',
-                        'jdbc.pass'  : 'hibernateormtest',
+                        'jdbc.user'  : 'hibernate_orm_test',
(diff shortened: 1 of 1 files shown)


gavinking:
I think you’re supposed to use `-Pdb=mysql_ci`. The plain `mysql ` version is for a local install of MySQL.

Sanne:
I know `mysql_ci` was meant originally for our ci systems - perhaps I'm also meant to use it locally now? But other databases do have credentials matching the docker script.

So perhaps I'm confused and since I'm uncertain I'm not pushing this yet .. but if I'm wrong would be nice to clarify the instructions.

sebersole:
Those `*_ci` databases use containers.  They are used in CI, but also locally to test inside containers you start with `./docker_db.sh`.

Have you looked at the bottom of `README.adoc`?

sebersole (commented):
We can apply this, but its unnecessary.

Sanne:
It's indeed in the README and I missed it - thanks will close my PR.

PR #6209, opened 2023-03-06
HHH-16015 Merge operation throws a NPE: Cannot invoke org.hibernate.property.access.spi.Setter.set(Object, Object) when using CompositeUserType

dreab8:
https://hibernate.atlassian.net/browse/HHH-16015

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/TypeHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/TypeHelper.java
@@ -11,6 +11,7 @@ import java.util.Map;
 import org.hibernate.Internal;
 import org.hibernate.bytecode.enhance.spi.LazyPropertyInitializer;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.metamodel.spi.EmbeddableInstantiator;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.property.access.internal.PropertyAccessStrategyBackRefImpl;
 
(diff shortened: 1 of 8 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/AbstractEmbeddableMapping.java
```diff
@@ -110,7 +111,12 @@ public void setValues(Object component, Object[] values) {
 		}
 		else {
 			for ( int i = 0; i < values.length; i++ ) {
-				getAttributeMapping( i ).getPropertyAccess().getSetter().set( component, values[i] );
+				final Setter setter = getAttributeMapping( i ).getPropertyAccess().getSetter();
+				// We need to check the setter is not null,
+				// i.e.  PropertyAccessCompositeUserTypeImpl#getSetter() returns null;
+				if ( setter != null ) {
```
beikov, line 117: Shouldn't we rather check outside of the loop if the `getRepresentationStrategy()` is someting custom and skip the whole loop altogether then? Also, it's a bit concerning that this method is called in the first place. Shouldn't we try to create a copy of the component instead?

dreab8: `TypeHelper#replaceAssociations()`, calls `ComnponentType#setPropertyValues`,
probably a better solution is to change this last method to :
```
public void setPropertyValues(Object component, Object[] values) {
    if ( compositeUserType != null ) {
	return;
    }
    mappingModelPart.getEmbeddableTypeDescriptor().setValues( component, values );
}
```
what do you think?

beikov: I don't know if `TypeHelper#replaceAssociations()` will actually "do something", but maybe it's better to create a new copy of the component and replace that in the respective context if `compositeUserType != null`

dreab8: it should already be a copy, from the `TypeHelper#replaceAssociations` doc
```
* If the corresponding type is a component type, then apply {@link Type#replace} across the component
* subtypes but do not replace the component value itself.
```

dreab8: working on a solution

hibernate-core/src/main/java/org/hibernate/type/TypeHelper.java
```diff
@@ -182,41 +183,60 @@ public static Object[] replaceAssociations(
...
+							targetComponentValues,
+							subtypes,
+							session,
+							null,
+							copyCache,
+							foreignKeyDirection
+					);
+					if ( target[i] != null ) {
+						if( !( compositeType instanceof ComponentType ) ){
+							compositeType.setPropertyValues( target[i], objects );
```
beikov, line 216: As far as I understand the code, this will just always fail if we have a `AnyType`. IMO this should all rather be like the following:

```java
if ( target[i] != null && compositeType instanceof ComponentType ) {
	final ComponentType componentType = (ComponentType) compositeType;
	if ( componentType.isCompositeUserType() ) {
		final EmbeddableInstantiator instantiator = ( (ComponentType) compositeType ).getMappingModelPart()
			.getEmbeddableTypeDescriptor()
			.getRepresentationStrategy()
			.getInstantiator();
		target[i] = instantiator.instantiate( () -> objects, session.getSessionFactory() );
	}
	else {
		compositeType.setPropertyValues( target[i], objects );
	}
}
```

beikov (commented):
One last improvement and then we can merge this :)

hibernate-core/src/main/java/org/hibernate/type/TypeHelper.java
dreab8: done

PR #6211, opened 2023-03-06
HHH-16218 CompoundNaturalIdMapping#calculateHashCode does no longer always return 0

sdusart:
The test case (CompoundNaturalIdCacheTest) is mainly a performance test but I added it in the unit tests of hibernate-core.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/internal/NaturalIdCacheKey.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/internal/NaturalIdCacheKey.java
@@ -38,7 +38,7 @@ public class NaturalIdCacheKey implements Serializable {
 	}
 
 	public NaturalIdCacheKey(Object naturalIdValues, EntityPersister persister, String entityName, SharedSessionContractImplementor session) {
-		this( persister.getNaturalIdMapping().disassemble( naturalIdValues, session ), entityName, session.getTenantIdentifier(), persister.getNaturalIdMapping().calculateHashCode( naturalIdValues ) );
+		this( persister.getNaturalIdMapping().disassemble( naturalIdValues, session ), entityName, session.getTenantIdentifier(), persister.getNaturalIdMapping().calculateHashCode( naturalIdValues, session ) );
 	}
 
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/cache/internal/NaturalIdCacheKey.java
```diff
@@ -38,7 +38,7 @@ public NaturalIdCacheKey(Object naturalIdValues, EntityPersister persister, Shar
 	}
 
 	public NaturalIdCacheKey(Object naturalIdValues, EntityPersister persister, String entityName, SharedSessionContractImplementor session) {
-		this( persister.getNaturalIdMapping().disassemble( naturalIdValues, session ), entityName, session.getTenantIdentifier(), persister.getNaturalIdMapping().calculateHashCode( naturalIdValues ) );
+		this( persister.getNaturalIdMapping().disassemble( naturalIdValues, session ), entityName, session.getTenantIdentifier(), persister.getNaturalIdMapping().calculateHashCode( naturalIdValues, session ) );
```
sebersole, line 41: You effectively do the disassemble twice, not sure how that is more performant.  Would be better to do that disassembly once.  Maybe make the disassembled state be a pre-condition to `#calculateHashCode`

sebersole: Another (probably better) option is to calculate the hash-code using the attributes

sdusart: Do you mean without disassembling ?
That's what I did initially but it broke a unit test (it would require to implement equals()/hashCode() for every attribute).

sdusart: I presume a valid correction will be brought by #6248 :)

beikov:
Fixed via https://github.com/hibernate/hibernate-orm/pull/6320

PR #6213, opened 2023-03-07
HHH-16223 Deprecate all "javax." constants defined in org.hibernate.cfg.AvailableSettings

Sanne:
https://hibernate.atlassian.net/browse/HHH-16223

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java
@@ -1698,7 +1698,7 @@ public interface AvailableSettings {
 	 * <p>
 	 * The default value is {@code /import.sql}.
 	 * <p>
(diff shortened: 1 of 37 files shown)


Sanne:
@sebersole ok I'll try to get to that but it means it will take some more time - wouldn't you prefer to treat that as a separate step?

sebersole:
I'll make an exception for you ;)

But it is a bad idea generally imo to accept changes to make a setting deprecated without adding the logging for that.

Sanne:
I'm wondering now if I'm not missing something; isn't it implied that all `javax.` properties are deprecated?

I didn't mean to "deprecate them" (now), I simply thought we had forgotten to mark these as being deprecated.

But sure no worries I'll figure out if I can add the logs; I thought this was already done.

Sanne:
I think I've don that now - indeed it was mostly done already :+1:

PR #6214, opened 2023-03-07
[6.2] HHH-16172 BasicCacheKeyImplementation must be public to support externalization.

pferraro:
https://hibernate.atlassian.net/browse/HHH-16172

Original HHH-16172 fix neglected to make this class public so that it can be instantiated outside the org.hibernate.cache.internal package.


pferraro:
@Sanne FYI

Sanne (approved):

Sanne:
Thanks @pferraro !

PR #6216, opened 2023-03-07
HHH-16249 - Add test

jrenaat:
Disable batching when no transaction is used in a stateless session

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/jdbc/Expectation.java
+++ b/hibernate-core/src/main/java/org/hibernate/jdbc/Expectation.java
@@ -18,7 +18,7 @@ import org.hibernate.HibernateException;
 public interface Expectation {
 
 	/**
-	 * Is it acceptable to combiner this expectation with statement batching?
+	 * Is it acceptable to combine this expectation with statement batching?
 	 *
 	 * @return True if batching can be combined with this expectation; false otherwise.
 	 */
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/mutation/DeleteCoordinator.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/mutation/DeleteCoordinator.java
@@ -133,7 +133,9 @@ public class DeleteCoordinator extends AbstractMutationCoordinator {
 		return session.getFactory()
 				.getServiceRegistry()
 				.getService( MutationExecutorService.class )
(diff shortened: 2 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/internal/StatelessSessionImpl.java
```diff
@@ -646,4 +647,20 @@ public boolean isStatelessSession() {
...
+		if ( isClosed() ) {
+			LOG.trace( "Already closed" );
+			return;
+		}
+
+		if ( LOG.isTraceEnabled() ) {
+			LOG.tracef( "Closing session [%s]", getSessionIdentifier() );
+		}
+
+		getJdbcCoordinator().executeBatch();
```
beikov, line 661: It seems `SessionImpl` has some methods which `StatelessSessionImpl` lacks that could affect what the inherited `close` method does:

```
	@Override
	public void delayedAfterCompletion() {
		TransactionCoordinator coordinator = getTransactionCoordinator();
		if ( coordinator instanceof JtaTransactionCoordinatorImpl ) {
			( (JtaTransactionCoordinatorImpl) coordinator).getSynchronizationCallbackCoordinator()
					.processAnyDelayedAfterCompletion();
		}
	}

	@Override
	protected boolean shouldCloseJdbcCoordinatorOnClose(boolean isTransactionCoordinatorShared) {
		if ( !isTransactionCoordinatorShared ) {
			return super.shouldCloseJdbcCoordinatorOnClose( isTransactionCoordinatorShared );
		}

		final ActionQueue actionQueue = getActionQueue();
		if ( actionQueue.hasBeforeTransactionActions() || actionQueue.hasAfterTransactionActions() ) {
			log.warn(
					"On close, shared Session had before/after transaction actions that have not yet been processed"
			);
		}
		return false;
	}
```

Can you please set a break point in `JdbcCoordinator.executeBatch()` to see how that method is called from `SessionImpl` and whether adding one or both methods I mentioned here to `StatelessSessionImpl` could be a "more appropriate fix".

jrenaat: Tbh I don't think either of these two places are appropriate for executing any lingering batch when the session is about to be closed, at least not in any similar form to what SessionImpl does in them

PR #6217, opened 2023-03-08
HHH-16260 - JdbcParameterRenderer not called with dynamic filters

sebersole:
https://hibernate.atlassian.net/browse/HHH-16260

Includes general improvements to filter handing.  Drops the need for separate "filter parameter" tracking.

Part of that change was to push a not strictly "clean" representation into the SQL AST.  Mainly it delays processing parameters until we are in the `SqlAstTranslator` so it can apply the renderer.  Its also what facilitates the improved parameter handling.

### DIFF
--- a/hibernate-agroal/src/test/resources/hibernate.properties
+++ b/hibernate-agroal/src/test/resources/hibernate.properties
@@ -11,6 +11,9 @@ hibernate.connection.username @jdbc.user@
 hibernate.connection.password @jdbc.pass@
 hibernate.connection.init_sql @connection.init_sql@
 
+# some Agroal tests fail with native markers.  not sure why yet
+hibernate.dialect.native_param_markers=false
+
 hibernate.jdbc.batch_size 10
 hibernate.connection.provider_class AgroalConnectionProvider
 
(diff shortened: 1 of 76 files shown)


beikov (approved):

sebersole:
@Sanne I added the fix for your `HHH-16256 - JdbcParameterRenderer to have an impact on write operations` issue here - just waiting for checks to pass.  Anyway, this addresses the issues I know about regarding parameter handling for reactive (aside from the legacy generators which we already talked about).

sebersole:
Going to add one more commit as well that drops `JdbcParameterRenderer` in favor of a setting to use Dialect native parameter markers.

sebersole:
> Going to add one more commit as well that drops `JdbcParameterRenderer` in favor of a setting to use Dialect native parameter markers.

Actually, I decided not to drop `JdbcParameterRenderer` yet.  For now, I just have the standard initiator consider whether to use (new method) `Dialect#getNativeParameterRenderer` .  

In the short term this lets us use `JdbcParameterRenderer` from those legacy SQL generators until we can get those converted to SQL AST.  I'll work on this last bit tomorrow.

sebersole:
Applied upstream

PR #6219, opened 2023-03-08
HHH-6215 Composite primary key @IdClass attribute mapping is borrowed from the first OneToMany backref and cannot be set

dreab8:
https://hibernate.atlassian.net/browse/HHH-16215

alternative solution to https://github.com/hibernate/hibernate-orm/pull/6218, basically it is a refactoring of Marco's PR on haw to detect if a `Fetchable` is selectable.


hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
```diff
@@ -688,6 +688,10 @@ private BiConsumer<Fetchable, Boolean> createFetchableBiConsumer(
 			LoaderSqlAstCreationState creationState,
 			ImmutableFetchList.Builder fetches) {
 		return (fetchable, isKeyFetchable) -> {
+			if ( !fetchable.isSelectable() ) {
```
beikov, line 691: I'm curious now. When you add that here, `BaseSqmToSqlAstConverter` is the only place where we still process the fetchable even though it is not selectable. Should we add the same condition there as well? What if we simply remove the fetchable from the `staticFetchableList` instead?

PR #6228, opened 2023-03-10
HHH-16279 Hibernate ORM with hibernate.hbm2ddl.import_files_sql_extractor = multi-line fails to process import.sql containing 

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16279

The patch seems to make sense, though I wouldn't be able to tell you why the parser was fine with an empty file before... There's probably some early check that's not part of the grammar.


yrodiere:
Merged, thanks

sebersole:
Grr was in the process of squashing and merging ;)

yrodiere:
Sorry :shrug:

sebersole:
No problem.  Just apply it to 6.2 branch as well please.

yrodiere:
Right. Done.

PR #6229, opened 2023-03-10
HHH-16280 Fix Jackson XML mapper support for Oracle Array data types

mbladel:
https://hibernate.atlassian.net/browse/HHH-16280


hibernate-core/src/main/java/org/hibernate/type/format/jackson/JacksonXmlFormatMapper.java
```diff
@@ -24,10 +32,13 @@ public final class JacksonXmlFormatMapper implements FormatMapper {
...
 
 	public JacksonXmlFormatMapper() {
-		this(new XmlMapper());
+		this( new XmlMapper() );
 	}
 
 	public JacksonXmlFormatMapper(ObjectMapper objectMapper) {
+		// needed to automatically find and register Jackson's jsr310 module for java.time support
+		objectMapper.findAndRegisterModules();
+		objectMapper.configure( SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false );
```
beikov, line 41: We should only do that for the `XmlMapper` that we create ourselves.

mbladel:
FTR there is an issue with how Jackson XML handles `null` strings inside arrays which causes them to be deserialized as empty values (`""`) instead. See https://github.com/FasterXML/jackson-dataformat-xml/issues/344.

In the meantime we're using a custom deserializer as a workaround.

hibernate-core/src/main/java/org/hibernate/type/format/jackson/JacksonXmlFormatMapper.java
mbladel: Just as a note, a mapper we receive would need to be configured like we expect to handle this cases (i.e. string / datetime arrays).

beikov: Yeah, I don't expect people to simply extend this class and passing something without first looking into what the base class does, so this is fine. Might be worth making `createXmlMapper` `protected` though along with the rebase as soon as https://github.com/hibernate/hibernate-orm/pull/6268 is merged

mbladel: The class is declared as `final` though, so it seems we aren't expecting anyone to extend this.

sebersole:
Have you checked if this works with https://github.com/hibernate/hibernate-orm/pull/6268?

sebersole:
Nevermind, that one is failing.  I'm going to integrate this one for CR4

sebersole:
Integrated into main and 6.2

PR #6230, opened 2023-03-10
Make it possible for Hibernate Reactive to plug in custom fetch initializers

DavideD:
Basically, it creates some protected methods or changes some existing ones so that Hibernate Reactive can instantiate its own classes 

Fix https://hibernate.atlassian.net/browse/HHH-16282


beikov:
I'm curious, why do you need this? This doesn't build on JDBC APIs directly, but only indirectly through `org.hibernate.sql.results.jdbc.spi.JdbcValues` being accessed in `org.hibernate.sql.results.internal.RowProcessingStateStandardImpl`. Also, I thought you implement JDBC APIs, so even that shouldn't be an issue.

PR #6234, opened 2023-03-11
[Question] Update collection of embeddables bug

DavideD:
It seem that updating an element in a collection of elements will delete the whole collection and only keep the new value.

Am I doing something wrong or is this a bug in ORM 6.2?


beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/6245

PR #6237, opened 2023-03-13
HHH-16189 wrong order by and group by generated

dreab8:
https://hibernate.atlassian.net/browse/HHH-16189

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/BasicValuedPathInterpretation.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/BasicValuedPathInterpretation.java
@@ -13,18 +13,21 @@ import java.util.function.Consumer;
 import org.hibernate.metamodel.MappingMetamodel;
 import org.hibernate.metamodel.mapping.BasicValuedModelPart;
 import org.hibernate.metamodel.mapping.EntityMappingType;
+import org.hibernate.metamodel.mapping.ManagedMappingType;
 import org.hibernate.metamodel.mapping.ModelPart;
+import org.hibernate.metamodel.mapping.ModelPartContainer;
 import org.hibernate.metamodel.model.domain.EntityDomainType;
+import org.hibernate.query.sqm.tree.from.SqmFrom;
 import org.hibernate.spi.NavigablePath;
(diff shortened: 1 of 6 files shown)


dreab8:
yeap. Derby, Oracle, DB2 and SQLServer were failing, they want the id to be present in both the group and order by.

I had also  changed https://github.com/hibernate/hibernate-orm/pull/6237/files#diff-d177a632e40c816cc34f0e75e48cfd23535d0f7c626597bfc4adbaf861b72514R79 adding ` || currentClause == Clause.ORDER || currentClause == Clause.HAVING`  to avoid failure of queries like
```
select c.id, sum(p.number) " +
"from Parent p " +
"inner join p.child c " +
"group by c.id " +
"order by c.id
```

and
 ```
"select c.id, sum(p.number) " +
"from Parent p " +
"inner join p.child c " +
"group by c.id " +
"having c.id > 5",
```

I'm going to check with embeddables

**Comment by dreab8** [comment]
@beikov added [EmbeddedIdOrderByAndAggregateFunctionTest](https://github.com/hibernate/hibernate-orm/pull/6237/files#diff-3da09d3c782286da823573df1bc455027f500f765068b42c95e642d41af1d178)

File: hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/BasicValuedPathInterpretation.java [hunk 1]
```diff
@@ -68,10 +72,24 @@ else if ( sqmPath.getLhs().getNodeType() instanceof EntityDomainType ) {
 			}
 		}
 
-		final BasicValuedModelPart mapping = (BasicValuedModelPart) tableGroup.getModelPart().findSubPart(
-				sqmPath.getReferencedPathSource().getPathName(),
-				treatTarget
-		);
+		final ModelPartContainer modelPart = tableGroup.getModelPart();
+		final BasicValuedModelPart mapping;
+		// In the select and group by clause we have to make sure we render the column of the target table,
```
beikov, line 77: Can you please add `order by` and `having` clauses to the comment?

beikov (commented):
Looks like `EmbeddedIdOrderByAndAggregateFunctionTest` is failing on PostgreSQL.

PR #6239, opened 2023-03-13
Failure merging a referenced entity

DavideD:
This test case seems to fail in ORM.
Am I doing something wrong?

Thanks

Note that the join column references a column that's not the identifier.


beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/6495

PR #6241, opened 2023-03-13
HHH-16290 - Mark `o.h.persister.entity` and `o.h.persister.collection` as internal

sebersole:
HHH-16290 - Mark `o.h.persister.entity` and `o.h.persister.collection` as internal


sebersole:
Applied upstream

PR #6242, opened 2023-03-13
Prefer `#value` over `#clause` for `@Where` and `@OrderBy`

sebersole:
HHH-16292 - Rename @OrderBy#clause to @OrderBy#value
HHH-16291 - Rename @Where#clause to @Where#value


beikov:
Just wanted to note that there is an existing issue for this: https://hibernate.atlassian.net/browse/HHH-16163

PR #6243, opened 2023-03-14
update maven dependency of hibernate-spatial

florianhof:
the new Maven group is org.hibernate.orm (and so referenced in Spring Boot 3)


beikov:
Thanks

PR #6245, opened 2023-03-14
HHH-16297 Removing an element from a collection of elements removes the whole collection

dreab8:
https://hibernate.atlassian.net/browse/HHH-16297

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
@@ -1337,6 +1337,7 @@ public class MappingModelCreationHelper {
 					basicElement.isColumnInsertable( 0 ),
 					basicElement.isColumnUpdateable( 0 ),
 					basicElement.isPartitionKey(),
+					true, // element collection does not support null elements
 					dialect,
 					creationProcess.getSqmFunctionRegistry()
 			);
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/SelectableMappingImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/SelectableMappingImpl.java
@@ -9,14 +9,12 @@ package org.hibernate.metamodel.mapping.internal;
 import java.util.Locale;
 
 import org.hibernate.dialect.Dialect;
-import org.hibernate.mapping.BasicValue;
 import org.hibernate.mapping.Column;
 import org.hibernate.mapping.Selectable;
 import org.hibernate.metamodel.mapping.JdbcMapping;
 import org.hibernate.metamodel.mapping.SelectableMapping;
 import org.hibernate.metamodel.mapping.SelectablePath;
 import org.hibernate.query.sqm.function.SqmFunctionRegistry;
-import org.hibernate.type.BasicType;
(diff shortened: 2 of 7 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/embeddable/EmbeddableElementCollectionUpdateTest.java
```diff
@@ -0,0 +1,160 @@
...
+			}
+			Person person = (Person) o;
+			return Objects.equals( name, person.name );
+		}
+
+	}
+
+	@Embeddable
+	public static class Address {
+		@Basic(optional = false)
```
beikov, line 130: Would be interesting to have a test case where one of the attributes is also optional, to see if the null checks are properly handled during deletion.

hibernate-core/src/main/java/org/hibernate/persister/collection/BasicCollectionPersister.java
```diff
@@ -699,7 +705,7 @@ private void applyDeleteRowRestrictions(
 						jdbcValueBindings,
 						null,
 						(valueIndex, bindings, noop, jdbcValue, jdbcValueMapping) -> {
-							if ( jdbcValueMapping.isNullable() || jdbcValueMapping.isFormula() ) {
+							if ( !elementColumnIsInPrimaryKey[valueIndex] || jdbcValueMapping.isFormula() ) {
```
beikov, line 708: The change is unnecessary, because `!elementColumnIsInPrimaryKey[valueIndex] == !jdbcValueMapping.isNullable()`. Please revert it.

hibernate-core/src/main/java/org/hibernate/persister/collection/BasicCollectionPersister.java
```diff
@@ -645,7 +645,13 @@ private RestrictedTableMutation<JdbcMutationOperation> generateDeleteRowAst(Muta
 				deleteBuilder.addKeyRestrictionsLeniently( pluralAttribute.getIndexDescriptor() );
 			}
 			else {
-				deleteBuilder.addKeyRestrictions( pluralAttribute.getElementDescriptor() );
```
beikov, line 648: `elementColumnIsInPrimaryKey` essentially is a check if `!isNullable()`. We worked on supporting nullable columns in the past, but I don't remember where that work went. Maybe you didn't create a PR for that back then?

Anyway, I think the correct solution for this problem would be to handle nullable columns properly by using the `distinct from` predicate like we did in our earlier work.

beikov (changes requested):

hibernate-core/src/main/java/org/hibernate/persister/collection/BasicCollectionPersister.java
dreab8: in this case is not true that `elementColumnIsInPrimaryKey essentially is a check if !isNullable()` we have columns that are null but they are PK. I just reintroduced what we do in 5. For what I remember the `distinc from` approach was super complicated

hibernate-core/src/test/java/org/hibernate/orm/test/embeddable/EmbeddableElementCollectionUpdateTest.java
dreab8: I'll add the test

PR #6246, opened 2023-03-14
HHH-16311 - Migrate away from UserType for enum handling

sebersole:
Simple proof for handling enum values with the UserType machinery


gavinking:
Does this eliminate all usages of `org.hibernate.type.EnumType`? (I ask because you didn't delete it.)

sebersole:
Well, that's the question isn't it ;)  As I said, this is all just a quick proof.  Still need to analyse all uses of `EnumeratedValueResolution#getLegacyResolvedBasicType`.  

One use-case that comes to mind immediately is the SQM handling of enum value names as literals (i.e. `... gender=MALE`).

sebersole:
Actually this was really easy to test.  There were just a few errors - looking through them

gavinking:
> One use-case that comes to mind immediately is the SQM handling of enum value names as literals (i.e. `... gender=MALE`).

this is one of the things that was causing me a lot of pain when I tried to handle mapping to Postgres enum types.

sebersole:
As a quick test I changed `EnumeratedValueResolution` to always return the `ConvertedBasicType`, even for the `#getLegacyResolvedBasicType` case.  That resulted in a few errors, but all were related to tests casting to the `CustomType` and `EnumType` classes directly.  I'll push here in a bit to see what affects it has across all the databases.

sebersole:
This latest push should be pretty complete.

One change I'll probably revert is that this push limits "enum short name" handling to just implicit and explicit `@Enumerated` mappings.  I was just playing around with hooks from `EnumeratedValueResolution` to think about native database type handling, aux database objects, etc.

gavinking:
OK then, I'll try making #6232 work on top of this patch, and see if it helps.

sebersole:
See especially `InFlightMetadataCollector#registerEnumMapping`...

I'm not yet sure what to think of the test failures from the last push.  They stem from the fact that we now interpret mappings like `@Enumerated(STRING) @Column(length=1)` using `Types.CHAR` / `Character`.  That change was never propagated into `EnumType` - meaning envers always saw `Types.VARCHAR` / `String`.  I'm inclined to think the consistency is a good thing.

I have not yet looked into the Oracle test failure from the second push

sebersole:
Integrated upstream

PR #6248, opened 2023-03-14
HHH-16281 Inconsistent Behaivor of L2 cache between Hibernate 5 and 6

dreab8:
https://hibernate.atlassian.net/browse/HHH-16281


hibernate-core/src/main/java/org/hibernate/type/descriptor/java/JavaTypedExpressible.java
```diff
@@ -11,4 +11,9 @@
  */
 public interface JavaTypedExpressible<T> {
 	JavaType<T> getExpressibleJavaType();
+
+	default int extractHashCode(T disassembledValue) {
```
beikov, line 15: I think this is unnecessary. We should just call the `hashCode` method on the disassembled state directly.

beikov:
During our last meeting we decided to use a different approach, which is to introduce a `addToCacheKey(MutableCacheKey)` on `JavaTypedExpressible` instead, which will add the constituent part and hash code to a `MutableCacheKey` directly. That way, we don't need to expose two methods and walk the model twice.

beikov:
When you work on this, could you please also make sure that we use the same approach for https://github.com/hibernate/hibernate-orm/pull/6211 @dreab8 ?

dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/6279

PR #6250, opened 2023-03-15
HHH-16310 Fix regressions in documentation and support for MultiTenantConnectionProvider

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16310

We probably also need an entry about `hibernate.multiTenancy` being removed in the 6.0 migration guide. I'll send another PR for that.


Sanne:
See also https://hibernate.atlassian.net/browse/HHH-16246

Sanne (approved):

yrodiere:
Thanks, merged and backported to 6.2.

PR #6252, opened 2023-03-15
HHH-16246 Document multitenancy changes

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16246

I *think* that's the gist of it, but @gavinking will know for sure.


gavinking (approved):
LGTM, thanks!

PR #6255, opened 2023-03-15
HHH-16316 Move version specific dialects of unsupported versions to hibernate-community-dialects

beikov:
https://hibernate.atlassian.net/browse/HHH-16316

### DIFF
--- a/documentation/src/main/asciidoc/userguide/chapters/domain/inheritance.adoc
+++ b/documentation/src/main/asciidoc/userguide/chapters/domain/inheritance.adoc
@@ -312,23 +312,6 @@ include::{extrasdir}/entity-inheritance-table-per-class-query-example.sql[]
 [IMPORTANT]
 ====
 Polymorphic queries require multiple UNION queries, so be aware of the performance implications of a large class hierarchy.
-
-Unfortunately, not all database systems support UNION ALL, in which case, UNION is going to be used instead of UNION ALL.
-
-The following Hibernate dialects support UNION ALL:
-
-- `AbstractHANADialect`
-- `AbstractTransactSQLDialect`
-- `CUBRIDDialect`
-- `DB2Dialect`
-- `H2Dialect`
-- `HSQLDialect`
-- `Ingres9Dialect`
-- `MySQL5Dialect`
-- `Oracle8iDialect`
-- `Oracle9Dialect`
-- `PostgreSQL81Dialect`
-- `RDMSOS2200Dialect`
 ====
 
 [[entity-inheritance-polymorphism]]
--- a/hibernate-core/src/main/java/org/hibernate/boot/registry/selector/internal/DefaultDialectSelector.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/registry/selector/internal/DefaultDialectSelector.java
@@ -9,58 +9,35 @@ package org.hibernate.boot.registry.selector.internal;
 import java.util.Objects;
 
 import org.hibernate.boot.registry.selector.spi.DialectSelector;
+import org.hibernate.boot.registry.selector.spi.StrategySelectionException;
 import org.hibernate.dialect.CockroachDialect;
-import org.hibernate.dialect.DB2390Dialect;
-import org.hibernate.dialect.DB2390V8Dialect;
 import org.hibernate.dialect.DB2400Dialect;
 import org.hibernate.dialect.DB2400V7R3Dialect;
-import org.hibernate.dialect.DB297Dialect;
 import org.hibernate.dialect.DB2Dialect;
 import org.hibernate.dialect.DB2iDialect;
 import org.hibernate.dialect.DB2zDialect;
 import org.hibernate.dialect.DerbyDialect;
-import org.hibernate.dialect.DerbyTenFiveDialect;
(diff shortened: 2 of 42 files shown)


hibernate-core/src/main/java/org/hibernate/boot/registry/selector/internal/DefaultDialectSelector.java
```diff
@@ -112,75 +89,85 @@ public Class<? extends Dialect> resolve(final String name) {
...
 			case "SybaseASE15":
-				return SybaseASE15Dialect.class;
 			case "SybaseASE157":
-				return SybaseASE157Dialect.class;
+				return findCommunityDialect( name );
 		}
 		return null;
 	}
 
+	private static Class<? extends Dialect> findCommunityDialect(String name) {
```
sebersole, line 154: Still not entirely sure about exception v. warning here.  As I understand, the incoming `name` is a value explicitly set for which Dialect to use.

What's the ultimate outcome in this case?  We fall back to database-metadata resolution I assume?

documentation/src/main/asciidoc/userguide/chapters/domain/inheritance.adoc
```diff
@@ -313,24 +313,6 @@ include::{extrasdir}/entity-inheritance-table-per-class-query-example.sql[]
...
-- `H2Dialect`
-- `HSQLDialect`
-- `Ingres9Dialect`
-- `MySQL5Dialect`
-- `Oracle8iDialect`
-- `Oracle9Dialect`
-- `PostgreSQL81Dialect`
-- `RDMSOS2200Dialect`
-====
-
```
yrodiere, line 333: Was that on purpose? It seems unrelated and breaks formatting (you're not closing the admonition block).

hibernate-core/src/main/java/org/hibernate/boot/registry/selector/internal/DefaultDialectSelector.java
yrodiere: +1 for exception. If something gets explicitly requested and we can't find it, there's a problem that should be addressed right now.

Falling back to database-metadata resolution is unlikely to help if the user selected a community (i.e. unsupported) dialect in the first place.

hibernate-core/src/main/java/org/hibernate/engine/jdbc/dialect/internal/DialectFactoryImpl.java
```diff
@@ -124,6 +153,20 @@ private Dialect constructDialect(Object dialectReference, DialectResolutionInfoS
...
+			final String dialectFqn = dialectReference.toString();
+			if ( LEGACY_DIALECTS.contains( dialectFqn ) ) {
+				throw new StrategySelectionException(
+						"Couldn't load the dialect class for the `hibernate.dialect` [" + dialectFqn + "], " +
+								"because the application is missing a dependency on the hibernate-community-dialects module. " +
+								"Hibernate 6.2 dropped support for database versions that are unsupported by vendors  " +
+								"and code for old versions was moved to the hibernate-community-dialects module. " +
+								"For further information, read https://in.relation.to/2023/02/15/hibernate-orm-62-db-version-support/",
+						e
+				);
```
yrodiere, line 166: :+1: great to give that info to users.

hibernate-core/src/test/java/org/hibernate/orm/test/strategyselectors/DefaultDialectSelectorTest.java
```diff
@@ -21,8 +25,6 @@ public class DefaultDialectSelectorTest {
 	@Test
 	public void verifyAllDialectNamingResolve() {
 		testDialectNamingResolution( DB2Dialect.class );
-		testDialectNamingResolution( DB2390Dialect.class );
-		testDialectNamingResolution( DB2390V8Dialect.class );
```
yrodiere, line 25: Should there be a dedicated test in the `hibernate-community-dialects` module, then? I understand we don't guarantee that dialects work, but I'd expect us to guarantee that name resolution, at least, does?

yrodiere (commented):
Thanks, a few comments below.

Please don't forget to update migration guides ;D

documentation/src/main/asciidoc/userguide/chapters/domain/inheritance.adoc
beikov: I'll fix the block closing, but the section is unnecessary, because all supported dialect versions support `union all` now.

hibernate-core/src/main/java/org/hibernate/boot/registry/selector/internal/DefaultDialectSelector.java
beikov: Ok, then I'll change to throwing an exception.

hibernate-core/src/test/java/org/hibernate/orm/test/strategyselectors/DefaultDialectSelectorTest.java
beikov: I'll move the test assertions into that module.

PR #6258, opened 2023-03-16
Reproducer for incorrect function argument domain type

jwgmeligmeyling:
Reproducer that argument type is not properly resolvable from SqmFunction return type resolvers.

Test fails with:

```
Unexpected type ==> expected: <org.hibernate.type.CustomType> but was: <org.hibernate.metamodel.model.domain.internal.BasicTypeImpl>
Expected :org.hibernate.type.CustomType
Actual   :org.hibernate.metamodel.model.domain.internal.BasicTypeImpl
````

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/annotations/type/dynamicparameterized/DynamicParameterizedTypeTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/annotations/type/dynamicparameterized/DynamicParameterizedTypeTest.java
@@ -23,22 +23,49 @@
  */
 package org.hibernate.orm.test.annotations.type.dynamicparameterized;
 
-import java.util.Date;
-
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jwgmeligmeyling** [comment]
A "work around" seems to be to get the entity name and attribute name from the SqmTypedNode and look it up in the Metamodel, although this requires access to the `SessionFactory` which is only accessible through a deprecated method:

```java
typeConfiguration.getSessionFactory().getMappingMetamodel()
									.getEntityDescriptor(((SqmRoot) ((SqmBasicValuedSimplePath) arguments.get(0)).getLhs()).getEntityName())
									.getPropertyType(((SqmBasicValuedSimplePath) arguments.get(0)).getNavigablePath().getLocalName());
							
```

beikov:
It looks like this was already merged and fixed as part of https://github.com/hibernate/hibernate-orm/pull/6780

PR #6262, opened 2023-03-16
HHH-16322 Merge of entities having a collection with orphanRemoval true fails when bytecode enhancement is enabled 

dreab8:
https://hibernate.atlassian.net/browse/HHH-16322


sebersole:
Applied to main and 6.2

PR #6263, opened 2023-03-16
HHH-15108 - tweak to not ignore exceptions in AggregateClassLoader.findClass()

jrenaat:


hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/AggregatedClassLoader.java
```diff
@@ -196,18 +196,21 @@ protected URL findResource(String name) {
 	@Override
 	protected Class<?> findClass(String name) throws ClassNotFoundException {
 		final Iterator<ClassLoader> clIterator = newClassLoaderIterator();
+		Throwable t = new Throwable();
```
yrodiere, line 199: You don't technically need to instantiate that throwable; see suggestion here and below.

```suggestion
		Throwable t = null;
```

hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/AggregatedClassLoader.java
```diff
@@ -196,18 +196,21 @@ protected URL findResource(String name) {
...
 		while ( clIterator.hasNext() ) {
 			final ClassLoader classLoader = clIterator.next();
 			try {
 				return classLoader.loadClass( name );
 			}
-			catch (Exception ignore) {
+			catch (Exception ex) {
+				t.addSuppressed( ex );
```
yrodiere, line 206:
```suggestion
				if ( t == null ) {
					t = ex;
				}
				else {
					t.addSuppressed(ex);
				}
```


Alternatively, if you feel like it, you could also copy-paste a util from Hibernate Search:

https://github.com/hibernate/hibernate-search/blob/f2ccc2370cfecaa8f9257f855f03a840e1a0a05e/util/common/src/main/java/org/hibernate/search/util/common/impl/Throwables.java#L53-L64

`org.hibernate.internal.util.ExceptionHelper` seems like a good place for that in Hibernate ORM.

It makes the code slightly more readable:

```suggestion
				t = ExceptionHelper.combine( t, ex );
```

See for example https://github.com/hibernate/hibernate-search/blob/f2ccc2370cfecaa8f9257f855f03a840e1a0a05e/integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/testsupport/util/LuceneTckBackendAccessor.java#L111-L113

hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/AggregatedClassLoader.java
```diff
@@ -196,18 +196,21 @@ protected URL findResource(String name) {
...
 			}
-			catch (LinkageError ignore) {
+			catch (LinkageError le) {
+				t.addSuppressed( le );
```
yrodiere, line 209:
```suggestion
				if ( t == null ) {
					t = le;
				}
				else {
					t.addSuppressed( le );
				}
```

hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/AggregatedClassLoader.java
```diff
@@ -196,18 +196,21 @@ protected URL findResource(String name) {
...
 			}
 		}
 
-		throw new ClassNotFoundException( "Could not load requested class : " + name );
+		throw new ClassNotFoundException( "Could not load requested class : " + name, t.getSuppressed().length > 0 ? t : null );
```
yrodiere, line 213:
```suggestion
		throw new ClassNotFoundException( "Could not load requested class : " + name, t );
```

yrodiere (approved):
LGTM, though I suggested some potential improvements. Which you can probably call premature optimization... YMMV :)

hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/AggregatedClassLoader.java
yrodiere: > I don't think having a Throwable around for the duration of the method would impact much in terms of memory and such?

It's more the call to `java.lang.Throwable#fillInStackTrace()` that annoys me, but yeah, I have no proof it's a problem, it's quite possibly just cargo cult.

Your PR, your rules :shrug: Feel free to merge :)

yrodiere (approved):

hibernate-core/src/main/java/org/hibernate/boot/registry/classloading/internal/AggregatedClassLoader.java
sebersole: Either way, eagerly creating the exception should be avoided.  We are creating it even if everything goes fine iiuc

jrenaat: Yeah, I agree, Yoann's remark on the call to fillInStackTrace made me reconsider

sebersole (approved):
Looks good!

PR #6268, opened 2023-03-17
HHH-16333 Handle converters properly in BasicPluralType

beikov:
https://hibernate.atlassian.net/browse/HHH-16333

### DIFF
--- a/migration-guide.adoc
+++ b/migration-guide.adoc
@@ -14,21 +14,4 @@ earlier versions, see any other pertinent migration guides as well.
 * link:{docsBase}/6.1/migration-guide/migration-guide.html[6.1 Migration guide]
 * link:{docsBase}/6.0/migration-guide/migration-guide.html[6.0 Migration guide]
 
-[[data-format-changes]]
-== Data format changes
-
-[[data-format-wrapper-byte-array]]
-=== Byte[] data format changes
-
-Prior to Hibernate 6.2, a `Byte[]` in the domain model was always just blindly converted to a `byte[]`,
-possibly running into a `NullPointerException` if the `Byte[]` contains a `null` array element.
-
-To fix this, a `Byte[]` is now serialized with Java serialization to a `byte[]`, which allows handling nulls better.
-This change is necessary to properly support plural basic types where the element type maps to the `Byte` Java type,
-i.e. an array or collection of `Enum`.
-
-Since the use of `Byte[]` in a domain model did not make sense so far because null elements were impossible,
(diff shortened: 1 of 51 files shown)


hibernate-core/src/main/java/org/hibernate/type/ArrayConverter.java
sebersole, line 1: I'd prefer this live with the other converter handling I think

beikov: It's an internal thing. I can make it an inner class if you prefer that.

sebersole (commented):
api/spi discussions aside - lgtm

hibernate-core/src/main/java/org/hibernate/type/ArrayConverter.java
beikov: Right, that is the status-quo. I would like to make it an inner class or move to an internal package though. Do you have a particular package where you'd want to see that class live?

migration-guide.adoc
```diff
@@ -14,21 +14,34 @@ earlier versions, see any other pertinent migration guides as well.
...
+The mappings for the wrapper array types `Byte[]` and `Character[]` have changed.
 
 Prior to Hibernate 6.2, a `Byte[]` in the domain model was always just blindly converted to a `byte[]`,
 possibly running into a `NullPointerException` if the `Byte[]` contains a `null` array element.
+`Character[]` was similarly blindly converted to `char[]`, potentially running into the same `NullPointerException` problem.
 
-To fix this, a `Byte[]` is now serialized with Java serialization to a `byte[]`, which allows handling nulls better.
-This change is necessary to properly support plural basic types where the element type maps to the `Byte` Java type,
-i.e. an array or collection of `Enum`.
+Not allowing `null` elements defeats the purpose of choosing the wrapper array type in the first place,
```
sebersole, line 26: I'm not sure this is true.  Its possible users map these assuming no null elements.  Now, is the wrapper form over the primitive form "correct" there?  In a clean-room, no.  But Hibernate does map existing schemas and models.

sebersole: I guess the concern is really that users have 2 options that are not immediately clear.  They are covered later by 

> A possible migration could involve the following steps in a migration script:

and

> Alternatively, to revert to pre-6.2 behavior ...

Maybe it makes more sense to highlight these as alternative options.  Not sure exactly how.  Maybe as list items?

PR #6271, opened 2023-03-17
Merge org.hibernate.testing.orm.ExceptionHelper into org.hibernate.internal.util.ExceptionHelper

jrenaat:


sebersole:
Hm, I'd have expected the other direction if anything.

PR #6274, opened 2023-03-17
HHH-16313 Throw an appropriate error when a mappedBy property references the wrong entity type

mbladel:
https://hibernate.atlassian.net/browse/HHH-16313

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/internal/BinderHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/internal/BinderHelper.java
@@ -1071,4 +1071,34 @@ public class BinderHelper {
 	public static boolean isDefault(XClass clazz, MetadataBuildingContext context) {
 		return context.getBootstrapContext().getReflectionManager().equals( clazz, void.class );
 	}
+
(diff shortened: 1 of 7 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/internal/BinderHelper.java
```diff
@@ -1071,4 +1071,33 @@ static boolean isCompositeId(XClass entityClass, XProperty idProperty) {
...
+				return;
+			}
+			else {
+				referencedClass = referencedClass.getSuperclass();
+			}
+		}
+		throw new AnnotationException(
+				"Collection '" + qualify( propertyHolder.getPath(), propertyName )
+						+ "' is 'mappedBy' a property named '" + mappedBy
+						+ "' which references the wrong entity type '" + referencedEntityName + "'"
```
beikov, line 1100: The message doesn't fit for ToOne attributes. Would be cool if you could adapt that a bit. Also, would be nice if you could add the expected entity name to the error message.

beikov (commented):
Looks good, just a minor comment on the exception message.

PR #6278, opened 2023-03-20
HHH-16336 Throw an appropriate error when a mappedBy property references the wrong entity type

mbladel:
https://hibernate.atlassian.net/browse/HHH-16336

Modify the `mappedBy` type checks to be more lenient in case of subtypes or entity types which map to the same table.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/internal/BinderHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/internal/BinderHelper.java
@@ -877,7 +877,7 @@ public class BinderHelper {
 			return metadataCollector.getPropertyAnnotatedWithMapsId( mappedClass, isId ? "" : propertyName );
 		}
 	}
-	
+
 	public static Map<String,String> toAliasTableMap(SqlFragmentAlias[] aliases){
 		final Map<String,String> ret = new HashMap<>();
 		for ( SqlFragmentAlias alias : aliases ) {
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/internal/BinderHelper.java
```diff
@@ -1085,13 +1086,14 @@ public static void checkMappedByType(
...
-			if ( referencedClass.getEntityName().equals( referencedEntityName ) ) {
+		final PersistentClass referencedClass = persistentClasses.get( referencedEntityName );
+		PersistentClass ownerClass = propertyHolder.getPersistentClass();
+		while ( ownerClass != null ) {
+			if ( checkReferencedClass( ownerClass, referencedClass ) ) {
 				return;
 			}
 			else {
-				referencedClass = referencedClass.getSuperclass();
+				ownerClass = ownerClass.getSuperclass();
```
beikov, line 1096: Theoretically, a `PersistentClass` could have a `MappedSuperclass` instead which then extends from a `PersistentClass`, so it might be worth using a similar approach as in `org.hibernate.metamodel.internal.MetadataContext#applyGenericEmbeddableProperties` to traverse the super types.

mbladel: Added similar logic to get the superclass.

beikov:
Looks like a temporary issue on your side Scott.

scottmarlow:
> > Looks like a temporary issue on your side Scott.
> 
> > Looks like a temporary issue on your side Scott.
> 
> Yeah, I was trying to build with Java 11, building with Java 17 worked fine! :-)

I cherry-picked this change to the `6.2` branch and it passes the `jpa/core/annotations/mapkeycolumn` Jakarta EE 10 Platform TCK tests!

PR #6282, opened 2023-03-21
Relax scopes for TableBased*Handler

DavideD:
So that Hibernate Reactive can create a custom LocalTemporaryTableStrategy

Fix https://hibernate.atlassian.net/browse/HHH-16346

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/temptable/TableBasedInsertHandler.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/temptable/TableBasedInsertHandler.java
@@ -76,7 +76,7 @@ public class TableBasedInsertHandler implements InsertHandler {
 	private final DomainParameterXref domainParameterXref;
 	private final JdbcParameter sessionUidParameter;
 
-	TableBasedInsertHandler(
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/temptable/TableBasedInsertHandler.java
```diff
@@ -337,6 +337,40 @@ else if ( entityDescriptor.getGenerator() instanceof OptimizableGenerator ) {
 		);
 	}
 
+	// For Hibernate Reactive
```
sebersole, line 340:
```suggestion
	/**
	 * For Hibernate Reactive
	 */
```

So it gets moved with any refactoring

sebersole (approved):
lgtm

hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/temptable/TableBasedInsertHandler.java
DavideD: Thanks,  would it be OK  sense to add an annotation in the future?

PR #6285, opened 2023-03-21
HHH-16248 Problem with batching and inheritance with @ManyToOne mapped in multiple subclasses

mbladel:
https://hibernate.atlassian.net/browse/HHH-16248


dreab8:
LGTM

PR #6288, opened 2023-03-22
HHH-16305 Error when using parameter as both a select function argument and in where clause

mbladel:
https://hibernate.atlassian.net/browse/HHH-16305

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/produce/function/ArgumentTypesValidator.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/produce/function/ArgumentTypesValidator.java
@@ -6,6 +6,10 @@
  */
 package org.hibernate.query.sqm.produce.function;
 
+import java.lang.reflect.Type;
+import java.sql.Types;
+import java.util.List;
+
 import org.hibernate.QueryException;
 import org.hibernate.metamodel.MappingMetamodel;
 import org.hibernate.metamodel.mapping.JdbcMapping;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -5180,6 +5185,10 @@ private void replaceJdbcParametersType(
 						parameters.set( 0, newJdbcParameter );
 						jdbcParameters.getJdbcParameters().remove( jdbcParameter );
 						jdbcParameters.getJdbcParameters().add( newJdbcParameter );
+						for ( SqmParameterInterpretation interpretation : this.sqmInterpretationsBySqmParam.get( sqmParameter ) ) {
+							// We need to clear to make sure we overwrite eventual already resolved expressions
+							interpretation.clearResolvedExpression();
```
beikov, line 5190: I'm kind of surprised this is necessary. What is calling `SqmParameterInterpretation#getResolvedExpression()` so early that this must be reset? Maybe we should try to defer this call instead?

mbladel: For reference, `ArgumentTypesValidator.ParameterDetector` is what is initializing the resolved expression "early" - before we replace the jdbc paramter type.

mbladel: I updated the solution which now checks for `JavaObjectType` instead of walking the tree to detect jdbc parameters.

beikov (approved):

PR #6292, opened 2023-03-23
HHH-16358 Make OneToMany with abstract TABLE_PER_CLASS element work again

beikov:
https://hibernate.atlassian.net/browse/HHH-16358
https://hibernate.atlassian.net/browse/HHH-16359
https://hibernate.atlassian.net/browse/HHH-16372

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
@@ -73,6 +73,7 @@ import org.hibernate.metamodel.mapping.SelectablePath;
 import org.hibernate.metamodel.mapping.VirtualModelPart;
 import org.hibernate.metamodel.model.domain.NavigableRole;
 import org.hibernate.metamodel.spi.RuntimeModelCreationContext;
+import org.hibernate.persister.collection.AbstractCollectionPersister;
 import org.hibernate.persister.collection.CollectionPersister;
 import org.hibernate.persister.collection.QueryableCollection;
 import org.hibernate.persister.collection.SQLLoadableCollection;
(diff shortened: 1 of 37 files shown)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/MappingModelCreationHelper.java
```diff
@@ -762,7 +764,7 @@ private static void interpretPluralAttributeMappingKeyDescriptor(
 
 			final BasicValuedModelPart simpleFkTargetPart = (BasicValuedModelPart) fkTargetPart;
 
-			final String keyTableExpression = getTableIdentifierExpression( bootValueMappingKey.getTable(), creationProcess );
+			final String keyTableExpression = collectionTableName;//getTableIdentifierExpression( bootValueMappingKey.getTable(), creationProcess );
```
beikov, line 767: This is for changing the `containingTableExpression` from e.g. `AbstractTablePerClassEntity` to `(select ... from table1 union all select ... from table2 ...)`, to match what the entity persister reports as its table name. Needed for some assert statements.

hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java
```diff
@@ -363,6 +362,8 @@ public AbstractCollectionPersister(
 		else {
 			elementPersister = null;
 		}
+		// Defer this after the element persister was determined, because it is needed in OneToManyPersister#getTableName()
+		spaces[0] = getTableName();
```
beikov, line 366: This changes the entry for the space 0 from e.g. `AbstractTablePerClassEntity` to `(select ... from table1 union all select ... from table2 ...)` for one-to-many associations, to match what the entity persister reports as its table name. Needed for some assert statements.

Relatedly, wouldn't it be more correct to copy the spaces of the element entity type into the collection spaces? AFAIU, the spaces affect cache invalidation, and in case of an abstract table per class entity element, we wouldn't really invalidate any caches though we certainly should.

PR #6296, opened 2023-03-27
Reproducer for HHH-16378 SQM fails to resolve target type of association defined in mappersuperclass with generics

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16378

See also #6297 for the same test on 5.6, which doesn't fail, showing HHH-16378 is a regression.


beikov:
Fixed via https://github.com/hibernate/hibernate-orm/pull/6311

PR #6298, opened 2023-03-27
HHH-16281 Inconsistent Behaivor of L2 cache between Hibernate 5 and 6

dreab8:
https://hibernate.atlassian.net/browse/HHH-16281


sebersole (commented):
lgtm

PR #6301, opened 2023-03-27
HHH-16255 Avoid confusing logs "Executing import script"

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16255

The first commit is just a cleanup, the actual change is in the second commit.


dreab8:
Instead of logging the message only when the target is the database, I would prefer to log it for all the targets adding the info about the executing target.

beikov:
I think it's fine to log only right before executing.

yrodiere:
So... ok to merge? :)

PR #6309, opened 2023-03-28
HHH-16385 Ensure QueryKey is serializable

clementdenis:
https://hibernate.atlassian.net/browse/HHH-16385

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cache/spi/QueryKey.java
+++ b/hibernate-core/src/main/java/org/hibernate/cache/spi/QueryKey.java
@@ -8,6 +8,7 @@ package org.hibernate.cache.spi;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.Objects;
 import java.util.Set;
 
@@ -65,7 +66,7 @@ public class QueryKey implements Serializable {
 	private final Integer firstRow;
 	private final Integer maxRows;
 	private final String tenantIdentifier;
-	private final Set<String> enabledFilterNames;
+	private final String[] enabledFilterNames;
 
 	/**
 	 * For performance reasons, the hashCode is cached; however, it is marked transient so that it can be
@@ -85,7 +86,8 @@ public class QueryKey implements Serializable {
 		this.firstRow = firstRow;
 		this.maxRows = maxRows;
 		this.tenantIdentifier = tenantIdentifier;
-		this.enabledFilterNames = enabledFilterNames;
+		// ensure enabledFilterNames is serializable
+		this.enabledFilterNames = enabledFilterNames.toArray( String[]::new );
 		this.hashCode = generateHashCode();
 	}
 
@@ -110,7 +112,7 @@ public class QueryKey implements Serializable {
 //		result = 37 * result + ( maxRows==null ? 0 : maxRows );
 		result = 37 * result + ( tenantIdentifier==null ? 0 : tenantIdentifier.hashCode() );
 		result = 37 * result + parameterBindingsMemento.hashCode();
-		result = 37 * result + ( enabledFilterNames == null ? 0 : enabledFilterNames.hashCode() );
+		result = 37 * result + Arrays.hashCode( enabledFilterNames );
 		return result;
 	}
 
@@ -146,7 +148,7 @@ public class QueryKey implements Serializable {
 			return false;
 		}
 
-		if ( ! Objects.equals( enabledFilterNames, that.enabledFilterNames ) ) {
+		if ( ! Arrays.equals( enabledFilterNames, that.enabledFilterNames ) ) {
 			return false;
 		}
 
--- a/hibernate-jcache/src/test/java/org/hibernate/orm/test/jcache/QueryCacheWithFilterTest.java
+++ b/hibernate-jcache/src/test/java/org/hibernate/orm/test/jcache/QueryCacheWithFilterTest.java
@@ -1,0 +1,135 @@
+package org.hibernate.orm.test.jcache;
+
+import java.util.List;
+
+import org.hibernate.annotations.Filter;
+import org.hibernate.annotations.FilterDef;
+import org.hibernate.annotations.ParamDef;
+import org.hibernate.cache.spi.RegionFactory;
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.query.Query;
+import org.hibernate.stat.CacheRegionStatistics;
+import org.hibernate.stat.spi.StatisticsImplementor;
+
+import org.hibernate.testing.orm.junit.DomainModel;
+import org.hibernate.testing.orm.junit.ServiceRegistry;
+import org.hibernate.testing.orm.junit.SessionFactory;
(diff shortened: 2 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/cache/spi/QueryKey.java
```diff
@@ -85,7 +87,8 @@ public QueryKey(
 		this.firstRow = firstRow;
 		this.maxRows = maxRows;
 		this.tenantIdentifier = tenantIdentifier;
-		this.enabledFilterNames = enabledFilterNames;
+		//ensure enabledFilterNames is serializable
+		this.enabledFilterNames = Collections.unmodifiableSet(new HashSet<>(enabledFilterNames));
```
beikov, line 91: Since we have to copy this anyway, I'd prefer to use a `String[]`. I see that it is hard to properly test this, so I won't mandate that you write that.

beikov: A test could be as simple as running query with enabled filters that puts into the cache. Before closing the session, mutate the filters. Then in a new session, call the same query again with the same filters. This should lead to a cache miss with this bug, but should be a cache hit without it.

clementdenis: I switched to String[] as requested.

I also added a test and it behaves as described, but I think it's a bit convoluted:
- I had to put it in the jcache module, because the default cache impl in tests does not fail on non-serializable keys (I did not check, but it's probably just a Map)
- In the jcache module, EHCache won't throw an error when encountering a non-serializable key, because of the default [resilience strategy](https://www.ehcache.org/documentation/3.7/resilience.html), but won't cache the value either
- I'm using Caffeine as the JSR107 implementationin my project, which fails with a serialization exception instead of failing silently without actually caching, and made the bug much more obvious
=> Isn't it possible to configure EHCache (at least for the hibernate tests) to a resilience strategy that would throw an exception (ThrowingResilienceStrategy seems to do exaclty that)

I also changed the base branch to 6.2, let me know if I need to rebase on main.

hibernate-jcache/src/test/java/org/hibernate/orm/test/jcache/QueryCacheWithFilterTest.java
```diff
@@ -0,0 +1,143 @@
...
+		);
+	}
+
+	@Test
+	public void testCachedQueryWithFilter(SessionFactoryScope scope) {
+		executeQueryWithCache( scope, false, 0 );
+		executeQueryWithCache( scope, false, 1 );
+		evictQueryRegion(scope);
+		executeQueryWithCache( scope, true, 0 );
+		executeQueryWithCache( scope, true, 1 );
```
beikov, line 60: To actually provoke the bug you'd need to mutate the enabled filters of a session after a query result was cached. Let's try this:

```java
scope.inTransaction(
	session -> {
		Query<Person> query = session.createQuery(
			"from Person p",
			Person.class
		);
		query.setCacheable( true );

		List<Person> resultList = query.getResultList();
		assertThat( resultList ).hasSize( 1 );

		session.enableFilter( "personName" ).setParameter( "name", "John" );
	}
);
scope.inTransaction(
	session -> {
		Query<Person> query = session.createQuery(
			"from Person p",
			Person.class
		);
		query.setCacheable( true );

		List<Person> resultList = query.getResultList();
		assertThat( resultList ).hasSize( 1 );

		assertThat( getQueryCacheRegionStatistics( session ).getHitCount() ).isEqualTo( 1 );
	}
);
```

clementdenis: Sorry I'm lost here:
- What I'm trying to test is that QueryKey is properly serializable so it can be used as a cache key => if it's not then the query is not cached, there's a miss the second time, my test validates that
- I don't understand what your test is trying to validate: how is mutating the filter relevant in this case?

Actually, I might have found a second bug: I noticed filter parameters are not part of the QueryKey (except for the tenant identifier). So this test will fail:

```java
scope.inTransaction(
		session -> {
			session.enableFilter( "personName" ).setParameter( "name", "John" );
			Query<Person> query = session.createQuery(
					"from Person p",
					Person.class
			);
			query.setCacheable( true );

			List<Person> resultList = query.getResultList();
			assertThat( resultList ).hasSize( 1 );
			session.enableFilter( "personName" ).setParameter( "name", "John" );
		}
);
scope.inTransaction(
		session -> {
			session.enableFilter( "personName" ).setParameter( "name", "Jack" );
			Query<Person> query = session.createQuery(
					"from Person p",
					Person.class
			);
			query.setCacheable( true );

			List<Person> resultList = query.getResultList();
			assertThat( resultList ).hasSize( 0 ); //fails here with 1
		}
);
```

Should I open another bug for this?

beikov: >What I'm trying to test is that QueryKey is properly serializable so it can be used as a cache key => if it's not then the query is not cached, there's a miss the second time, my test validates that
>I don't understand what your test is trying to validate: how is mutating the filter relevant in this case?

Your original fix was to create a copy of the `enabledFilterNames`, and while looking through the code, I saw that this is necessary because the underlying collection can be mutated. Even if `HashMap.KeySet` is not serializable, the problem is deeper than just the serializability concern.

>Should I open another bug for this?

IMO the `parameterBindingsMemento` should actually contain the filter parameter values. Yes, please open a new JIRA ticket for this.

clementdenis: OK, I could finally reproduce the QueryKey mutability issue reliably, but it requires a second filter to be enabled after 
the query, otherwise an emptySet() is passed to the QueryKey the first time. So your test becomes this:
```java
scope.inTransaction(
		session -> {
			session.enableFilter( "personName" )
					.setParameter( "name", "John" );
			Query<Person> query = session.createQuery(
					"from Person p",
					Person.class
			);
			query.setCacheable( true );

			List<Person> resultList = query.getResultList();
			assertThat( resultList ).hasSize( 1 );

			session.enableFilter( "personName2" )
					.setParameter( "name", "John" );
		}
);
scope.inTransaction(
		session -> {
			session.enableFilter( "personName" )
					.setParameter( "name", "John" );
			Query<Person> query = session.createQuery(
					"from Person p",
					Person.class
			);
			query.setCacheable( true );

			List<Person> resultList = query.getResultList();
			assertThat( resultList ).hasSize( 1 );

			assertThat( getQueryCacheRegionStatistics( session ).getHitCount() ).isEqualTo( 1 );
		}
);
```

Also, the problem does not happen when using a jcache-compatible cache provider, as the QueryKey is serialized before comparison. So I've added the test in the hibernate-core module, that uses an in-memory cache backed by a simple ConcurrentMap

I still need to keep my test in the hibernate-cache module to validate the serialization issue.

I've also created HHH-16617 for the filter parameters missing from the cache key.

beikov:
Thanks for the pre-work @clementdenis. @mbladel will add some finishing touches through https://github.com/hibernate/hibernate-orm/pull/6658 and make sure this will be merged soon.

PR #6311, opened 2023-03-28
HHH-16378 SQM fails to resolve target type of association defined in mappersuperclass with generics

mbladel:
https://hibernate.atlassian.net/browse/HHH-16378

This change includes *concrete* type resolution for both associations (that fixes the issue) and basic-valued paths.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/metamodel/internal/AttributeFactory.java
+++ b/hibernate-core/src/main/java/org/hibernate/metamodel/internal/AttributeFactory.java
@@ -23,6 +23,7 @@ import org.hibernate.mapping.Map;
 import org.hibernate.mapping.OneToMany;
 import org.hibernate.mapping.PersistentClass;
 import org.hibernate.mapping.Property;
+import org.hibernate.mapping.SimpleValue;
 import org.hibernate.mapping.Value;
 import org.hibernate.metamodel.AttributeClassification;
 import org.hibernate.metamodel.RepresentationMode;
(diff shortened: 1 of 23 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/tree/domain/AbstractSqmPath.java
```diff
@@ -163,9 +177,11 @@ public SqmExpression<Class<? extends T>> type() {
 	@Override
 	@SuppressWarnings("unchecked")
 	public SqmPath<?> get(String attributeName) {
-
-		final SqmPathSource<?> subNavigable = getReferencedPathSource().getSubPathSource( attributeName );
-
+		final SqmPathSource<?> resolvedModel;
+		final SqmPathSource<?> pathSource = ( resolvedModel = getResolvedModel() ) != null ?
```
beikov, line 181: When does `getResolvedModel()` return `null`? IMO it should never do that.

mbladel: It can happen for `SqmCteRoot` and  `SqmDerivedRoot` since `getModel()` returns null. An alternative would be overriding the `getResolvedModel` for those two but I was not sure what would be appropriate.

Also using `getReferencedPathSource()` instead of `getModel()` when the pathSource is not generic would work, but that would change the semantics of `getResolvedModel` for some cases IMO (e.g. treated paths).

PS: one example of how `@Nullable` would help :smile:

beikov (commented):
Looks good except for the `determineSimpleType` implementation which I believe is not fully correct yet.

hibernate-core/src/main/java/org/hibernate/metamodel/internal/AttributeFactory.java
```diff
@@ -221,7 +227,19 @@ private <Y> DomainType<Y> determineSimpleType(ValueContext typeContext) {
 	public static <Y> DomainType<Y> determineSimpleType(ValueContext typeContext, MetadataContext context) {
 		switch ( typeContext.getValueClassification() ) {
 			case BASIC: {
-				return context.resolveBasicType( typeContext.getJpaBindableType() );
+				Class returnedClass = typeContext.getJpaBindableType();
+				if ( returnedClass.isAssignableFrom( Object.class ) ) {
```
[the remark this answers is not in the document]
mbladel: It should be, but we were not building *concrete* identifier attributes apart from `Composite` types. I reviewed that logic and added a separate test for basic properties/identifiers.

beikov (approved):

PR #6312, opened 2023-03-28
HHH-16390 - Execution of non-batched statements do not force execution of current batch

sebersole:
https://hibernate.atlassian.net/browse/HHH-16390


sebersole:
Applied

PR #6316, opened 2023-03-29
HHH-16261 JPA metamodel generator is not considering Java 14's records

slinstaedt:
Also:
- keep backward compatibility with compilers, not supporting records


sebersole:
Have a look at `hibernate-core/src/test/java17` and how that is handled in `hibernate-core/hibernate-core.gradle`

slinstaedt:
Thanks for the hint. I will do so tomorrow.

mbladel:
Superseded by https://github.com/hibernate/hibernate-orm/pull/6345.

PR #6318, opened 2023-03-29
fix wrong parameter order

awagner73:


beikov:
Thanks for the fix. Fixed this via https://hibernate.atlassian.net/browse/HHH-16404

PR #6319, opened 2023-03-29
HHH-16396 HQL with SubQuery having same alias of root Query generates wrong SQL

dreab8:
https://hibernate.atlassian.net/browse/HHH-16396

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/ast/spi/SimpleFromClauseAccessImpl.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/ast/spi/SimpleFromClauseAccessImpl.java
@@ -9,10 +9,12 @@ package org.hibernate.sql.ast.spi;
 import java.util.HashMap;
 import java.util.Map;
 
-import org.hibernate.metamodel.mapping.EmbeddableValuedModelPart;
+import org.hibernate.metamodel.mapping.ModelPartContainer;
 import org.hibernate.metamodel.model.domain.NavigableRole;
 import org.hibernate.spi.NavigablePath;
 import org.hibernate.sql.ast.SqlTreeCreationLogger;
+import org.hibernate.sql.ast.tree.from.CorrelatedTableGroup;
+import org.hibernate.sql.ast.tree.from.PluralTableGroup;
 import org.hibernate.sql.ast.tree.from.TableGroup;
 import org.hibernate.sql.ast.tree.from.VirtualTableGroup;
 
@@ -53,9 +55,9 @@ public class SimpleFromClauseAccessImpl implements FromClauseAccess {
 	@Override
 	public TableGroup findTableGroupForGetOrCreate(NavigablePath navigablePath) {
 		final TableGroup tableGroup = findTableGroup( navigablePath );
-		if ( parent != null && tableGroup instanceof VirtualTableGroup && tableGroup.getModelPart() instanceof EmbeddableValuedModelPart ) {
+		if ( parent != null && tableGroup != null && navigablePath.getParent() != null ) {
 			final NavigableRole navigableRole = tableGroup.getModelPart().getNavigableRole();
-			if ( navigableRole != null ) {
+			if ( navigableRole != null && navigableRole.getParent() != null ) {
 				// Traverse up the navigable path to the point where resolving the path leads us to a regular TableGroup
 				NavigableRole parentRole = navigableRole.getParent();
 				NavigablePath parentPath = navigablePath.getParent();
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/sql/ast/spi/SimpleFromClauseAccessImpl.java
```diff
@@ -75,6 +75,18 @@ public TableGroup findTableGroupForGetOrCreate(NavigablePath navigablePath) {
 		return tableGroup;
 	}
 
+	private TableGroup locateTableGroup(NavigablePath navigablePath) {
+		final TableGroup tableGroup = tableGroupMap.get( navigablePath );
+		if ( tableGroup != null || parent == null ) {
+			return tableGroup;
+		}
+		final NavigablePath navigablePathParent = navigablePath.getParent();
+		if ( navigablePathParent == null || tableGroupMap.get( navigablePathParent ) == null ) {
```
beikov, line 84: I think this is actually not so easy. We also have to consider grandparents etc.

beikov: Consider a slightly more complicated query like this:

```
select
 e.embeddable.otherEntity.id," +
 (select e.aString from MyEntity e where e.embeddable.otherEntity.id = :anotherEntityId)
from MyEntity e where e.id = :id
```

mbladel: I was thinking we could generalize the check in `findTableGroupForGetOrCreate` which cycles to find the very first parent to verify all table groups and not just embeddables.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/SimpleFromClauseAccessImpl.java
```diff
@@ -52,7 +52,7 @@ public TableGroup findTableGroup(NavigablePath navigablePath) {
 
 	@Override
 	public TableGroup findTableGroupForGetOrCreate(NavigablePath navigablePath) {
-		final TableGroup tableGroup = findTableGroup( navigablePath );
+		final TableGroup tableGroup = locateTableGroup( navigablePath );
 		if ( parent != null && tableGroup instanceof VirtualTableGroup && tableGroup.getModelPart() instanceof EmbeddableValuedModelPart ) {
```
mbladel, line 56:
```suggestion
		if ( parent != null && tableGroup != null ) {
```

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/SimpleFromClauseAccessImpl.java
```diff
@@ -75,6 +75,18 @@ public TableGroup findTableGroupForGetOrCreate(NavigablePath navigablePath) {
...
+		final NavigablePath navigablePathParent = navigablePath.getParent();
+		if ( navigablePathParent == null || tableGroupMap.get( navigablePathParent ) == null ) {
+			return parent.findTableGroup( navigablePath );
+		}
+		return null;
+	}
+
 	private TableGroup getUnderlyingTableGroup(VirtualTableGroup virtualTableGroup) {
 		final TableGroup tableGroup = virtualTableGroup.getUnderlyingTableGroup();
 		if ( tableGroup instanceof VirtualTableGroup ) {
```
mbladel, line 92:
```suggestion
	private TableGroup getUnderlyingTableGroup(TableGroup tableGroup) {
		if ( tableGroup instanceof VirtualTableGroup ) {
			return getUnderlyingTableGroup( ( (VirtualTableGroup) tableGroup ).getUnderlyingTableGroup() );
		}
		return tableGroup;
	}
```

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/SimpleFromClauseAccessImpl.java
beikov: This would be nice, but not sure if the code below works fine with non embeddable paths like in this case. Ideally, we should change the code below though instead of adding this `locateTableGroup` method.

dreab8:
@beikov  added the test you suggested and pairing with @mbladel for a different solution

sebersole:
I'm taking a look at the failures..

sebersole:
These are real... `StackOverflowException` .  I have a solution I am testing atm

sebersole:
Applied.  Thanks guys!

PR #6327, opened 2023-03-30
5.6 pagination parameter generation error

uxfix:
I set page = 1 size =15

But the final generated sql statement does limit 15,15


uxfix:
I think I understand where the problem is, your first page starts from 0

PR #6329, opened 2023-03-30
HHH-16403 Avoid unnecessary wrapping for exceptions/errors thrown by getters/setters

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16403


beikov (approved):

beikov:
Please backport that to 6.2

yrodiere:
@beikov Why? That's an improvement and a behavior change, it doesn't fix a regression in ORM 6.

beikov:
Ok, if you feel like this has the potential for being a breaking change. To me, this is more of a bug fix i.e. we should never wrap errors anywhere

yrodiere:
> Ok, if you feel like this has the potential for being a breaking change.

I just saw code translating any exception coming from Hibernate as a HTTP 400 (bad request) in their Jax-RS layer, so... I'm afraid everything is possible at this point :/

Let's backport if someone actually experiences this problem in a real application.

PR #6331, opened 2023-03-30
HHH-16396 HQL with SubQuery having same alias of root Query generates wrong SQL

sebersole:
https://hibernate.atlassian.net/browse/HHH-16396

Minor fix for the failures in https://github.com/hibernate/hibernate-orm/pull/6319


sebersole:
Applied

PR #6339, opened 2023-03-31
HHH-16414 Improve table group resolution for subqueries using the same alias as parent

mbladel:
https://hibernate.atlassian.net/browse/HHH-16414

PR #6341, opened 2023-04-01
HHH-16413 Added test for issue

aspan:


mbladel:
Thanks for the reproducer, integrated in https://github.com/hibernate/hibernate-orm/pull/6339.

PR #6344, opened 2023-04-03
HHH-16420 - Make CriteriaBuilder.function() supports array result type

quaff:
Fix https://hibernate.atlassian.net/browse/HHH-16420
Fix https://hibernate.atlassian.net/browse/HHH-16419 Partially

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/spi/TypeConfiguration.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/spi/TypeConfiguration.java
@@ -61,13 +61,16 @@ import org.hibernate.resource.beans.internal.FallbackBeanInstanceProducer;
 import org.hibernate.resource.beans.spi.ManagedBean;
 import org.hibernate.resource.beans.spi.ManagedBeanRegistry;
 import org.hibernate.service.ServiceRegistry;
+import org.hibernate.type.BasicArrayType;
 import org.hibernate.type.BasicType;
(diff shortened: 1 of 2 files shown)


quaff:
Maybe it's better to push down the enhancement to `TypeConfiguration::getBasicTypeForJavaType`?

beikov:
This should definitely rather be handled in `org.hibernate.type.BasicTypeRegistry#getRegisteredType` i.e. if the type can not be found in the basic type registry, lookup the `JavaType` first and construct the `BasicType` from that similar to how `org.hibernate.boot.model.process.internal.InferredBasicValueResolver#from` does it for the `reflectedJtd` when the `explicitJdbcType` is null.

quaff:
I've given up changing `BasicTypeRegistry`, it break many tests I can not fix, now the changes applied to `TypeConfiguration::getBasicTypeForJavaType`, feel free to close it if you have better solution.

beikov:
>This should definitely rather be handled in `org.hibernate.type.BasicTypeRegistry#getRegisteredType` i.e. if the type can not be found in the basic type registry, lookup the `JavaType` first and construct the `BasicType` from that similar to how `org.hibernate.boot.model.process.internal.InferredBasicValueResolver#from` does it for the `reflectedJtd` when the `explicitJdbcType` is null.

I gave you pretty exact instructions on how to implement this. If you can't manage to do that, then I'll try to work on this as part of my array function extensions.

PR #6345, opened 2023-04-03
Fix for Jira issue HHH-16261

cigaly:
Do not ignore JDK 14 records

### DIFF
--- a/tooling/metamodel-generator/hibernate-jpamodelgen.gradle
+++ b/tooling/metamodel-generator/hibernate-jpamodelgen.gradle
@@ -43,6 +43,40 @@ compileTestJava {
 }
 
 
+// Tests with records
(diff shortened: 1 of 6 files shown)


tooling/metamodel-generator/hibernate-jpamodelgen.gradle
```diff
@@ -43,6 +43,40 @@ compileTestJava {
 }
 
 
+// Tests with records
+if ( gradle.ext.javaVersions.test.release.asInt() >= 17 && gradle.ext.javaToolchainEnabled ) {
```
beikov, line 47: Can you please add a separate source set like is for the hibernate-core module? I'll approve running the workflow, but I'm pretty sure the run will fail because of this change, since we require compatibility with Java 11

PR #6346, opened 2023-04-03
HHH-16421 - Add PostgresQL native any function

quaff:
https://hibernate.atlassian.net/browse/HHH-16421

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQLDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQLDialect.java
@@ -29,6 +29,7 @@ import org.hibernate.dialect.aggregate.AggregateSupport;
 import org.hibernate.dialect.aggregate.PostgreSQLAggregateSupport;
 import org.hibernate.dialect.function.CommonFunctionFactory;
 import org.hibernate.dialect.function.PostgreSQLMinMaxFunction;
+import org.hibernate.dialect.function.PostgreSQLNativeAnyFunction;
 import org.hibernate.dialect.function.PostgreSQLTruncFunction;
 import org.hibernate.dialect.function.PostgreSQLTruncRoundFunction;
 import org.hibernate.dialect.identity.IdentityColumnSupport;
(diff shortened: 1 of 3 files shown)


gavinking (changes requested):
This change is not correct, as indicated by the failing tests.

The HQL `any()` function is an aggregate function that accepts a boolean value.

It is different to the HQL/SQL `any` qualifier.

quaff:
> This change is not correct, as indicated by the failing tests.
> 
> The HQL `any()` function is an aggregate function that accepts a boolean value.
> 
> It is different to the HQL/SQL `any` qualifier.

How to use PostgresQL native any function? It conflict with HQL any function.

quaff:
@gavinking I've updated this PR to register `native_any` function as compromise.

beikov:
As part of https://github.com/hibernate/hibernate-orm/discussions/5562, we can make sure that the any "function"/quantifier works with arrays, but adding a function is IMO not the way to go. You can do this in your own project if you need that right now, but in Hibernate Core, we strive for a more generally applicable solution.

quaff:
> As part of #5562, we can make sure that the any "function"/quantifier works with arrays, but adding a function is IMO not the way to go. You can do this in your own project if you need that right now, but in Hibernate Core, we strive for a more generally applicable solution.

HQL `any` is alias of `bool_or`, which means only booleans accepted then apply logical OR operator, It's not the same thing as PostgresQL native `any`, the gap is not about arrays, but the semantic conflicts.

beikov:
The function could have multiple meanings depending on the type of the argument, or we could also consider using the `array_contains` function i.e. in HQL `where array_contains(array, column)` renders to SQL as `where any(array) = column`.

quaff:
> The function could have multiple meanings depending on the type of the argument, or we could also consider using the `array_contains` function i.e. in HQL `where array_contains(array, column)` renders to SQL as `where any(array) = column`.

One function have multiple meanings is not a good idea IMO, `array_contains` function is the missing piece.
FYI: PG need `any(array)` be rhs like `column = any(array)`.

gavinking:
@beikov we can close this one, right?

beikov:
We will add the `array_contains` function when working on https://hibernate.atlassian.net/browse/HHH-16780

PR #6349, opened 2023-04-04
HHH-16429 WF SessionFactoryTestCase test fails

dreab8:
https://hibernate.atlassian.net/browse/HHH-16429

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
@@ -850,7 +850,7 @@ public abstract class AbstractEntityInitializer extends AbstractFetchParentAcces
 			);
 		}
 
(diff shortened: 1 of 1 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/AbstractEntityInitializer.java
```diff
@@ -850,8 +850,6 @@ private void initializeEntityInstance(Object toInitialize, RowProcessingState ro
 			);
 		}
 
-		assert concreteDescriptor.getIdentifier( entityInstance, session ) != null;
```
beikov, line 853: Shouldn't this rather be the following?

```suggestion
		assert concreteDescriptor.getIdentifier( toInitialize, session ) != null;
```

dreab8: let me try

PR #6351, opened 2023-04-04
HHH-16349 ORA-00979 not a group by expression when using a polymorphic entity in HQL group by

mbladel:
https://hibernate.atlassian.net/browse/HHH-16349


dreab8:
LGTM

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/EntityValuedPathInterpretation.java
```diff
@@ -318,7 +319,11 @@ public static <T> EntityValuedPathInterpretation<T> from(
 			if ( discriminatorMapping != null ) {
 				discriminatorMapping.forEachSelectable( selectableConsumer );
 			}
-			entityMappingType.forEachSelectable( selectableConsumer );
+			entityMappingType.visitFetchables( fetchable -> {
```
beikov, line 322: Minor nitpick. Could you please use a loop based on `getFetchable(int)` and `getNumberOfFetchables()`

mbladel: Done!

PR #6352, opened 2023-04-04
HHH-16418 - Added test and improvement for issue

jrenaat:


hibernate-core/src/main/java/org/hibernate/internal/CoreMessageLogger.java
```diff
@@ -1824,4 +1824,8 @@ void attemptToAssociateProxyWithTwoOpenSessions(
 			id = 514)
 	void postgreSQLJdbcDriverNotAccessible();
 
+	@Message(value = "The identity select string is undefined for the dialect in use. Check the identity generation configuration, in particular the 'hibernate.jdbc.use_get_generated_keys' setting",
```
beikov, line 1827:
```suggestion
	@Message(value = "Can't retrieve the generated identity value, because the dialect does not support selecting the last generated identity and 'hibernate.jdbc.use_get_generated_keys' was disabled",
```

PR #6356, opened 2023-04-05
HHH-16363 OneToMany object properties not set when EmbeddedId and it holds an OneToOne relationship to itself

mbladel:
https://hibernate.atlassian.net/browse/HHH-16363
https://hibernate.atlassian.net/browse/HHH-16473


beikov:
This will be fixed via HHH-16382 which I have been working on for quite a while now, so I think we should only keep the test.

beikov (approved):

PR #6358, opened 2023-04-05
HHH-16435 - Discriminators are ignored when Filter is active

jrenaat:


beikov (approved):
LGTM. Please also backport to 6.2

PR #6362, opened 2023-04-06
HHH-16436 Fix Numeric Overflow Exception

flashvayne:
If we set type "Integer" for NUMBER(10,0), Numeric Overflow Exception occurs when some data stored in db (eg: 3210987654) is bigger than Integer.MAX_VALUE.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java
@@ -736,12 +736,8 @@ public class OracleDialect extends Dialect {
 					// Don't infer TINYINT or SMALLINT on Oracle, since the
 					// range of values of a NUMBER(3,0) or NUMBER(5,0) just
 					// doesn't really match naturally.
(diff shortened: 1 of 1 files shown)


gavinking:
(And $2^{31}-1$ is a 10-digit number, so this is correct.)

flashvayne:
But some datas like '3210987654' can be saved in column NUMBER(10,0), which is bigger than Integer.MAX_VALUE.

gavinking:
> But some datas like '3210987654' can be saved in column NUMBER(10,0), which is bigger than Integer.MAX_VALUE.

That's correct. But Oracle doesn't give us a way to create a numeric column with exactly 31 binary digits so this is the best we can possibly do. [The contrary option would be to use `NUMERIC(9,0)` to store integers, and then *some `int` values would not fit*. Which is obviously much worse.]

Of course in practice one is *extremely* unlikely to encounter numeric values like `3210987654` in a `NUMBER(10,0)` column, since everyone sensible declares column types with *at least* one decimal digit of overhead. So this "problem" is entirely hypothetical.

PR #6363, opened 2023-04-06
HHH-16433 Fix forced follow on locking with order by

jhyoty:
https://hibernate.atlassian.net/browse/HHH-16433

Simple "SELECT x ... ORDER BY x.id FOR UPDATE" query does not need follow on locking in Oracle. Avoiding follow-on locking is important for correct semantics if this is used with "skip locked" queue implementation.

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/OracleSqlAstTranslator.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/OracleSqlAstTranslator.java
@@ -97,7 +97,7 @@ public class OracleSqlAstTranslator<T extends JdbcOperation> extends SqlAstTrans
 			Boolean followOnLocking) {
 		LockStrategy strategy = super.determineLockingStrategy( querySpec, forUpdateClause, followOnLocking );
 		final boolean followOnLockingDisabled = Boolean.FALSE.equals( followOnLocking );
-		if ( strategy != LockStrategy.FOLLOW_ON && querySpec.hasSortSpecifications() ) {
+		if ( strategy != LockStrategy.FOLLOW_ON && querySpec.hasSortSpecifications() && hasLimit() ) {
 			if ( followOnLockingDisabled ) {
 				throw new IllegalQueryOperationException( "Locking with ORDER BY is not supported" );
 			}
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/dialect/functional/OracleFollowOnLockingTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/dialect/functional/OracleFollowOnLockingTest.java
@@ -230,6 +230,28 @@ public class OracleFollowOnLockingTest extends
 
 
 	@Test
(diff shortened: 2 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/OracleSqlAstTranslator.java
```diff
@@ -97,7 +97,7 @@ protected LockStrategy determineLockingStrategy(
 			Boolean followOnLocking) {
 		LockStrategy strategy = super.determineLockingStrategy( querySpec, forUpdateClause, followOnLocking );
 		final boolean followOnLockingDisabled = Boolean.FALSE.equals( followOnLocking );
-		if ( strategy != LockStrategy.FOLLOW_ON && querySpec.hasSortSpecifications() ) {
+		if ( strategy != LockStrategy.FOLLOW_ON && querySpec.hasSortSpecifications() && hasLimit() ) {
```
beikov, line 100: Did you try removing the whole if block already? Maybe this is not necessary at all.

jhyoty: Yes, removing the test breaks the following tests in org.hibernate.orm.test.dialect.functional.OracleFollowOnLockingTest:
* testPessimisticLockWithMaxResultsAndOrderByThenFollowOnLocking
* testPessimisticLockWithMaxResultsAndOrderByWhileExplicitlyDisablingFollowOnLockingThenFails

Those tests seem valid to me.

beikov: I just played around a bit myself and it seems that Oracle allows to use rownum filtering, so a limit is actually allowed. IMO the if-block that checks for the order by should be removed.

jhyoty: I re-checked, and without the if block the tests fail (using `-Pdb=oracle`, which seems to use 21c) with SQL grammar exception:
`
org.hibernate.exception.SQLGrammarException: JDBC exception executing SQL [select p1_0.id,p1_0.name from Product p1_0 order by p1_0.id fetch first ? rows only for update] [ORA-02014: cannot select FOR UPDATE from view with DISTINCT, GROUP BY, etc.
] 
`
Looks like some further changes would be needed to make Hibernate to fall back to the (pre-12c) rownum filtering in this case.

jhyoty: Noticed that in [org.hibernate.dialect.OracleDialect#useFollowOnLocking](https://github.com/hibernate/hibernate-orm/blob/959858f8bcc702e8a0d8313cc885a7abd6110b7f/hibernate-core/src/main/java/org/hibernate/dialect/OracleDialect.java#L1093) (apparently used when adjusting native queries?), the test checks for hasLimit() with order by when deciding about follow on locking use.

PR #6364, opened 2023-04-06
HHH-16443 Convert SqlStatementLogger into a Service

DavideD:
This way is possible to get the logger without having to initialize the JdbcServices service.

Something that I need for Hibernate Reactive.

Before I clean up the changes, is this acceptable?


gavinking:
Concept looks fine to me, but perhaps we should wait to see what @sebersole thinks?

sebersole:
Meh, if you need it.

DavideD:
> Meh, if you need it.

Good enough for me.
I will revert the changes, if I can come up with a different approach

PR #6365, opened 2023-04-06
HHH-16444 When logging the selected dialect, log the db version too

DavideD:
Fix for https://hibernate.atlassian.net/browse/HHH-16444

This is what the log will look like:
```
INFO SQL dialect [vert.x-worker-thread-0] HHH000400: Using dialect: org.hibernate.dialect.MariaDBDialect, version: 10.11
```


gavinking:
Excellent idea!

PR #6366, opened 2023-04-06
HHH-16444 When logging the selected dialect, log the db version too

DavideD:
Same as #6365 but for `main`

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/dialect/HANADialectTestCase.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/dialect/HANADialectTestCase.java
@@ -29,11 +29,10 @@ import jakarta.persistence.GeneratedValue;
 import jakarta.persistence.GenerationType;
 import jakarta.persistence.Id;
 import jakarta.persistence.Table;
-import org.hamcrest.MatcherAssert;
 
-import static org.hamcrest.core.Is.is;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/dialect/HANADialectTestCase.java
```diff
@@ -53,7 +53,7 @@ public void testSqlGeneratedForIdentityInsertNoColumns() {
 					catch (MappingException e) {
 						MatcherAssert.assertThat(
 								e.getMessage(),
-								is( "The INSERT statement for table [EntityWithIdentity] contains no column, and this is not supported by [" + HANAColumnStoreDialect.class.getName() + "]" )
+								startsWith( "The INSERT statement for table [EntityWithIdentity] contains no column, and this is not supported by [" + HANAColumnStoreDialect.class.getName() )
```
jrenaat, line 56: I think you might be missing the final "]" ?

DavideD: With the new log message the version might change. Isn't it enough to test the beginning of the message?
I guess I can use a regex

jrenaat: Ah, never mind, did not look at what you were actually doing, my bad :facepalm:

PR #6368, opened 2023-04-06
HHH-16409 SQLGrammarException when selecting property of grouped join association

mbladel:
https://hibernate.atlassian.net/browse/HHH-16409

This PR reworks `EntityValuedPath`s expansion when they're used as `group by` expressions.

Other than fixing the related issue, that occurred when selecting only an attribute of a joined path and ordering by the path itself, it should also address expansion for implicit joins (which were previously ignored) by applying the same logic.


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/EntityValuedPathInterpretation.java
```diff
@@ -284,10 +292,28 @@ public static <T> EntityValuedPathInterpretation<T> from(
 			EntityValuedModelPart mapping,
 			EntityValuedModelPart treatedMapping,
 			SqmToSqlAstConverter sqlAstCreationState) {
+		final boolean expandToAllColumns;
+		final Clause currentClause = sqlAstCreationState.getCurrentClauseStack().getCurrent();
+		if ( currentClause == Clause.GROUP || currentClause == Clause.ORDER ) {
+			final SqmQuerySpec<?> querySpec = (SqmQuerySpec<?>) sqlAstCreationState.getCurrentSqmQueryPart();
```
beikov, line 298: I'd prefer if we could pass this as argument to the `from` method instead of adding `getCurrentSqmQueryPart`

beikov (commented):
Other than what I commented, this looks good to me.

beikov (approved):

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/internal/EntityValuedPathInterpretation.java
mbladel: This was needed because this `from` method can get called from different places for e.g. implicit joins, and I didn't want to add a parameter to all methods when it's only used for this specific check.

PR #6374, opened 2023-04-06
fix some slightly misleading javadoc about the @Where annotation

gavinking:
@sebersole please check this


hibernate-core/src/main/java/org/hibernate/annotations/Where.java
```diff
@@ -48,10 +48,11 @@
  * List&lt;Document&gt; documents;
  * </pre>
  * <p>
- * By default, {@code @Where} restrictions declared for an entity are
+ * By default, {@code @Where} restrictions declared for an entity are also
```
sebersole, line 51: "... are also be disabled ..."

gavinking: > By default, {@code @Where} restrictions declared for an entity are also applied when loading associations of that entity type. This behavior can be disabled using the setting....

PR #6375, opened 2023-04-07
HHH-16425 - Can not perform some aggregates over subquery

cigaly:
See [jira issue 16425](https://hibernate.atlassian.net/jira/software/c/projects/HHH/issues/HHH-16425)

With Hibernate 6.2

`select a.name, sum(a.amount) from (select name as name, amount as amount from EntityB) as a group by a.name`

is generating error message.

With Hibernate 6.1 same problem is with min and max


cigaly:
Fixed by [PR 6501](https://github.com/hibernate/hibernate-orm/pull/6501)

PR #6376, opened 2023-04-07
Fix test that caused sporadic CI failures and shorten entity names for ORA11

mbladel:
Should also fix some [timestamp failures on s390x](https://ci.hibernate.org/job/hibernate-orm-pipeline/job/main/1721/testReport/junit/org.hibernate.orm.test.timezones/UTCNormalizedInstantTest/Build___s390x_h2___Test___test_SessionFactoryScope_/).

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/timezones/UTCNormalizedInstantTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/timezones/UTCNormalizedInstantTest.java
@@ -1,92 +1,88 @@
 package org.hibernate.orm.test.timezones;
 
-import jakarta.persistence.Entity;
-import jakarta.persistence.GeneratedValue;
-import jakarta.persistence.Id;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.TimeZone;
+
 import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.dialect.Dialect;
 import org.hibernate.dialect.SybaseDialect;
+import org.hibernate.type.descriptor.DateTimeUtils;
+
 import org.hibernate.testing.orm.junit.DomainModel;
 import org.hibernate.testing.orm.junit.SessionFactory;
 import org.hibernate.testing.orm.junit.SessionFactoryScope;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/timezones/UTCNormalizedInstantTest.java
```diff
@@ -34,8 +37,8 @@ public class UTCNormalizedInstantTest {
 		});
 		scope.inSession( s-> {
 			Zoned z = s.find(Zoned.class, id);
-			assertEquals( instant, z.utcInstant );
-			assertEquals( instant, z.localInstant );
+			assertThat( z.utcInstant ).isCloseTo( instant, within( 1, ChronoUnit.MILLIS ) );
```
beikov, line 40: Is this change for Sybase? If so, I think it would be nice to try using this instead `long millis = System.currentTimeMillis(); Instant instant = Instant.ofEpochMilli( millis - millis % 3 );` and add a comment that Sybase has a precision of 1/3 milliseconds.

beikov (approved):

hibernate-core/src/test/java/org/hibernate/orm/test/timezones/UTCNormalizedInstantTest.java
mbladel: No this was failing on the pipeline with h2 on s390x, see [this example](https://ci.hibernate.org/job/hibernate-orm-pipeline/job/main/1740/testReport/junit/org.hibernate.orm.test.timezones/UTCNormalizedInstantTest/Build___s390x_h2___Test___test_SessionFactoryScope_/).

I noticed the extracted instant was always rounded to milliseconds, so that is what I tried to check. Don't know if there's a better way to handle this.

mbladel: Thanks for the pointer, I've updated the test with the rounding logic and removed the duplicated tests just for Sybase since they were no longer needed

beikov (approved):

PR #6378, opened 2023-04-07
HHH-16347 The ORDER BY clause in a Window Function doesn't work when using multiple columns

mbladel:
https://hibernate.atlassian.net/browse/HHH-16347

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
@@ -1414,7 +1414,10 @@ public class SemanticQueryBuilder<R> extends HqlParserBaseVisitor<Object> implem
 		return expressions;
 	}
 
-	private SqmExpression<?> resolveOrderByOrGroupByExpression(ParseTree child, boolean definedCollate) {
+	private SqmExpression<?> resolveOrderByOrGroupByExpression(
+			ParseTree child,
+			boolean definedCollate,
+			boolean allowPositionalOrAliases) {
 		final SqmCreationProcessingState processingState = getCurrentProcessingState();
 		final SqmQuery<?> processingQuery = processingState.getProcessingQuery();
 		final SqmQueryPart<?> queryPart;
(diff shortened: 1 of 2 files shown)


beikov (commented):
I actually think that we should rather disable the alias and positional support for order by clauses of window functions and ordered set aggregates, because I think that it might be confusing if we "support" this as the alias/position is used in a nested context.

beikov:
The problematic query is much simple: `select id as id, dense_rank() over (order by id) from EntityA a`
The problem is that we interpret the text `id` in `SemanticQueryBuilder#resolveOrderByOrGroupByExpression` as alias, which is wrong and problematic. In fact, even the query `select id as id, dense_rank() over (order by 1) from EntityA a` would result in the same problematic interpretation.

The reason why I think this is problematic is that there is no way to do forward references like e.g. `select dense_rank() over (order by id), id as id from EntityA a`
The order by in the dense_rank of this query would refer to the attribute `id` of `a`, whereas the previous query would refer to the select item named `id`.

IMO we should simply skip the resolving of the terminal text in `SemanticQueryBuilder#resolveOrderByOrGroupByExpression` when we are in the OVER clause. If you look into how we handle the partition by clause (the group by equivalent for window functions) right now, you'll see that we don't support aliases there. The "support" of aliases in the order by clause of a window function and ordered-set aggregate function is just wrong.

I guess it's best to move the logic of `visitOrderByClause` into a new overloaded method that additionally receives a flag `allowPositionalOrAliases`, which passes this flag through to a new overloaded method `visitSortSpecification`. In `visitSortSpecification`, based on the flag, you call `visitSortExpression` or `ctx.getChild( 0 ).accept( this )` directly.

hibernate-core/src/main/java/org/hibernate/query/hql/internal/SemanticQueryBuilder.java
```diff
@@ -1629,7 +1651,11 @@ public SqmSortSpecification visitSortSpecification(HqlParser.SortSpecificationCo
 
 	@Override
 	public SqmExpression<?> visitSortExpression(HqlParser.SortExpressionContext ctx) {
-		return resolveOrderByOrGroupByExpression( ctx.getChild( 0 ), ctx.getChildCount() > 1 );
+		return visitSortExpression( ctx, true );
+	}
+
+	public SqmExpression<?> visitSortExpression(HqlParser.SortExpressionContext ctx, boolean allowPositionalOrAliases) {
+		return resolveOrderByOrGroupByExpression( ctx.getChild( 0 ), ctx.getChildCount() > 1, allowPositionalOrAliases );
```
mbladel, line 1658: I updated the PR with @beikov's suggestions, disallowing alias/positional based order by in OVER and WITHIN GROUP clauses.

Impl. note: I had to pass the flag down to `visitSortExpression` because calling `ctx.getChild( 0 ).accept( this )` was not enough to resolve the order-by expression. The logic is now managed in `resolveOrderByOrGroupByExpression`.

beikov (approved):

PR #6381, opened 2023-04-07
HHH-16163 replace @OrderBy and @Where by @SQLOrder and @SQLRestriction

gavinking:
See https://hibernate.atlassian.net/browse/HHH-16163


gavinking:
I'll think I'll merge this, since I believe we all agreed on the basic idea here. If anyone dislikes the names `SQLOrder` and `SQLRestriction`, we can easily rename them, just speak up on HHH-16163.

PR #6389, opened 2023-04-11
HHH-16397 Wrong SQL when HQL has subquery using FK attribute of parent query root

mbladel:
https://hibernate.atlassian.net/browse/HHH-16397


beikov (approved):
This looks so much cleaner now. Great work!

PR #6393, opened 2023-04-11
HHH-16438 - Add test for issue

jrenaat:
Left outer joins do not work with single table polymorphic entities because of the discriminator in the where clause

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -3166,6 +3166,7 @@ public abstract class BaseSqmToSqlAstConverter<T extends Statement> extends Base
 	}
 
 	private TableGroup consumeEntityJoin(SqmEntityJoin<?> sqmJoin, TableGroup lhsTableGroup, boolean transitive) {
+		final List<Predicate> predicates = new ArrayList<>();
 		final EntityPersister entityDescriptor = resolveEntityPersister( sqmJoin.getReferencedPathSource() );
 
 		final SqlAstJoinType correspondingSqlJoinType = sqmJoin.getSqmJoinType().getCorrespondingSqlJoinType();
@@ -3174,16 +3175,26 @@ public abstract class BaseSqmToSqlAstConverter<T extends Statement> extends Base
 				sqmJoin.getNavigablePath(),
 				sqmJoin.getExplicitAlias(),
 				null,
-				() -> (predicate) -> additionalRestrictions = combinePredicates( additionalRestrictions, predicate ),
+				() -> (predicate) -> {
+					switch ( correspondingSqlJoinType ) {
+						case INNER:
+						case LEFT:
+						case FULL:
+							predicates.add( predicate );
+							break;
+						default:
+							additionalRestrictions = combinePredicates( additionalRestrictions, predicate );
+					}
+				},
 				this
-		);
+				);
 		getFromClauseIndex().register( sqmJoin, tableGroup );
 
 		final TableGroupJoin tableGroupJoin = new TableGroupJoin(
 				sqmJoin.getNavigablePath(),
 				correspondingSqlJoinType,
 				tableGroup,
-				null
+				predicates.size() == 1 ? predicates.get(0) : null
 		);
 
 		// add any additional join restrictions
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3166,24 +3166,33 @@ private TableGroup consumeCrossJoin(SqmCrossJoin<?> sqmJoin, TableGroup lhsTable
...
 				null,
-				() -> (predicate) -> additionalRestrictions = combinePredicates( additionalRestrictions, predicate ),
-				this
-		);
+				() -> (predicate) -> {
+					if ( canUseInnerJoins || correspondingSqlJoinType == SqlAstJoinType.RIGHT ) {
+						additionalRestrictions = combinePredicates( additionalRestrictions, predicate );
+					}
+					else {
+						predicates.add( predicate );
```
jrenaat, line 3184: @beikov you said this should be done for everything but cross joins, but i rather suspect that it should only be done for left joins (not sure about the full join)

hibernate-core/src/test/java/org/hibernate/orm/test/join/JoinWithSingleTableInheritanceTest.java
```diff
@@ -0,0 +1,256 @@
...
+					r3.setSomeOtherId( 21 );
+					s.persist( r3 );
+					RootOne r4 = new RootOne(4);
+					s.persist( r4 );
+				}
+		);
+
+		scope.inTransaction(
+				s -> {
+					List<Object[]> l = s.createSelectionQuery( "select r.id, r.someOtherId from RootOne r full join ChildEntityA ce order by ce.id, r.id", Object[].class ).list();
```
beikov, line 198: You need a join predicate for full joins

hibernate-core/src/test/java/org/hibernate/orm/test/join/JoinWithSingleTableInheritanceTest.java
```diff
@@ -0,0 +1,256 @@
...
+					s.persist( r2 );
+					RootOne r3 = new RootOne(3);
+					r3.setSomeOtherId( 21 );
+					s.persist( r3 );
+				}
+		);
+
+		scope.inTransaction(
+				s -> {
+					List<Object[]> l = s.createSelectionQuery( "select r.id, r.someOtherId from RootOne r join ChildEntityA", Object[].class ).list();
```
beikov, line 125: A cross join needs to be spelled out i.e. `RootOne r cross join ChildEntityA a`. Usually, only a cross join may omit the `on` clause, so this should IMO actually fail with a `SemanticQueryException`, thought that's a different topic for which you can start a discussion and create a JIRA issue. In this test, you don't need to test cross joins. I figured that `consumeEntityJoin` is never called with `SqlAstJoinType.CROSS`

jrenaat: Indeed I saw that for those tests it seemed to always be 'inner'

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3174,16 +3175,26 @@ private TableGroup consumeEntityJoin(SqmEntityJoin<?> sqmJoin, TableGroup lhsTab
...
+					switch ( correspondingSqlJoinType ) {
+						case INNER:
+						case LEFT:
+						case FULL:
+							predicates.add( predicate );
+							break;
+						default:
+							additionalRestrictions = combinePredicates( additionalRestrictions, predicate );
+					}
+				},
```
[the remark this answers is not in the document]
jrenaat: This makes the right join break though, i forgot that when i committed the suggestion; to return the correct result, in the right join the discriminator condition needs to be in the where clause

hibernate-core/src/test/java/org/hibernate/orm/test/join/JoinWithSingleTableInheritanceTest.java
jrenaat: I added 1=1, but keeps failing on postgresql with that weird error, so i excluded the full join test for it.

beikov: Why don't you use the same predicate that you use for other join tests? PG works fine with sane queries ;)

hibernate-core/src/test/java/org/hibernate/orm/test/join/JoinWithSingleTableInheritanceTest.java
beikov: A query with an entity join that is lacking an `on` clause, like your example, should fail with a `SemanticQueryException`, because it is only legal to omit the `on` clause for cross joins.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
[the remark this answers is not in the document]
beikov: Well, maybe your expectation/assertion was wrong 😅
The fix as I suggested it should be correct

PR #6394, opened 2023-04-12
HHH-16458 JDBC statement is not closed when DeferredResultSetAccess fails to execute a query

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16458

Backported to branch 6.2 as #6396.


yrodiere:
Thanks, yes I will have a look, I've been absorbed by other things unfortunately.

yrodiere:
Should be good now, let's hope this didn't break the test for other DBs... :crossed_fingers:

PR #6397, opened 2023-04-12
HHH-16366 UnknownTableReferenceException when selecting entity using IdClass

dreab8:
https://hibernate.atlassian.net/browse/HHH-16366


beikov:
The tests were integrated as part of https://github.com/hibernate/hibernate-orm/pull/6435

PR #6398, opened 2023-04-12
Reproducer for HHH-16459 Bytecode-enhanced inline dirty tracking ignores generic associations from mapped superclasses

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16459

@dreab8 do you think you could have a look when you have some time? The reproducer is in `ToOneInheritedAssociationWithGenericsTest`.


dreab8:
Hi @yrodiere ,
thanks for the reproducer I'll give a look as soon as I finished with an issue I'm working on with Marco

mbladel:
Superseded by https://github.com/hibernate/hibernate-orm/pull/6404.

PR #6399, opened 2023-04-12
HHH-16460 Upgrade to Jandex 3.1.0

Sanne:
https://hibernate.atlassian.net/browse/HHH-16460


Sanne:
switched to jandex 3.1.1

PR #6400, opened 2023-04-12
Reproducer for HHH-16461 @Version + session.refresh(entity, LockMode.PESSIMISTIC_WRITE) leads to StaleObjectStateException

yrodiere:

See #6401 for a backport of the reproducer to 5.6 (which does not fail).


beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/6406

PR #6401, opened 2023-04-12
Backport to 5.6 of reproducer for HHH-16461 Test @Version + session.refresh(entity, LockMode.PESSIMISTIC_WRITE) leads to StaleObjectStateException

yrodiere:

Backport of the reproducer from #6400 to branch 5.6.

The test does not fail, proving this is indeed a regression in 6.x.


yrodiere:
Fixed in https://github.com/hibernate/hibernate-orm/pull/6406

PR #6404, opened 2023-04-12
HHH-16459 Bytecode-enhanced inline dirty tracking ignores generic associations from mapped superclasses

mbladel:
https://hibernate.atlassian.net/browse/HHH-16459

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/bytecode/enhancement/dirty/DirtyTrackingInheritanceWithGenericsTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/bytecode/enhancement/dirty/DirtyTrackingInheritanceWithGenericsTest.java
@@ -1,0 +1,114 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.orm.test.bytecode.enhancement.dirty;
+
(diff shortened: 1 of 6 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/bytecode/enhancement/dirty/DirtyTrackingInheritanceWithGenericsTest.java
```diff
@@ -0,0 +1,114 @@
...
+			return id;
+		}
+
+		public void setId(Long id) {
+			this.id = id;
+		}
+	}
+
+	@Entity( name = "ChildItemOne" )
+	public static class ChildItemTwo extends Item<Integer, OtherTwo> {
```
yrodiere, line 101: Both entities have the same name? Not that it matters in this test, but still...

```suggestion
	@Entity( name = "ChildItemOne" )
	public static class ChildItemOne extends Item<String, OtherOne> {
		@Id
		@GeneratedValue
		private Long id;

		public Long getId() {
			return id;
		}

		public void setId(Long id) {
			this.id = id;
		}
	}

	@Entity( name = "ChildItemTwo" )
	public static class ChildItemTwo extends Item<Integer, OtherTwo> {
```

yrodiere (commented):
Thanks for working on this!

I noticed a minor typo, but I'll let Andrea comment on the fix itself :)

hibernate-core/src/test/java/org/hibernate/orm/test/bytecode/enhancement/dirty/DirtyTrackingInheritanceWithGenericsTest.java
mbladel: Thanks, fixed this just in case!

beikov (approved):
LGTM

PR #6406, opened 2023-04-13
HHH-16461 @Version + session.refresh(entity, LockMode.PESSIMISTIC_WRITE) leads to StaleObjectStateException

dreab8:

I checked and it seems that in 5.6 we execute the version check only for hql queries https://github.com/hibernate/hibernate-orm/blob/5.6/hibernate-core/src/main/java/org/hibernate/loader/Loader.java#L1703

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmJdbcExecutionContextAdapter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmJdbcExecutionContextAdapter.java
@@ -70,4 +70,8 @@ public class SqmJdbcExecutionContextAdapter extends BaseExecutionContext {
 		return true;
 	}
 
+	@Override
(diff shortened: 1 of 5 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmJdbcExecutionContextAdapter.java
```diff
@@ -70,4 +70,8 @@ public boolean hasQueryExecutionToBeAddedToStatistics() {
 		return true;
 	}
 
+	@Override
+	public boolean upgradeLocks() {
+		return true;
```
beikov, line 75: This means that only HQL/Criteria and native queries can upgrade locks. Is that intentional i.e. did we not upgrade locks for `session.find()` before?

beikov (approved):

PR #6415, opened 2023-04-17
HHH-16479 Generic enum in @MappedSuperclass fails with java.lang.IllegalArgumentException: Named type  [...] did not implement BasicType nor UserType

dreab8:
https://hibernate.atlassian.net/browse/HHH-16479

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/boot/model/TypeDefinition.java
+++ b/hibernate-core/src/main/java/org/hibernate/boot/model/TypeDefinition.java
@@ -29,6 +29,7 @@ import org.hibernate.resource.beans.spi.ManagedBean;
 import org.hibernate.resource.beans.spi.ManagedBeanRegistry;
 import org.hibernate.type.BasicType;
 import org.hibernate.type.CustomType;
+import org.hibernate.type.JavaObjectType;
 import org.hibernate.type.SerializableType;
 import org.hibernate.type.Type;
 import org.hibernate.type.descriptor.converter.spi.BasicValueConverter;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/boot/model/TypeDefinition.java
```diff
@@ -221,64 +222,81 @@ public MutabilityPlan getMutabilityPlan() {
...
-				public MutabilityPlan getMutabilityPlan() {
-					// a TypeDefinition does not explicitly provide a MutabilityPlan (yet?)
-					return resolved.isMutable()
-							? getDomainJavaType().getMutabilityPlan()
-							: ImmutableMutabilityPlan.instance();
-				}
-			};
+		if ( indicators.getEnumeratedType() != null ) {
+			assert typeImplementorClass.isInterface();
+			return createBasicTypeResolution( new JavaObjectType(), typeImplementorClass, indicators, typeConfiguration );
```
mbladel, line 232: I think this assertion might fail if the user doesn't specify an `extends` interface in the type parameter.
For example, in the PR'S test, if the superclass is defined like this:
```
@MappedSuperclass
public static abstract class GenericBaseEntity<T> {
```
then `typeImplementorClass` might be `java.lang.Object`.

mbladel:
LGTM, only one comment about an assertion.

hibernate-core/src/main/java/org/hibernate/boot/model/TypeDefinition.java
dreab8: @mbladel I added a test, is it correct?

mbladel: Yes, the tests are correct. Turns out what I was worried about is not a problem, since `java.lang.Object` is registered as `BasicType` so it's resolved earlier.

beikov (approved):

PR #6421, opened 2023-04-18
HHH-16392 @Where annotation generates bad query with entity name instead of alias in association subquery

mbladel:
https://hibernate.atlassian.net/browse/HHH-16392

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/SqmMutationStrategyHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/SqmMutationStrategyHelper.java
@@ -6,17 +6,24 @@
  */
 package org.hibernate.query.sqm.mutation.internal;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 
 import org.hibernate.engine.jdbc.spi.JdbcServices;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.metamodel.mapping.EntityMappingType;
 import org.hibernate.metamodel.mapping.PluralAttributeMapping;
 import org.hibernate.metamodel.mapping.internal.EmbeddedAttributeMapping;
+import org.hibernate.persister.internal.SqlFragmentPredicate;
 import org.hibernate.sql.ast.tree.delete.DeleteStatement;
 import org.hibernate.sql.ast.tree.from.NamedTableReference;
+import org.hibernate.sql.ast.tree.from.TableGroup;
 import org.hibernate.sql.ast.tree.from.TableReference;
+import org.hibernate.sql.ast.tree.predicate.FilterPredicate;
+import org.hibernate.sql.ast.tree.predicate.Junction;
 import org.hibernate.sql.ast.tree.predicate.Predicate;
 import org.hibernate.sql.exec.spi.ExecutionContext;
 import org.hibernate.sql.exec.spi.JdbcParameterBindings;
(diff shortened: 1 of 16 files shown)


hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -6017,19 +6018,24 @@ public void visitTableReferenceJoin(TableReferenceJoin tableReferenceJoin) {
...
 		final String dmlTargetTableAlias = getDmlTargetTableAlias();
 		if ( dmlTargetTableAlias != null && dmlTargetTableAlias.equals( columnReference.getQualifier() ) ) {
-			// todo (6.0) : use the Dialect to determine how to handle column references
-			//		- specifically should they use the table-alias, the table-expression
-			//			or neither for its qualifier
-
-			final String tableExpression = getCurrentDmlStatement().getTargetTable().getTableExpression();
-			// Qualify the column reference with the table expression only in subqueries
-			final boolean qualifyColumn = !queryPartStack.isEmpty();
+			final ColumnQualifierSupport qualifierSupport = getDialect().getMinimumColumnQualifierSupport();
```
mbladel, line 6021: The changes in the third commit with the introduction of `ColumnQualifierSupport` were mainly needed for Oracle, since with that dialect table aliases are always rendered in the from clause (even for update and delete statements) and we were qualifying the `@Where` clause columns with the table name, which caused an unknown column error.

A simpler `Dialect.supportsUnqualifiedColumns()`, returning `true` by default and `false` for Oracle, could've been enough to solve this but since I noticed this `todo (6.0)` I tried implementing the full feature. This way, we could support dialects *requiring* explicit table expression, even though right now none of the supported ones do AFAIK.

hibernate-core/src/main/java/org/hibernate/sql/ast/tree/delete/DeleteStatement.java
```diff
@@ -27,10 +27,12 @@ public class DeleteStatement extends AbstractMutationStatement {
 
 	public static final String DEFAULT_ALIAS = "to_delete_";
 	private final Predicate restriction;
+	private final Predicate subqueryRestriction;
```
beikov, line 30: This should go away and the predicate for the cleanup subquery should be built on demand.

hibernate-core/src/main/java/org/hibernate/dialect/ColumnQualifierSupport.java
```diff
@@ -0,0 +1,32 @@
...
+	 * Qualify the column using the table expression,
+	 * ignoring a possible table alias.
+	 */
+	TABLE_EXPRESSION,
+
+	/**
+	 * Qualify the column using the table alias, whenever available,
+	 * and fallback to the table expression.
+	 */
+	TABLE_ALIAS,
```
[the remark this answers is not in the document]
mbladel: I have done this, but only when visiting delete and update statements. I think we still need to keep the `rendersTableReferenceAlias` method in `renderNamedTableReference` to render aliases for non UPDATE/DELETE clauses.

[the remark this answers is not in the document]
beikov: Ok right, then let's re-model that method maybe to not `return false;` for `UPDATE`/`DELETE`, but instead return whether ` == DmlTargetColumnQualifierSupport.TABLE_ALIAS`

hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/SqmMutationStrategyHelper.java
```diff
@@ -179,4 +186,83 @@ private static void cleanUpCollectionTable(
...
+			SharedSessionContractImplementor session) {
+		if ( predicate instanceof FilterPredicate || predicate instanceof SqlFragmentPredicate ) {
+			return getBaseRestrictions( entityDescriptor, tableGroup, session ).get( 0 );
+		}
+		else if ( predicate instanceof Junction ) {
+			final Junction original = (Junction) predicate;
+			if ( original.getPredicates().size() > 1 ) {
+				final Junction junction = new Junction(
+						original.getNature(),
+						original.getPredicates(),
```
beikov, line 207: I think this should create a copy.

hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/SqmMutationStrategyHelper.java
```diff
@@ -179,4 +186,83 @@ private static void cleanUpCollectionTable(
...
+						original.getExpressionType()
+				);
+				final Predicate secondToLastPredicate = junction.getPredicates().get( junction.getPredicates().size() - 2 );
+				final Predicate lastPredicate = junction.getPredicates().get( junction.getPredicates().size() - 1 );
+				int filterPredicateIndex = -1;
+				int fragmentPredicateIndex = -1;
+				if ( lastPredicate instanceof Junction ) {
```
beikov, line 214: Can you explain that a little? I'm not sure what this is for.

beikov: Add a comment about the case maybe. Also, after copying the predicates list, you'll have to change the code below a bit by adding `predicate = junction;`

hibernate-core/src/main/java/org/hibernate/dialect/ColumnQualifierSupport.java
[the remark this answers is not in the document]
mbladel: To do this we would need access to the `Dialect`, which we don't have since the method is `static`. It's also `public`, so I didn't want to change the signature. Should I create another method (non-static or even 2 params) and deprecate this one?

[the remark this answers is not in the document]
beikov: It's incubating, so no need to keep the old method.

hibernate-core/src/main/java/org/hibernate/query/sqm/mutation/internal/SqmMutationStrategyHelper.java
mbladel: It's needed when there is both an explicit where in the mutation query and e.g. both a filter and a discriminator or a filter and a sql-restriction. Added a comment to clarify this.

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
```diff
@@ -5478,13 +5479,13 @@ protected final void renderTableReferenceIdentificationVariable(TableReference t
 		}
 	}
 
-	public static boolean rendersTableReferenceAlias(Clause clause) {
+	public boolean rendersTableReferenceAlias(Clause clause) {
```
beikov, line 5482: Please also make it protected

beikov (approved):

hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
mbladel: Done :)

PR #6436, opened 2023-04-21
HHH-16500 Avoid unnecessary reflection in StreamDecorator

dreis2211:
See https://hibernate.atlassian.net/browse/HHH-16500


beikov:
Stream decorators should be removed. If you want to contribute, we'd welcome a PR that removes all the classes.

beikov:
You have to search for the JIRA ticket for this, but essentially, a `Stream` has a `close` method which the user must invoke to close underlying resources. That's simply how the JDK specifies this, and we follow the model. Trying to be smart and closing IO after executing terminal operations is not the way to go.

PR #6440, opened 2023-04-23
Add @NonNull for the return type of getReferencedProperty

datavisortedyu:
This PR adds @NonNull for the return type of `getReferencedProperty` method to signify that the return value of the method is not null.


datavisortedyu:
This is my first contribution.
If committer deems this PR valid, I will create HHH- issue.

Thanks

gavinking:
There's no point in adding a single piecemeal `@NotNull` annotation. It should be done everywhere or not at all.

PR #6448, opened 2023-04-24
HHH-16125 PostgreSQL enum types + reimplementation of enum support

gavinking:
First try was #6232.

This time things went *somewhat* more smoothly, except that I ran into a problem with enum types in foreign keys.

I'm now concerned that we'll have to abandon the whole `specialiedTypeDeclaration` thing I added to let converters have a vote on the generated DDL types. In place of that I would probably introduce `SqlTypes.ENUM` and pass the enum type along with the `Size` class. So there would be a real `DdlType` for native enums.


gavinking:
Superseded by #6480.

PR #6456, opened 2023-04-25
HHH-16514  test demonstrating issue with DDL generation

gavinking:
results from PK columns getting sorted, but FK columns not


sebersole:
6468

PR #6467, opened 2023-04-26
HHH-16515 - Add several spi packages to nullness checking

jrenaat:


hibernate-core/src/main/java/org/hibernate/bytecode/spi/BytecodeEnhancementMetadata.java
```diff
@@ -82,7 +84,7 @@ void injectEnhancedEntityAsProxyInterceptor(
 	 */
 	LazyAttributeLoadingInterceptor extractInterceptor(Object entity) throws NotInstrumentedException;
```
beikov, line 85:
```suggestion
	@Nullable LazyAttributeLoadingInterceptor extractInterceptor(Object entity) throws NotInstrumentedException;
```

hibernate-core/src/test/java/org/hibernate/orm/test/multitenancy/schema/TenantResolverConfigurationTest.java
```diff
@@ -35,7 +35,7 @@ public void testConfiguration() throws Exception {
 	}
 
 	private static class TestCurrentTenantIdentifierResolver implements CurrentTenantIdentifierResolver {
-		private String currentTenantIdentifier;
+		private String currentTenantIdentifier = "acme";
```
beikov, line 38: That seems irrelevant.

beikov: That's actually a design question we have to ask ourselves. I think that right now, we allow null return values from the `resolveCurrentTenantIdentifier` method. Anyway, there are other implementations that also potentially return null, so either we fix all of them or none :)
Fixing the implementation could also mean checking for a null value and throwing an exception btw, which I would consider to be more correct.

beikov: Please revert the change

beikov (approved):

PR #6468, opened 2023-04-26
HHH-16514 - Property sorting can lead to incorrect column mappings with derived embeddable keys

sebersole:
https://hibernate.atlassian.net/browse/HHH-16514


beikov (approved):

sebersole:
Integrated upstream

PR #6472, opened 2023-04-27
Correction to allow System.out

jrenaat:


hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java
```diff
@@ -138,9 +139,11 @@ public H2Dialect() {
 		this( MINIMUM_VERSION );
 	}
 
+	@AllowSysOut
 	public static void main(String[] args) {
 		System.out.println( new H2Dialect().getMinimumSupportedVersion() );
 	}
```
mbladel, line 145: I doubt having a `main` in a Dialect class was intended, I think we should actually just remove the method.

mbladel (approved):

PR #6484, opened 2023-05-02
HHH-15726 Fix treat disjunction handling and improve pushdown

beikov:
https://hibernate.atlassian.net/browse/HHH-15726

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
@@ -8,10 +8,10 @@ package org.hibernate.persister.entity;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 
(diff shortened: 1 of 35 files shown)


hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
```diff
@@ -660,12 +661,36 @@ public FilterAliasGenerator getFilterAliasGenerator(String rootAlias) {
...
+			}
+			final EntityPersister persister = mappingMetamodel.getEntityDescriptor( entry.getKey() );
+			// Filtering for abstract entities makes no sense, so ignore that
+			// Also, it makes no sense to filter for any of the super types,
+			// as the query will contain a filter for that already anyway
+			if ( !persister.isAbstract() && !isTypeOrSuperType( persister ) && useKind == EntityNameUse.UseKind.TREAT ) {
+				containsTreatUse = true;
+				break;
+			}
+		}
```
gavinking, line 688: @beikov this code looks a bit strange. The first `continue` skips anything other than `TREAT` and `FILTER`, but than later it only actually does work for `TREAT`, so why not just test for `TREAT` right up front?

PR #6487, opened 2023-05-02
HHH-16469 Hibernate repeatedly issues identical SELECT queries to load an optional one-to-one association 

dreab8:
https://hibernate.atlassian.net/browse/HHH-16469

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/BatchEntitySelectFetchInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/BatchEntitySelectFetchInitializer.java
@@ -40,7 +40,7 @@ public class BatchEntitySelectFetchInitializer extends AbstractBatchEntitySelect
 	protected void registerResolutionListener() {
 		parentAccess.registerResolutionListener( parentInstance -> {
 			final AttributeMapping parentAttribute = getParentEntityAttribute( referencedModelPart.getAttributeName() );
-			if ( parentAttribute != null ) {
+			if ( parentAttribute != null && !firstEntityInitializer.isEntityInitialized() ) {
 				getParentInfos().add( new ParentInfo( parentInstance, parentAttribute.getStateArrayPosition() ) );
 			}
 		} );
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
@@ -31,6 +31,7 @@ import org.hibernate.sql.results.graph.entity.EntityInitializer;
 import org.hibernate.sql.results.graph.entity.EntityLoadingLogging;
 import org.hibernate.sql.results.graph.entity.LoadingEntityEntry;
 import org.hibernate.sql.results.jdbc.spi.RowProcessingState;
+import org.hibernate.sql.results.spi.LoadContexts;
 
 import static org.hibernate.internal.log.LoggingHelper.toLoggableString;
 
@@ -139,10 +140,8 @@ public class EntitySelectFetchInitializer extends AbstractFetchParentAccess impl
 			return;
 		}
 
-		final LoadingEntityEntry existingLoadingEntry = session
-				.getPersistenceContext()
-				.getLoadContexts()
-				.findLoadingEntityEntry( entityKey );
+		final LoadContexts loadContexts = session.getPersistenceContext().getLoadContexts();
+		final LoadingEntityEntry existingLoadingEntry = loadContexts.findLoadingEntityEntry( entityKey );
 
 		if ( existingLoadingEntry != null ) {
 			if ( EntityLoadingLogging.DEBUG_ENABLED ) {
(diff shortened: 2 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/BatchEntitySelectFetchInitializer.java
```diff
@@ -40,7 +40,7 @@ public BatchEntitySelectFetchInitializer(
 	protected void registerResolutionListener() {
 		parentAccess.registerResolutionListener( parentInstance -> {
 			final AttributeMapping parentAttribute = getParentEntityAttribute( referencedModelPart.getAttributeName() );
-			if ( parentAttribute != null ) {
+			if ( parentAttribute != null && !firstEntityInitializer.isEntityInitialized() ) {
```
beikov, line 43: Can you explain the need for this condition please?

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
```diff
@@ -209,7 +215,17 @@ public void initializeInstance(RowProcessingState rowProcessingState) {
 		if ( lazyInitializer != null ) {
 			lazyInitializer.setUnwrap( unwrapProxy );
 		}
+
 		isInitialized = true;
+
+		if ( entityInstance == null ) {
+			rowProcessingState.getJdbcValuesSourceProcessingState()
```
beikov, line 222: Maybe move that up into the if that already checks if `entityInstance` is `null`. Also, a code comment that explains why the `LoadingEntityEntry` is registered would be nice. While we are at it, shouldn't we always register a `LoadingEntityEntry` when `existingLoadingEntry` is null?

beikov (commented):
It would be great if you could make the two tests for e.g. H2 only and assert that the queries match exactly :)

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/BatchEntitySelectFetchInitializer.java
dreab8: if the parent entity is initialized, there is no need to add it to the batch loading.

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/fetch/depth/DepthOneBatchTest.java
```diff
@@ -0,0 +1,388 @@
...
+			this.agencyId = agencyId;
+		}
+
+		@Column(name = "AGENCY_DETAIL")
+		public String getAgencyDetail() {
+			return agencyDetail;
+		}
+
+		public void setAgencyDetail(String agencyDetail) {
+			agencyDetail = agencyDetail;
```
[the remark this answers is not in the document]
beikov: Seems you are missing a `this.`

hibernate-core/src/test/java/org/hibernate/orm/test/mapping/fetch/depth/DepthOneTest.java
```diff
@@ -0,0 +1,382 @@
...
+			this.agencyId = agencyId;
+		}
+
+		@Column(name = "AGENCY_DETAIL")
+		public String getAgencyDetail() {
+			return agencyDetail;
+		}
+
+		public void setAgencyDetail(String agencyDetail) {
+			agencyDetail = agencyDetail;
```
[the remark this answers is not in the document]
beikov: Seems you are missing a `this.`

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/BatchEntitySelectFetchInitializer.java
beikov: Wouldn't we skip running the initializer in this case anyway because the entity would have already been loaded?

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
dreab8: not sure

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
```diff
@@ -193,6 +199,11 @@ public void initializeInstance(RowProcessingState rowProcessingState) {
 			if ( toOneMapping.getNotFoundAction() == NotFoundAction.EXCEPTION ) {
 				throw new FetchNotFoundException( entityName, entityIdentifier );
 			}
+			rowProcessingState.getJdbcValuesSourceProcessingState()
```
beikov, line 202: In `AbstractEntityInitializer` we always register the loading entity entry on resolve, so IMO we should do this for select fetching as well.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
dreab8: I may be wrong but this seems to me the only case where this is really needed

beikov: What if we have two associations referring to the same entity type with batching. Then I think we'd try to load the entities twice since we don't register the loading entity entry.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
dreab8: with `internaLoad()` if the entity is not null we have it in the PC, so no need to register it, and before calling the `internalLoad()` we have no entityInstance to register.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/BatchEntitySelectFetchInitializer.java
dreab8: no, we don't

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
dreab8: if you give me a test and I'll check

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
beikov: Ok, you are right. We always check the PC first before looking for a loading entity entry, so this is fine.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializer.java
beikov: You're right, I checked the code again and this is fine.

beikov (approved):

PR #6489, opened 2023-05-02
HHH-16538 Keep using `javax.validation.ConstraintViolation` as fallback

mkurz:
Fixes https://hibernate.atlassian.net/browse/HHH-16538

Should also be ported to the `main` branch (and maybe to 6.1 and/or 6.0 if they are still maintained?)


sebersole:
Ended up just removing `BV_CHECK_CLASS`.  Thanks for calling our attention to it!

PR #6494, opened 2023-05-02
HHH-16541 - (Jakarta EE 10 Platform TCK) jpa/core/lock/query/Client.java#getResultListTest1 test regression

sebersole:
https://hibernate.atlassian.net/browse/HHH-16541

Unable to reproduce so far outside of the TCK


beikov (commented):
Just some small oversights

beikov:
After debugging the TCK test, I figured the problem was that we forced using follow-on locking because of a `LEFT` `TableJoin` referring to an uninitialized `LazyTableGroup`, which revealed a bug in follow-on locking. I fixed both issues with my PR https://github.com/hibernate/hibernate-orm/pull/6568

PR #6495, opened 2023-05-03
HHH-16298 Failure merging a referenced entity 

dreab8:
https://hibernate.atlassian.net/browse/HHH-16298

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/EntityType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/EntityType.java
@@ -13,19 +13,24 @@ import java.util.Map;
 import org.hibernate.AssertionFailure;
 import org.hibernate.HibernateException;
 import org.hibernate.MappingException;
+import org.hibernate.bytecode.enhance.spi.interceptor.EnhancementAsProxyLazinessInterceptor;
 import org.hibernate.engine.internal.ForeignKeys;
 import org.hibernate.engine.spi.EntityUniqueKey;
 import org.hibernate.engine.spi.Mapping;
 import org.hibernate.engine.spi.PersistenceContext;
+import org.hibernate.engine.spi.PersistentAttributeInterceptor;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.internal.util.ReflectHelper;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.persister.entity.Joinable;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/type/EntityType.java
```diff
@@ -460,14 +465,33 @@ else if ( value == null ) {
...
-			EntityPersister entityPersister = getAssociatedEntityPersister( session.getFactory() );
-			Object propertyValue = entityPersister.getPropertyValue( value, uniqueKeyPropertyName );
+			if ( value instanceof HibernateProxy ) {
+			/*
+				If the value is a Proxy and the property access is field, the value returned by
+			 	`attributeMapping.getAttributeMetadata().getPropertyAccess().getGetter().get( object )`
+			 	is always null except for the id, we need the to use the proxy implementation to
+			 	extract the property value.
+			 */
+				value = ( (HibernateProxy) value ).getHibernateLazyInitializer().getImplementation();
```
mbladel, line 475:
```suggestion
			final LazyInitializer lazyInitializer = extractLazyInitializer( value );
			if ( lazyInitializer != null ) {
				/*
				If the value is a Proxy and the property access is field, the value returned by
			 	`attributeMapping.getAttributeMetadata().getPropertyAccess().getGetter().get( object )`
			 	is always null except for the id, we need the to use the proxy implementation to
			 	extract the property value.
				*/
				value = lazyInitializer.getImplementation();
```

hibernate-core/src/main/java/org/hibernate/type/EntityType.java
```diff
@@ -460,14 +465,33 @@ else if ( value == null ) {
...
+			}
+			else if ( ManagedTypeHelper.isPersistentAttributeInterceptable( value ) ) {
```
mbladel, line 477:
```suggestion
			else if ( isPersistentAttributeInterceptable( value ) ) {
```

mbladel (commented):
Other than some slight refactoring LGTM.

hibernate-core/src/main/java/org/hibernate/type/EntityType.java
mbladel: We usually use `extractLazyInitializer` to handled cases like this.

hibernate-core/src/main/java/org/hibernate/type/EntityType.java
mbladel: Import static `isPersistentAttributeInterceptable` as well.

beikov (approved):

PR #6503, opened 2023-05-04
HHH-16370 Using MapKey on ManyToMany leads to wrong insert SQL

dreab8:
https://hibernate.atlassian.net/browse/HHH-16370


beikov (commented):
It seems to me like the changes in `ManyToManyCollectionPart` might be unnecessary, because the `SelectableMapping` of the `foreignKey` should have the correct `insertable`/`updatable` configs. Is the change in `MapBinder` maybe already enough?

beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/6646

PR #6504, opened 2023-05-04
HHH-16505 Interpretation of "bare" left-join alias reference in HQL

mbladel:
https://hibernate.atlassian.net/browse/HHH-16505

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -235,6 +235,7 @@ import org.hibernate.query.sqm.tree.from.SqmEntityJoin;
 import org.hibernate.query.sqm.tree.from.SqmFrom;
 import org.hibernate.query.sqm.tree.from.SqmFromClause;
 import org.hibernate.query.sqm.tree.from.SqmJoin;
+import org.hibernate.query.sqm.tree.from.SqmQualifiedJoin;
 import org.hibernate.query.sqm.tree.from.SqmRoot;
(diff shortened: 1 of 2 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3719,9 +3728,12 @@ private Expression visitTableGroup(TableGroup tableGroup, SqmFrom<?, ?> path) {
 			if ( inferredEntityMapping == null ) {
 				// When the inferred mapping is null, we try to resolve to the FK by default, which is fine because
 				// expansion to all target columns for select and group by clauses is handled in EntityValuedPathInterpretation
-				if ( entityValuedModelPart instanceof EntityAssociationMapping && ( (EntityAssociationMapping) entityValuedModelPart ).isFkOptimizationAllowed() ) {
+				if ( entityValuedModelPart instanceof EntityAssociationMapping
+						&& ( (EntityAssociationMapping) entityValuedModelPart ).isFkOptimizationAllowed()
+						&& !isLeftJoinWithCondition( path ) ) {
```
beikov, line 3733: Actually, it doesn't matter what the join type is. Even for `inner` it is problematic to use the FK if there are additional join conditions.

beikov (commented):
Other than what I commented on, this looks good to me.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
mbladel: You're right, thanks for the review - updated the PR.

beikov (approved):

PR #6505, opened 2023-05-04
HHH-16551- discriminator column of type CHAR should be of length 1

cigaly:
Test case and patch for jira issue [HHH-16551](https://hibernate.atlassian.net/browse/HHH-16551)

[HHH-16551]: https://hibernate.atlassian.net/browse/HHH-16551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


beikov:
Thanks for the test a fix for this issue!

Did you check if Hibernate 5 also generated a schema with a length of 1? I guess it's ok to default to size 1 in case of `CHAR`, but I would not throw an exception if a greater length is configured, unless Hibernate 5 behaved the same way.
Also, please take some inspiration from `org.hibernate.orm.test.annotations.beanvalidation.DDLTest` for writing the test for this.

cigaly:
It seems that Hibernate 5 is ignoring length attribute, so I've removed check (and exception).

Now test is similar to those in DDLTest class.

beikov:
Thanks!

PR #6507, opened 2023-05-04
HHH-16553 Test reproducing a stackoverflow when loading chained entities

gtoison:
When attempting to load entities chained together I get a stack overflow


gtoison:
The issue was rejected: https://hibernate.atlassian.net/browse/HHH-16553

PR #6510, opened 2023-05-05
HHH-16556 Correct MS SQL 2016 deprecation version

vdotjansen:
Use version 13 as suggestion in the deprecation warning instead of 11 which is version MS SQL 2012


gavinking:
Thanks!

PR #6511, opened 2023-05-05
HHH-16557 Testcase and bugfix proposal

pb00067:
CachedMutableNaturalIdTest.java contains a new method testCreateDeleteRecreate which fails without the fix 
NaturalIdReadWriteAccess.java contains a fix proposal.
I ran the hibernate-core tests and all 12882 tests are passing.

https://hibernate.atlassian.net/browse/HHH-16557


hibernate-core/src/main/java/org/hibernate/cache/spi/support/NaturalIdReadWriteAccess.java
```diff
@@ -74,7 +74,7 @@ public boolean afterInsert(SharedSessionContractImplementor session, Object key,
 		try {
 			writeLock().lock();
 			Lockable item = (Lockable) getStorageAccess().getFromCache( key, session );
-			if ( item == null ) {
+			if ( item == null || item.getValue() != value) {
```
mbladel, line 77: Rather than checking the item's value I think the cache entry should not have been created in the first place. Accessing the cache after the `persist`/`remove` in a single transaction would lead to incorrectly finding an entry.

It would be great if you could check why the first transaction creates it even though the entity is deleted and see if we can avoid it.

pb00067: >>It would be great if you could check why the first transaction creates it even though the entity is deleted 

Thanks for reviewing this. You can find an explanation of what's happening in the [Jira Issue](https://hibernate.atlassian.net/browse/HHH-16557) (second comment).
I don't remember exactly why on inserts the storage in NaturaIdCache were deferred to afterTransactionCompletion.
I only can immagine that in the moment of insert (persist-call) the naturalId fields might not yet be initialized/set to their final values and therefore it was decided to deffer it to afterTransactionCompletion.
Im pretty sure that when we not deffer the insert to afterTransactionCompletion some other test will fail.
I know the proposed solution is not very neat, but I don't know how to do it better.

>>Accessing the cache after the persist/remove in a single transaction would lead to incorrectly finding an entry.
Not exactly. Only internally we will produce a cache hit, resolving the naturalId values to the entities primary key.
Then when hiberante tries to resolve the entity itself by the PK it recognizes that it does not exist,
so from user point of view nothing is broken.

mbladel: The cache item is saved after the transaction completes to make sure we create it only if the insert was successful. You can see this in `NaturalIdResolutionsImpl`, where `NaturalIdDataAccess#afterInsert` is called.

That would be a great starting point to start trying to see if we can avoid the cache item creation in the first place.

pb00067: Well there might be different approaches, each own with its pros and cons

1. Before processing the InsertActionQueue, compare it's elements with the DeleteActionQueue and annihilate creations and deletions for the same entity. 

   Pros: avoids useles cache item creation/eviction not only in NaturalIdCache but also in entity shared cache.  
   Cons: might have performance impact and other side-effects.

2. Deffer also delete of NaturalIdCache eviction to afterDelete (afterTransactionCompletion).

pb00067: @mbladel Overthinkig these 3 alternatives, I think that alternative n.2 is the best, most straight forward approach.

hibernate-core/src/main/java/org/hibernate/engine/spi/NaturalIdResolutions.java
```diff
@@ -68,7 +68,7 @@ void manageSharedResolution(
 	/**
 	 * Removes any cross-reference from the L2 cache
 	 */
-	void removeSharedResolution(Object id, Object naturalId, EntityMappingType entityDescriptor);
+	void removeSharedResolution(Object id, Object naturalId, EntityMappingType entityDescriptor, boolean delayToAfterTransactionCompletion);
```
beikov, line 71: Need to retain backwards compatibility since this is an SPI.

```suggestion
	void removeSharedResolution(Object id, Object naturalId, EntityMappingType entityDescriptor, boolean delayToAfterTransactionCompletion);
	
	default void removeSharedResolution(Object id, Object naturalId, EntityMappingType entityDescriptor, boolean delayToAfterTransactionCompletion) {
		removeSharedResolution( id, naturalId, entityDescriptor );
	}
```

hibernate-core/src/main/java/org/hibernate/engine/internal/NaturalIdResolutionsImpl.java
```diff
@@ -381,7 +381,7 @@ private void manageSharedResolution(
 	}
 
 	@Override
-	public void removeSharedResolution(Object id, Object naturalId, EntityMappingType entityDescriptor) {
+	public void removeSharedResolution(Object id, Object naturalId, EntityMappingType entityDescriptor, boolean delayToAfterTransactionCompletion) {
```
beikov, line 384:
```suggestion
	public void removeSharedResolution(Object id, Object naturalId, EntityMappingType entityDescriptor) {
		removeSharedResolution( id, naturalId, entityDescriptor, false );
	}
	
	public void removeSharedResolution(Object id, Object naturalId, EntityMappingType entityDescriptor, boolean delayToAfterTransactionCompletion) {
```

beikov:
After you update the PR, I will discuss with the rest of the team whether we want this fix just for the `main` branch, or if the bug is critical enough to also port this to 6.2.

beikov:
Please also squash the commits and rebase so that this PR only shows a single commit.

pb00067:
I don;t know hw to do it,
git rebase -i HEAD~2  
show me an almost endless list of commits.

beikov:
First you rebase as is i.e. `git rebase upstream/6.2` and then you can squash commits with `git rebase -i HEAD~2`
``

pb00067:
I have following errore when rebasing 
git rebase origin/6.2
error: The following untracked working tree files would be overwritten by checkout:
        documentation/src/main/asciidoc/integrationguide/Preface.adoc
Please move or remove them before you switch branches.
Aborting
error: could not detach HEAD

Then by removing this file I run into further problems.
I think im faster making a new banch with a new PR

pb00067:
> After you update the PR, I will discuss with the rest of the team whether we want this fix just for the `main` branch, or if the bug is critical enough to also port this to 6.2.

We plan to migrate directly to hibernate6.4, so for me the backport to 6.2 is not important

beikov:
Then please provide a PR against 6.4 with a single commit.

hibernate-core/src/main/java/org/hibernate/engine/spi/NaturalIdResolutions.java
pb00067: @beikov this your suggestion does not compile:
Duplicate method removeSharedResolution(Object, Object, EntityMappingType, boolean) in type NaturalIdResolutions

pb00067:
> Then please provide a PR against 6.4 with a single commit.

https://github.com/hibernate/hibernate-orm/pull/8335

PR #6515, opened 2023-05-05
HHH-16558 Detach removes NaturalIdCrossReference in shared cache too

pb00067:
According documentation detach should only detach/evict the object from persistent contect (first level cache) not from second level cache too.

CachedMutableNaturalIdTest.java: New testmethod which fails without the fix
NaturalIdResolutionsImpl.java: fix, remove unneeded stuff


sebersole (approved):

pb00068:
>>We need to squash and rebase this before integrating it upstream. I can do that (which ultimately means rejecting this PR) or you can do it and force push to your branch. Let me know.

Please do it. It does not matter to me if it needs the PR being rejected.

>> Also, did you open a PR against main? 
No, I did open PR against branch [hibernate:6.2](https://github.com/hibernate/hibernate-orm/tree/6.2) which I think is the current development branch.

>>Or should I just cherry-pick this there?
Sounds good.

PR #6517, opened 2023-05-05
HHH-16472 Treated left join with restriction on entity results in incorrect query

mbladel:
https://hibernate.atlassian.net/browse/HHH-16472

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
@@ -4983,7 +4983,6 @@ public abstract class BaseSqmToSqlAstConverter<T extends Statement> extends Base
 			if ( entityNames.isEmpty() ) {
 				continue;
 			}
-			registerTypeUsage( tableGroup );
 
 			final ModelPartContainer modelPart = tableGroup.getModelPart();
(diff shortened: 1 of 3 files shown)


beikov:
Depends on https://github.com/hibernate/hibernate-orm/pull/6484

mbladel:
> Depends on #6484

Since that change is now merged I've rebased this PR and adapted the solution to work with the new treat handling.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -5000,19 +4999,38 @@ private Predicate consumeConjunctTreatTypeRestrictions(Map<TableGroup, Map<Strin
 					tableGroup,
 					this
 			);
+			// We need to check if this is a treated left or full join, which case we should
+			// allow null discriminator values to maintain correct semantics
+			final TableGroupJoin join = getParentTableGroupJoin( tableGroup );
+			final boolean allowNulls = join != null && ( join.getJoinType() == SqlAstJoinType.LEFT || join.getJoinType() == SqlAstJoinType.FULL );
```
mbladel, line 5005: Updated the solution to allow nulless in the discriminator predicate (i.e. added `or t.disc_col is null`) when the treated table group is a `LEFT` or `FULL` join.

beikov (approved):
LGTM

PR #6519, opened 2023-05-05
HHH-16545 PersistenceUtil.MetadataCache needs to be threadsafe

Sanne:
https://hibernate.atlassian.net/browse/HHH-16545

Backporting this to 5.6

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/jpa/internal/util/PersistenceUtilHelper.java
+++ b/hibernate-core/src/main/java/org/hibernate/jpa/internal/util/PersistenceUtilHelper.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.jpa.internal.util;
 
+import java.io.ObjectStreamException;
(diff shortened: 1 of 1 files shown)


AntonGabov:
Hello @Sanne!

Thanks for the awesome fix, we encountered the bug with not thread-safe access to WeakHashMap on production server :( 

Maybe you know when the new version of hibernate 5.6 is released?

Sanne:
@AntonGabov sorry there currently are no plans for another 5.6 release, the focus of the team is on 6.x+
 - https://github.com/hibernate/hibernate-orm/wiki/Huge-Project,-Small-Team

It's not absolutely excluded that we might do any more 5.x release, but it would be exceptional: please don't don't count on it if you're under time pressure. Feel free to quote me to justify upgrading.

If you need this urgently and can't upgrade yet, I'd suggest to incorporate the patches you need in a custom build.

PR #6522, opened 2023-05-06
HHH-16548 allow Configuration to specify URLs to be scanned

gavinking:
Add new operations to `Configuration` to allow archive scanning.

See https://hibernate.atlassian.net/browse/HHH-16548

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/cfg/Configuration.java
+++ b/hibernate-core/src/main/java/org/hibernate/cfg/Configuration.java
@@ -28,6 +28,7 @@ import org.hibernate.boot.Metadata;
 import org.hibernate.boot.MetadataBuilder;
 import org.hibernate.boot.MetadataSources;
 import org.hibernate.boot.SessionFactoryBuilder;
+import org.hibernate.boot.archive.scan.spi.ScanEnvironment;
 import org.hibernate.boot.internal.ClassmateContext;
 import org.hibernate.boot.jaxb.spi.Binding;
 import org.hibernate.boot.model.FunctionContributor;
@@ -95,7 +96,17 @@ import static java.util.Collections.emptyList;
  *     // set a configuration property
  *     .setProperty(AvailableSettings.DATASOURCE,
  *                  "java:comp/env/jdbc/test")
- *     .getSessionFactory();
+ *     .buildSessionFactory();
+ * </pre>
+ * <p>
+ * Alternatively, a {@code Configuration} may scan archives or directories
+ * for annotated classes and XML mappings.
+ * <pre>
+ * SessionFactory factory = new Configuration()
+ *     .setRootArchiveUrl(new URL("file://...")
+ *     .setProperty(AvailableSettings.DATASOURCE,
+ *                  "java:comp/env/jdbc/test")
+ *     .buildSessionFactory();
  * </pre>
  * <p>
  * In addition, there are convenience methods for adding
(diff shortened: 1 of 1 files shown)


gavinking:
@maxandersen Does this work for you?

maxandersen:
what I do is `java -jar generated-random-name-myscript.jar` ... if grok this change right I'll need to know the location of dynamically created jar to make the scanning happen? 

no built-in way of hibernate to look into classpath? ie. `java.class.path` ?

sebersole:
> what I do is `java -cp generated-random-name-myscript.jar:dep1:dep2` ... if grok this change right I'll need to know the location of dynamically created jar to make the scanning happen?
> 
> no built-in way of hibernate to look into classpath? ie. `java.class.path` ?

It **could**.  But that's a potentially large number of files to scan

gavinking:
> if grok this change right I'll need to know the location of dynamically created jar to make the scanning happen?

@maxandersen The URL can be a directory.

maxandersen:
> > what I do is `java -cp generated-random-name-myscript.jar:dep1:dep2` ... if grok this change right I'll need to know the location of dynamically created jar to make the scanning happen?
> 
> > 
> 
> > no built-in way of hibernate to look into classpath? ie. `java.class.path` ?
> 
> 
> 
> It **could**.  But that's a potentially large number of files to scan

even if it would only scan those locations that has a persistence.xml or some other marker file or manifest entry?

maxandersen:
> That's not how discovery at the JPA level works. If you have "non root" archives, those will not have any markers.

I'm sorry to be being slow here - but isn't persistence.xml used normally to mark archives/classes dir as a location to scan for entities?

maxandersen:
here is the simplest usecase I got (using jbang but same applies to if i used maven or gradle):

```
///usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS org.hibernate:hibernate-core:6.3.0-SNAPSHOT
//DEPS com.h2database:h2:2.1.214

import static java.lang.System.*;

import java.net.MalformedURLException;
import java.net.URL;

import org.hibernate.cfg.AvailableSettings;
import org.hibernate.cfg.Configuration;

import jakarta.persistence.Entity;
import jakarta.persistence.GeneratedValue;
import jakarta.persistence.GenerationType;
import jakarta.persistence.Id;


public class scan {

    @Entity
    static public class MyEntity {
        @Id
        @GeneratedValue(strategy = GenerationType.IDENTITY)
        Long id;
        String name;
        String description;
    }

    public static void main(String... args) throws MalformedURLException {
        var factory = new Configuration()
        .setRootArchiveUrl(new URL(args[0]))
        .setProperty(AvailableSettings.DIALECT, "org.hibernate.dialect.H2Dialect")
        .setProperty(AvailableSettings.URL, "jdbc:h2:mem:test")
        .setProperty(AvailableSettings.HBM2DDL_AUTO, "create-drop")
        .buildSessionFactory();

        var session = factory.openSession();

        var entity = new MyEntity();
        entity.name = "max";
        entity.description = "test";

        session.persist(entity);

        session.close();

        session = factory.openSession();

        session.createQuery("from java.lang.Object", MyEntity.class).list().forEach(out::println);

        out.println("Hello World");
    }
}
```

Here I make it work using:

`jbang scan.java file://`jbang info tools --select applicationJar scan.java``

That jbang info tools command just gets the generated jar or folder.

Its flexible but it just seems quite weird to me, requiring having to pass directory or jar to the app itself.

I grok it can be expensive to scan the full classpath but hey - why not allow for it to get started ?
but if that is considered too easy; scan a class path for jars or dirs with a persistence.xml or some other marker that indicates to Hibernate  - let me go look.

Then no need to pass in self-referencing  directories or jars.

gavinking:
@sebersole @maxandersen so what's the conclusion on this one?

Max: is this actually helpful to you?

Steve: is this something we should just add because it's generally-useful and makes it easier to make use of already-existing functionality?

gavinking:
In principle, we could still do this one, if there's any demand for it. But the problem is solved more elegantly in #10271.

PR #6528, opened 2023-05-08
HHH-16542 - Bad get/is handling with bytecode enhancement

sebersole:
https://hibernate.atlassian.net/browse/HHH-16542


sebersole:
Applied

PR #6531, opened 2023-05-09
Reproducer for HHH-16573 Embeddable element collection with updateable = false leads to NullPointerException

yrodiere:
https://hibernate.atlassian.net/browse/HHH-16573

Creating as draft because this is just a reproducer, not a fix.


mbladel:
Integrated in https://github.com/hibernate/hibernate-orm/pull/6540.

PR #6533, opened 2023-05-09
HHH-16517 H6: Problems with flush when batch is ON (without @DynamicUpdate)

mbladel:
https://hibernate.atlassian.net/browse/HHH-16517


beikov (approved):
Looks ok to me

PR #6535, opened 2023-05-09
HHH-16549 - Fix potential NPE in LoadQueryInfluencers

jrenaat:


hibernate-core/src/main/java/org/hibernate/engine/spi/LoadQueryInfluencers.java
```diff
@@ -52,6 +52,8 @@ public class LoadQueryInfluencers implements Serializable {
 
 	private Boolean readOnly;
 
+	private boolean isMutable;
```
beikov, line 55: Please remove the field. There is no need for it. Just do a null check for `sessionFactory` in `checkMutability`

jrenaat: No problem, I created it to avoid having to do the comparison several times, but I acknowledge it makes little to no difference in this case.

beikov (approved):

PR #6537, opened 2023-05-10
Gradle add test logging exceptionFormat = 'full'

dreab8:
Add more info to the Gradle logging , useful with github failures


beikov (approved):
LGTM

PR #6540, opened 2023-05-10
HHH-16573 Embeddable element collection with updateable = false leads to NullPointerException

dreab8:
https://hibernate.atlassian.net/browse/HHH-16573


beikov (approved):

cigaly:
Unfortunately, this will not fix the problem. Try slightly more complicated test case:

```
	@Test
	public void insert(SessionFactoryScope scope) {
		var id = scope.fromTransaction( session -> {
			var entity = new MyEntity();
			entity.getMyEmbeddables().add( new MyEmbeddable( "first" ) );
			entity.getMyEmbeddables().add( new MyEmbeddable( "third" ) );
			session.persist( entity );
			return entity.getId();
		} );
		scope.inTransaction( session -> {
			MyEntity entity = session.find( MyEntity.class, id );
			assertThat( entity ).isNotNull();
			assertThat( entity.getMyEmbeddables() ).hasSize( 2 );
			entity.getMyEmbeddables().add( 1, new MyEmbeddable( "second" ) );
		} );
		scope.inTransaction( session -> {
			MyEntity entity = session.find( MyEntity.class, id );
			assertThat( entity ).isNotNull();
			List<MyEmbeddable> myEmbeddables = entity.getMyEmbeddables();
			assertThat( myEmbeddables ).hasSize( 3 );
			assertThat( myEmbeddables.get( 0 ).getEmbeddedProperty() ).isEqualTo( "first" );
			assertThat( myEmbeddables.get( 1 ).getEmbeddedProperty() ).isEqualTo( "second" );
			assertThat( myEmbeddables.get( 2 ).getEmbeddedProperty() ).isEqualTo( "third" );
		} );
	}
```

This will in first step add two embeddables to entity, then in second step it will insert thrid embeddable between those two. Expected result should be first-second-third, but actual result is first-third-third.

You can see the reason for this by removing updatable=false:

```
insert into MyEntity (id) values (1);

insert into MyEntity_myEmbeddables (MyEntity_id,myEmbeddables_ORDER,embeddedProperty) values (1,0,'first');

insert into MyEntity_myEmbeddables (MyEntity_id,myEmbeddables_ORDER,embeddedProperty) values (1,1,'third');

select m1_0.id from MyEntity m1_0 where m1_0.id=1;

select m1_0.MyEntity_id, m1_0.myEmbeddables_ORDER, m1_0.embeddedProperty from MyEntity_myEmbeddables m1_0 where m1_0.MyEntity_id=1;

update MyEntity_myEmbeddables set embeddedProperty='second' where MyEntity_id=1 and myEmbeddables_ORDER=1;

insert into MyEntity_myEmbeddables (MyEntity_id,myEmbeddables_ORDER,embeddedProperty) values (1,2,'third'):
```

When embeddable is inserted between two existing, resulting SQL is to update existing second one with new value, and then inserting value of existing second to third position.

Proper behaviour wiil be either to delete (all or changed onyl) and the insert (all or changed only) or to change calue of order column of existing and then insert new.

dreab8:
Hi @cigaly, 

thanks for the test, I'm going to add it and work for a solution.

hibernate-core/src/main/java/org/hibernate/persister/collection/BasicCollectionPersister.java
```diff
@@ -502,7 +526,8 @@ private void applyUpdateRowValues(
 				jdbcValueBindings,
 				null,
 				(valueIndex, bindings, y, jdbcValue, jdbcValueMapping) -> {
```
beikov, line 528: I think it's easier to understand this way.

```suggestion
				hasIndex()
				? (valueIndex, bindings, y, jdbcValue, jdbcValueMapping) -> {
					if ( jdbcValueMapping.isFormula() ) {
						return;
					}
					bindings.bindValue(
							jdbcValue,
							jdbcValueMapping,
							ParameterUsage.SET
					);
				}
				: (valueIndex, bindings, y, jdbcValue, jdbcValueMapping) -> {
					if ( !jdbcValueMapping.isUpdateable() ) {
						return;
					}
					bindings.bindValue(
							jdbcValue,
							jdbcValueMapping,
							ParameterUsage.SET
					);
				},
				session
		);
```

dreab8: done :)

hibernate-core/src/main/java/org/hibernate/persister/collection/BasicCollectionPersister.java
```diff
@@ -501,16 +525,28 @@ private void applyUpdateRowValues(
...
+							}
+							bindings.bindValue(
+									jdbcValue,
+									jdbcValueMapping,
+									ParameterUsage.SET
+							);
+						}
+						:
+						(valueIndex, bindings, y, jdbcValue, jdbcValueMapping) -> {
+							if ( !jdbcValueMapping.isUpdateable() || jdbcValueMapping.isFormula() ) {
```
beikov, line 541: Shouldn't this be enough since formulas are never updatable?

```suggestion
							if ( !jdbcValueMapping.isUpdateable() ) {
```

dreab8: it should but not 100% sure it is. Checking....

dreab8: it seems that at the moment `isFormula` does not always imply `!isUpdatable`.

beikov: It should maybe, wdyt?

dreab8: yes but it seems to me a topic unrelated to this issue, probably better to create a separate Jira

PR #6545, opened 2023-05-10
HHH-16579 - Add constructor for CockroachDialect that parses the version without querying the db

DavideD:
Fix [HHH-16579](https://hibernate.atlassian.net/browse/HHH-16579)

[HHH-16579]: https://hibernate.atlassian.net/browse/HHH-16579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrenaat:
@DavideD is using `DatabaseVersion make(Integer major, Integer minor, Integer micro)` perhaps an option?

PR #6548, opened 2023-05-10
HHH-16578 - Add test and fix for issue

jrenaat:


cigaly:
This is not a bug. Property level is 32 bit integer (java.lang.Integer or primitive int), but corresponding column is tinyint which can accept only 8 bit numbers (see [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/integer-types.html)). After changing type of property level from int to byte validation will pass without problems.

hibernate-core/src/main/java/org/hibernate/tool/schema/internal/ColumnDefinitions.java
```diff
@@ -43,6 +43,21 @@ static boolean hasMatchingType(Column column, ColumnInformation columnInformatio
...
 	}
 
+	private static boolean hasSameColumnType(String metadataColumnTypeString, String columnInformationColumnTypeString) {
+		/*
+		 tokenize the metadata column type string, since that could include more than just the type information, such as e.g. 'unsigned',
+		 and we can't just check if it contains the other one, since that would fail with e.g. tinyint and int
+		 the column information type string, as per InformationExtractorJdbcDatabaseMetaDataImpl.addColumns (L160) should already
+		 only contain a single string
+		*/
+		for (String s : metadataColumnTypeString.split( " " ) ) {
```
beikov, line 53: Just for the record, it seems the problem is that `ColumnInformation#getTypeName` only returns `tinyint` because it already tokenizes the type and throws away the `unsigned` part.
IMO we should rather fix this in `InformationExtractorJdbcDatabaseMetaDataImpl` to not throw away the `unsigned` part.

beikov (approved):

PR #6551, opened 2023-05-11
HHH-16555 Join fetch query in combination with the inheritance pattern not working in hibernate 6

mbladel:
https://hibernate.atlassian.net/browse/HHH-16555


beikov (approved):
LGTM

PR #6554, opened 2023-05-11
Reproducer for HHH-16591 @org.hibernate.annotations.JdbcType does not override the `hibernate.type.preferred_duration_jdbc_type` type

marko-bekhta:
https://hibernate.atlassian.net/browse/HHH-16591

Only a test to reproduce the issue and a small update to the doc to use a correct annotation (I didn't add a suggestion to use `@JdbcType`  since that would require explicitly naming a JdbcType class, which would (most likely) be DB-specific)


sebersole:
https://github.com/hibernate/hibernate-orm/pull/6562 includes this PR plus a proposed fix though it is currently failing.

yrodiere:
Looks like https://github.com/hibernate/hibernate-orm/pull/6562 is passing now

beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/6562

PR #6558, opened 2023-05-12
HHH-16572 - Incorrect enhancement for PROPERTY attributes with mismatched field and method names

sebersole:
https://hibernate.atlassian.net/browse/HHH-16572


sebersole:
This first commit is mostly a rough copy of the `ClassDetails` (and friends) parts of the "annotation source" work.  It is a basis for a better understanding of the domain model when deciding what and how to enhance.  

Without big changes to `EnhancementContext`, we are never going to be able to get this 100% correct.  And really, the "annotation source" work will make this all faster and even more robust.

PR #6562, opened 2023-05-13
HHH-16591 - @org.hibernate.annotations.JdbcType does not override the `hibernate.type.preferred_duration_jdbc_type` type

cigaly:
See Jira issue [HHH-16591](https://hibernate.atlassian.net/browse/HHH-16591)


beikov:
Since @gavinking did these changes, he should chime in here.

gavinking:
> Since @gavinking did these changes, he should chime in here.

I did? Don't really recall that. Which changes precisely?

Anyway I guess I agree that the annotation should override the global setting.

gavinking:
OK I do remember that change but it was a change which only impacted h2 and no other database.

And I *did* update the docs in to reflect that change, in the same PR. Perhaps I missed something.

Anyway it doesn't impact the issue of whether the global setting or the annotation takes precedence. Naively it seems to me like it's the annotation which should "win":

gavinking:
Wait, are you saying that when it was `INTERVAL_SECOND` that it would switch the type to `NUMERIC` if you made the `scale` explicit in `@Column`? How did that work exactly? Seems a sort of weird side-effect. Perhaps I'm misunderstanding.

gavinking:
Oh I see you just mean that the fallback type for `INTERVAL_SECOND` in `NUMERIC`, not `BIGINT`.

gavinking:
> Oh I see you just mean that the fallback type for `INTERVAL_SECOND` in `NUMERIC`, not `BIGINT`.

No, wait, that's not it. The default is still `NUMERIC`, unless you change it via the property setting. So specifying an explicit precision should still work perfectly fine.

Not sure what you're saying here. (But I think you must mean `precision`, not `scale`.)

hibernate-core/src/main/java/org/hibernate/boot/model/process/spi/MetadataBuildingProcess.java
```diff
@@ -732,6 +730,42 @@ private static void adaptToPreferredSqlTypeCode(
...
+				jdbcTypeRegistry.getDescriptor( preferredSqlTypeCodeForDuration ),
+				"duration"
+		);
+		basicTypeRegistry.register(
+				durationType,
+				Duration.class.getSimpleName(),
+				Duration.class.getName()
+		);
+	}
+
```
gavinking, line 768: I don't understand what the difference between these methods is. Looks like they could be one method parameterized by a `Class`, an `int`, and a `String`. 

And then these two and also `adaptToPreferredSqlTypeCodeForInstant()` would go away.

cigaly:
@gavinking Guess that those last changes should do what are talking about

beikov (approved):

beikov:
Thanks @cigaly!

PR #6565, opened 2023-05-14
HHH-16560: Support composite EmbeddedId classes.

stevestorey:
In 6.2.2 (via commit 995aa53fc0bae734324c16bf2d53080aa65bb079), an assertion failure was triggered when an `@Embeddable` ID class also contained another `@Embeddable`. I'm proposing to restore the <=6.2.1 behaviour instead of triggering the assertion failure here.

/cc @beikov as you added the assertion in the first place, and I don't know whether my proposal is in fact sensible or valid, but I'm hoping you can keep my new test to detect regressions ...


beikov:
We'll look into this in the coming weeks. Thanks for the report.

mbladel:
Superseded by https://github.com/hibernate/hibernate-orm/pull/6660.

PR #6569, opened 2023-05-15
HHH-16576 Problem in query involving union on MariaDB and Oracle

mbladel:
https://hibernate.atlassian.net/browse/HHH-16576


beikov (approved):
Looks reasonable. Thanks for the fix!

PR #6578, opened 2023-05-16
HHH-16494 Right or full joining on single tablepolymorphic entities gives incorrect results

mbladel:
https://hibernate.atlassian.net/browse/HHH-16494


hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
```diff
@@ -1407,6 +1416,17 @@ public void pruneForSubclasses(TableGroup tableGroup, Map<String, EntityNameUse>
...
 			tableReferenceJoins.removeIf( join -> !retainedTableReferences.contains( join.getJoinedTableReference() ) );
 		}
+
+		if ( needsTreatDiscriminator ) {
+			// We have to add the discriminator condition to the root table join to make sure
+			// we're only retrieving the correct subtype records from the shared joined table
+			final String discriminatorFragment = getPrunedTableExpression(
+					entityNameUses,
+					metamodel,
+					tableGroup.getPrimaryTableReference().getIdentificationVariable()
```
beikov, line 1426: I don't think this will always be the correct table reference. Isn't the discriminator only on the root table?

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
```diff
@@ -1407,6 +1416,17 @@ public void pruneForSubclasses(TableGroup tableGroup, Map<String, EntityNameUse>
...
+			);
+			tableReferenceJoins.get( 0 ).applyPredicate( new SqlFragmentPredicate( discriminatorFragment ) );
```
beikov, line 1428: Do you know for sure that the first join is always the root table join?

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3304,16 +3304,21 @@ private TableGroup consumeEntityJoin(SqmEntityJoin<?> sqmJoin, TableGroup lhsTab
...
-				() -> p -> predicate.set( combinePredicates( predicate.get(), p ) ),
+				() -> p -> needsTreat.setValue( true ),
 				this
-				);
+		);
 		getFromClauseIndex().register( sqmJoin, tableGroup );
 
+		if ( needsTreat.getValue() ) {
+			// Register new treat to apply the discriminator condition to the table reference itself, see #pruneTableGroupJoins
+			registerEntityNameUsage( tableGroup, EntityNameUse.TREAT, entityDescriptor.getEntityName() );
```
beikov, line 3314: Don't we need something similar for association joins as well? What happens if we do `from EntityA a right join a.association`?

hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java
```diff
@@ -3028,6 +3030,63 @@ else if ( discriminatorValue == NOT_NULL_DISCRIMINATOR ) {
...
+		if ( frag.getValues().size() == discriminatorSQLValues.size() ) {
+			// Nothing to prune if we filter for all subtypes
+			return null;
+		}
+
+		if ( containsNotNull ) {
+			final StringBuilder sb = new StringBuilder();
+			String lhs;
+			if ( isDiscriminatorFormula() ) {
+				lhs = StringHelper.replace( getDiscriminatorFormulaTemplate(), Template.TEMPLATE, "t" );
```
beikov, line 3070:
```suggestion
				lhs = StringHelper.replace( getDiscriminatorFormulaTemplate(), Template.TEMPLATE, alias );
```

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
mbladel: Yes the discriminator condition should always use the root table alias.

I believed `tableGroup.getPrimaryTableReference()` always returned a reference to the root table for joined subclasses, with table reference joins to the subtype-specific tables. Am I wrong about that?

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
mbladel: From what I could understand, the primary table reference is the root table and the joins "add" the joined-subclasses tables - so the first join in the list should always start from the root table.

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
mbladel: FYI `AbstractEntityPersister#createRootTableGroup` when creating the table reference, when `needsDiscriminator()` is true (like in this case) always uses the root table name:
```
final TableReference rootTableReference = new NamedTableReference(
		needsDiscriminator() ? getRootTableName() : getTableName(),
		sqlAliasBase.generateNewAlias()
);
```

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
beikov: There might be a case where there is no table reference join i.e. when pruning all subclass tables. Here a sample query: `select 1 from RootEntity r where r.rootEntityAttribute = 1`

mbladel: Right, I guess this is a problem only if `RootEntity` is explicitly treated When table reference joins are empty we could resort to the subquery-style filter: `select 1 from (select * from RootEntity t where t.disc_col = '1') r where r.rootEntityAttribute = 1`

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
mbladel: I'm not sure, but I can test this. I guess this could be a problem for associations that use a specific inheritance subtype, or even the root type but the query specifies a subtype-specific path thus implicitly treating it (though this second case should might already be covered by the new entity-name uses).

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
mbladel: Why not? If the discriminator condition is applied to the root table it should be fine.

Alternatively, there's always the subquery approach (though it might perform worse).

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
mbladel: I've updated the logic to always substitute the root table expression with a subquery containing the discriminator predicate.

I've also confirmed that the tableGroup's primary table reference might not be the root table, and that the first join is not always the one containing it. Also, I've fixed an issue where for explicitly treated joins we were removiong the root table's join causing an error.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
mbladel: I've added 3 more tests that do the same checks as the entity join ones but with an association. I also updated the logic to add a treat for an attribute join to an inheritance subtype entity to ensure we only retrieve records of that type.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3216,6 +3216,18 @@ private TableGroup consumeAttributeJoin(
...
 		registerPluralTableGroupParts( joinedTableGroup );
+
+		if ( !sqmJoin.hasTreats() && sqmJoin.getNodeType().getSqmPathType() instanceof EntityDomainType ) {
+			final EntityDomainType<?> entityDomainType = (EntityDomainType<?>) sqmJoin.getNodeType().getSqmPathType();
+			final String entityName = entityDomainType.getHibernateEntityName();
+			final EntityPersister entityDescriptor = domainModel.findEntityDescriptor( entityName );
+			if ( entityDescriptor != null && entityDescriptor.getSuperMappingType() != null ) {
+				// This is a non-treated join with a subtype entity, register a TREAT entity name use
+				// to filter only the correctly typed records
+				( (SqmJoin) sqmJoin ).treatAs( entityDomainType );
```
beikov, line 3227: I don't know if changing the SQM tree here is a good idea. Can you please register an entity name use instead?

mbladel: I have, but now we're getting some test failures specific to Derby (looks like these might be related to their optimizer, but the generated SQL queries themselves look fine to me).

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3114,10 +3114,12 @@ private void registerEntityNameUsage(
...
-				actualTableGroup.resolveTableReference(
-						null,
-						subType.getEntityPersister().getMappedTableDetails().getTableName()
-				);
+				if ( useKind == EntityNameUse.UseKind.PROJECTION ) {
+					actualTableGroup.resolveTableReference(
+							null,
+							subType.getEntityPersister().getMappedTableDetails().getTableName()
+					);
+				}
```
mbladel, line 3122: I have changed this to only resolve subtype table references when the use kind is `PROJECTION` to avoid creating "useless" treats. This was causing for example in joined inheritance all subtype tables to be left-joined while none of their columns were needed in the query.

This also got rid of the error triggered just for Derby due to an [optimizer bug](https://issues.apache.org/jira/browse/DERBY-7154).

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3403,16 +3415,21 @@ private TableGroup consumeEntityJoin(SqmEntityJoin<?> sqmJoin, TableGroup lhsTab
 				sqmJoin.getNavigablePath(),
 				sqmJoin.getExplicitAlias(),
 				null,
-				() -> p -> predicate.set( combinePredicates( predicate.get(), p ) ),
+				() -> p -> needsTreat.setValue( true ),
```
beikov, line 3418: I think passing anything other than `null` for the `additionalPredicateCollectorAccess` could become a problem if we are entity joining a multi-table entity, given how the implementation in `AbstractEntityPersister#createRootTableGroup` currently looks like. When `additionalPredicateCollectorAccess` is not `null`, the join predicate for the other tables of the entity will be passed to that collector instead of the `TableReferenceJoin`.

In a query like `select b.superclassAttribute from EntityA a join EntityASub b on a.id = b.id` we might produce wrong SQL because the predicate would be missing for the join from `EntityASub` to it's superclass table.

I don't know a great fix for this yet. Maybe we need to simply ignore `additionalPredicateCollectorAccess` in the table reference join creation. If we can do that, I'd like you to rename that parameter to `discriminatorPredicateCollectorAccess` to make the intent clear.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
beikov: I think the problem with doing this is that we can't determine the correct object type for the table group if we don't join all subclass tables. Not sure if my example here will reproduce the problem that I am thinking about, but let's consider the following query: `select 1 from EntitySuperclass a where treat(a as EntitySubclass).superclassProperty = 1`

This should produce a predicate like `case when a1_1.id is not null then 1 when a1_2.id is not null then 2 ... else 0 end in (2, 3) and a1_0.superclassProperty = 1`. If we don't resolve the subclass table references, we would not get a result if there is a `EntitySubSubclass` with `superclassProperty` having a value of `1`.

Note that this only is a problem for joined inheritance if there is no discriminator column, so maybe you will have to add a condition to that if e.g. `if ( useKind == EntityNameUse.UseKind.PROJECTION || isJoinedWithoutDiscriminator( persister ) ) {`

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3336,6 +3338,16 @@ private TableGroup consumeAttributeJoin(
 			registerEntityNameProjectionUsage( sqmJoin, getActualTableGroup( joinedTableGroup, sqmJoin ) );
 		}
 		registerPathAttributeEntityNameUsage( sqmJoin, ownerTableGroup );
+		if ( !sqmJoin.hasTreats() && sqmJoin.getNodeType().getSqmPathType() instanceof EntityDomainType ) {
+			final EntityDomainType<?> entityDomainType = (EntityDomainType<?>) sqmJoin.getNodeType().getSqmPathType();
+			final String entityName = entityDomainType.getHibernateEntityName();
+			final EntityPersister entityDescriptor = domainModel.findEntityDescriptor( entityName );
```
beikov, line 3344: To avoid doing a map lookup it might be possible to use `(EntityPersister) joinedTableGroup.getModelPart().getPartMappingType()` instead, which should never be null.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3336,6 +3338,16 @@ private TableGroup consumeAttributeJoin(
...
+			if ( entityDescriptor != null && entityDescriptor.getSuperMappingType() != null ) {
+				// This is a non-treated join with a subtype entity, register a TREAT entity name use
```
beikov, line 3346:
```suggestion
				// This is a non-treated join with a superclass, register a TREAT entity name use
```

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3336,6 +3338,16 @@ private TableGroup consumeAttributeJoin(
...
+				// to filter only the correctly typed records
```
beikov, line 3347: Can you also add an explanation to the code comment about why this is necessary?

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
```diff
@@ -1355,15 +1358,22 @@ public void pruneForSubclasses(TableGroup tableGroup, Map<String, EntityNameUse>
...
 			}
+			final String tableName = persister.getTableName();
 			final TableReference mainTableReference = tableGroup.getTableReference(
 					null,
-					persister.getTableName(),
+					tableName,
 					false
 			);
-			if ( mainTableReference == null ) {
-				throw new UnknownTableReferenceException( persister.getTableName(), "Couldn't find table reference" );
```
beikov, line 1364: I guess you removed the exception throwing because of not doing `resolveTableReference` anymore for `EntityNameUse.TREAT`?

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
```diff
@@ -1394,7 +1412,12 @@ public void pruneForSubclasses(TableGroup tableGroup, Map<String, EntityNameUse>
 					tableReferenceJoins.add( join );
 				}
 				else {
-					final String tableExpression = oldJoin.getJoinedTableReference().getTableExpression();
+					final String tableExpression = joinedTableReference.getTableExpression();
+					final int index = indexOf( getSubclassTableNames(), tableExpression );
+					if ( index > -1 && isClassOrSuperclassTable[index] ) {
+						// Always retain joins to superclass tables
```
beikov, line 1418: Please explain why this is necessary. Would be a bummer if we can't remove unused super class table joins anymore.

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
```diff
@@ -1408,6 +1431,44 @@ public void pruneForSubclasses(TableGroup tableGroup, Map<String, EntityNameUse>
 		else {
 			tableReferenceJoins.removeIf( join -> !retainedTableReferences.contains( join.getJoinedTableReference() ) );
 		}
+
+		if ( needsTreatDiscriminator ) {
+			// We have to apply the discriminator condition to the root table reference join
+			boolean applied = applyDiscriminatorPredicate(
```
beikov, line 1437: It seems to me that this must always succeed, as `needsTreatDiscriminator` can only be `true` if `needsDiscriminator()` returns `true` and in `AbstractEntityPersister#createRootTableGroup`, we always create the primary table reference with the `getRootTableName()` if `needsDiscriminator() == true`.

So IMO the whole block in this if should be simplified to

```java
			if ( tableReferenceJoins.isEmpty() ) {
				final String discriminatorPredicate = getPrunedDiscriminatorPredicate( entityNameUses, metamodel, "t" );
				if ( discriminatorPredicate != null ) {
					final NamedTableReference tableReference = (NamedTableReference) tableGroup.getPrimaryTableReference();
					tableReference.setPrunedTableExpression( "(select * from " + getRootTableName() + " t where " + discriminatorPredicate + ")" );
				}
			}
			else {
				final TableReferenceJoin join = tableReferenceJoins.get( 0 );
				assert join.getJoinType() == SqlAstJoinType.INNER : "Found table reference join with root table of non-INNER type: " + join.getJoinType();
				final String discriminatorPredicate = getPrunedDiscriminatorPredicate(
						entityNameUses,
						metamodel,
						tableGroup.getPrimaryTableReference().getIdentificationVariable()
				);
				join.applyPredicate( new SqlFragmentPredicate( discriminatorPredicate ) );
			}
```

beikov (changes requested):

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
mbladel: This is not always true, as the table gorup might be created by a `LazyTableGroup` supplier like `ToOneAttributeMapping#createTableGroupInternal` which does not use the root table name. You probably don't remember but in my first iteration I also supposed the first join to always be the one with the root table, but our test suite confirmed that's often not the case.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
mbladel: I do not think this is an actual problem. From what I could see, `createCaseExpression` already resolves all needed table references through `registerTypeUsage(TableGroup tableGroup)`, so this should always work. I did a quick test and I believe my change here should be fine.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
mbladel: I tried this but I'm sure this would work, as we would need to also handle collections and we would need to differentiate between element / index descriptors. Any idea if there's an alternative (other than what I've already done that requires a map lookup)?

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
mbladel: > to filter only the correctly typed records

This is why it's necessary. I've expanded the comment a bit more but lmk if it's not clear enough.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
mbladel: I can't quite follow you on this, I did try ignoring `additionalPredicateCollectorAccess` when creating the table reference join but a test fails (`CustomSQLSecondaryTableTest`).

Not sure this would be a problem for joins though, wouldn't those create `LazyTableGroup`s?

beikov: Forget that, I misread `AbstractEntityPersister#createRootTableGroup`

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
beikov: I guess you'd have to do something like this:

```
final TableGroup elementTableGroup = joinedTableGroup instanceof PluralTableGroup 
  ? ( (PluralTableGroup) joinedTableGroup ).getElementTableGroup()
  joinedTableGroup;
final EntityPersister entityDescriptor = (EntityPersister) realTableGroup.getModelPart().getPartMappingType();
```

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
beikov: Ok, but can `tableReferenceJoins` be empty or i.e. reasonably not contain the root table?

mbladel: Well, if it's empty the `getPrimaryTableReference()` will always return the root table (I already use that in my code). it will also always contain the root table, but to find the correct join we have to check all `tableReferenceJoins` until we find it like I do in the for loop.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
mbladel: Couldn't we need to do this for the plural `getIndexTableGroup()` instead of the element though?

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
beikov: The code you pointed me to in `ToOneAttributeMapping#createTableGroupInternal` suggests that if we have a mapping like:

```java
@ManyToOne
Subclass association;
```

That the root table might not be part of the joins for a query e.g. `select 1 from EntityA a join a.association`.

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
beikov: A join on a `PluralJoin` is always against the element. A join against the index part will be a `PluralPartJoin` AFAIR.

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
mbladel: Well, if that's the case we wouldn't need the discriminator to be applied, since the subtype's table would be directly joined. So maybe I just need to remove the `assert applied`?

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
mbladel: I've applied your suggestion, slightly modified it to account for all cases.

hibernate-core/src/main/java/org/hibernate/persister/entity/JoinedSubclassEntityPersister.java
```diff
@@ -1376,6 +1386,29 @@ public void pruneForSubclasses(TableGroup tableGroup, Map<String, EntityNameUse>
...
+				boolean applied = applyDiscriminatorPredicate(
+						tableReferenceJoins.get( 0 ),
+						(NamedTableReference) tableGroup.getPrimaryTableReference(),
+						entityNameUses,
+						metamodel
+				);
+				for ( int i = 0; !applied && i < tableReferenceJoins.size(); i++ ) {
+					final TableReferenceJoin join = tableReferenceJoins.get( i );
+					applied = applyDiscriminatorPredicate( join, join.getJoinedTableReference(), entityNameUses, metamodel );
+				}
```
beikov, line 1409: I think I'd feel more comfortable if you add an assert at the end, and if it fails, we have to think about how to handle this scenario.

```suggestion
				}
				assert applied;
```

beikov (approved):

beikov (approved):

PR #6580, opened 2023-05-16
HHH-16611 : Sybase schema support in NameQualifierSupport

sebersole:
HHH-16611 : Sybase schema support in NameQualifierSupport


beikov:
It seems the jConnect JDBC driver doesn't support schemas so I figured that we can't say Sybase ASE "supports" `BOTH`. Try running the test suite with the jConnect driver and you will see some tests that fail due to `AbstractMethodError`. We can surely try working around that, though I wanted to share the troubles I had in the past with you.

sebersole:
Beyond annoying they still don't make their driver available.

sebersole:
Well, actually these tests are using jTDS still right?  And its failing there.

What's odd is that their documentation explicitly states they do support schemas and catalogs - though granted nothing specific about the JDBC driver, just what the database supports.

beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/7138

PR #6582, opened 2023-05-17
proposal for NativeQuery.setParameterEscapes()

gavinking:
ability to override the syntax for named/ordinal parameters in native SQL queries


beikov:
Is this still useful? Should we discuss it during the team meeting?

PR #6583, opened 2023-05-17
HHH-16570 Batch fetch with FetchMode JOIN might lead to multiplication in OneToMany items

dreab8:
https://hibernate.atlassian.net/browse/HHH-16570

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/loader/ast/internal/LoaderSelectBuilder.java
@@ -354,6 +354,14 @@ public class LoaderSelectBuilder {
 		this.entityGraphTraversalState = entityGraphTraversalState;
 		this.forceIdentifierSelection = forceIdentifierSelection;
 		this.jdbcParameterConsumer = jdbcParameterConsumer;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/test/java/org/hibernate/orm/test/cid/CompositeIdTest.java
```diff
@@ -156,8 +156,9 @@ public void testCompositeIds(SessionFactoryScope scope) {
 					statementInspector.clear();
 					Order o2 = new Order( c );
 					o2.setOrderDate( Calendar.getInstance() );
-					statementInspector.assertExecutedCount( 1 );
-					statementInspector.assertNumberOfOccurrenceInQuery( 0, "join", 1 );
+					statementInspector.assertExecutedCount( 2 );
```
beikov, line 159: If this change is required it would mean that the `index` mapping in `Customer.hbm.xml` is ignored. The `Customer#orders` collection should be a `LIST`, not a `BAG` and hence this change should not be necessary.

dreab8: i checked an i 5.6 we execute a separate query for the lineitems, anywway i'll double check

beikov (approved):

PR #6586, opened 2023-05-17
HHH-16626 - JPA hint for Session (EntityManager) level tenant-id

sebersole:
https://hibernate.atlassian.net/browse/HHH-16626


sebersole:
Applied upstream

PR #6597, opened 2023-05-19
HHH-16559 Batch fetch does not initialize bi-directional entity in HQL

mbladel:
https://hibernate.atlassian.net/browse/HHH-16559

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializerBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializerBuilder.java
@@ -86,6 +86,9 @@ public class EntitySelectFetchInitializerBuilder {
 		if ( !entityPersister.isBatchLoadable() || creationState.isScrollResult() ) {
 			return BatchMode.NONE;
 		}
+		else if ( creationState.isDynamicInstantiation() ) {
+			return BatchMode.BATCH_INITIALIZE;
+		}
 		while ( parentAccess.isEmbeddableInitializer() ) {
 			final EmbeddableInitializer embeddableInitializer = parentAccess.asEmbeddableInitializer();
 			final EmbeddableValuedModelPart initializedPart = embeddableInitializer.getInitializedPart();
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/batch/BatchFetchInstantiationTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/batch/BatchFetchInstantiationTest.java
@@ -1,0 +1,134 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package org.hibernate.orm.test.batch;
+
+import org.hibernate.cfg.AvailableSettings;
+
+import org.hibernate.testing.orm.junit.DomainModel;
+import org.hibernate.testing.orm.junit.Jira;
+import org.hibernate.testing.orm.junit.ServiceRegistry;
+import org.hibernate.testing.orm.junit.SessionFactory;
+import org.hibernate.testing.orm.junit.SessionFactoryScope;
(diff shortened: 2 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializerBuilder.java
```diff
@@ -83,7 +83,7 @@ private static BatchMode determineBatchMode(
 			EntityPersister entityPersister,
 			FetchParentAccess parentAccess,
 			AssemblerCreationState creationState) {
-		if ( !entityPersister.isBatchLoadable() || creationState.isScrollResult() ) {
+		if ( !entityPersister.isBatchLoadable() || creationState.disableBatching() ) {
```
beikov, line 86: Wouldn't it be better to use `BatchMode.BATCH_INITIALIZE` if there is a dynamic instantiation referring to the fetch owner (i.e. the `FetchParentAccess` root)?

mbladel: I did a quick test and `BATCH_INITIALIZE` would work in this very simple case; I did test and verify that in 5 we disabled batching for dynamic instantiation queries so I thought that doing the same would be safest.

Sadly we can't determine if the `FetchParentAccess` root is a dynamic instantiation because when the entity result initializer is created from a `DynamicInstantiationResult`, the latter is a first level `DomainResult` so the fetch parent access is always set to `null`.

beikov: Ok, then this is fine. Let's not overcomplicate this.

mbladel: Actually, I think `BATCH_INITIALIZE` is fine here, it will always eagerly load any associations and that's the same we did in 5. We also use it for cacheable entities, so I don't see why it wouldn't work here.

I'll also cache the check for dynamic instantiations so we don't have to repeat it.

beikov (approved):

hibernate-core/src/test/java/org/hibernate/orm/test/batch/BatchFetchInstantiationTest.java
```diff
@@ -0,0 +1,134 @@
...
+			final EntityA result = query.getSingleResult();
+			assertThat( result.getEntityB().getName() ).isEqualTo( "entity_b" );
+		} );
+	}
+
+	@Test
+	public void testDynamicInstantiation(SessionFactoryScope scope) {
+		scope.inTransaction( session -> {
+			final TypedQuery<MyPojo> query2 = session.createQuery(
+					String.format( "select new %s(t) from EntityA t where t.id = ?1", MyPojo.class.getName() ),
```
gavinking, line 74: This is not necessary anymore.

gavinking:
Folks, sorry to be the asshole here again but....

I don't see how the  _correct_ fix here would be to disable batch fetching.

I mean of course I could be missing something but naively I don't see how batch fetching is in any way incompatible with wrapping the entity in another class, at least not in principle, so this limitation doesn't seem well-motivated at all.

Indeed, I'm pretty sure that if you used my _implicit_ instantiation stuff [here](https://github.com/hibernate/hibernate-orm/pull/6685) you would find that it "just works".

Now, sure, I understand that, OK, H5 had this as a limitation, but that doesn't mean H6 should have it.

Wouldn't it be much better to actually _fix_ this arbitrary-looking limitation? Is it really that hard?

mbladel:
The limitation is there because the dynamic instantiation constructor could access properties that are yet not initialized when using `BatchMode.BATCH_LOAD`, for example associated entities, like in the reproducer for this issue.

Note that in the end we opted for *not* disabling batching, but we use `BatchMode.BATCH_INITIALIZE` which still uses batching but initializes the entity eagerly so we can make sure we don't pass an incomplete state to the dynamic instantiation constructor. This is the same behavior as when caching is enabled on the entity being batched.

Also I've quickly tried this on the *implicit* instantiation PR but with a query selecting an entity type I get a `QueryTypeMismatchException`, maybe I'm doing something wrong there.

PR #6606, opened 2023-05-20
HHH-16649 tests demoing that Hibernate.remove() doesn't work at all

gavinking:
See https://hibernate.atlassian.net/browse/HHH-16649.

CC: @jrenaat


jrenaat:
Indeed.
Afaict it's because `BasicCollectionPersister.doProcessQueuedOps`is not implemented (states 'nothing to do', which doesn't seem right?)

PR #6608, opened 2023-05-20
upgrade JDBC drivers

gavinking:


gavinking:
Damn, hit https://hibernate.atlassian.net/browse/HHH-15889?filter=21172 again. We _really_ should report that one!

gavinking:
@beikov I had to disable [one of your tests](https://hibernate.atlassian.net/browse/HHH-16317) on Oracle

beikov:
I reported the driver issue to @loiclefevre from Oracle. He suggests to upgrade the driver to 21.9.0.0 for now.

settings.gradle
```diff
@@ -183,25 +183,29 @@ dependencyResolutionManagement {
             version( "h2gis", h2gisVersion )
             version( "hsqldb", hsqldbVersion )
             version( "derby", derbyVersion )
-            version( "pgsql", "42.5.0" )
-            version( "mysql", "8.0.27" )
-            version( "oracle", "21.8.0.0" )
+            version( "pgsql", "42.6.0" )
+            version( "mysql", "8.0.33" )
+            version( "oracle", "23.2.0.0" )
+            version( "mariadb", "2.7.9" )
```
Sanne, line 189: Any specific reason to not have opted for 3.1.4 (latest of mariadb) ?
 - https://mariadb.com/kb/en/about-mariadb-connector-j/

Sanne:
> I reported the driver issue to @loiclefevre from Oracle. He suggests to upgrade the driver to 21.9.0.0 for now.

so.. should we do that? I wonder if we're not better off disabling that specific test on Oracle (it's broken after all, and apparently tracked appropriately), at least we'd be in a position to better detect other regressions, should they happen.

beikov:
I'm ok with disabling the test for now and re-enabling it with the next driver update. Just note that the JDBC driver has a bug, the test is fine.

beikov:
Created https://hibernate.atlassian.net/browse/HHH-16675 for this.

settings.gradle
gavinking: Yes because of https://hibernate.atlassian.net/browse/HHH-15889?filter=21172

jeandelavarene (commented):
A TIMESTAMP column doesn't include any timezone and hence can't be converted to an OffsetDateTime. This is why the latest driver throws an exception. The previous versions of the driver was assuming a timezone (UTC) for the TIMESTAMP. This was wrong and corrected through a bug fix.

PR #6610, opened 2023-05-20
HHH-16651, HHH-16659 introduce config property to enable subselect fetching

gavinking:


gavinking:
Alright, so for the next step of this work&mdash;which is going to be the really nice bit&mdash;I actually need to merge this with the work in #6617.

So, I think the best idea is to just merge all this, which is a really great cleanup and makes the whole machinery of `LoadQueryInfluencers` more powerful, as well as adding very nice features.

If there are specific changes I need to roll back, which is totally possible, or specific things you folks dislike, no problem, I'll roll them back by hand, or revert the individual commit if possible, no big deal.

So please folks, take the time to review what I've done here because it's quite significant.

PR #6614, opened 2023-05-22
introduce @View annotation

gavinking:
Thought on this, everyone?


gavinking:
> Looks ok to me, but how does hbm2ddl work with views?

WDYM? Isn't that the whole point of this change?

I know there's one missing thing here: I need to fix it to export views after tables, but that should also be straightforward I guess.

I mean, I *could* go to the whole effort of introducing a `View` object and a `StandardViewExporter` but it just doesn't seem worth the effort right now...

beikov:
Sorry, hit comment too fast. I meant to ask how the validation or update process works for views.

gavinking:
> Sorry, hit comment too fast. I meant to ask how the validation or update process works for views.

Ah good point. Maybe we should just skip em.

sebersole:
As we discussed in chat, we can at least perform some validation of views.  Update... probably nothing

gavinking:
Actually yeah, looking at what validator does, I think it should work perfectly fine for views.

PR #6615, opened 2023-05-22
HHH-16627 Duplicate records being created for many-to-many relation implemented as two one-to-many relations with cascading + HHH-16628 Two transactions within the same session cause unsaved transient instance error

dreab8:
https://hibernate.atlassian.net/browse/HHH-16627

https://hibernate.atlassian.net/browse/HHH-16628

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
+++ b/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
@@ -148,6 +148,9 @@ public class StatefulPersistenceContext implements PersistenceContext {
 	// May be empty or not contains all relation
 	private IdentityHashMap<Object,Object> parentsByChild;
 
+	// The key is the entity that has been merged and the value is the merged copy created by the merge process
+	private IdentityHashMap<Object,Object> mergedEntities;
+
 	private int cascading;
 	private int loadCounter;
 	private int removeOrphanBeforeUpdatesCounter;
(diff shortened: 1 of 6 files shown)


hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java
```diff
@@ -148,6 +149,8 @@ the following fields are used in all circumstances, and are not worth (or not su
 	// May be empty or not contains all relation
 	private IdentityHashMap<Object,Object> parentsByChild;
 
+	private IdentitySet<Object> mergedEntity;
```
sebersole, line 152: Minor - this should be pluralized: `mergedEntities`

sebersole (commented):
Not a huge fan of the need to add a new Map here, but I don't have a good alternative atm.

Shouldn't we clear `mergedEntities` after flushing?

dreab8:
Hi@sebersole,
 I found a problem that my solution does not fix,
given:

```
Student s1 = session.createQuery( "from Student where name = :name", Student.class )
							.setParameter( "name", STUDENT_NAME ).uniqueResult();

Course c1 = session.createQuery( "from Course where name = :name", Course.class )
							.setParameter( "name", COURSE_1_NAME ).uniqueResult();

Enrollment e1 = new Enrollment( 10 );

s1.addEnrollment( e1 );

c1.addEnrollment( e1 );

Student mergedStudent = session.merge( s1 );
Course mergedCourse = session.merge( c1 );
```

the second merge trigger the persist of a second `Enrollment` , working on this and I'll check also if we can clear the `mergedEntities` after `flush()`.

PR #6621, opened 2023-05-22
HHH-16651 support all fetch styles in fetch profiles

gavinking:
Finally, after all these years!


gavinking:
Note that with this PR, we can close https://hibernate.atlassian.net/browse/HHH-3414.

Yeah, **3141**, created August 5, 2008.

PR #6658, opened 2023-05-25
HHH-16385, HHH-16594, HHH-16617 Query cache key fixes regarding parameter and filters

mbladel:
This PR expands on the solution for [HHH-16385](https://hibernate.atlassian.net/browse/HHH-16385) proposed in https://github.com/hibernate/hibernate-orm/pull/6309, adding tests and fixes for [HHH-16594](https://hibernate.atlassian.net/browse/HHH-16594) and [HHH-16617](https://hibernate.atlassian.net/browse/HHH-16617) as well.

[HHH-16385]: https://hibernate.atlassian.net/browse/HHH-16385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HHH-16594]: https://hibernate.atlassian.net/browse/HHH-16594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HHH-16617]: https://hibernate.atlassian.net/browse/HHH-16617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingsImpl.java
```diff
@@ -167,30 +173,49 @@ public void visitBindings(@SuppressWarnings("rawtypes") BiConsumer action) {
...
+			for ( String paramName : filterDefinition.getParameterNames() ) {
+				final Object paramValue = filter.getParameter( paramName );
+				final JdbcMapping jdbcMapping = filterDefinition.getParameterJdbcMapping( paramName );
+				final MutableCacheKeyImpl cacheKey = new MutableCacheKeyImpl( 1 );
+				addBasicValueToCacheKey( cacheKey, paramValue, jdbcMapping, session );
+				cacheKeys.add( cacheKey );
+			}
+		}
+		// Sort the parameter cache key list using each key's hash code to ensure consistency
+		cacheKeys.sort( Comparator.comparingInt( k -> k.hashCode ) );
```
beikov, line 214: This shouldn't be necessary. We should only need a single `MutableCacheKeyImpl` like before. I think the problems you might have encountered are because we don't add anything to `MutableCacheKeyImpl` for `null` values, which is a problem in general.

mbladel: I think it is necessary, how can we be sure that the parameter values are always added to the cache key in the same order? The reproducer for HHH-16594 causes parameters to be in different orders each time the criteria query is executed.

hibernate-core/src/main/java/org/hibernate/engine/internal/CacheHelper.java
```diff
@@ -35,4 +41,28 @@ public static Object fromSharedCache(
 		return cachedValue;
 	}
 
+	public static void addBasicValueToCacheKey(
+			MutableCacheKeyBuilder cacheKey,
+			Object value,
+			JdbcMapping jdbcMapping,
+			SharedSessionContractImplementor session) {
+		if ( value == null ) {
+			return;
```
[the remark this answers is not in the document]
mbladel: You're right, I had copied it from `BasicValuedMapping`'s implementation but I guess that's fine for that too.

hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingsImpl.java
beikov: I think we have to make sure that `parameterBindingMap` is a `LinkedHashSet`, then we have a fully deterministic ordering which we would need for replicated/clustered caches anyway.

mbladel: Updated the solution, now we ensure the parameters are added to the cache key in the order they were processed through a custom `LinkedIdentityHashMap` and there's no need of manual sorting.

beikov (approved):

PR #6663, opened 2023-05-25
HHH-16699  promote repeat() function

gavinking:
I know I tested this before, but I forget how many dbs it fails on...


documentation/src/main/asciidoc/userguide/chapters/query/hql/QueryLanguage.adoc
```diff
@@ -892,6 +892,7 @@ Naturally, there are a good number of functions for working with strings.
 | `left()` | The leftmost characters of a string | `left(string, length)` | &cross; / &cross;
 | `right()` | The rightmost characters of a string | `right(string, length)`  | &cross; / &cross;
 | `replace()` | Replace every occurrence of a pattern in a string | `replace(string, pattern, replacement)` | &cross; / &cross;
+| `repeat()` | Concatenate a string with itself multiple times | `replace(string, times)` | &cross; / &cross;
```
andrei-ivanov, line 895: the 3rd column doesn't seem correct

PR #6679, opened 2023-05-28
HHH-16606 - Class cast exception when retrieving only single property of array type using criteria query

cigaly:
Jira issue [HHH-16606](https://hibernate.atlassian.net/browse/HHH-16606)

When only single property of array type is selected using criteria query with explictely specified (array) result type, returned result is wrapped into object array. Thi is caused by wrongly determined result transformer as RowTransformerArrayImpl instead of RowTransformerSingularReturnImpl.

[HHH-16606]: https://hibernate.atlassian.net/browse/HHH-16606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cigaly:
Not every array result type should be handled by RowTransformerArrayImpl.
If there is single selection of same type as result type, it should be handled by RowTransformerSingularReturnImpl.

beikov:
Thanks for your great work. I simplified your fix a bit as I believe the `RowTransformerArrayImpl` should only be used if the result type is `Object[]`.

Superseded by https://github.com/hibernate/hibernate-orm/pull/6729

PR #6681, opened 2023-05-29
HHH-16702-Added support for transaction management of tongweb application servers

shijianliang:
HHH-16702-Added support for transaction management of tongweb application servers


beikov:
If you want to add new features/improvements, please target the main branch. Old branches only get bug fixes.

PR #6683, opened 2023-05-29
HHH-16612 NPE on JDBC type code SQL types JSON in combination with explicit table name owner

cigaly:
See [Jira issue HHH-16612](https://hibernate.atlassian.net/browse/HHH-16612)


hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EmbeddableMappingTypeImpl.java
```diff
@@ -182,7 +182,7 @@ private EmbeddableMappingTypeImpl(
 				updatable = componentProperty.isUpdateable();
 			}
 			this.aggregateMapping = SelectableMappingImpl.from(
-					bootDescriptor.getOwner().getTable().getName(),
+					bootDescriptor.getOwner().getTable().getQualifiedTableName().render(),
```
mbladel, line 185:
```suggestion
					bootDescriptor.getOwner().getTable().getQualifiedName( creationContext.getSqlStringGenerationContext() ),
```
We need this in case default schema or default catalog are configured

hibernate-core/src/test/java/org/hibernate/orm/test/embeddable/JdbcTypeCodeJsonWithTableSchemaNameTest.java
```diff
@@ -0,0 +1,137 @@
...
+import jakarta.persistence.Table;
+import jakarta.persistence.Tuple;
+import jakarta.transaction.Transactional;
+
+import static org.hibernate.testing.transaction.TransactionUtil.doInJPA;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@RequiresDialect(PostgreSQLDialect.class)
+@JiraKey( "HHH-16612" )
+public class JdbcTypeCodeJsonWithTableSchemaNameTest extends BaseEntityManagerFunctionalTestCase {
```
mbladel, line 32: It would be nice if you could use the annotation test class style (`@SessionFactory`, `@ServiceRegistry` and `@DomainModel`) instead of extending the old base class. You also don't need the `@Transactional` annotation on test methods.

hibernate-core/src/main/java/org/hibernate/metamodel/mapping/internal/EmbeddableMappingTypeImpl.java
cigaly: Changed code

hibernate-core/src/test/java/org/hibernate/orm/test/embeddable/JdbcTypeCodeJsonWithTableSchemaNameTest.java
cigaly: Changed

mbladel:
@cigaly Thank you very much for the great work. I've been working on a very similar issue and I've included your test and fix in this PR: https://github.com/hibernate/hibernate-orm/pull/6739.

PR #6692, opened 2023-05-30
HHH-16574, HHH-16693 Fix join creation on treated and generic joined paths

mbladel:
https://hibernate.atlassian.net/browse/HHH-16574
https://hibernate.atlassian.net/browse/HHH-16693

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/hql/internal/QualifiedJoinPathConsumer.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/hql/internal/QualifiedJoinPathConsumer.java
@@ -21,6 +21,7 @@ import org.hibernate.query.sqm.tree.cte.SqmCteStatement;
 import org.hibernate.query.sqm.tree.domain.SqmCteRoot;
 import org.hibernate.query.sqm.tree.domain.SqmPath;
 import org.hibernate.query.sqm.tree.domain.SqmPolymorphicRootDescriptor;
+import org.hibernate.query.sqm.tree.domain.SqmTreatedPath;
 import org.hibernate.query.sqm.tree.from.SqmCteJoin;
 import org.hibernate.query.sqm.tree.from.SqmEntityJoin;
 import org.hibernate.query.sqm.tree.from.SqmFrom;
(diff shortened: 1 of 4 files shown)


hibernate-core/src/main/java/org/hibernate/query/hql/internal/QualifiedJoinPathConsumer.java
```diff
@@ -178,8 +179,11 @@ private static SqmFrom<?, Object> createJoin(
 			boolean isTerminal,
 			boolean allowReuse,
 			SqmCreationState creationState) {
+		final SqmPathSource<?> pathSource = lhs instanceof SqmTreatedPath ?
+				( (SqmTreatedPath<?, ?>) lhs ).getTreatTarget() :
+				lhs.getResolvedModel();
 		//noinspection unchecked
-		final SqmPathSource<Object> subPathSource = (SqmPathSource<Object>) lhs.getReferencedPathSource().getSubPathSource( name );
+		final SqmPathSource<Object> subPathSource = (SqmPathSource<Object>) pathSource.getSubPathSource( name );
```
mbladel, line 186: As an alternative to this check we could make `getReferencedPathSource()` return the `treatTarget` entity domain type for all treated `SqmJoin`s, like we already do for `SqmTreatedRoot`.

I've done some testing on this and found that, while it would require changing quite a bit of the treated SQM api classes, this should work too. Open to opinions on what is best.

beikov: >As an alternative to this check we could make getReferencedPathSource() return the treatTarget entity domain type for all treated SqmJoins, like we already do for SqmTreatedRoot.

That solution sounds more correct to me.

mbladel: I've created a separate PR with that https://github.com/hibernate/hibernate-orm/pull/6704.

beikov:
Closing this in favor of https://github.com/hibernate/hibernate-orm/pull/6704

PR #6701, opened 2023-05-31
HHH-16727 StatelessSession.upsert()

gavinking:
Here's a POC for the `upsert()` method of `StatelessSession`, but I really need @sebersole to take a look at how I've done this.


gavinking:
Merging this, with very prominent `@Incubating` annotation.

PR #6704, opened 2023-05-31
HHH-16574, HHH-16693 Fix join creation on treated and generic joined paths

mbellade:
Alternative solution to https://github.com/hibernate/hibernate-orm/pull/6692.

https://hibernate.atlassian.net/browse/HHH-16574
https://hibernate.atlassian.net/browse/HHH-16693

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/query/hql/internal/QualifiedJoinPathConsumer.java
+++ b/hibernate-core/src/main/java/org/hibernate/query/hql/internal/QualifiedJoinPathConsumer.java
@@ -178,8 +178,15 @@ public class QualifiedJoinPathConsumer implements DotIdentifierConsumer {
 			boolean isTerminal,
 			boolean allowReuse,
 			SqmCreationState creationState) {
+		final SqmPathSource<?> referencedPathSource = lhs.getReferencedPathSource();
+		// We need to use referencedPathSource when it is not generic since the getResolvedModel() method would
+		// return the association attribute as a path source and for treated paths that might correspond to a
+		// different entity type (usually the first in alphabetical order) and not the correct treat target
+		final SqmPathSource<?> pathSource = referencedPathSource.isGeneric() ?
+				lhs.getResolvedModel() :
+				referencedPathSource;
 		//noinspection unchecked
-		final SqmPathSource<Object> subPathSource = (SqmPathSource<Object>) lhs.getReferencedPathSource().getSubPathSource( name );
+		final SqmPathSource<Object> subPathSource = (SqmPathSource<Object>) pathSource.getSubPathSource( name );
 		if ( allowReuse && !isTerminal ) {
 			for ( SqmJoin<?, ?> sqmJoin : lhs.getSqmJoins() ) {
 				if ( sqmJoin.getAlias() == null && sqmJoin.getReferencedPathSource() == subPathSource ) {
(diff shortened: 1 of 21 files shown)


hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
```diff
@@ -3393,8 +3393,12 @@ private TableGroup consumeEntityJoin(SqmEntityJoin<?> sqmJoin, TableGroup lhsTab
 				null,
 				() -> p -> predicate.set( combinePredicates( predicate.get(), p ) ),
 				this
-				);
+		);
 		getFromClauseIndex().register( sqmJoin, tableGroup );
+		// For joins, we also need to register the table groups for the treats
```
beikov, line 3398: Might be worth extracting that to a method, because we would have to do the same for normal roots and cross joins, as well as do something similar for attribute joins.

hibernate-core/src/main/java/org/hibernate/query/hql/internal/QualifiedJoinPathConsumer.java
```diff
@@ -178,8 +178,12 @@ private static SqmFrom<?, Object> createJoin(
 			boolean isTerminal,
 			boolean allowReuse,
 			SqmCreationState creationState) {
+		final SqmPathSource<?> referencedPathSource = lhs.getReferencedPathSource();
+		final SqmPathSource<?> pathSource = referencedPathSource.isGeneric() ?
```
beikov, line 182: I'm a bit confused, shouldn't `lhs.getResolvedModel()` always produce the desired outcome?

mbladel: Not for treated joins, since `getResolvedModel()` for non-generic paths returns `getModel()` we would get the non-treated association attribute as a path source, which would report a wrong associated entity type (first one in alphabetical order).

mbladel: We could override this behavior in the `SqmTreatedPath` interface with a default `getResolvedModel()` implementation that returns the treat target entity type, but I wasn't sure if that is correct from the JPA-model side.

beikov: Ok, makes sense. Thanks

hibernate-core/src/main/java/org/hibernate/query/sqm/sql/BaseSqmToSqlAstConverter.java
mbladel: We already did this for attribute joins (I copied this from there), but we did not for from clause roots and cross joins. I extracted this to a method and added it for all cases you mentioned.

hibernate-core/src/main/java/org/hibernate/query/hql/internal/QualifiedJoinPathConsumer.java
mbladel: I've added a comment to clarify why we need to use the referenced path source, let's see if a `getResolvedNodeType()` or similar could be useful in other places in the future.

PR #6708, opened 2023-05-31
HHH-15929 Handle the possibility of different JdbcMappings for the same column

beikov:
DO NOT MERGE YET. Let @sebersole decide if we want to backport this.

https://hibernate.atlassian.net/browse/HHH-15929

Also fixes https://hibernate.atlassian.net/browse/HHH-16250 and https://hibernate.atlassian.net/browse/HHH-16490

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
+++ b/hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
@@ -759,7 +759,8 @@ public class SingleTableEntityPersister extends AbstractEntityPersister {
 					tableName,
 					() -> columnConsumer -> columnConsumer.accept(
 							tableName,
-							constraintOrderedKeyColumnNames[tablePosition]
+							constraintOrderedKeyColumnNames[tablePosition],
+							getIdentifierMapping()::getJdbcMapping
 					)
 			);
 		}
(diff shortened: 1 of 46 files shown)


hibernate-core/src/main/java/org/hibernate/persister/entity/SingleTableEntityPersister.java
```diff
@@ -759,7 +759,8 @@ public void visitConstraintOrderedTables(ConstraintOrderedTableConsumer consumer
 					tableName,
 					() -> columnConsumer -> columnConsumer.accept(
 							tableName,
-							constraintOrderedKeyColumnNames[tablePosition]
+							constraintOrderedKeyColumnNames[tablePosition],
+							getIdentifierMapping().getJdbcMappings()
```
Sanne, line 763: It's probably worth extracing these from the hot loop? I see `getIdentifierMapping()` is typically trivial, but `getJdbcMappings` is sometimes not.

Sanne: Also it's a quadratic loop :/ And I see several `getJdbcMappings` implementations creating defensive list copies.

It seems to call for a change in `getJdbcMappings()` to expose an immutable list by contract, so to avoid all such copies - regardless of this loop.

PR #6711, opened 2023-06-01
Migrating `Serializable[] getPropertySpaces()` to `String[] getPropertySpaces()`

Sanne:
Migrating `Serializable[] getPropertySpaces()` to `String[] getPropertySpaces()` : we seem to require these to be String[] and need to occasionally do some casting; might as well make it explicit?

Please let me know if this is considered an SPI breaking change, I could do it with a deprecation cycle and choosing a new methodname... I hope that's unnecessary in this case.


beikov (approved):
I guess this SPI change is fine for 6.3 i.e. main, though I'm not sure if we consider `EntityPersister`/`CollectionPersister` API or SPI. Wdyt @sebersole?

Sanne:
Conflicts resolved .. I'll merge soon?

PR #6714, opened 2023-06-02
Deprecate method getJdbcMappings() and prepare for its removal

Sanne:
A possibly better alternative of #6712 : 

In this case I deprecate the method but don't remove it yet; I still rework all client code to not use it so a subsequent removal will be trivial.

Benefit: could be backported.

Drawback: the deprecation of the method apparently has puzzling effect on javac's ability to handle multiple inheirtance gracefully, I had to re-implement the method even in classes which wouldn't be affected by the removal just to keep it from no longer compiling.


Sanne:
superseded by #6715 : both deprecating and then removing, so we can backport and cleanup.

PR #6717, opened 2023-06-02
@jrenaat HHH-16515 - Nullability work continued

jrenaat:


hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/JBossStandAloneJtaPlatform.java
```diff
@@ -27,6 +27,7 @@ public class JBossStandAloneJtaPlatform extends AbstractJtaPlatform {
 	private static final WildFlyStandAloneJtaPlatform wildflyBasedAlternative = new WildFlyStandAloneJtaPlatform();
 
 	@Override
+	@SuppressWarnings("nullness")
```
beikov, line 30: Use the helper class approach we talked about to reduce code duplication and handle nullness in that helper correctly.

hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/JBossStandAloneJtaPlatform.java
```diff
@@ -48,6 +49,7 @@ protected TransactionManager locateTransactionManager() {
 	}
 
 	@Override
+	@SuppressWarnings("nullness")
```
beikov, line 52: Use the helper class approach we talked about to reduce code duplication and handle nullness in that helper correctly.

hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/WebSphereLibertyJtaPlatform.java
```diff
@@ -58,16 +61,16 @@
 	}
 
 	public int getCurrentStatus() throws SystemException {
-		return retrieveTransactionManager().getStatus();
+		return NullnessUtil.castNonNull( retrieveTransactionManager() ).getStatus();
 	}
 
-	public Object getTransactionIdentifier(Transaction transaction) {
+	public @Nullable Object getTransactionIdentifier(Transaction transaction) {
```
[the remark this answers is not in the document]
jrenaat: Kinda odd that CodeQL complains about `getTransactionIdentifier`, but the missing @Override on  `canRegisterSynchronization`,  `getCurrentStatus`, `registerSynchronization` doesn't seem to bother it

hibernate-core/src/main/java/org/hibernate/engine/profile/FetchProfile.java
```diff
@@ -111,19 +111,6 @@ public void addFetch(final Fetch fetch) {
...
-				// we need to go back and ignore that previous bag join fetch.
-				if ( containsJoinFetchedBag ) {
-					// just for safety...
-					if ( fetches.remove( bagJoinFetch.getAssociation().getRole() ) != bagJoinFetch ) {
-						LOG.unableToRemoveBagJoinFetch();
-					}
-					bagJoinFetch = null;
-					containsJoinFetchedBag = false;
-				}
-
```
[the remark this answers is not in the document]
jrenaat: IIRC i removed this because `containsJoinFetchedBag` is never assigned, therefore always by default false, hence never going to enter that if anyway. Following the same reasoning `containsJoinFetchedBag` is also useless.

hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/WebSphereLibertyJtaPlatform.java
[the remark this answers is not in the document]
beikov: It tries to only report problems of lines that changed AFAIU, so that's why we are seeing this. It would be nice if you could fix that though.

hibernate-core/src/main/java/org/hibernate/engine/profile/FetchProfile.java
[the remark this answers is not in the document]
beikov: Please revert that, let's not add unrelated changes. The `isContainsJoinFetchedBag()` method where this field is used might be deprecated, but who know what is using this class.

hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/WebSphereLibertyJtaPlatform.java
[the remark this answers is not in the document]
jrenaat: With fix that I assume you reffered to adding @Override where it was missing? Or was that about CodeQL only reporting on what was changed?

hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/JBossStandAloneJtaPlatform.java
jrenaat: The main problem with this (for the CF) is the "invoke(null)" part; even if you check that the obtained Method reference is static (only case where you can pass in a null object reference), CF still complains.
So, I'm not seeing what you mean, basically.

beikov: I suppose you will need a stub file entry for `Method.invoke()` then and annotate `@Nullable`.

hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/AbstractJtaPlatform.java
```diff
@@ -31,12 +31,20 @@ public abstract class AbstractJtaPlatform
 	private boolean cacheUserTransaction;
 	private ServiceRegistryImplementor serviceRegistry;
 
-	private final JtaSynchronizationStrategy tmSynchronizationStrategy = new TransactionManagerBasedSynchronizationStrategy( this );
+	private final TransactionManagerAccessBridge transactionManagerAccess = new TransactionManagerAccessBridge();
+
+	private final JtaSynchronizationStrategy tmSynchronizationStrategy = new TransactionManagerBasedSynchronizationStrategy( transactionManagerAccess );
```
beikov, line 36: I'd suggest simply inlining the implementation of `TransactionManagerBasedSynchronizationStrategy` into `AbstractJtaPlatform`. It's all internal, so no worries about backwards compatibility.

hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/WebSphereJtaPlatform.java
```diff
@@ -46,7 +48,7 @@ public WebSphereJtaPlatform() {
 			throw new JtaPlatformException( "Could not locate WebSphere TransactionManager access class" );
 		}
 
-		this.transactionManagerAccessClass = tmAccessClass;
+		this.transactionManagerAccessClass = NullnessUtil.castNonNull( tmAccessClass );
```
beikov, line 51: I think it might be better to replace

`private final Class transactionManagerAccessClass;`

with

`private final Method getTransactionManagerMethod;`

and this here with `this.getTransactionManagerMethod = NullnessUtil.castNonNull( NullnessUtil.castNonNull( tmAccessClass ).getMethod( "getTransactionManager" ) )`

This way, you should be able to avoid the `@SuppressWarnings("nullness")` in `locateTransactionManager`

jrenaat: Removing that `@SuppressWarnings("nullness")` in locateTransactionManager doesn't need anything really, I just forgot to remove it, since it was combined with another warning I overlooked it.

hibernate-core/src/main/java/org/hibernate/engine/transaction/jta/platform/internal/AbstractJtaPlatform.java
jrenaat: I inlined it as an inner class, and I really prefer that over having the platform implementing the interface.

beikov (approved):

PR #6721, opened 2023-06-03
HHH-16605 Allow integer <=> boolean comparisons

ctabin:
Hello,

I'm trying to fix [HHH-16605](https://hibernate.atlassian.net/browse/HHH-16605) because it worked in version 6.1.7. Actually, it will still crash on H2 (but at the JDBC layer) **unless you use MODE=MYSQL** for the JDBC connection.

I'm not quite sure of the fix because it seems that it should be an option in the [H2Dialect](https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/dialect/H2Dialect.java) but it seems that there is no check on the dialect in [SqmCriteriaNodeBuilder](https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/query/sqm/internal/SqmCriteriaNodeBuilder.java#L242).

It seems that [this PR](https://github.com/hibernate/hibernate-orm/pull/6189) is causing the regression.

Best regards,
Cedric

[HHH-16605]: https://hibernate.atlassian.net/browse/HHH-16605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/IntegerJavaType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/IntegerJavaType.java
@@ -99,6 +99,8 @@ public class IntegerJavaType extends AbstractClassJavaType<Integer>
 			case "java.lang.Byte":
 			case "short":
 			case "java.lang.Short":
+			case "boolean":
+			case "java.lang.Boolean":
 				return true;
 			default:
 				return false;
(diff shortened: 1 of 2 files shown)


beikov:
See the comments on the issue. Will be done as part of https://hibernate.atlassian.net/browse/HHH-16769

PR #6724, opened 2023-06-04
HHH-16608 literal predicates can match against any type

ctabin:
https://hibernate.atlassian.net/browse/HHH-16608

After #6189 some queries against MariaDB/MySQL (and also on H2 with either mode) are broken because those engines accept predicates against values that are enclosed in literal, such as `SELECT * FROM table t WHERE t.some_date = '2023-01-01'` and hence that type cannot be refused by Hibernate.

Maybe that should be something configurable from the Dialect :thinking:


ctabin:
@beikov I just saw [your comment](https://github.com/hibernate/hibernate-orm/pull/6189#issuecomment-1453419092) about the migration of `p.dateField > '2023-01-01'` to `p.dateField > date 2023-01-01` but that would imply some huge migrations in some codebase.

Instead of simply reenabling the literal beahavoir as this PR, maybe it would be also interesting to add a new configuration property (in the `persistence.xml`) that allow to keep the old behavoir (eg comparing dates against string, but also on other values) ? :thinking:

beikov:
We decided that this kind of usage was not documented to be supported, but acknowledge the usefulness of it, so we created the improvement issue [HHH-16769](https://hibernate.atlassian.net/browse/HHH-16769) for this.

You're very welcome to contribute this general HQL improvement in the way how it is described in the JIRA issue. Join us on v if you want to discuss the design.

PR #6741, opened 2023-06-06
HHH-16740 + HHH-16750 Fetching an element collection on an entity with a composite id fails when batch size > 1

dreab8:
https://hibernate.atlassian.net/browse/HHH-16740
https://hibernate.atlassian.net/browse/HHH-16750


dreab8:
superseded by https://github.com/hibernate/hibernate-orm/pull/6758

PR #6744, opened 2023-06-06
HHH-16742 - Add test for issue

jrenaat:
Already fixed by https://github.com/hibernate/hibernate-orm/pull/6685


hibernate-core/src/test/java/org/hibernate/orm/test/jpa/query/TupleQueryTest.java
```diff
@@ -85,6 +91,28 @@ public void testGetAlias(EntityManagerFactoryScope scope) {
...
+
+					Tuple result = query.getResultList().get(0);
+					assertNotNull( result );
+					assertEquals( 2, result.getElements().size() );
+					try {
+						// Should not throw IllegalArgumentException
+						result.get("fn", String.class);
+						result.get("fn_2", String.class);
+					}
+					catch (IllegalArgumentException iae) {
```
beikov, line 109: Please remove the try-catch part. If an exception is thrown, the test will fail anyway.

beikov (approved):

PR #6747, opened 2023-06-07
Test non-standard access with extended bytecode enhancement

yrodiere:
I originally wrote this to compare the behavior with what we get in Quarkus (where part of the bytecode enhancement is done by Panache instead).

Turns out there is nothing wrong and these tests already pass, but we might as well merge this to get more non-regression tests?

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/bytecode/enhancement/basic/ExtendedEnhancementNonStandardAccessTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/bytecode/enhancement/basic/ExtendedEnhancementNonStandardAccessTest.java
@@ -1,0 +1,765 @@
+package org.hibernate.orm.test.bytecode.enhancement.basic;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.hibernate.Hibernate;
+
+import org.hibernate.testing.bytecode.enhancement.BytecodeEnhancerRunner;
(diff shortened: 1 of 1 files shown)


sebersole:
> i.e. when exposing fields through accessors with non-standard names

Hm, I'm surprised that works as it fails on the TCK and in my reproducer for those tests.

yrodiere:
> > i.e. when exposing fields through accessors with non-standard names
> 
> Hm, I'm surprised that works as it fails on the TCK and in my reproducer for those tests.

You need extended enhancement for that to work.

sebersole:
@scottmarlow - is Wild Fly using extended enhancement?

scottmarlow:
> @scottmarlow - is Wild Fly using extended enhancement?

Please remind me how extended enhancement is enabled for runtime enhancement?

sebersole:
You probably have an implementation of `org.hibernate.bytecode.enhance.spi.EnhancementContext`

scottmarlow:
> @scottmarlow - is Wild Fly using extended enhancement?

No, WildFly is not using extended enhancement.

scottmarlow:
> We could try that as an option in the TCK runs and see if it addresses the original failures

Would could WildFly do differently than org.hibernate.bytecode.enhance.spi.DefaultEnhancementContext that could help the TCK tests to pass?

yrodiere:
Tests are passing... I'll merge.

PR #6748, opened 2023-06-07
HHH-16761 Throw an appropriate error when an entity defines more @Id fields than its @IdClass

mbladel:
https://hibernate.atlassian.net/browse/HHH-16761


beikov (approved):

gavinking:
Nice.

PR #6750, opened 2023-06-08
HHH-16759 - Merge fails when entity has an Embedded Java record

cigaly:
Jira issue [HHH-16759](https://hibernate.atlassian.net/browse/HHH-16759)

[HHH-16759]: https://hibernate.atlassian.net/browse/HHH-16759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/type/ComponentType.java
+++ b/hibernate-core/src/main/java/org/hibernate/type/ComponentType.java
@@ -26,6 +26,7 @@ import org.hibernate.engine.spi.Mapping;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.generator.Generator;
+import org.hibernate.internal.util.ReflectHelper;
 import org.hibernate.internal.util.StringHelper;
 import org.hibernate.internal.util.collections.ArrayHelper;
 import org.hibernate.mapping.Component;
@@ -49,6 +50,7 @@ import org.hibernate.usertype.CompositeUserType;
  */
 public class ComponentType extends AbstractType implements CompositeTypeImplementor, ProcedureParameterExtractionAware {
 	private final Class<?> componentClass;
+	private final boolean mutable;
 
 	private final String[] propertyNames;
 	private final Type[] propertyTypes;
@@ -112,6 +114,7 @@ public class ComponentType extends AbstractType implements CompositeTypeImplemen
 		else {
 			this.compositeUserType = null;
 		}
+		this.mutable = !ReflectHelper.isRecord( componentClass ) && ( compositeUserType == null || compositeUserType.isMutable() );
 	}
 
 	private boolean isAggregate() {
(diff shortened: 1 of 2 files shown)


cigaly:
Passing all JDK 17 test with H2 and PostgreSQL

jrenaat:
Looks good, but there is a small type in the test's name, if you could correct that, please?

hibernate-core/src/main/java/org/hibernate/type/ComponentType.java
```diff
@@ -623,7 +630,7 @@ public CascadeStyle getCascadeStyle(int i) {
 
 	@Override
 	public boolean isMutable() {
-		return compositeUserType == null || compositeUserType.isMutable();
+		return (compositeUserType == null || compositeUserType.isMutable()) && !ReflectHelper.isRecord( componentClass );
```
beikov, line 633: IMO this should be stored in a field in `ComponentType`, as the `isRecord` check is not cheap.

cigaly: You are right, I will fix that.

cigaly:
@jrenaat 
> Looks good, but there is a small type in the test's name, if you could correct that, please?

I am not sure that I see what you are talking about.

hibernate-core/src/test/java17/org/hibernate/orm/test/records/MergeRecordPropetyTestCase.java
```diff
@@ -0,0 +1,75 @@
...
+import jakarta.persistence.Id;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@DomainModel(annotatedClasses = {
+		MergeRecordPropetyTestCase.MyEntity.class
+})
+@SessionFactory
+@JiraKey("HHH-16759")
+public class MergeRecordPropetyTestCase {
```
jrenaat, line 21: "MergeRecordPropetyTestCase" should be "MergeRecordPropertyTestCase" (it's one of those things that either you see it at once, or you will never see it - our brain plays this kind of tricks on us ;) )

cigaly: I see ... time to check my eyesight :-(

hibernate-core/src/main/java/org/hibernate/type/ComponentType.java
```diff
@@ -623,7 +632,7 @@ public CascadeStyle getCascadeStyle(int i) {
 
 	@Override
 	public boolean isMutable() {
-		return compositeUserType == null || compositeUserType.isMutable();
+		return componentClassIsNotRecord && ( compositeUserType == null || compositeUserType.isMutable() );
```
beikov, line 635:
```suggestion
		return mutable;
```

hibernate-core/src/main/java/org/hibernate/type/ComponentType.java
```diff
@@ -69,6 +71,7 @@ public ComponentType(Component component, int[] originalPropertyOrder, MetadataB
 		this.componentClass = component.isDynamic()
 				? Map.class
 				: component.getComponentClass();
+		this.componentClassIsNotRecord = !ReflectHelper.isRecord( componentClass );
```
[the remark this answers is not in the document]
cigaly: OK, but I will move this at the bottom of constructor, after compositeUserType is set.

PR #6756, opened 2023-06-08
[HHH-16772] Added a check method to `SetFieldOnArgument` setter bytecode generator to check if value type is 64-bit and adjust operand stack size accordingly.

aurimasniekis:
I found a bug on bytecode generated `HibernateAccessOptimizer` code, which when using types of 64-bit would cause `java.lang.VerifyError: Operand stack overflow` exception, duo to small operand stack size. I added logic to switch operand stack size to 3 if value type is 64-bit, and 2 when it's 32-bit value.

I was concerned that getter would be the same issue but apparently not and according to GPT:

> When returning a value for a long type, the maxstack is 2 because:
>
> 1. The first slot (0) in the stack is used to store the this reference - the object on which the method is being called.
> 2. The second and third slots (1 and 2) are used to store the long value. However, when the lreturn instruction is executed, it pops the long value from the stack to be returned, which actually only consumes one additional stack slot for the method invocation (similar to int).
> 
> The Java Virtual Machine (JVM) treats the stack as an array of stack slots, and the size of a long value is twice that of an int value (64-bit vs 32-bit). However, the JVM uses 32-bit stack slots. So, when returning a long value, only an additional slot is occupied by the long return value (not two slots) due to the value being popped by lreturn.
> 
> The return instruction for the long value consumes the additional slot occupied by the long return value, effectively leaving only two total stack slots used for the method invocation. That's why the maxstack for returning a long value is 2


hibernate-core/src/test/java/org/hibernate/bytecode/internal/bytebuddy/GenerateProxiesTest.java
```diff
@@ -43,6 +47,28 @@ public void generateFastClassAndReflectionOptimizer() {
 		assertNotNull( reflectionOptimizer.getInstantiationOptimizer().newInstance() );
 	}
 
+	@Test
+	public void generateFastMappedSuperclassAndReflectionOptimizer() {
```
beikov, line 51: Could you please add `@JiraKey("HHH-16772")` to the method?

beikov (approved):
Looks good to me. After adding the annotation I asked for, this can be merged.

PR #6772, opened 2023-06-12
HHH-16781 @JavaType annotation is ignored when basic type is already registered for same JdbcType/java class pair

cigaly:
See [Jira issue HHH-16781](https://hibernate.atlassian.net/browse/HHH-16781)


beikov:
Thanks for the test and attempt to fix the problem. I think though that the `BasicTypeRegistry#resolve` method simply does the wrong thing right now and needs to be fixed.

Superseded by https://github.com/hibernate/hibernate-orm/pull/7191

PR #6774, opened 2023-06-12
HHH-16782 Allow query plan caching of criteria queries

beikov:
https://hibernate.atlassian.net/browse/HHH-16782

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/jpa/HibernateHints.java
+++ b/hibernate-core/src/main/java/org/hibernate/jpa/HibernateHints.java
@@ -201,4 +201,26 @@ public interface HibernateHints {
 	 * @see jakarta.persistence.EntityManager#setProperty(String, Object)
 	 */
 	String HINT_JDBC_BATCH_SIZE = "org.hibernate.jdbcBatchSize";
+
+	/**
+	 * Hint to enable or disable the query plan caching.
+	 * <p>
+	 * By default, query plan caching is enabled for HQL queries
+	 * and immutable criteria queries i.e. created with {@link org.hibernate.cfg.AvailableSettings#CRITERIA_COPY_TREE}.
+	 * Query plan caching can be disabled for any query by setting this property to {@code false}.
+	 * Query plan caching can be enabled for mutable criteria queries by setting this property to {@code true}.
+	 * <p>
+	 * Setting this property to {@code true} for mutable criteria queries can lead to cache trashing,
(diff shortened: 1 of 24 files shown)


hibernate-core/src/main/java/org/hibernate/jpa/HibernateHints.java
```diff
@@ -201,4 +201,18 @@ public interface HibernateHints {
...
+	 * <p>
+	 * A value of {@code true} enables follow-on-locking, whereas a value
+	 * of {@code false} disables it. If the value is {@code null}, the
+	 * dialect itself will determine whether follow-on locking is used.
+	 *
+	 * @see org.hibernate.query.SelectionQuery#setQueryPlanCacheable
+	 *
+	 * @since 6.3
+	 */
+	String HINT_QUERY_PLAN_CACHEABLE = "hibernate.query.plan.cacheable";
```
dreab8, line 217: it seems that the javadoc is wrong

hibernate-core/src/main/java/org/hibernate/query/sqm/internal/QuerySqmImpl.java
```diff
@@ -633,6 +636,14 @@ protected ScrollableResultsImplementor doScroll(ScrollMode scrollMode) {
 	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 	// Select query plan
 
+	@Override
+	public boolean isQueryPlanCacheable() {
+		return CRITERIA_HQL_STRING.equals( hql )
+				// For criteria queries, query plan caching requires an explicit opt-in
+				? getQueryOptions().getQueryPlanCachingEnabled() == Boolean.TRUE
```
dreab8, line 643: what about a warn about the risk of enabling it for non immutable criteria.

beikov: Can you elaborate a bit? Why would we do that? Do you mean the warning is appropriate because the caching of such query plans could trash the cache? To me, it kind of feels unnecessary to emit a warning, because people will have to opt-in to this anyway. I can write a warning into the Javadoc or explain how the cache trashing could happen in the `@implNote` of the setting.

dreab8: explaining this in the the javadoc is perfect

PR #6777, opened 2023-06-12
HHH-16744 Wrong class created when reusing a join column for different entities in multiple ManyToOne

dreab8:
https://hibernate.atlassian.net/browse/HHH-16744


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/BatchInitializeEntitySelectFetchInitializer.java
```diff
@@ -45,17 +45,14 @@ protected void registerResolutionListener() {
 	}
 
 	@Override
-	public void resolveKey(RowProcessingState rowProcessingState) {
+	public void resolveInstance(RowProcessingState rowProcessingState) {
 		if ( state != State.UNINITIALIZED ) {
```
beikov, line 49: Please remove this as it is handled in `resolveKey` now apparently.

hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/AbstractBatchEntitySelectFetchInitializer.java
```diff
@@ -82,10 +70,37 @@ public void initializeInstance(RowProcessingState rowProcessingState) {
...
 
+	protected void resolveKey(
+			RowProcessingState rowProcessingState,
+			ToOneAttributeMapping referencedModelPart,
+			FetchParentAccess parentAccess) {
+		if ( state != State.UNINITIALIZED ) {
+			return;
+		}
+		if ( !isAttributeAssignableToConcreteDescriptor( parentAccess, referencedModelPart ) ) {
+			return;
```
beikov, line 81: I don't know how often this gets called, but I think it might be beneficial to set `state = State.MISSING;` before returning.

PR #6780, opened 2023-06-13
HHH-16749 Reuse BasicType from Value for JPA Metamodel instead of creating dedicated BasicDomainType instances

beikov:
https://hibernate.atlassian.net/browse/HHH-16749


sebersole (commented):
lgtm

PR #6781, opened 2023-06-13
Test against JDK 22 EA

yrodiere:


yrodiere:
Converting to draft as it seems Gradle 7 is not able to handle JDK 22 for some reason, even in toolchains.
See https://github.com/hibernate/hibernate-reactive/pull/1669

yrodiere:
Rebased on main. Now that we're on Gradle 8, hopefully the build will pass on JDK 22 as well.

PR #6783, opened 2023-06-13
Parameterize Lock tests to additionally run with batch fetching enabled

marko-bekhta:
When the way how `SingleIdEntityLoader` is created got changed here to use a `BatchLoaderFactory` service:
https://github.com/hibernate/hibernate-orm/commit/25a23fd1c0462015697d1d13eab2ff4eae813ce5#diff-8404725a304ff77d25bf88d040cc8285d82fe35d4381a7fe684d7c26624281fbR1023-R1025

It caused a problem with the lock mode not being applied if batch fetching is enabled (`hibernate.default_batch_fetch_size>1`) 
But it got already fixed by changes made in HHH-16651. But since those changes are only in the main branch, how should we handle this for 6.2 (quarkus is affected by this problem)? 

This PR makes the lock test parameterized so that we cover cases when batch fetching is enabled.

### DIFF
--- a/hibernate-core/src/test/java/org/hibernate/orm/test/jpa/lock/LockTest.java
+++ b/hibernate-core/src/test/java/org/hibernate/orm/test/jpa/lock/LockTest.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.orm.test.jpa.lock;
 
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


dreab8:
this [PR](https://github.com/hibernate/hibernate-orm/pull/6824) tries to fix the issue

beikov:
Superseded by https://github.com/hibernate/hibernate-orm/pull/6824

PR #6784, opened 2023-06-13
HHH-16773 Introduce a Dialect flag for databases that support group by / order by functional dependency

mbladel:
https://hibernate.atlassian.net/browse/HHH-16773

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
+++ b/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
@@ -5299,4 +5299,12 @@ public abstract class Dialect implements ConversionContext, TypeContributor, Fun
 	public DmlTargetColumnQualifierSupport getDmlTargetColumnQualifierSupport() {
 		return DmlTargetColumnQualifierSupport.NONE;
 	}
+
+	/**
+	 * Get this dialect's level of support for primary key functional dependency analysis
+	 * within {@code GROUP BY} and {@code ORDER BY} clauses.
+	 */
+	public FunctionalDependencyAnalysisSupport getFunctionalDependencyAnalysisSupport() {
+		return FunctionalDependencyAnalysisSupportImpl.NONE;
+	}
 }
--- a/hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/CockroachLegacyDialect.java
+++ b/hibernate-community-dialects/src/main/java/org/hibernate/community/dialect/CockroachLegacyDialect.java
@@ -19,26 +19,13 @@ import java.util.TimeZone;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import jakarta.persistence.TemporalType;
-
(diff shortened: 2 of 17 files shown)


hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
```diff
@@ -5299,4 +5299,12 @@ public String getRowIdColumnString(String rowId) {
 	public DmlTargetColumnQualifierSupport getDmlTargetColumnQualifierSupport() {
 		return DmlTargetColumnQualifierSupport.NONE;
 	}
+
+	/**
+	 * Indicates whether this dialect supports primary key functional dependency within
+	 * {@code GROUP BY} and {@code ORDER BY} clauses for the indicated entity mapping type.
+	 */
+	public boolean supportsPKFunctionalDependency(EntityMappingType entityMappingType) {
```
gavinking, line 5307: I know this is completely anal by can we please avoid stuff like "PK" and "FK" in method names?

mbladel: No problem at all, wdyt about `supportsGroupAndOrderByFunctionalDependency`?

beikov: How about letting the Dialect return a `FunctionalDependencyAnalysisSupport` enum instead and offer the following values?

* `NONE` - No support for the analysis
* `TABLE_REFERENCE` - Only supports the analysis for a single table reference, i.e. no support for inner joined tables
* `TABLE_REFERENCE_AND_CONSTANTS` - like `TABLE_REFERENCE` but also supports constants (for table per class strategy)
* `TABLE_GROUP` - Full support for the analysis

Alternatively, another option could be to make `FunctionalDependencyAnalysisSupport` an interface that has some _supports_-methods like `supportsAnalysis()`, `supportsConstants()`, `supportsInnerJoinedTables()` etc.

mbladel: I've applied @beikov's first suggestion, returning an enum (with slightly different values) and moving the logic to the entity valued path interpretation. I'm open to changing it again, LMK what you prefer.

beikov: After giving this some more thought, I think the more future proof approach would be to turn the enum into an interface and the constants to implementations of that interface.
In case we have to change something or add checks, the interface approach is more flexible IMO.

mbladel: Allright, I've converted it to an interface and updated the checks.

beikov (approved):

hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java
gavinking: Thanks 🙏

PR #6789, opened 2023-06-14
HHH-16743 StackOverflowError when loading a ManyToOne whith @Proxy(lazy=false)

dreab8:
https://hibernate.atlassian.net/browse/HHH-16743

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializerBuilder.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializerBuilder.java
@@ -105,7 +105,10 @@ public class EntitySelectFetchInitializerBuilder {
 			return NONE;
 		}
 		else if ( creationState.isDynamicInstantiation() ) {
-			return BatchMode.BATCH_INITIALIZE;
+			if ( canBatchInitializeBeUsed( entityPersister ) ) {
+				return BatchMode.BATCH_INITIALIZE;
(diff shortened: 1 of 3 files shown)


hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntitySelectFetchInitializerBuilder.java
```diff
@@ -134,6 +134,11 @@ else if ( creationState.isDynamicInstantiation() ) {
 			if ( cacheAccess != null ) {
 				// Do batch initialization instead of batch loading if the parent entity is cacheable
 				// to avoid putting entity state into the cache at a point when the association is not yet set
+				if ( entityPersister.getRepresentationStrategy()
```
beikov, line 137: Should this rather be part of the first if condition in this method?

dreab8: not really, it fine if it is a BATCH_LOAD but probably we have to do this check also at line 108 and 123 where we may return BATCH_INITIALIZE.

dreab8: @beikov what do you think of the new change?

beikov: LGTM

PR #6793, opened 2023-06-14
HHH-16678 AssertionError in QueryLiteral.<init> when using "update versioned Entity ..." if @Version is a long

mbladel:
Same as https://github.com/hibernate/hibernate-orm/pull/6792, in main there was no error because the assertion in `QueryLiteral` was removed but changing it for consistency.


beikov (approved):

beikov:
Could you please try re-adding the assertion? I'd like to know if it was removed "just because", or if there is actually a test that fails with the assertion.

beikov (approved):

PR #6796, opened 2023-06-14
HHH-16589 Limit in clause padding to Dialect.getInExpressionCountLimit and fix negated in predicate splitting

beikov:
https://hibernate.atlassian.net/browse/HHH-16589

### DIFF
--- a/hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
+++ b/hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java
@@ -6839,13 +6839,6 @@ public abstract class AbstractSqlAstTranslator<T extends JdbcOperation> implemen
 			}
 		}
 
-		inListPredicate.getTestExpression().accept( this );
-		if ( inListPredicate.isNegated() ) {
-			appendSql( " not" );
-		}
-		appendSql( " in (" );
-		String separator = NO_SEPARATOR;
-
 		int bindValueCount = listExpressions.size();
 		int bindValueMaxCount = bindValueCount;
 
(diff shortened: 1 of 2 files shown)


Adrodoc:
Would it be possible to change the algorithm for calculating the in-clause size to the one implemented in #6555? The current algorithm makes little sense in my opinion.

Consider a expression count limit of 1000. The current algorithm will use four different statements for 16000, 16001, 16002 and 16003 in clause parameters, because the power of two padding is only applied to the remainder. This is extremely inefficient use of the statement cache of the database. But the whole point of padding is to use the statement cache efficiently. Compare this to the power of two padding applied for a dialect without an expression count limit: All four expression counts would use the same statement with 16384 bind variables.
The algorithm implemented in #6555 achieves efficient use of the statement cache by applying a power of two padding up to the expression count limit and then continues with a power of two times expression count limit padding.

Here is a table comparing the growth of bind variables between the current algorithm and the proposed algorithm for an expression limit of 100. You can see that the proposed algorithm follows an exponential curve, just like the padding for dialects without an expression limit, so it requires only a logarithmic amount of cached statements, whereas the current algorithm follows a linear curve and thus requires a linear amount of cached statements.

| current | proposed | power of two (without expression limit) |
|-|-|-|
|1|1|1|
|2|2|2|
|4|4|4|
|8|8|8|
|16|16|16|
|32|32|32|
|64|64|64|
|100|100|128|
|101|200|256|
|102|400|512|
|104|800|1024|
|108|1600|2048|
|116|3200|4096|
|132|6400|8192|
|164|12800|16384|
|200|25600|32768|

beikov:
You can create an improvement request for this in JIRA and start a discussion on our chat platform Zulip about this, or since you have a concrete proposal, you can even start a [GitHub discussion](https://github.com/hibernate/hibernate-orm/discussions). This JIRA is about fixing a bug though, so let's limit it to exactly that.
Please don't forget that a number of 100 is unrealistic. Take a look into the dialects to see what limits real databases actually have - usually something around 1000+.

Adrodoc:
> You can create an improvement request for this in JIRA and start a discussion on our chat platform Zulip about this, or since you have a concrete proposal, you can even start a [GitHub discussion](https://github.com/hibernate/hibernate-orm/discussions). This JIRA is about fixing a bug though, so let's limit it to exactly that. Please don't forget that a number of 100 is unrealistic. Take a look into the dialects to see what limits real databases actually have - usually something around 1000+.

I chose 100 to make the table not be super long in hopes that this would illustrate the difference better. I know that 1000 is a more realistic limit, but it doesn't really change anything.
Sure if you want this PR to focus on the bug I have nothing against that. I am glad if it gets fixed so that I can upgrade the applications in the company I am working for :)

beikov:
It's documented here: https://docs.jboss.org/hibernate/orm/6.2/userguide/html_single/Hibernate_User_Guide.html#:~:text=hibernate.query.in_clause_parameter_padding%20(e.g.%20true%20or%20false%20(default%20value)) but doesn't go into how it behaves when splitting the in-list. Either way, such an improvement should not be done in a bug fix release.

Adrodoc:
@beikov I opened [HHH-16826](https://hibernate.atlassian.net/browse/HHH-16826) for the algorithm change. Do you think I should still write something in Zulip or GitHub Discussions? I don't quite know how the Hibernate team handles improvements like this.

beikov:
@Adrodoc I'll discuss this during our meeting tomorrow and will reply to you through a comment on the JIRA issue about what we decided/discussed.

PR #6798, opened 2023-06-14
HHH-16799 Backport 6.2: Don't fall back to field access type too early

marko-bekhta:
https://hibernate.atlassian.net/browse/HHH-16799

While the test case itself may look a bit silly, it was as close as I got to get to the same state as in the original issue caused by using Kotlin in Quarkus 🙈 (https://github.com/quarkusio/quarkus/issues/33740)

If the fix isn't ok -- let me know 😄


marko-bekhta:
hey hey, @dreab8 just pinging you to get this PR on your radar, maybe there'll be time to get it in for the next release.

